PR #1967, opened 2018-03-22
checkout exempted from csrf

### LINKED ISSUE #160: Collections are returned even if empty
Django saves CharField as empty string instead of null. This causes get_products_from_collection method to return objects, which don't have collection specified, as it only checks for null

szymi-:
I want to merge this change because it fixes #160. This behaviour (exemption from csrf check) is consistent with behaviour of django-payments, where some other views are also exempt.

### DIFF
--- a/saleor/order/views.py
+++ b/saleor/order/views.py
@@ -8,6 +8,7 @@ from django.http import Http404, HttpResponseForbidden
 from django.shortcuts import get_object_or_404, redirect
 from django.template.response import TemplateResponse
 from django.utils.translation import pgettext_lazy
+from django.views.decorators.csrf import csrf_exempt
 from payments import PaymentStatus, RedirectNeeded
 
 from . import FulfillmentStatus, OrderStatus
@@ -141,6 +142,7 @@ def cancel_payment(request, order):
     return HttpResponseForbidden()
 
 
+@csrf_exempt
 def checkout_success(request, token):
     """Redirect user after placing an order.
 


elwoodxblues:
Thanks for your contribution @szymi-! I'm not the original maintainer of `django-payments` so it's hard for me to say anything about those `csrf_exempts` in there and if using them isn't actually a security issue.
As for the change you've made -`checkout_success` serves as success callback for payment gateways, but it also does some logic like saving a registration form and assigning user to a order. I think such logic should remain protected by CSRF mechanism. Maybe this logic should be moved to a separate view and we should have one that will only serve as payment callback, so it can safely process also POST requests send by payment gateway, as it happens in case of Dotpay.

szymi-:
Thanks for your comments @elwoodxblues. Separating these two views sounds like a reasonable thing to do, I'll give it a shot when I get a chance.

szymi-:
I played a little with the payment flow. After placing an order and paying using Dotpay, I get redirected to an url, say:

    https://<HOST>/order/ce50ea32-50e0-4754-9316-23bf90d05e2a/checkout-success/?status=OK

I think the problem is that, as specified in their current docs, Dotpay sends a POST request to this url instead of just GET, with status=OK in the request body. When I performed a purchase without logging in, using the same email address as a registered user has, apart from just the "Thank you for your order" message, I got a login form which said "It looks like you already are our customer! Please log in to connect order with your account.", which is the form you mention. When I looked at the form in page source, I can see the csrf token is there, so when the user decides to log in directly from this page, they will log in successfully. So the csrf_exempt decorator only means that you can send POST request to this particular URL without token, however any forms in this page will have csrf token set.

I did not find any forms using the url /order/<uuid>/ as "action", so as far as I can tell using this @csrf_exempt decorator in this view should not be a problem. Please let me know what you think.

elwoodxblues:
Hey @szymi-, I've discussed the problem with rest of the team and we agreed that the correct solution would be to split the views and move our registration-related logic to a separate view. I'll open an issue for that.

szymi-:
Okay, thanks.

PR #1973, opened 2018-03-23
Allow to create attribute values from product form

elwoodxblues:
This PR is follow-up to #1964. Just as with variant attributes, it introduces the ability to create new attribute values from the product form. 

### Pull Request Checklist

### DIFF
--- a/saleor/dashboard/product/forms.py
+++ b/saleor/dashboard/product/forms.py
@@ -137,7 +137,53 @@ class ProductTypeForm(forms.ModelForm):
         return data
 
 
-class ProductForm(forms.ModelForm):
+class AttributesMixin(object):
+    """Form mixin that dynamically adds attribute fields."""
+
+    available_attributes = ProductAttribute.objects.none()
+
+    # Name of a field in self.instance that hold attributes HStore
+    model_attributes_field = None
+
+    def __init__(self, *args, **kwargs):
+        if not self.model_attributes_field:
+            raise Exception(
+                'model_attributes_field must be set in subclasses of '
+                'AttributesMixin.')
+
+    def prepare_fields_for_attributes(self):
+        initial_attrs = getattr(self.instance, self.model_attributes_field)
+        for attribute in self.available_attributes:
+            field_defaults = {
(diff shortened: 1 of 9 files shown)


saleor/dashboard/product/forms.py
```diff
@@ -209,10 +209,16 @@ def save(self, commit=True):
...
         for attr in self.product_attributes:
             value = self.cleaned_data.pop(attr.get_formfield_name())
-            if isinstance(value, AttributeChoiceValue):
-                attributes[smart_text(attr.pk)] = smart_text(value.pk)
-            else:
-                attributes[smart_text(attr.pk)] = value
+
+            # if the passed attribute value is a string,
+            # create the attribute value.
+            if not isinstance(value, AttributeChoiceValue):
```
akjanik: This block is the same as in `VariantAttributeForm`, maybe we can make a function from it?

elwoodxblues:
@akjanik You were right about duplicated code, so I've extracted a form mixin that handles dynamic attributes fields in the same way in both product and variant form. I think the code looks a bit better now.

Also, I've taken the opportunity and dropped `get_attribute` and `set_attribute` functions from models, that were used mostly in tests. There is definitely too many confusing utility functions related to attributes, that we should eventually organize.

saleor/dashboard/product/forms.py
```diff
@@ -137,7 +137,53 @@ def clean(self):
...
+            if attribute.has_values():
+                field = ModelChoiceOrCreationField(
+                    queryset=attribute.values.all(), **field_defaults)
+            else:
+                field = forms.CharField(**field_defaults)
+            self.fields[attribute.get_formfield_name()] = field
+
+    def iter_attribute_fields(self):
+        for attr in self.available_attributes:
+            yield self[attr.get_formfield_name()]
```
mad-anne: Shouldn't it be rather `self[attr.get_formfield_name()]`?

elwoodxblues: You can access fields in this way as well. It was here before so I didn't change it.

PR #1980, opened 2018-03-26
Rewrite page forms

dominik-zeglen:
Refactor page forms to match `Form` component pattern

### DIFF
--- a/saleor/static/dashboard-next/page/views/PageUpdateForm.tsx
+++ b/saleor/static/dashboard-next/page/views/PageUpdateForm.tsx
@@ -7,9 +7,12 @@ import * as React from "react";
 import { Redirect } from "react-router";
 
 import { DialogContentText } from "material-ui/Dialog";
+import { pageListUrl } from "..";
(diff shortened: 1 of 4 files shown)


saleor/static/dashboard-next/page/views/PageUpdateForm.tsx
```diff
@@ -70,90 +77,126 @@ export class PageUpdateForm extends React.Component<
...
+                { called, data: deleteResult, error, loading: deleteInProgress }
               ) => {
                 if (called && !deleteInProgress) {
                   if (error) {
                     console.error(error);
                     return;
                   }
-                  if (data && data.pageDelete.errors.length) {
+                  if (
+                    deleteResult &&
```
patrys: Do we need to check if the result exists if we know that the mutation was called, finished loading and did not yield errors?

dominik-zeglen: There is wisdom in those words 🤔

PR #1996, opened 2018-03-28
Drop Stock model

### LINKED ISSUE #1994: Drop Stock model
We should simplify our inventory model that currently assumes that a variant can have multiple stock records. It's a specific case that could be implemented as a customization by those who actually need it. In our case, it would be much simpler (in terms of UI and the code) while still covering most use cases, if we moved quantity tracking to the ProductVariant model.

As a result, ProductVariant would have additional fields:
- `quantity`
- `quantity_allocated`
- `cost_price`

Pacu2:
This PR drops Stock Model and moves related fields to ProductVariant instead.
close #1994 
Variant details view:
variant_dashboard_details
Variant form view:
variant_example_v2

### Pull Request Checklist


saleor/dashboard/order/forms.py
```diff
@@ -152,25 +151,25 @@ def __init__(self, *args, **kwargs):
     def clean_quantity(self):
         quantity = self.cleaned_data['quantity']
         delta = quantity - self.initial_quantity
-        stock = self.instance.stock
-        if stock and delta > stock.quantity_available:
+        variant = self.instance.variant
```
NyanKiyoshi: I'm not sure why instead of putting a column ["product" that is referencing the table "Product"](https://github.com/mirumee/saleor/blob/master/saleor/order/models.py#L160-L164), we don't put a column "variant" referencing "ProductVariant" (also see https://github.com/mirumee/saleor/blob/master/saleor/order/utils.py#L124-L127, where we are creating order lines from variants).

templates/dashboard/product/product_variant/form.html
```diff
@@ -76,10 +76,20 @@
                 <div class="row">
                   {{ form.name|materializecss }}
                 </div>
+                <p class="card-title">{% trans "Stock" %}</p>
```
NyanKiyoshi: What about:
- moving this section under the `SKU` field,
- renaming the label "quantity" to "stock quantity",
- and renaming the label "price override" to "selling price"?

saleor/dashboard/order/forms.py
Pacu2: It will be done in this PR as well, just taking it down one by one

templates/dashboard/product/product_variant/form.html
Pacu2: Moving stock to the right doesn't look good when variant has no attributes, will think about this UI further later

NyanKiyoshi: Okay then, sounds fair!

> - renaming the label "quantity" to "stock quantity",
> - and renaming the label "price override" to "selling price override" (and thus "Selling price on base product form label)

I think those two could be great to add as they would be more descriptive.

patrys: @Pacu2 I don't think we want to support variants with no attributes.

Pacu2: Take a look at the new screenshot

patrys: For new screenshot:

* I don't think "number in stock" belongs in the wide section as it we are unlikely to deal with stock in excess of ten digits 😉 
* sale price and cost price are different width
* sale price has double height which looks kind of odd (in other forms only the first line of the form behaves this way)

saleor/order/utils.py
```diff
@@ -113,27 +113,25 @@ def add_variant_to_order(
         order, variant, total_quantity) if add_to_existing else total_quantity
     price = variant.get_price_per_item(discounts)
     while quantity_left > 0:
```
elwoodxblues: This `while` loop was useful when we had multiple stocks that we had to iterate over to check quantity in each of them. Do we still need it? Maybe this function could be simplified a bit.

templates/dashboard/product/product_variant/form.html
Pacu2: Screenshot updated according to @patrys comments

saleor/core/utils/random_data.py
```diff
@@ -304,23 +303,14 @@ def create_product(**kwargs):
...
-    return Stock.objects.create(**defaults)
-
-
 def create_variant(product, **kwargs):
     defaults = {
-        'product': product}
+        'product': product,
+        'quantity': fake.random_int(1, 50),
+        'cost_price': fake.money(),
+        'quantity_allocated': fake.random_int(1, 50)}
```
mad-anne: Shouldn't `quantity_allocated` be always less than `quantity`?

NyanKiyoshi:
Hello! After some discussions with the people I'm working with, it was decided and concluded that they really need the stock location to be kept.

But, by stock location, we hear that we have: a variant (or a product that has no variants) that has:
- a stock count,
- a stock location,
- a cost price,
- a selling price.

My first thought is this that:
- we could and can keep this those stock location and still drop the multiple stock mechanism, which is the goal this PR,
- keep the a stock location model (`StockLocation {pk, name, variants}`),
- then, apply a migration to the model StockLocation to move current variants stocks into variants, and,
- apply the stock location pk to the variant instance, as well during the migration process.

To summarize:

```
StockLocation
=============
name: str


ProductVariant
=============
stock_count: int
stock_allocated: int
stock_cost: Money
stock_location: ForeignKey[StockLocation]
```

I do I agree on the fact that we should keep the location name and not drop it, and think it may be really unwanted by some people.

Pacu2:
@NyanKiyoshi Could you please explain me the case?
What's the workflow(situation) when one does not need multiple stocks, but needs to keep StockLocation, to separate one stock from the other?

I don't think that's a common case, from shops I've worked with you either have multiple stocks or you don't. If you don't, why is the StockLocation needed?

NyanKiyoshi:
Seems fair... Especially as SKU should handle this as well. I will discuss tomorrow with the team and keep you updated.

To at least explain, we have a quite a set of products. Each of them are from a given manufacturer, let's say manufacturer A, I would say 80% of the time it's the same manufacturer. 10% of B, 8% of C and 2% of D.

So, what was requested to me, when I discussed about fully-dropping the stock locations, was to keep that because it's useful to them to explicitly (fully) remind them what is the manufacturer of this product. 

I do think it may not be a bad idea to keep a stock location name field (as a FK to avoid data redundancy).

I agree this is quite a weird feature. I'm still building my commerce knowledge, so I don't really have a really criticizing mind as of now. There are others weird features that were requested, they are implemented but flagged as open to be changed on my roadmap, I'm also planning on seeing with Saleor's team for thoughts on some cases (especially as I'm slowing but surely trying to shrink the difference gap between Saleor and our store's source code).

Thanks a lot.

patrys:
@NyanKiyoshi What if we introduced a manufacturer field on a product similarly to what Shopify does for example?

NyanKiyoshi:
@patrys just took a look


Seems nice to answer this issue!

PR #2006, opened 2018-03-30
[WIP] Add JWT-based auth

### LINKED ISSUE #2011: JWT-based authentication
Currently we depend on existing cookies for authentication. The backend now supports JWT-based auth tokens and we should switch to those.

The `tokenAuth` mutation returns a token for a valid email and password.

We've discussed the architecture with @maarcingebala and we've planned it as follows:

* `<AuthProvider>` internally takes care of persisting and restoring tokens from local storage (browser persistence API)
* `<AuthProvider>` internally wraps all the necessary `<Migration>` objects
* `<AuthProvider>` passes the user object, a login method, and a logout method to its callable children
* `<AuthProvider>` puts an object containing the above in a React context provider
* `<AuthClient>` is a React context consumer that extracts the above and provides it to its own children
* The login page uses the client to access the login method
* When the user is not logged in, the main router `<Switch>` is replaced with a single `<Route>` containing the login page

TODO:

* [ ] We should have a way to either ask for only staff users or to ask for a certain list of scopes when logging in to avoid situations where a regular customer logs into the dashboard and is greeted with an empty interface.

patrys:
This adds a login dialog and the ability to send and refresh the JWT token.

Fixes #2011 

TODO:
* [x] introduce a credential store component that is able to re-render its children when the auth token appears and disappears (in future when logged in user changes)
* [ ] create our own Mutation and Query components that handle errors (I want to avoid having to copy-paste error handling boilerplate like we are currently forced to do)
* [x] notify the store that the current token is no longer valid (the above components could use context to communicate with the credential store)
* [x] make sure it makes sense to persist the JWT auth token using localStorage (it should be safe as it's a client token but the token itself is only valid for a very short time and needs to be refreshed often)
* [x] fetch user details and permissions along with the token
* [x] create components that conditionally hide their children if the currently logged in user does not have sufficient permissions
* [x] when the user clicks "Remember me", store the auth token in `localStorage`, otherwise use `sessionStorage`

### Pull Request Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,14 +26,14 @@
         "@babel/traverse": "7.0.0-beta.44",
         "@babel/types": "7.0.0-beta.44",
         "babylon": "7.0.0-beta.44",
-        "convert-source-map": "^1.1.0",
-        "debug": "^3.1.0",
-        "json5": "^0.5.0",
-        "lodash": "^4.2.0",
-        "micromatch": "^2.3.11",
-        "resolve": "^1.3.2",
-        "semver": "^5.4.1",
-        "source-map": "^0.5.0"
+        "convert-source-map": "1.5.1",
+        "debug": "3.1.0",
+        "json5": "0.5.1",
+        "lodash": "4.17.10",
+        "micromatch": "2.3.11",
+        "resolve": "1.5.0",
+        "semver": "5.5.0",
+        "source-map": "0.5.7"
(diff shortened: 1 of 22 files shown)


patrys:
Rebased on current branch. We should consider dropping the refresh interval before merging. I'll open separate TODOs for the subtasks.

PR #2020, opened 2018-04-02
Added a box to accept cookies and learn more

NyanKiyoshi:
Hello,

I merged some changes that were done on our side about cookie laws. This PR shows a box to accept cookies and contains a link to the privacy policy.

#### Screenshot
Screenshot

Also note that this cookie box is not GDPR compliant. As it would require some changes on Saleor's code from my understanding (didn't dig this through yet). So this is just a simple "I accept" mechanism that sets a cookie to remember it (expiring after 3 months).

### To be done
- [x] Generate a page with the slug 'privacy', so users can customize the privacy page.
  - This page would contain base privacy information (mostly about Saleor);
  - This page would contain a list of cookies we use, and also the expiration time.
- [x] Update documentation.

This page was already done long ago, but I will have to translate it from French to English (also, while speaking about languages, I'm planing on implementing a clean and user friendly i18n on models, but I don't have any ETA, it's on my spring roadmap).

**If this PR is good to Mirumee, I would just need some help:** where do I generate the privacy page data? In `random_data.py`?

### Need feedback on visual rendering

I'm not sure about i18n, do you think, it's better to keep the above screenshot or put two buttons like below?

Screenshot

```diff
Index: templates/base.html
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- templates/base.html	(revision 98b4e7815e81c2d778c5198fcd342c5f256f0e86)
+++ templates/base.html	(revision )
@@ -274,10 +274,11 @@
       This website uses cookies to improve user experience.
       By using this website, you are accepting our policy about cookies.
     {% endblocktrans %}
-    <a href="{% url "page:details" slug="privacy" %}">
-      {% trans "Read further." context "Cookie prompter box, read more link after the information text" %}
-    </a>
-  </p>
+  </p>
+
+  <a class="btn btn-secondary btn-sm" href="{% url "page:details" slug="privacy" %}">
+    {% trans "Read further" context "Cookie prompter box, read more link after the information text" %}
+  </a>
 
   <button class="btn btn-primary btn-sm">
     {% trans "I accept" context "Cookie prompter box, accept/ close button" %}
```

**Pro:** i18n friendly, as there is not risk to create a bad order between the text and the link in some languages as the link now separated (see the above diff).

**Con:** not user friendly, as the user eyes may not pop directly on the button or may miss the button (tried with a bigger button, it's pretty much the same and uglier).

### Pull Request Checklist

[...]

PR #2021, opened 2018-04-02
Customers must accept *our* selling contract before placing their order

NyanKiyoshi:
Hello!

This PR has for goal to require the customers to accept the store's selling contract before placing any order. As this is required by most of the countries of Europe.

The checkout process is quite a mess at the moment... I'm not sure if you are planing on reviewing Saleor's order process. Hit me up if you need any help.

Still the same issue as #2020, I'm not sure where I should integrate a default page data creation (for the slug: `selling-contract`).

-----

### To be done
- [x] Generate a page with the slug 'privacy', so users can customize the privacy page.
  - This page would contain base privacy information (mostly about Saleor);
  - This page would contain a list of cookies we use, and also the expiration time.
- [x] Update documentation;
- [ ] Implement #2038 once available into master.

-----

**Edit: can't do much about the code climate issue, it would require to refactor the checkout. It's better to let it be for now and open another PR for this.**

-----

### Screenshot
screenshot

### Pull Request Checklist


saleor/checkout/views/summary.py
```diff
@@ -46,6 +46,11 @@ def handle_order_placement(request, checkout):
     return redirect_url
 
 
+def get_contract_acceptance_form(data):
+    contract_form = ContractAcceptanceForm(data or None)
```
akjanik: As data is required argument here, I think `or None` can be removed.

patrys: If you remove `or None` then the form will be invalid by default.

PR #2023, opened 2018-04-02
Allow to set a custom domain to the S3 buckets

NyanKiyoshi:
Hello!

This PR has for goal to allow users to set their S3 bucket's domain (e.g. cloudfront).

**Two new settings keys were introduced:**

| Settings Key | Information |
|---|---|
| `AWS_STATIC_CUSTOM_DOMAIN`  |  Sets the static S3 bucket's domain |
| `AWS_MEDIA_CUSTOM_DOMAIN`  |  Sets the media S3 bucket's domain |

---

*Note: I'm using `AWS_S3_CUSTOM_DOMAIN` as `AWS_STATIC_CUSTOM_DOMAIN` for the environment, because I prefer to avoid to set the S3 bucket's domain as a global and default value for any bucket (for example: the media bucket would take this value if `AWS_MEDIA_CUSTOM_DOMAIN` is not set, which may (will in most cases) result into 404s).*

### Pull Request Checklist


Pacu2:
This PR is a nice start for making saleor more CloudFront-friendly out of the box, I think we can make it even better:
- Env variables could be listed in `configuration.rst` file
- This is a good candidate for integration guide
@elwoodxblues  what do you think?

patrys:
I think with CloudFront in place there is little gain from using a separate domain for static files and media. Especially so with HTTP/2.

NyanKiyoshi:
I will edit the docs later today + [`s3.rst`](https://github.com/mirumee/saleor/blob/master/docs/deployment/s3.rst)

NyanKiyoshi:
I updated the S3 deployment documentation, I also to cut the S3 documentation into three sections: base configuration (global) which are the credentials, then we have the media and static bucket.

Hope it looks good.

PR #2024, opened 2018-04-02
Allow to force SSL (301 all HTTP connections to HTTPS)

NyanKiyoshi:
Hello!

This change introduce a new settings key: `FORCE_SSL` for users that are using (enable) SSL in Saleor. 
`FORCE_SSL` is using (setting) `SECURE_SSL_REDIRECT` of `django.middleware.security.SecurityMiddleware` which permanently redirect every HTTP connections to HTTPS (301).

### Pull Request Checklist

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -93,6 +93,9 @@ EMAIL_USE_SSL = email_config['EMAIL_USE_SSL']
 ENABLE_SSL = ast.literal_eval(
     os.environ.get('ENABLE_SSL', 'False'))
 
+if ENABLE_SSL and ast.literal_eval(os.environ.get('FORCE_SSL', 'False')):
+    SECURE_SSL_REDIRECT = True
+
 DEFAULT_FROM_EMAIL = os.environ.get('DEFAULT_FROM_EMAIL')
 ORDER_FROM_EMAIL = os.getenv('ORDER_FROM_EMAIL', DEFAULT_FROM_EMAIL)
 
@@ -148,6 +151,7 @@ SECRET_KEY = os.environ.get('SECRET_KEY')
 
 MIDDLEWARE = [
     'django.contrib.sessions.middleware.SessionMiddleware',
+    'django.middleware.security.SecurityMiddleware',
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',


Pacu2:
I wonder what are the advantages of having `SecurityMiddleware` if one doesn't want to `FORCE_SSL`?

Personally I'm not a fan of, think that such things could be handled at uwsgi/CDN level.

NyanKiyoshi:
Maybe we should append `SecurityMiddleware` to the middlewares if `FORCE_SSL` is True?

patrys:
I would keep SecurityMiddleware in the middleware stack but would prefer to set `SECURE_SSL_REDIRECT = not DEBUG` instead of introducing a separate environment variable.

elwoodxblues:
We could enable `SecurityMiddleware` conditionally as you propose. Also, instead of introducing new environmental variables we were thinking that `SECURE_SSL_REDIRECT` could be enabled when `DEBUG=False`.

NyanKiyoshi:
Issue with `SECURE_SSL_REDIRECT` to `True` when `DEBUG=False` is that, it may create an infinite loop in some cases/ some configurations. The case I'm thinking of, is when someone uses both SSL but also uses a proxy (with no HTTPS to the origin) that is going through HTTPS (client to proxy), and then to HTTP (proxy to web server).

patrys:
I think Django already supports detecting secure reverse proxies based on HTTP headers though?

NyanKiyoshi:
True. From django documentation, we could use [`SECURE_PROXY_SSL_HEADER`](https://docs.djangoproject.com/en/1.10/ref/settings/#std:setting-SECURE_PROXY_SSL_HEADER) to mark it secure.

```patch
Index: saleor/settings.py
<+>UTF-8
--- saleor/settings.py	(revision 138bc8f720d2dc0363be2153a6415926ae68f423)
+++ saleor/settings.py	(revision )
@@ -134,8 +134,9 @@
 
 ENABLE_SSL = get_bool('ENABLE_SSL', False)
 
-if ENABLE_SSL and get_bool('FORCE_SSL', False):
-    SECURE_SSL_REDIRECT = True
+if ENABLE_SSL:
+    SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTOCOL', 'https')
+    SECURE_SSL_REDIRECT = not DEBUG
 
 DEFAULT_FROM_EMAIL = os.environ.get('DEFAULT_FROM_EMAIL')
 ORDER_FROM_EMAIL = os.getenv('ORDER_FROM_EMAIL', DEFAULT_FROM_EMAIL)
```

PR #2025, opened 2018-04-02
Filter button is now a real button (for mobile users)

NyanKiyoshi:
Hello!

After two weeks or so of testing in post-production, I'm opening this PR to replace the simple text button to toggle filter menu, to a real button. Otherwise, users don't know it's a button but just a label.

----

#### After

----

#### After

----

#### After

----

### Pull Request Checklist

### DIFF
--- a/saleor/static/scss/components/_filters.scss
+++ b/saleor/static/scss/components/_filters.scss
@@ -83,9 +83,6 @@
     }
   }
   &__title {
-    @include media-breakpoint-down(sm) {
-      padding: $global-padding *2 0;
-    }
     @media (max-width: 370px) {
       padding-top: 5rem;
     }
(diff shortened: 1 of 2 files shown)


Pacu2:
Wonder if dropdown isn't a better fit?
eg. as seen on amazon 
filters_amazon

NyanKiyoshi:
I think it would require to change the logic on the toggle button. As of now, I think it's preferable to just make the toggle label/ button visible to the user.
Then, we could try to think and implement a better way for storefront 2.0.

elwoodxblues:
I agree that "Filters" looks like a label on mobile and we should definitely improve the UX here. By looking at the screens I'm not convinced that the default button styling fits there. I'll hand it to someone from our design team to take a look.

NyanKiyoshi:
Okay then!

PR #2026, opened 2018-04-03
Generate ProductVariant name based on attributes

### LINKED ISSUE #1957: Drop variant name field
Currently `ProductVariant.name` field is not particularly useful. Moreover, it causes some confusion in code when it comes to variant representation. and should be dropped. In my opinion, we should drop this field and use variant attributes as a name. We already do it, but our model allows variants without attributes which doesn't make much sense and we should eventually change it, but for now we could use SKU as fallback.

Pacu2:
close #1957 

ProductVariant name is autogenerated on ProductVariantForm's save, based on attributes assigned to it

Also cleaned up Product's utils module and its tests, splitting it to smaller modules

### Pull Request Checklist


saleor/dashboard/product/forms.py
```diff
@@ -123,19 +124,34 @@ def clean(self):
...
-                if query.exists():
-                    msg = pgettext_lazy(
-                        'Product type form error',
-                        'Some products of this type have more than '
-                        'one variant.')
-                    self.add_error('has_variants', msg)
+        if not self.instance.pk:
+            return data
+
+        variants_changed = (
```
elwoodxblues: Django's [has_changed](https://docs.djangoproject.com/en/2.0/ref/forms/fields/#checking-if-the-field-data-has-changed) wouldn't help here?

saleor/dashboard/product/forms.py
```diff
@@ -123,19 +124,34 @@ def clean(self):
...
+            if query.exists():
+                msg = pgettext_lazy(
+                    'Product type form error',
+                    'Some products of this type have more than '
+                    'one variant.')
+                self.add_error('has_variants', msg)
+
+        # Some variant attributes could be removed so name should be updated
+        # accordingly
+        initial_attributes = set(list(self.instance.variant_attributes.all()))
```
elwoodxblues: This `clean` is huuuge. Maybe we could benefit from having 2-3 helper functions that do these particular pieces of logic?

saleor/dashboard/product/forms.py
```diff
@@ -256,9 +272,17 @@ class Meta:
             'price_override': pgettext_lazy(
                 'Override price', 'Override price'),
             'name': pgettext_lazy('Product variant name', 'Name')}
+        help_texts = {
```
elwoodxblues: Maybe `name` field shouldn't be included in the form, since it's not editable. It will be displayed somewhere in the page title or breadcrumbs anyway.

saleor/product/utils.py
```diff
@@ -377,3 +351,56 @@ def get_product_list_context(request, filter_set):
         'sort_by_choices': SORT_BY_FIELDS,
         'now_sorted_by': now_sorted_by,
         'is_descending': is_descending}
+
+
+def get_product_attributes_data(product):
```
elwoodxblues: This `utils` module has become quite cluttered, so maybe we should think of splitting it into multiple files. E.g., these attributes-related functions could be stored in a separate file `attributes.py`. I'm thinking of such structure:
```
product
  utils
    __init__.py
    attributes.py
    availability.py
    variant_picker.py
    etc.
```
For now, we could only introduce `attributes.py` and keep the rest of utility functions in `__init__.py` (which could import utilities from all submodules) and clean up the rest in a separate PR. Up to you.

saleor/dashboard/product/forms.py
Pacu2: Unfortunately not

PR #2038, opened 2018-04-04
Protected pages & Selling contract + Privacy pages

### LINKED ISSUE #2022: Selling contract creation and customization
Hello,

As already made by #2021, we, Saleor, would need to allow staff users to add and customize the selling contract of their store.

I would like to discuss with Saleor's team on how to implement the later (selling contract customization process).

### First solution — a simple page

The easiest (as of #2021), would be to create and edit a simple page through Saleor's page mechanism. And thus, of course, if the selling contract page is missing, the user would see a 404 error, so Saleor won't depend on it.

### Second solution — store settings

The more user/ staff friendly (+ we could restrict access through permissions), would be to add options to the store settings, just like Shopify does. Or another way, as we will need to set some settings, for VAT for example, would be inside a law settings or something.

shopify screenshot

----

What are your thoughts on this?

NyanKiyoshi:
This PR has for goal to implement the suggestions of #2022 and close #2022. Closing this PR will also make both #2020 and #2021 ready for merge (after review)—Edit: I may have to do a few edits to #2020 and #2021 when this PR will be merged.

### Details
- This PR introduce a new, and temporary mechanism of page protection. Which means, nobody can delete it from dashboard (not even the super user), to avoid any breaking in dependencies;
- It will also generate a privacy and selling contract page(s) on migration and on populate db (which will check for existing page).

Then, will be dropped with a better solution later on, depending how it will be decided to proceed and to implement it.

### Page protection
- The PR adds a new settings key named `PROTECTED_PAGES`, which contains page `slugs`;
- It also adds a new property `is_protected` to the `Page` model which returns `True` if the slug is in `PROTECTED_PAGES`;
- The view responsible for the page deletion in dashboard return an error modal if the page is protected. Note, I didn't decide to hide the 'Remove button' as the user may wonder why the page doesn't have one;
- The page form disables the slug field if the page is flagged as protected, which blocks any bypassing or inadvertent changes/ breakings.

### Todo
- [x] Implement the checks in GraphQL;
- [x] Add tests;
- [x] Generate privacy and selling contract page(s);
- [x] Generate from migrations;
- [x] Document the changes;

### Pull Request Checklist


NyanKiyoshi:
This PR is ready to be reviewed.

default_pages/privacy-policy.md
```diff
@@ -0,0 +1,39 @@
+# Privacy Policy
```
elwoodxblues: We've decided that we don't want to maintain default policies with Saleor at the moment. We only want to give a simple mechanism to allow defining it, but it's the shop owner/operator who is responsible for formulating those policies.

saleor/dashboard/page/decorators.py
```diff
@@ -0,0 +1,16 @@
+from functools import wraps
+
+from django.shortcuts import get_object_or_404
+
+from ...page.models import Page
+from .modals import modal_protected_page
+
+
+def unprotected_page_required(view):
```
elwoodxblues: Do we really need the decorator for that purpose? In case of `page_delete` view we could check `page.is_protected` in the view. We could consider adding such decorator when we have more views requiring it.

saleor/dashboard/page/modals.py
```diff
@@ -0,0 +1,11 @@
+from django.template.response import TemplateResponse
+
+PROTECTED_PAGE_ERROR_MODAL_PATH = (
+    'dashboard/page/modals/protected_page_error_modal.html')
+
+
+def modal_protected_page(request, page):
```
elwoodxblues: I think this logic could be moved directly to the `page_delete` view.

saleor/graphql/core/mutations.py
```diff
@@ -102,12 +102,22 @@ def __init_subclass_with_meta__(
         cls._update_mutation_arguments_and_fields(
             arguments=arguments, fields=fields)
 
+    @classmethod
+    def _delete_instance(cls, instance):
```
elwoodxblues: I like this idea, we were thinking something similar in "create" and "update" mutations, so that you don't have to override the whole `mutate` method which contains a lot of boilerplate. I think we could rename it to `delete_instance` - since we want to give the ability to override it, it shouldn't be "private" (according to Python's naming conventions).

saleor/graphql/page/decorators.py
```diff
@@ -0,0 +1,23 @@
+from functools import wraps
+
+from graphql_jwt.exceptions import PermissionDenied
+
+from ...page.models import Page
+
+
+def must_be_unprotected(fn):
```
elwoodxblues: Again, why the decorator? Ideally, this simple condition could be checked inside overwritten `delete_instance` method.

saleor/core/templatetags/urls.py
```diff
@@ -12,3 +13,47 @@ def build_absolute_uri(request, location):
...
+        return internal_pages[internal_name]
+
+    raise ValueError((
+        '\'{}\' is not a know internal page name. '
+        'Please check the \'INTERNAL_PAGES\' settings key if you edited it.'
+    ).format(internal_name))
+
+
+@register.simple_tag
+def get_internal_page_url(internal_name):
```
elwoodxblues: Why and where do we need this? I thought staff will manually add those pages to menus it they need it.

elwoodxblues:
This is great work @NyanKiyoshi, thanks! However, I think what would be the best for us now is a simple mechanism that adds those pages (privacy, refund and terms of service) in migrations, even without any content (or with "Example privacy policy page" placeholder text).

PR #2043, opened 2018-04-05
Add Customer Notes

### LINKED ISSUE #2013: Add notes to customers
Hello, this is a quick feature suggestion for Saleor.

- Be able to add notes to customers through the dashboard (e.g. self/team-reminders, warning, information about customer, etc.);
- Be able to see those notes when we are creating an order for a given customer, as it may be useful or important (needs #1971).

I recently saw that shopify has this feature as well, but didn't try it out.

----

As screenshots are speaking a lot, here is the current implementation:

### Order creation, with this customer selected

----

I want your thoughts about this feature, I think it would be great to add it to everyone.

koradon:

We want a simple customer note functionality.
This will be helpful when a staff member will create order from dashboard.

Fixes #2013 

### Pull Request Checklist


koradon:
Some print screens:
screenshot from 2018-04-06 12-44-27
screenshot from 2018-04-06 12-44-20
screenshot from 2018-04-06 12-43-38

templates/dashboard/customer/detail.html
```diff
@@ -56,6 +56,18 @@ <h4>
               </div>
             {% endif %}
           </div>
+          <div class="row">
+            {% if customer.note %}
```
Pacu2: Shouldn't it be one line up?

saleor/dashboard/customer/views.py
```diff
@@ -93,3 +97,24 @@ def ajax_users_list(request):
     users = [
         {'id': user.pk, 'text': user.ajax_label} for user in queryset]
     return JsonResponse({'results': users})
+
+
+@staff_member_required
+@permission_required('account.edit_user')
+def customer_add_note(request, customer_pk):
+    customer = get_object_or_404(User, pk=customer_pk)
+    print(customer)
```
akjanik: debug print!

saleor/core/models.py
```diff
@@ -0,0 +1,14 @@
...
+
+class BaseNote(models.Model):
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL, blank=True, null=True,
+        on_delete=models.SET_NULL)
+    date = models.DateTimeField(db_index=True, auto_now_add=True)
+    content = models.TextField()
+    is_public = models.BooleanField(default=True)
+
+    class Meta:
```
elwoodxblues: This model has to be abstract. We don't want to save instances of `BaseNote` in the database, but only the derived models such as `OrderNote` or `CustomerNote`.

saleor/order/migrations/0045_auto_20180405_0656.py
```diff
@@ -0,0 +1,60 @@
+# Generated by Django 2.0.3 on 2018-04-05 11:56
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def move_data_to_base_note(apps, schema_editor):
+    OrderNote = apps.get_model('order', 'OrderNote')
+    BaseNote = apps.get_model('core', 'BaseNote')
```
elwoodxblues: I think if `BaseNote` was abstract we wouldn't have to do this.

saleor/core/models.py
elwoodxblues: This model has to be abstract. We don't want to save instances of `BaseNote` in the database, but only the derived models such as `OrderNote` or `CustomerNote`.

saleor/dashboard/customer/views.py
```diff
@@ -44,7 +45,10 @@ def customer_details(request, pk):
             'default_billing_address', 'default_shipping_address')
     customer = get_object_or_404(queryset, pk=pk)
     customer_orders = customer.orders.all()
-    ctx = {'customer': customer, 'customer_orders': customer_orders}
+    notes = customer.notes.all()
```
Pacu2: Shouldn't we prefetch notes in aboves queryset?

saleor/core/models.py
```diff
@@ -0,0 +1,15 @@
...
+        abstract = True
+        ordering = ('date', )
```
Pacu2: Default ordering is by pk, in this case, ordering by date added does the same thing, I suppose it's not needed

patrys: We should also make sure that ordering is always stable, it's possible for two dates to be equal (even if unlikely) so if we really want to sort by date, I suggest that we sort by `('date', 'pk')`.

koradon: Wa agreed that ordering by pk will be enough.

PR #2059, opened 2018-04-10
Reduce number of queries

Pacu2:
This PR reduces number of queries fired in several places in the dashboard

### DIFF
--- a/saleor/order/models.py
+++ b/saleor/order/models.py
@@ -132,17 +132,27 @@ class Order(models.Model):
     def get_absolute_url(self):
         return reverse('order:details', kwargs={'token': self.token})
 
-    def get_last_payment_status(self):
+    def get_last_payment(self):
+        """Query-friendly method for getting last payment from cached
+        payments.
+        """
         payments = list(self.payments.all())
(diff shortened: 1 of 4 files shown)


saleor/order/models.py
```diff
@@ -132,14 +132,26 @@ def __str__(self):
...
+        """Query-friendly method for getting last payment from cached
+        payments.
+        """
+        payments = list(self.payments.all())
+        if not payments:
+            return
+        payment_mapping = {payment.pk: payment for payment in payments}
+        latest_payment_pk = max(payment_mapping.keys())
+        last_payment = payment_mapping[latest_payment_pk]
+        return last_payment
```
mad-anne: We have `PaymentQuerySet` which has overriden `last()` method - seems it's not used now. And I wonder if we could make this method easier, maybe this would work:
```payments = list(self.payments.all())
return payments[0] if payments else None```

Comment (patrys) [reply saleor/order/models.py]: Or even:

```python
from operator import attrgetter

return max(self.payments.all(), key=attrgetter('pk'))
```

Issue #1999, opened 2018-03-28
Display logo in emails without internet access


NyanKiyoshi:
I really think we would shouldn't be putting [our logo as a link to our remote server](
https://github.com/mirumee/saleor/blob/master/templates/templated_email/source/shared/header.mjml#L8).

The reasons are:
- It may generate useless bandwidth to our server/ static server and use customers data;
- We are dependent on a remote server which is not "reliable" (may happen to be unavailable, slow, ...);
- Requires explicit permission from the customer on mail clients most of the time.

Also, despite allowing the remote pictures on the gmail mail client on Android, I never managed to make it load the logo. Not sure why. With, and without static only server.

![](https://i.imgur.com/Nowplxv.png)

(*note: I took the screenshot of one of the local test server emails, as I don't want to display sensitive information. But of course, 127.0.0.1 would not work without any doubt*).

---

**I think that it would be better to just (put) include the SVG file in the mails instead of linking it**, the SVG logo and SVG logos in general have really small footprint.

---

Edit: if we have a S3 bucket for static files, we get both the website domain and the s3 bucket domain (see on the right, where there are black squares).
![](https://i.imgur.com/xWWGcaq.png)

elwoodxblues:
I'm not sure how does embedding SVGs in email templates work, but from what I recall it's still not supported by some of the popular inboxes.

In the line, that you highlighted, the path is constructed in a wrong way - the `{% static %}` template tag resolves the absolute URL when `STATIC_URL`.

elwoodxblues:
@mad-anne Could you take a look when you have time? As far as I remember you were implementing those templates.

patrys:
It's possible to attach the file as an "inline" multipart attachment and reference that in the "src" attribute.

cmoralesmx:
Is this issue still relevant? I would like to work on it. @patrys's solution seems doable, I think I can implement this.

maarcingebala:
This issue is not relevant anymore as we removed a logo from emails. This is part of reworking the email mechanism where we want to move email templates out from the Saleor core repository.

PR #2060, opened 2018-04-10
Fix generating path to placeholder

elwoodxblues:
### Pull Request Checklist


patrys:
@artursmet I'm afraid this could break CloudFront and result in the storefront linking directly to S3.

NyanKiyoshi:
There is a similar issue with #1999 if you look at the end of the issue, on the screenshots. But I'm thinking about opening a PR that will include/ copy the SVG file into the mail source if that's sound okay, I never got any feedback.

elwoodxblues:
@patrys I think it should solve the issue for now. If this is not the right approach, then we have a more general problem since this is the standard way to how we currently reference static files.

PR #2085, opened 2018-04-16
Fix main menu displaying child items

eskerda:
Current menu implementation displays child `MenuItems` on the main nav. 

#### How to reproduce

1. With sample data
2. Edit a menu item on the dashboard
3. Add a menu item

#### Before

image

#### After

image

### Pull Request Checklist

### DIFF
--- a/saleor/core/templatetags/shop.py
+++ b/saleor/core/templatetags/shop.py
@@ -23,5 +23,7 @@ def menu(context, site_menu=None, horizontal=False):
         return
     menus = context[NAVIGATION_CONTEXT_NAME]
     menu = next((menu for menu in menus if menu.pk == site_menu.pk), None)
-    menu_items = menu.items.all() if menu else None
+    if not menu:
+        return
+    menu_items = [item for item in menu.items.all() if item.parent_id is None]
     return {'menu_items': menu_items, 'horizontal': horizontal}
--- a/tests/test_templatetags.py
+++ b/tests/test_templatetags.py
@@ -1,6 +1,6 @@
 from django.urls import reverse
 
-from saleor.core.templatetags.shop import get_sort_by_url
+from saleor.core.templatetags.shop import get_sort_by_url, menu
 
 
 def test_sort_by_url_ascending(admin_client, default_category):
@@ -21,3 +21,9 @@ def test_sort_by_url_descending(admin_client, default_category):
     result = get_sort_by_url(response.context, 'name', descending=True)
     expected = url + '?sort_by=-name'
     assert result == expected
+
+
+def test_menu(client, menu_with_items):
+    response = client.get(reverse('home'))
+    result = menu(response.context, menu_with_items)
+    assert all((i for i in result['menu_items'] if i.parent_id is None))


Pacu2:
My proposition is:
in `saleor/core/context_processor.py` we should prefetch `items__parent` as well, and remove unnecessary filter 
Then in `shop.py` we could filter out menus without parent from cached queryset, like:
```
    menu_items = None
    if menu:
        menu_items = [item for item in menu.items.all() if item.parent is None]
```

eskerda:
@patrys @Pacu2 Updated code, let me know if this is what you had in mind.

Also, I think this is slightly related to https://github.com/mirumee/saleor/issues/2049, so maybe it's worth adding a couple of tests to a new `tests/test_menu.py`. The existing fixture for menu items would already cover this case.

Pacu2:
@eskerda You're totally right, this should be well tested, do you mind adding tests for this functions, or shall we handle that in a separate issue?

eskerda:
@Pacu2 sure, no problem. I think it's alright with just one or two integration tests testing what appears on the homepage.

saleor/core/templatetags/shop.py
```diff
@@ -23,5 +23,7 @@ def menu(context, site_menu=None, horizontal=False):
         return
     menus = context[NAVIGATION_CONTEXT_NAME]
     menu = next((menu for menu in menus if menu.pk == site_menu.pk), None)
-    menu_items = menu.items.all() if menu else None
+    if not menu:
+        return {'menu_items': None, 'horizontal': horizontal}
```
eskerda: I guess it is alright to just `return` in this case

saleor/core/templatetags/shop.py
```diff
@@ -23,5 +23,7 @@ def menu(context, site_menu=None, horizontal=False):
         return
     menus = context[NAVIGATION_CONTEXT_NAME]
     menu = next((menu for menu in menus if menu.pk == site_menu.pk), None)
-    menu_items = menu.items.all() if menu else None
+    if not menu:
+        return
+    menu_items = filter(lambda i: i.parent_id is None, menu.items.all())
```
Pacu2: Personally I prefer list comprehension, it's more readable

eskerda: Sure, no worries

NyanKiyoshi:
I don't get what is the issue.

Shouldn't [this line](https://github.com/mirumee/saleor/blob/master/saleor/core/utils/random_data.py#L592) already handle this issue? I don't think this is the job of the tag to check that. 

```python
    if not top_menu.items.exists():
        # create a menu entry for all root categories
        categories = Category.objects.filter(parent__isnull=True)
```

I agree with @Pacu2, `menu.items.all()` is generating a "useless" SQL request, menus should be fetched with children.

Pacu2:
@NyanKiyoshi That's not an issue with `populatedb`, if you assign an `MenuItem` as a children, in the navigation tab, it will display as a parent as well.

NyanKiyoshi:
Thanks @Pacu2!

While I'm looking at the template `menu.html`, I'm not sure that [`item.get_children()`](https://github.com/mirumee/saleor/blob/master/templates/menu.html#L4-L23) is hitting the cache, even with a prefetch? I would need to check this but I cannot right now.

NyanKiyoshi:
@eskerda after digging through and looking through the menu implementation of Saleor, I did not find any solution that seemed good enough. 
I would, at least for now, suggest to keep your solution, which is to iterate the menu items and look for root nodes. But as I said above, your current changes are making Python to iterate, to process twice the node list. I would suggest to move this logic to `menu.html` as we are already iterating there :+1: 

```diff
Index: templates/menu.html
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- templates/menu.html	(date 1523903767000)
+++ templates/menu.html	(date 1523909488000)
@@ -1,44 +1,46 @@
 {% url 'home' as home_url %}
 <ul class="menu {% if horizontal %}nav mb-4 mb-md-0{% endif %}{% if request.get_full_path == home_url %} no-border{% endif %}">
   {% for item in menu_items %}
-  {% with children=item.get_children %}
-    <li class="{% if horizontal %}nav-item{% endif %} {% if children %}nav-item__dropdown{% endif %} menu__item">
-      <a class="{% if horizontal %}nav-link{% endif %}" href="{{ item.get_url }}">
-        {{ item }}
-      </a>
-      {% if children %}
-        <div class="{% if horizontal %}nav-item__dropdown-content{% else %}nav-item__submenu{% endif %}">
-          <div class="container">
-            <ul>
-            {% for child in children %}
-              <li>
-                <a href="{{ child.get_url }}">
-                  {% if horizontal %}
-                    <strong>{{ child }}</strong>
-                  {% else %}
-                    {{ child }}
-                  {% endif %}
-                </a>
-              </li>
-              {% with grandchildren=child.get_children %}
-                {% if grandchildren %}
-                <ul>
-                  {% for grandchild in grandchildren %}
-                    <li>
-                      <a href="{{ grandchild.get_url }}">
-                        {{ grandchild }}
-                      </a>
-                    </li>
-                  {% endfor %}
-                </ul>
-                {% endif %}
-              {% endwith %}
-            {% endfor %}
-            </ul>
-          </div>
-        </div>
-      {% endif %}
-    </li>
-  {% endwith %}
+    {% if not item.parent_id %}
+      {% with children=item.get_children %}
+        <li class="{% if horizontal %}nav-item{% endif %} {% if children %}nav-item__dropdown{% endif %} menu__item">
+          <a class="{% if horizontal %}nav-link{% endif %}" href="{{ item.get_url }}">
+            {{ item }}
+          </a>
+          {% if children %}
+            <div class="{% if horizontal %}nav-item__dropdown-content{% else %}nav-item__submenu{% endif %}">
+              <div class="container">
+                <ul>
+                {% for child in children %}
+                  <li>
+                    <a href="{{ child.get_url }}">
+                      {% if horizontal %}
+                        <strong>{{ child }}</strong>
+                      {% else %}
+                        {{ child }}
+                      {% endif %}
+                    </a>
+                  </li>
+                  {% with grandchildren=child.get_children %}
+                    {% if grandchildren %}
+                    <ul>
+                      {% for grandchild in grandchildren %}
+                        <li>
+                          <a href="{{ grandchild.get_url }}">
+                            {{ grandchild }}
+                          </a>
+                        </li>
+                      {% endfor %}
+                    </ul>
+                    {% endif %}
+                  {% endwith %}
+                {% endfor %}
+                </ul>
+              </div>
+            </div>
+          {% endif %}
+        </li>
+      {% endwith %}
+    {% endif %}
   {% endfor %}
 </ul>
```

elwoodxblues:
Thanks for your contribution @eskerda!

PR #2124, opened 2018-04-25
Add storyshots

dominik-zeglen:
Integrates Jest snapshot tests with Storybook.


patrys:
Do we need both `@babel/core` and `babel-core`?

dominik-zeglen:
Yes, it's Storyshots' peer dependency and without it, Storyshots won't work due to version mismatch (`6.26.0` vs `7.0.0`).

PR #2134, opened 2018-04-27
Feature: put a recaptcha check on account forms

NyanKiyoshi:
Hello,

This is a little feature we (on our side) planned into implementing to Saleor, to add extra checks. This pull request adds a simple field to account forms to check if the user is human to restrict more the access to bots.

This feature is by default disabled and can be enabled/ toggled from settings.py, through the environment. The field is added through the creation of the base form (lazily).

screenshot

### Pull Request Checklist


saleor/account/forms.py
```diff
@@ -9,6 +10,14 @@
 from .i18n import AddressMetaForm, get_address_form_class
 
 
+class FormWithReCaptcha(forms.BaseForm):
+    def __new__(cls, *args, **kwargs):
+        if settings.ENABLE_RECAPTCHA:
+            cls.base_fields['_captcha'] = ReCaptchaField(
+                label=pgettext_lazy('Captcha field label', 'Are you human?'))
```
Pacu2: I'm not a fan of this label, as usually `Recaptcha` is self explanatory

saleor/settings.py
```diff
@@ -468,3 +473,15 @@ def get_host():
...
     'top_menu_name': 'navbar',
     'bottom_menu_name': 'footer'}
+
+
+NOCAPTCHA = True
+
+ENABLE_RECAPTCHA = get_bool('ENABLE_RECAPTCHA', False)
+
+if 'RECAPTCHA_PUBLIC_KEY' in os.environ:
+    RECAPTCHA_PUBLIC_KEY = os.environ['RECAPTCHA_PUBLIC_KEY']
```
Pacu2: What's the downside of having `None` `RECAPTCHA_PUBLIC_KEY`?
Shouldn't we preset some values for local development, or will it pass anyway?

saleor/settings.py
```diff
@@ -468,3 +473,15 @@ def get_host():
 DEFAULT_MENUS = {
     'top_menu_name': 'navbar',
     'bottom_menu_name': 'footer'}
+
+
+NOCAPTCHA = True
+
+ENABLE_RECAPTCHA = get_bool('ENABLE_RECAPTCHA', False)
+
```
Pacu2: I'd group all those variables as they're treating same subject

saleor/settings.py
```diff
@@ -468,3 +473,15 @@ def get_host():
 DEFAULT_MENUS = {
     'top_menu_name': 'navbar',
     'bottom_menu_name': 'footer'}
+
+
+NOCAPTCHA = True
+
+ENABLE_RECAPTCHA = get_bool('ENABLE_RECAPTCHA', False)
```
Pacu2: Wonder if we shouldn't enable it by default, @patrys what do you think?

Pacu2:
Great PR overall, thinking about improvemen: We could disable 'send' button untill Recaptcha completes.

Something like:
in forms.py:
```
captcha = ReCaptchaField(attrs={'callback': 'recaptchaCallback'})
```
in template.html:
```
<form method="post" action=".">
    {% csrf_token %}
    {{ form }}
    <div class="row">
        <script>
            function recaptchaCallback() {
                document.getElementById('registerButton').disabled = false;
            }
        </script>
        <button id="registerButton" type="submit" class="button" disabled>
        {% trans "Register account" %}
        </button>
    </div>
</form>
```

saleor/settings.py
NyanKiyoshi: Hey! The issue right now is that the dependency is doing a `getattr(settings, key, test_api_key)`. So, setting to None, would make it use the key `None`.

https://github.com/praekelt/django-recaptcha/blob/34af16ba1ec8f27d0aeab1ab88d7a35a7b76b1ce/captcha/fields.py#L33

What I did think we could also do, is to set ourselves the [constants](https://github.com/praekelt/django-recaptcha/blob/develop/captcha/constants.py#L4-L5) through `environ.get`.

saleor/settings.py
NyanKiyoshi: Sure! Just note that `NOCAPTCHA` is to use the new (latest) version of recaptcha (which is the "Are you humain? Just check the box.")

NyanKiyoshi:
Sounds great! I will look into that when I'm available!

saleor/settings.py
patrys: I'd say enable it if we have a key. In future we should likely move the key from the environment to the settings screen.

Pacu2:
@NyanKiyoshi Since it's optional, I'd not worry about Chinese-compatible solution in this issue. Recaptcha is well known, will do the job if one need it.

saleor/account/forms.py
NyanKiyoshi: What about an empty label?
screenshot

NyanKiyoshi:
@Pacu2 I think we could let the submit button as it is. It would probably be better if we decide to add other solution (as said above, a Chinese compatible solution) as otherwise it would add Javascript code as we would need to iterate the forms to look for captcha fields.

saleor/account/forms.py
Pacu2: Looks fine to me, I think that recaptcha is labeled as seen on the picture

NyanKiyoshi:
All good! Thanks!

saleor/account/forms.py
```diff
@@ -9,6 +10,16 @@
 from .i18n import AddressMetaForm, get_address_form_class
 
 
+class FormWithReCaptcha(forms.BaseForm):
+    def __new__(cls, *args, **kwargs):
+        if settings.ENABLE_RECAPTCHA:
```
Pacu2: Can we get rid of `ENABLE_RECAPTCHA` bool in favor of checking the presence of private and public recaptcha keys?

NyanKiyoshi: Well... Yes, we could set the keys to None. And then check if both are set. Thus, users will no longer have access to the default development keys. 
If we do like so, we need to update the docs with the developments keys as note.

As you prefer.

saleor/settings.py
```diff
@@ -13,7 +14,14 @@ def get_list(text):
     return [item.strip() for item in text.split(',')]
 
 
-DEBUG = ast.literal_eval(os.environ.get('DEBUG', 'True'))
+def get_bool(name, default_value):
```
akjanik: I wonder if we can find more descriptive name for this. `get_env_variable_value`? I'd look for something not too long, but providing enough information.

NyanKiyoshi: `get_bool_from_env` maybe?

akjanik: Yes, sounds better. Shall we add an `try` that warns if someone set the env variables incorrect i.e. `DEBUG="tru"`?

elwoodxblues:
@NyanKiyoshi Thank you for this new feature! We will mention it in the new release 🎉

Issue #2088, opened 2018-04-16
Cannot add user to group in dashboard interface


raybesiga:
### I wish to a user to group on the dashboard interface

I wish to create a group and add users to it on the dashboard interface but I get a `string index out of range` error

### Steps to reproduce the problem

1. Log into dashboard
2. Create group
3. Add user by inputting email address and choosing group
4. Get the following error message

![screen shot 2018-04-16 at 15 44 47](https://user-images.githubusercontent.com/405697/38809708-24a8f21e-418d-11e8-99e2-55259a6cf337.png)

Below is the stack trace

```
Environment:


Request Method: POST
Request URL: http://www.shopihato.com/dashboard/staff/staff-create/

Django Version: 2.0.2
Python Version: 3.6.3
Installed Applications:
['storages',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.sitemaps',
 'django.contrib.sites',
 'django.contrib.staticfiles',
 'django.contrib.auth',
 'django.contrib.postgres',
 'django.forms',
 'saleor.account',
 'saleor.discount',
 'saleor.product',
 'saleor.cart',
 'saleor.checkout',
 'saleor.core',
 'saleor.graphql',
 'saleor.order.OrderAppConfig',
 'saleor.dashboard',
 'saleor.shipping',
 'saleor.search',
 'saleor.site',
 'saleor.data_feeds',
 'versatileimagefield',
 'django_babel',
 'bootstrap4',
 'django_fsm',
 'django_prices',
 'django_prices_openexchangerates',
 'graphene_django',
 'mptt',
 'payments',
 'webpack_loader',
 'social_django',
 'django_countries',
 'django_filters',
 'django_celery_results',
 'impersonate',
 'phonenumber_field']
Installed Middleware:
['django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django_babel.middleware.LocaleMiddleware',
 'saleor.core.middleware.discounts',
 'saleor.core.middleware.google_analytics',
 'saleor.core.middleware.country',
 'saleor.core.middleware.currency',
 'saleor.core.middleware.site',
[...]

PR #2147, opened 2018-05-07
Documentation about mail integrations with SMTP relays

NyanKiyoshi:
This PR introduce a few new sections to help the user to setup the email settings in Saleor, especially for setting up SMTP relays (sendgrid, mailgun and mailjet).

---

It also add a fix for #2088. Which, instead of setting the `DEFAULT_FROM_EMAIL` value to `None`, it sets the value to the email username, which means, for example, in the case of gmail: hello@gmail.com. 

The reason is quite simple: in the email url we have to set the username to an address mail for SMTP servers (not for every relays though).

e.g.: `smtp://my.gmail.username:my-password@smtp.gmail.com:465/?ssl=True`

And, in the case the user did not set any email, it won't fail, as the target is the console and not a SMTP server (or relay server).

The documentation is also now saying in relays that we need to set this value for some relays.

### Pull Request Checklist


NyanKiyoshi:
Hi,

I will add SES (as SMTP) as well as it may really be useful for users for their deployment.

And some more information about how to customize the sender information and subject prefix.

joshkpeterson:
Setting up Saelor for the first time. Without the fix here, my installation on heroku was crashing. Kind of alarming such a small thing could bring down the app. Would be great to get this merged in as this was a distraction for me.

NyanKiyoshi:
Did someone had the opportunity to look over this pull request?

docs/guides/email_integration.rst
```diff
@@ -0,0 +1,112 @@
...
+You can set the environment variable ``EMAIL_URL`` to the SMTP URL,
+which will contain a straightforward value as shown in below examples.
+
+
+.. table::
+
+   ==========================================================  ================================================================================
+   Description                                                 URL
+   ==========================================================  ================================================================================
+   GMail with **SSL on**.                                      ``smtp://my.gmaail.username@gmail.com:my-password@smtp.gmail.com:465/?ssl=True``
```
patrys: It says `gmaail`. Typo?

PR #2156, opened 2018-05-08
Sortable attribute choice values

### LINKED ISSUE #2055: Sort filters values
Hello!

This issue has been discovered some time ago and fixed on our production version. I would like to take some to discuss about a final patch for this issue.

### Filters values sorting
Here is an example, we have 100g, 150g, 200g and 1kg. But we now have a few new products (yay!): 50g and 2kg.

result - screenshot

So, here we go, we add everything, everything is great and we take a look into the groceries category. Issue: the filters values are unordered. Which means, the customer will have a few logical issues when filtering.

### Proposition
I'm thinking, maybe we could reorder the values when we are adding them (by adding a `position` column for example and cache the sorted data).

---

Edit: while I'm looking at my Saleor checklist, I also note that we shouldn't be showing filters containing no values. This case was happening when using attributes containing no pk values, only text values in products.

### LINKED ISSUE #2119: Remove "color" field from AttributeChoiceValue
Field `color` in `AttributeChoiceValue` seems to an obsolete legacy one that could be removed after investigating and ensuring it is not used.

mad-anne:
I want to merge this change because it adds ability to sort attribute choice values from dashboard, which guarantees constant order of filters in the storefront. It also unifies code standing behind sortable items (both frontend and backend), such as `ProductImage`, `AttributeChoiceValue` and `MenuItem` models.

Closes #2055, closes #2119 

### TODO's

* [x] Isolate common `SortableModel`
* [x] Add docs about sorting attribute choice values
* [x] Remove unused `color` attribute from `AttributeChoiceValue`
* [x] Add index to `sort_order` column

### Pull Request Checklist


saleor/core/models.py
```diff
@@ -12,3 +13,27 @@ class BaseNote(models.Model):
 
     class Meta:
         abstract = True
+
+
+class SortableModel(models.Model):
+    sort_order = models.PositiveIntegerField(editable=False)
```
Pacu2: If we want to order by this field, it should be indexed

PR #2166, opened 2018-05-14
Add collection endpoints in API

### LINKED ISSUE #2114: Add collections endpoint in API
We would like to be able to mutate and query for collections, filtering by name (partial match).

akjanik:
Closes #2114 

~~While working on this, I hit the issue with too many db queries when creating `Collection`, which appears as well in creating `Product Type` and I suppose any mutation that take list of GraphQL ids as an argument - `get_node` function at the end calls `cls._meta.model.objects.get(pk=id)` for each `id` provided.
I did not find or overlooked solution in `graphene-django`, so with this PR I introduced `MirumeeNode` (name is temporary) that mimics greatly `graphene.Node` but returns multiple objects with one db query. I think that we could even change old `get_node` function to use `get_nodes(...)[0]`.~~

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/api.py
+++ b/saleor/graphql/api.py
@@ -20,9 +20,10 @@ from .product.mutations import (
     ProductVariantDeleteMutation, ProductVariantUpdateMutation,
     ProductImageCreateMutation)
 from .product.resolvers import (
-    resolve_attributes, resolve_categories, resolve_products,
-    resolve_product_types)
-from .product.types import Category, Product, ProductAttribute, ProductType
+    resolve_attributes, resolve_categories, resolve_collections,
+    resolve_products, resolve_product_types)
+from .product.types import (
+    Category, Collection, Product, ProductAttribute, ProductType)
 from .utils import get_node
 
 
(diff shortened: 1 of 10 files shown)


saleor/graphql/product/resolvers.py
```diff
@@ -41,5 +25,25 @@ def resolve_attributes(category_id, info):
...
 
+def resolve_categories(info, level=None):
+    qs = models.Category.objects.all()
+    if level is not None:
+        qs = qs.filter(level=level)
+    return qs.distinct()
+
+
+def resolve_collections(info):
+    return models.Collection.objects.all()
```
elwoodxblues: We need to make sure that we exclude unpublished collections for unauthorized requests.

saleor/graphql/api.py
```diff
@@ -37,6 +40,11 @@ class Query(graphene.ObjectType):
     category = graphene.Field(
         Category, id=graphene.Argument(graphene.ID),
         description='Lookup a category by ID.')
+    collection = graphene.Field(
+        Collection, id=graphene.Argument(graphene.ID), slug=graphene.String(),
```
elwoodxblues: Is it possible to query collection by slug?

akjanik: I decided not to query by slug.

PR #2193, opened 2018-05-20
Fix calculus for orders with VAT

### LINKED ISSUE #2191: Unable to pay using PayPal when Taxes are applied
Pay an order using PayPal.

1. Make sure taxes are correctly enabled;
2. Create an order from front store as an European (e.g.: shipping to France);
3. Try to pay using PayPal;
4. It should fail, PayPal should answer by 'Bad Request'.

### What I expected to happen
Be able to pay using PayPal.

#### Request
```json
{
    "redirect_urls": {
        "return_url": "http://localhost:8000/fr/process/bc5dd0d5-0aef-4698-b7a6-8255a5fc6706/",
        "cancel_url": "http://localhost:8000/fr/process/bc5dd0d5-0aef-4698-b7a6-8255a5fc6706/"
    },
    "payer": {
        "payment_method": "paypal"
    },
    "intent": "sale",
    "transactions": [{
        "item_list": {
            "items": [{
                "currency": "EUR",
                "name": "Spot",
                "sku": "123",
                "quantity": "1",
                "price": "172.80"
            }]
        },
        "amount": {
            "currency": "EUR",
            "total": "172.80",
            "details": {
                "subtotal": "144.00",
                "tax": "28.80",
                "shipping": "0.00"
            }
        },
        "description": "Commande #6"
    }]
}
```

#### Response

```json
{
    "details": [{
        "issue": "Item amount must add up to specified amount subtotal (or total if amount details not specified)",
        "field": "transactions[0]"
    }],
    "name": "VALIDATION_ERROR",
    "message": "Invalid request - see details",
    "information_link": "https://developer.paypal.com/docs/api/payments/#errors"
}
```

NyanKiyoshi:
This PR fixes #2191. Which was caused by a miss calculation from Saleor side.

After some debugging, I discovered that Saleor was trying to subtract a gross total with gross items, a gross shipping price and total tax.

In other words:
```
subtotal = total.gross - shipping.gross - tax

--- e.g.:
Item Net Price = 22.95
Item Gross Price = 27.54 (@ 20%)

Shipping Net = 62.73
Shipping Gross = 75.28 (@ 20%)

Total Net = 85.68
Total Gross = 102.82

Total Tax = 17.14

--- Wrong calculus made by Saleor:
subtotal = total.gross - shipping.gross - tax
subtotal = 102.82 - 75.28 - 17.14
subtotal = 10.40

expected subtotal = item net = 22.95

10.40 != 22.95

----------------------
PR fixed calculus:
----------------------
subtotal = total.gross - shipping.net - tax
subtotal = 102.82 - 62.73 - 17.14
subtotal = 22.95

expected subtotal = item net = 22.95

22.95 == 22.95
```

That's almost it... The second issue, was Saleor was passing the items unit price as gross, but django-payments seems to be expecting a Net price. At least, PayPal, did expect a net price.

So, the issue was that PayPal, when summing the items, and then adding the shipping and the tax, was getting a value higher than the real total.

It was pretty much all.

----

Tested against:
- [x] Single line with unique and multiple quantities;
- [x] Multiple lines with random quantities;
- [x] Vouchers;
- [x] Without VAT.

But I'm not 100% sure if it doesn't break some other providers.

Was successfully tested against the following providers:
- [x] PayPal;
- [x] Stripe;
- [x] DotPay.

Should not make an issues for others (assumptions).

---

Maybe in the future we could improve the tests on that side, to make sure those kind of bugs does not happen?

Something that could be possible, is to add a few more tests and take the providers keys (client id, secret keys, pins, etc.) from the environment when launching tests (separated tests from the other  tests, or ignored when not set in environment). 
Those environment values would be set on Travis as hidden values (and thus we would have to make sure we don't leak any sensitive data over debug logging or else).

Let me know what you think, it would be quite easy to implement such integration tests.

### Pull Request Checklist


mad-anne:
Hi @NyanKiyoshi,
changes look good 🙂 I found two other similar places, where shipping amount is not handled correctly:
* random_data file, line 388,
* OrderMarkAsPaidForm, line 371.
Maybe it does not affect payments directly - for now - but we should correct it.

NyanKiyoshi:
Thanks! I'm looking into it!

Just a weird issue: impersonating users doesn't let us show their orders.

scr
scr

NyanKiyoshi:
Is it safe to let this one be? Or should we change it? I don't know how this part works, but I would guess that we may want to add tax to the delivery using a sum, so we may want a net delivery price.

https://github.com/mirumee/saleor/blob/908e03ea430b285d0acf36532f59b4961cf55ae8/saleor/core/analytics.py#L45-L47

elwoodxblues:
Since `order.total.tax` already includes the shipping tax, I guess we should pass shipping net price: `shipping=order.shipping_price.net`. It would seem to be correct with examples in GA docs:

https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#measuring-transactions

They don't specify which values are net or gross and also say nothing about how to deal with the shipping tax. I guess it's up to the shop administrators and to what they want to see in GA dashboard.

elwoodxblues:
@NyanKiyoshi Thanks for the PR! Small note for future - in case of conflicts, we prefer to fix them using `git rebase` rather than merging the master branch, to make the history log cleaner.

NyanKiyoshi:
Okay, thanks for that!

PR #2194, opened 2018-05-21
Order endpoints

akjanik:
I want to merge this change because it adds ability to query orders with API.
Note: it require changes from
https://github.com/mirumee/django-prices/issues/61
so build will fail until above PR is merged and released.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/filters.py
+++ b/saleor/graphql/order/filters.py
@@ -1,0 +1,39 @@
+from django.db.models import Q
+from django_filters import (
+    CharFilter, DateFromToRangeFilter, NumberFilter)
+from graphene_django.filter.filterset import GlobalIDMultipleChoiceFilter
+
+from ...order import models
+from ..core.filters import DistinctFilterSet
+
+
+class OrderFilter(DistinctFilterSet):
+    """
+    id - graphene type ID
+    order_id - database primary key
+    """
+    id = GlobalIDMultipleChoiceFilter(name='id', label='GraphQL ID')
+    order_id = NumberFilter(method='order_id_lookup', label='Database ID')
+    created__gte = CharFilter(
+        name='created', lookup_expr='gte', label='ISO 8601 standard')
+    created__lte = CharFilter(
+        name='created', lookup_expr='lte', label='ISO 8601 standard')
+
+    user = CharFilter(
+        method='filter_by_order_customer')
+
+    class Meta:
+        model = models.Order
+        fields = {
+            'status': ['exact'],
+            'total_net': ['exact', 'lte', 'gte']}
+
+    def order_id_lookup(self, queryset, name, value):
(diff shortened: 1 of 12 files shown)


saleor/graphql/order/filters.py
```diff
@@ -0,0 +1,39 @@
...
+from django_filters import (
+    CharFilter, DateFromToRangeFilter, NumberFilter)
+from graphene_django.filter.filterset import GlobalIDMultipleChoiceFilter
+
+from ...order import models
+from ..core.filters import DistinctFilterSet
+
+
+class OrderFilter(DistinctFilterSet):
+    """
```
patrys: This is not a valid docstring, please test with `pydocstyle`

saleor/graphql/order/resolvers.py
```diff
@@ -0,0 +1,24 @@
...
+from .types import Order
+
+
+def resolve_orders(info):
+    user = info.context.user
+    if user.is_anonymous:
+        raise PermissionDenied('You have no permission to see this')
+    if user.get_all_permissions() & {'order.view_order', 'order.edit_order'}:
+        return models.Order.objects.all().distinct()
+    return models.Order.objects.filter(user=user).distinct()
```
elwoodxblues: If I'm correct this queryset would be used for example on customer's account page in the storefront to show his orders, right? In the current account view we use this queryset: `user.orders.confirmed()`. We could reuse it here to be consistent.

saleor/graphql/order/resolvers.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def resolve_order(info, id):
+    order = get_node(info, id, only_type=Order)
+    user = info.context.user
+    if (order.user == user
+        or user.is_superuser
```
elwoodxblues: Do we have to explicitly check `is_superuser`? I think `get_all_permissions` would return all permissions anyway if the user is a superuser.

saleor/graphql/order/resolvers.py
akjanik: Correct. I will add some docstrings explaining it and take a look at this queryset.

saleor/graphql/order/types.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Order(CountableDjangoObjectType):
+    order_id = graphene.Int(
+        description='Order number.')
+
+    class Meta:
+        description = 'Represents an order in the shop.'
+        interfaces = [relay.Node]
+        model = models.Order
+        exclude_fields = []
```
elwoodxblues: We should exclude `total_net` and `total_gross` in favor of having only `total` field to access both values. The same applies to the `shipping_price` fied. `TaxedMoney` type should also give access to `tax` property.

PR #2204, opened 2018-05-23
Add ability to retrieve multiple nodes with one query

akjanik:
I'd like to have a tool to get many objects of selected type at once instead of iterating over `get_node` function like we are doing right now.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/utils.py
+++ b/saleor/graphql/utils.py
@@ -16,14 +16,14 @@ def get_node(info, id, only_type=None):
 
 
 def get_nodes(ids, graphene_type):
-    """Return queryset of nodes of proper type."""
+    """Return a list of nodes of proper type."""
     pks = []
     for graphql_id in ids:
         _type, _id = from_global_id(graphql_id)
         assert str(graphene_type) == _type, (
             'Must receive an {} id.').format(graphene_type._meta.name)
         pks.append(_id)
-    nodes =  graphene_type._meta.model.objects.filter(pk__in=pks)
+    nodes = list(graphene_type._meta.model.objects.filter(pk__in=pks))
     if not nodes:
         raise Exception(
             "Could not resolve to a nodes with the global id list of '%s'."
--- a/saleor/graphql/utils.py
+++ b/saleor/graphql/utils.py
@@ -1,6 +1,7 @@
 import graphene
 from django.utils.encoding import smart_text
 from django.utils.text import slugify
+from graphql_relay import from_global_id
 
 from ..product.models import AttributeChoiceValue, ProductAttribute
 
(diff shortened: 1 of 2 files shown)


saleor/graphql/utils.py
```diff
@@ -14,6 +15,17 @@ def get_node(info, id, only_type=None):
...
 
+def get_nodes(ids, graphene_type):
+    """Return queryset of nodes of proper type."""
+    pks = []
+    for graphql_id in ids:
+        _type, _id = from_global_id(graphql_id)
+        assert str(graphene_type) == _type, (
+            'Must receive an {} id.').format(graphene_type._meta.name)
+        pks.append(_id)
+    return graphene_type._meta.model.objects.filter(pk__in=pks)
```
patrys: Don't we care about the order? Do we want to return `None` where no match exists?

akjanik: I made this with mutations in mind, where we are providing few global_id, for example products, and want to add them to collection - in this case, for me the order does not matter.
I do not recall such case, when we care about the order.
I will add case for `None`, similar to `get_node` function.

elwoodxblues: It feels natural to return an empty list if no matches are found.

patrys: It may or may not. For example if I send a list of five Node IDs, is it natural for the API to silently only assign three of them or are there cases where a missing object is a validation error?

akjanik: @elwoodxblues Then why do we raise `Exception` in `get_node` if no node was found? I find it a bit inconsistent. For me it is more elaborate to communicate, that ids are wrong than just return empty list.

@patrys I was thinking about how to handle such case too. Do you think it would be better to remain it as it is or do another check, for example for length of input list and output queryset and raising an error if they are not equal (maybe with message, which ids were not found)?

patrys: I don't think comparing length will work. What if I pass `ids=[1, 1, 1]`?

saleor/graphql/utils.py
```diff
@@ -14,6 +15,27 @@ def get_node(info, id, only_type=None):
...
 
+def get_nodes(ids, graphene_type):
+    """Return queryset of nodes of proper type."""
+    pks = []
+    for graphql_id in ids:
+        _type, _id = from_global_id(graphql_id)
+        assert str(graphene_type) == _type, (
+            'Must receive an {} id.').format(graphene_type._meta.name)
+        pks.append(_id)
+    nodes =  graphene_type._meta.model.objects.filter(pk__in=pks)
```
patrys: Maybe cast it to a list explicitly here? Otherwise you're technically returning a QuerySet that could still be operated upon.

akjanik: Sure.

PR #2205, opened 2018-05-23
Mutation to create product image

### LINKED ISSUE #1997: Figure out file upload for the GraphQL API
How do we upload files?

Do we use `FormData` APIs to attach files as parts of a multipart POST? Do we send files as base64-encoded string variables?

elwoodxblues:
Fixes #1997 

This PR adds the ability to upload product image and answers the question of how do we support file upload in GraphQL by using multipart POST requests.

File upload itself is not an issue, but I failed to use our `ModelFormMutation` mechanisms and found it much easier to do it with a plain Graphene mutation. It worries me that our form-based mutations are not so easy to understand and customize when looking at it after some break.

Anyway, for now, I guess we the solution for the upload. I'll try to examine the complexity of ModelFormMutation separately and think of better options.

### Pull Request Checklist


elwoodxblues:
I tested the upload with `curl` (apparently none of the GUI clients that I used supports upload in GraphQL):

```curl localhost:8000/graphql/ -F query='mutation { productImageCreate(productId: "<product_id>") { productImage { id image url order} }}' -F image=@<image_path.jpg>```

**Comment by patrys** [comment]
Since we need to write dedicated forms anyway, what are the pros of sticking to form classes?

**Comment by patrys** [comment]
How easy/how hard is handling this kind of upload in Apollo?

**Comment by elwoodxblues** [comment]
@patrys I'm not sure about the complexity in Apollo, I assumed that since I'm able to manually construct the request using the standard tool and successfully upload the file, we'll somehow be able to integrate it on frontend side. I'll take a look how Apollo handles upload and see if my solution needs adjustments.

Apollo examples:
https://github.com/jaydenseric/apollo-upload-client
https://gist.github.com/tvorogme/1c0794e1455bf3fc61162e2915ec2462

File: tests/api/test_product.py [hunk 1]
```diff
@@ -778,3 +778,29 @@ def test_product_type_delete_mutation(admin_api_client, product_type):
     assert data['productType']['name'] == product_type.name
     with pytest.raises(product_type._meta.model.DoesNotExist):
         product_type.refresh_from_db()
+
+
+def test_product_image_create_mutation(admin_api_client, product):
+    query = """
+    mutation createProductImage($productId: ID!) {
+        productImageCreate(productId: $productId) {
```
akjanik: Do you think is there any way to test `file` argument here?

elwoodxblues:
I adjusted the upload mutation to the [spec](https://github.com/jaydenseric/graphql-multipart-request-spec) which is also respected by Apollo client. I had to use [this package](https://github.com/lmcgartland/graphene-file-upload), as by default Graphene doesn't support this spec.

I also had to make a change in tests for GraphQL views. The client we were using to send requests was using `multipart` content type, which worked, but conflicted with `graphql-multipart-request-spec`. Right now all standard queries are sent as `application/json`.

PR #2206, opened 2018-05-23
Merge Checkout with Cart

### LINKED ISSUE #2004: Saving checkouts in database
Checkout should have a dedicated model that would be stored in the database. This will allow other features such as:
- sharing checkouts
- creating better checkout API (multiple calls to create/edit checkout)
- tracking abandoned checkouts

mad-anne:
I want to merge this change because it closes #2004. Checkout class is removed and all data needed for that process are now saved in database within Cart model. Also, the business logic standing behind handling shipping and billing addresses in checkout is simplified.

### TODO's
* [x] Simplify and refactor existing tests
* [x] Test new functionalities
* [x] Merge checkout and cart folders, keep distinct urls

### Pull Request Checklist


elwoodxblues:
This looks good! One thing that I'm not convinced of is the file structure, particularly the nested `/cart/checkout/` directory. I think "checkout" is a more general notion and "cart" is only a part of the whole flow, so maybe this app should be called "checkout" instead?

NyanKiyoshi:
I still don't get one part of the Checkout code: why are cart session expiring after 10 years?

https://github.com/mirumee/saleor/blob/e2fb44dfcb466413bfc0c402d8b7e36f2cfcbca9/saleor/cart/utils.py#L20-L23

In France, we have a 13 months limit for cookies, so I always have to lower this value in production.

patrys:
10 years sounds... interesting. I'd say a month is more than enough for a cart. I'd also like to know why we can't store it in session instead of a separate cookie.

saleor/checkout/utils.py
```diff
@@ -1,27 +1,619 @@
...
+
+    Remove previously saved address if not connected to any user.
+    """
+    has_address_changed = (
+        not address and cart.shipping_address or
+        address and not cart.shipping_address or
+        address and cart.shipping_address and address != cart.shipping_address)
+    if has_address_changed:
+        remove_old_address = (
+            cart.shipping_address and (not cart.user or (
```
Pacu2: Not a fan of such nesting, can we introduce some variables?

saleor/checkout/utils.py
```diff
@@ -57,16 +649,162 @@ def _get_product_or_category_voucher_discount_for_checkout(voucher, checkout):
...
+        order_data['discount_amount'] = cart.discount_amount
+        order_data['discount_name'] = cart.discount_name
+
+    order = Order.objects.create(**order_data)
+
+    for line in cart:
+        add_variant_to_order(
+            order, line.variant, line.quantity, discounts, taxes)
+
+    if voucher:
```
Pacu2: Can we merge with above check?

saleor/checkout/utils.py
```diff
@@ -57,16 +649,162 @@ def _get_product_or_category_voucher_discount_for_checkout(voucher, checkout):
...
+        'shipping_method').unique_for_country_code(country_code)
+    if cart.shipping_method not in valid_methods:
+        cart.shipping_method = None
+        cart.save()
+        return False
+    return True
+
+
+@transaction.atomic
+def create_order(cart, tracking_code, discounts, taxes):
```
Pacu2: This function is almost 80 lines long, can we introduce some type of helpers here?

saleor/checkout/utils.py
```diff
@@ -1,27 +1,619 @@
...
+
+    return addresses_form, address_form, updated
+
+
+def change_billing_address_in_cart(cart, address):
+    """Save billing address in cart if changed.
+
+    Remove previously saved address if not connected to any user.
+    """
+    has_address_changed = (
```
Pacu2: can we introduce some kind of variables, or use `all`?  We are checking presence of 7 variables here

saleor/checkout/utils.py
```diff
@@ -1,27 +1,619 @@
...
+        'total_with_shipping': total_with_shipping,
+        'local_total_with_shipping': local_total_with_shipping}
+
+
+def find_open_cart_for_user(user):
+    """Find an open cart for the given user."""
+    carts = user.carts.all()
+    if len(carts) > 1:
+        logger.warning('%s has more than one open basket', user)
+        for cart in carts[1:]:
```
Pacu2: Just checking, the first cart is the newest? Or shall we order above queryset by pk/created/last modified?

saleor/checkout/utils.py
mad-anne: I will extract a separate function checking if an address has changed and should be removed.

saleor/checkout/utils.py
mad-anne: Yes, I don't see anything against that.

saleor/checkout/utils.py
mad-anne: You are right, it was hard to read. I added some helpers.

saleor/checkout/utils.py
mad-anne: As above, I moved it to a new function.

saleor/checkout/utils.py
mad-anne: First one is the newest - we track it by Cart's `last_change` field.

PR #2226, opened 2018-05-29
Fix storyshots

dominik-zeglen:
Due to overlook, snapshots taken in storyshots were too shallow, testing completely nothing. This PR is an attempt to fix it.

### DIFF
--- a/saleor/static/dashboard-next/storybook/Stories.test.ts
+++ b/saleor/static/dashboard-next/storybook/Stories.test.ts
@@ -8,6 +8,8 @@ configure({ adapter: new Adapter() });
 
 initStoryshots({
   configPath: "saleor/static/dashboard-next/storybook/",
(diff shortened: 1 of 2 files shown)


patrys:
Maybe we should use the regular renderer instead of the shallow one?

dominik-zeglen:
Consider situation, where the DOM tree has 4 levels:
```
A
|---B
    |---C
        |---D
```
Now let's say that component D has changed (unintended change) and you run snapshot tests.
With shallow renderer, you get one clear message that something has changed in component D's DOM. When using regular, you get error bloat from all 4 components, 3 of them being completely meaningless. Now imagine what would happen if not one but a couple of low-level components had been changed.

I'd stay with shallow renderer, beacuse it provides a lot better problem understanding than simply informing developer that something is not right.

PR #2229, opened 2018-05-29
More descriptive shipping methods list

### LINKED ISSUE #2216: Missing informations while choosing shipping method for order
When creating order from dashboard and choosing shipping method, we would like to have information about shipping country. For example, one may same shipping provider, but different prices for different countries.
image

mad-anne:
I want to merge this change because it closes #2216.

The list of shipping methods is restricted only to methods available in the delivery country (which is took from a shipping address). If none is provided, then all methods are returned. This can end up with multiple inserts from the same provider. To make a displayed destination more descriptive, a country name is added now. Also sorting is improved - first by shippng method name and then by price.

### Pull Request Checklist

### DIFF
--- a/saleor/dashboard/order/views.py
+++ b/saleor/dashboard/order/views.py
@@ -717,7 +717,7 @@ def change_fulfillment_tracking(request, order_pk, fulfillment_pk):
 def ajax_order_shipping_methods_list(request, order_pk):
     order = get_object_or_404(Order, pk=order_pk)
     queryset = ShippingMethodCountry.objects.select_related(
-        'shipping_method').order_by('price').all()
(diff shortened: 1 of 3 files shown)


saleor/dashboard/order/views.py
```diff
@@ -717,7 +717,7 @@ def change_fulfillment_tracking(request, order_pk, fulfillment_pk):
 def ajax_order_shipping_methods_list(request, order_pk):
     order = get_object_or_404(Order, pk=order_pk)
     queryset = ShippingMethodCountry.objects.select_related(
-        'shipping_method').order_by('price').all()
+        'shipping_method').order_by('shipping_method__name', 'price')
```
NyanKiyoshi: Shouldn't we index it then?

PR #2243, opened 2018-06-01
Fix to allow viewing and editing of stock for non-variant products

lampyjon:
A previous change appears to have removed the ability to view or edit stock if a product doesn't have variants. I think this is the right location to add in the required functionality based on the variant dashboard code, but apologies if there is a better approach!

### DIFF
--- a/templates/dashboard/product/detail.html
+++ b/templates/dashboard/product/detail.html
@@ -231,6 +231,39 @@
                 </table>
               </div>
             </div>
+      {% if no_variants %}
+          <div class="card">
+            <div class="data-table-header">
+              <div class="data-table-title">
+                <h5>
+                  {% trans "Inventory" context "Product inventory card header" %}
+                </h5>
+              </div>
+            </div>
+            <div class="data-table-container">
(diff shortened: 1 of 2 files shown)


NyanKiyoshi:
Could you align tags and statements to the rest of the code?

templates/dashboard/product/detail.html
```diff
@@ -231,6 +231,39 @@ <h6>
                 </table>
               </div>
             </div>
+      {% if no_variants %}
```
NyanKiyoshi: 1. Could you keep the same contexts as the variant's table to avoid duplicating strings?

https://github.com/mirumee/saleor/blob/a54c91605bd0def1eb052bdbef83c16103d2adab/templates/dashboard/product/product_variant/detail.html#L145-L173

2. Please keep the code aligned from the rest (as said before, just a reminder);
3. Instead of duplicating all this code, I think it would be best to create a new snippet and use a include instead (using the `with` keyword) in variant details template, and for this one here.

elwoodxblues:
This was resolved in #2233

PR #2247, opened 2018-06-03
Feature: allow to disable stock management

NyanKiyoshi:
Closes https://github.com/mirumee/saleor/issues/1979.

This adds a new field `track_inventory` on the variant model, which toggles to inventory management (allocations and checks).

This also adds a new settings (on the SiteSettings model), that allows to set the default inventory tracking for the variant creation form.

It's only there to allow users to save time and prevent them from accidentally setting an unwanted inventory management.

### Pull Request Checklist


saleor/product/models.py
```diff
@@ -185,6 +185,11 @@ class ProductVariant(models.Model):
         Product, related_name='variants', on_delete=models.CASCADE)
     attributes = HStoreField(default={}, blank=True)
     images = models.ManyToManyField('ProductImage', through='VariantImage')
+    handle_stock = models.BooleanField(
```
elwoodxblues: Although I understood the purpose of this field from reading the code, this name sounds a bit too general to me. What about `track_inventory`? I think it would be more understandable.

saleor/product/models.py
```diff
@@ -185,6 +185,11 @@ class ProductVariant(models.Model):
...
+        default=True,
+        help_text=pgettext_lazy(
+            'product variant handle stock field help text',
+            'Allow or not to automatically handle stock units from orders'))
```
elwoodxblues: We tend to keep help texts in forms rather than in models. Also, if we changed `handle_stock` to `track_inventory` we would have to rephrase it to something like: "Automatically track this product's inventory".

saleor/site/models.py
```diff
@@ -22,6 +22,12 @@ class SiteSettings(models.Model):
     include_taxes_in_prices = models.BooleanField(default=True)
     display_gross_prices = models.BooleanField(default=True)
     charge_taxes_on_shipping = models.BooleanField(default=True)
+    handle_stock_by_default = models.BooleanField(
+        default=True,
+        help_text=pgettext_lazy(
```
elwoodxblues: Again, this help_text should be moved to the form. I think we need to explain precisely what this feature does, in a user-friendly way. I'm thinking about a checkbox with descriptive label and help_text:
- label: "Enable inventory tracking for newly created products."
- help text: "Inventory tracking may be modified at the product-level. Changing this setting won't affect existing products."

What do you think?

NyanKiyoshi:
@elwoodxblues all good for me, I will do the changes. Thanks!

Issue #2222, opened 2018-05-29
Add endpoints for menus


elwoodxblues:
Add GraphQL endpoints for handling storefront's menus.

Queries:
- [x] menu($id)
- [x] menus()

Mutations:
- [x] menuCreate()
- [x] menuUpdate()
- [x] menuDelete()
- [x] menuItemCreate($menuId: $ID)
- [x] menuItemUpdate($id: $ID)
- [x] menuItemDelete($id: $ID)

We will also need a mutation to assign top and footer menus, but that's related to the `SiteSettings` model.

PR #2256, opened 2018-06-06
Menu endpoints

akjanik:
It is a part of #2222. It is yet to decide how to implement mutations - sticking up with old mutation forms or wait for a new shiny mutation with `<Type_name>Input`.

### Pull Request Checklist


saleor/graphql/api.py
```diff
@@ -50,6 +53,16 @@ class Query(graphene.ObjectType):
         Collection, query=graphene.String(
             description=DESCRIPTIONS['collection']),
         description='List of the shop\'s collections.')
+    menu = graphene.Field(
+        Menu, id=graphene.Argument(graphene.ID),
+        description='Lookup a menu by ID.')
+    menus = DjangoFilterConnectionField(
+        Menu, description="List of the shop\'s menus.")
+    menu_item = graphene.Field(
```
elwoodxblues: I'm not sure if we need to access menu items directly, do we? I guess it would be enough and more common to fetch the menu object and access its items from there.

akjanik: As far as I remember we had a discussion about it and we stated that we may want to have this endpoint, so we can open `Edit menu` view and assign items created before to the menu - similar to how we currently manage `attributes` of `Product Types`.
Did our assumptions change?

elwoodxblues: I guess they didn't. It's fine if there's a valid use-case.

saleor/graphql/menu/types.py
```diff
@@ -0,0 +1,23 @@
...
+from ...menu import models
+from ..core.types import CountableDjangoObjectType
+
+
+class Menu(CountableDjangoObjectType):
+    class Meta:
+        description = """Represents a single menu - an object that is used
+        to help navigate through the store."""
+        interfaces = [relay.Node]
+        filter_fields = ['id', 'name']
```
elwoodxblues: Do we need the filtering by `id` here, since we have an endpoint to fetch a particular menu? Also, since we've introduced `query` in some of the other queries, I'd prefer to stay consistent with them and drop filtering by `name` in favor of `query`. We could stick to that convention in all queries.

saleor/graphql/menu/types.py
```diff
@@ -0,0 +1,23 @@
...
+        model = models.Menu
+
+
+class MenuItem(CountableDjangoObjectType):
+    class Meta:
+        description = """Represents a single item of the related menu.
+        Can store categories, collection or pages."""
+        interfaces = [relay.Node]
+        exclude_fields = ['lft', 'rght', 'tree_id']
+        filter_fields = ['id', 'name']
```
elwoodxblues: The same comments as for Menu applies here.

saleor/graphql/menu/types.py
```diff
@@ -0,0 +1,23 @@
...
+class Menu(CountableDjangoObjectType):
+    class Meta:
+        description = """Represents a single menu - an object that is used
+        to help navigate through the store."""
+        interfaces = [relay.Node]
+        filter_fields = ['id', 'name']
+        model = models.Menu
+
+
+class MenuItem(CountableDjangoObjectType):
```
elwoodxblues: One problem with this type is that the consumer of the API has to fetch all of the related objects (page, collection, category) in order to get the actual link. Instead, the type should look more or less like this:

```
type MenuItem {
  id: ID!
  title: String!
  url: String!
}
```

where `url` is automatically derived from the model by the backend. The client shouldn't have to worry about which relation is actually when he only needs the proper URL.

saleor/graphql/menu/mutations.py
```diff
@@ -0,0 +1,134 @@
...
+    def user_is_allowed(cls, user, input):
+        return user.has_perm('menu.edit_menu')
+
+    @classmethod
+    def clean_input(cls, info, instance, input, errors):
+        cleaned_input = super().clean_input(info, instance, input, errors)
+        items = [
+            cleaned_input.get('page'), cleaned_input.get('collection'),
+            cleaned_input.get('url'), cleaned_input.get('category')]
+        items = list(filter(None.__ne__, items))
```
elwoodxblues: A more idiomatic way in Python would be to use a list comprehension instead of `filter`.

saleor/graphql/menu/mutations.py
```diff
@@ -0,0 +1,134 @@
...
+        if len(items) > 1:
+            field = 'items'
+            msg = 'More than one item provided.'
+            cls.add_error(errors=errors, field=field, message=msg)
```
elwoodxblues: Variables `field` and `msg` doesn't seem necessary here, the code would be clean enough without them.

saleor/graphql/menu/mutations.py
```diff
@@ -0,0 +1,134 @@
...
+    category = graphene.ID(description='Category to which item points.')
+    collection = graphene.ID(description='Collection to which item points.')
+    page = graphene.ID(description='Page to which item points.')
+
+
+class MenuCreate(ModelMutation):
+    class Arguments:
+        input = MenuInput(
+            required=True,
+            description='Fields required to create a menu.')
```
elwoodxblues: This description (as well as the one in "update" mutation below) needs more explanation, that only one of the four possible items will be respected as a related one.
Maybe this input should take only `url` or `id` - if ID is provided we query against the supported types of related items in the resolver or use the given `url` as a target.

saleor/graphql/menu/types.py
akjanik: I am not sure about this one. `query` was intended to search in _many_ fields like name, address, country etc, for a quick filter. Here we have one field, so I would go for `name__Icontains` - it is technically the same, but we remain consistent at least.

elwoodxblues: In my opinion, `query` sounds like a good name for any kind of text-based searching/matching/filtering - it's easy to understand when you look at it and definitely more user-friendly than `name__Icontains`. To be honest I'm not a fan of this `django-filters` integration, mostly because of the ugly looking argument names it generates. Maybe we could customize it. Also, I've read that a good practice is to keep the arguments list as short as possible.

saleor/graphql/menu/mutations.py
akjanik: I have changed description in `MenuItemCreate/Update` - here you only need to provide menu name.

PR #2257, opened 2018-06-06
Minor fixes

dominik-zeglen:
I want to merge this change because it fixes `<DateFormatter />` snapshots and adds missing `key` prop to `<TimelineNode />` components in order history section.


patrys:
Let's rebase and update the stories once we merge #2258 as it replaces most if not all snapshots.

PR #2263, opened 2018-06-07
Add `customers` app

dominik-zeglen:
Quick preview:

zrzut ekranu 2018-06-07 o 16 29 41
zrzut ekranu 2018-06-07 o 16 30 01
zrzut ekranu 2018-06-07 o 16 30 20
zrzut ekranu 2018-06-07 o 16 30 12


Pacu2:
Isn’t “price” label adjusted to the right, while all the others are aligned to the left?

Not sure if that’s intended or not.

dominik-zeglen:
It is planned. Every numeric ( therefore comparable ) values should be aligned to the right, so should be its label. It is easier to read and compare values this way.

mirekm:
Including `#` @dominik-zeglen? ;)

dominik-zeglen:
Okay, numeric and comparable :)

PR #2273, opened 2018-06-11
Refactor mutations

elwoodxblues:
This PR aims to refactor ModelFormMutation to a simpler mechanism which doesn't use forms under the hood.

### Pull Request Checklist


saleor/graphql/core/mutations.py
```diff
@@ -65,146 +47,159 @@ def _update_mutation_arguments_and_fields(cls, arguments, fields):
...
         if not model:
-            raise ImproperlyConfigured(
-                'model is required for ModelDeleteMutation')
-
-        _meta = ModelDeleteMutationOptions(cls)
-
-        if not return_field_name:
-            return_field_name = get_model_name(model)
+            raise ImproperlyConfigured('model is required for ModelMutation')
+        _meta = ModelMutationOptions(cls)
```
akjanik: Do we have to provide `_meta=None` argument if it is being overwritten here anyway?

saleor/graphql/core/mutations.py
```diff
@@ -65,146 +47,159 @@ def _update_mutation_arguments_and_fields(cls, arguments, fields):
...
-        abstract = True
+    @classmethod
+    def _clean_input(cls, info, instance, input, errors):
+        InputCls = getattr(cls.Arguments, 'input')
+        cleaned_input = {}
+        for field_name, field in InputCls._meta.fields.items():
+            if field_name in input:
+                value = input[field_name]
+                if value is not None:
+                    # FIXME: maybe we could have custom input field type that takes
```
akjanik: This may be an inspiration
https://github.com/graphql-python/graphene-django/blob/master/graphene_django/forms/forms.py

saleor/graphql/core/mutations.py
```diff
@@ -65,146 +47,159 @@ def _update_mutation_arguments_and_fields(cls, arguments, fields):
...
-        model = cls._meta.model
-        model_type = registry.get_type_for_model(model)
-        instance = get_node(info, id, only_type=model_type)
-        instance.delete()
-        field_name = cls._meta.return_field_name
-        kwargs = {field_name: instance}
-        return cls(**kwargs)
+    def add_error(cls, errors, field, message):
+        errors.append(Error(field=field, message=message))
+        return errors
```
akjanik: Is a return needed here?

saleor/graphql/core/mutations.py
```diff
@@ -65,146 +47,159 @@ def _update_mutation_arguments_and_fields(cls, arguments, fields):
...
-        """
-        form_kwargs = cls.get_form_kwargs(root, info, **kwargs)
-        form = cls._meta.form_class(**form_kwargs)
-        cls._form = form
-        if form.is_valid():
-            instance = form.save()
-            return instance
-        return None
+    def user_is_allowed(cls, user, input):
+        return True
```
akjanik: I think there should be some kind of check here.

saleor/graphql/core/mutations.py
```diff
@@ -65,146 +47,159 @@ def _update_mutation_arguments_and_fields(cls, arguments, fields):
...
+        if hasattr(field.type, 'of_type'):
+            return field.type.of_type == target_type
+        return field.type == target_type
 
-class ModelFormMutation(BaseMutation):
-
-    class Meta:
-        abstract = True
+    @classmethod
+    def _clean_input(cls, info, instance, input, errors):
```
akjanik: I think argument `errors` should be dropped from here. It is not being used here apart from being returned in the same way as while called.
For me it is a bit misleading, for example in `mutate` method below, `errors = []` is being initialised, then this function is called, which does nothing and then there is a call to `_clean_instance` which checks for errors.

saleor/graphql/product/mutations.py
```diff
@@ -2,112 +2,139 @@
...
-            products = set(get_nodes(product_ids, Product))
-            kwargs['data']['products'] = products
-        return kwargs
+    def user_is_allowed(cls, user, input):
+        return user.has_perm('collection.edit_collection')
 
 
-class CollectionUpdate(StaffMemberRequiredMixin, ModelFormUpdateMutation):
-    permissions = 'collection.edit_collection'
+class CollectionUpdate(ModelMutation):
```
akjanik: From what I see, right now there are two ways to add products to collection: using this mutation and `CollectionAddProducts` below.
Maybe one could modify `_clean_input` to remove proper key from dict; or, we decide that we update list of products every time, by dropping `CollectionAddProducts`.

saleor/graphql/core/mutations.py
elwoodxblues: The idea is that at any step of the mutation execution you are able to add your own errors. It's not used in the original implementation, but if you customized `clean_input` and you had some custom conditions to validate the data, you'd be able to do this.

saleor/graphql/core/mutations.py
elwoodxblues: I'll add docs to these methods.

saleor/graphql/core/mutations.py
elwoodxblues: We can get rid of it.

saleor/graphql/core/mutations.py
elwoodxblues: I'll take a look. I have some ideas how to do this but I want to do it separately, to not block this PR anymore.

saleor/graphql/core/mutations.py
elwoodxblues: Edit: at any "validation" step so `clean_input` or `clean_instance`.

PR #2277, opened 2018-06-12
Various changes and fixes on storefront user deletion feature

NyanKiyoshi:
This change requires the user to post to confirm account deletion instead of a simple click/ get. Which as it was discussed, could be exploitable through CSRF, or could be triggered by anti-spam or anti-virus mechanism or browser preloading.

This also redirect the user to the homepage with a success message.

### Remaining things to check/ discuss about
- [x] Require the user to be logged in before deleting (to prevent users deleting others' accounts);
- [x] Migration issue setting the same token to everyone.

### Pull Request Checklist


Pacu2:
@NyanKiyoshi Can we align the content of the account_prompt_delete to the center?

NyanKiyoshi:
My local assets were outdated, that was why; sorry about that!

@Pacu2 not sure why we were passing a `pk` to the 'request delete account' endpoint... I removed it as we are only expecting the current user.

PR #2285, opened 2018-06-18
Update products.rst

nobism:
Adding package size variant to the coffee example, this was it's consistent with the game example which lists all the game's variants.

I want to merge this change because... I think the examples should be consistent.

### Pull Request Checklist

### DIFF
--- a/docs/architecture/products.rst
+++ b/docs/architecture/products.rst
@@ -104,11 +104,11 @@ Your shop sells Coffee from around the world. Customer can order 1kg, 500g and 2
 
 .. table:: Product
 
-   ============  ================  =================  =================================
-   Product type  Name              Country of origin  Description
-   ============  ================  =================  =================================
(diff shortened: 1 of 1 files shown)


elwoodxblues:
Hi @nobism, thanks for the PR! I think what we have now is correct - "Country of origin" is an attribute at the product level, while "Package size" is specific to variants (look again at the product type "Coffe"). That's why in the table with product data we have only one of them, while the other is in the table with variants. That's how it works in Saleor.

On the other hand, in case of "game" example, both attributes are defined at the product level (again, look at the product type definition) - that's why we included them all in the table with example product data.

elwoodxblues:
I agree it's not that clear from the beginning. It's on our list to make the UX of product management views more intuitive so that the whole concept is better understood. We'll address this issue in the forthcoming Dashboard 2.0.

Issue #2189, opened 2018-05-19
No payment status update on the order page.


maltitco:
Example:
![nagranie ekranu 2018-05-19 o 12 08 49](https://user-images.githubusercontent.com/1561204/40267570-eb763ace-5b5e-11e8-8621-63b8aa3953dd.gif)

maarcingebala:
Fixed: #2348

PR #2288, opened 2018-06-18
fixes issues: #2189

maltitco:
Now in the right order panel, displayed is the last payment status.

### DIFF
--- a/templates/dashboard/order/detail.html
+++ b/templates/dashboard/order/detail.html
@@ -111,7 +111,7 @@
             {% trans "Payments" context "Order summary card header" %}
           </h5>
           {% if payment %}
-            {% render_status payment.status payment.get_status_display %}
+            {% render_status order.get_last_payment_status order.get_last_payment_status_display %}
           {% else %}
             <span class="label">
               {% trans "N/A" context "Payment status not available" %}


templates/dashboard/order/detail.html
```diff
@@ -111,7 +111,7 @@ <h5 class="data-table-title">
             {% trans "Payments" context "Order summary card header" %}
           </h5>
           {% if payment %}
-            {% render_status payment.status payment.get_status_display %}
+            {% render_status order.get_last_payment_status order.get_last_payment_status_display %}
```
elwoodxblues: I think we should instead make sure that the `payment` instance which is passed from the view is the right one. Above there is the `{% if payment %}` check which is here on purpose. Could you instead see what's going on in the `order_details` view, that the returned payment is incorrect?

maltitco: OK, I will try to check.

maltitco: @elwoodxblues Thanks for the tips of what to look for.
In `order_details` view, we have `payment = order.payments.last()` and it should be `payment = order.payments.first()` because the query set sorts the payment from the latest to the oldest one - which is confirmed by the tab with the list of payments.

elwoodxblues: In order/models.py we also have functions `get_last_payment_status` and `get_last_payment_status_display` - they're used to display the payment status in the orders list. If you look at the code they both use `max()` to determine the most recent ID. This seems safe as it doesn't depend on the sorting order. Could we have a simple function `order.get_last_payment()` that would do the same logic to fetch the last payment that user has made? Then we could reuse it in those two for getting the status.

elwoodxblues: The problem with `queryset.first()` and `queryset.last()` is that it depends on the sorting order defined in the metaclass. If someone changes it in the future than the template could break again. If we had this small function, we could protect ourselves from that.

maltitco: Can we, however, use my PR proposal?
I know that there is `{% if payment %}`, but if it is a payment, it will display the correct order status, if there is no payment will go to `{% else %}`.

maltitco: I can be wrong and if it is, please let me know which way to go to solve this problem.

elwoodxblues: Let's keep the template unchanged and only modify [this line](https://github.com/mirumee/saleor/blob/master/saleor/dashboard/order/views.py#L124) in the view to get the last payment properly (as it it's done in `get_last_payment_status`).

maarcingebala:
This issue was fixed in #2348.

PR #2305, opened 2018-06-22
Helpful changes to docs and settings for new devs

jfialkoff:
This is a reboot of #1885 after having removed instructions to install Postgres and Node via Homebrew. Left in are instructions for creating a new database and role, and modification to `ALLOWED_HOSTS`.

It eases setup and prevents a possible obstacle from surfacing.

### Pull Request Checklist


saleor/settings.py
```diff
@@ -340,7 +340,8 @@ def get_host():
 
 TEST_RUNNER = ''
 
-ALLOWED_HOSTS = get_list(os.environ.get('ALLOWED_HOSTS', 'localhost'))
+ALLOWED_HOSTS = get_list(os.environ.get('ALLOWED_HOSTS',
+                                        'localhost,127.0.0.1'))
```
NyanKiyoshi: Could you put this `os.environ.get(...)` on a new line? Exactly like this:
```python
ALLOWED_HOSTS = get_list(
    os.environ.get('ALLOWED_HOSTS', 'localhost,127.0.0.1'))
```

docs/gettingstarted/installation-macos.rst
```diff
@@ -18,7 +18,18 @@ PostgreSQL
 
 Saleor needs PostgreSQL version 9.4 or above to work. Get the macOS installer from the `PostgreSQL download page <https://www.postgresql.org/download/macosx/>`_.
 
-Make sure you keep track of the password you set for the administration account during installation.
+Then, create a new database and user. Make sure you keep track of the password you set for the administration account during installation.
```
NyanKiyoshi: Maybe you could change to "Then, you **will** need to create a new database and user.", letting know the user further instructions and code blocks are coming.

NyanKiyoshi: > Make sure you keep track of the password you set for the administration account during installation.

What do you mean? To make sure to remember the password to access the database (e.g.: development tools)?

docs/gettingstarted/installation-macos.rst
```diff
@@ -18,7 +18,18 @@ PostgreSQL
...
+.. code-block:: console
+
+ $ psql postgres
+
+and from the Postgres prompt:
+
+.. code-block:: psql
+
+ # CREATE ROLE saleor WITH LOGIN PASSWORD 'saleor' SUPERUSER;
+ # CREATE DATABASE saleor;
```
NyanKiyoshi: I would put the two instructions into two separated blocks for better understanding and structure:
1. Create the database;
2. Create the superuser with a password.

jfialkoff:
Hmm, I see that there's now some instructions near the bottom for creating a new Postgres user. I don't see instructions on how to create the DB. It seems to me to make sense to group the database creation with the installation of Postgres but I could see it either way. Let me know your preference and I'll update.

docs/gettingstarted/installation-macos.rst
NyanKiyoshi: You should not run those commands as root but as `postgres` user. Or it doesn't work that way on MacOS?

patrys: On macOS you don't need a special system user.

The default prompt after issuing `psql postgres` is `postgres=#` though so I'm not sure `=>` is a good prompt.

NyanKiyoshi: Okay then!

saleor/settings.py
jfialkoff: Sure

docs/gettingstarted/installation-macos.rst
jfialkoff: Actually, I'm a little confused here too. I install Postgres via homebrew. With homebrew there's no setting up an administration account. Are we certain that this occurs when installing via the installer? We might want to remove the whole "Make sure ..." part.

jfialkoff:
I updated the commit to reflect the changes we've discussed.

docs/gettingstarted/installation-macos.rst
```diff
@@ -121,6 +118,7 @@ Installation
 
    .. code-block:: console
 
+    $ createdb saleor
```
NyanKiyoshi: Could you move it to its own section (as `6. Create the database`)?

patrys: I'd not create the database until the role exists as you'd normally pass the role as `--owner` to `createdb`.

PR #2308, opened 2018-06-22
Enhance Dockerfile to use multi-stage build

fpytloun:
This will build all assets and virtualenv and result in image that is
ready to be used for production use and has smallest possible size
(478MB).

I didn't tested docker-compose.yml changes much but I suppose it should be working for development in the same manner as before this change.

### Pull Request Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,8 @@ ENV PYTHONUNBUFFERED 1
 
 RUN \
   apt-get -y update && \
-  apt-get install -y gettext python3-virtualenv virtualenv && \
+  apt-get install -y gettext virtualenv && \
+  # Cleanup apt cache
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,56 @@
-FROM python:3.6
+### Build python virtualenv
+FROM python:3.6-stretch as build-python
 ENV PYTHONUNBUFFERED 1
 
 RUN \
   apt-get -y update && \
-  apt-get install -y gettext && \
-  apt-get clean
+  apt-get install -y gettext python3-virtualenv virtualenv && \
+  apt-get clean && \
+  rm -rf /var/lib/apt/lists/*
 
+RUN virtualenv -p python3 /app/virtualenv
 ADD requirements.txt /app/
-RUN pip install -r /app/requirements.txt
+RUN /app/virtualenv/bin/pip install -r /app/requirements.txt
+
+### Build static assets
+FROM node:8.6.0 as build-nodejs
+# Install node_modules
+ADD webpack.config.js app.json package.json package-lock.json /app/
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,18 +1,56 @@
-FROM python:3.6
+### Build python virtualenv
+FROM python:3.6-stretch as build-python
 ENV PYTHONUNBUFFERED 1
 
 RUN \
   apt-get -y update && \
-  apt-get install -y gettext && \
-  apt-get clean
+  apt-get install -y gettext python3-virtualenv virtualenv && \
```
NyanKiyoshi: Why installing both `python3-virtualenv` and `virtualenv`? `virtualenv` alone should be enough (which depends on 	`python3-virtualenv`).

Dockerfile
```diff
@@ -1,18 +1,56 @@
...
+  apt-get clean && \
+  rm -rf /var/lib/apt/lists/*
```
NyanKiyoshi: You could maybe add a comment saying `empty the APT cache`?

patrys:
I'm not sure what to make of that. One one hand it provides a usable image after just one build. On the other hand:
* I'm not sure what we gain from introducing virtualenvs inside a container.
* We hardly ever deploy static assets as part of the docker container. They usually end up on S3 or a similar service and are served by the load balancer. It's a waste of resources to use a synchronous Python worker to serve all traffic.
* It looks like the entire virtualenv will be rebuilt each time Webpack needs to rebundle assets. This combined with the above makes me reluctant.

The build process becomes very complex and we don't seem to gain much. In the past we've discussed adding a separate docker-compose container to run `webpack --watch`.

patrys:
@fpytloun Thank you for your first contribution to Saleor, please see above for some things I'd like to discuss before we decide how to proceed with your changes.

Dockerfile
fpytloun: This was required in jessie, you are right that it should not be needed now

Dockerfile
fpytloun: Ok

patrys:
@fpytloun Do you have an example of what dev tooling is skipped when using virtualenv? I'm not sure I follow. All of our tooling is installed from `requirements.txt` so it would also end up in the virtualenv?

fpytloun:
Image that's used to build virtualenv is python:3.6-stretch vs. final image is build using python:3.6-slim-stretch.
The first one contains build tools (pip, dev libs, etc.) so it is bigger. Second one doesn't have these tools but ships only bare interpreter. Check the size difference:
```
python                         3.6-slim-stretch    29ea9c0b39c6        7 weeks ago         138MB
python                         3.6-stretch         29d2f3226daf        7 weeks ago         912MB
```

So my image is 138MB + virtualenv size + assets size = only 478MB
When using `3.6-slim-stretch` and avoid using virtualenv, it would be over 1GB.

PR #2312, opened 2018-06-23
Fixed wrong references to the cart namespace

### LINKED ISSUE #2311: Checkout with an empty cart generates infinite HTTP redirection

1. Checkout with an empty cart ([one click example](https://demo.getsaleor.com/en/checkout/)).

### What I expected to happen
Be redirect to the cart (as it was some time ago).

NyanKiyoshi:
The merge of checkout and cart (#2206) missed some references that were targeting the cart namespace instead of the checkout namespace. This fixes that.

It also fixes the infinite HTTP redirection happening when the user was checking out without a cart (closes #2311).

### Pull Request Checklist


maarcingebala:
Thank you @NyanKiyoshi, good catch!

PR #2332, opened 2018-06-29
Add `attributes` section

dominik-zeglen:
I want to merge this change because it adds `attributes` section to the new dashboard.

### Screenshots
zrzut ekranu 2018-06-29 o 17 54 31
zrzut ekranu 2018-06-29 o 17 54 37
zrzut ekranu 2018-06-29 o 17 54 42
zrzut ekranu 2018-06-29 o 17 54 49
zrzut ekranu 2018-06-29 o 17 55 01
zrzut ekranu 2018-06-29 o 17 55 07
zrzut ekranu 2018-06-29 o 17 55 13


patrys:
What if adding a new value was as easy as clicking the last row that says "New value" in grey?

NyanKiyoshi:
Could be a simple text input field with a submit button that can be triggered by hitting 'enter' and that doesn't lose focus when hit 'enter' (submit).

So we can add quickly all the values we want without ever touching the mouse.

Issue #2353, opened 2018-07-03
Documentation of "Docker for Development" not working


carstenblt:
Forgive me if this is too obvious, I'm completely new to Docker and this took me a few minutes to figure out. However, the documentation says:

5. Prepare the database

$ docker-compose run web python manage.py migrate
$ docker-compose run web python manage.py collectstatic
$ docker-compose run web python manage.py populatedb --createsuperuser

Which does not work. What works is 
$ docker-compose run web /app/virtualenv/bin/python manage.py ...

since the container is setup with a virtualenv.

nobism:
Is your `virtualenv` activated? If so then I think it should have worked, otherwise you're just specifying exactly which python interpreter to use. If you run `python --version`, `python3 --version` and `python3.6 --version` you might get 3 different answers, that's the main reason why `virtualenv` was created.

On a related note, I'm trying to run up Saleor using the production instructions, but I have a feeling that they're incomplete as well. I'm also new to Docker so it's taking me some time to work out what's going on.

carstenblt:
Well, running the container with
$ docker-compose run xxx
apparently does not activate the environment. So either it's specifying the correct interpreter or changing the Docker container to do that automatically - which I don't know how to do.
$ docker-compose up
however does switch to the virtualenv correctly, the server starts up without problems.

nobism:
What error message(s) were you getting exactly? Also, do you happen to have a folder called `virtualenv` in the project's base directory? The `Dockerfile` says to copy everything from the base directory into the container, that's the only way that you could possibly run python from `/app/virtualenv/...`.

carstenblt:
Oh, yeah:
`Traceback (most recent call last):
  File "manage.py", line 8, in <module>
    from django.core.management import execute_from_command_line
ModuleNotFoundError: No module named 'django'`

No, and as far as I can tell the Dockerfile does not copy anything but runs pip within the container.

patrys:
This is a valid problem and was introduced by #2308

carstenblt:
#2357 fixes it for me.

maarcingebala:
Should be resolved by #2362.

PR #2357, opened 2018-07-04
Fixing issues introduced with new Docker changes

ilanponimansky:
Relates to https://github.com/mirumee/saleor/issues/2353

In https://github.com/mirumee/saleor/pull/2308, useful Docker features were introduced however this broke migrations, as well as introducing, in my opinion, unnecessary usage of virtualenv. This PR removes virtualenv which fixes migrations. 

I have confirmed migrations run and I am able to launch the site without issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,9 +9,9 @@ RUN \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
-RUN virtualenv -p python3 /app/virtualenv
 ADD requirements.txt /app/
-RUN /app/virtualenv/bin/pip install -r /app/requirements.txt
+RUN mkdir -p /app/pip && pip install -r /app/requirements.txt -t /app/pip
+
 
 ### Build static assets
 FROM node:8.6.0 as build-nodejs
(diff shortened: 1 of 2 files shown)


nobism:
Would this require updates to the development and/or deployment instructions for Docker? If so, could they please be updated? That way maybe more people can help out if they have a clearer understanding of the procedure for setting up Docker.

carstenblt:
@nobism This actually fixes the documentation again, see #2353.

ilanponimansky:
The fundamental procedures have not changed so nothing is needed from the end-user or developer. I can't think of a scenario where this will break current implementations. 

The documentation does need to be updated IMO, however that is outside the scope of this PR. I will open another one with my updates in the near future.

NyanKiyoshi:
#2360 should fix the issue as well if you can take a look, it would be better to keep the virtual environment to stay isolated from the OS.

ilanponimansky:
I see what you are saying about virtualenv and in general, it is better to use it however I question the benefits of it in a container, which is already as isolated as you will get.  Seems like it just adds unnecessary complexity.

NyanKiyoshi:
It always a better practice to isolate the working python environment from the OS's Python environment as the system is also using Python (especially as Debian is using quite a lot Python).

It also keeps the docker environment cleaner.

patrys:
@NyanKiyoshi I don't think that's the case with containerized apps. We choose a very specific version of the Python image for a reason. I don't see much value in separating the app from the system since in this case the app is very much the entire system.

patrys:
I've submitted a slightly different solution to the same problem that skips virtualenv creation and allows unprefixed binaries to be used.

maarcingebala:
Fixed in #2362

PR #2370, opened 2018-07-08
Use safe filter on product.description

EmlynC:

Parse product description through safe filter, else it renders the HTML tags

### Screenshots

Before:
image

After:
image

### Pull Request Checklist


maarcingebala:
Thanks, I thought we had that covered when we were adding the HTML editor, but maybe the issue was introduced accidentally later.

PR #2391, opened 2018-07-12
Convert snake_case field names to camelCase

### LINKED ISSUE #2386: Mutation form errors should be always returned in camelCase
There are situations, where field names in `errors` property are written in `snake_case` instead of `camelCase`. Most likely it happens because of Django's form validation returning raw field name instead of camelCased.

akjanik:
Fixes #2386 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/core/mutations.py
+++ b/saleor/graphql/core/mutations.py
@@ -90,6 +90,7 @@ class ModelMutation(BaseMutation):
         As a result of this method, the `errors` argument is updated with an
         Error object to be returned in mutation result.
         """
+        field = snake_to_camel_case(field)
         errors.append(Error(field=field, message=message))
 
     @classmethod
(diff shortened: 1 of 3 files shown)


saleor/graphql/core/utils.py
```diff
@@ -4,3 +4,11 @@ def clean_seo_fields(data):
     if seo_fields:
         data['seo_title'] = seo_fields.get('title')
         data['seo_description'] = seo_fields.get('description')
+
+
+def snake_to_camel_case(name):
+    """Convert snake_case variable name to camelCase."""
+    if isinstance(name, str):
+        splitted = name.split('_')
```
patrys: "Split" is an irregular verb and its past participle is also "split".

saleor/graphql/core/mutations.py
```diff
@@ -173,6 +174,7 @@ def clean_instance(cls, instance, errors):
                 if field in cls._meta.exclude:
                     continue
                 for message in message_dict[field]:
+                    field = snake_to_camel_case(field)
```
maarcingebala: Shouldn't we call `snak_to_camel_case` inside `add_error` instead? We would then ensure that every error you add to the response (as you can do it outside of `clean_instance`) is correctly formatted.

PR #2396, opened 2018-07-13
Update CircleCI API call

artursmet:
I want to merge this change because since we changed main Saleor to CircleCI 2.0 we also have to update demo deployment hooks.


artursmet:
It's set in CircleCI dashboard

PR #2397, opened 2018-07-13
Set variant name in mutation

### LINKED ISSUE #2395: Variant name doesn't change when changing variant attributes in API
I'm trying to update variant attribute through the API. Doing that in the old dashboard updates the name of the variant which is later displayed in UI.

1. Run `productVariantUpdate` mutation and change attributes of a variant.

### What I expected to happen
The variant's name should be updated based on new attributes.

akjanik:
Closes #2395 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/mutations/products.py
+++ b/saleor/graphql/product/mutations/products.py
@@ -335,7 +335,8 @@ class ProductVariantCreate(ModelMutation):
 
     @classmethod
     def save(cls, info, instance, cleaned_input):
-        instance.name = get_name_from_attributes(instance)
+        attributes = instance.product.product_type.variant_attributes.all()
+        instance.name = get_name_from_attributes(instance, attributes)
         instance.save()
 
     @classmethod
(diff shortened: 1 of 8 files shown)


saleor/graphql/product/mutations/products.py
```diff
@@ -294,7 +296,7 @@ class ProductVariantInput(graphene.InputObjectType):
     quantity = graphene.Int(
         description='The total quantity of this variant available for sale.')
     track_inventory = graphene.Boolean(
-        required=True,
+        required=False,
```
maarcingebala: I intentionally set `required=true` on all boolean fields when I was integrating the API with the new dashboard. There was a problem with null values being passed into model boolean fields which resulted in a crash at the database level. The issue is to yet be investigated, but at that moment that was the quick solution that I found.

akjanik:
@patrys Thank you for the feedback, I think it looks better now.

tests/api/test_product.py
```diff
@@ -1080,3 +1082,56 @@ def test_remove_products_to_collection(
...
+        'variantAttributes': variant_attributes_ids})
+    response = admin_api_client.post(
+        reverse('api'), {'query': query, 'variables': variables})
+    content = get_graphql_content(response)
+    assert 'errors' not in content
+    product.refresh_from_db()
+    variant = product.variants.first()
+    attribute = product.product_type.variant_attributes.first()
+    value = attribute.values.first().name
+    assert variant.name == value
```
patrys: How do we know the name was not correct all along? Shouldn't the test explicitly set it to something different first?

PR #2422, opened 2018-07-18
Refactor Sales & Vouchers

Pacu2:
Remarkable things done in this PR:
- [x] `Sale` has `start_date` and `end_date`
- [x] `apply_to` field on `Voucher` was dropped as it presence was confusing and mostly useless
- [x] Added `countries`on `Voucher` to track countries to which free shipping should apply to (all if empty)
- [x] Added `products` on `Voucher` to discount selected products (all if empty)
- [x] Renamed `limit` to `min_amount_spent` on the `Voucher` due to its confusing name
- [x] Added `collections` on `Voucher` to discount selected collections (all if empty)
- [x] Unified some methods for adding voucher on the storefront and the dashboard

TODO:
- [x] Tests
- [x] Migrations
- [x] Docstrings / help texts
- [x] Ability to discount each item individually instead of the whole order? (eg. 15$ off each item instead of 15$ off whole order)
- [ ] ...

and many many more

### Pull Request Checklist


saleor/dashboard/discount/forms.py
```diff
@@ -123,94 +139,79 @@ def country_choices():
...
-
-    limit = MoneyField(
-        min_value=ZERO_MONEY, required=False,
-        currency=settings.DEFAULT_CURRENCY,
-        label=pgettext_lazy(
-            'Lowest value for order to be able to use the voucher',
-            'Only if order is over or equal to'))
-    apply_to = forms.ChoiceField(
+    min_amount_spent = MinAmountSpent
+    countries = forms.MultipleChoiceField(
```
NyanKiyoshi: Do we really want to limit the choice on the shipping methods?

If I have a country "Rest of the World" in my shipping methods and I want to add a Sale for France for example, I'm not able as France is *not* in the shipping method as France itself.

I think we probably just could retake safely all the country choices without looking over the shipping methods.

https://github.com/mirumee/saleor/blob/a0b69706d6dc6942e252ef9b4977d6121f34b458/saleor/account/i18n.py#L259-L262

Pacu2: You can pick 'Rest of the world' inside of the discount as well, I think at some point you might want to offer a free shipping only to specific countries, as shipping abroad might be quite costly

NyanKiyoshi: Yes, we can pick "Rest of the World" but not "France" if as you say, we want to make discount for that country.

Pacu2: I think the root of this dispute is the misleading name of `Rest of the world` country choice, in the codebase, we're calling it `Any Country` which, I think, it's way more suitable for it purpose.

Will change the country choice name, and display countries accordingly

PR #2427, opened 2018-07-18
Validate mime type in productImage mutations

### LINKED ISSUE #2321: Mutation productImageCreate doesn't validate file extension
The mutation `productImageCreate` which is used to upload product images from the dashboard doesn't validate file extensions. Now it is possible to upload e.g. an audio file.

oldPadavan:

Fixes #2321

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/mutations/products.py
+++ b/saleor/graphql/product/mutations/products.py
@@ -448,7 +448,7 @@ class ProductImageCreate(ModelMutation):
     def clean_input(cls, info, instance, input, errors):
         cleaned_input = super().clean_input(info, instance, input, errors)
         uploaded_image = cleaned_input['image']
-        if not uploaded_image.content_type.startswith('image'):
+        if not uploaded_image.content_type.startswith('image/'):
             cls.add_error(errors, 'image', 'Invalid file type')
         return cleaned_input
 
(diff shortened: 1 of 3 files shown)


saleor/graphql/product/mutations/products.py
```diff
@@ -461,6 +469,14 @@ class Meta:
         description = 'Updates a product image.'
         model = models.ProductImage
 
+    @classmethod
```
NyanKiyoshi: You should not write the same code twice.

saleor/graphql/product/mutations/products.py
```diff
@@ -444,6 +444,14 @@ class Meta:
         description = 'Creates a product image.'
         model = models.ProductImage
 
+    @classmethod
+    def clean_input(cls, info, instance, input, errors):
+        cleaned_input = super().clean_input(info, instance, input, errors)
+        uploaded_image = cleaned_input['image']
+        if not uploaded_image.content_type.startswith('image'):
```
NyanKiyoshi: The correct mime type is `image/...` (https://www.iana.org/assignments/media-types/media-types.xhtml#image), you should check it's starting with `image/`.

saleor/graphql/product/mutations/products.py
maarcingebala: True, `ProductImageUpdate` could inherit from `ProductImageCreate`.

oldPadavan:
By the way, I've noticed the test `test_product_image_update_mutation` seems to be actually testing `productImageCreate` mutation instead of `productImageUpdate`
```
def test_product_image_update_mutation(admin_api_client, product_with_image):
    product = product_with_image
    query = """
    mutation updateProductImage($image: Upload!, $alt: String, $product: ID!) {
        productImageCreate(input: {image: $image, alt: $alt, product: $product}) {
            productImage {
                alt
            }
        }
    }
    """
```

PR #2430, opened 2018-07-19
Introducing Pipenv

jxltom:
Pipenv is a quite popular and easy tool for managing python packages. It is also supported by Heroku seamlessly.

```pipenv install``` or ```pipenv install --dev``` is for installing locked version of packages.
```pipenv lock``` is for locking current package versions.

This PR introduces pipenv by converting ```requiremnents.in``` to ```Pipfile``` and lock packages versions to ```Pipfile.lock``` by ```pipenv lock```.

### Pull Request Checklist


maarcingebala:
Thanks for this PR @jxltom! I haven't used Pipenv yet so I don't have any opinion about it. It looks promising but I'd like to test it locally first to see how it works.

jxltom:
Of course, please try. It is like composer for php, npm for nodejs. It is much easier to manage packages using pipenv than using requirements.txt and requirements-dev.txt. :smile:

jxltom:
Drop ```requirements.*``` supports or support both requirements.* and pipenv? The pipenv itself does not conflict with ```requirements.*```

Heroku's buildpack also supports both. It will try to install from Pipfile.lock if it exists, otherwise it will install from ```requirements.*```.

I suggests dropping ```requirements.in``` but keep ```requirements.txt``` and ```requirements-dev.txt```. So user who don't use pipenv can still install by ```pip install -r requirements.txt```. For users (especially developers) using pipenv, they can install packages by ```pipenv install --dev``` and update ```requirements.txt``` easily by ```pipenv lock -r requirements.txt```

patrys:
I don't have a preference but I'd like to have the following two bases covered:

1. there should be one obvious way to install the project locally and one way to change the requirements so that a hypothetical lock file does not drift away from the existing requirements file
2. the Docker image should not have to deal with virtual environments since it's already a container

I'm not sure if we can cover both but if we can't I think from maintainability point of view it's easier to sacrifice the latter rather than the former.

patrys:
Perhaps we could just use `pipenv lock --requirements [--dev]` and rebuild the requirement files?

jxltom:
Sure, that's what I'm suggesting.

1.1. For develepers, they should update ```requirements*.txt``` if the pinned package versions are updated in ```Pipfile```. This can be done by ```pipenv lock -r > requirements.txt``` and ```pipenv lock -r --dev > requirements-dev.txt```

1.2. Then, no matter for users or developers, they all can install packages locally by ```pip install requirements.txt``` and ```pip install requirements-dev.txt```

1.3. For developers or users who prefer pipenv, they can install packages locally by ```pipenv install``` or ```pipenv install --dev```

2. In Docker, packages can be installed by ```pipenv install --system``` which will install packages to system-wide without creating virtual envs. Or, ```pip install requirements.txt``` is also OK since ```requirements.txt``` is also pinned to proper packge versions by ```pipenv```

jxltom:
This PR is updated since there are some changes in ```requirements_dev.txt```. Some comments are also added to make it easier to compare ```Pipfile``` with ```requirements.in``` line-by-line. After PR is approved, these comments can be safely removed.

maarcingebala:
Thanks @jxltom, looks good, but the tests are failing now due to `django-debug-toolbar` moved to dev requirements and not being installed on CI, could you take a look at that?

jxltom:
@maarcingebala Thanks for reminding. Fixed by installing ```requirments_dev.txt``` in tox since CI/CD relys pytest and all other dev packages.

maarcingebala:
🎉

PR #2432, opened 2018-07-19
Celery support in heroku

jxltom:
Support celery worker in Heroku

- ```celeryworker``` is added in Procfile

- CloudAMQP addon in heroku is used as celery broker. Saleor will read ```CLOUDAMQP_URL``` as broker url if ```CELERY_BROKER_URL``` is not provided.

### Pull Request Checklist


maarcingebala:
Couldn't we use Redis as a broker? We already have it in Heroku's configuration, also settings.py supports it.

jxltom:
The heroku-redis addon is enabled in Heroku indeed, but it is only set for django cache. Check the configurations here https://github.com/mirumee/saleor/blob/575718ad095434e15037b90af14f18a44697c647/saleor/settings.py#L48.

It is fine to set redis as celery broker as well. But rabbitmq is more effecient and stable than redis as a celery broker, IMHO. So in this PR, amqp addmon is added.

patrys:
Does this change mean we can no longer use the free credits to deploy a dev environment for free?

jxltom:
@patrys It is still free. ```cloudamqp:lemur``` is a free plan. Refer to https://elements.heroku.com/addons/cloudamqp.

Besides, an extra dyno such as a celery worker with default ```web``` dyno are also free.

patrys:
I was more worried about the extra worker dyno, previously the free tier did not include it. Thanks for your quick response.

jxltom:
The doc https://www.heroku.com/pricing says heroku's free plan includes 1 web and 1 worker with 1000 free dyno hours if account is verified. So enabling one extra worker dyno is free.

Even using 1 web and 1 worker will exceed 1000 free dyno hours, it is fine to just disable worker dyno in heroku's dashboard. By the way, by default the worker dyno in Heroku in disabled after app is deployed. For enable it, user has to enable it in heroku's dashboard manually.

patrys:
@jxltom One thing to consider is that 1000 hours is not enough to run two dynos for one month so the dev environment may not be usable 24/7.

jxltom:
Yes, indeed. If we would like to keep the demo in heroku run 24/7, we can not run both web and worker dynos.

But if it is necessary, we could just disable the worker dyno in Heroku's dashboard.

IMHO, though the worker dyno may be disabled in Heroku but the code should support it. So users could try this feature when they would like to and it makes easier to test the worker when needed.

jxltom:
Add it makes possible to deploy to heroku by [Deploy](https://heroku.com/deploy?template=https://github.com/mirumee/saleor) in one click with both web/worker, which can be used for production use, for paid users in Heroku or users who would like to try the demo with full features.

PR #2434, opened 2018-07-19
2291/drop view permission

Pacu2:
- [x] Dropped several permissions for product management (`edit_properties`, `edit_category`...) unifying them to `edit_product`
- [x] Unified `edit_vouchers` and `edit_sales` to `edit_discount`
- [x] Dropped `view permissions` leaving only `edit permissions` 
- [x] As a side effect, cleaned up and deleted several duplicated tests from `test_permissions.py`
- [x] Fixed minor bugs with incorrect permissions in graphql api
- [x] Changed permission naming convention from`edit_module` to `manage_modules`
- [x] updated permission names to better fit their current usage
new_permission_page

I've also renamed edit permissions and updated their description to fit their new role better


NyanKiyoshi:
@Pacu2 you forgot to commit the migrations files.

templates/dashboard/staff/detail.html
```diff
@@ -70,7 +68,6 @@
                 </div>
```
NyanKiyoshi: Shouldn't it better to call the template `form` instead of `detail`? It's kind of confusing when looking for the form template as the template is mostly a form.

maarcingebala:
@Pacu2 I'd simplify those permission descriptions visible in the staff form. E.g. instead of saying "Capabilities to view, create, modify and delete vouchers" it is enough to say "Manage vouchers". In the case of products, collections, categories I would also say "Manage products", as collection and categories are part of that. Instead of "Capabilities to impersonate the customers" we can say simply "Impersonate customers".

Issue #2292, opened 2018-06-19
Mutation to reset password by email


maarcingebala:
We have mutations to create users through the API: `customerCreate` and `staffCreate`, but they don't allow to set the password. Both of them should be extended with an extra argument: `send_password_email: Boolean`. If set to true, this argument would trigger sending an email with a link to set the password.

We should also add the mutation to trigger this action once the user is created:
`passwordReset(id: ID!)` - clear password and send an email with a link to set it up.

patrys:
If `passwordReset` requires a user `ID` then we create a chicken and egg problem for regular users where if you don't know your current password there's no way for you to obtain your user ID.

elwoodxblues:
True, this would work only in the dashboard where we have the users list. Then we could instead use `email` since it is unique. This mutation should also only send the email, without clearing the password, as I originally wrote.

PR #2452, opened 2018-07-20
Password reset emails mutations

oldPadavan:

implements #2292 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -1,9 +1,23 @@
 import graphene
 from django.contrib.auth.tokens import default_token_generator
+from django.core.exceptions import ObjectDoesNotExist
+from django.utils.encoding import force_bytes
+from django.utils.http import urlsafe_base64_encode
+from graphql_jwt.decorators import permission_required
 
-from ...account import models
+from ...account import emails, models
 from ...core.permissions import MODELS_PERMISSIONS, get_permissions
-from ..core.mutations import ModelMutation
+from ..core.mutations import BaseMutation, ModelMutation
+from ..core.types import Error
+
+
+def send_user_password_reset_email(user):
+    context = {
+        'email': user.email,
+        'uid': urlsafe_base64_encode(force_bytes(user.pk)).decode(),
+        'token': default_token_generator.make_token(user)
+    }
+    emails.send_password_reset_email.delay(context, user.email)
 
 
 class UserInput(graphene.InputObjectType):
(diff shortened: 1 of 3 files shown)


oldPadavan:
I'm not sure if I'm passing all required arguments to `context` variable - I don't have email set up in development environment, so I haven't tested whether email is really sent

saleor/graphql/account/mutations.py
```diff
@@ -12,22 +31,39 @@ class UserInput(graphene.InputObjectType):
...
         description = 'Creates a new customer.'
         exclude = ['password']
         model = models.User
 
+    @classmethod
+    def save(cls, info, user, cleaned_input):
+        user.save()
+        if cleaned_input.get('send_password_email'):
+            send_user_password_reset_email(user)
+            return cls.success_response(user)
```
maarcingebala: This line can be dropped, as `ModelMutation` doesn't expect any return value from `save`. It is the `mutate` method that handles returning GraphQL responses.

saleor/graphql/account/mutations.py
```diff
@@ -1,9 +1,28 @@
...
+
+def send_user_password_reset_email(user, use_https=False):
+    site = Site.objects.get_current()
+    context = {
+        'email': user.email,
+        'uid': urlsafe_base64_encode(force_bytes(user.pk)).decode(),
+        'token': default_token_generator.make_token(user),
+        'site_name': site.name,
+        'domain': site.domain,
+        'protocol': 'https' if use_https else 'http'
```
maarcingebala: In `settings.py` we have `ENABLE_SSL` setting that determines whether to use https, so the argument `use_https` could be dropped.

saleor/graphql/account/mutations.py
```diff
@@ -130,3 +173,23 @@ def clean_input(cls, info, instance, input, errors):
...
+    @classmethod
+    @permission_required('account.manage_users')
+    def mutate(cls, root, info, email):
+        try:
+            user = models.User.objects.get(email=email)
+        except ObjectDoesNotExist:
+            return cls(
+                errors=[
+                    Error(field='email',
+                          message="User with this email doesn't exist")])
```
maarcingebala: To be consistent with formatting in rest of the file, please use single quotes here.

maarcingebala: We use a bit different line-breaking style - `field` should go under Error and `message` under `field`, like this:
```
Error(
    field='email',
    message="User with this email doesn't exist")
```
You may try using `yapf` for code formatting, it should handle it for you.

tests/api/test_account.py
```diff
@@ -414,3 +415,148 @@ def test_set_password(user_api_client, customer_user):
 
     customer_user.refresh_from_db()
     assert customer_user.check_password(password)
+
+
+@patch('saleor.account.emails.send_password_reset_email.delay')
+def test_create_user_password_reset_email(
```
maarcingebala: We already have a test for the `customerCreate` mutation.  I think we can extend it with the additional variable `send_mail: True` and test email sending there.

tests/api/test_account.py
```diff
@@ -414,3 +415,148 @@ def test_set_password(user_api_client, customer_user):
...
+    assert send_password_reset_mock.call_count == 1
+    args, kwargs = send_password_reset_mock.call_args
+    call_context = args[0]
+    call_email = args[1]
+    assert call_email == email
+    assert 'token' in call_context
+
+
+@patch('saleor.account.emails.send_password_reset_email.delay')
+def test_create_staff_password_reset_email(
```
maarcingebala: Same here, I would do the necessary changes in `test_staff_create` to include the email testing there.

saleor/graphql/account/mutations.py
```diff
@@ -1,9 +1,29 @@
...
-from ...account import models
+from ...account import emails, models
 from ...core.permissions import MODELS_PERMISSIONS, get_permissions
-from ..core.mutations import ModelMutation
+from ..core.mutations import BaseMutation, ModelMutation
+from ..core.types import Error
+
+
+def send_user_password_reset_email(user):
+    site = Site.objects.get_current()
```
maarcingebala: I just realized that we can save this database query. We have the middleware `saleor.core.middleware.site` that fetches the current `site` and assigns it to the `request`. In Graphene we can access it in resolvers: `info.context.site` so we would have to pass it to this function as a second argument.

maarcingebala:
@oldPadavan, after merging your PR with address mutation there are some conflicts, could you resolve them?

Issue #2159, opened 2018-05-10
Drop is_public flag from OrderNote model


dominik-zeglen:
As the first step to refactoring order notes, I would like to propose removing `is_public` flag from the model. As I spoke with @patrys, we decided that notes should be visible to staff only, and communication with customers should be moved elsewhere.

PR #2465, opened 2018-07-24
Drop note is public 2159

oldPadavan:

implements #2159

### Pull Request Checklist


oldPadavan:
Although in #2159 it was mentioned the notes should be visible to staff only, I left visible to the user its own notes. If it's not ok, I'll modify the view

saleor/dashboard/order/forms.py
```diff
@@ -248,14 +248,10 @@ def save(self, commit=True):
...
         widgets = {
             'content': forms.Textarea()}
-        labels = {
-            'content': pgettext_lazy('Order note', 'Note'),
-            'is_public': pgettext_lazy(
-                'Allow customers to see note toggle',
-                'Customer can see this note')}
+        labels = {'content': pgettext_lazy('Order note', 'Note')}
 
     def send_confirmation_email(self):
```
Pacu2: I think this should be removed as well, was used for notifying an user about new notes on his order

saleor/order/models.py
```diff
@@ -322,6 +322,7 @@ class Meta:
 class OrderNote(BaseNote):
     order = models.ForeignKey(
         Order, related_name='notes', on_delete=models.CASCADE)
+    is_public = None
```
Pacu2: Why is that needed?

saleor/dashboard/order/forms.py
Pacu2: I think that was used for notifying a customer about a new note on his order, could be removed as well

saleor/order/models.py
oldPadavan: `OrderNote` inherits from Abstract base class `BaseNote`. In order to remote the field `is_public` from `OrderNote` I set it to None. From docs:
> Fields inherited from abstract base classes can be overridden with another field or value, or be removed with None.

saleor/dashboard/order/forms.py
oldPadavan: Thanks for comments, I missed it

saleor/order/views.py
```diff
@@ -31,10 +31,10 @@ def details(request, token):
     orders = orders.select_related(
         'billing_address', 'shipping_address', 'user')
     order = get_object_or_404(orders, token=token)
-    notes = order.notes.filter(is_public=True)
+    user = request.user if request.user.is_authenticated else None
+    notes = order.notes.filter(user=user)
```
Pacu2: I see some trouble here, let's say if staff user added some note, and his account was deleted after, then `OrderNote` he added will be preserved, but `user` will be set to `None`. This may cause some data leakage. 

I wonder if we shouldn't leave `is_public` on the `OrderNote` to determine if the note was added by the customer or not, without a possibility to create a public note from the dashboard 

Or shall we leave `OrderNote` to customers only, and introduce another way for staff to add notes about orders/ fulfillments etc. 
cc @maarcingebala @NyanKiyoshi

maarcingebala: I'd remove this line at all and disable rendering notes in the customer's view. As in the original issue - those notes should be rewritten so that they're only avaialble to staff users.

To keep the customer notes functionality, I would add a simple Textfield in Cart and in Order models. We would use that to store a one-time note from the customer and display it somewhere in the dashboard.

oldPadavan: Thanks for Your feedback! I have a couple of questions. Should these new notes be visible to the customer? Should these fields be added in this PR or it should be a separate PR?

maarcingebala: If users can add notes to their order I think they should also be able to see it later in the storefront's order details page. I would prefer to do those changes in this PR to keep the "order notes" funtionality available on `master`.

maarcingebala:
@oldPadavan A few remarks about this PR:
- I think we could entirely drop the `BaseNote` abstract model. It was a missed concept and it won't be used anywhere else except `OrderNote`, so we could simply move its fields to `OrderNote`.
- It would be great if you could remove those "merge" commits. We prefer to use `git rebase` to keep up with `master` as it produces cleaner commit history.

maarcingebala:
Btw. I just noticed that we've already merged 7 PRs from you. We really appreciate your contribution! I'd like to reach you outside of Github, could you please drop me an email to `marcin@mirumee.com` when you have time and I'll write you back? Thanks 🙂

oldPadavan:
I noticed there's also `CustomerNote` model which inherits from `BaseNote`

maarcingebala:
You're right - we also use it to add multiple notes in the customer details page. Personally, I think it's overcomplicated now because there is also a simple text field that allows defining customer notes there. But let's keep it then, I don't want to change too much in one PR.

saleor/checkout/utils.py
```diff
@@ -828,7 +828,8 @@ def _fill_order_with_cart_data(order, cart, discounts, taxes):
             order, line.variant, line.quantity, discounts, taxes)
 
     if cart.note:
-        order.notes.create(user=order.user, content=cart.note)
+        order.customer_note = cart.note
+        order.save()
```
Pacu2: if saving just one field, we could use `updated_fields` here

saleor/order/migrations/0051_order_customer_note.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.0.3 on 2018-07-31 14:05
```
Pacu2: I think we could squash this migration with the previous one

tests/test_order.py
```diff
@@ -443,3 +433,24 @@ def test_create_user_after_order(order, client):
     user = User.objects.filter(email='hello@mirumee.com').first()
     assert user is not None
     assert user.orders.filter(token=order.token).exists()
+
+
+def test_view_order_detals(order, client):
```
Pacu2: Shouldn't it be `details` instead of `detals`?

saleor/order/forms.py
```diff
@@ -49,11 +49,11 @@ def __init__(self, *args, **kwargs):
         self.fields['email'].widget = forms.HiddenInput()
 
 
-class OrderNoteForm(forms.ModelForm):
+class CustomerNoteForm(forms.ModelForm):
+    customer_note = forms.CharField(
+        max_length=250, required=False, strip=True, label=False,
```
Pacu2: Aren't we missing a label here?

oldPadavan: There was no label in note form during checkout so I didn't add it here as well. I can add a label or simply a header like "Add a note to Your order" or leave it as it is. How would You advice is better?

saleor/checkout/utils.py
NyanKiyoshi: We could pass this field data at creation time instead:
https://github.com/mirumee/saleor/blob/a945504c5fc1e1d7471212681761751864e8773e/saleor/checkout/utils.py#L828-L836

saleor/order/forms.py
NyanKiyoshi: The current one was "Add note to order", it seems good enough.

saleor/order/models.py
```diff
@@ -90,6 +90,7 @@ class Order(models.Model):
         decimal_places=settings.DEFAULT_DECIMAL_PLACES, default=0)
     discount_name = models.CharField(max_length=255, default='', blank=True)
     display_gross_prices = models.BooleanField(default=True)
+    customer_note = models.TextField(blank=True, default='')
```
NyanKiyoshi: Isn't a `TextField` a bit huge?

saleor/order/forms.py
```diff
@@ -49,11 +49,11 @@ def __init__(self, *args, **kwargs):
         self.fields['email'].widget = forms.HiddenInput()
 
 
-class OrderNoteForm(forms.ModelForm):
+class CustomerNoteForm(forms.ModelForm):
```
NyanKiyoshi: I think the class should be named `OrderNoteForm` instead of `CustomerNoteForm`.

saleor/dashboard/order/forms.py
```diff
@@ -248,18 +247,10 @@ def save(self, commit=True):
 class OrderNoteForm(forms.ModelForm):
     class Meta:
         model = OrderNote
-        fields = ['content', 'is_public']
+        fields = ['content']
```
NyanKiyoshi: Do we still want to use a one to many relation for the dashboard note form?

saleor/order/views.py
```diff
@@ -31,12 +31,9 @@ def details(request, token):
...
-    notes = order.notes.filter(is_public=True)
-    ctx = {'order': order, 'notes': notes}
-    if order.is_open():
-        user = request.user if request.user.is_authenticated else None
-        note = OrderNote(order=order, user=user)
-        note_form = OrderNoteForm(request.POST or None, instance=note)
+    ctx = {'order': order}
+    if order.is_open() and not order.customer_note:
+        note_form = CustomerNoteForm(request.POST or None, instance=order)
         ctx.update({'note_form': note_form})
```
NyanKiyoshi: What about we set `note_form` to `None` and then, we move `ctx = {...}` at the end and add everything we need to the context to save some useless instructions?

```python
def details(request, token):
    note_form = None
    orders = Order.objects.confirmed().prefetch_related(
        'lines__variant', 'fulfillments', 'fulfillments__lines',
        'fulfillments__lines__order_line')
    orders = orders.select_related(
        'billing_address', 'shipping_address', 'user')
    order = get_object_or_404(orders, token=token)
    if order.is_open() and not order.customer_note:
        note_form = CustomerNoteForm(request.POST or None, instance=order)
        ctx.update({'note_form': note_form})
        if request.method == 'POST':
            if note_form.is_valid():
                note_form.save()
                return redirect('order:details', token=order.token)
    fulfillments = order.fulfillments.filter(
        status=FulfillmentStatus.FULFILLED)
    ctx = {
        'order': order, 'fulfillments': fulfillments, 'note_form': note_form}
    return TemplateResponse(request, 'order/details.html', ctx)

File: templates/dashboard/order/detail.html [hunk 12]
```diff
@@ -518,6 +518,22 @@ <h5 class="data-table-title">
...
+      <div class="card order-notes{% if not order.customer_note %} print-hide{% endif %}" id="customer-note">
+        <div class="card-content">
+          <span class="card-title black-text">
+            {% trans "Customer note" context "Customer order note card title" %}
+          </span>
+          {% if order.customer_note %}
+            <p>{{ order.customer_note }}</p>
+          {% else %}
+            <p>
+              {% trans "There's no note from customer associated with this order." context "No customer order note card message" %}
```
NyanKiyoshi: Suggested change:
> There **is** no note from **the** customer associated with this order.

Or maybe we could just hide the whole card?

templates/order/details.html
```diff
@@ -63,35 +63,11 @@ <h2 class="order-details__header">
...
-            {% else %}
-              {% blocktrans with date=note.date %}
-                {{ date }} by anonymous customer
-              {% endblocktrans %}
-            {% endif %}
-          </small>
-          <p>{{ note.content }}</p>
-        {% endfor %}
+      {% if order.customer_note %}
+        <h2 class="order-notes__header">{% trans "Customer note" context "Customer order note" %}</h2>
```
NyanKiyoshi: I would say "Your notes" instead since it is in the customer's panel.

saleor/order/models.py
oldPadavan: `Cart` also contains `note` which is later copied to `Order.customer_note` (previously was added to the list of `OrderNote`). That field is `TextField`, so I decided to keep it the same in `Order`.

templates/order/details.html
oldPadavan: Sounds better, just I would rename to "Your note" as it could be at most 1 note

saleor/order/forms.py
oldPadavan: I've added a header to the template

saleor/dashboard/order/forms.py
oldPadavan: Do You mean to leave only one-time comment here as well?

maarcingebala: This one should use a relation - admins are able to add multiple notes to an order.

Issue #2220, opened 2018-05-28
Countries in address form are not sorted alphabetically due to translation


mad-anne:
List of countries in the AddresForm used during checkout is not sorted alphabetically due to translation. As a result, when language is set to different than English, sorting doesn't change.

Below is a screenshot showing that in polish translation Czarnogóra (eng. *Montenegro*)  appears after Mongolia.

![checkout-list](https://user-images.githubusercontent.com/20802901/40622260-ed949ffa-62a0-11e8-9baa-69c507a1bf27.png)

DigitalProFound:
can anyone help me to install this on my domain ? please

PR #2475, opened 2018-07-27
Add translation interface

### LINKED ISSUE #815: Translation for models
As we discussed with @patrys, it will be useful to add translation feature to saleor models (mostly products, probably some CMS in future).

A few assumptions:
 
Create a model that adds translation to a Product instance and is connected to a product with FK.

* Use Prefetch objects
* Adding new language can't modify the DB schema
 * API like `product.translated.name` - how do we handle multiple languages there?
* Use Django Prefetch objects
* ~Only products with complete translation will be available in translated version (if not fields are translated, product will be displayed only in default language)~ Only complete translation can be added to a product

Pacu2:
translation_interface

close #815 

In this PR:
- [x] added translation interface for all models used in the storefront (eg. title, description, seo_title, seo_description ...)
- [x] Changed the way Category's URL is generated due to extra queries fired (it uses slug instead of building fully nested category path)
- [x] prepared storefront for displaying translated properties (just add translations and you’re ready to go!) 
- [x] tests 
- [x] docs
- [x] migrations

Things that Needs to be done after merging this PR:
- [ ] interface for adding translations in the dashboard
- [ ] graphql api for translations 

### Pull Request Checklist


Pacu2:
Functionality is fully working (excluding some debugging code left), but have few potential drawbacks:
- MPTT models do not come well with the translations, some internal logic used for getting children/ancestors is breaking the prefetches. It's especially visible for the navigation, producing whooping 30 database queries per page.
**solution:** As Menu can be changed in a limited number of places, I propose we parse it to JSON, store on the model, and serve the JSON version to avoid all the queries. 

- On the `Order`, there's `product_name` as string used for identifying a purchased product, at this moment it's saved as a translated version(the one that customer saw at the checkout)
**solution:** Add product_name_translated or product_name_native to allow easier identifying purchased products. The translated version could be used in the storefront, while native - in the dashboard

patrys:
In future we should also implement the translation fallback logic from gettext but I think it's a separate task: https://code.djangoproject.com/ticket/29144

Pacu2:
I've also fixed #2220 in this PR

Pacu2:
Helper function that can be used for creating polish translations for all the content:
https://gist.github.com/Pacu2/6f4d08f037f468ba411c66de44435a16

Pacu2:
Migrations will be added after merging #2422

NyanKiyoshi:
@Pacu2 could you add tests to check when `json_content` of the top and the bottom menu is empty, `home` doesn't crash? (it does crash right now)

Pacu2:
The internal logic of MPTT models (eg. Category) used for getting ancestors/descendants is breaking the prefetches.  The number of queries is rising as we're moving down the categories tree.

The menu was reworked into JSON to avoid it, categories hopefully will be dropped soon.

saleor/core/utils/translations.py
```diff
@@ -0,0 +1,28 @@
+from django.utils.translation import get_language
+
+
+class TranslationWrapper(object):
+    def __init__(self, instance, locale):
+        self.instance = instance
+        self.translation = next((
+            t for t in instance.translations.all()
```
maarcingebala: In this line, we assume that the relation `instance.translations` exists, but it may be undefined if the model with translated fields and reverse relation `translations` was not defined. I think we should add a check here and throw an error in that case (we can implement it later).

maarcingebala:
We will also have to add some docs explaining how it works, but I'll open a separate issue for that.

NyanKiyoshi:
@Pacu2 I still have Saleor crashing when I migrate from an older menu, so when the menu is empty and no populatedb was done (`json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`).

Pacu2:
@NyanKiyoshi There was some miscommunication on what was the empty menu, I thought u meant menu unassigned from settings, not the one that was left after creating the store, without running the populated command.

Should be fixed on https://github.com/mirumee/saleor/pull/2522/files

PR #2479, opened 2018-07-29
2148/user can edit own email address

### LINKED ISSUE #2148: Give the user a way to edit their email address from their profile
Hi!

I just remembered that, as users, we cannot edit our email address in our account in Saleor. We should add this feature in Saleor.

Note: I assign this task to myself. I will put it on my todolist.

It's rather simple: 
1. Add a edit button next to the email address in `/profile/`, which would link to a form in the `account` namespace;
2. Then, once changed (need to check if the data is changed or not, in the form), we would put a pending verification (TTL of 15 minutes) on the new user email address which would activate the email address.

tomcio1205:
close #2148


saleor/account/forms.py
```diff
@@ -127,3 +127,27 @@ def send_mail(
...
+
+
+class EmailChangeForm(forms.Form):
+    """ Allow changing email """
+
+    email = forms.EmailField(
+        label=pgettext('New email', 'Email'), max_length=75)
+
+    error_messages = {
+        'email': pgettext_lazy('New email address cannot be the same'
```
Pacu2: `pgettext` takes two arguments, first is the context of the translated string, second is the string to be translated. You're missing the first one here

saleor/account/forms.py
```diff
@@ -127,3 +127,27 @@ def send_mail(
         # template, we remove it from the context.
         del context['user']
         emails.send_password_reset_email.delay(context, to_email)
+
+
+class EmailChangeForm(forms.Form):
+    """ Allow changing email """
+
+    email = forms.EmailField(
+        label=pgettext('New email', 'Email'), max_length=75)
```
Pacu2: I think context could be a little more expanded here, like `Label of the user's email field`

Also, it's quite dangerous to limit the `max_length` at the form level, we need to make sure that this length is always the same as on the Model's field.

saleor/account/forms.py
```diff
@@ -127,3 +127,27 @@ def send_mail(
         # template, we remove it from the context.
         del context['user']
         emails.send_password_reset_email.delay(context, to_email)
+
+
+class EmailChangeForm(forms.Form):
```
Pacu2: Can we use `ModelForm` here? It will simplify things out, passing user to the constructor won't be necessary, as we could take it from the instance

tomcio1205:
I have some questions. Firstly the class that I created "EmailChangeForm" raising ValidationError if new email is the same as current email, but the error messages are not showing in view. So I added condition in email_edit method (views.py) where I check if form has errors and then call messages.error. I know that it not the best option but I don't know why ValidationError not working. Secondly I have question how to include TTL in email user.

templates/account/details.html
```diff
@@ -152,6 +134,28 @@ <h3 class="text-sm-center">
         </div>
       </div>
       <div class="tab-pane fade" id="settings" role="tabpanel">
+        {% if user.email %}
```
Pacu2: I think it can never be false, as we always have the user's email. 

There's one case, where `user.password` might be false and it's when one signed up via Google/Facebook social login

saleor/account/views.py
```diff
@@ -156,3 +158,20 @@ def account_delete_confirm(request, token):
...
+def email_edit(request):
+    form = EmailChangeForm(request, data=request.POST or None)
+
+    if request.method == 'POST':
+        if form.is_valid():
+            form.clean_email()
+            messages.success(request, pgettext(
+                'Storefront message', 'Email successfully changed.'))
+        elif form.errors:
+            messages.error(request, pgettext(
```
Pacu2: I think we should stick to the form's errors, trying to attach the error message to the most precise item possible. 

form's field -> whole form (ValidationError) -> messages

saleor/account/views.py
```diff
@@ -156,3 +158,20 @@ def account_delete_confirm(request, token):
...
         request, 'account/account_delete_prompt.html')
+
+
+@login_required
+def email_edit(request):
+    form = EmailChangeForm(request, data=request.POST or None)
+
+    if request.method == 'POST':
+        if form.is_valid():
+            form.clean_email()
```
patrys: Why is `clean_email` called here?

saleor/account/views.py
```diff
@@ -156,3 +158,20 @@ def account_delete_confirm(request, token):
...
+            messages.success(request, pgettext(
+                'Storefront message', 'Email successfully changed.'))
```
patrys: Is this correct? Is there code that actually changes the email address?

tomcio1205: @patrys It will be as you saying. Firstly we will be sending email to the old address. I'm making changes in code according to @Pacu suggestions, so it will be completely different. I'm was made this PR because I didn't know how to resolve this issue :)

tomcio1205:
I have some questions regarding my last commits. First of all, I want to ask if this make any sense. When user click form button I check which button was clicked and based on this i get right form using get_forms_by_method_or_button method. If button for email changing was clicked I check if address is different than current and then send email (on old address)with confirmation link. And second I don't know how to get new address email when user will click on link. Should I pass new email with url param? and then get this in method that handle confirmation or maybe there is better solution?

NyanKiyoshi:
1. You need to edit the `action` attribute of the form and thus handle the `POST` in a separated view;
2. You can temporary store the new email to the `Account` and retrieve it when the user clicks on the confirmation url.

tomcio1205:
@NyanKiyoshi  I did it before but I don't know what should return in this view. 
`
@login_required
def details(request):

    email_form, password_form = get_forms_by_method_or_button(request)
    orders = request.user.orders.confirmed().prefetch_related('lines')
    orders_paginated = get_paginator_items(
        orders, settings.PAGINATE_BY, request.GET.get('page'))
    ctx = {'addresses': request.user.addresses.all(),
           'orders': orders_paginated,
           'change_password_form': password_form,
           'change_email_form': email_form}

    return TemplateResponse(request, 'account/details.html', ctx)
`
This method is for handle account details view, and both form (password and email change) are returned in context of this function, so when somebody click the submit button for email changing and I handle this action in view e.g. /email/change should I redirect to details view with passed email form, or somehow return email form only? Before when I returned email form I had errors like

> Parameter "form" should contain a valid Django Form

For the second point - so I need to add new field in the models.py where I will be store temporary email ?

saleor/account/forms.py
```diff
@@ -120,10 +121,56 @@ def get_users(self, email):
...
+
+
+class EmailChangeForm(forms.ModelForm):
+    """ Allow changing email """
+
+    user_email = forms.EmailField(
+        label=pgettext("Label of the user's email field", "User Email"))
+
+    error_messages = {
+        'user_email': pgettext_lazy("Change email error",
```
Pacu2: You should start from the new line

saleor/account/forms.py
```diff
@@ -120,10 +121,56 @@ def get_users(self, email):
...
+                                    ' as your current email address'),
+    }
+
+    class Meta:
+        model = User
+        fields = ['user_email']
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.user = self.instance
```
Pacu2: Can we use `self.instance` through this function?

saleor/account/forms.py
```diff
@@ -120,10 +121,56 @@ def get_users(self, email):
...
+    def clean(self):
+        data = self.cleaned_data
+        if self.user.email == data['user_email']:
+            self.add_error('user_email', self.error_messages['user_email'])
+        return data
+
+    def send_mail(self, request=None, use_https=False):
+        to_email = self.user.email
+        new_email = self.data['user_email']
+        user = self.user
```
Pacu2: I think it doesn't make any improvement in this function

saleor/account/views.py
```diff
@@ -156,3 +171,26 @@ def account_delete_confirm(request, token):
...
     return TemplateResponse(
         request, 'account/account_delete_prompt.html')
+
+
+@login_required
+def email_edit(request):
+    form = EmailChangeForm(data=request.POST or None, instance=request.user)
+
+    if request.method == 'POST':
+        if form.is_valid():
```
Pacu2: can be one line

saleor/account/views.py
```diff
@@ -156,3 +171,26 @@ def account_delete_confirm(request, token):
...
+            form.send_mail()
+            messages.success(request, pgettext(
+                'Storefront message', 'Email successfully changed.'))
+    return form
+
+
+@login_required
+def email_change_confirm(request, token=None):
+    user = request.user
```
Pacu2: I think that's unnecessary

templates/templated_email/account/email_change_confirm.email
```diff
@@ -0,0 +1,27 @@
...
+
+{% block plain %}
+{% blocktrans context "Email change confirmation text" %}
+Hi!
+
+You're receiving this e-mail because you or someone else has requested changing email for your user account at {{ domain }}.
+It can be safely ignored if you did not request a email change. Click the link below to activate your new email address ({{ new_email }}).
+{% endblocktrans %}
+
+{{ protocol}}://{{ domain }}{% url 'account:email-change-confirm' token=token %}
```
Pacu2: You should put this link on the backend, and pass it to the template

saleor/account/views.py
```diff
@@ -75,17 +75,32 @@ def password_reset_confirm(request, uidb64=None, token=None):
...
+def get_forms_by_method_or_button(request):
+    if request.method == 'POST':
+        if 'email_change' in request.POST:
+            password_form = ChangePasswordForm(data=None, user=request.user)
+            email_form = email_edit(request)
+        elif 'password_change' in request.POST:
+            password_form = get_or_process_password_form(request)
+            email_form = EmailChangeForm(data=None)
+        return email_form, password_form
+    else:
```
Pacu2: This `else` statement is not necessary, as you'll return from the above's `if` anyway

saleor/account/forms.py
```diff
@@ -120,10 +121,56 @@ def get_users(self, email):
...
+        site_name = current_site.name
+        domain = current_site.domain
+        context = {
+            'new_email': new_email,
+            'email': to_email,
+            'domain': domain,
+            'site_name': site_name,
+            'token': str(user.token),
+            'protocol': 'https' if use_https else 'http',
+        }
```
Pacu2: We have `no hanging brackets` policy, should close on the previous line  :)

templates/account/email_change_confirm.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+
+{% load i18n %}{% load staticfiles %}
```
Pacu2: Should be on separate lines

saleor/account/forms.py
```diff
@@ -120,10 +124,81 @@ def get_users(self, email):
...
+        data = self.cleaned_data
+        if self.instance.email == data['user_email']:
+            self.add_error('user_email', self.error_messages['same_email'])
+        elif last_request and (last_request + timedelta(minutes=10) >
+                               timezone.now()):
+            self.add_error(
+                'user_email', self.error_messages['requests_exceeded'])
+        else:
+            try:
+                user = User.objects.get(email=data['user_email'])
```
Pacu2: You might want to take a look at [exists](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#django.db.models.query.QuerySet.exists) queryset method

saleor/account/emails.py
```diff
@@ -32,3 +32,17 @@ def send_account_delete_confirmation_email(token, recipient_email):
         from_email=settings.DEFAULT_FROM_EMAIL,
         recipient_list=[recipient_email],
         context=ctx)
+
+
+@shared_task
+def send_user_email_change_confirmation(context, recipient):
+    reset_url = build_absolute_uri(
+        reverse(
+            'account:email-change-confirm', kwargs={'token':
```
Pacu2: I think this type of line breaks is against our formatting rules, there's `yapf` config in the repo, should help you prevent such things

saleor/account/forms.py
```diff
@@ -120,10 +124,81 @@ def get_users(self, email):
...
+            except User.DoesNotExist:
+                user = None
+            finally:
+                if user:
+                    self.add_error(
+                        'user_email', self.error_messages['user_exists'])
+        return data
+
+    def send_mail(self, request=None):
```
Pacu2: I think we might pass just the current site instead of the whole request object

saleor/account/models.py
```diff
@@ -120,7 +120,7 @@ class User(PermissionsMixin, AbstractBaseUser):
     default_billing_address = models.ForeignKey(
         Address, related_name='+', null=True, blank=True,
         on_delete=models.SET_NULL)
-
+    email_change_requested_on = models.DateTimeField(null=True)
```
Pacu2: Do we want to keep a track on when was the last email change?

If that's just for the error purpose, we could read the email back from cache and determine if the time passed or not

saleor/account/views.py
```diff
@@ -156,3 +172,51 @@ def account_delete_confirm(request, token):
...
         request, 'account/account_delete_prompt.html')
+
+
+@login_required
+def email_edit(request):
+    form = EmailChangeForm(data=request.POST or None, instance=request.user)
+
+    if request.method == 'POST' and form.is_valid():
+        cache.set(
+            str(request.user.pk) + '_email_field', form.data['user_email'],
```
Pacu2: I think we should create a constant for email field cache key, that would improve the readability

templates/templated_email/source/email_change_confirm.mjml
```diff
@@ -0,0 +1,30 @@
...
+          {% trans "Hi!" context "Email change confirmation text" %}
+        </mj-text>
+        <mj-text>
+          {% blocktrans trimmed context "Email change confirmation text" %}
+            You're receiving this e-mail because you or someone else has requested changing email for your user account at {{ domain }}.<br/>
+            It can be safely ignored if you did not request a email change. Click the link below to activate your new email address ({{ new_email }}).
+          {% endblocktrans %}
+        </mj-text>
+        <mj-text>
+          <a href="{{ protocol}}://{{ domain }}{% url 'account:email-change-confirm' token=token %}">
```
Pacu2: Can we glue this url on the backend, and pass it to the template?

saleor/account/forms.py
```diff
@@ -120,10 +119,73 @@ def get_users(self, email):
...
+            html_email_template_name=None):
         # Passing the user object to the Celery task throws an
         # error "'User' is not JSON serializable". Since it's not used in our
         # template, we remove it from the context.
         del context['user']
         emails.send_password_reset_email.delay(context, to_email)
+
+
+class EmailChangeForm(forms.ModelForm):
+    """ Allow changing email """
```
Pacu2: I think this docstring is not especially helpful.

It should be added without spaces around `"""` and end with a full stop.

saleor/account/forms.py
```diff
@@ -120,10 +119,73 @@ def get_users(self, email):
...
+
+    user_email = forms.EmailField(
+        label=pgettext("Label of the user's email field", "User Email"))
+
+    error_messages = {
+        'requests_exceeded':
+        pgettext_lazy(
```
Pacu2: This should start on the previous line

saleor/account/forms.py
```diff
@@ -120,10 +119,73 @@ def get_users(self, email):
...
+            'Max requests exceeded',
+            'Sorry, but it looks like You actually trying to reset'
+            ' Your email address again. Please check Your email inbox first.'),
+        'same_email':
+        pgettext_lazy(
+            "Email can't be the same", 'New email address cannot be the '
+            'same as your current email address'),
+        'user_exists':
+        pgettext_lazy(
+            "User already exists",
```
Pacu2: Should be single quotes

saleor/account/forms.py
tomcio1205: @Pacu2 so it should looks like this?
`        'same_email': pgettext_lazy(
            "Email can't be the same", 'New email address cannot be the '
            'same as your current email address'),`

I don't know wy my code formatter split this to new lines....

Pacu2: Right, but with an extra indent in front of `Email can't`...

maarcingebala:
I'm closing this PR as it's inactive and I think this whole feature requires a bit more investigation and planning. What we should take into account:
- since `email` is one of the main identifiers of a customer in the shop, we should probably require customers to confirm the operation by providing their password as well,
- the new email address needs to be confirmed before we can actually change it,
- as PWA storefront is going to become the default storefront I would consider adding this feature only at the API level, without support in the old storefront.

PR #2482, opened 2018-07-30
Fix order of translated country choices

### LINKED ISSUE #2220: Countries in address form are not sorted alphabetically due to translation

Pacu2:
Close #2220 

### Pull Request Checklist

### DIFF
--- a/saleor/account/i18n.py
+++ b/saleor/account/i18n.py
@@ -4,7 +4,6 @@ import i18naddress
 from django import forms
 from django.forms.forms import BoundField
 from django.utils.translation import pgettext_lazy
-from django_countries import countries
 from django_countries.data import COUNTRIES
 from phonenumber_field.formfields import PhoneNumberField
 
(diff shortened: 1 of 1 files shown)


saleor/account/i18n.py
```diff
@@ -122,6 +123,7 @@ class Meta:
     def __init__(self, *args, **kwargs):
         autocomplete_type = kwargs.pop('autocomplete_type', None)
         super().__init__(*args, **kwargs)
+        self.fields['country'].choices = countries
```
patrys: Why does this work?

Pacu2: I think the countries order was taken as defined in the model, not being sorted accordingly to the selected language

PR #2483, opened 2018-07-30
Fix language picker

### LINKED ISSUE #2188: Issues with language selector
Hi, here are a few issues I encountered so far with this feature:

## 1 - [Fixed: #2306] Splitting/ resizing windows
As I have reported multiple times, resizing windows is breaking the language selector menu. Which is really annoying for people splitting the windows.

### How to reproduce
1. Resize the window to a smaller size, that would trigger the mobile view;
2. Resize back to the desktop view;
3. Try to use the language;
4. You should not be able anymore.

## 2 - Mouse hovering
Hovering the mouse over the language in the select, is using the default inactive cursor instead of a pointer. Which makes the user think there are no actions.

## 3 - Hard to select the wanted language
A big issue of the current design is that we are totally unable to click on the language selector and then type the first characters to select the language we want. Which may be an issue for some more advanced users.

## 4 - Unusable on mobile
The language selection menu is *totally* unusable on mobile. At least on Firefox Focus on Android (which is based on Android's Chromium, very likely to be the same for Chrome then).

We have a few centimeters to select the wanted language, and thus to scroll, as well. We can scroll down, but not up.

[scr](https://i.imgur.com/dl1Qqr7.png)

## 5 - [Patch @ #2350] Redirection after selecting a language
A big issue is that when we want to switch the language of a page, we are redirected to the home page, instead of reloading/ redirecting to the newly selected language.

And hitting back, is, of course, making us go back to the previous language as frontstore URLs are specifying the used language.

### How to reproduce
1. Go anywhere you want on front store, let's say **the cart**;
2. Let's say I'm currently using Saleor in English, I switch (in cart) to German;
3. I'm redirected to the home page and (of course) cannot go back to get the language I wanted.

It's a real issue in some pages, like products. Users have to find back their product or their page to view the page in their language, as they switched the language for this reason.

### LINKED ISSUE #2508: Cursor type on language selector
When changing language, two types of cursor (`default` and `auto`) appers when hovering on language options:

saleor-lang-selector

Probably should be `cursor: pointer;` on each language option.

Pacu2:
close #2188 
close #2508 
Behavior on the desktop is the same as it was previously (except for fixing some minor bugs)
language_picker


NyanKiyoshi:
There are a few issues/ details I saw that could be a bit annoying.

1. Labels having a non-clickable area generated by the storefront label margin ([here](https://github.com/mirumee/saleor/blob/9688a7f1af8edca1c4514270f0950e0ebeead43d/saleor/static/scss/components/_forms.scss#L146)). Thus we are getting a pointer cursor when hovering those areas and no effects on click.
2. Then, we have a label displayed as `inline-block` (instead of `block`) generating once again a non-clickable area with a pointer cursor.

templates/_language_picker_modal.html
```diff
@@ -0,0 +1,30 @@
...
+      </div>
+      <form id="language-picker" action="{% url 'set_language' %}" method="POST">
+      <div class="modal-body">
+          {% csrf_token %}
+          <input name="next" type="hidden" value="{{ request.get_full_path }}">
+          {% for language in languages %}
+            <div class="radio">
+              <label>
+                <input type="radio" name="language" value="{{ language.code}}" {% if language.code == LANGUAGE_CODE %}checked{% endif %}>
+                {{ language.name_local }}
```
NyanKiyoshi: Some languages are starting with a capital, some don't (e.g.: French is "français" because in French language name in sentences don't start with a capital; Then we have "English" which does start with a capital in sentences).

We should add a CSS rule or add the filter `capfirst`.

saleor/static/scss/components/_language-picker.scss
```diff
@@ -0,0 +1,32 @@
...
+    display: inline-block;
+    text-align: left;
+    vertical-align: middle;
+  }
+  @media (min-height: 500px) {
+    .modal-body {
+      height: 25em;
+    }
+  }
+  @media (min-height: 800px) {
```
NyanKiyoshi: Why not using breaking point mixins? (`media-breakpoint-down` and `media-breakpoint-up`).

Defaults are `sm: 576px, md: 768px`.

templates/_language_picker_modal.html
patrys: I'm not sure this is a problem. In fact it could be considered a grammar error to spell "polski" with a capital letter unless at the start of a sentence.

NyanKiyoshi: That's interesting. Alright then!

PR #2487, opened 2018-07-31
Site settings endpoints

maarcingebala:
This PR implements all queries mentioned in #2272. While working on this, I've changed the structure of `graphql/core/types` - I've split the types into separate files mostly to have `shop` related queries and resolvers in one file as it may grow bigger over time.

### Pull Request Checklist


saleor/graphql/core/types/money.py
```diff
@@ -0,0 +1,81 @@
...
+        reduced_rates = self.data.get('reduced_rates', {}) or {}
+        return [
+            ReducedRate(rate=rate, type=type)
+            for type, rate in reduced_rates.items()]
+
+
+class ReducedRate(graphene.ObjectType):
+    rate = graphene.Float(
+        description='Reduced VAT rate in percent.', required=True)
+    type = graphene.String(description='A type of goods.', required=True)
```
akjanik: I recommend to avoid overwriting language specific names like `type`. I think something like `rate_type` would be understandable.

saleor/graphql/core/types/shop.py
```diff
@@ -0,0 +1,119 @@
...
+        return format_permissions_for_display(permissions)
+
+    def resolve_phone_prefixes(self, info):
+        return list(COUNTRY_CODE_TO_REGION_CODE.keys())
+
+    @permission_required('site.manage_settings')
+    def resolve_tax_rates(self, info):
+        return vatlayer_models.VAT.objects.order_by('country_code')
+
+    @permission_required('site.manage_settings')
```
akjanik: I wonder if we need to set up permission check here. How is information about tax rates (I guess they are available on some economic law websites) more restricted to users than (crypto)currencies, served countries or phone prefixes used by shop?

maarcingebala: We fetch tax rates from Vatlayer API which except for simple basic plan has also paid ones. I think we shouldn't expose this data to the world as someone might use our API to get data the others have to pay for.

patrys: I second @maarcingebala's sentiment here, while tax rates are public domain information we wouldn't want someone to start using our demo site as a data source for a third party app.

PR #2534, opened 2018-08-09
Fix static urls for public path with a custom path and templates statics

### LINKED ISSUE #2512: Debug toolbar, static files fault
please see:
0fffd44822 1

### LINKED ISSUE #2515: Webpack doesn't look into the right folder when STATIC_URL is supplied
It seems like there is an issue with the configuration of webpack. Which is described as below.

Webpack is building under the folder named `assets` (of which the parent is `static`).

The webpack configuration seems to be expecting the user to supply the base static folder to look into. Which, by default is `/static/`. If, the environment variable `STATIC_URL` is not set, the webpack configuration sets the assets path to `/static/assets/`, which appends to the default `/static/`.

Now, if we are supplying the environment variable `STATIC_URL` to webpack, it make sense that the webpack would expect the assets folder to be appended to the base static folder. But webpack doesn't manually appends `assets` to the base path.

As a result, webpack is looking for the assets at the root static folder instead of its generated folder (`assets`).

As of now, the way to prevent conflicts between the `settings.py`, django side, would be to supply a `STATIC_URL` for request serving; and a different `STATIC_URL` including the `assets` folder in the URL on compilation time of webpack (for `webpack-bundle.json` thus). Which can make sense, but would be really confusing for the users.

1. Set the environment variable `STATIC_URL` to `/static/` and keep it for every command below;
2. Compile the assets **in production mode**;
3. `webpack-bundle.json` should have `publicPath` set to `/static/` instead of `/static/assets/`;
4. Launch Saleor;
5. Make a request on the home page;
6. Saleor will find anything non-related to the webpack bundles (e.g.: placeholder images) but not the assets.

### Proposed solution
As the target folder (`assets`) is hard coded in the config, thus non changeable by the user, I would propose to join `process.env.STATIC_URL` with the folder `assets`.

```patch
Index: webpack.config.js
--- webpack.config.js	(revision 89489fa751617fdb1e60f508ef6eb05931398c5b)
+++ webpack.config.js	(date 1533322641000)
@@ -11,11 +11,13 @@
 var output;
 
 if (process.env.NODE_ENV === 'production') {
+  const baseStaticPath = process.env.STATIC_URL || '/static/';
+  const publicPath = path.join(baseStaticPath, 'assets/');
   output = {
     path: resolve('saleor/static/assets/'),
     filename: '[name].[chunkhash].js',
     chunkFilename: '[name].[chunkhash].js',
-    publicPath: process.env.STATIC_URL || '/static/assets/'
+    publicPath: publicPath
   };
   fileLoaderPath = 'file-loader?name=[name].[hash].[ext]';
   extractCssPlugin = new MiniCssExtractPlugin({

```

NyanKiyoshi:
This fixes #2515 and mentioned issues by @2XL.
Closes #2512 as well.

### Pull Request Checklist

### DIFF
--- a/templates/dashboard/includes/_filters.html
+++ b/templates/dashboard/includes/_filters.html
@@ -9,7 +9,7 @@
       <div class="col s12">
         <span class="card-title">{% trans 'Filters' context 'Dashboard card title' %}</span>
         {{ filter.get_summary_message }}
-        <span class="collapse-activate hide-on-large-only"><svg data-src="{% static '/images/filter-icon.svg' %}" /></span>
+        <span class="collapse-activate hide-on-large-only"><svg data-src="{% static 'images/filter-icon.svg' %}" /></span>
       </div>
       <div class="col s12 hide-on-large-only chips-container">
         {% for chip in chips %}
--- a/saleor/core/templatetags/placeholder.py
+++ b/saleor/core/templatetags/placeholder.py
@@ -6,4 +6,4 @@ register = Library()
 
 @register.simple_tag
 def placeholder(size):
-    return static('/images/placeholder{size}x{size}.png'.format(size=size))
+    return static('images/placeholder{size}x{size}.png'.format(size=size))
--- a/saleor/dashboard/templatetags/utils.py
+++ b/saleor/dashboard/templatetags/utils.py
@@ -132,13 +132,13 @@ def sorting_header(context, field, label, is_wide=False):
             # new_sort_by is used to construct a link with already toggled
             # sort_by value
             new_sort_by = '-%s' % field
-            sorting_icon = static('/images/arrow-up-icon.svg')
+            sorting_icon = static('images/arrow-up-icon.svg')
         else:
             # enable descending sort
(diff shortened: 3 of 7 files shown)


templates/dashboard/includes/_filters.html
```diff
@@ -9,7 +9,7 @@
       <div class="col s12">
         <span class="card-title">{% trans 'Filters' context 'Dashboard card title' %}</span>
         {{ filter.get_summary_message }}
-        <span class="collapse-activate hide-on-large-only"><svg data-src="{% static '/images/filter-icon.svg' %}" /></span>
+        <span class="collapse-activate hide-on-large-only"><svg data-src="{% static 'images/filter-icon.svg' %}" /></span>
```
Pacu2: I think this happens in few other places as well, if you look for ` "/` or ` '/` in .html files, do you mind fixing it as well?

NyanKiyoshi: You are right, I forgot to check both. :+1:

PR #2549, opened 2018-08-10
Refactor autocomplete widgets

### LINKED ISSUE #2530: Refactor autocomplete widgets
At this moment autocomplete widget logic are messy and they require storing value as
```typescript
interface AutocompleteFieldValue {
  label: string;
  value: string;
}
```
which only brings more and more form data processing. They also support both static list and list fetching, which causes them to lose benefits of strict typechecking. The idea to refactor them is to:
- [ ] make widgets stateful and hold current label in component state
- [ ] separate components that are able to fetch choice list from those which cannot
- [ ] use `fuzzaldrin` library to sort static choice lists ( https://github.com/atom/fuzzaldrin )
- [ ] decide if we should drop multiple choice autocomplete widget with fetching ability
- [ ] refactor logic used in views

dominik-zeglen:
Resolves #2530 

### Pull Request Checklist


maarcingebala:
We agreed with @dominik-zeglen that this can be closed due to a lot of conflicts, we'll open a new PR for that.

PR #2569, opened 2018-08-15
2121/Add placeholders, remove label

### LINKED ISSUE #2121: Rename address fields
screenshot

I think we should rename 'Address' and 'Address' to 'Address 1' and 'Address 2' or, even better, we could change the logic of addresses.

For example, Amazon, we have a field set "Street", then inside it:
- Street and number, P.O. box,
- Apartment, suite, building, floor, etc.

screenshot

We don't need to use placeholders, event just as labels "Address 1 (Street and number, ...)", "Address 2 (...)".

I get negative feedbacks on this part as being confusing, and get the questions "What is it?" and "What is address 1 and 2?.

tomcio1205:
close #2121

### DIFF
--- a/saleor/account/i18n.py
+++ b/saleor/account/i18n.py
@@ -90,34 +90,32 @@ class AddressForm(forms.ModelForm):
         model = Address
         exclude = []
         labels = {
-            'first_name':
-            pgettext_lazy('Personal name', 'Given name'),
-            'last_name':
-            pgettext_lazy('Personal name', 'Family name'),
-            'company_name':
-            pgettext_lazy(
+            'first_name': pgettext_lazy(
+                'Personal name', 'Given name'),
+            'last_name': pgettext_lazy(
+                'Personal name', 'Family name'),
+            'company_name': pgettext_lazy(
                 'Company or organization', 'Company or organization'),
-            'street_address_1':
-            pgettext_lazy('Address', 'Address'),
-            'street_address_2':
-            pgettext_lazy('Address', ''),
-            'city':
-            pgettext_lazy('City', 'City'),
-            'city_area':
-            pgettext_lazy('City area', 'District'),
-            'postal_code':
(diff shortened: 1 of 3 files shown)


saleor/account/i18n.py
```diff
@@ -76,45 +73,69 @@ def clean(self):
...
-        ('street_address_1', 'address-line1'),
-        ('street_address_2', 'address-line2'),
-        ('city', 'address-level2'),
-        ('postal_code', 'postal-code'),
-        ('country_area', 'address-level1'),
-        ('country', 'country'),
-        ('city_area', 'address-level3'),
-        ('phone', 'tel'),
-        ('email', 'email')]
+        ('first_name',
```
Pacu2: Something strange happened to this formatting, previous one was good tho

saleor/account/i18n.py
```diff
@@ -76,45 +73,69 @@ def clean(self):
...
     class Meta:
         model = Address
         exclude = []
         labels = {
-            'first_name': pgettext_lazy(
-                'Personal name', 'Given name'),
-            'last_name': pgettext_lazy(
-                'Personal name', 'Family name'),
-            'company_name': pgettext_lazy(
+            'first_name':
```
Pacu2: Same here, previous formatting was OK.

saleor/account/i18n.py
```diff
@@ -250,8 +275,9 @@ def construct_address_form(country_code, i18n_rules):
 
 for country in COUNTRIES.keys():
     try:
-        country_rules = i18naddress.get_validation_rules(
-            {'country_code': country})
+        country_rules = i18naddress.get_validation_rules({
+            'country_code':
```
Pacu2: Previous one was good tho

Pacu2:
@tomcio1205 something happened to the code formatting,  apart from that LGTM

Issue #2507, opened 2018-08-03
Need update psycopg2-binary==2.7.4 to 2.7.5


mmmcorpsvit:
psycopg2-binary==2.7.4
have setup issues on Windows (in my case) please use 2.7.5

Pacu2:
What's the problem with the 2.7.4 version?

mmmcorpsvit:
2.7.4 not build corrent on python 3.7 on Windows 10

patrys:
There are probably no wheel files for Python 3.7.

Issue #1809, opened 2018-02-21
'cu' is not a valid country code


elwoodxblues:
`faker` generates addresses with the country code `CU` which is not supported by [google-i18n-address](https://github.com/mirumee/google-i18n-address). This causes an error when accessing the address instances. It seems that Google doesn't have this code in [their repository](https://chromium-i18n.appspot.com/ssl-address). A quick workaround would be to blacklist this country code in our random data generator.

elwoodxblues:
A much better idea would be to have a default address form which will be the fallback in cases when localized address form can't be provided.

PR #2601, opened 2018-08-16
Update deps

patrys:
This updates most of the outdated dependencies.

I was unable to bump our Django version to 2.1 because of https://github.com/graphql-python/graphene-django/pull/492

### Pull Request Checklist

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -11,8 +11,8 @@ django-cache-url>=1.0.0
 django-countries>=5.2
 django-filter==1.1
 django-fsm>=2.6.0
-django-graphql-jwt==0.1.9
-django-impersonate==1.3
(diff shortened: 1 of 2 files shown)


NyanKiyoshi:
Note: this closes #2507.

NyanKiyoshi:
This also fixes #1809 as google-i18n-address 2.3.1 adds Cuba (CU) and other missing countries. (c.f. https://github.com/mirumee/google-i18n-address/pull/36)

Now, do we still want to implement the change suggested by @maarcingebala?
> A much better idea would be to have a default address form which will be the fallback in cases when localized address form can't be provided.

I feel like it may not be needed as it is very unlikely that the user will have an unknown country code saved into the database. But I may be wrong.

PR #2626, opened 2018-08-18
Add support to google cloud storage

sergioisidoro:
Adding support for google cloud storage through django-storages.

https://github.com/mirumee/saleor/issues/2533
Deploying saleor to google cloud with Kubernetes and having aws buckets for storage is an inconvenience. This is a draft to add support for google cloud storage.

In order to prevent some confusions in case both settings are present, AWS has precedence over Gcloud. Raising exception if both settings are defined could also be an option, but having the error on django runserver time can be an inconvenience.

### Pull Request Checklist

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -400,7 +400,7 @@ if 'GOOGLE_APPLICATION_CREDENTIALS' not in os.environ:
     GS_CREDENTIALS = os.environ.get('GS_CREDENTIALS')
 
 GS_PROJECT_ID = os.environ.get('GS_PROJECT_ID')
-GS_AUTO_CREATE_BUCKET = os.environ.get('GS_AUTO_CREATE_BUCKET', False)
+GS_AUTO_CREATE_BUCKET = get_bool_from_env('GS_AUTO_CREATE_BUCKET', False)
 
 if AWS_STORAGE_BUCKET_NAME:
     STATICFILES_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -389,11 +389,26 @@ AWS_S3_CUSTOM_DOMAIN = os.environ.get('AWS_STATIC_CUSTOM_DOMAIN')
 AWS_SECRET_ACCESS_KEY = os.environ.get('AWS_SECRET_ACCESS_KEY')
 AWS_STORAGE_BUCKET_NAME = os.environ.get('AWS_STORAGE_BUCKET_NAME')
 
+# Google cloud storage
+GS_BUCKET_NAME = os.environ.get('GS_BUCKET_NAME')
+GS_MEDIA_BUCKET_NAME = os.environ.get('GS_MEDIA_BUCKET_NAME')
+GS_PROJECT_ID = os.environ.get('GS_PROJECT_ID')
+GS_CREDENTIALS = os.environ.get('GS_CREDENTIALS')
+GS_AUTO_CREATE_BUCKET = os.environ.get('GS_AUTO_CREATE_BUCKET', False)
+
+# Alternatively keep the ENV varibable GOOGLE_APPLICATION_CREDENTIALS instead of PROJ_ID and others
+
 if AWS_STORAGE_BUCKET_NAME:
     STATICFILES_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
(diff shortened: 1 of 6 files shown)


sergioisidoro:
Please be aware of https://github.com/jschneier/django-storages/issues/491
One option is to wait for this to get merged, another is to implement the same workaround.

saleor/settings.py
```diff
@@ -384,11 +384,29 @@ def get_host():
...
 AWS_STORAGE_BUCKET_NAME = os.environ.get('AWS_STORAGE_BUCKET_NAME')
 
+# Google cloud storage
+GS_STORAGE_BUCKET_NAME = os.environ.get('GS_STORAGE_BUCKET_NAME')
+GS_MEDIA_BUCKET_NAME = os.environ.get('GS_MEDIA_BUCKET_NAME')
+
+# If GOOGLE_APPLICATION_CREDENTIALS is set with path to JSON credentials,
+# there is no need to load OAuth token
+# See https://django-storages.readthedocs.io/en/latest/backends/gcloud.html
+if not os.environ.get("GOOGLE_APPLICATION_CREDENTIALS"):
```
NyanKiyoshi: Don't retrieve the value if you don't need it, just do the following instead:
```python
if 'GOOGLE_APPLICATION_CREDENTIALS' not in os.environ
```

sergioisidoro: Totally right - Copy pasta to blame
Thanks!

sergioisidoro:
Hi ! 
I've tested that this works in a staging environment. But this PR is getting stale.
What should be done to get this forward?

maarcingebala:
Hey, thanks for this PR! What we should only change is to rebase it against the current `master` to resolve conflicts in requirements files. We have recently switched to `pipenv` and don't use `requirements.in` anymore. Instead, we should add necessary packages to `Pipfile` and generate a new `Pipfile.lock`. Would you be able to take care of that?

sergioisidoro:
Great. I'll try to rebase this tonight 🙂

sergioisidoro:
Hey there !

So, I had to re-write some of the history with rebase for clarity, and to make the git history a bit more understandable.

I also did:
```
pipenv lock -r -d > requirements_dev.txt  
pipenv lock -r > requirements.txt  
```
for pip requirements backwards compatibility. Most docs still reference `pip install -r requirements.txt` so people who start for the first time in this project might have a harsh time the first time they try to setup the env. I'm not sure if I should or if you have any CI making that.

Also, please note that `security/snyk` also uses requirements.txt so I guess keeping the file updated is a requirement?

Anyway, tell me what else can I do. 
BR.
Sergio

NyanKiyoshi:
Sounds good to me. Everything in the [guide](https://docs.getsaleor.com/en/latest/customization/backend.html) is respected.

NyanKiyoshi:
I still wonder, though, wouldn't it better to put the google cloud requirements apart as optional requirements? But I never figured how to do add extras requires in pipenv (from what I understood in the past, was it's not possible), e.g. doing: `pip install -e .[google-cloud]`.

sergioisidoro:
@NyanKiyoshi yeah, it kind of makes sense, but then should we also do the same for AWS related packages? I guess we would need to add safeguards to the imports as well, as the packages might not be installed.

NyanKiyoshi:
Yes, but I don't know how this should handled, it's up to Mirumee to decide what is the proper way. I guess that if we assume that Saleor should be fully configurable without any additional requirements (e.g.: give the staff a way to edit storages from the dashboard for example), we would not create extra dependencies, it would not make sense. Just like what is planned: adding a way to configure payment methods from the dashboard, so we need all the providers to be ready to be up and running.

Then, if this behavior is, as it seems, not supported by pipenv, I guess that it would be safer to not just separate as it would add another way of installing packages, etc. (through a `setup.py` file), which would really confuse the users more than anything else.

sergioisidoro:
@maarcingebala Status? Another merge has created conflicts in the requirements. I guess I could update the requirements again, but it would be of no use if this gets stalled again.

maarcingebala:
Hey @sergioisidoro, sorry for no response from our side! We had been very busy with the release of Saleor 2.0. We also had some problems with `Pipfile.lock` due to broken dependencies between packages. Now everything should be all right. Do you mind updating this PR once again and pinging me once you do it?

maarcingebala:
Closing in favor of #4127

PR #2635, opened 2018-08-20
[WIP] Fix language codes

### LINKED ISSUE #2634: LANGUAGES = ('zh-tw', _('Chinese(Taiwan)')) needs to be changed into 'zh-hant'
...
Fix the languages problem of Chinese (Taiwan/HongKong)

...
Change `LANGUAGES = ('zh-tw', _('Chinese(Taiwan)'))` into 
`LANGUAGES = ('zh-hant', _('Chinese(Taiwan)'))`
…

### Screenshots or mockups

wx20180820-150041 2x

Pacu2:
Fixed incorrect language codes close #2634

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -63,7 +63,7 @@ LANGUAGES = [
     ('de', _('German')),
     ('en', _('English')),
     ('es', _('Spanish')),
-    ('fa-ir', _('Persian (Iran)')),
+    ('fa', _('Persian (Iran)')),
     ('fr', _('French')),
     ('hu', _('Hungarian')),
     ('it', _('Italian')),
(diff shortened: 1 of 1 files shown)


saleor/settings.py
```diff
@@ -63,7 +63,7 @@ def get_bool_from_env(name, default_value):
     ('de', _('German')),
     ('en', _('English')),
     ('es', _('Spanish')),
-    ('fa-ir', _('Persian (Iran)')),
+    ('fa', _('Persian (Iran)')),
```
patrys: This is incorrect. `fa` is generic Farsi (Persian), not the Iranian variant.

saleor/settings.py
```diff
@@ -80,7 +80,7 @@ def get_bool_from_env(name, default_value):
     ('uk', _('Ukrainian')),
     ('vi', _('Vietnamese')),
     ('zh-hans', _('Chinese')),
-    ('zh-tw', _('Chinese (Taiwan)'))]
+    ('zh-hant', _('Chinese (Taiwan)'))]
```
patrys: These two should get renamed to "Chinese (simplified)" and "Chinese (traditional)" respectively.

patrys: We also need to rename the locale directories, otherwise both will fall back to `zh`.

PR #2643, opened 2018-08-20
Unify mutation handling

### LINKED ISSUE #2640: Unify mutation handling
We need to rework how mutations are handled and what they return to avoid complex logic on the client side.

patrys:
Introduce TypedMutation, a mutation class factory that centralizes the logic.

Fixes #2640 

### Pull Request Checklist


akjanik:
In `ProductImage(Create/Update/Delete)` mutation, does changing from `ModelMutation` to just `BaseMutation` simplify using them (for example in frontend)? Or is there another reason we do not like `ModelMutation` anymore?
I see some repeated logic there, but what more concerns me, that now accessing some data, for example `alt` of an image is a bit more complicated
https://github.com/mirumee/saleor/pull/2643/files#diff-7e654175e9389b4b0155558cb4dd534aR894

patrys:
@akjanik We want to return more than the instance so that all clients does not need to reimplement the same cache updating logic.

Accessing data from mutations is usually not an issue as it's mostly there to feed the cache.

I've updated the code to reflect my recent discussion with @maarcingebala so please re-review.

PR #2652, opened 2018-08-21
Handle unpublished menu items

### LINKED ISSUE #2137: Handle not published items in Menu
At this moment, one can add not published (Collection / Page) to the menu. Such item will be listed on the Storefront, but cannot be viewed by regular user(returning 404-instead).

My proposition is:
- [ ] Display a warning when one adds not published item(Collection / Page) to the Menu
- [ ] Display a warning if any MenuItem contains not published Collection / Page 
~- [ ] Hide items linking to not published instances, on the storefront~

Related discussion in #2109

oldPadavan:

closes #2137 

### Screenshots

2018-09-06-115119_1920x951_scrot
2018-09-06-115127_1920x951_scrot

### Pull Request Checklist

### DIFF
--- a/saleor/dashboard/menu/utils.py
+++ b/saleor/dashboard/menu/utils.py
@@ -100,8 +100,10 @@ def get_menu_obj_text(obj):
     elif isinstance(obj, Page) and obj.is_visible and obj.available_on:
         return pgettext(
             'Menu item page hidden status',
-            'Hidden (will become visible on %s)' % (localize(
-                obj.available_on)))
+            'Hidden (will become visible on %(available_on_date)s)' % (
+                {'available_on_date': localize(obj.available_on)}))
     else:
         return pgettext(
-            'Menu item published status', '%s (Not published)' % str(obj))
+            'Menu item published status',
+            '%(menu_item_name)s (Not published)' % {
+                'menu_item_name': str(obj)})
--- a/saleor/dashboard/menu/utils.py
+++ b/saleor/dashboard/menu/utils.py
@@ -102,8 +102,7 @@ def get_menu_obj_text(obj):
             'Menu item page hidden status',
             'Hidden (will become visible on %(available_on_date)s)' % (
                 {'available_on_date': localize(obj.available_on)}))
-    else:
-        return pgettext(
-            'Menu item published status',
-            '%(menu_item_name)s (Not published)' % {
-                'menu_item_name': str(obj)})
+    return pgettext(
+        'Menu item published status',
+        '%(menu_item_name)s (Not published)' % {
+            'menu_item_name': str(obj)})
--- a/saleor/dashboard/menu/utils.py
+++ b/saleor/dashboard/menu/utils.py
@@ -1,4 +1,5 @@
 from django.db.models import Q
+from django.utils.translation import pgettext
 
 from ...menu.models import Menu, MenuItem
 from ...page.models import Page
@@ -90,3 +91,8 @@ def get_menus_that_needs_update(collection=None, categories=None, page=None):
     menus_to_be_updated = MenuItem.objects.filter(q).distinct().values_list(
         'menu', flat=True)
     return menus_to_be_updated
+
+
+def get_menu_obj_text(obj):
+    return str(obj) + ('' if obj.is_published else pgettext(
+        'Menu item pubslished status', ' (Not published)'))
--- a/saleor/dashboard/menu/utils.py
+++ b/saleor/dashboard/menu/utils.py
@@ -95,7 +95,7 @@ def get_menus_that_needs_update(collection=None, categories=None, page=None):
 
 
 def get_menu_obj_text(obj):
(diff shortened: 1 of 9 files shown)


saleor/dashboard/menu/views.py
```diff
@@ -120,7 +119,6 @@ def menu_item_create(request, menu_pk, root_pk=None):
         msg = pgettext_lazy(
             'Dashboard message', 'Added menu item %s') % (menu_item,)
         messages.success(request, msg)
-        update_menu(menu)
```
Pacu2: Why is that? The menu is kept as `JSON` and we need to update it each time it changed.

oldPadavan: Hm... Very strange. It seems that I accidentally restored some changes from disk which caused these lines to be deleted and I missed it when reviewed my own commit. I apologize for inconvenience and will correct it tomorrow

saleor/dashboard/menu/utils.py
```diff
@@ -92,3 +93,8 @@ def get_menus_that_needs_update(collection=None, categories=None, page=None):
     menus_to_be_updated = MenuItem.objects.filter(q).distinct().values_list(
         'menu', flat=True)
     return menus_to_be_updated
+
+
+def get_menu_obj_text(obj):
+    return str(obj) + ('' if obj.is_published else pgettext(
```
Pacu2: So I have a doubt if that would work properly (Adding a translated string to nontranslated one), it might be not clear for the translators.

I'd say we need 3 cases here:
- Regular menu item
- Menu item available on a certain date (from the future)
- Menu item not published

and I'd add individual translation for 2nd and 3rd one

oldPadavan: You mean it's necessary to split the status "Not published" into 2 cases: 1) Not published collection or invisible page 2) Visible page since some data from the future?
Regarding concatenation of translated string with non-translated - it's possible to add an additional field "status" which will contain translated string to JSON response and join "text" and "status" on front-side, but I'm afraid it can complicate a little bit JS code - now there's common initialization for all selects. What would You advice?

Pacu2: I’d just split it with if-else statement.

If not obj.is_published:
   Return pgettext(„Some description”, „%(page_name)s (not published)”  % str(page))
Elif obj.available_on > today:
   Return ...
Return str(obj)

oldPadavan: Ok, then I'll add additional status for menu items visible from a certain date in other places where `MenuItem` status appears as well

saleor/product/models.py
```diff
@@ -52,6 +52,10 @@ def get_absolute_url(self):
             'product:category',
             kwargs={'slug': self.slug, 'category_id': self.id})
 
+    @property
+    def is_published(self):
```
Pacu2: I think this is unnecessary, we could just use a `getattr` with a fallback to `True` if it's not present.

saleor/dashboard/menu/utils.py
```diff
@@ -92,3 +94,16 @@ def get_menus_that_needs_update(collection=None, categories=None, page=None):
...
     return menus_to_be_updated
+
+
+def get_menu_obj_text(obj):
+    if obj.is_published:
+        return str(obj)
+    elif isinstance(obj, Page) and obj.is_visible and obj.available_on:
+        return pgettext(
+            'Menu item page hidden status',
+            'Hidden (will become visible on %s)' % (localize(
```
Pacu2: For translations, we should always use a meaningful formatting. 

Translator won't know what goes under `%s`, formatting it as `%(available_on_date)s` will give a nice feel, of how it will look like on the real site.

saleor/product/models.py
oldPadavan: Sorry, I didn't understand Your proposal. My idea was to have `is_published` property in `Page`, `Category` and `Collection` to have possibility to check published status of `MenuItem` as `linked_object.is_published`

Pacu2: Instead of using `linked_object.is_published` you can use `getattr(linked_object, 'is_published', True)`

Which will return True if `is_published` is not present on the model.

oldPadavan: Got it, thanks!

saleor/dashboard/menu/utils.py
```diff
@@ -92,3 +94,18 @@ def get_menus_that_needs_update(collection=None, categories=None, page=None):
...
+
+def get_menu_obj_text(obj):
+    if getattr(obj, 'is_published', True):
+        return str(obj)
+    elif isinstance(obj, Page) and obj.is_visible and obj.available_on:
+        return pgettext(
+            'Menu item page hidden status',
+            'Hidden (will become visible on %(available_on_date)s)' % (
+                {'available_on_date': localize(obj.available_on)}))
+    else:
```
Pacu2: This else is unnecessary as above ifs returns anyway

PR #2668, opened 2018-08-23
Enable Pythonic environment variables

ajostergaard:
I want to merge this change because it makes environment variable more Pythonic (intuitive) and reduces newby error.

No issues reported as far as I know but I did see one on Gitter:
Stephen Moloney @stephenmoloney 14:56
I though that get_bool_from_env would coerce it...
to False

Not sure how to add tests for this but if that's possible, happy to do so.

### Pull Request Checklist

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -18,6 +18,10 @@ def get_list(text):
 def get_bool_from_env(name, default_value):
     if name in os.environ:
         value = os.environ[name]
+        # Enable Pythonic environment variables,
+        # i.e. DEBUG='' or DEBUG= is Falsy but would generate an error
+        if value == '':
+            return False
         try:
             return ast.literal_eval(value)
         except ValueError as e:


saleor/settings.py
```diff
@@ -18,6 +18,10 @@ def get_list(text):
 def get_bool_from_env(name, default_value):
     if name in os.environ:
         value = os.environ[name]
+        # Enable Pythonic environment variables,
+        # i.e. DEBUG='' or DEBUG= is Falsy but would generate an error
+        if value == '':
+            return False
```
NyanKiyoshi: The value should be defaulting to to whatever is `default_value`. If the value is \*not\* set. Which was the expected behavior from him. 

Otherwise it doesn't make sense (for me at least).

I would suggest to replace `if name in os.environ: ...` by a `if value: ...` which would handle the case where the value is fully unset (not in the environment) or partially unset (an empty placeholder in the environment).

I also think that we probably shouldn't use `ast.literal_eval(value)` for parsing a boolean, as we don't have any use to compile the argument to Python code. A simple string check that would say whether the value is considered to be the `True` or `False` literal would be enough. Maybe we could drop the case sensitivity.

ajostergaard: I disagree regarding the default, a Python user coming to this would expect, at least, "" and '' to equate to False ala bool("").

I agree that the function could be improved but am not expecting the powers that be to entertain large changes which would require much more testing.

stephenmoloney: I had expected a variable which was explicitly set as follows: `VAR=` to default to a falsey response... ie `False`. I'm not a python developer though so I don't know the ways things normally go.

jxltom: Explicit is better than implicit :D
Just set set ```VAR=0``` or any other false value to make it work as ```False```

PR #2678, opened 2018-08-24
Unify resolving nodes in Graphene

maarcingebala:
This PR unifies the way we resolve nodes from global IDs. Assumptions about the new approach:
- return `null` when ID is unresolved in queries
- return a user error when ID is unresolved in mutations

### Pull Request Checklist


maarcingebala:
@patrys I'm planning to open more similar PRs soon to clean up this code, I'll try to do something about this method as well.

PR #2685, opened 2018-08-24
Make Decimal fields null if no value supplied

### LINKED ISSUE #2672: productVariantCreate mutation fails
Mutation `productVariantCreate` fails in the new dashboard when creating a variant without "cost price" or "price override" values.

Mutation variables that resulted in an error:

```
{
  "attributes": [
    {
      "slug": "cover",
      "value": "soft"
    }
  ],
  "costPrice": "",
  "priceOverride": "",
  "product": "UHJvZHVjdDo2Mg==",
  "quantity": "5",
  "sku": "cover-soft",
  "trackInventory": true
}
```

Error:
```
{"errors":[{"message":"Variable \"$costPrice\" got invalid value \"\".\nExpected type \"Decimal\", found \"\".","locations":[{"line":103,"column":60}]}]}
```

The problem is that the fields that failed are of `Decimal` type which doesn't accept an empty string `""`  value. I ran the mutation manually and set `null` instead of empty strings and it worked.

Create a product variant in the new dashboard.

1. Create a new product.
2. Try to add a new variant with empty "cost price" and "price override" fields.

### What I expected to happen
A product variant should be added.

dominik-zeglen:
I want to merge this change because it fixes bug causing error when no `costPrice` or `priceOverride` were provided.
Fixes #2672 

### Pull Request Checklist

### DIFF
--- a/saleor/static/dashboard-next/products/views/ProductVariantCreate.tsx
+++ b/saleor/static/dashboard-next/products/views/ProductVariantCreate.tsx
@@ -12,6 +12,7 @@ import {
   productVariantCreateQuery,
   TypedProductVariantCreateQuery
 } from "../queries";
+import { decimal } from "../../misc";
 
 interface ProductUpdateProps {
   productId: string;
@@ -77,12 +78,8 @@ export const ProductVariant: React.StatelessComponent<ProductUpdateProps> = ({
                       variantCreate({
                         variables: {
                           attributes: formData.attributes,
-                          costPrice: formData.costPrice
-                            ? formData.costPrice
-                            : null,
-                          priceOverride: formData.priceOverride
-                            ? formData.priceOverride
-                            : null,
+                          costPrice: decimal(formData.costPrice),
+                          priceOverride: decimal(formData.priceOverride),
                           product: productId,
                           quantity: formData.stock,
                           sku: formData.sku,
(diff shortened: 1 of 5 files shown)


maarcingebala:
Do we have any other occurrences of `Decimal` variables? I'd make sure that we cover all of them.

saleor/static/dashboard-next/products/views/ProductVariantCreate.tsx
```diff
@@ -79,8 +79,12 @@ export const ProductVariant: React.StatelessComponent<ProductUpdateProps> = ({
                       variantCreate({
                         variables: {
                           attributes: formData.attributes,
-                          costPrice: formData.costPrice,
-                          priceOverride: formData.priceOverride,
+                          costPrice: formData.costPrice
```
patrys: What was submitted before? Shouldn't there be a helper function to do it?

dominik-zeglen: An empty string. It seems that we need to pass either string (which cannot be empty) or null. The problem is that cannot solve this problem by using just TS types, unless we pass `--passthrough-custom-scalars` option to `apollo-codegen` and write constructor for all those custom types.

dominik-zeglen:
Now all decimals had been covered.

PR #2699, opened 2018-08-27
Lock django version smaller than 2.1 since django-filter 1.1 doesn't support 2.1

jxltom:
Saleor requires django-filter 1.1 but django-filter 1.1 does not support django 2.1 since django 2.1 removes ```django.db.models.sql.constants.QUERY_TERMS``` but django-filter 1.1 will try to import it.

So ```Django>=1.11,<2.1``` is added in ```requirements.in```

### Pull Request Checklist

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -2,7 +2,7 @@ Babel>=2.5
 bleach<3.0
 boto3
 cairosvg
-Django>=1.11
+Django>=1.11,<2.1
 dj_database_url>=0.3.0
 dj_email_url>=0.0.4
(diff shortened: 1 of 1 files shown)


jxltom:
Refer to https://github.com/django/django/pull/10227

patrys:
We want to move to Django 2.1 and `django-filter` 2.x but are currently blocked by `graphene-django`.

jxltom:
OK. django-filter 2.0 has already supported django 2.1

PR #2711, opened 2018-08-28
Send payment notification on payment status change.

BenSturmfels:
I could be wrong, but I think this is meant to say `send_payment_notification` rather than `send_order_notification`.

Before this change, the order notifications are sent twice (once after order placed, once after payment) while payment notifications aren't sent at all.

With this change, the order notification is sent after order placed and the payment notification is sent after payment is confirmed.

### DIFF
--- a/saleor/order/signals.py
+++ b/saleor/order/signals.py
@@ -3,7 +3,7 @@ import logging
 from django.utils.translation import pgettext_lazy
 
 from ..core import analytics
-from .emails import send_order_confirmation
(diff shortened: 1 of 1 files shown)


BenSturmfels:
Thanks for all your work on Saleor by the way - I'm really enjoying working the the system.

maarcingebala:
Good catch! Thanks for your first contribution to Saleor. I'm happy that it works for you well!

PR #2726, opened 2018-08-28
getsaleor.com Add production webpack config

benekex2:
Ref mirumee/mirumee#496

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -7,10 +7,11 @@ const CleanWebpackPlugin = require('clean-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const PostcssPresetEnv = require('postcss-preset-env');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
+const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin');
 
 const sourceDir = path.join(__dirname, './src/');
 const distDir = path.join(__dirname, './dist/');
-const devMode = process.env.NODE_ENV !== 'production';
+const devMode = process.argv.indexOf('-p') < 0;
 
 module.exports = {
   resolve: {
@@ -30,7 +31,7 @@ module.exports = {
   },
   output: {
     path: distDir,
-    filename: 'js/[name].js',
+    filename: devMode ? 'js/[name].js' : 'js/[name].[contenthash].js',
     publicPath: '/',
   },
   devtool: 'source-map',
@@ -78,7 +79,7 @@ module.exports = {
           {
             loader: 'file-loader',
             options: {
-              name: '[name].[ext]',
+              name: '[name].[hash].[ext]',
               outputPath: 'fonts/',
             },
           },
@@ -90,7 +91,7 @@ module.exports = {
           {
             loader: 'file-loader',
             options: {
-              name: '[name].[ext]',
+              name: '[name].[hash].[ext]',
               outputPath: 'images/',
             }
           },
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -7,10 +7,11 @@ const CleanWebpackPlugin = require('clean-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const PostcssPresetEnv = require('postcss-preset-env');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
+const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin');
 
 const sourceDir = path.join(__dirname, './src/');
 const distDir = path.join(__dirname, './dist/');
-const devMode = process.env.NODE_ENV !== 'production';
+const devMode = process.argv.indexOf('-p') < 0;
```
mociepka: Why that? Normally we use environments.

benekex2: Because webpack does not set `process.env.NODE_ENV` is always `undefined`

mociepka: It does https://webpack.js.org/api/cli/#shortcuts

mociepka: But we shouldn't use `-p`. We are defining npm scripts but the division between builds should be done on the build step.

Building for production should be run as:
```
NODE_ENV=production npm build
```

Without the env a development build should be run. With envs it is more elastic (easy for docker, heroku, aws etc.)

patrys: We should use `-p`. And it is available inside the config:

```js
module.exports = (env, argv) => {
  if (argv.mode === "production") {
    console.log("WOO");
  }
  return {
    // ...
  }
}
```

mociepka: If we go with `-p` how production build should be run? You want to run it:

1. explicit `webpack -p`
2. by script `npm run build -- -p`
3. introduce new script `npm run build:prod`

Which is why I prefer env approach. In addition `-p` sets also `--optimize-minimize` which inject `UglifyJsPlugin` I prefer to set it manually.

patrys: We usually have `build` that launches `webpack -p` and `start` that launches `webpack -d --watch`.

mirekm: If everything else in the entire Saleor system is differentiated by the env variable I would probably lean toward `NODE_ENV=production` with consistency in mind.

benekex2: Or we should split config and manually set env: 
`new webpack.DefinePlugin({ "process.env.NODE_ENV": JSON.stringify("production") }),`

dominik-zeglen: I don't get it. Why would we want production build without optimizations?
And if we don't, what do we get by explicitly setting them in the config?

patrys: Why invent our own way of handling Webpack modes?

Webpack already has mode switching built into it. It does not depend on the environment, it's explicitly passed on the command line. (In fact we can't `export NODE_ENV=production` as that would break `npm install`).

Just do what the docs say and instead of exporting a config object, export a function that accepts `env` and `argv` and use `argv.mode` to determine which mode was requested on the command line.

We don't have to maintain separate config files for production and development. We don't need to merge configs in code.

The default plugins set by production mode and development mode are good, there's no reason to list them manually. Same for the devtool config.

patrys: FYI: we also plan to switch Saleor's main Webpack config to use `argv.mode`. Currently there are surprising situations where Webpack thinks it's running in production mode but our config tells it to use the development build of React because `NODE_ENV` was not set.

mociepka: Having those all comments in mind I propose to use one config both for production and development. 
The distinction between the two will be made by `const devMode = argv.mode !== "production"` and contitions inside the `webpack.config.js` file.

For building perspective command `npm build` will be running `webpack -p` which is always a production build (we won't be using `NODE_ENV ` env from the outside at all).  From working perspective command `npm start` will be running `webpack -d --watch`.

Did I missed somethng?

dominik-zeglen: Looks good to me 👍

PR #2729, opened 2018-08-28
2614/Add protocol option to email template context

### LINKED ISSUE #2614: [Templates] Logo URL is broken in mail body when remove account
Logo broken in mail body when remove account

1. login
2. enter to account page
3. click "remove account"
4. see broken logo url, must be like http://localhost/static/images/logo-document.svg , but got http://127.0.0.1/en/products/mccarthy-llc-2/%3C%3C%20MISSING%20VARIABLE%20%22protocol%22%20%3E%3E://localhost:80/static/images/logo-document.svg . In another mail (in my case order details logo show normal).

### Screenshots
7b790d456a 1

To read mail use: https://github.com/mirumee/saleor/issues/2610

tomcio1205:
I want to merge this change because it resolves #2614

### DIFF
--- a/saleor/account/emails.py
+++ b/saleor/account/emails.py
@@ -26,7 +26,11 @@ def send_account_delete_confirmation_email(token, recipient_email):
     site = Site.objects.get_current()
     delete_url = build_absolute_uri(
         reverse('account:delete-confirm', kwargs={'token': token}))
-    ctx = {'site_name': site.name, 'domain': site.domain, 'url': delete_url}
+    ctx = {
+        'site_name': site.name,
+        'domain': site.domain,
+        'url': delete_url,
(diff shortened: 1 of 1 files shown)


Pacu2:
Thanks! Your changes indeed solve the problem, therefore I think we should not glue the URL on the templates(which leads to the bugs like this one), instead, we should pass the full absolute URL as the context variable, from the view to the template

Pacu2:
I've created an issue for fixing the URL creation in email's templates #2730

tomcio1205:
@Pacu2 so I must change all templates which are uses the header.mjml shared file. Should I do this on this branch ?

tomcio1205:
@Pacu2  I think it could be close now, because pr #2734 also resolve this issue

PR #2734, opened 2018-08-29
2730/Remove gluing urls in templates

### LINKED ISSUE #2614: [Templates] Logo URL is broken in mail body when remove account

### LINKED ISSUE #2730: Get rid of manually generating URLs in the email templates
In some places, in the emails templates we are creating full absolute URLs in such ways:
```
{{ protocol}}://{{ domain }}{% url 'account:reset-password-confirm' uidb64=uid token=token %}
```
Which leads to hard to catch bugs. 

All URLs should be created on the backend, passed to the templates, instead of gluing protocol/domain/URL manually in the templates. 
This could be also an template tag.

tomcio1205:
I want to merge this change because it resolve #2730, also resolve #2614 

### Pull Request Checklist


templates/templated_email/account/password_reset.email
```diff
@@ -8,11 +8,11 @@
 {% blocktrans context "Password reset e-mail text" %}
 Hi!
 
-You're receiving this e-mail because you or someone else has requested a password for your user account at {{ domain }}.
+You're receiving this e-mail because you or someone else has requested a password for your user account at {{ absolute_url }}.
 It can be safely ignored if you did not request a password reset. Click the link below to reset your password.
 {% endblocktrans %}
 
-{{ protocol}}://{{ domain }}{% url 'account:reset-password-confirm' uidb64=uid token=token %}
+{{ absolute_url }}{% url 'account:reset-password-confirm' uidb64=uid token=token %}
```
Pacu2: Could we generate the whole url on the backend?  Like, it's better now, but still joining two parts in the template.

saleor/account/emails.py
```diff
@@ -24,12 +30,14 @@ def send_password_reset_email(context, recipient):
 @shared_task
 def send_account_delete_confirmation_email(token, recipient_email):
     site = Site.objects.get_current()
-    absolute_url = build_absolute_uri(location=None)
+    logo_url = build_absolute_uri(
```
NyanKiyoshi: Could you refactor a little the code remove the repeated code? For example, have ` get_email_base_context()` to return the context (`{'sitename': ..., 'domain': ..., 'logo_url':: ...}`) then you add whatever you need to the retrieved base context.

That would be great :+1:

saleor/account/emails.py
```diff
@@ -23,12 +27,22 @@ def send_password_reset_email(context, recipient):
...
+    ctx = get_email_base_context()
+    ctx.update({'url': delete_url})
     send_templated_mail(
         template_name='account/account_delete',
         from_email=settings.DEFAULT_FROM_EMAIL,
         recipient_list=[recipient_email],
         context=ctx)
+
+
+def get_email_base_context():
```
NyanKiyoshi: It would be better to define it to `core.email` instead of duplicating it, otherwise it doesn't fix the issue I mentioned. :+1:

saleor/account/emails.py
```diff
@@ -23,12 +27,22 @@ def send_password_reset_email(context, recipient):
 
 @shared_task
 def send_account_delete_confirmation_email(token, recipient_email):
-    site = Site.objects.get_current()
     delete_url = build_absolute_uri(
         reverse('account:delete-confirm', kwargs={'token': token}))
-    ctx = {'site_name': site.name, 'domain': site.domain, 'url': delete_url}
+    ctx = get_email_base_context()
+    ctx.update({'url': delete_url})
```
NyanKiyoshi: I would recommend not calling `update(...)` (everywhere you put it) as there is no use to do that, it costs more CPU time for nothing in the current context.

Just do `ctx['delete_url'] = delete_url`

saleor/order/emails.py
```diff
@@ -18,13 +19,11 @@
...
+    logo_url = build_absolute_uri(
+        location=None) + static('images/logo-document.svg')
     order_url = build_absolute_uri(
         reverse('order:details', kwargs={'token': order_token}))
-    ctx = {
-        'protocol': 'https' if settings.ENABLE_SSL else 'http',
-        'site_name': site.name,
-        'domain': site.domain,
-        'url': order_url}
+    ctx = {'site_name': site.name, 'logo_url': logo_url, 'url': order_url}
```
NyanKiyoshi: That one is still a duplicated code as well.

tomcio1205:
@NyanKiyoshi thanks for all suggestions. Can You tell me if core.utils.email is also good location for store this method?

NyanKiyoshi:
@tomcio1205 to define all bases from emails, it would be better to put them inside the core module (`core.emails`) especially as getting the base email context wouldn't really be an utility call. In my opinion, of course.

tomcio1205:
@NyanKiyoshi so as You suggested, I moved this method to core.emails :)

saleor/core/emails.py
```diff
@@ -0,0 +1,14 @@
+from django.contrib.sites.models import Site
+from django.templatetags.static import static
+
+from ..core.utils import build_absolute_uri
+
+
+def get_email_base_context():
+    site = Site.objects.get_current()
+    logo_url = build_absolute_uri(
+        location=None) + static('images/logo-document.svg')
```
Pacu2: We should not build urls manually in any case. Instead of `None` you might to pass an actual `static` url to the `build_absolute_uri` function.

`static` will resolve to full absolute url if one uses, for eg. external cloud service to serve the images.
```
In [2]: build_absolute_uri('http://google.pl')
Out[2]: 'http://google.pl'

In [4]: build_absolute_uri(location=None) + 'http://google.pl'
Out[4]: 'http://localhost:8000http://google.pl'
```

saleor/account/emails.py
```diff
@@ -12,8 +12,11 @@ def send_password_reset_email(context, recipient):
     reset_url = build_absolute_uri(
         reverse(
             'account:reset-password-confirm',
-            kwargs={'uidb64': context['uid'], 'token': context['token']}))
-    context['reset_url'] = reset_url
+            kwargs={
+                'uidb64': context['uid'],
+                'token': context['token']}))
+    context = get_email_base_context()
+    context.update({'reset_url': reset_url})
```
Pacu2: This update is unnecessary as we can modify the dict directly.

saleor/core/emails.py
patrys: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

tests/test_emails.py
```diff
@@ -12,8 +12,7 @@
 
 def test_get_email_base_context(site_settings):
     site = site_settings.site
-    logo_url = build_absolute_uri(
-        location=None) + static('images/logo-document.svg')
+    logo_url = build_absolute_uri(static('images/logo-document.svg'))
```
NyanKiyoshi: Could you, if you have some spare time, add a test to ensure the S3 bucket case is correctly handled? This would fix on of the issue mentioned in #1999.

That would be great :tada:

tomcio1205: @NyanKiyoshi I read this issue but I'm not pretty sure how can I test it. Could you explain what exactly should be tested?

NyanKiyoshi: Here is an example of a test testing the set-up of a S3 bucket:
https://github.com/mirumee/saleor/blob/cefc896d18aba433c5baee92345760e0a4ed5397/tests/test_core.py#L240-L246

Pacu2: We could just mock `static` to return a full URL instead of presetting the whole S3 bucket

saleor/account/emails.py
```diff
@@ -9,14 +9,14 @@
...
-    reset_url = build_absolute_uri(
+    url = build_absolute_uri(
         reverse(
             'account:reset-password-confirm',
             kwargs={
                 'uidb64': context['uid'],
                 'token': context['token']}))
     context = get_email_base_context()
-    context['reset_url'] = reset_url
+    context['url'] = url
```
NyanKiyoshi: Could you keep meaningful variable names everywhere for the URLs? (like it was before, `reset_url`, `password_set_url`)

saleor/account/emails.py
```diff
@@ -23,10 +26,10 @@ def send_password_reset_email(context, recipient):
 
 @shared_task
 def send_account_delete_confirmation_email(token, recipient_email):
-    site = Site.objects.get_current()
     delete_url = build_absolute_uri(
         reverse('account:delete-confirm', kwargs={'token': token}))
-    ctx = {'site_name': site.name, 'domain': site.domain, 'url': delete_url}
+    ctx = get_email_base_context()
+    ctx['url'] = delete_url
```
Pacu2: I think this URL could use more meaningful name

saleor/dashboard/emails.py
```diff
@@ -28,12 +29,8 @@ def send_set_password_email(staff):
...
 @shared_task
 def send_promote_customer_to_staff_email(staff):
-    site = Site.objects.get_current()
-    ctx = {
-        'protocol': 'https' if settings.ENABLE_SSL else 'http',
-        'domain': site.domain,
-        'url': build_absolute_uri(reverse('dashboard:index')),
-        'site_name': site.name}
+    ctx = get_email_base_context()
+    ctx['url'] = build_absolute_uri(reverse('dashboard:index'))
```
Pacu2: I think this URL could use more meaningful name

saleor/order/emails.py
```diff
@@ -37,7 +24,9 @@ def collect_data_for_email(order_pk, template):
     """
     order = Order.objects.get(pk=order_pk)
     recipient_email = order.get_user_current_email()
-    email_context = get_email_context(order.token)
+    email_context = get_email_base_context()
+    email_context['url'] = build_absolute_uri(
```
Pacu2: I think this URL could use more meaningful name

templates/templated_email/source/confirm_note.mjml
```diff
@@ -14,7 +14,7 @@
         </mj-text>
         <mj-text>
           {% blocktrans trimmed context "Note confirmation e-mail text" %}
-            A new note has been added to your order. To see details please visit: <a href="{{ url }}">{{ url }}</a>
+            A new note has been added to your order. To see details please visit: <a href="{{ order_details_url }}">{{ order_details_url }}</a>
```
maarcingebala: The old `url` context variable is also used in the plain-text version of this email in file "confirm_note.email", we should change it there as well. Except for that, I think we have covered all of the URLs.

maarcingebala:
Thanks for your effort @tomcio1205 🎉

PR #2738, opened 2018-08-29
Allow creating menu items in menu create mutation

### LINKED ISSUE #2720: Create menu items in menuCreate mutation
It would be very handy if API user would be able to create the first level of menu items at the menu creation moment. This way we could reduce the number of database queries needed to create a menu and speed up the whole process. 

Change `MenuInput` signature to
```
type MenuInput {
  name: String!
  items: [MenuItemCreate!]
}
```

maarcingebala:
Fixes #2720

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/menu/mutations.py
+++ b/saleor/graphql/menu/mutations.py
@@ -3,22 +3,13 @@ from graphql_jwt.decorators import permission_required
 
 from ...menu import models
 from ..core.mutations import BaseMutation, ModelDeleteMutation, ModelMutation
+from ..product.types import Category, Collection
+from ..page.types import Page
 from .types import Menu
 
 
-class MenuInput(graphene.InputObjectType):
-    name = graphene.String(description='Name of the menu.')
-
-
 class MenuItemInput(graphene.InputObjectType):
-    menu = graphene.ID(
-        description='Menu to which item belongs to.', name='menu')
     name = graphene.String(description='Name of the menu item.')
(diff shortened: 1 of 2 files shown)


saleor/graphql/menu/mutations.py
```diff
@@ -42,6 +52,44 @@ class Meta:
...
+                if category:
+                    category = cls.get_node_or_error(
+                        info, category, errors, 'items', Category)
+                    item['category'] = category
+                if collection:
+                    collection = cls.get_node_or_error(
+                        info, collection, errors, 'items',
+                        only_type=Collection)
+                    item['collection'] = collection
+                if page:
```
NyanKiyoshi: Unless I am missing something, we should say `elif` for those two last blocks as there is only zero to one true condition.

saleor/graphql/menu/mutations.py
```diff
@@ -42,6 +52,45 @@ class Meta:
...
+            collection = item.get('collection')
+            page = item.get('page')
+            url = item.get('url')
+            if len([i for i in [category, collection, page, url] if i]) > 1:
+                cls.add_error(
+                    errors, 'items', 'More than one item provided.')
+            else:
+                if category:
+                    category = cls.get_node_or_error(
+                        info, category, errors, 'items', Category)
```
akjanik: Small nit here, I'd like to be consistent and use `only_type` keyword as in calls below.

Issue #2744, opened 2018-08-30
Turn Order.paymentStatus field into an enum


patrys:
Currently `Order.status` is an `OrderStatus` enum but `Order.paymentStatus` is a `String`.

We should make both enums so clients can know all possible values up-front.

Issue #2154, opened 2018-05-08
Order events should contain type and parameters fields


dominik-zeglen:
At this moment order events consist with `content` and `user` fields, which are just fine to just display them, but in upcoming dashboard's version, we should expand this section a little, making the distinction between different types of events. Therefore I would like to propose introducing additional fields `type` and `parameters`, which would precisely describe what happened in the exact event, allowing us to display additional actions, information or just change appearance.

Also, please do not save translated messages, as they are currently being.

patrys:
Blocks part of #2639

maarcingebala:
Related issue: #2617

patrys:
Closed by #2811

PR #2754, opened 2018-08-30
[WIP] Implement order list and parts of order details

### LINKED ISSUE #2639: Implement order section in dashboard
Connect order section components to API and add them to the dashboard side menu, to make orders manageable.

patrys:
⚠️ Work in progress!

Fixes #2639 

Re #2587 

### TODO
* [x] phone numbers returned as single string while input expects prefix + number #2765
* [x] Order type does not have subtotal #2764
* [x] Order type does not have total authorized #2763
* [x] ~~Order type does not have total refunded (and neither does the Payment model)~~
* [x] order lines do not provide a mechanism to ask for a product thumbnail #2858
* [x] payment statuses are not an enum #2744
* [x] there are no order event types #2154 
* [x] there is a function called `Ø` that could probably use a better name
* [x] order summary uses fixed-height boxes with all labels grouped in one div and all values in another (this breaks when we need to show additional rows and is terrible for a11y) #2861
* [x] shipping method dialog is named "Edit customer details"
* [x] ~~it's impossible to select anything from the list of shipping methods without typing first~~
* [x] shipping method selection dialog does not indicate the previous value
* [x] there's no way I found to edit either of the addresses #2862
* [x] order cancellation can be repeated even if order was already cancelled #2808
* [x] there's no UI to select whether stock should be restored on cancellation #2868
* [x] `onCustomerEmailClick` sounds like it should really just be an `<a href="mailto:...">` #2869
* [x] `fulfillmentCreate(input: { order: ID!, ... })` should be `orderCreateFulfillment(order: ID!, input: ...)` and should return `order` #2809
* [x] there's no "success" confirmation of any of the operations, be it a visual cue or an explicit toaster message 
* [x] `orderAddNote` should take `order: ID!` an argument and should return `order` #2810
* [x] `Form` > `OrderFulfillmentDialog` should become `OrderFulfillmentDialog` > `OrderFulfillmentForm` > `Form`
* most mutations are not connected:
    * [x] cancel order
    * [x] customer edit - UI is not connected to a callback #2870
    * [x] add product #2854
    * change shipping method
        * [x] UI is not connected to a callback #2867
        * [x] mutation missing #2807 
    * [x] fulfillment - UI is not connected to a callback
    * [x] capture - UI is not connected to a callback
    * [x] refund - UI is not connected to a callback
    * [x] void
    * [x] add order note - UI is not connected to a callback #2866

### Pull Request Checklist


dominik-zeglen:
Very nice work with `Ø` function! But I feel like it should be introduced in separate PR, since this is project-wide change.

maarcingebala:
@dominik-zeglen Eventually this function will be used across the whole project, but extracting a separate PR now would only add extra work and dependencies between PRs which slows us down. I think we can as well introduce it incrementally.

dominik-zeglen:
Blocked by #2849

dominik-zeglen:
Blocked by #2854

PR #2781, opened 2018-09-04
Allow to query variants by list of ids

### LINKED ISSUE #2741: Add productVariants query
I would like to get more than one variant from API by ids, right now we support only the `productVariant(id: "xyz"}` query.

I need this feature for the cart feature in the client app. A query which I am thinking of should look like:
```graphql
  productVariants(ids: ["UHJvZHVjdFZhcmlhbnQ6Mg==", "JHDaFZhcmlhbnQ6Mg=="]) {
    id,
    stockQuantity,
    costPrice {
      currency,
      amount
    },
    product {
      id,
      name,
      thumbnailUrl
    }
  }
}
```

artursmet:
I want to merge this change because it closes #2741 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/resolvers.py
+++ b/saleor/graphql/product/resolvers.py
@@ -3,8 +3,8 @@ from django.db.models import Q
 
 from ...product import models
 from ...product.utils import products_with_details
-from ..utils import filter_by_query_param
-from .types import Category
+from ..utils import filter_by_query_param, get_nodes
+from .types import Category, ProductVariant
 
 PRODUCT_SEARCH_FIELDS = ('name', 'description', 'category__name')
 CATEGORY_SEARCH_FIELDS = ('name', 'slug', 'description', 'parent__name')
@@ -65,3 +65,7 @@ def resolve_products(info, category_id, query):
 
 def resolve_product_types():
     return models.ProductType.objects.all().distinct()
+
+
+def resolve_product_variants(info, ids):
+    return get_nodes(ids, graphene_type=ProductVariant)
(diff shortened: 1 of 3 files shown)


saleor/graphql/product/resolvers.py
```diff
@@ -65,3 +65,7 @@ def resolve_products(info, category_id, query):
 
 def resolve_product_types():
     return models.ProductType.objects.all().distinct()
+
+
+def resolve_product_variants(info, ids):
```
maarcingebala: To unify this endpoint the rest of connection-endpoints in our API I would make the `ids` argument optional. If it's not provided the endpoint would behave just as all other paginated connections. But when providing the list of `ids` we would filter the results and only return the requested IDs.

mociepka: That requires some extra permissions. A normal user won't be able to list all checkouts only its own.

maarcingebala: I assume you'll use a different query for checkouts, this one only returns variants. But it is true that similarly to how we filter products in the `products` query, here we should make sure that we don't expose variants of unpublished products publicly.

mociepka: Sorry for the misunderstanding, mu ming was attached to checkout API.

maarcingebala: Wat? 😄

maarcingebala: image
(credits: @patrys)

artursmet:
@maarcingebala I refactored the endpoint to be more flexible as you suggested. Please review again.

PR #2782, opened 2018-09-04
Return only static URLs for menu items

### LINKED ISSUE #2767: MenuItem's url should be null if not static
If MenuItem points to an object (for example collection), its `url` prop still returns url, which is unnecessary and introduces additional steps in creating form. We should avoid returning urls from API as client's app can have different url structure and make this field useless if generated.

1. Execute query 
```
{
  menuItems {
    edges {
      node {
        url
        category {
          id
        }
      }
    }
  }
}
```

### What I expected to happen
If url is not static, this field should be `null`

artursmet:
I want to merge this change because it closes #2767 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/menu/types.py
+++ b/saleor/graphql/menu/types.py
@@ -31,4 +31,4 @@ class MenuItem(CountableDjangoObjectType):
         model = models.MenuItem
 
     def resolve_url(self, info):
-        return self.get_url()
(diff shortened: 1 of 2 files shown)


maarcingebala:
True, `resolve_url` resolver becomes redundant now, that we return just the `url` field.

PR #2785, opened 2018-09-04
Add separate input for ProductVariantCreate and ProductVariantUpdate

### LINKED ISSUE #2674: It's possible to reparent a product variant using the GraphQL mutation
Currently `productVariantCreate` and `productVariantUpdate` share the same input type.

This causes `productVariantCreate` to list `product` (an ID field) as optional (which is invalid).

It also causes `productVariantUpdate` to accept a new value for `product` which should be illegal as it would move a variant between two products.

oldPadavan:

closes #2674 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/mutations/products.py
+++ b/saleor/graphql/product/mutations/products.py
@@ -294,16 +294,13 @@ class ProductDelete(ModelDeleteMutation):
         return user.has_perm('product.manage_products')
 
 
-class ProductVariantInput(graphene.InputObjectType):
+class ProductVariantPropertiesInput(graphene.InputObjectType):
     attributes = graphene.List(
         AttributeValueInput,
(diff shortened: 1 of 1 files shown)


saleor/graphql/product/mutations/products.py
```diff
@@ -294,16 +294,13 @@ def user_is_allowed(cls, user, input):
         return user.has_perm('product.manage_products')
 
 
-class ProductVariantInput(graphene.InputObjectType):
+class ProductVariantPropertiesInput(graphene.InputObjectType):
```
maarcingebala: I think the word "properties" may be misinterpreted as relating to product attributes. I would keep the name ProductVariantInput" here. The other name, "ProductVariantCreateInput" is fine.

PR #2786, opened 2018-09-04
Separate input for ProductCreate and ProductUpdate

### LINKED ISSUE #2677: It's possible to change the product's type using the GraphQL mutation
Currently `productCreate` and `productUpdate` share the same input type.

This causes `productCreate` to list `productType` (an ID field) as optional (not providing it is invalid).

It also causes `productUpdate` to accept a new value for `productType` which should be illegal as it could turn a book into a pair of jeans (and make all product and variant attributes invalid).

oldPadavan:

closes #2677 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/mutations/products.py
+++ b/saleor/graphql/product/mutations/products.py
@@ -196,7 +196,7 @@ class AttributeValueInput(InputObjectType):
         required=True, description='Value of an attribute.')
 
 
-class ProductInput(graphene.InputObjectType):
+class ProductPropertiesInput(graphene.InputObjectType):
     attributes = graphene.List(
         AttributeValueInput,
         description='List of product attributes.')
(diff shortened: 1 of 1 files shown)


saleor/graphql/product/mutations/products.py
```diff
@@ -196,7 +196,7 @@ class AttributeValueInput(InputObjectType):
         required=True, description='Value of an attribute.')
 
 
-class ProductInput(graphene.InputObjectType):
+class ProductPropertiesInput(graphene.InputObjectType):
```
maarcingebala: @oldPadavan I would do a similar change as in your other PR - let's change it here to `ProductInput` and have dedicated `ProductCreateInput` only for product creation.

maarcingebala:
Thanks for your contribution @oldPadavan!

PR #2796, opened 2018-09-06
Return absolute urls in media

### LINKED ISSUE #2718: Urls returned by GraphQL API should be absolute
Right now default media storage returns relative assets URLs. For instance:
```
/media/__sized__/products/saleor/static/placeholders/t-shirts/6-crop-c0-5__0-5-255x255-70.jpg
```
For API consumers will be much easier to work with relative URLs like:
```
https://example.com/media/__sized__/products/saleor/static/placeholders/t-shirts/6-crop-c0-5__0-5-255x255-70.jpg
```

My current solution fixes only product images and looks like:

```
def resolve_thumbnail_url(self, info, *, size=None):
        if not size:
            size = 255
        thumbnail_url = get_thumbnail(self.get_first_image(), size)
        is_absolute_url = urlparse(thumbnail_url).netloc
        if not is_absolute_url:
            return info.context.build_absolute_uri(thumbnail_url) 
        return thumbnail_url
```

But we need a solution which will be working for all media urls.

artursmet:
I want to merge this change because it closes #2718 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/types.py
+++ b/saleor/graphql/product/types.py
@@ -216,25 +216,6 @@ class ProductType(CountableDjangoObjectType):
             user=user).filter(product_type=self).distinct()
 
 
-class Collection(CountableDjangoObjectType):
-    products = DjangoFilterConnectionField(
-        Product, filterset_class=ProductFilterSet,
-        description='List of collection products.')
-
-    class Meta:
-        description = "Represents a collection of products."
-        exclude_fields = ['voucher_set', 'sale_set', 'menuitem_set']
-        filter_fields = {
-            'name': ['exact', 'icontains', 'istartswith']}
-        interfaces = [relay.Node]
-        model = models.Collection
-
-    def resolve_products(self, info, **kwargs):
-        user = info.context.user
-        return products_with_details(
-            user=user).filter(collections=self).distinct()
-
-
 class ImageWithThumbnail(graphene.ObjectType):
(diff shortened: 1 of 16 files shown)


saleor/product/templatetags/product_images.py
```diff
@@ -80,7 +80,7 @@ def get_thumbnail_size(size, method):
 
 
 @register.simple_tag()
-def get_thumbnail(instance, size, method='crop'):
+def get_thumbnail(instance, size, method='thumbnail'):
```
patrys: Could we either drop the method param entirely or make it explicitly required?

saleor/graphql/product/types.py
```diff
@@ -216,10 +216,26 @@ def resolve_products(self, info, **kwargs):
             user=user).filter(product_type=self).distinct()
 
 
+class ImageWithThumbnail(graphene.ObjectType):
```
patrys: Shouldn't products use this type as well? I think `Image` would be a better name, Shopify has a similar type and it contains the `originalUrl`, a `transformedUrl` (thumbnail), `id` and `altText`.

maarcingebala: +1 for `Image`

maarcingebala: Ideally this should be the common interface for all the images exposed by the API, including product images.

PR #2805, opened 2018-09-08
nodejs-legacy is no more available

hasher-hasher:
The [nodejs-legacy package](https://packages.debian.org/search?searchon=names&keywords=nodejs-legacy) is now installed through a node syslink with the distribution official repositories installation of nodejs on Debian and Ubuntu based systems.


patrys:
Thank you for your first pull request! ✨

PR #2811, opened 2018-09-10
Refactor OrderHistoryEntry events

### LINKED ISSUE #2617: Order events timeline
Extend the current timeline feature from (order history) to cover the whole order's lifecycle and meet with the requirements of new UI mockups for Dashboard 2.0.

### LINKED ISSUE #2810: Dashboard 2.0: `orderAddNote` mutation should return the modified order
Since `orderAddNote` operates on orders it should return the modified `order` aside from the newly created order note.

It should also take `order: ID!` as an argument separate from `input` to match other mutations and to allow for easy variable binding on the client side.

Pacu2:
close #2617 
close #2810 

Reworked OrderHistoryEntry to work with Enum instead of hardcoding the messages.

### Pull Request Checklist


saleor/graphql/order/types.py
```diff
@@ -1,11 +1,41 @@
...
+        return None
+
+    @staticmethod
+    def parse_literal(node):
+        return node
+
+
+class OrderEvent(CountableDjangoObjectType):
+    event = OrderEventsEnum(description='Order event type')
+    parameters = ParametersScalar(
```
patrys: I think we could do a better job here by just explicitly listing the fields we have along with their types. Otherwise we'll have to handle a custom scalar type in client code and maintain documentation out-of-channel (as the internal structure of a custom scalar cannot be expressed using the schema).

saleor/dashboard/order/views.py
```diff
@@ -614,14 +629,18 @@ def fulfill_order_lines(request, order_pk):
...
+                user=request.user,
+                type=OrderEvents.FULFILLMENT_FULFILLED_ITEMS.value)
             if form.cleaned_data.get('send_mail'):
                 send_fulfillment_confirmation.delay(order.pk, fulfillment.pk)
-                send_mail_msg = pgettext_lazy(
-                    'Dashboard message related to an order',
-                    'Shipping confirmation email was sent to user'
-                    '(%(email)s)') % {'email': order.get_user_current_email()}
-                order.history.create(content=send_mail_msg, user=request.user)
+                order.events.create(
```
NyanKiyoshi: What do you think about writing the event in the celery task? When it is done? As the mail can fail to be sent and raise an exception for example.

Pacu2: Initially, it will be possible to resend the email  directly from the event, in case of error event won't be added at all.
I think the solution might be adding a status to the event (like STARTED/ERROR/FINISHED) but for now I think it's out of scope

PR #2814, opened 2018-09-10
Expose one price for variants in the API

### LINKED ISSUE #2794: Return a variant price instead of the priceOverride
In my impression, `priceOverride` is an internal functionality which shouldn't be available in the API.
In addition, when I want to show a variant price I need to always introduce logic like:
```
let variantPrice = variants[0].priceOverride || product.price
```

I propose to change the attribute `priceOverride` to `price` which shows product price when the override is empty.

artursmet:
I want to merge this change because it closes #2794

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/types.py
+++ b/saleor/graphql/product/types.py
@@ -90,6 +90,7 @@ class ProductVariant(CountableDjangoObjectType):
         Money,
         description="""Override the base price of a product if necessary.
         A value of `null` indicates that the default product price is used.""")
+    price = graphene.Field(Money, description="Price of the product variant.")
     attributes = graphene.List(
         SelectedAttribute,
         description='List of attributes assigned to this variant.')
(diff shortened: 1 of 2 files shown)


saleor/graphql/product/types.py
```diff
@@ -114,6 +115,13 @@ def resolve_attributes(self, info):
     def resolve_margin(self, info):
         return get_margin_for_variant(self)
 
+    def resolve_price(self, info):
+        return (
+            self.price_override if self.price_override else self.product.price)
```
Pacu2: Might not be a case, but what is `self.price_override` is set to 0?

saleor/graphql/product/types.py
```diff
@@ -90,6 +90,7 @@ class ProductVariant(CountableDjangoObjectType):
         Money,
         description="""Override the base price of a product if necessary.
         A value of `null` indicates that the default product price is used.""")
+    price = graphene.Field(Money, description="Price of a product")
```
Pacu2: I think it should be "**the** product **variant.**"""

PR #2816, opened 2018-09-10
Add access control headers

### LINKED ISSUE #2815: Add access control headers to GraphQL API
I would like to be able to access API from a domain different that API in the browser

I will add proper headers in the `FileUploadGraphQLView` which is our main one.

mociepka:
Closes #2815

### DIFF
--- a/saleor/graphql/file_upload/views.py
+++ b/saleor/graphql/file_upload/views.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from graphene_django.views import GraphQLView
 import json
 import graphene
@@ -7,6 +8,18 @@ import graphene
 
 
 class FileUploadGraphQLView(GraphQLView):
+
+    def dispatch(self, request, *args, **kwargs):
+        # Handle options method the GraphQlView restricts it.
+        if request.method == 'OPTIONS':
+            response =  self.options(request, *args, **kwargs)
+        else:
+            response = super().dispatch(request, *args, **kwargs)
+        # Add access control headers
+        response['Access-Control-Allow-Origin'] = ','.join(settings.ALLOWED_HOSTS)
+        response['Access-Control-Allow-Methods'] = 'POST, OPTIONS'
+        response['Access-Control-Allow-Headers'] = 'Origin, Content-Type, Accept, Authorization'
+        return response
 
     @staticmethod
     def get_graphql_params(request, data):
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -372,7 +372,7 @@ bootstrap4 = {
 TEST_RUNNER = ''
 
 ALLOWED_HOSTS = get_list(
-    os.environ.get('ALLOWED_HOSTS', 'localhost,127.0.0.1'))
+    os.environ.get('ALLOWED_HOSTS', '*'))
 
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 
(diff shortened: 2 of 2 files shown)


saleor/settings.py
```diff
@@ -372,7 +372,7 @@ def get_host():
 TEST_RUNNER = ''
 
 ALLOWED_HOSTS = get_list(
-    os.environ.get('ALLOWED_HOSTS', 'localhost,127.0.0.1'))
+    os.environ.get('ALLOWED_HOSTS', '*'))
```
Pacu2: I'm not sure what are the downsides of having a wildcard as ALLOWED_HOSTS?

mociepka: This is only fall back for dev. Downsides on prod: https://docs.djangoproject.com/en/2.1/ref/settings/#allowed-hosts

patrys: Why do we need to change that?

saleor/graphql/file_upload/views.py
```diff
@@ -8,6 +9,18 @@
...
 class FileUploadGraphQLView(GraphQLView):
 
+    def dispatch(self, request, *args, **kwargs):
+        # Handle options method the GraphQlView restricts it.
+        if request.method.lower() == 'options':
+            response =  self.options(request, *args, **kwargs)
+        else:
+            response = super().dispatch(request, *args, **kwargs)
+        # Add access control headers 
+        response['Access-Control-Allow-Origin'] = ','.join(settings.ALLOWED_HOSTS)
```
patrys: Do we need to limit this to a specific list of hosts? Most APIs just return `Access-Control-Allow-Origin: *`.

NyanKiyoshi: I think it would still be better to allow the user to provide their own value if they want to only allow a few hosts. And default the value to `*`.

NyanKiyoshi: Shouldn't we check if the method is `OPTIONS` or `POST`?

saleor/graphql/file_upload/views.py
```diff
@@ -8,6 +9,18 @@
 
 class FileUploadGraphQLView(GraphQLView):
 
+    def dispatch(self, request, *args, **kwargs):
+        # Handle options method the GraphQlView restricts it.
+        if request.method.lower() == 'options':
```
NyanKiyoshi: Is there really a case where the HTTP method is not passed as uppercase?

saleor/settings.py
mociepka: @patrys if your API will be on the different domain that client browser will restrict requests ([CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS)).

This code sets proper headers for allowed host domains. So in the scenario:
* API on api.mystore.com,
* client on mystore.com,

you need to set `ALLOWED_HOSTS` to `["api.mystore.com", "mystore.com"]` and everything works.

In addition local development with API on `localhost:8000` and client `localhost:3000` is also covered.

saleor/graphql/file_upload/views.py
mociepka: It is more like my habit with keys from HTTP header. I will change that.

saleor/settings.py
patrys: Why would it be on a different domain during local development? I'm against putting `*` here as default as people will leave it as it in production mode which is a potential attack vector. CORS is in no way related to ALLOWED_HOSTS and in my opinion it should not be.

artursmet: I'd leave it as previous - it will force people to set it correctly in production without a way to leave it too open.

maarcingebala: I agree that we shouldn't set `*` here. I'll change that.

PR #2824, opened 2018-09-11
Drop jsonfield package for django's native JSONField

### LINKED ISSUE #2802: Use Django's JSONField instead of external package
Since v1.9 Django comes with builtin `JSONField`, we could start using it instead of an external `jsonfield` package.

Potential drawbacks:
- It's PostgreSQL-specific field
But we are supporting only PostgreSQL + HStoreField(postgres-specific as well) is already in use

Docs:
https://docs.djangoproject.com/en/2.1/ref/contrib/postgres/fields/#jsonfield

the-bionic:

resolve #2802
I could not completely remove the package from requirements.txt because django-prices depends on it.
### Pull Request Checklist


patrys:
I don't think you planned to include the `package-lock.json` change with a purely backend PR.

Pacu2:
Looks great, just one thing, we've just merged #2811, which introduced another `JSONField`on `OrderEvents` model, do you mind rebasing this PR against latest master and changing it in there too?

Also, I think that `jsonfield` should be dropped from the requirements as well.

the-bionic:
- Sure, not a problem. 

- As regards dropping it from the requirements, django-prices has a depen

Pacu2:
I think we might just drop it from requirements.in then, as it’s not required directly by Saleor

the-bionic:
Alright, on it.

saleor/order/models.py
```diff
@@ -329,8 +329,8 @@ class OrderEvent(models.Model):
     order = models.ForeignKey(
         Order, related_name='events', on_delete=models.CASCADE)
     parameters = JSONField(
-        blank=True, default={},
-        dump_kwargs={'cls': CustomJsonEncoder, 'separators': (',', ':')})
+        blank=True, default=dict,
+        encoder=CustomJsonEncoder)
```
Pacu2: I think it will all fit on line 331.

Also, I suppose we need to generate a new migration to reflect the `encoder` change on the `parameters` field, do we?

the-bionic:
Review has been factored in. Please take a look again.

Pacu2:
Looks good to me, thanks for contributing to Saleor! 

Will merge the PR tomorrow.

the-bionic:
Happy to help!

Pacu2:
There's a problem when migrating from the old instance with JSON data created (eg. checkout from the master, generate data via `populatedb`, pull this branch, run new migrations)
```
In [3]: Menu.objects.first().json_content
Out[3]: '"\\"[{\\\\\\"url\\\\\\": \\\\\\"/en/products/category/apparel-1/\\\\\\", \\\\\\"name\\\\\\": \\\\\\"Apparel\\\\\\", \\\\\\"translations\\\\\\": {}, \\\\\\"child_items\\\\\\": []}, {\\\\\\"url\\\\\\": \\\\\\"/en/products/category/accessories-2/\\\\\\", \\\\\\"name\\\\\\": \\\\\\"Accessories\\\\\\", \\\\\\"translations\\\\\\": {}, \\\\\\"child_items\\\\\\": []}, {\\\\\\"url\\\\\\": \\\\\\"/en/products/category/groceries-3/\\\\\\", \\\\\\"name\\\\\\": \\\\\\"Groceries\\\\\\", \\\\\\"translations\\\\\\": {}, \\\\\\"child_items\\\\\\": [{\\\\\\"url\\\\\\": \\\\\\"/en/products/category/coffees-4/\\\\\\", \\\\\\"name\\\\\\": \\\\\\"Coffees\\\\\\", \\\\\\"translations\\\\\\": {}, \\\\\\"child_items\\\\\\": []}, {\\\\\\"url\\\\\\": \\\\\\"/en/products/category/candies-5/\\\\\\", \\\\\\"name\\\\\\": \\\\\\"Candies\\\\\\", \\\\\\"translations\\\\\\": {}, \\\\\\"child_items\\\\\\": []}]}, {\\\\\\"url\\\\\\": \\\\\\"/en/products/category/books-6/\\\\\\", \\\\\\"name\\\\\\": \\\\\\"Books\\\\\\", \\\\\\"translations\\\\\\": {}, \\\\\\"child_items\\\\\\": []}]\\""'
```
The ugly quick fix I've tried 
```
In [13]: json.loads(json.loads(json.loads(Menu.objects.first().json_content)))
Out[13]:
[{'child_items': [],
  'name': 'Apparel',
  'translations': {},
  'url': '/en/products/category/apparel-1/'},
 {'child_items': [],
  'name': 'Accessories',
  'translations': {},
  'url': '/en/products/category/accessories-2/'},
 {'child_items': [{'child_items': [],
    'name': 'Coffees',
    'translations': {},
    'url': '/en/products/category/coffees-4/'},
   {'child_items': [],
    'name': 'Candies',
    'translations': {},
    'url': '/en/products/category/candies-5/'}],
  'name': 'Groceries',
  'translations': {},
  'url': '/en/products/category/groceries-3/'},
 {'child_items': [],
  'name': 'Books',
  'translations': {},
  'url': '/en/products/category/books-6/'}]
```
But have not investigated what are the differences between `jsonfield`'s `JSONField` and the one that Django offers.

the-bionic:
Alright. I will look into this.

the-bionic:
I used the interactive shell and tried the following commands:
>>> from saleor.menu.models import Menu
>>> val = Menu.objects.first().json_content
>>> print(val)

This is the result I got:

``` javascript
[{"url": "/en/products/category/apparel-1/", "name": "Apparel", "translations": {}, "child_items": []}, {"url": "/en/products/category/accessories-2/", "name": "Accessories", "translations": {}, "child_items": []}, {"url": "/en/products/category/groceries-3/", "name": "Groceries", "translations": {}, "child_items": [{"url": "/en/products/category/coffees-4/", "name": "Coffees", "translations": {}, "child_items": []}, {"url": "/en/products/category/candies-5/", "name": "Candies", "translations": {}, "child_items": []}]}, {"url": "/en/products/category/books-6/", "name": "Books", "translations": {}, "child_items": []}]
```

Pacu2:
Steps to reproduce for me:
1. Checkout to latest master branch
2. Fill some JSONField with data (eg. create Menus using `populatedb` command)
3. Download this branch, apply its migration
4. Open shell and check `json_content` of the Menu
It should not be a valid JSON (at least in my case)

Is that how you've reproduced it?

Pacu2:
@the-bionic 
Populatedb command lives in
https://github.com/mirumee/saleor/blob/master/saleor/core/management/commands/populatedb.py

the-bionic:
I have noticed some differences in how they both work:

```jsonfield's```  ```JSONField```` saves the value as ```text``` in the database while ```django's``` native ```JSONField``` saves its own as ```jsonb,``` which is native to postgres too. The difference is what is causing those inconsistencies.

Here is a snippet from  ```jsonfield's``` ```JSONField``` to buttress my point:

``` python
class JSONField(JSONFieldBase, models.TextField):
    """JSONField is a generic textfield that serializes/deserializes JSON objects"""
    form_class = JSONFormField

    def dumps_for_display(self, value):
        kwargs = {"indent": 2}
        kwargs.update(self.dump_kwargs)
        return json.dumps(value, ensure_ascii=False, **kwargs)
```

the-bionic:
> Steps to reproduce for me:
> 
>     1. Checkout to latest master branch
> 
>     2. Fill some JSONField with data (eg. create Menus using `populatedb` command)
> 
>     3. Download this branch, apply its migration
> 
>     4. Open shell and check `json_content` of the Menu
>        It should not be a valid JSON (at least in my case)
> 
> 
> Is that how you've reproduced it?

Yes. I did this.

the-bionic:
I think the data should not be populated as ```strings``` anymore, but as ```dictionaries```.
It will make things like this possible:

``` python
Dog.objects.create(name='Rufus', data={
     'breed': 'labrador',
     'owner': {
         'name': 'Bob',
         'other_pets': [{
             'name': 'Fishy',
         }],
     },
 })
Dog.objects.create(name='Meg', data={'breed': 'collie', 'owner': None})

Dog.objects.filter(data__breed='collie')
<QuerySet [<Dog: Meg>]>
````

This part of the codebase will be touched to return a dictionary:
https://github.com/mirumee/saleor/blob/34bc56a89caa3d65fc2dd852a5d1dc4bb7cec101/saleor/dashboard/menu/utils.py#L39

I also don't know how many more lines like there are.
What do you think?

**Comment by Pacu2** [comment]
@the-bionic I agree, we could rework this to work with dictionaries instead.
Searching for `json.dumps` would help us localize all the places (as far as I've briefly checked that's the only one)
Apart from that, we could add empty migrations to convert all string-based JSONFields to dictionaries via python scripts.

**Comment by the-bionic** [comment]
I am worried about something though, for the JSONFields to be changed from ```text``` -- the one ```jsonfield``` provides --- to ```jsonb``` (the one ```django``` provides), the migrations need to be done on a fresh database, because no migration files was generated when we changed the fields to  ```django's``` native version.

We should take a new approach to it, like the one in this [stackoverflow](https://stackoverflow.com/questions/41683385/how-should-i-upgrade-from-bradjaspers-django-jsonfield-to-djangos-built-in-jso) answer, then I migrate the data from string-base to dictionaries.

**Comment by the-bionic** [comment]
Updated PR. Should work as expected now.

File: saleor/menu/migrations/0008_menu_json_content_new.py [hunk 2]
```diff
@@ -0,0 +1,28 @@
...
+from django.db import migrations
+import json
+
+
+class Migration(migrations.Migration):
+
+    def populate_data(apps, schema_editor):
+        Menu = apps.get_model('menu', 'Menu')
+        for menu in Menu.objects.all():
+            if isinstance(menu.json_content, str):
```
Pacu2: @the-bionic I'm still having an issue with the `Menu` being a str. 
I wonder if we can just change from `if` to `while` in here, wouldn't that solve the problem?

Is it possible that we've stored a `json_content` as a `str`, inside of text-based JSONField? Wouldn't that make it the double nested string?

the-bionic: @Pacu2 Yes that did it. I did not realize it was a deeply nested string. The logic for resolving the nested string, though, is a little bit more than changing the ```if``` condition to a ```while``` loop since we transferring the content of a text-based JSONField to a jsonb-based JSONField, before dropping the text-based one.

PR updated again. Thanks.

Pacu2:
Works like a charm. Thank you very much for taking care of this issue.  The scope was bigger than planned, but you've handled it like a boss.

Congrats on your first contribution to Saleor!

the-bionic:
I'm humbled. Thanks for being patient with me. :)

PR #2826, opened 2018-09-11
Fix django 2.1 compatibility issues

### LINKED ISSUE #2616: Django 2.1 compatibility
We should switch our supported Django version to the following list:
* Django 1.11 (current LTS)
* Django 2.1 (latest stable)

Current blockers:
* [x] `graphene-django` depends on an old version of `django-filters` (https://github.com/graphql-python/graphene-django/pull/492)
* [ ] WeightInput passes floats to its base class which is a DecimalField
* [ ] Some form widgets pass `renderer` to functions that don't expect it

NyanKiyoshi:
This fixes the remaining compatibility issues.

May close #2616.

### Pull Request Checklist

### DIFF
--- a/saleor/core/weight.py
+++ b/saleor/core/weight.py
@@ -70,8 +70,8 @@ class WeightInput(forms.TextInput):
             return value.value
         return value
 
-    def render(self, name, value, attrs=None):
-        widget = super().render(name, value, attrs=attrs)
+    def render(self, name, value, attrs=None, renderer=None):
+        widget = super().render(name, value, attrs=attrs, renderer=renderer)
         unit = get_default_weight_unit()
         translated_unit = dict(WeightUnits.CHOICES)[unit]
         return render_to_string(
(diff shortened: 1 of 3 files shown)


saleor/core/weight.py
```diff
@@ -70,16 +70,16 @@ def format_value(self, value):
...
+        widget = super().render(name, value, attrs=attrs, renderer=renderer)
         unit = get_default_weight_unit()
         translated_unit = dict(WeightUnits.CHOICES)[unit]
         return render_to_string(
             self.template,
             {'widget': widget, 'value': value, 'unit': translated_unit})
 
 
-class WeightField(forms.DecimalField):
+class WeightField(forms.FloatField):
```
NyanKiyoshi: @Pacu2 could you take a look at this? To tell if we keep it that way or refactor the validation. I have an issue with `measurement.measures.Weight` converting the given double value to a float.
```python
    def to_python(self, value):
        value = super().to_python(value)
        if value is None:
            return value
        unit = get_default_weight_unit()
        return Weight(**{unit: value})
```

Because `DecimalField.validate` is expecting a double value, not a float.

Pacu2: I think we can keep it this way, it's more reasonable than working around the `Weight` instance (which already has its issues...)

Pacu2:
Thank you for this!

PR #2828, opened 2018-09-12
Update hstore field default value and add missing migrations for product model

jxltom:
This PR fixed the warrning when django startup as followings.

```
Performing system checks...

System check identified some issues:

WARNINGS:
product.Product.attributes: (postgres.E003) HStoreField default should be a callable instead of an instance so that it's not shared between all field instances.
	HINT: Use a callable instead, e.g., use `dict` instead of `{}`.
product.ProductVariant.attributes: (postgres.E003) HStoreField default should be a callable instead of an instance so that it's not shared between all field instances.
	HINT: Use a callable instead, e.g., use `dict` instead of `{}`.

System check identified 2 issues (0 silenced).
```

This PR also add the missing migrations for ```producttype``` in ```product``` model

### Pull Request Checklist


Pacu2:
Thanks for contributing!

PR #2831, opened 2018-09-12
Add orderUpdateShipping mutation

### LINKED ISSUE #2807: Dashboard 2.0: We need a mutation to change the shipping method of an order
I couldn't find a mutation that allows an order's shipping method to be changed. There is an `orderUpdate` mutation that looks like a candidate for this field but I'd like to see it split into separate mutations instead of being an all-in-one function that it currently is.

Pacu2:
close #2807 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/mutations/orders.py
+++ b/saleor/graphql/order/mutations/orders.py
@@ -161,8 +161,8 @@ class OrderUpdateShipping(BaseMutation):
             order.shipping_method_name = None
             order.save(
                 update_fields=[
-                    'shipping_method', 'shipping_price',
-                    'shipping_method_name'])
+                    'shipping_method', 'shipping_price_net',
+                    'shipping_price_gross', 'shipping_method_name'])
             return OrderUpdateShipping(order=order)
 
         method = cls.get_node_or_error(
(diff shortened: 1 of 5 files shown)


saleor/graphql/order/mutations/orders.py
```diff
@@ -62,6 +90,59 @@ class Meta:
...
+        if not input['shipping_method']:
+            if order.is_shipping_required():
+                cls.add_error(
+                    errors, 'shippingMethod',
+                    'Shipping method is required for this order.')
+                return OrderUpdateShipping(errors=errors)
+            order.shipping_method = None
+            order.shipping_price = ZERO_TAXED_MONEY
+            order.shipping_method_name = None
+            order.save()
```
maarcingebala: We could use `update_fields` here and in the `save` below.

saleor/graphql/order/mutations/orders.py
```diff
@@ -62,6 +90,59 @@ class Meta:
...
+        if errors:
+            return OrderUpdateShipping(errors=errors)
+
+        order.shipping_method = method
+        order.shipping_price = method.get_total_price(info.context.taxes)
+        order.shipping_method_name = method.name
+        order.save()
+        return OrderUpdateShipping(order=order)
+
+    order = graphene.Field(
```
maarcingebala: Class fields should appear at the top of the class declaration, above the metaclasses and methods. From what I see our codebase lacks consistency in this area so it may be a good idea to do the overall refactoring, but that's a separate PR. Let's fix at least this line in this PR.

PR #2835, opened 2018-09-12
Add OrderEvents docs

Pacu2:
close #2811


docs/architecture/events.rst
```diff
@@ -0,0 +1,44 @@
...
+    when certain actions are completed, such us creating the order,
+    cancelling fulfillment or completing a payment.
+
+Order events
+------------
+
++--------------------------+----------------------------------------------------------------+
+| Code                     | Description                                                    |
++--------------------------+----------------------------------------------------------------+
+| ``placed``               | An order was placed by the customer.                           |
```
patrys: Since we expose those as an enum in the API they end up being uppercase so tables like this can be a bit confusing. Should we add a column for values returned by the GraphQL API? Should we switch the backend to also use uppercase values (this would require a migration and might break existing scripts)?

Pacu2: For now, we might just add another column here to note down the GraphQL Enum value, and consider transforming all of the choices to uppercase.

docs/architecture/events.rst
```diff
@@ -0,0 +1,44 @@
+Events
+======
+
+.. note::
+    Events are autogenerated and will be created automatically
```
NyanKiyoshi: Maybe it would be fitting better that way since we are speaking about events:
> and will be **triggered** automatically

PR #2838, opened 2018-09-12
Remove Django 2.1 from allowed failures

Pacu2:
I think after resolving #2616 we can remove Django 2.1 from allowed failures on Travis CI


patrys:
We should also remove Django 2.0 from the test matrix. Our support policy is latest stable + last LTS.

PR #2841, opened 2018-09-13
Add checks if Order mutations can be performed

### LINKED ISSUE #2808: Dashboard 2.0: Order mutations don't check if an order is in the correct state
Mutations like order cancellation, payment release etc. can be called over and over on the same order ID without errors:

screenshot

Pacu2:
Some of them were missing

close #2808 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/mutations/orders.py
+++ b/saleor/graphql/order/mutations/orders.py
@@ -19,6 +19,41 @@ def try_payment_action(action, money, errors):
         errors.append(Error(field='payment', message=str(e)))
 
 
+def clean_order_cancel(order, errors):
+    if order and not order.can_cancel():
+        errors.append(
+            Error(
+                field='order',
+                message='This order can\'t be canceled.'))
+    return errors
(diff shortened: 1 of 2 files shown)


saleor/graphql/order/mutations/orders.py
```diff
@@ -19,6 +19,41 @@ def try_payment_action(action, money, errors):
...
+            Error(
+                field='payment',
+                message='There\'s no payment associated with the order.'))
+        return errors
+    if payment.status != PaymentStatus.PREAUTH:
+        errors.append(
+            Error(
+                field='payment',
+                message='Only pre-authorized payments can be captured'))
+    try_payment_action(payment.capture, amount, errors)
```
maarcingebala: I see some inconsistency here - rest of those `clean_*` functions only check if action preconditions are true and return errors respectively, while this one as well tries to perform an action. Is that on purpose? I'd expect only logical checks here.

PR #2848, opened 2018-09-14
Return OrderEventsEmail as Enum in GraphQL API

Pacu2:
I've missed it in #2811, OrderEventsEmail should be returned as an Enum


patrys:
Have you checked if it affects the schema? Does it need to be rebuilt?

maarcingebala:
@Pacu2 Please rebase this branch with `master`.

Issue #2291, opened 2018-06-19
Simplify permission management


maarcingebala:
We think that the current permission management flow may be overcomplicated in many cases. It might be enough if we only had general permissions, so e.g. instead of "edit products" and "view product" we could have one general "products" permission. 

There is also question whether we need permission groups at this moment. Maybe assigning simple general permissions directly to user accounts would be enough.

I'd like to hear some input on this before we decide what approach to take.

PR #2850, opened 2018-09-14
Remove Group permissions from API

Pacu2:
Removed leftovers from Group permissions as they're not used

related #2291


patrys:
@Pacu2 did you make sure that Graphene does not automatically define this field for us? It's not in the exclude list.

Pacu2:
@patrys You're right, it's automatically defined, will exclude it in another PR.

PR #2851, opened 2018-09-14
Handle user_email resolver with an empty string

### LINKED ISSUE #2849: Creating empty order draft causes API to explode
To get order draft details from API.

Execute this query
```
{
  orders {
    edges {
      node {
        id
        userEmail
      }
    }
  }
}
```

### What I expected to happen
Definitely not to throw an error.

Pacu2:
As the field was nonnullable, we could not return a null value for it in the API 

close #2849 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/types.py
+++ b/saleor/graphql/order/types.py
@@ -98,6 +98,8 @@ class Order(CountableDjangoObjectType):
     events = graphene.List(
         OrderEvent,
         description='List of events associated with the order.')
+    user_email = graphene.String(
+        required=False, description='Email address of the customer.')
 
     class Meta:
         description = 'Represents an order in the shop.'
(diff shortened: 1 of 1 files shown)


Pacu2:
@akjanik I've forgotten to include `null=True` on the model

patrys:
I think it's easier to redefine the field to not be required at API level.

Pacu2:
@patrys There's a problem with the resolvers, if user_email is equal to the empty string, I think we should return a null instead, which is not possible at this moment due to the field being not nullable.

patrys:
If we override the field to make it not required at GraphQL level then we can safely return null for empty strings.

Pacu2:
@patrys Right, got it

PR #2853, opened 2018-09-14
Add attribute value types in API

### LINKED ISSUE #2793: Add ability to store color and images in the variant attribute model
I would like to be able to render variant picker with color or image like adidas.com does:
image

Currently, an attribute has only one field: name. In the past, we have the attribute type model but that was over-engineered. 

I propose to introduce a new `value` field (CharField) in the attribute model (not required if empty filled by the slug). This field could be filled with a color (#aabbcc), an image url (https://a.b/image.svg) or a string. 

API consumer will decide what to do with those values (or just omit them). For my data picker, I could simple match regexps for url and color and decide how to render my widget.

In the future dashboard could be doing the same and provide a nice interface for color picker or images selector.

```graphql
query Product {
  product(id: "UHJvZHVjdDo0") {
    id
    variants {
      edges {
        node {
          id
          sku
          name
          attributes {
            attribute {
              id
              slug
              name
            }
            value {
              id
              name
              value
            }
          }
        }
      }
    }
  }
}

```

artursmet:
I want to merge this change because it closes #2793 

### Pull Request Checklist


patrys:
Please rebuild the schema so we know what the result looks like.

saleor/graphql/product/types.py
```diff
@@ -35,17 +44,33 @@ def resolve_attribute_list(attributes):
     return attributes_list
 
 
+def resolve_attribute_value_type(attribute_value):
+    color_pattern = '^(#[0-9a-fA-F]{3}|#(?:[0-9a-fA-F]{2}){2,4}|(rgb|hsl)a?\((-?\d+%?[,\s]+){2,3}\s*[\d\.]+%?\))$'  # noqa
+    color_pattern = re.compile(color_pattern)
```
NyanKiyoshi: Could you define it as a constant at the top of the file instead?

saleor/graphql/product/types.py
```diff
@@ -35,17 +46,32 @@ def resolve_attribute_list(attributes):
     return attributes_list
 
 
+def resolve_attribute_value_type(attribute_value):
```
Pacu2: Wouldn't that be more convenient to just define `type` on the `ProductAttributeValue`, instead of guessing what's the value?

saleor/graphql/product/types.py
```diff
@@ -35,17 +46,32 @@ def resolve_attribute_list(attributes):
...
+    color_pattern = re.compile(COLOR_PATTERN)
```
NyanKiyoshi: You still need to compile the regex at the top of the file, then you will use the compiled regex as is. Otherwise it doesn't make sense to use `re.compile`.

saleor/graphql/product/types.py
```diff
@@ -35,17 +46,32 @@ def resolve_attribute_list(attributes):
...
+    if color_pattern.match(attribute_value):
+        return ProductAttribuetValueType.COLOR
+    if 'gradient(' in attribute_value:
+        return ProductAttribuetValueType.GRADIENT
+    if '://' in attribute_value:
+        return ProductAttribuetValueType.URL
+    return ProductAttribuetValueType.STRING
```
NyanKiyoshi: Typo: 
> ProductAttrib**ue**tValueType

saleor/graphql/product/types.py
```diff
@@ -16,6 +18,15 @@
 from ..utils import get_database_id
 from .filters import ProductFilterSet
 
+COLOR_PATTERN = r'^(#[0-9a-fA-F]{3}|#(?:[0-9a-fA-F]{2}){2,4}|(rgb|hsl)a?\((-?\d+%?[,\s]+){2,3}\s*[\d\.]+%?\))$'  # noqa
+
+
+class ProductAttribuetValueType(graphene.Enum):
```
NyanKiyoshi: Typo: 
> ProductAttrib**ue**tValueType

PR #2894, opened 2018-09-18
Update shipping zones GraphQL interface

### LINKED ISSUE #2792: Shipping zones in GraphQL have a very non-GraphQL interface
Current signature is:

```graphql
{
  shippingZones(
    before: String
    after: String
    first: Int
    last: Int
    name_Icontains: String
    countries_Icontains: String
    shippingMethods_Price_Gte: Float
    shippingMethods_Price_Lte: Float
  ) {
    edges {
      node {
        id: ID!
        name: String!
        countries: [String]
        shippingMethods {
          edges {
            node {
              id: ID!
              name: String!
              type: ShippingMethodType!
              price: Money
              minimumOrderPrice: Money
              maximumOrderPrice: Money
              minimumOrderWeight: Money
              maximumOrderWeight: Money
            }
          }
        }
        priceRange: MoneyRange
      }
    }
  }
}
```

I don't like the fact that we have two levels of pagination here. 

All `Money` and `MoneyRange` fields are optional, I don't think that's intended.

It would be nice to document whether we intend clients do display both zone and method names or just the method names.

The input fields generated by `django-filters` are ugly and we should replace them but I think the ones supported are not very useful anyway.

It would be super useful to be able to just say `shippingMethods(forOrder: "123")` instead of having to reimplement filtering by countries, price ranges, and weight in the client.

Another approach is to do what Shopify is doing and provide one API for fetching the list of zones and methods and another to provide just the ones that make sense in given context:

```graphql
{
  checkout(id: "123") {
    availableShippingRates {
      id
      name
      price
    }
  }
  order(id: "abc") {
    availableShippingRates {
      id
      name
      price
    }
  }
}
```

Pacu2:
close #2792 

### DIFF
--- a/saleor/graphql/shipping/types.py
+++ b/saleor/graphql/shipping/types.py
@@ -17,8 +17,7 @@ ShippingMethodTypeEnum = graphene.Enum(
 
 
 class ShippingMethod(DjangoObjectType):
-    type = ShippingMethodTypeEnum(
-        required=True, description='Type of the shipping method.')
+    type = ShippingMethodTypeEnum(description='Type of the shipping method.')
 
     class Meta:
         description = """
--- a/saleor/graphql/shipping/types.py
+++ b/saleor/graphql/shipping/types.py
@@ -62,6 +62,7 @@ class ShippingZone(CountableDjangoObjectType):
     def resolve_shipping_methods(self, info):
         return self.shipping_methods.all()
 
+
 class WeightScalar(Scalar):
 
     @staticmethod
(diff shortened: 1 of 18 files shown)


Pacu2:
React part needs to be updated by the React hero @dominik-zeglen

saleor/graphql/shipping/types.py
```diff
@@ -11,16 +11,41 @@
...
+        ShippingMethod,
+        description=(
+            'List of shipping methods available for orders'
+            ' shipped to countries within this shipping zone.'))
 
     class Meta:
-        description = 'Represents a shipping zone in the shop.'
+        description = """
+            Represents a shipping zone in the shop. Zones are the concept
+            used only for groupping shipping methods in the dashboard,
```
maarcingebala: Typo: groupping -> grouping

PR #2899, opened 2018-09-18
Improve authentication loading state

### LINKED ISSUE #2555: Improve user authentication state
Currently, if a user is unauthenticated, but has token saved in local storage, we display the login form regardless of the fact that there is a possibility of skipping this form. In my opinion, in that particular case, we should display a spinner (or another loading widget) to let the user know that they are being logged in, rather than showing form for a split second.

dominik-zeglen:
Fixes #2555


dominik-zeglen:
@pwgryglak we need a design for spinner or something to cover state where a user has saved credentials, but server is verifying it.
Scenario: logged user hits refresh.

PR #2901, opened 2018-09-18
Fulfillment mutations improvements

### LINKED ISSUE #2809: Dashboard 2.0: Replace `fulfillmentCreate` mutation with `orderCreateFulfillment`
Creating an order fulfillment is an operation on the order object so it should follow the naming scheme of other mutations. It should return `order` as well as the newly created `fulfillment`.

It should also take `order: ID!` as a named argument outside of `input` like other mutations do. This allows the client to bind the `order` variable separately from the user input:

```jsx
<OrderFulfillmentCreate variables={{ order: order.id }}>
  {createFulfillment => (
    // ...
  )}
</OrderFulfillmentCreate>
```

### LINKED ISSUE #2843: Refactor `fulfillmentUpdate` mutation
`fulfillmentUpdate` mutation should be renamed to `orderUpdateFulfillmentTrackingInfo` to clarify what exactly would be updated since fulfillment lines are immutable. Its signature should also be changed to 
```graphql
orderUpdateFulfillmentTrackingInfo(id: ID!, input: { trackingNumber: String! }, notifyUser: Boolean)
```
so we won't mix object properties and action options in one input.

Pacu2:
close #2809 
close #2843 

### Pull Request Checklist


saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -130,13 +135,19 @@ def save(cls, info, instance, cleaned_input):
                     'email_type': OrderEventsEmails.FULFILLMENT.value},
                 type=OrderEvents.EMAIL_SENT.value,
                 user=info.context.user)
+        return FulfillmentCreate(fulfillment=instance, order=order)
```
maarcingebala: I'm not sure this would work as the result of `save` method is not used anywhere in `ModelMutation`. To override the mutation response you could override `success_response`.

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -56,96 +48,140 @@ class FulfillmentCancelInput(graphene.InputObjectType):
...
-            send_fulfillment_confirmation.delay(order.pk, instance.pk)
-            order.events.create(
-                parameters={
-                    'email': order.get_user_current_email(),
-                    'email_type': OrderEventsEmails.FULFILLMENT.value},
-                type=OrderEvents.EMAIL_SENT.value,
-                user=info.context.user)
+            send_fulfillment_confirmation_to_customer(
+                order, fulfillment, user)
+        return FulfillmentCreate(fulfillment=fulfillment, order=order)
```
maarcingebala: Is the result of this `return` used anywhere?

Pacu2:
React part needs to be updated by the true and only ReactJS hero @dominik-zeglen

dominik-zeglen:
Creating fulfillment group without tracking number returns this:
```
{"errors":[{"message":"'tracking_number'","locations":[{"line":2,"column":3}],"path":["orderFulfillmentCreate"]}],"data":{"orderFulfillmentCreate":null}}
```

PR #2907, opened 2018-09-19
Remove label of the second address in checkout page

jxltom:
It looks like https://github.com/mirumee/saleor/pull/2569 adds additional padding top for second adddress but forgets to remove second address's label which makes the alignment not horizontal, such as in https://demo.getsaleor.com/zh-tw/checkout/shipping-address/.

Here is a screenshot.

screenshot from 2018-09-19 15-14-03

### Screenshots

Here is the screenshopt after this PR fixes this issue.

image

### Pull Request Checklist

### DIFF
--- a/saleor/account/i18n.py
+++ b/saleor/account/i18n.py
@@ -98,8 +98,7 @@ class AddressForm(forms.ModelForm):
                 'Company or organization', 'Company or organization'),
             'street_address_1': pgettext_lazy(
                 'Address', 'Address'),
-            'street_address_2': pgettext_lazy(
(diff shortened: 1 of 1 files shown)


Pacu2:
We should just change https://github.com/mirumee/saleor/blob/master/saleor/account/i18n.py#L101
to 
```
...
            'street_address_2': '',
...
```

jxltom:
You can right, we should just remove the label. Now this PR is updated.

Pacu2:
Thanks for catching this!

PR #2910, opened 2018-09-19
Add mutations to manage order lines in draft orders

### LINKED ISSUE #2854: Add mutation for order lines in draft order
We should be able to add, update and remove order lines without sending the whole list to a server. Currently, it is not possible as `draftOrderUpdate` takes only `DraftOrderInput!`, which then takes `lines: [OrderLineInput]`, replacing the whole list rather than doing any operations on it.

maarcingebala:
Fixes #2854

### Pull Request Checklist


saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -225,3 +206,118 @@ def mutate(cls, root, info, id):
             type=OrderEvents.PLACED_FROM_DRAFT.value,
             user=info.context.user)
         return DraftOrderComplete(order=order)
+
+
+class DraftOrderLineAdd(BaseMutation):
```
Pacu2: I wonder if this kind of mutations cannot be used for both draft and non-draft orders?

Pacu2: Also, I think we are using `Create` instead of `Add` usually

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -225,3 +206,118 @@ def mutate(cls, root, info, id):
...
+            return DraftOrderLineAdd(errors=errors)
+
+        if order.status != OrderStatus.DRAFT:
+            cls.add_error(
+                errors, 'order_id', 'Only draft orders can be edited.')
+
+        line = None
+        if not errors:
+            line = add_variant_to_order(
+                order, variant, input['quantity'], allow_overselling=True)
```
Pacu2: I think we should check if quantity is larger than 0

saleor/graphql/order/mutations/draft_orders.py
maarcingebala: Adding order lines is valid only for draft orders. On branch `checkout-2.0` there is a similar mutation for adding lines to checkout, but I'd keep them separated.

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -225,3 +206,118 @@ def mutate(cls, root, info, id):
...
+                deallocate_stock(variant, line.quantity)
+            db_id = line.id
+            line.delete()
+            line.id = db_id
+            recalculate_order(order)
+        return DraftOrderLineDelete(
+            errors=errors, order=order, order_line=line)
+
+
+class DraftOrderLineUpdate(ModelMutation):
```
Pacu2: I wonder if we shouldn't delete the line, if it's quantity is set to 0?

saleor/graphql/order/mutations/draft_orders.py
maarcingebala: True. I'll add a manual check but it would be very helpful if we were able to use Django validators for that purpose. We'll have to think about something like that in our `BaseMutation`.

Pacu2: Related: https://github.com/mirumee/saleor/issues/2912
At this moment 0 is allowed value on the quantity field in the models.

saleor/graphql/order/mutations/draft_orders.py
maarcingebala: I decided to hold on with implementing this. I think that in terms of API it might be unexpected behavior since there is a dedicated mutation to remove order lines. I'd keep it as it is and see how it works in our frontend.

PR #2925, opened 2018-09-20
Adjust minimum validators on CartLine, OrderLine and FulfillmentLine

### LINKED ISSUE #2912: It should not be possible to create a line with 0 quantity
At this moment one can create `CartLine` / `OrderLine` / `FulfillmentLine` with 0 `quantity`
We should adjust a validator on the models to prevent such actions, as it doesn't make sense to have an item with no quantity assigned.

the-bionic:
close #2912

### Pull Request Checklist


Pacu2:
I think you need to include migrations for those changes.

the-bionic:
Yeah, that's right.

the-bionic:
I'll do that right away.

the-bionic:
How about dropping the ```MaxValueValidator``` on ```quantity_fulfilled``` as well?

Pacu2:
Thank you for picking this up!

the-bionic:
Glad to be of assistance.

PR #2933, opened 2018-09-21
Fixes applicable shipping method is not filtered by the discounts

jxltom:
If there is a sale on a product, a price based shipping method is filtered by product's original price instead of its discount price. So it may be shown in checkout's shipping method view but can not proceed to checkout's summary view since the latter view will take discounts into account when validating shipping method.

A price based shipping method should be filtered according to order's subtotal after discount. This PR fixes this issue.

### Pull Request Checklist


Pacu2:
Good catch

PR #2946, opened 2018-09-22
Flexible manage settings, add critical warnings on main page, add override any template or static files

mmmcorpsvit:
1. Add warning messages when Debug=True and SECRET_KEY is default (you can define list of default values in settings.IS_BAD_SECRET_KEY)
2. Add feature for easy define LANGUAGES const, use SALEOR_LANGUAGES comma separated env
3. For work with env use [Django-environ](https://github.com/joke2k/django-environ), this package already contain in django-graphql-jwt package
4. Add simple template tag for extract any settings value **saleor\core\templatetags\attributes.py** function settings_value. Simple:
in any template
```html
{% load attributes %}

      {% settings_value "IS_BAD_SECRET_KEY" as bad_key %}
      {% if bad_key %}
        <div class="bg-warning">
          Warning, <a href="https://saleor.readthedocs.io/en/latest/gettingstarted/configuration.html">
          SECRET_KEY is default!</a>
         </div>

```
5. Add feature for override any template file, just create folder **templates_local** and put override file. Also you can override any static file, put to **static_local**. Need you help to deep test. I don't work with webpack and another JS-hell)
6. And main. Use can now define you os.env in declarative way from **.env** file. Example:
```python
# simple define variables

DEBUG=True
SECRET_KEY=your-secret-key

DJANGO_ADMINS=Full Name <email-with-name@example.com>,anotheremailwithoutname@example.com
INTERNAL_IPS=127.0.0.1,127.0.0.2

DEBUG_TOOLBAR=True

REDIS_URL=rediscache://127.0.0.1:6379/1

DATABASE_URL=pgsql://saleor:saleor@127.0.0.1:5432/saleor

TIME_ZONE=Europe/Kiev
SALEOR_LANGUAGES=en,uk,ru

EMAIL_URL=smtp://user@:password@localhost:25

CELERY_TIMEZONE=UTC
CELERY_BROKER=amqp://
```
How it work? On load settings.py parse .env file (if file not present just show warning) and set variables to system os.env. Now we can simple use old way load (standard os.env) or more progressive new way presented there (with type cast, default values etc...). 
7. Also you can use local_settings.py file to override or extend settings. This code do that:
``` python
try:
    from settings_local import *
except ImportError:
    pass
```
8. This all what need for minimize change settings code of original Saleor and easy load Saleor from GitHub and start work, because all settings now contain default values (critical values: Debug and SECRET_KEY now show warning on main page).

### Screenshots

Warning messages in main page (ugly, but this good, because you see this messages)
b4423f8a68

Sample .env file, this simple format support begin from "#"-comment, to support $-aliases. This form file also wide use in Docker and have many bash samples to use (load, execute...). Also suport syntax highlight in top IDE PyCharm. So now you can very quickly set/delete/edit any os variables just edit file.
bc3893ff91

Easy Override any template or static files:
cd5a2f40b6

P.S. I'm 2 days test and fix all errors and bugs, but can be new, need good test. If you dont wont use .env+localsettings.py you can use all way - os.env, must all work fine. Also i investigate way for hook and replace any file load without change Saleor core files. 
I do this, because must be simple way to update Saleor core in already cloned file with minimal conflicts.

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -87,9 +87,9 @@ LANGUAGES = [
     ('zh-hans', _('Chinese')),
     ('zh-tw', _('Chinese (Taiwan)'))]
 
-_tmp = env('SALEOR_LANGUAGES')
-if _tmp:
-    LANGUAGES = [i for i in LANGUAGES if i[0] in _tmp]
+_tmp_saleor_languages = env('SALEOR_LANGUAGES')
+if _tmp_saleor_languages:
+    LANGUAGES = [i for i in LANGUAGES if i[0] in _tmp_saleor_languages]
 
 
 LOCALE_PATHS = [os.path.join(PROJECT_ROOT, 'locale')]


mmmcorpsvit:
I try many times fix Code climate errors, but this impossible, need you help please. 
This is my 3 PR, i really don't know how fix. Ignore error, try fix error from some conf files? Can any explain please.

P.R. Sorry my poor English.

saleor/settings.py
```diff
@@ -542,3 +566,10 @@ def get_host():
 
 SERIALIZATION_MODULES = {
     'json': 'saleor.core.utils.json_serializer'}
+
+
+# you can override settings also in local_settings.py
+try:
+    from settings_local import *  # noqa
```
Pacu2: You should use `DJANGO_SETTINGS_MODULE` and point it to your `settings_local`, instead of importing `settings_local` to the normal `settings` file
https://docs.djangoproject.com/en/2.1/topics/settings/#envvar-DJANGO_SETTINGS_MODULE

saleor/settings.py
```diff
@@ -81,16 +86,23 @@ def get_bool_from_env(name, default_value):
     ('vi', _('Vietnamese')),
     ('zh-hans', _('Chinese')),
     ('zh-tw', _('Chinese (Taiwan)'))]
+
+_tmp = env('SALEOR_LANGUAGES')
+if _tmp:
```
Pacu2: Please use meaningful variables

.gitignore
```diff
@@ -18,8 +18,9 @@
 *.log
 *.pot
 *.pyc
-local_settings.py
+settings_local.py
```
Pacu2: What is this change for?

mmmcorpsvit: This change do navigate in project tree more simple, because usualy use alphabetical sort file order. Same way too use for tenplates_local, static_local

saleor/settings.py
mmmcorpsvit: This not same. Django documentation say what you can define Django settings file, but don't say how apply settings on already forked project with minimal changing. I try fix this situation.
Lets see:
settings.py - big, around 550 lines! Copy and full edit - bad way. My way - allow don't do any changing in original settings.py file, just define variables from os.env or if you wish from .env file. In any way no conflicts. If you still need change  some settings you can too use settings_local.py file (example for define not trivial - Logging settings).

I don't say what my solution perfect, but Saleor don't have any solution for small pain update core code. Can explain how you company work with this problem?

saleor/settings.py
mmmcorpsvit: sorry my english, meaningful variables - you mean what _tmp variable must have more explained name? for example _tmp_saleor_languages?
i good understand? fixed

.gitignore
Pacu2: What with the developers that already have `local_settings.py` in their projects dirs?

mmmcorpsvit: Saleor have only one link to local_settings, - in .gitignore
we can use any name local_settings file, this don't have big weight

saleor/settings.py
mmmcorpsvit: 1. my solution give you choise: os.env and/or .env file, solution - full compatible, this not replacement! What edit more simple os variables or file? Where we can search any string on os env or in files??? etc... I think simple - use .env file, you already use it, see common.env))) Just try on clear machine clone Saleor with my PR and without and compare how much time you lost just for run Saleor with default configuration. I love just do clone to favorite IDE and simple run code, i don't wont do what must work from box. If this don't work from box - maybe this don't general work? I remember when first time clone Saleor - and i have simple questions: why i must set secret_key, i just wont run this, why developers dont set any default secret key? I just wont run, i select from oscar, odoo etc.... i need simple way to run!...What, how in 2018 year shop-framework dont have demo docker??? Next shock - what tab release is not release)))) I see all time questions like this on gitter-chat, almost every day.
2. why what you say don't present in docs in "how customize code"? Many peoples just see files and this use how docs. Have one good old rule - good software - don't need help docs. 
3. Saleor - great shop, but must be more fork-based framework friendly from box! I think...

patrys: @mmmcorpsvit Please follow my line of though for a second:

1. Create a file named `local_settings.py`

2. Have the contents say:

    ```python
    from saleor.settings import *

    TEMPLATES[0]['DIRS'].append('local_templates')
    ```

3. Launch Django with `--settings-module=local_settings`

Everything should work with none of the proposed changes applied.

As for `local_settings` vs. `settings_local`: is it better to have local settings closer to global settings and local templates closer to global templates or is it better to have local overrides close to each other (ie. have local templates next to local settings)?

PR #2949, opened 2018-09-23
Add attribute filters in subcategories

### LINKED ISSUE #2697: Attribute filters are not available in subcategories

I'm trying to filter products in subcategories by attributes of this products

1.  Create category and then create 2 or more subcategory
2.  Add product to last subcategory
3. Select category, than subcategory and try to filter products by attributes

### What I expected to happen

Attribute filters are not available in subcategories! Filters by attributes available only in last subcategory. For example, i have category "Phones" with subcategories by companies names -> phone model. If i select last subcategory "Iphone 8", i got all filters, but if i select subcategory "Apple", i got filter only by price range.

So, how to enable these filters? Can i modify some template or python code by myself, or you guys will do it some later?

### Screenshots
nofitlers
filters

pitero92:
This PR fixes a problem with availability of filters in subcategories

Now filters are avaible in subcategories
example

Fixes #2697


saleor/product/filters.py
```diff
@@ -88,10 +88,10 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def _get_product_attributes_lookup(self):
-        return Q(product_types__products__category=self.category)
+        return Q(product_types__products__category=self.get_descendants(include_self=True))
```
Pacu2: `self` is the `ProductCategoryFilter` instance, I think you've meant `self.category`?

pitero92: Yes, I'm gonna push improved commit in few minutes

maarcingebala:
@pitero92 Thanks for opening the PR! We care not only about the quality of our code but also about the quality of all the written part on Github 😉. It would be great if you could:
- Edit the issue description with an explanation of the problem it fixes and concerning the issue it resolves (if there is one). You should use [Github syntax for closing issues](https://help.github.com/articles/closing-issues-using-keywords/).
- Edit your commit messages to follow [these rules](https://github.com/erlang/otp/wiki/writing-good-commit-messages) of good commit messages (also applies to the PR title).

In case of any question feel free to ask. Thanks!

pitero92:
@maarcingebala Thank You for your feedback.

I've already corrected what you asked for.
It was my first PR, I hope that with next one will go better.

filters.py
```diff
@@ -0,0 +1,106 @@
+from collections import OrderedDict
```
maarcingebala: Is this whole file used? I think it was added accidentally.

saleor/product/filters.py
```diff
@@ -88,10 +88,10 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def _get_product_attributes_lookup(self):
-        return Q(product_types__products__category=self.category)
+        return Q(product_types__products__category=self.category.get_descendants(include_self=True))
```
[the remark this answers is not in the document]
Pacu2: Right, I'd also like to see a test for it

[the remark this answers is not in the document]
pitero92: @maarcingebala could You explain your 1st point? 
This line should look something like `return Q(product_types__products__category__in(categories))`?
Where `categories = self.category.get_descendants(include_self=True)`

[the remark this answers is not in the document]
Pacu2: @pitero92 Please refer to the Django's docs regarding the queryset methods: https://docs.djangoproject.com/en/2.1/ref/models/querysets/#in

Getting the category's descendants is correct.

filters.py
```diff
@@ -88,10 +88,11 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def _get_product_attributes_lookup(self):
-        return Q(product_types__products__category=self.category.get_descendants(include_self=True))
+        categories = self.category.get_descendants(include_self=True)
+        return Q(product_types__products__category__in=categories)
 
     def _get_variant_attributes_lookup(self):
-        return Q(product_variant_types__products__category=self.category.get_descendants(include_self=True))
+        return Q(product_variant_types__products__category__in=categories)
```
Pacu2: Categories will be undefined at this point

filters.py
```diff
@@ -83,12 +83,12 @@ def _get_attribute_choices(self, attribute):
 
 
 class ProductCategoryFilter(ProductFilter):
+    categories = self.category.get_descendants(include_self=True)
```
pitero92: @Pacu2  now is better?

Pacu2: I think you should take a different approach in resolving bugs.

I think that you could start the Saleor on your local instance and reproduce the situation listed in the issue to confirm that the bug appears in your local copy of Saleor.

Having a bug reproduced, you could make attempts to fix it, as in the more complex situation it might not be possible to fix it via the guessing game.

It's not better, as that would throw an error as `self` will be undefined at this point. You should define that `categories` in each function calling for an attributes lookup.

maarcingebala:
Closing due to inactivity.

maarcingebala:
Continued in #3535

PR #2951, opened 2018-09-24
Add defaultVariant argument to createProduct mutation

### LINKED ISSUE #2724: Mutation to create products doesn't create the default variant
1. In the new dashboard, create a product of a type that doesn't use variants.

### What I expected to happen
A product is created correctly, but it is missing the default variant, that should be created in that case as well.

oldPadavan:

closes #2724 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/mutations/products.py
+++ b/saleor/graphql/product/mutations/products.py
@@ -196,6 +196,24 @@ class AttributeValueInput(InputObjectType):
         required=True, description='Value of an attribute.')
 
 
+class ProductVariantInput(graphene.InputObjectType):
+    attributes = graphene.List(
+        AttributeValueInput, required=False,
+        description='List of attributes specific to this variant.')
+    cost_price = Decimal(description='Cost price of the variant.')
+    price_override = Decimal(
+        description='Special price of the particular variant.')
+    sku = graphene.String(description='Stock keeping unit.')
+    quantity = graphene.Int(
+        description='The total quantity of this variant available for sale.')
+    track_inventory = graphene.Boolean(
+        description="""Determines if the inventory of this variant should
+        be tracked. If false, the quantity won't change when customers
+        buy this item.""")
+    weight = WeightScalar(
+        description='Weight of the Product Variant.', required=False)
+
+
 class ProductInput(graphene.InputObjectType):
     attributes = graphene.List(
         AttributeValueInput,
@@ -225,6 +243,11 @@ class ProductCreateInput(ProductInput):
     product_type = graphene.ID(
         description='ID of the type that product belongs to.',
         name='productType', required=True)
+    default_variant = ProductVariantInput(
+        required=False, description="""
+            Default variant to assign to product of type that does not have
+            variants. If product type has variants and default variant is give,
+            returns an error""")
 
 
 class ProductCreate(ModelMutation):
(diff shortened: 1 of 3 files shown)


saleor/graphql/product/mutations/products.py
```diff
@@ -236,6 +259,21 @@ class Meta:
...
 
+    @classmethod
+    def clean_default_variant(cls, cleaned_input, product_type, errors):
+        default_variant = cleaned_input.pop('default_variant', None)
+        if default_variant and product_type.has_variants:
+            cls.add_error(
+                errors, 'default_variant', 'Product type has variants')
+        elif default_variant and not product_type.has_variants:
+            cleaned_input['default_variant'] = \
+                ProductVariantCreate.clean_attributes_input(
```
maarcingebala: Shouldn't we call `ProductVariantCreate. clean_input` here? From reading this code I assume that `clean_attributes_input` would only clean the attributes data, while it should clean the whole variant data.

saleor/graphql/product/mutations/products.py
```diff
@@ -260,6 +303,20 @@ def clean_input(cls, info, instance, input, errors):
...
+    @classmethod
+    def save(cls, info, instance, cleaned_input):
+        default_variant_data = cleaned_input.get('default_variant', None)
+        instance.save()
+        if default_variant_data:
+            default_variant_instance = models.ProductVariant()
+            default_variant_instance = ProductVariantCreate.construct_instance(
+                default_variant_instance, default_variant_data)
+            default_variant_instance.product = instance
+            default_variant_instance.save()
```
maarcingebala: We should also trigger somewhere `full_clean` on the newly created variant to validate the model. In the regular ModelMutation flow it's done in the `mutate` function through `clean_instance`. I'm not sure if we will be able to fit this call anywhere in this flow. Since this whole mutation got a bit complex now, maybe it would be much easier to stop using `ModelMutation` here, use `BaseMutation` instead and writing completely custom `mutate` to handle all that logic? We do that for some other complex mutations. What do you think?

saleor/graphql/product/mutations/products.py
```diff
@@ -236,6 +259,21 @@ class Meta:
         description = 'Creates a new product.'
         model = models.Product
 
+    @classmethod
+    def clean_default_variant(cls, cleaned_input, product_type, errors):
+        default_variant = cleaned_input.pop('default_variant', None)
+        if default_variant and product_type.has_variants:
+            cls.add_error(
+                errors, 'default_variant', 'Product type has variants')
+        elif default_variant and not product_type.has_variants:
```
NyanKiyoshi: That's a bit of a waste to evaluate twice the same conditions. Why not doing
```python
if default_variant:
    if product_type.has_variants:
         pass  # ...
    else:
         pass  # ...
```

It adds another block level, but removes useless CPU instructions (assuming Python does not optimize the code).

saleor/graphql/product/mutations/products.py
```diff
@@ -236,6 +259,21 @@ class Meta:
...
+            cleaned_input['default_variant'] = \
+                ProductVariantCreate.clean_attributes_input(
+                    product_type, default_variant, errors)
+        elif not default_variant and not product_type.has_variants:
```
NyanKiyoshi: Again, the logic should probably use some rethinking to avoid repetitions.

saleor/graphql/product/mutations/products.py
oldPadavan: Totally agree this looks better. Initially I also used nested if blocks, but I had to choose the current implementation because commit didn't pass `codeclimate` check - cognitive complexity of this method was above the threshold (nested if's add additional point to cognitive complexity score).

maarcingebala: True, but sometimes code written in such manner is much more readable and just feels more natural. There are some places in our code which don't pass this check but in my opinion, this is acceptable if we all agree on readability.

saleor/graphql/product/mutations/products.py
oldPadavan: I've made an attempt to implement all necessary functionality using `ModelMutation`. Frankly I don't really like how `clean_input` of `ProductVariantCreate` is called for `default_variant`. The issue was that `ProductVariantCreate.clean_input` requires `product` to be already created and accessible from variant instance or fetched by its id. On the other hand as I said before implementation of validation logic for `default_variant` separately from `ProductVariantCreate` leads to code duplication. I don't know what's better in this case. How do You think? Or maybe You can advice on another solution

maarcingebala:
Closing in favor of #3505

PR #2976, opened 2018-09-26
Ability to create values for Attribute in the create mutation

### LINKED ISSUE #2967: Allow creating attributes with values in one mutation
We need to change the signature of `productAttributeCreate` mutation to allow creating attribute values in the same mutation. This is required by the new UI for the product attributes dashboard.

Pacu2:
close #2967 
Added ability to create AttributeValues together with the Attribute in the mutation.

### DIFF
--- a/saleor/graphql/product/mutations/attributes.py
+++ b/saleor/graphql/product/mutations/attributes.py
@@ -68,7 +68,7 @@ class AttributeCreate(ModelMutation):
                     if field == 'attribute':
                         continue
                     for message in validation_errors.message_dict[field]:
-                        error_field = 'values-%(field)s' % {'field': field}
+                        error_field = 'values:%(field)s' % {'field': field}
                         cls.add_error(errors, error_field, message)
         return cleaned_input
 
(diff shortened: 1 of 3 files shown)


saleor/graphql/product/mutations/attributes.py
```diff
@@ -45,6 +49,43 @@ class Meta:
...
+            attribute_value = models.AttributeValue(**value_data)
+            try:
+                attribute_value.full_clean()
+            except ValidationError as validation_errors:
+                for field in validation_errors.message_dict:
+                    # Attribute instance is not created yet so we cannot add it
+                    if field == 'attribute':
+                        continue
+                    for message in validation_errors.message_dict[field]:
+                        error_field = 'values-%(field)s' % {'field': field}
```
maarcingebala: We implemented a similar idea [in other mutation](https://github.com/mirumee/saleor/blob/master/saleor/graphql/product/mutations/products.py#L347) but used colons to separate field and attribute name. I'd use the same format here. Later we should think of some standardization and probably a helper method to generate such errors.

PR #2977, opened 2018-09-26
Use fields default value if no value is provided and field is not nullable

### LINKED ISSUE #2973: Product type mutations should be able to reset charge_taxes flag
To create product type, that wouldn't be charged with taxes.

1. Execute `productTypeCreate` mutation and set `taxRate: null`
2. Get an error

### What I expected to happen
To mutation pass successfully.

Pacu2:
This is a fallback for fields that have a default, were nulled in the API, but are not nullable at the database level.

### DIFF
--- a/saleor/graphql/core/mutations.py
+++ b/saleor/graphql/core/mutations.py
@@ -211,11 +211,13 @@ class ModelMutation(BaseMutation):
         opts = instance._meta
 
         for f in opts.fields:
-            if not f.editable or isinstance(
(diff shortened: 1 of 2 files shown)


maarcingebala:
I think the problem is that there is no way to set empty tax rate (which is possible in the old dashboard and the model allows it as well). This PR seems useful anyway and I would keep that changes, but maybe we should also provide a solution that would allow passing an empty string (or add a TaxRateType that would set an empty string).

PR #2982, opened 2018-09-27
Change lines in Order type to graphene.List

### LINKED ISSUE #2900: Consider changing order lines to a list
Currently, the `lines` field in the `Order` type is a connection. I think we should simplify this and change it to a simple `graphene.List`. I cannot imagine a situation where pagination of those lines should be necessary. 

The query currently looks like this:

```
{
  order {
    id
    lines {
      edges {
        node {
          quantity
          productName
        }
      }
    }
  }
}
```

We could change it to:
```
{
  order {
    id
    lines {
      quantity
      productName
    }
  }
}
```

oldPadavan:

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/types.py
+++ b/saleor/graphql/order/types.py
@@ -87,6 +87,27 @@ class FulfillmentLine(CountableDjangoObjectType):
         exclude_fields = ['fulfillment']
 
 
+class OrderLine(CountableDjangoObjectType):
+    thumbnail_url = graphene.String(
+        description='The URL of a main thumbnail for the ordered product.',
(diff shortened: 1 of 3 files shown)


Pacu2:
Thanks for picking this up!

@dominik-zeglen do we need to upgrade the react part as well?

dominik-zeglen:
Definitely, but it should be easy to fix

Pacu2:
Thanks for picking this up! I've merged this branch into https://github.com/mirumee/saleor/pull/3009, where we will fix the Schema, Types and ReactJS part

PR #2988, opened 2018-09-28
Add bulk operations to Attributes mutations

### LINKED ISSUE #2981: Add addValues and removeValues arguments to attributeUpdate mutation
We should be able to do a bulk update of attribute values. I'd like to propose adding `removeValues: [ID!]` i `addValues:[String!]` arguments to `attributeUpdate`.

Pacu2:
This PR adds ability to bulk update/delete attribute values from the Attribute

Also, I've reworked Attribute and AttributeValue mutations to make it consistent with other we've introduced previously.

close #2981

### DIFF
--- a/saleor/graphql/product/mutations/attributes.py
+++ b/saleor/graphql/product/mutations/attributes.py
@@ -1,37 +1,32 @@
 import graphene
+from django.core.exceptions import ValidationError
+from django.db.models import Q
+from django.template.defaultfilters import slugify
+from graphql_jwt.decorators import permission_required
 
 from ....product import models
-from ...core.mutations import ModelDeleteMutation, ModelMutation
-from django.core.exceptions import ValidationError
+from ...core.mutations import BaseMutation, ModelDeleteMutation, ModelMutation
+from ..types import Attribute, AttributeValue
 
 
-class AttributeValueCreateInput(graphene.InputObjectType):
-    attribute = graphene.ID(
-        required=False,
-        description='Attribute to which value will be assigned.',
-        name='attribute')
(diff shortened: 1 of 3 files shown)


saleor/graphql/product/mutations/attributes.py
```diff
@@ -103,43 +194,77 @@ def user_is_allowed(cls, user, input):
...
 
     class Meta:
-        description = 'Creates an attribute choice value.'
-        model = models.AttributeValue
+        description = 'Creates an AttributeValue.'
 
     @classmethod
-    def user_is_allowed(cls, user, input):
-        return user.has_perm('product.manage_products')
+    def clean_inputt(cls, input, errors, attribute, attribute_value_pk=None):
```
NyanKiyoshi: inputt 👍

saleor/graphql/product/mutations/attributes.py
```diff
@@ -42,52 +37,148 @@ class Meta:
...
                     for message in validation_errors.message_dict[field]:
-                        error_field = 'values:%(field)s' % {'field': field}
+                        error_field = '%(values_field)s:%(field)s' % {
+                            'values_field': cls.ATTRIBUTE_VALUES_FIELD,
+                            'field': field}
                         cls.add_error(errors, error_field, message)
+        return errors
+
+    @classmethod
+    def clean_input(cls, info, instance, input, errors):
```
NyanKiyoshi: Could probably add a few comments/ refactor as this block is quite big?

Pacu2: That's still heavy WIP

PR #2999, opened 2018-10-01
Naming consistency of API mutations

### LINKED ISSUE #2757: Make API field names more consistent
We should introduce some project-wide pattern for endpoints and mutations naming.
In some parts of the project, we have different styles of naming for mutations and it makes hard to figure out if some API scenarios are covered or not.

I recommend to keep following approach:
`<Entity><Action` examples:
`categoryCreate`
`orderUpdate`

now we have for example `orderAddNote` and it should be refactored to `orderNoteAdd`
Also, we should consider using the same verbs for common operations like `Create`, `Update` or `Delete`. Now it's mixed up sometimes.

The second recommendation is to sort alphabetically and group mutations in `api.py` file because now it's hard to find interesting mutations there.

What do you think?

the-bionic:
resolves #2757

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/api.py
+++ b/saleor/graphql/api.py
@@ -31,10 +31,10 @@ from .order.mutations.draft_orders import (
     DraftOrderLineCreate, DraftOrderLineDelete, DraftOrderLineUpdate,
     DraftOrderUpdate)
 from .order.mutations.fulfillments import (
-    FulfillmentCancel, FulfillmentCreate, FulfillmentUpdateTracking)
+    FulfillmentCancel, FulfillmentCreate, FulfillmentTrackingUpdate)
 from .order.mutations.orders import (
-    OrderAddNote, OrderCancel, OrderCapture, OrderMarkAsPaid, OrderRefund,
-    OrderRelease, OrderUpdate, OrderUpdateShipping)
+    OrderNoteAdd, OrderCancel, OrderCapture, OrderMarkAsPaid, OrderRefund,
+    OrderRelease, OrderUpdate, OrderShippingUpdate)
 from .page.resolvers import resolve_pages, resolve_page
 from .page.types import Page
 from .page.mutations import PageCreate, PageDelete, PageUpdate
@@ -45,8 +45,8 @@ from .product.mutations.attributes import (
     AttributeUpdate)
 from .product.mutations.products import (
     CategoryCreate, CategoryDelete, CategoryUpdate,
-    CollectionAddProducts, CollectionCreate, CollectionDelete,
-    CollectionRemoveProducts, CollectionUpdate, ProductCreate,
+    CollectionProductsAdd, CollectionCreate, CollectionDelete,
+    CollectionProductsRemove, CollectionUpdate, ProductCreate,
     ProductDelete, ProductUpdate, ProductTypeCreate,
     ProductTypeDelete, ProductImageCreate, ProductImageDelete,
     ProductImageReorder, ProductImageUpdate, ProductTypeUpdate,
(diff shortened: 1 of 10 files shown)


saleor/graphql/api.py
```diff
@@ -313,21 +314,14 @@ class Mutations(graphene.ObjectType):
...
-    draft_order_line_create = DraftOrderLineCreate.Field()
-    draft_order_line_delete = DraftOrderLineDelete.Field()
-    draft_order_line_update = DraftOrderLineUpdate.Field()
-    draft_order_update = DraftOrderUpdate.Field()
     order_fulfillment_cancel = FulfillmentCancel.Field()
     order_fulfillment_create = FulfillmentCreate.Field()
-    order_fulfillment_update_tracking = FulfillmentUpdateTracking.Field()
-    order_add_note = OrderAddNote.Field()
+    order_fulfillment_update_tracking = FulfillmentTrackingUpdate.Field()
+    order_add_note = OrderNoteAdd.Field()
```
NyanKiyoshi: We have order_**add_note** but then we have **NoteAdd**.  We also have other cases below where we also have *ObjectName***Action** and not **Action**ObjectName. Plus, "OrderAddNote" sounds more correct English.

the-bionic: I think ```CreateOrder``` sounds more correct than ```OrderCreate```, but that is the convention. Same applies to OrderNoteAdd. 

Should OrderNoteCreate be used here instead?

I will update the variable names, must have missed that.

the-bionic: I think ```CreateOrder``` sounds more correct than ```OrderCreate```, but that is the convention. Same applies to OrderNoteAdd. 

Should OrderNoteCreate be used here instead?

I will update the variable names, must have missed that.

saleor/graphql/api.py
```diff
@@ -269,14 +269,15 @@ def resolve_address_validator(self, info, input):
...
 class Mutations(graphene.ObjectType):
-    assign_navigation = AssignNavigation.Field()
+    address_create = AddressCreate.Field()
+    address_update = AddressUpdate.Field()
+    address_delete = AddressDelete.Field()
 
-    token_create = CreateToken.Field()
-    token_refresh = graphql_jwt.Refresh.Field()
-    token_verify = VerifyToken.Field()
+    assign_navigation = AssignNavigation.Field()
```
maarcingebala: This should be renamed to `navigation_assign`, according to the <enitity><action> rule.

saleor/graphql/api.py
```diff
@@ -286,24 +287,24 @@ class Mutations(graphene.ObjectType):
     category_delete = CategoryDelete.Field()
     category_update = CategoryUpdate.Field()
 
+    collection_create = CollectionCreate.Field()
+    collection_update = CollectionUpdate.Field()
+    collection_delete = CollectionDelete.Field()
+    collection_products_add = CollectionProductsAdd.Field()
```
maarcingebala: As suggested [in this comment](https://github.com/mirumee/saleor/issues/2757#issuecomment-426213619) I would consider "collection" to be the entity. So the original name was correct.

saleor/graphql/api.py
```diff
@@ -377,6 +359,25 @@ class Mutations(graphene.ObjectType):
...
 
+    shop_domain_update = ShopDomainUpdate.Field()
+    shop_settings_update = ShopSettingsUpdate.Field()
+    homepage_collection_update = HomepageCollectionUpdate.Field()
+
+    staff_create = StaffCreate.Field()
+    staff_update = StaffUpdate.Field()
+    staff_delete = StaffDelete.Field()
+
+    set_password = SetPassword.Field()
```
maarcingebala: `password_set`

maarcingebala:
@the-bionic Changes look good, but we would also have to regenerate GraphQL schema (`npm run build-schema`) and ensure that the build of frontend code isn't failing after that.

maarcingebala:
@the-bionic There has been a lot of changes in the API recently and this PR is not up-to-date with them. Will you have to take a look at it in the nearest future? I just need to know whether I can expect some updates from your side. If not, maybe I'll get someone from our team to take this issue.

PR #3016, opened 2018-10-03
This Supposed to Be a Simple Test Client Change, You Won't Believe What This Developer Did Next!

### LINKED ISSUE #2650: API tests don't test permissions
Tests suite should be precise. Now all non-user actions are tested using `admin_api_client` fixture which operates on superuser privileges level. This way, we don't test if staff API endpoints contain correct permissions checking.
In the future, it may lead to a potential security issue.

### What I expected to happen
Staff API endpoints tests should use `staff_api_client` fixture and add correct permissions at the beginning of the test. Insufficient permissions test case should be covered as well.

### LINKED ISSUE #3003: Refactor API client in tests
Calling API in tests requires us to write some repetitive code:
```
admin_api_client.post(reverse('api'), {'query': query, 'variables': variables})
```

Both the URL and structure of the data dictionary are the same in every API, so we could adjust the `post` method of our `ApiClient` and have something like:
```
admin_api_client.post(query, variables)
```

We could also experiment with passing permissions to be used for a particular call, e.g.
```
admin_api_client.post(query, variables, permissions=[permission_manage_orders])
```

## TODO
- [x] New method `post_grapqhl` for simplified query posting
- [ ] All tests tested with `staff_api_client` instead of `admin_api_client`
- [ ] `admin_api_client` is removed
- [ ] API-related fixtures should be moved to `tests/api/conftest.py`

Pacu2:
In this PR:
- `post_graphql` was adjusted to automatically test permissions if provided
- Changed admin_api_client to staff_api_client(no more superuser-testing)
- Moved api-related fixtures to tests/api/conftest.py
- Basically, added proper permissions to all API tests

close #3003 
close #2650

### DIFF
--- a/tests/api/conftest.py
+++ b/tests/api/conftest.py
@@ -1,0 +1,79 @@
+import json
+
+import pytest
+
+from django.shortcuts import reverse
+from django.test.client import MULTIPART_CONTENT, Client
+from graphql_jwt.shortcuts import get_token
+
+from .utils import assert_no_permission
+
+API_PATH = reverse('api')
+
+
+class ApiClient(Client):
+    """GraphQL API client."""
+
+    def __init__(self, *args, **kwargs):
+        user = kwargs.pop('user')
+        self.user = user
+        self.token = get_token(user)
+        super().__init__(*args, **kwargs)
+
+    def _base_environ(self, **request):
+        environ = super()._base_environ(**request)
+        environ.update({'HTTP_AUTHORIZATION': 'JWT %s' % self.token})
+        return environ
+
+    def post(self, data=None, **kwargs):
+        """Send a POST request.
+
+        This wrapper sets the `application/json` content type which is
+        more suitable for standard GraphQL requests and doesn't mismatch with
+        handling multipart requests in Graphene.
+        """
+        if data:
+            data = json.dumps(data)
+        kwargs['content_type'] = 'application/json'
+        return super().post(API_PATH, data, **kwargs)
+
+    def post_graphql(
+            self, query=None, variables=None, permissions=None, **kwargs):
+        """Dedicated helper for posting GraphQL queries.
+
+        Sets the `application/json` content type and json.dumps the variables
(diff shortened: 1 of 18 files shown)


tests/api/conftest.py
```diff
@@ -0,0 +1,79 @@
...
+        more suitable for standard GraphQL requests and doesn't mismatch with
+        handling multipart requests in Graphene.
+        """
+        if data:
+            data = json.dumps(data)
+        kwargs['content_type'] = 'application/json'
+        return super().post(API_PATH, data, **kwargs)
+
+    def post_graphql(
+            self, query=None, variables=None, permissions=None, **kwargs):
```
patrys: How useful is the case where `query` is empty? I'd like to propose an alternative signature:

```python
def post_graphql(self, query, *, variables=None, permissions=None, **kwargs):
```

tests/api/conftest.py
```diff
@@ -0,0 +1,81 @@
...
+        if data:
+            data = json.dumps(data)
+        kwargs['content_type'] = 'application/json'
+        return super().post(API_PATH, data, **kwargs)
+
+    def post_graphql(self, query, variables=None, permissions=None, **kwargs):
+        """Dedicated helper for posting GraphQL queries.
+
+        Sets the `application/json` content type and json.dumps the variables
+        if present.
```
NyanKiyoshi: Maybe you could also note/ say that if permission were passed it checks (assert) that the permissions are correctly handled.

Pacu2: Right, will update the docstring soon

tests/api/conftest.py
patrys: Out of curiosity, why skip the single asterisk in the final implementation?

Pacu2: Couldn't find a use for it in our current test structure, what's more, it enforced to pass the variables as the keyword adding more boilerplate.

PR #3021, opened 2018-10-03
Fix default product tax rate in Dashboard 1.0

### LINKED ISSUE #2953: Product form tax_rate
When I edit a product on the dashboard I would like to have the product tax_rate as the selected option. Instead the product type tax_rate is selected.

1. Set a tax_rate on a product type
2. Set a different tax rate on a product
3. Go to the dashboard and edit the product
4. When you look at the tax_rate select field, the selected value is the product type tax_rate 

### What I expected to happen
When I edit a product I would like to have the product tax_rate (if set) as the selected option.

### Screenshots
issue_saleor

**System information**
Operating system: OSX 10.13
Browser: Firefox

Thank you.

tomcio1205:
I want to merge this change because it resolve #2953 

### Pull Request Checklist

### DIFF
--- a/saleor/dashboard/product/forms.py
+++ b/saleor/dashboard/product/forms.py
@@ -241,7 +241,8 @@ class ProductForm(forms.ModelForm, AttributesMixin):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         product_type = self.instance.product_type
-        self.initial['tax_rate'] = product_type.tax_rate
+        if self.instance.tax_rate:
+            self.initial['tax_rate'] = self.instance.tax_rate
         self.available_attributes = (
             product_type.product_attributes.prefetch_related('values').all())
         self.prepare_fields_for_attributes()
--- a/saleor/product/models.py
+++ b/saleor/product/models.py
@@ -81,7 +81,7 @@ class ProductType(models.Model):
         'Attribute', related_name='product_variant_types', blank=True)
     is_shipping_required = models.BooleanField(default=False)
     tax_rate = models.CharField(
-        max_length=128, default=DEFAULT_TAX_RATE_NAME, blank=True,
+        max_length=128, blank=True,
         choices=TaxRateType.CHOICES)
     weight = MeasurementField(
         measurement=Weight, unit_choices=WeightUnits.CHOICES,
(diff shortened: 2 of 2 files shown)


tomcio1205:
@Pacu2 I'm not sure how to add placeholder in select field. Should I add next option with product type tax rate but with empty value and set it as default when no tax rate is assigned on the Product ?

Pacu2:
I think we could change the empty value description.

Let's say:
If there's no tax_rate defined on ProductType we should display `--------` as it is at this moment. 
In other cases, we could display something like (greyed out) `VAT 5% (Inherited from ProductType)` (or something shorter)

patrys:
If we're going to use the default value then I suggest it says something like "Controlled by product type (currently: ____)".

BTW, do we ever set explicit tax percentage (eg. 5%) instead of tax groups ("hospitality")?

Pacu2:
We are always using tax groups, `VAT 5%` was made up by me

saleor/product/models.py
```diff
@@ -81,7 +81,7 @@ class ProductType(models.Model):
         'Attribute', related_name='product_variant_types', blank=True)
     is_shipping_required = models.BooleanField(default=False)
     tax_rate = models.CharField(
-        max_length=128, default=DEFAULT_TAX_RATE_NAME, blank=True,
+        max_length=128, blank=True,
```
maarcingebala: I think that using `DEFAULT_TAX_RATE_NAME` here and in the Product model is not problematic and we could keep it. What needs to be fixed is only the dashboard form.

maarcingebala: I think Django generates migrations for such changes, if yes, we should provide them with in PR.

saleor/dashboard/product/forms.py
```diff
@@ -241,7 +241,8 @@ class Meta:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         product_type = self.instance.product_type
-        self.initial['tax_rate'] = product_type.tax_rate
+        if self.instance.tax_rate:
```
maarcingebala: If `instance.tax_rate` is None we want to render the default from the product type. I'd suggest:

```
self.initial['tax_rate'] = self.instance.tax_rate or product_type.tax_rate
```

maarcingebala:
@tomcio1205 Would you have time to take a look at this PR again? It requires a few little changes and it would be ready to merge.

maarcingebala:
@tomcio1205 We'd need to rebase it with `master` as well.

maarcingebala:
Closing in favor of #3880.

PR #3035, opened 2018-10-04
Add --dev flag in pipenv install

maarcingebala:
Install dev dependencies on CircleCI.


patrys:
I think the proper solution would be to move debug toolbar from dev dependencies to regular dependencies since it's being imported by production code.

PR #3039, opened 2018-10-04
Adding an IF Condition to DEBUG / development dependencies in Dockerfile

### LINKED ISSUE #3038: Enable DEBUG option for build docker image

In some cases it may be interesting to be able to build the docker image with the development dependencies.

For example if I want to have a development environment not only on a local machine but also posted on some server.

I am not a docker master but I understand that if the image is not built with the development dependencies it will not be possible run the container with debug mode on.

Add a IF condition bound to the DEBUG parameter for the build-python image

```
### Build and install packages
FROM python:3.6 as build-python

RUN \
  apt-get -y update && \
  apt-get install -y gettext && \
  # Cleanup apt cache
  apt-get clean && \
  rm -rf /var/lib/apt/lists/*

ARG DEBUG

# Install Python dependencies
RUN pip install pipenv
ADD Pipfile /app/
ADD Pipfile.lock /app/
WORKDIR /app
RUN \
    if [ "$DEBUG" = "True" ] ; \
    then pipenv install --system --deploy --dev ; \
    else pipenv install --system --deploy ; \
    fi

### Build static assets
FROM node:10 as build-nodejs

.
.
.
.
```

It may be related to issue: #3019 

And PR's: #3018  #3015 #3035 #3036

daltonsena:

Close #3038 

### Pull Request Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,12 +8,18 @@ RUN \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
+ARG DEBUG
+
 # Install Python dependencies
 RUN pip install pipenv
 ADD Pipfile /app/
 ADD Pipfile.lock /app/
 WORKDIR /app
-RUN pipenv install --system --deploy --dev
+RUN \
+  if [ "$DEBUG" = "True" ] ; \
+  then pipenv install --system --deploy --dev ; \
+  else pipenv install --system --deploy ; \
+  fi
 
 ### Build static assets
 FROM node:10 as build-nodejs


jxltom:
Since by default ```DEBUG``` is ```True``` in saleor and then debugtoolbar and other dev packages are needed, so ```--dev``` is required for ```pipenv install``` anyway.

To solve this, I would recommend set ``` env``` DEBUG same as ```arg``` DEBUG, then:

- When ```docker build --build-arg DEBUG=True``` , ```pipenv install --system --deploy``` is performed and saleor will load DEBUG from env as False which enables production mode.

- When ```docker build --build-arg``` is not provide or not equal to True , ```pipenv install --system --deploy --dev``` is performed and saleor will load DEBUG from env as True which enables development mode.

By default, ```DEBUG`` should be ```False``` which enables production mode.

Besides, if this PR is accepted, it will be nice to also update doc https://saleor.readthedocs.io/en/latest/customization/docker.html and https://saleor.readthedocs.io/en/latest/customization/docker.htmlhttps://github.com/mirumee/saleor/blob/master/docs/deployment/docker.rst by reflecting the usage of ```--build-arg DEBUG=True|False```

jxltom:
Update: since https://github.com/mirumee/saleor/pull/3043, maybe this will be unneccessary?

patrys:
Yes, I think this won't be needed. Installing linters in docker is not a typical use case.

daltonsena:
Makes a lot of sense!
I'm going to close this pull request.

Thank you guys

PR #3040, opened 2018-10-04
Fixed tons of deprecated warnings using latest pytest

jxltom:
After https://github.com/mirumee/saleor/pull/2430 pytest is updated to latest version. Since pytest 3.8.0, ```DeprecationWarning``` and ```PendingDeprecationWarning``` are now shown by default if no other warning filter is configured. https://github.com/pytest-dev/pytest/blob/master/CHANGELOG.rst#features. This makes the test in CI having tons of deprecated warning messages.

This PR ignores deprecation warning for third-party packages and update saleor to remove the deprecated warnings.

### Pull Request Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,7 +3,6 @@ from unittest.mock import MagicMock, Mock
 
 import pytest
 
-import django
 from django.contrib.auth.models import Permission
 from django.contrib.sites.models import Site
 from django.core.files import File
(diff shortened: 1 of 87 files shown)


Pacu2:
Related:
https://github.com/mirumee/django-prices/pull/67
https://github.com/coddingtonbear/django-measurement/pull/83

patrys:
Didn't the contrib.staticfiles move to core in 2.x? We still need to support the current LTS which is 1.11.

jxltom:
@pacu2 Thanks for the follow-ups. I didn't ignore all the third-party packages' deprecated warnings but only most of them.  This is because the warnings may go sway as the packages upgrade.

For now, there are 20+ warnings left which is acceptable for me.

jxltom:
@patrys Don't worry. In django 1.11 (and of course django 2.x), https://github.com/django/django/blob/1.11/django/templatetags/static.py exists for [backwards compatibility](https://github.com/django/django/blob/58f02c498b659a906e9c30d946bd89bedc4717e5/django/contrib/staticfiles/templatetags/staticfiles.py#L9). They are just being depreated from django 2.x

tests/conftest.py
```diff
@@ -36,6 +37,33 @@
 from saleor.site.models import AuthorizationKey, SiteSettings
 
 
+def pytest_collection_modifyitems(items):
```
patrys: I'd love to merge this without this fixture. I fear this will immediately become unmaintained and will result in an unpleasant surprise in future.

jxltom: This fixture ignores deprecated warning in django-measurement and django-impersonate. Without this, there still will be amount of warnings in CI.

Or just disable all deprecation warning by default? But this may bypass some useful warning as well.

patrys: I'd rather see more warnings than necessary than accidentally miss some of the ones I care about. Does `pytest -Wd` still display the ones ignored here?

jxltom: No, it looks python's -Wd option won't override the markers added in ```pytest_collection_modifyitems```.

jxltom:
Update: give up 😭 Let it be. I can not make it work by applying ```module``` or ```once``` actions without fixtures.

Hope the remains warnings will be gone soon after django-measurement upgrades.

PR #3044, opened 2018-10-05
Support latest django-graphql-jwt

jxltom:
For making PR not so radical, https://github.com/mirumee/saleor/pull/2430 locked django-graphql-jwt to lower than 0.1.13.

In [django-graphql-jwt 0.1.13](https://github.com/flavors/django-graphql-jwt/blob/master/CHANGES.rst#0113), ```GraphQLJWTError``` is renamed to ```JSONWebTokenError``` and it does not provide backward compatibility.

Since it actually doesn't break compatibility with saleor and to have new features and fewer bugs, let's support the latest version and lock it to ```>=0.1.13```

### Pull Request Checklist


patrys:
Not sure why but seemingly unrelated dependency hashes change in this PR.

jxltom:
Because we didn't pin specific version in Pipfile. For example, we only pin ```Django = ">=1.11"```, so if there are any new releases such as ```2.1.12```, ```pipenv lock``` will upgrade these packages to the latest version only if they match the requirments in Pipfile.

It has same behaviour with ```pip-tools```.

It looks like there is no way to only upgrade only a specific package but leave others untouched, except you pin all packages with specifc version by ```==some_version```. But doing that will break the easiness of upgrading pacakges.

patrys:
`pip-tools` allows you to upgrade specific dependencies, I kind of expected `pipenv` to do the same.

jxltom:
Unfortunately, it is not supported yet https://github.com/pypa/pipenv/issues/966. 😢

PR #3047, opened 2018-10-05
Silent collectstatic's output in CI especially for Django 1.11

jxltom:
By default, the verbosity for collectstatic command is 1. But in django 1.11, the verbosity for logging ```coping files ...``` is also 1. This generates around 1000 lines in collectstatic output, which makes it difficult to find useful output in CI results.

https://github.com/django/django/blob/3d0344dc402e11c57b65ecf57481e681f790db2e/django/contrib/staticfiles/management/commands/collectstatic.py#L362

This PR silents the collectstatic's output by setting verbosity to 0.

### Pull Request Checklist


Pacu2:
What if the collectstatic fails on the CI?

jxltom:
@Pacu2 Good question. The ```verbosity``` argument only controls the logging to ```stdout``` which is called by collectstatics intently.

I'm sure the stderr will still be displayed in CI if there are any exceptions or errors in running collectstatic

jxltom:
@patrys Thanks for the suggestion. It sounds like another PR though, since it also relates with the frontend.

PR #3050, opened 2018-10-05
Do not collect unnecessary static files

jxltom:
As mentioned in https://github.com/mirumee/saleor/pull/3047#issuecomment-427364593, unnecessary static files are collected by collectstatic.

It will be more elegant to move to frontend's source files to such as ```/src```, but this sounds like a big refactor and many existed pull requests needs to rebase.

So this PR fixed this issue tempararily.

### Pull Request Checklist

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -123,7 +123,8 @@ STATICFILES_DIRS = [
     ('assets', os.path.join(PROJECT_ROOT, 'saleor', 'static', 'assets')),
     ('favicons', os.path.join(PROJECT_ROOT, 'saleor', 'static', 'favicons')),
     ('images', os.path.join(PROJECT_ROOT, 'saleor', 'static', 'images')),
-    ('dashboard', os.path.join(PROJECT_ROOT, 'saleor', 'static', 'dashboard'))]
+    ('dashboard_images', os.path.join(
+        PROJECT_ROOT, 'saleor', 'static', 'dashboard', 'images'))]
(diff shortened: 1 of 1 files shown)


jxltom:
@patrys With this solution, did I miss anything?

patrys:
It's hard to tell as IIRC we have the `AppDirectoriesFinder` enabled which could be collecting everything under `saleor/static` regardless of the directory list.

jxltom:
From my understanding, the ```AppDirectoriesFinder``` will only collect static files under ```static``` directory under each installed app. We've never put ```saleor``` under ```INSTALLED_APPS``` but instead of only putting such as ```saleor.checkout```, ```saleor.core``` and etc. So I think it should be fine?

patrys:
Let's see which files get picked up on CI.

jxltom:
Only django 1.11 shows collected files, please check https://travis-ci.org/mirumee/saleor/jobs/437655354

patrys:
The list looks correct but I've just noticed that two things change in the `STATICFILES_DIRS`:

1. The prefix changes from `dashboard` to `dashboard_images`, I do not see any template changes that reflect that
2. The output path, previously files would end up as `STATIC_ROOT/images/foo.bar` now the same file will become `STATIC_ROOT/foo.bar` once collected as the source root is specified to be one level deeper.

jxltom:
Thanks. Good catch! Let me think how to fix that.

jxltom:
Update: should be fixed now.

PR #3055, opened 2018-10-05
Add .gitattributes file to hide generated files diffs

### LINKED ISSUE #2673: Set up .gitattributes to mark generated files as such
Some of the files are generated by scripts (like `package-lock.json`, the GraphQL schema, `gql-types.ts` and so on), we should tell GitHub that those are generated to make them appear collapsed in pull requests by default. I don't think we need to mark them as binary though.

NyanKiyoshi:
Closes #2673.

This hides diffs from:
- Pipenv lock files;
- GraphQL generated files (types and schemas);
- mjml precompiled files;
- django generated migration files;
- pot files;
- svg files.

### Pull Request Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,18 +1,12 @@
 # Package manager lock files
-Pipfile.lock linguist-generated
-package-lock.json linguist-generated
-
-# Generated GraphQL schemas
-*.graphql linguist-generated
-
-# Generated GraphQL types
-gql-types.ts linguist-generated
+Pipfile.lock linguist-generated -diff -merge
+package-lock.json linguist-generated -diff -merge
 
 # Generated django migration files
 *_auto_*.py linguist-generated
 
 # Generated locale pot files
-*.po linguist-generated
+*.po -merge
 
 # Software made SVG files
 *.svg linguist-generated
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,21 @@
+# Package manager lock files
+Pipfile.lock linguist-generated
+package-lock.json linguist-generated
+
+# Generated GraphQL schemas
+*.graphql linguist-generated
+
+# Generated GraphQL types
+gql-types.ts linguist-generated
+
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,21 @@
+# Package manager lock files
+Pipfile.lock linguist-generated
+package-lock.json linguist-generated
+
+# Generated GraphQL schemas
+*.graphql linguist-generated
```
patrys: I'm 👎 on this particular change. That's the only way to understand how a backend change affects the API.

.gitattributes
```diff
@@ -0,0 +1,21 @@
...
+
+# Generated GraphQL types
+gql-types.ts linguist-generated
```
patrys: I think we got rid of that?

.gitattributes
```diff
@@ -0,0 +1,21 @@
...
+
+# Generated django migration files
+*_auto_*.py linguist-generated
+
+# Generated locale pot files
+*.po linguist-generated
```
patrys: These are not really generated.

.gitattributes
```diff
@@ -0,0 +1,21 @@
+# Package manager lock files
+Pipfile.lock linguist-generated
+package-lock.json linguist-generated
```
patrys: Could we also add `-diff` and `-merge` for these? It rarely makes sense to resolve these by hand.

.gitattributes
NyanKiyoshi: How do we want to handle them? Do we want to show the diffs?

patrys: I'd like to see a diff on GitHub but would explicitly tell git not to attempt a three-way merge.

NyanKiyoshi:
@Pacu2 you can remove `Help Wanted`, thanks again.

NyanKiyoshi:
@patrys I updated the changes, let me know if it's good now or not.

PR #3057, opened 2018-10-06
Allow more variables to be set through os_env

stephenmoloney:

Some variables need to be changed and it is more convenient to do so through the ENV
variables rather than rewriting the `settings.py` file.
Currently, for the helm charts, I need to add a `custom_settings.py` file to allow for this. Now, I 
will leave the ability to add a `custom_settings.py` in the helm charts as users may have specific
reasons for that but I'd prefer that the default case is that it is not required.

What does this commit/MR/PR do?

- Allows more variables to be set via os_env
- Makes app more configurable from os_env rather than
settings.py
- Changes the default timezone to UTC rather than `America/Chicago`

Why is this commit/MR/PR needed?

- Try to move away from relying on changing settings.py as much as
possible.
- More flexibility

### Pull Request Checklist


saleor/settings.py
```diff
@@ -83,9 +83,9 @@ def get_bool_from_env(name, default_value):
     ('zh-hans', _('Chinese')),
     ('zh-tw', _('Chinese (Taiwan)'))]
 LOCALE_PATHS = [os.path.join(PROJECT_ROOT, 'locale')]
-USE_I18N = True
-USE_L10N = True
-USE_TZ = True
```
stephenmoloney: Should this not be `False` by default. See docs here. https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-USE_TZ

docs/gettingstarted/configuration.rst
```diff
@@ -76,3 +76,91 @@ Environment variables
...
+  Whether one should be logged out after changing one's password. Defaults to `False`
+
+``DB_SEARCH_ENABLED``
+  Whether or not to enable the search queries in the database. Will be superceded
+  by elasticsearch if elasticsearch is enabled. Defaults to `True`
+
+``SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL``
+
+``NOCAPTCHA``
+  Whether to activate google's nocaptha feature for logging in.
```
NyanKiyoshi: As the comment says in the `settings`, it's not a toggle for the Google's noCaptcha feature, at all. The value should not, by any mean be changed to `False`. The old Google's captcha (reCaptcha) is no longer existing and supported by Google.

To enable the noCaptcha feature ("Are you human? Just check the box the prove it."), you need to provide API keys, otherwise it stays disabled.

saleor/settings.py
```diff
@@ -525,13 +530,12 @@ def get_host():
 # This enable the new 'No Captcha reCaptcha' version (the simple checkbox)
 # instead of the old (deprecated) one. For more information see:
 #   https://github.com/praekelt/django-recaptcha/blob/34af16ba1e/README.rst
-NOCAPTCHA = True
+NOCAPTCHA = get_bool_from_env('NOCAPTCHA', True)
```
NyanKiyoshi: Could you change the comment to say reCaptcha (v1) was shutdown instead of deprecated?
(And restore the `True` non-changeable constant as explained in my above comment).

docs/gettingstarted/configuration.rst
stephenmoloney: Ah, you're right, I forgot that after the break !

saleor/settings.py
stephenmoloney: :+1: will do

docs/gettingstarted/configuration.rst
```diff
@@ -76,3 +76,91 @@ Environment variables
...
+``AVAILABLE_CURRENCIES``
+  A comma separated list of the currencies to be available for selection.
+
+``ACCOUNT_ACTIVATION_DAYS``
+
+``LOGIN_REDIRECT_URL``
+  Redirect url on unsuccessful login attempts
+
+``LOW_STOCK_THRESHOLD``
+  Low stock flag threshold for products
```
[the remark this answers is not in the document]
stephenmoloney: :+1:  will do

docs/gettingstarted/configuration.rst
```diff
@@ -76,3 +76,91 @@ Environment variables
...
+
+``MAX_CART_LINE_QUANTITY``
+
+``PAGINATE_BY``
+  Number of products per page to display in public customer facing site
```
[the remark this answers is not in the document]
stephenmoloney: :+1: yes

docs/gettingstarted/configuration.rst
```diff
@@ -76,3 +76,91 @@ Environment variables
...
+  The placeholder image url of size 120 x 120
+
+``PLACEHOLDER_IMAGE_255``
+  The placeholder image url of size 255 x 255
+
+``PLACEHOLDER_IMAGE_540``
+  The placeholder image url of size 540 x 540
+
+``PLACEHOLDER_IMAGE_1080``
+  The placeholder image url of size 1080 x 1080
```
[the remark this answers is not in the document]
stephenmoloney: good suggestion, can you paste some idea exactly what you mean though ?

docs/gettingstarted/configuration.rst
```diff
@@ -76,3 +76,91 @@ Environment variables
...
+
+``DEFAULT_PLACEHOLDER``
+  The url for the default placeholder. Defaults to the `PLACEHOLDER_IMAGE_255` url.
```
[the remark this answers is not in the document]
stephenmoloney: sounds good to me

docs/gettingstarted/configuration.rst
```diff
@@ -76,3 +76,91 @@ Environment variables
...
+
+``LOGOUT_ON_PASSWORD_CHANGE``
+  Whether one should be logged out after changing one's password. Defaults to `False`
+
+``DB_SEARCH_ENABLED``
+  Whether or not to enable the search queries in the database. Will be superceded
```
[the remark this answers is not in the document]
stephenmoloney: Can someone write this please? I'm not sure what it does exactly.

docs/gettingstarted/configuration.rst
```diff
@@ -76,3 +76,91 @@ Environment variables
...
+  by elasticsearch if elasticsearch is enabled. Defaults to `True`
+
+``SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL``
```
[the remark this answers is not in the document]
stephenmoloney: So, I will remove this altogether I think.

docs/gettingstarted/configuration.rst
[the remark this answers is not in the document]
stephenmoloney: I think the repitition is ok and it is explicit.

saleor/settings.py
```diff
@@ -325,9 +325,9 @@ def get_bool_from_env(name, default_value):
 VATLAYER_ACCESS_KEY = os.environ.get('VATLAYER_ACCESS_KEY')
 VATLAYER_USE_HTTPS = get_bool_from_env('VATLAYER_USE_HTTPS', False)
 
-ACCOUNT_ACTIVATION_DAYS = 3
```
stephenmoloney: Where in saleor is this actually used ? I am not sure what to write in the documentation for this variable.

Pacu2: I think it's legacy

stephenmoloney:
Any word on this ? 

I'd like to close it if it is not suitable for merge.

saleor/settings.py
```diff
@@ -414,13 +417,19 @@ def get_host():
...
+        'PLACEHOLDER_IMAGE_120', 'images/placeholder120x120.png'),
+    255: os.environ.get(
+        'PLACEHOLDER_IMAGE_255', 'images/placeholder255x255.png'),
+    540: os.environ.get(
+        'PLACEHOLDER_IMAGE_540', 'images/placeholder540x540.png'),
+    1080: os.environ.get(
+        'PLACEHOLDER_IMAGE_1080', 'images/placeholder1080x1080.png')}
+
+DEFAULT_PLACEHOLDER = os.environ.get(
+    'DEFAULT_PLACEHOLDER_IMAGE', os.environ.get('PLACEHOLDER_IMAGE_255'))
```
NyanKiyoshi: What about `DEFAULT_PLACEHOLDER = PLACEHOLDER_IMAGES[255]` instead of introducing a new environment variable?

docs/gettingstarted/configuration.rst
```diff
@@ -76,3 +76,74 @@ Environment variables
...
+``DASHBOARD_SEARCH_LIMIT``
+  Number of search results to display after a product search. Defaults to `5`.
+
+Placeholder images:
+``PLACEHOLDER_IMAGE_60`` - The relative path of the 60 x 60 placeholder image. Defaults to `images/placeholder60x60.png`.
+``PLACEHOLDER_IMAGE_120`` - The relative path of the 120 x 120 placeholder image. Defaults to `images/placeholder120x120.png`.
+``PLACEHOLDER_IMAGE_255`` - The relative path of the 255 x 255 placeholder image. Defaults to `images/placeholder255x255.png`.
+``PLACEHOLDER_IMAGE_540`` - The relative path of the 540 x 540 placeholder image. Defaults to `images/placeholder540x540.png`.
+``PLACEHOLDER_IMAGE_1080`` - The relative path of the 1080 x 1080 placeholder image. Defaults to `images/placeholder1080x1080.png`.
+``DEFAULT_PLACEHOLDER`` - The default placeholder value. Defaults to the `PLACEHOLDER_IMAGE_255` value.
```
NyanKiyoshi: This you test the above lines results? Seems broken. Also inline code is a double grave accent instead of a single one in RST (` ``something`` `).

> @NyanKiyoshi -- You probably could pack `60x60, 120x120, ..., 1080x1080` together instead of repeating the same sentence and also precise it can be an absolute path?

> @stephenmoloney -- good suggestion, can you paste some idea exactly what you mean though ?

I was thinking of a table, maybe. (Sorry for the late answer, I'm swamped by work). Something like that:

---

You can customize the placeholder images, which come to the following dimensions:


| environment variable name | dimensions | default value |
| -- | -- | -- |
| `PLACEHOLDER_IMAGE_60` | 60x60 | `images/placeholder60x60.png`
| `PLACEHOLDER_IMAGE_120` | 120x120 | `images/placeholder120x120.png`
| `PLACEHOLDER_IMAGE_255` | 255x255 | `images/placeholder255x255.png`
| `PLACEHOLDER_IMAGE_540` | 540x540 | `images/placeholder540x540.png`
| `PLACEHOLDER_IMAGE_1080` | 1080x1080 | `images/placeholder1080x1080.png`

stephenmoloney: Good suggestion, thanks.

patrys: Why is it useful to override the paths to placeholders instead of providing custom images under the same path?

stephenmoloney: The benefit of this change is minimal no doubt but providing this flexibility does no harm really. 
Overriding is optional, the default is to use the images under the same path.
For example, perhaps you want to keep less divergence of your fork with upstream.

patrys: It does have potential impact of exceeding the maximum allowed environment size. For example Amazon Elastic Beanstalk has a total environment limit set to 4kB. Anything in excess is silently dropped from the environment configuration.

stephenmoloney: Ok, I'll remove this block of code and rebase.

stephenmoloney:
@patrys 
Ok, I'll close this PR.

PR #3060, opened 2018-10-07
Issue 2749, added alt text to cart icon so that it will say Items if …

### LINKED ISSUE #2749: [Templates] [Cosmetics] Add Alt text to cart
Add Alt text to 
`<svg data-src="{% static "images/cart.svg" %}" width="35" height="30"/>`
in **base.html**
details: https://www.w3.org/wiki/HTML/Usage/TextAlternatives

MartinSeibert:

Issue 2749 suggests that there should be a text alternative in case the cart icon doesn't load. I added a text default to "Items" in case the icon doesn't load.

Fixes #2749

### Pull Request Checklist

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -149,7 +149,7 @@
                   {% trans "Your Cart" context "Main navigation item" %}
                 </span>
                 <div class="navbar__brand__cart__icon">
-                  <svg data-src="{% static "images/cart.svg" %}" width="35" height="30"/>
+                  <svg data-src="{% static "images/cart.svg" %}" width="35" height="30" alt="Items"/>
                 </div>
                 <span class="badge {% if not cart_counter %}empty{% endif %}">
                   {% if cart_counter %}


templates/base.html
```diff
@@ -149,7 +149,7 @@
                   {% trans "Your Cart" context "Main navigation item" %}
                 </span>
                 <div class="navbar__brand__cart__icon">
-                  <svg data-src="{% static "images/cart.svg" %}" width="35" height="30"/>
+                  <svg data-src="{% static "images/cart.svg" %}" width="35" height="30" alt="Items"/>
```
maarcingebala: I think that according to the good practices of writing alt texts we should be more descriptive here and say rather something like "An icon of a shopping cart." or even "An icon of a shopping cart possibly with the number of items added to the cart".

patrys: I think in this particular case the icon is purely decorative as the textual label besides it does a good job describing what the button does. Decorative elements should have an empty alt.

NyanKiyoshi: Plus, the label right next to the icon is saying "Your cart" then we have the cart's item count badge; it's already really descriptive and comprehensible without any alt text, even for blind users' accessibility.

<p align='center'>screenshot</p>

NyanKiyoshi: I'm not sure what's the question? You can edit the file and then commit. If you want to clean the commit history, you can rewrite the history using `git reset HEAD~1 --soft` (I would only recommend if you are used to `git` though), then commit over it and force push.

maarcingebala:
In the review, it was suggested that alt text is not necessary in this case.

PR #3076, opened 2018-10-09
Move debugtoolbar and silk to regular

jxltom:
As decided here https://github.com/mirumee/saleor/issues/3042#issuecomment-427298303, packages related with codebase directly are moved to regular requirements.

It is waiting for https://github.com/mirumee/saleor/pull/3044 to be merged first, otherwise there will be conflicts since they are related with requirements files.

Since these packages are moved to regular requirements, ```--dev``` in building docker image is unnecessary.

Fixes: https://github.com/mirumee/saleor/issues/3042

### Pull Request Checklist


jxltom:
BTW, there some issues with recent pipenv version (2018.6.25/2018.7.1). If there are any issues relating with locking, please use ```11.9.0```.

Since https://github.com/pypa/pipenv/pull/2940 and https://github.com/pypa/pipenv/pull/2935, which fixed  amounts of parsing and locking issues, are merged, I' sure the last release of pipenv, which will be soon I think, will be a stable one.

```pip install https://github.com/pypa/pipenv/archive/master.zip``` also works if anyone would like to try the latest pipenv version.

Pacu2:
I don't think we wanted to move `Silk` do the regular requirements, only debug toolbar is used if one uses DEBUG mode,  Silk is the matter of choice and needs to be enabled manually.

jxltom:
If that is the case, https://github.com/mirumee/saleor/pull/3039 may still be reasonable. Someone may want to enable silk in docker. If it is put in dev packages, it can never be enabled in docker because it is not installed.

Also, in heroku, it won't install dev packages by default. So one can never enable silk in server deployed by heroku since it won't be installed in heroku at all, if it is put in dev packages.

Pacu2:
To be honest I doubt that Silk should ever make it to the production environment.

jxltom:
Actually I think even debugtoolbar should never be used in production server. And that is why I comment here https://github.com/mirumee/saleor/pull/3043#issuecomment-427249575.

But it looks like https://github.com/mirumee/saleor/pull/3043#issuecomment-427291080 think it should be regular packages even it is controlled by environment variables.

I'm fine with installing these kind of packages in production server as long as it is under control and environment variables for enabling them are never be enabled in production server.

Pacu2:
@jxltom We've discussed it internally, all dependencies (even optional like Debug toolbar or Silk) should live inside of the regular requirements. 

dev_requirements.txt are just the tools that are helping developers in doing their jobs, like linters etc.

jxltom:
Cool. A discussion to clarify this is very helpful.

maarcingebala:
Build on CircleCI failed due to a missing `pytest` dependency. I think we should move it to regular dependencies as well since it is required by the test suite, along with all of its plugins.

https://circleci.com/gh/mirumee/saleor/3605

jxltom:
@maarcingebala Oh, I didn't know there are also tests are running in docker containers. https://github.com/mirumee/saleor/pull/3076/files#diff-3254677a7917c6c01f55212f86c57fbf is should be reverted.

PR #3082, opened 2018-10-09
Add mutations to manage authorization keys

### LINKED ISSUE #3080: Mutations to mange authorization keys
Add mutations to add/remove authorization keys.

maarcingebala:
Fixes #3080 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/api.py
+++ b/saleor/graphql/api.py
@@ -66,7 +66,8 @@ from .shipping.mutations import (
 
 from .shop.types import Shop
 from .shop.mutations import (
-    ShopDomainUpdate, ShopSettingsUpdate, HomepageCollectionUpdate)
+    AuthorizationKeyAdd, AuthorizationKeyDelete, ShopDomainUpdate,
+    ShopSettingsUpdate, HomepageCollectionUpdate)
 
 
 class Query(graphene.ObjectType):
@@ -271,6 +272,9 @@ class Query(graphene.ObjectType):
 
 
 class Mutations(graphene.ObjectType):
+    authorization_key_add = AuthorizationKeyAdd.Field()
+    authotization_key_delete = AuthorizationKeyDelete.Field()
+
     assign_navigation = AssignNavigation.Field()
 
     token_create = CreateToken.Field()
--- a/saleor/graphql/api.py
+++ b/saleor/graphql/api.py
@@ -273,7 +273,7 @@ class Query(graphene.ObjectType):
 
 class Mutations(graphene.ObjectType):
     authorization_key_add = AuthorizationKeyAdd.Field()
-    authotization_key_delete = AuthorizationKeyDelete.Field()
+    authorization_key_delete = AuthorizationKeyDelete.Field()
 
     assign_navigation = AssignNavigation.Field()
 
--- a/saleor/graphql/shop/mutations.py
+++ b/saleor/graphql/shop/mutations.py
@@ -1,11 +1,11 @@
-
 import graphene
 from graphql_jwt.decorators import permission_required
 
+from ...site import models as site_models
 from ..core.mutations import BaseMutation
 from ..core.types.common import WeightUnitsEnum
 from ..product.types import Collection
-from .types import Shop
+from .types import AuthorizationKey, AuthorizationKeyType, Shop
 
 
 class ShopSettingsInput(graphene.InputObjectType):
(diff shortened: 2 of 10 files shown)


saleor/graphql/api.py
```diff
@@ -271,6 +272,9 @@ def resolve_address_validator(self, info, input):
 
 
 class Mutations(graphene.ObjectType):
+    authorization_key_add = AuthorizationKeyAdd.Field()
+    authotization_key_delete = AuthorizationKeyDelete.Field()
```
Pacu2: typo `authotization` -> `authorization`

saleor/graphql/shop/mutations.py
```diff
@@ -108,3 +108,69 @@ def mutate(cls, root, info, collection):
...
+class AuthorizationKeyAdd(BaseMutation):
+    authorization_key = graphene.Field(
+        AuthorizationKey, description='Newly added authorization key.')
+
+    class Meta:
+        description = 'Adds an authorization key.'
+
+    class Arguments:
+        key_type = AuthorizationKeyType(
+            required=True, description='Type of authorization key to add.')
```
Pacu2: I think we are missing `an` before the `authorization` (as in the description below)

saleor/graphql/shop/mutations.py
```diff
@@ -108,3 +108,69 @@ def mutate(cls, root, info, collection):
...
+    @permission_required('site.manage_settings')
+    def mutate(cls, root, info, key_type, input):
+        errors = []
+        if site_models.AuthorizationKey.objects.filter(name=key_type).exists():
+            cls.add_error(
+                errors, 'key_type', 'Authorization key already exists.')
+            return AuthorizationKeyAdd(errors=errors)
+
+        site_settings = info.context.site.settings
+        instance = site_models.AuthorizationKey.objects.create(
```
Pacu2: Do we want to save it at this point, or should we just create an instance, without saving it, then run `cls.clean_instance`, and if there's no errors - save it?

saleor/graphql/shop/mutations.py
```diff
@@ -108,3 +108,69 @@ def mutate(cls, root, info, collection):
...
+        return AuthorizationKeyAdd(authorization_key=instance)
+
+
+class AuthorizationKeyDelete(BaseMutation):
+    authorization_key = graphene.Field(
+        AuthorizationKey, description='Auhtorization key that was deleted.')
+
+    class Arguments:
+        key_type = AuthorizationKeyType(
+            required=True, description='Type of key to delete.')
```
Pacu2: `Type of a key`

saleor/graphql/shop/mutations.py
```diff
@@ -108,3 +108,69 @@ def mutate(cls, root, info, collection):
...
+
+    class Meta:
+        description = 'Deletes an authorization key.'
+
+    @classmethod
+    @permission_required('site.manage_settings')
+    def mutate(cls, root, info, key_type):
+        errors = []
+        try:
```
Pacu2: Personally I like `first()` a little bit more than try-except, I find it more readable

saleor/graphql/shop/types.py
```diff
@@ -53,7 +62,10 @@ class Shop(graphene.ObjectType):
         Geolocalization,
         description='Customer\'s geolocalization data.')
     authorization_keys = graphene.List(
-        AuthorizationKey, description='List of configured authorization keys.',
+        AuthorizationKey,
+        description='''List of configured authorization keys. Authorization
+        keys are used to enable thrid party OAuth authorization (currently
```
Pacu2: To be precise it's `OAuth2`

maarcingebala: Is the concept of `client_id` and `client_secret` fields specific only to OAuth2 and not valid in OAuth1? I meant general concept of OAuth.

saleor/graphql/shop/mutations.py
maarcingebala: `try..except` pattern is also used across our codebase and API as well, both are readable in my opinion.

saleor/graphql/shop/types.py
Pacu2: As far as I know it's the same concept, matter of fact if we want to say precisely that OAuth2 authorization is supported by us, or that this feature can be used for OAuth in general

maarcingebala: `python-social-auth` which we use under the hood provides backends for both OAuth1 and OAuth2 authorizations. All of them could be potentially enabled in Saleor, so I think that generally saying about OAuth here should be fine.

PR #3083, opened 2018-10-09
2200/Checking if user is already impersonating

### LINKED ISSUE #2200: Unable to impersonate a user when already impersonating
To impersonating another user while impersonating someone, because I'm too lazy to click on 'Stop impersonating' :wink: 

1. Impersonate a customer;
2. Impersonate someone else;
3. The impersonating should have failed, meaning that you are still impersonating the same user (which may be confusing).

### What I expected to happen
To be impersonating the newly selected user.

### What happened instead/how it failed
Still impersonating the previous user and no message to warn me about it.

---

It's quite a simple bug: when wanting to impersonate a user from dashboard (`/dashboard/customers/`), we are clicking on a link that is pointing to `/impersonate/<uid>/`. So we are pointed outside of the dashboard.

As of now, Saleor is excluding dashboard from being impersonated, so if in a dashboard view we are trying to read the current user, we will get the admin user, not the impersonated user.

Now, we are trying to impersonate someone, we hit `/impersonate/<uid>`, which is not excluded from impersonation. So we are hitting this view as the impersonated, not as the admin. Which gets us denied from impersonating, and thus, redirects us to the homepage.

So, the fix is fairly simple: exclude `/impersonate/` from impersonation in settings. That's it.

**But now...** The question is: why should the impersonation views be outside of dashboard? Shouldn't we move them over `dashboard.customer`? Wouldn't it make more sense to do that way, and thus not having to add a new exclusion?

Thanks a lot for your feedback!

tomcio1205:
I want to merge this change because it resolve #2200 

### Pull Request Checklist

### DIFF
--- a/saleor/core/views.py
+++ b/saleor/core/views.py
@@ -3,7 +3,7 @@ import json
 from django.contrib import messages
 from django.template.response import TemplateResponse
 from django.utils.translation import pgettext_lazy
-from impersonate.views import impersonate as orig_impersonate
+from impersonate.views import impersonate as orig_impersonate, stop_impersonate
 
 from ..account.models import User
 from ..dashboard.views import staff_member_required
@@ -32,6 +32,9 @@ def styleguide(request):
 
 
 def impersonate(request, uid):
+    if request.user.is_impersonate:
+        stop_impersonate(request)
+        request.user = request.impersonator
     response = orig_impersonate(request, uid)
     if request.session.modified:
         msg = pgettext_lazy(
--- a/saleor/core/views.py
+++ b/saleor/core/views.py
@@ -33,7 +33,6 @@ def styleguide(request):
 
 def impersonate(request, uid):
     if request.user.is_impersonate:
-        stop_impersonate(request)
         request.user = request.impersonator
     response = orig_impersonate(request, uid)
     if request.session.modified:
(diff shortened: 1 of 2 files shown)


saleor/core/views.py
```diff
@@ -32,6 +32,9 @@ def styleguide(request):
 
 
 def impersonate(request, uid):
+    if request.user.is_impersonate:
+        stop_impersonate(request)
+        request.user = request.impersonator
```
NyanKiyoshi: I'm not sure about those changes. (Sorry I don't have enough time to test this right now).

From the [source code](https://bitbucket.org/petersanchez/django-impersonate/src/ba74bb2af5adbfb4219434ed7c14ffd486a71eb7/impersonate/views.py?at=default&fileviewer=file-view-default#views.py-48:79), I feel like it is not needed to call this view, it doesn't seem to be doing anything needed to replace the currently impersonated user, as what we want to do here is to replace the currently impersonated user by a new UID. And calling something that acts as a view seems really odd.

tomcio1205:
@NyanKiyoshi You have right. Calling this method wasn't necessary but I think I still need this line
`    if request.user.is_impersonate:
        request.user = request.impersonator`
because if I we don't change requesting user, impersonating user has no permission to call impersonate method.

maarcingebala:
Closing this since this is a stale PR and we're not sure if this is the correct solution.

Issue #3049, opened 2018-10-05
FileNotFoundError: [Errno 2] No such file or directory: '/app/saleor/static/assets'


joshuacox:
### What I'm trying to achieve

docker-compose run web python3 manage.py collectstatic 

### Steps to reproduce the problem
1.  clone saleor
1. docker-compose build
1. docker-compose run web python3 manage.py migrate
1. then:
```
$ docker-compose run web python3 manage.py collectstatic   
Starting saleor-redis         ... done
Starting saleor-postgres      ... done
Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 316, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 353, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.6/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 188, in handle
    collected = self.collect()
  File "/usr/local/lib/python3.6/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 105, in collect
    for path, storage in finder.list(self.ignore_patterns):
  File "/usr/local/lib/python3.6/site-packages/django/contrib/staticfiles/finders.py", line 125, in list
    for path in utils.get_files(storage, ignore_patterns):
  File "/usr/local/lib/python3.6/site-packages/django/contrib/staticfiles/utils.py", line 23, in get_files
    directories, files = storage.listdir(location)
  File "/usr/local/lib/python3.6/site-packages/django/core/files/storage.py", line 313, in listdir
    for entry in os.listdir(path):
FileNotFoundError: [Errno 2] No such file or directory: '/app/saleor/static/assets'
```

### What I expected to happen

not certain, but looking in that directory there is no `assets`:

```
$ docker-compose run web /bin/bash -c 'ls -alh /app/saleor/static'
Starting saleor-postgres      ... done
Starting saleor-elasticsearch ... done
total 40K
[...]

Issue #2642, opened 2018-08-20
Docker installation problem


saikia1:
I was trying to run the application using docker-compose

### Steps to reproduce the problem
1. `docker-compose up`

2. visit localhost:8000

### What I expected to happen
To land in the webpage

### Error
Traceback (most recent call last):
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 35, in inner
web_1     |     response = get_response(request)
web_1     |   File "./saleor/core/middleware.py", line 80, in middleware
web_1     |     request.site = Site.objects.get_current()
web_1     |   File "./saleor/site/patch_sites.py", line 24, in new_get_current
web_1     |     site = self.prefetch_related('settings').filter(pk=site_id)[0]
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 306, in __getitem__
web_1     |     qs._fetch_all()
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 1179, in _fetch_all
web_1     |     self._result_cache = list(self._iterable_class(self))
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 53, in __iter__
web_1     |     results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/db/models/sql/compiler.py", line 1068, in execute_sql
web_1     |     cursor.execute(sql, params)
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 100, in execute
web_1     |     return super().execute(sql, params)
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 68, in execute
web_1     |     return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 77, in _execute_with_wrappers
web_1     |     return executor(sql, params, many, context)
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 85, in _execute
web_1     |     return self.cursor.execute(sql, params)
web_1     |   File "/usr/local/lib/python3.6/site-packages/django/db/utils.py", line 89, in __exit__
web_1     |     raise dj_exc_value.with_traceback(traceback) from exc_value
[...]

Issue #2506, opened 2018-08-03
Docker - Error reading ./webpack-bundle.json.


apburton84:
### What I'm trying to achieve

Follow the install instructions.

### Steps to reproduce the problem

1. docker-compose build
2. docker-compose run web python3 manage.py migrate
3. docker-compose run web python3 manage.py collectstatic
4. docker-compose run web python3 manage.py populatedb --createsuperuser
5. docker-compose up
6. Goto http://127.0.0.1:8000

### What I expected to happen

I expect to see the Saleor homepage

#### What I actually happens

    Error reading ./webpack-bundle.json. Are you sure webpack has generated the file and the path is correct?

**System information**
Operating system: MacOS High Sierra 10.13.3
Browser: Chrome 67.0.3396.99 (Official Build) (64-bit)
Docker Engine: 18.06.0-ce
Docker Compose: 1.22.0
Docker Machine: 0.15.0

Running ```yarn run build-assets``` does not resolve the issue, as yarn is not installed on the docker container.  #874 

It seems that ```/app/webpack-bundle.json``` is not being copied from build-nodejs (node:8.6.0) as instructed in the Dockerfile.

    COPY --from=build-nodejs /app/webpack-bundle.json /app/

```npm run build-assets --production``` which runs ```webpack -p --progress``` should have created ```/app/webpack-bundle.json```?

A log of the output of docker-compose up

    Step 25/35 : COPY --from=build-nodejs /app/webpack-bundle.json /app/   ---> 68c7bcd63e49

and docker history shows the ```/app/webpack-bundle.json``` being copied to the container.

    68c7bcd63e49        20 minutes ago      /bin/sh -c #(nop) COPY file:8bbd3e3203b94232…   1.18kB

https://stackoverflow.com/questions/51672994/saleor-docker-error-reading-webpack-bundle-json

**Comment by michaeldel** [comment]
The actual issue here is that the `/app` directory is mapped to the project dir through a volume because of this:
```yml
services:
  ...
  web:
    ...
    volumes:
      - .:/app:Z
```
Because all of these files indeed are in the built image (here the user is `root` because I have remove that `useradd` in `Dockerfile`):
```
$ docker run -it $(basename $(pwd)) ls -lh | grep -E 'bundle|static'
drwxr-xr-x 11 root root 4.0K Aug 10 12:19 static
-rw-r--r--  1 root root 1.2K Aug 10 12:16 webpack-bundle.json
```
One solution is to compile static files from host:
```
[...]

PR #3091, opened 2018-10-10
Remove host mounted volume in docker-compose

tiangolo:
Remove host mounted volume in docker-compose as the app expects files created at build time that doesn't exist in the mounted source files.

Even though I like the development trick of mounting the source code during development to be able to "hot reload" instead of having to build the image from scratch again, it doesn't work in this case, because it expects some files and directories to exists.

It expects:

* a directory `/app/static` , [created at build time](https://github.com/mirumee/saleor/blob/master/Dockerfile#L64)
* a directory `/app/media`, [created at build time](https://github.com/mirumee/saleor/blob/master/Dockerfile#L64)

As I wanted to develop locally with the mounted app directory, I tried creating them by hand, but the files are created and owned by my own Linux user, with `uid` `1000`, and the Linux user created inside the container is a new user (`saleor`), and as it is created as a system user, it has `uid` `999`, so it cannot create anything inside those directories.

I tried [commenting out the lines to create a custom user inside the container instead of the default `root` user](https://github.com/mirumee/saleor/blob/master/Dockerfile#L63).

But then I discovered that it also expects:

* A file `/app/webpack-bundle.json` [copied from the previous Docker build stage, created at build time](https://github.com/mirumee/saleor/blob/master/Dockerfile#L54)

...and probably some other files in subsequent steps.

## I want to merge this change because...

With the current layout of the source files and architecture, it's not feasible to mount the local app directory for development. If it is wanted / required, while a new layout is defined, this would simplify testing Saleor locally with Docker.

## Relevant issues

Directly causes: #3049
I can't say that it causes #2642 , but after adding the directories manually, it causes the problems described in these subsequent comments: https://github.com/mirumee/saleor/issues/2642#issuecomment-414584455 https://github.com/mirumee/saleor/issues/2642#issuecomment-419433915 
https://github.com/mirumee/saleor/issues/2642#issuecomment-427600603

Then, after adding the directories manually and removing the custom user or having any other workaround, it causes: #2506

### Pull Request Checklist

1. [n/a] Privileged views and APIs are guarded by proper permission checks.
1. [n/a] All visible strings are translated with proper context.
1. [n/a] All data-formatting is locale-aware (dates, numbers, and so on).
1. [n/a] Database queries are optimized and the number of queries is constant.
1. [n/a] The code is documented (docstrings, project documentation).
1. [n/a] GraphQL schema and type definitions are up to date.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,8 +13,6 @@ services:
       - saleor-backend-tier
     ports:
       - 8000:8000
-    volumes:
-      - .:/app:Z
     env_file: common.env
     depends_on:
       - db
--- a/docs/customization/docker.rst
+++ b/docs/customization/docker.rst
@@ -50,6 +50,4 @@ Usage
     $ docker-compose up
 
 
-By default, the application is started in debug mode, will automatically reload code and is configured to listen on port ``8000``.
-
-
+By default, the application is started in debug mode and is configured to listen on port ``8000``.


patrys:
I agree that we should drop this, we should also move the static assets outside of the saleor directory and then re-evaluate which parts should be shadowed by local directories and which should not.

artursmet:
I agree @patrys, but maybe we could merge this one first and then work on moing assets outside of saleor directory?

patrys:
Yes, they should be separate changes.

ajostergaard:
Would it make sense to include removing ", will automatically reload code" from https://saleor.readthedocs.io/en/latest/customization/docker.html as part of this change.

tiangolo:
Good point @ajostergaard :+1: . 

I just updated it.

tiangolo:
:tada:

PR #3093, opened 2018-10-10
Docker compose override

tiangolo:
Include a `docker-compose.override.yml` file for the local development overrides as it is [read by default by Docker Compose](https://docs.docker.com/compose/extends/). 

This allows having the same configuration and current development experience, but the original `docker-compose.yml` file can be kept without exposing the ports, for example, to be used during deployments.

### Pull Request Checklist

1. [n/a] Privileged views and APIs are guarded by proper permission checks.
1. [n/a] All visible strings are translated with proper context.
1. [n/a] All data-formatting is locale-aware (dates, numbers, and so on).
1. [n/a] Database queries are optimized and the number of queries is constant.
1. [n/a] The code is documented (docstrings, project documentation).
1. [n/a] GraphQL schema and type definitions are up to date.

### DIFF
--- a/docs/customization/docker.rst
+++ b/docs/customization/docker.rst
@@ -18,7 +18,7 @@ You will need to install Docker and `docker-compose <https://docs.docker.com/com
 
 .. note::
 
-   Our configuration uses a ``docker-compose.override.yml`` that exposes PostgreSQL, Redis and Elasticsearch ports. If you don't want to expose the ports you can tell Docker Compose to do not include the additional configurations (ports) in the ``docker-compose.override.yml`` file by specifying ``docker-compose.yml`` with the ``-f`` option, as in ``docker-compose -f docker-compose.yml up -d``.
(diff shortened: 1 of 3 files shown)


docs/customization/docker.rst
```diff
@@ -18,7 +18,7 @@ You will need to install Docker and `docker-compose <https://docs.docker.com/com
 
 .. note::
 
-   Our configuration exposes PostgreSQL, Redis and Elasticsearch ports. If you have problems running this docker file because of port conflicts, you can remove ``ports`` section from ``docker-compose.yml``.
+   Our configuration uses a ``docker-compose.override.yml`` that exposes PostgreSQL, Redis and Elasticsearch ports. If you don't want to expose the ports you can tell Docker Compose to do not include the additional configurations (ports) in the ``docker-compose.override.yml`` file by specifying ``docker-compose.yml`` with the ``-f`` option, as in ``docker-compose -f docker-compose.yml up -d``.
```
NyanKiyoshi: > you can tell Docker Compose **to do not** include the additional configurations

I think you should change it to "not to include".

tiangolo: Thanks! Done.

maarcingebala:
Please see PR #3657 which was partially based on your work. Thanks!

tiangolo:
Thanks!

PR #3095, opened 2018-10-11
Perm enum

### LINKED ISSUE #2986: Use enum as permission code
Code transparency and type safety.

Return `enum` instead of `string` as permission code.

the-bionic:
resolves #2986

### Pull Request Checklist


tests/api/test_shop.py
```diff
@@ -12,7 +13,7 @@ def test_query_authorization_keys(
     query = """
     query {
         shop {
-            authorizationKeys {
+            s {
```
Pacu2: Not sure why this change is needed?

Pacu2:
Could you please fix codeclimate issues?

the-bionic:
Yes, I am fixing that.

Pacu2:
The ReactJS part needs adjustment, are you able to handle that or shall we?

the-bionic:
Alright. I'll handle it.

the-bionic:
It just occurred to me that this is going to need some reworking as well:

``` python
def format_permissions_for_display(permissions):
    """Transform permissions queryset into PermissionDisplay list.
    Keyword arguments:
    permissions - queryset with permissions
    """
    return [
        PermissionDisplay(
            code='.'.join(
                [permission.content_type.app_label, permission.codename]), name=permission.name) 
                for permission in permissions]
```
to return an enum so the frontend can perform its permission checks. Right now it is comparing ```string``` to ```PermissionEnum```, which is crashing it.

Here is what I came up with:
``` python
formatted_permissions = []
    for permission in permissions:
        codename = '.'.join(
            [permission.content_type.app_label, permission.codename])
        formatted_permissions.append(
            PermissionDisplay(
                code=graphene.Enum(
                    'PermissionEnum', str_to_enum(codename), codename),
                name=permission.name))
    return formatted_permissions
```

saleor/graphql/core/types/common.py
```diff
@@ -59,8 +61,13 @@ class LanguageDisplay(graphene.ObjectType):
     language = graphene.String(description='Language.', required=True)
 
 
+PermissionEnum = graphene.Enum(
+    'PermissionEnum', [(str_to_enum(codename), codename)
```
Pacu2: I think we should split it by the dot here, as we did in `split_permission_codename` function, so the final permission will be
`MANAGE_USERS` 
instead of 
`ACCOUNT_MANAGE_USERS`

the-bionic: I'll resolve that.

the-bionic:
``` python
def format_permissions_for_display(permissions):
    """Transform permissions queryset into PermissionDisplay list.
    Keyword arguments:
    permissions - queryset with permissions
    """
    return [
        PermissionDisplay(
            code='.'.join(
                [permission.content_type.app_label, permission.codename]),
name=permission.name) for permission in permissions]
```
This is going to need some modification. Because the frontend is going to be making its checks based on the ```PermissionEnum```, and not on ```strings```.

I was doing this:
``` python
formatted_permissions = []
    for permission in permissions:
        codename = '.'.join(
            [permission.content_type.app_label, permission.codename])
        formatted_permissions.append(
            PermissionDisplay(
                code=graphene.Enum(
                    'PermissionEnum', str_to_enum(codename), codename),
                name=permission.name))
    return formatted_permissions
```
What do you think?

Pacu2:
@the-bionic If the result is as expected, I'm fine with that

Pacu2:
Or actually, you can just use `getattr` and existing `PermissionEnum` 
Like `code=getattr(PermissionEnum, codename)`

the-bionic:
I used ```Permission.get(codename)``` which does about the same thing.

the-bionic:
Thanks! I am on it.

I have reverted the snapshot I updated to what it was before. I will investigate further and let you know you know if I need help.

the-bionic:
```npm run storybook``` runs without errors, but 2 tests are still failing though.

**Comment by Pacu2** [comment]
@the-bionic In such case, please take a look if stories in storybook are rendered properly, especially configuration page stories.

**Comment by the-bionic** [comment]
@Pacu2 it renders okay (I am not sure how it would look when it doesn't.) After inspecting the differences in the snapshot, it turns out
```
<input
                              class="MuiSwitchBase-input-id"
    -                         name="account.impersonate_users"
    +                         name="IMPERSONATE_USERS"
                              type="checkbox"
                            />
```

That is happening in an input tag. That is what is breaking the test.

Pacu2:
@the-bionic I think that's a correct behavior, as we've changed strings to Enums, we should adjust those tests and are good to go.

the-bionic:
It can be inspected here, too:
https://travis-ci.org/mirumee/saleor/jobs/440580916

the-bionic:
Great. I'll update the snapshot. 

You can also see here for the full details: https://travis-ci.org/mirumee/saleor/jobs/440580916

Pacu2:
@dominik-zeglen we good?

PR #3101, opened 2018-10-12
Add first and last name in user

fowczarek:
want to merge this change because resolve https://github.com/mirumee/saleor/issues/2324 

### Pull Request Checklist


NyanKiyoshi:
What's the reason(s) behind those changes?

saleor/account/models.py
```diff
@@ -139,6 +141,15 @@ class Meta:
                     'Permission description', 'Impersonate customers.')))
 
     def get_full_name(self):
+        if self.first_name or self.last_name:
+            not_empty_names = filter(None, [self.first_name, self.last_name])
```
maarcingebala: I was wondering if we could simplify the logic here and return the "full name" only if we have both first name and last name (the same in case of getting them from the address). Otherwise we would return just an email. E.g. if we had only "John" or just "Doe" somewhere in the UI as the user representation - does it provide any useful value? In that case, "john.doe@example.com" feels more valuable.

maarcingebala: Not related to my previous comment, but generally we rather don't use `filter` in our Python code. Here we have only (and always) two list items to check so we might as well write something like: `('%s %s' % (first_name, last_name)).strip()` to get the same result. In case a list contains more elements, a more idiomatic way to do the filtering is by using list comprehensions.

saleor/account/views.py
```diff
@@ -96,6 +98,15 @@ def get_or_process_password_form(request):
     return form
 
 
+def get_or_process_name_form(request):
+    form = NameForm(data=request.POST or None, instance=request.user)
+    if form.is_valid():
+        form.save()
+        messages.success(request, pgettext(
+            'Storefront message', 'Data successfully changed.'))
```
maarcingebala: We know what data was changed and we could give a more user-friendly feedback. What do you think about something like: "Account successfully updated"?

templates/dashboard/customer/list.html
```diff
@@ -52,8 +52,8 @@
                   <tr data-action-go="{% url "dashboard:customer-details" pk=customer.pk %}">
                     {% with customer.default_billing_address as default_address %}
                       <td>
-                        {% if default_address %}
-                          {{ default_address.first_name }} {{ default_address.last_name }}
+                        {% if customer.get_full_name != customer.email %}
```
maarcingebala: I'm not sure this check is necessary since `get_full_name` defaults to `email` if the full name wasn't assembled properly.

maarcingebala:
This PR looks really good but I have a few more general notes.
- We're missing the migration file, as there were changes in the `User` model. This is also the reason why the tests failed on CI.
- Since we added a new form and put some logic into `get_full_name` function, it is a good idea to add some tests for that as well. I'm thinking of a simple test for `get_full_name` that checks that the full name is assembled properly depending on data set in newly added `User` fields or in the billing address. We could also add a test for the `account` view, where we would submit some `POST` data required by this view and assert that after successful response, the user instance was updated as expected.

tests/test_account.py
```diff
@@ -285,3 +285,24 @@ def test_view_account_delete_confirm(customer_user, authorized_client):
     assert response.status_code == 302
     customer_user = User.objects.filter(pk=customer_user.pk).first()
     assert customer_user is None
+
+
+def test_add_names_to_user(customer_user):
```
Pacu2: I think we are testing form here, so the test name should be slightly different

tests/test_account.py
```diff
@@ -285,3 +285,24 @@ def test_view_account_delete_confirm(customer_user, authorized_client):
...
+        {'first_name': 'Jan', 'last_name': 'Nowak'},
+        instance=customer_user)
+    name_form.is_valid()
+    name_form.save()
+    updated_user = User.objects.get(pk=customer_user.pk)
+    assert updated_user.first_name == 'Jan'
+    assert updated_user.last_name == 'Nowak'
+
+
+def test_add_names_to_user_post(customer_user, authorized_client):
```
Pacu2: I think we could change the convention to `test_view_functionname` or even `test_functionname`

tests/dashboard/test_customer.py
```diff
@@ -34,6 +35,33 @@ def test_customer_create(admin_client):
...
+
+def test_add_customer_form():
+    user_count = User.objects.count()
+    customer_form = CustomerForm({'email': 'customer01@example.com'})
+    customer_form.is_valid()
+    customer_form.save()
+    assert User.objects.count() == user_count + 1
+
+
+def test_add_names_to_customer(customer_user):
```
Pacu2: I think we are testing edit, instead of adding here.

saleor/account/models.py
fowczarek: What if a person has only one name?

templates/dashboard/customer/list.html
fowczarek: If we don't check this then in some case email appears in 2 column in the customer list.

tests/dashboard/test_customer.py
```diff
@@ -34,6 +35,33 @@ def test_customer_create(admin_client):
...
+
+def test_add_customer_form():
+    user_count = User.objects.count()
+    customer_form = CustomerForm({'email': 'customer01@example.com'})
+    customer_form.is_valid()
+    customer_form.save()
+    assert User.objects.count() == user_count + 1
+
+
+def test_edit_customer_names(customer_user):
```
Pacu2: Test edit_customer_form?

saleor/search/documents.py
```diff
@@ -67,12 +71,34 @@ class Meta:
...
 
     def prepare_user(self, instance):
         if instance.user:
             return instance.user.email
         return instance.user_email
 
+    def prepare_first_name(self, instance):
+        user = instance.user
+        if user:
+            if user.first_name:
```
Pacu2: It looks like a copy of the `prepare_first_name` function from the above,  can we move it into the util, replace in both functions and test? Same with the function below

NyanKiyoshi: In addition to what @Pacu2 suggested, could you add a little docstring summarizing what the method does so we keep the code clean?

saleor/search/documents.py
```diff
@@ -67,12 +62,20 @@ class Meta:
...
+    first_name = fields.StringField()
+    last_name = fields.StringField()
 
     def prepare_user(self, instance):
         if instance.user:
             return instance.user.email
         return instance.user_email
 
+    def prepare_first_name(self, instance):
+        get_user_first_name(instance.user)
```
Pacu2: I think you should return it here

saleor/search/documents.py
```diff
@@ -67,12 +62,20 @@ class Meta:
...
+
+    def prepare_last_name(self, instance):
+        get_user_last_name(instance.user)
```
Pacu2: Same here

tests/test_account.py
```diff
@@ -213,6 +213,77 @@ def test_user_ajax_label_without_address(admin_user):
...
+
+@pytest.mark.parametrize(
+    "first_name, default_billing_address_first_name, result", [
+        ('John', 'Arnold', 'John'),
+        ('John', '', 'John'),
+        ('', 'Arnold', 'Arnold'),
+        ('', '', '')])
+def test_get_user_first_name(first_name, default_billing_address_first_name,
+                             result, address):
+    copied_address = address.get_copy()
```
Pacu2: Do we need to copy it?

saleor/dashboard/customer/forms.py
```diff
@@ -4,6 +4,13 @@
 from ...account.models import CustomerNote, User
 
 
+def get_name_placeholder(name):
+    return pgettext_lazy(
+        'Customer form: Name field placeholder',
+        '%(name)s (Inherit from default biling address)') % {
```
Pacu2: Typo, should be `billing`

PR #3106, opened 2018-10-12
Add createdb note to getting started for Windows

### LINKED ISSUE #2680: Update Windows installation guide
In windows installation guide we've missed a point, stating that database should be created. Something like:

```
Create a PostgreSQL database:

Unless configured otherwise the store will use saleor as the database name.

   $ createdb saleor
```

However, this information is present for both Mac and Linux

ajostergaard:
I want to merge this change because it fixes #2680 

### Pull Request Checklist

### DIFF
--- a/docs/gettingstarted/installation-windows.rst
+++ b/docs/gettingstarted/installation-windows.rst
@@ -104,7 +104,7 @@ All commands need to be performed in either PowerShell or a Command Shell.
 
    .. note::
 
-       Database name is extracted from the ``DATABASE_URL`` environmental variable. If absent it defaults to ``saleor``.
+       Database name is extracted from the ``DATABASE_URL`` environment variable. If absent it defaults to ``saleor``.
 
 #. Prepare the database:
 
--- a/docs/gettingstarted/installation-windows.rst
+++ b/docs/gettingstarted/installation-windows.rst
@@ -98,6 +98,14 @@ All commands need to be performed in either PowerShell or a Command Shell.
 
    Unless configured otherwise the store will use ``saleor`` as both username and password. Remeber to give your user the ``SUPERUSER`` privilege so it can create databases and database extensions.
 
+#. Create a PostgreSQL database
+
+   See `PostgreSQL's createdb command <https://www.postgresql.org/docs/current/static/app-createdb.html>`_ for details.
+
+   .. note::
+
+       Database name is extracted from the ``DATABASE_URL`` environmental variable. If absent it defaults to ``saleor``.
+
 #. Prepare the database:
 
    .. code-block:: powershell


docs/gettingstarted/installation-windows.rst
```diff
@@ -98,6 +98,14 @@ All commands need to be performed in either PowerShell or a Command Shell.
 
    Unless configured otherwise the store will use ``saleor`` as both username and password. Remeber to give your user the ``SUPERUSER`` privilege so it can create databases and database extensions.
 
+#. Create a PostgreSQL database
+
+   See `PostgreSQL's createdb command <https://www.postgresql.org/docs/current/static/app-createdb.html>`_ for details.
+
+   .. note::
+
+       Database name is extracted from the ``DATABASE_URL`` environmental variable. If absent it defaults to ``saleor``.
```
NyanKiyoshi: > **The** database name

> the ``DATABASE_URL`` environmental variable

It's an **environment variable**, it's not an ecological variable.

---

Also, I can't really manage to make much sense to the edits. Why not just just taking or basing your changes on the Linux or MacOS docs?

https://github.com/mirumee/saleor/blob/d9467f967cabe365b50e913e004abd22f8c42a94/docs/gettingstarted/installation-macos.rst#L125-L131

ajostergaard: @NyanKiyoshi the text is copy pasted from the Linux page.

NyanKiyoshi: I really think that just redirecting the user to another documentation, just to very (very) implicitly tell them to do `createdb saleor` is a very bad idea. It will confuse many users and make them give up on the installation has they would have to search the information or simply not get the point.

Also, are the postgres binaries in the Windows' `PATH`? Or does the user has to set it/ tell Windows where to look for `createdb`?

ajostergaard: I don't disagree but this patches a significantly worse issue with the docs and does it quickly without causing problems.

ajostergaard:
@patrys given your comment [here](https://github.com/mirumee/saleor/issues/2680#issuecomment-429843353) can we agree that the docs are correct with this PR. Not at all optimal - certainly scope for improvement - but as they are identical to the Linux instructions and are completely correct can we at least accept this PR and then discuss improvements in new issues. This PR is merely to plug a gaping hole in the Windows docs. :)

patrys:
Yes but in that case I'd drop the entire thing about the `SUPERUSER`. We can move it to the documentation on testing.

ajostergaard:
@patrys digging a little deeper, looks to me like additional permissions may be required during migration as that is when the PostgreSQL extensions are enabled. According to Django docs this requires SUPERUSER:

> Creating the extension requires a database user with superuser privileges. If the Django database user doesn’t have superuser privileges, you’ll have to create the extension outside of Django migrations with a user that has the appropriate privileges. In that case, connect to your Django database and run the query CREATE EXTENSION IF NOT EXISTS hstore;.

PR #3111, opened 2018-10-13
Build a docker image with dev dependencies as optional

stephenmoloney:
What does this commit/MR/PR do?

- Build a docker image with dev dependencies as optional

Why is this commit/MR/PR needed?

- Avoid forcing the inclusion of dev dependencies

### Pull Request Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,7 @@ jobs:
           command: |
             docker build \
               --build-arg STATIC_URL=$STATIC_URL \
+              --build-arg MODE=prod \
               -t mirumee/saleor:latest .
       - run:
           name: Run tests
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,16 @@
 ### Build and install packages
 FROM python:3.6 as build-python
 
+ARG MODE
+ENV MODE ${MODE:-dev}
+
+RUN \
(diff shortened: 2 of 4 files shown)


ajostergaard:
@stephenmoloney any chance of documenting this new feature?

stephenmoloney:
@ajostergaard 

If this PR is accepted, I will add documentation, sure. 

I'm not sure either this is the best approach.

The alternative is to create a separate `Dockerfile.prod` for production images without `pipenv`

Relates to https://github.com/mirumee/saleor/issues/3110

I do think there should be a production ready - production only docker image though.

patrys:
I think the only reason we've added `--dev` was because `silk` and `django-debug-toolbar` were placed in the wrong section of the `Pipfile`. We do not need to install dev tools inside the Docker container so the non-production mode is not really needed, we should instead make sure that all actual dependencies are in the default dependencies section and not in the dev dependencies section.

stephenmoloney:
ok, so if i add `silk and django-debug-toolbar` to the default deps and remove `--dev`, is that sufficient ?

I can close this PR and open a new one with those changes.

patrys:
@stephenmoloney Yes, move those (and any other conditionally enabled deps if you find any) and remove `--dev` from the install command in Dockerfile.

stephenmoloney:
Ref #3117

PR #3115, opened 2018-10-15
Escape IDs

### LINKED ISSUE #2657: Escape IDs in urls
To fix a potential router crash when objects ID contains slash (which could happen, since base64 allows it), matching as ID only its part.

1. Find an object, whose ID contains the slash character
2. Go to its details page
3. Contemplate the error

### What I expected to happen
To display the details page instead of an error.

### Proposed solution
Use `encodeURIComponent()` where applicable.

dominik-zeglen:
Resolves #2657

### DIFF
--- a/saleor/static/dashboard-next/orders/index.tsx
+++ b/saleor/static/dashboard-next/orders/index.tsx
@@ -13,8 +13,8 @@ const OrderList: React.StatelessComponent<RouteComponentProps<any>> = ({
 }) => {
   const qs = parseQs(location.search.substr(1));
   const params = {
-    after: qs.after,
-    before: qs.before
+    after: decodeURIComponent(qs.after),
+    before: decodeURIComponent(qs.before)
   };
   return <OrderListComponent params={params} />;
 };
@@ -22,7 +22,7 @@ const OrderList: React.StatelessComponent<RouteComponentProps<any>> = ({
 const OrderDetails: React.StatelessComponent<RouteComponentProps<any>> = ({
   match
 }) => {
-  return <OrderDetailsComponent id={match.params.id} />;
+  return <OrderDetailsComponent id={decodeURIComponent(match.params.id)} />;
 };
 
 const Component = ({ match }) => (
--- a/saleor/static/dashboard-next/orders/index.tsx
+++ b/saleor/static/dashboard-next/orders/index.tsx
@@ -13,8 +13,8 @@ const OrderList: React.StatelessComponent<RouteComponentProps<any>> = ({
 }) => {
   const qs = parseQs(location.search.substr(1));
   const params = {
-    after: decodeURIComponent(qs.after),
-    before: decodeURIComponent(qs.before)
+    after: qs.after,
+    before: qs.before
   };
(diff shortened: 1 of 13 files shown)


saleor/static/dashboard-next/categories/index.tsx
```diff
@@ -12,20 +12,25 @@ const CategoryPaginator: React.StatelessComponent<RouteComponentProps<any>> = ({
 }) => {
   const qs = parseQs(location.search.substr(1));
   const params = {
-    after: qs.after,
-    before: qs.before
+    after: decodeURIComponent(qs.after),
```
patrys: Shouldn't this be handled by `qs.parse()`?

saleor/static/dashboard-next/orders/index.tsx
```diff
@@ -22,7 +22,7 @@ const OrderList: React.StatelessComponent<RouteComponentProps<any>> = ({
 const OrderDetails: React.StatelessComponent<RouteComponentProps<any>> = ({
   match
 }) => {
-  return <OrderDetailsComponent id={match.params.id} />;
+  return <OrderDetailsComponent id={decodeURIComponent(match.params.id)} />;
```
patrys: How is this not decoded already?

saleor/static/dashboard-next/orders/views/OrderDetails.tsx
```diff
@@ -558,7 +560,9 @@ export const OrderDetails: React.StatelessComponent<OrderDetailsProps> = ({
                                     label: country.country
                                   }))}
                                   onProductClick={id => () =>
-                                    navigate(productUrl(id))}
+                                    navigate(
+                                      productUrl(encodeURIComponent(id))
```
patrys: The `productUrl` function should be responsible for that.

saleor/static/dashboard-next/orders/views/OrderList.tsx
```diff
@@ -30,7 +30,9 @@ export const OrderList: React.StatelessComponent<OrderListProps> = ({
             pushMessage({
               text: i18n.t("Order draft succesfully created")
             });
-            navigate(orderUrl(data.draftOrderCreate.order.id));
+            navigate(
+              orderUrl(encodeURIComponent(data.draftOrderCreate.order.id))
```
patrys: The `orderUrl` function should be responsible for that.

saleor/static/dashboard-next/productTypes/index.tsx
```diff
@@ -13,14 +13,10 @@ export const productTypeListUrl = "/productTypes";
 const ProductTypeList: React.StatelessComponent<RouteComponentProps<{}>> = ({
   location
 }) => {
-  const queryString = parseQs(location.search.substr(1));
+  const qs = parseQs(location.search.substr(1));
   const params = {
```
patrys: How about:

```
const { after, before } = qs;

...

return <ProductTypeListComponent params={{ after, before }} />;
```

And if we're here, why is there `params` instead of explicitly accepting `after` and `before` as props?

Issue #2778, opened 2018-09-03
Dashboard 2.0: Implement home widgets


patrys:
We need to define which widgets are currently implementable and add them to the home screen.

maarcingebala:
@pwgryglak We'll need a new homepage mockup that will include widgets that we'll agree on here.

pwgryglak:
Information that we had from meeting with our users (Merchandising and Customer Care) regarding dashboard:

Probably a task list elements with shortcuts to problematic occurences:
Out of Stock Products,
Orders problems
Uncaptured Payments

Analytics information:
Top Sellers
Sales Today 
Orders Today 
Visits Today
Most Clicked Products
all of them with ability to choose time range

maarcingebala:
We agreed on new components that we're able to provide and we got the mockups on Figma. @Bonifacy1 is to implement the components, I'll think of API changes required to implement those components and will implement them in a separate PR.

PR #3146, opened 2018-10-19
Add queries for dashboard homepage

maarcingebala:
This PR adds queries necessary to render the new dashboard's homepage. I also made one unrelated change (if you don't like it I can revert it), which is extracting product related queries and mutations from the `api.py` file into dedicated `products/schema.py`. I just wanted to check how would that work and it seems promising - we could eventually refactor the whole API to follow that pattern.

Related: #2778 

### Pull Request Checklist


saleor/graphql/order/resolvers.py
```diff
@@ -38,3 +58,11 @@ def resolve_shipping_methods(obj, info):
     return qs.applicable_shipping_methods(
         price=obj.get_subtotal().gross.amount, weight=obj.get_total_weight(),
         country_code=obj.shipping_address.country.code)
+
+
+def resolve_order_events(info):
+    # Filter only selected events to be displayed on homepage.
+    types = [
```
akjanik: Do we want types to be hardcoded here? If yes, then maybe we should rename resolver to something more suggestive, for example `resolve_placed_orders`.

saleor/graphql/product/types.py
```diff
@@ -154,6 +160,11 @@ def resolve_price(self, info):
     def resolve_price_override(self, info):
         return self.price_override
 
+    def resolve_quantity_ordered(self, info):
+        # This field is added through annotation when using the
+        # `resolve_top_products` resolver.
```
akjanik: I couldn't find this one.

saleor/order/models.py
```diff
@@ -29,18 +29,34 @@
 
 class OrderQueryset(models.QuerySet):
     def confirmed(self):
+        """Return draft orders."""
```
akjanik: Did you mean `non-draft`?

tests/api/test_order.py
```diff
@@ -137,7 +139,27 @@ def test_order_query(
     assert expected_method.type.upper() == method['type']
 
 
-def test_order_events_query(
+def test_order_status_filter_param(user_api_client):
```
akjanik: Please add some asserts/checks to the test. I think we should have got fixtures for `ready_to_capture/fulfill` orders, if not, we can always create them.

tests/api/test_order.py
```diff
@@ -186,6 +208,24 @@ def test_order_events_query(
...
+                    date
+                    type
+                }
+            }
+        }
+    }
+    """
+    response = staff_api_client.post_graphql(
+        query, permissions=[permission_manage_orders])
+    get_graphql_content(response)
```
akjanik: Again, content may be received, but does it returns what was intended?

saleor/graphql/order/resolvers.py
maarcingebala: Those events will be hardcoded for now. Maybe I should rename it to `resolve_homepage_activity` as it is supposed to return some (not all) events suitable for dashboard homepage. Later this would include also product- or user-related events.

akjanik: That is fine. I would just leave a comment explaining why and for what this resolver is, so if some dev would fork project, he would easily find a place where he can customize his own dashboard activity (or remove, if not needed).

PR #3148, opened 2018-10-20
Return default placeholder when no image

### LINKED ISSUE #3120: Getting error when category has no background image
1. Create category with no background image
2. Exec query
```
{
  category {
    backgroundImage {
      url
    }
  }
}
```
3. Get an error
```
{
  "errors": [
    {
      "message": "The 'background_image' attribute has no file associated with it.",
      "locations": [
        {
          "line": 6,
          "column": 7
        }
      ],
      "path": [
        "categories",
        "edges",
        1,
        "node",
        "backgroundImage",
        "url"
      ]
    }
  ]
}
```

### What I expected to happen
To successfully fetch category data.

bogdal:
I want to merge this change because it fixes #3120

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/types.py
+++ b/saleor/graphql/product/types.py
@@ -178,7 +178,7 @@ class Image(graphene.ObjectType):
         description = 'Represents an image.'
 
     def resolve_url(self, info, size=None):
-        if size:
+        if size or not self:
             return get_thumbnail(self, size, method='thumbnail')
         return self.url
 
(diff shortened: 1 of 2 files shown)


Pacu2:
I'm not sure what's the tactic for the new storefront, but currently, we are not displaying placeholders for missing category/collection images, just hiding them instead.

maarcingebala:
I agree with @Pacu2 here, we should rather return `null` when an image is not set. It should be within the client app responsibility to handle that case properly e.g. by rendering a placeholder image. I'll open a separate PR with a fix for this issue. Thank you for your effort @bogdal, I'll be able to use the tests you wrote in my PR.

patrys:
Agreed, placeholders (and how they look) are within the branding domain so the client should be responsible for picking an image (or not displaying one at all).

PR #3183, opened 2018-10-26
Add validation error code to API

dominik-zeglen:
I want to merge this change because it adds error codes to API.

Example: 
```
mutation {
  categoryUpdate(id:"Q2F0ZWdvcnk6Ng==", input:{slug:"a b"}) {
    errors{
      field
      message
      code
    }
  }
}
```
outputs
```
{
  "data": {
    "categoryUpdate": {
      "errors": [
        {
          "field": "slug",
          "message": "Wpisz poprawną uproszczoną nazwę zawierającą jedynie litery, cyfry, podkreślenia i myślniki.",
          "code": "invalid"
        }
      ]
    }
  }
}
```

TODO:
1. [ ] Use enum as `code` instead of string
2. [ ] Ensure that all custom errors in various mutations have `code` property 

It allows front end to format its own messages instead of relying on API output.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/core/mutations.py
+++ b/saleor/graphql/core/mutations.py
@@ -57,7 +57,7 @@ class BaseMutation(graphene.Mutation):
         cls._meta.fields.update(fields)
 
     @classmethod
-    def add_error(cls, errors, field, message):
+    def add_error(cls, errors, field, message, code=None):
         """Add a mutation user error.
 
         `errors` is the list of errors that happened during the execution of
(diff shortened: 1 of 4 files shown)


maarcingebala:
I like the idea and would love to see that implemented in entire API. I agree that we should use enums for codes, but I'm also fine with introducing changes gradually :)

maarcingebala:
Adding error codes to API is something that we have to definitly work on, but I'd do that in a new PR. There is a PR (#3891 ) that reworks error handling in API and it will be much easier to add statuses once it's merged.

PR #3186, opened 2018-10-27
A fix and refactor for docker host mounted volumes

bennylb:
This directly addresses PR #3091 which removed bind mounts as a result of file permission/ownership issues and some assets that were being created during image build time only (see #3091). These changes remove `Dockerfile` `RUN` instructions that created assets during image build and moves them to the `dev-example.Makefile` to instead be used at runtime along with [gosu](https://github.com/tianon/gosu). The additional `docker-compose.dev.yml` can be reduced and merged together with PR #3093 and the base `docker-compose.yml` in future to provide the first example use case described in the [docker documentation - Share Compose configurations](https://docs.docker.com/compose/extends/).

Most importantly these additions:

- Mean that changes made on the host are imediately reflected in the container (no need to rebuild image)
- Form a basis to handle file ownership on the host and container given the host `uid` and `gid` with [gosu](https://github.com/tianon/gosu)
- Have no side effects and are completely optional

If the description needs expanding or any changes/additions need to be made feedback is much appreciated. :)

### Relevant issues/PRs

- #3091 
- #3093 

### Pull Request Checklist

1. [NA] Privileged views and APIs are guarded by proper permission checks.
1. [NA] All visible strings are translated with proper context.
1. [NA] All data-formatting is locale-aware (dates, numbers, and so on).
1. [NA] Database queries are optimized and the number of queries is constant.
1. [NA] The code is documented (docstrings, project documentation).
1. [NA] GraphQL schema and type definitions are up to date.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -4,7 +4,7 @@ services:
   web:
     build:
       context: .
-      dockerfile: ./Dockerfile.dev
+      dockerfile: ./dev.Dockerfile
       args:
         STATIC_URL: '/static/'
         HOST_UID:
@@ -50,7 +50,7 @@ services:
   celery:
     build:
       context: .
-      dockerfile: ./Dockerfile.dev
+      dockerfile: ./dev.Dockerfile
       args:
         STATIC_URL: '/static/'
         HOST_UID:
--- a/dev-example.Makefile
+++ b/dev-example.Makefile
@@ -1,7 +1,7 @@
 # A quick and conveniant way to use this Makefile is by creating a
 # make shell alias like the following:
 #
-# alias saleor='make -f Makefile.dev.example'
+# alias saleor='make -f dev-example.Makefile'
 #
 # then use it as you would make:
 #


Dockerfile.dev
```diff
@@ -0,0 +1,64 @@
+### Build and install packages
+FROM python:3.6 as build-python
+
+ARG STATIC_URL
+
```
stephenmoloney: perhaps removes space here

Dockerfile.dev
```diff
@@ -0,0 +1,64 @@
...
+ARG SKIP_LOCK
+
```
stephenmoloney: perhaps removes space here

Dockerfile.dev
```diff
@@ -0,0 +1,64 @@
...
+FROM python:3.6-slim
+
+ARG STATIC_URL
+ENV STATIC_URL ${STATIC_URL:-/static/}
+
+ARG SKIP_LOCK
+
+ARG HOST_UID
+ARG HOST_GID
+ENV APP_USER=${HOST_UID:-saleor}
```
stephenmoloney: The fallback here is `saleor` which is a username and not a uid.
I thought a uid was a number like `1001` ?
Not 100% myself but please clarify.

Makefile.dev.example
```diff
@@ -0,0 +1,66 @@
...
+# alias saleor='make -f Makefile.dev.exmaple'
+#
+# then use it as you would make:
+#
+# saleor <target>
+
+# SKIP_LOCK=
+SKIP_LOCK=--skip-lock
+HOST_UID=1000
+HOST_GID=100
```
stephenmoloney: why 100 ?

Dockerfile.dev
stephenmoloney: I'd like to suggest the following slight change:

-  remove APP_USER as is not needed I think but HOST_UID is needed. I am open to correction on this.
- explicit command line arguments for `useradd`, it's easier to read when going through the `Dockerfile` to figure out what is going on
- Provide fallbacks for HOST_UID and HOST_GID, user can provide his/her own though. Don't depend on a makefile exclusively. Not everyone uses makefiles.

```bash
ARG HOST_UID
ARG HOST_GID
ENV HOST_UID=${HOST_UID:-1001} \
    HOST_GID=${HOST_GID:-1001}

RUN useradd --non-unique --uid ${HOST_UID}  --gid  ${HOST_GID} --create-home saleor && \
    install -d -o ${HOST_UID} -g ${HOST_GID} /app/node_modules && \
    install -d -o ${HOST_UID} -g ${HOST_GID} /app/media && \
    install -d -o ${HOST_UID} -g ${HOST_GID} /app/static
```

Dockerfile.dev
bennylb: Sure, we can remove the spaces.

Dockerfile.dev
bennylb: Yes APP_USER is redundant so that make more sense. A fallback uid and gid of 1001 can be used but they will have be set explicitly to the host uid and gid for the configuration to work as gosu requires them. That can be done through the Makefile, a shell script, common.env, docker-compose.dev.yml or on the command line with --build-args=

Makefile.dev.example
```diff
@@ -0,0 +1,66 @@
...
+# alias saleor='make -f Makefile.dev.example'
+#
+# then use it as you would make:
+#
+# saleor <target>
+
+# --build-arg=SKIP_LOCK=$(SKIP_LOCK) can be added to the 
+# image-build target below to build with latest pipfile.
+# SKIP_LOCK=--skip-lock
+HOST_UID=1000 # Change this to your systems userid
```
stephenmoloney: @bennylb 

How about this instead?

```
HOST_UID="$(id -u)"
HOST_UID=${HOST_UID:-1000}
HOST_GID="$(id -g)"
HOST_GID=${HOST_GID:-1000}
```

docker-compose.dev.yml
```diff
@@ -0,0 +1,91 @@
+version: '2.3'
+
+services:
+  web:
+    build:
+      context: .
+      dockerfile: ./Dockerfile.dev
+      args:
+        STATIC_URL: '/static/'
```
stephenmoloney: How about this

```
  web:
    build:
      context: .
      dockerfile: ./Dockerfile.dev
      args:
        STATIC_URL: '/static/'
        HOST_UID="$(id -u)"
        HOST_GID="$(id -g)"
```

Note: untested but I think something like this should be ok. Will this then work without the makefile ?
It would be great if this PR worked with and without the makefile.

Dockerfile.dev
```diff
@@ -0,0 +1,57 @@
...
+RUN pip install pipenv
+ADD Pipfile /app/
+ADD Pipfile.lock /app/
+WORKDIR /app
+RUN pipenv install $SKIP_LOCK --system --deploy --dev
+
+
+### Final image
+FROM python:3.6-slim
+ARG STATIC_URL
```
stephenmoloney: This is not really 100% part of this PR but as I see more ENV and ARG variables added, the disorganization grows.

Would you mind putting all the variables together like so:

```
ARG STATIC_URL
ARG SKIP_LOCK
ARG HOST_UID
ARG HOST_GID
ENV STATIC_URL=${STATIC_URL:-/static/} \
  HOST_UID=${HOST_UID:-1000} \
  HOST_GID=${HOST_GID:-HOST_UID} \
  PORT=8000 \
  PYTHONUNBUFFERED=1 \
  PROCESSES 4
```

docker-compose.dev.yml
```diff
@@ -0,0 +1,91 @@
...
+    image: library/redis:latest
+    restart: unless-stopped
+    networks:
+      - saleor-backend-tier
+    volumes:
+      - saleor-redis:/data
+    ports:
+      - 6379:6379
+
+  celery:
```
stephenmoloney: Same here too maybe

```
  celery:
    build:
      context: .
      dockerfile: ./Dockerfile.dev
      args:
        STATIC_URL: '/static/'
        HOST_UID="$(id -u)"
        HOST_GID="$(id -g)"
```

Makefile.dev.example
bennylb: Yeah this is a good idea. I'll make the change.

Dockerfile.dev
```diff
@@ -0,0 +1,57 @@
...
+ENV STATIC_URL ${STATIC_URL:-/static/}
+ARG SKIP_LOCK
```
stephenmoloney: unused variable `SKIP_LOCK`

docker-compose.dev.yml
stephenmoloney: Yea, I just tried the function `id -u` and it won't accept it.

An alternative is 

```
  web:
    build:
      context: .
      dockerfile: ./Dockerfile.dev
      args:
        STATIC_URL: '/static/'
        HOST_UID: ${UID}
        HOST_GID: ${GID}
```

and I think that will work. Gonna confirm it now.

bennylb: Ok by omitting default values args will take on the value of whatever variable has been defined in the environment with the same name.

So:

```
STATIC_URL: '/static/'
HOST_UID:
HOST_GID:
```

```
export HOST_UID=$(id -u) && export HOST_GID=$(id -g) && \
docker-compose --file docker-compose.dev.yml build --no-cache
```

would do the trick.

tiangolo:
Just an idea, I've seen editors and IDEs detect as `Dockerfile` format files named `Dockerfile` or named `*.dockerfile`. 

I have used `Dockerfile.ext` before and it seemed more intuitive to me, but automatic language detection didn't seem to work.

So maybe it could be `dev.dockerfile` instead of `Dockerfile.dev`.

bennylb:
> Just an idea, I've seen editors and IDEs detect as `Dockerfile` format files named `Dockerfile` or named `*.dockerfile`.
> 
> I have used `Dockerfile.ext` before and it seemed more intuitive to me, but automatic language detection didn't seem to work.
> 
> So maybe it could be `dev.dockerfile` instead of `Dockerfile.dev`.

Agreed, the thought had already crossed my mind. I'll make the change. It can easily be changed back if it causes confusion.

bykof:
When will you merge this branch, are there any plans?

patrys:
I don't think there are any current plans to merge this. We would need to discuss it first as it adds a whole lot of moving parts to Saleor and I feel it works around the problem instead of addressing it directly.

stephenmoloney:
This PR allows one to set a uid somewhat dynamically (I say somewhat because one has to rebuild the image if one wants a different uid but a default uid of 1001 is a good default I think)

> @stephenmoloney If you forced the UID to always be 1001 in the useradd command (or used an existing user with the same UID) then the result would be the same?

When setting security contexts for containers in kubernetes, it only knows uids (and not usernames)
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container

I've changed it for developing the `saleor-helm` repo but it would be helpful if this was possible upstream too. Ideally, I'd do the same for the group too.

https://github.com/stephenmoloney/saleor-helm/blob/49d654cac3950fbbc3858b2a03318b0b35ed6262/Dockerfile#L55-L59
image

stephenmoloney:
I can't see the benefit of using a names user with an automatically assigned uid vs just setting the uid arbitrarily.

Openshift (redhat k8s provider) has some documentation on this and the Dockerfile would not be suitable on an openshift instance without further modification.

image

source:
https://docs.okd.io/latest/creating_images/guidelines.html#openshift-specific-guidelines

bennylb:
@patrys I can go ahead and close the PR but before dismissing it completely, do you have an alternative solution for local development with docker that we may not know of?

bennylb:
Another thought. I could just move it to a separate repository so that those in the community that are interested can include it as a submodule if they choose.

bennylb:
For anyone that may find this functionality useful for local development I've moved these files to a separate [repo](https://github.com/bennylb/saleor-docker-local). Please feel free to open a PR to add and expand on the very basic functionality it has currently. :)

onahkenneth:
@patrys See [laradock](https://github.com/laradock/laradock) on how to build docker containers.
The current implementation is not there yet especially with mounted volumes.

patrys:
@onahkenneth see #3657 by @michaljelonek

maarcingebala:
How should we progress with this PR? @patrys @michaljelonek Do you think it's ready to merge or it requires some particular changes?

maarcingebala:
Closing this due to inactivity. Let me know if we need to get back to it.

PR #3189, opened 2018-10-29
Optimize updating variant name

### LINKED ISSUE #3053: Optimize updating variant names when attribute changes
Function `saleor.graphql.product.utils.update_variants_names` is meant to update variant names e.g. when product attributes changes within a product type ( used for example in `productTypeUpdate` mutation). The problem is that if we have a lot of variants it gets terribly slow (I've tested with ~800 variants). We should think if we can optimize this function somehow, as it causes noticable lags in the new dashboard's product type management.

fowczarek:
I want to merge this change because resolve #3053 

### Pull Request Checklist

### DIFF
--- a/saleor/product/tasks.py
+++ b/saleor/product/tasks.py
@@ -5,7 +5,9 @@ from .utils.attributes import get_name_from_attributes
 
 
 def _update_variants_names(instance, saved_attributes):
-    """update saved variants name for products type."""
+    """After change in attribute value name, all product variant names
+    should be updated. Products variant name created from attribute
+    names of variants."""
     initial_attributes = set(instance.variant_attributes.all())
     attributes_changed = initial_attributes.intersection(saved_attributes)
     if not attributes_changed:
--- a/saleor/product/tasks.py
+++ b/saleor/product/tasks.py
@@ -5,6 +5,7 @@ from .utils.attributes import get_name_from_attributes
 
 
 def _update_variants_names(instance, saved_attributes):
+    """update saved variants name for products type."""
     initial_attributes = set(instance.variant_attributes.all())
     attributes_changed = initial_attributes.intersection(saved_attributes)
     if not attributes_changed:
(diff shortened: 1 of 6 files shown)


saleor/product/tasks.py
```diff
@@ -0,0 +1,25 @@
+from celery import shared_task
+
+from .models import ProductVariant
+from .utils.attributes import get_name_from_attributes
+
+
+def _update_variants_names(instance, saved_attributes):
```
Pacu2: Some docstring would be nice

tests/api/test_product.py
```diff
@@ -1240,9 +1239,10 @@ def test_unassign_not_assigned_variant_image(
     assert content['data']['variantImageUnassign']['errors'][0]['field'] == (
         'imageId')
```
Pacu2: I think we are missing empty line here

tests/test_product_attributes.py
```diff
@@ -67,3 +70,18 @@ def test_generate_name_from_values():
 def test_generate_name_from_values_empty():
     name = generate_name_from_values({})
     assert name == ''
+
+
+def test_product_type_update_changes_variant_name(product_type):
+    variant = product_type.variant_attributes.first()
+    variant.name = "test_name"
```
Pacu2: Single quotes

tests/test_product_attributes.py
```diff
@@ -67,3 +70,18 @@ def test_generate_name_from_values():
...
+    variant.save()
+    _update_variants_names(product_type, set([variant]))
```
Pacu2: No need to use set if passing one item

saleor/product/tasks.py
```diff
@@ -5,6 +5,7 @@
 
 
 def _update_variants_names(instance, saved_attributes):
+    """update saved variants name for products type."""
```
Pacu2: I'd like to see an explanation, of how the variant's name is composed as that's not obvious

Pacu2: Also, we should start with capital letter

PR #3194, opened 2018-10-30
Drop Pipfile.lock

maarcingebala:
There is a conflict in dependencies which makes it impossible to generate a new Pipfile.lock:

```
Warning: Your dependencies could not be resolved. You likely have a mismatch in your sub-dependencies.
  You can use $ pipenv install --skip-lock to bypass this mechanism, then run $ pipenv graph to inspect the situation.
  Hint: try $ pipenv lock --pre if it is a pre-release dependency.
Could not find a version that matches elasticsearch<6.0.0,<7.0.0,==5.5.3,>=5.0.0,>=5.4,>=6.0.0
```

As a result our builds on CI are failing now. As a workaround I dropped the lock file for now to fix failling builds. I'd like to fix it properly after incoming October release.

### Pull Request Checklist


patrys:
Won't this break Docker?

maarcingebala:
Pipenv installs deps also if lock file is not present. I also added `--skip-lock` to not create one in Docker (which would fail anyway due to versions conflict).

PR #3202, opened 2018-10-31
Query logged-in user data

### LINKED ISSUE #3185: Query logged in user (self) on GraphQL API
Query the API for information on the currently logged in customer. Currently, only staff are allowed to query information on users. Customers should at least be able to query their own information.

Allow customers to query their own profile information.

This is possible using the `tokenVerify` mutation, but using a mutation to query information isn't convenient. It cannot be used with relay's `QueryRenderer` for example.

fowczarek:
I want to merge this change because resolve #3185 

We added anonymous API client used in tests. We should use this API client in every test which doesn't require the authenticated client. 
Issue #3260 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -152,14 +152,14 @@ class CustomerUpdate(CustomerCreate):
         model = models.User
 
 
-class LoggedCustomerUpdate(CustomerCreate):
+class LoggedUserUpdate(CustomerCreate):
     class Arguments:
         input = UserAddressInput(
             description='Fields required to update logged in customer.',
             required=True)
 
     class Meta:
-        description = 'Updates an existing logged in customer.'
+        description = 'Updates data of the logged in user.'
         exclude = ['password']
         model = models.User
 
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -170,6 +170,13 @@ class LoggedCustomerUpdate(CustomerCreate):
         id = graphene.Node.to_global_id('User', user.id)
         return user.has_perm('account.manage_users') or id == input['id']
 
+    @classmethod
+    def mutate(cls, root, info, **data):
+        if data['id'] == '':
+            user = info.context.user
+            data['id'] = graphene.Node.to_global_id('User', user.id)
+        return super().mutate(root, info, **data)
+
 
 class UserDelete(ModelDeleteMutation):
     class Meta:
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -62,11 +62,15 @@ class UserInput(graphene.InputObjectType):
     note = graphene.String(description='A note about the user.')
 
 
-class CustomerInput(UserInput):
+class UserAddressInput(graphene.InputObjectType):
     default_billing_address = AddressInput(
         description='Billing address of the customer.')
     default_shipping_address = AddressInput(
         description='Shipping address of the customer.')
+
+
+class CustomerInput(UserInput, UserAddressInput):
+    pass
 
 
 class UserCreateInput(CustomerInput):
@@ -146,6 +150,25 @@ class CustomerUpdate(CustomerCreate):
         description = 'Updates an existing customer.'
         exclude = ['password']
         model = models.User
+
+
+class LoggedCustomerUpdate(CustomerCreate):
+    class Arguments:
+        id = graphene.ID(
+            description='ID of a customer to update.', required=True)
+        input = UserAddressInput(
+            description='Fields required to update logged in customer.',
+            required=True)
+
+    class Meta:
+        description = 'Updates an existing logged in customer.'
+        exclude = ['password']
+        model = models.User
+
+    @classmethod
+    def user_is_allowed(cls, user, input):
+        id = graphene.Node.to_global_id('User', user.id)
+        return user.has_perm('account.manage_users') or id == input['id']
 
 
 class UserDelete(ModelDeleteMutation):
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -155,7 +155,7 @@ class CustomerUpdate(CustomerCreate):
 class LoggedUserUpdate(CustomerCreate):
     class Arguments:
         input = UserAddressInput(
-            description='Fields required to update logged in customer.',
+            description='Fields required to update logged in user.',
             required=True)
 
     class Meta:
@@ -165,8 +165,7 @@ class LoggedUserUpdate(CustomerCreate):
 
     @classmethod
     def user_is_allowed(cls, user, input):
-        id = graphene.Node.to_global_id('User', user.id)
-        return user.has_perm('account.manage_users') or id == input['id']
+        return user.is_authenticated
 
     @classmethod
     def mutate(cls, root, info, **data):
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -154,8 +154,6 @@ class CustomerUpdate(CustomerCreate):
 
 class LoggedCustomerUpdate(CustomerCreate):
     class Arguments:
-        id = graphene.ID(
-            description='ID of a customer to update.', required=True)
         input = UserAddressInput(
             description='Fields required to update logged in customer.',
             required=True)
@@ -172,9 +170,8 @@ class LoggedCustomerUpdate(CustomerCreate):
 
     @classmethod
     def mutate(cls, root, info, **data):
-        if data['id'] == '':
-            user = info.context.user
-            data['id'] = graphene.Node.to_global_id('User', user.id)
+        user = info.context.user
+        data['id'] = graphene.Node.to_global_id('User', user.id)
         return super().mutate(root, info, **data)
 
 
--- a/saleor/graphql/account/resolvers.py
+++ b/saleor/graphql/account/resolvers.py
@@ -16,6 +16,8 @@ USER_SEARCH_FIELDS = (
 
 def resolve_user(info, id):
     logged_user = info.context.user
+    if id == "":
+        return logged_user
     user = graphene.Node.get_node_from_global_id(info, id, User)
     if logged_user.has_perm('account.manage_users') or user == logged_user:
         return user
--- a/saleor/graphql/account/resolvers.py
+++ b/saleor/graphql/account/resolvers.py
@@ -16,7 +16,7 @@ USER_SEARCH_FIELDS = (
 
 def resolve_user(info, id):
     logged_user = info.context.user
-    if id == "":
+    if not id:
         return logged_user
     user = graphene.Node.get_node_from_global_id(info, id, User)
     if logged_user.has_perm('account.manage_users') or user == logged_user:
--- a/saleor/graphql/account/resolvers.py
+++ b/saleor/graphql/account/resolvers.py
@@ -1,3 +1,4 @@
+import graphene
 import graphene_django_optimizer as gql_optimizer
 from django.db.models import Q
 from i18naddress import get_validation_rules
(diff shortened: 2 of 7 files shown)


tests/api/test_account.py
```diff
@@ -143,6 +143,30 @@ def test_query_user(staff_api_client, customer_user, permission_manage_users):
...
+def test_non_staff_user_can_only_see_his_user_data(user_api_client,
+                                                   staff_user):
+    query = """
+    query User($id: ID!) {
+        user(id: $id) {
+            email
+        }
+    }
+    """
+    ID = graphene.Node.to_global_id('User', staff_user.id)
```
Pacu2: Not sure if that should be a const

tests/api/test_account.py
```diff
@@ -143,6 +143,30 @@ def test_query_user(staff_api_client, customer_user, permission_manage_users):
...
+    variables = {'id': ID}
+    response = user_api_client.post_graphql(query, variables)
+    content = get_graphql_content(response)
+    data = content['data']['user']
+    assert data is None
+    user = user_api_client.user
```
Pacu2: Please put a comment before each case or split it up between the two tests, so it will be nice and clear what we are testing and where

saleor/graphql/account/mutations.py
```diff
@@ -148,6 +152,32 @@ class Meta:
...
+        exclude = ['password']
+        model = models.User
+
+    @classmethod
+    def user_is_allowed(cls, user, input):
+        id = graphene.Node.to_global_id('User', user.id)
+        return user.has_perm('account.manage_users') or id == input['id']
+
+    @classmethod
+    def mutate(cls, root, info, **data):
```
Pacu2: Addresses should be validated in here, there's `i18nMixin` with `validate_address` function

saleor/graphql/account/mutations.py
```diff
@@ -148,6 +152,32 @@ class Meta:
...
+        if data['id'] == '':
```
Pacu2: Can we use `if not data['id']`, or is there a case when ID will evaluate to False?
Also, do we even need to ask for `id` if this mutation can be used without it?

saleor/graphql/account/resolvers.py
```diff
@@ -1,18 +1,29 @@
...
 
 USER_SEARCH_FIELDS = (
     'email', 'default_shipping_address__first_name',
     'default_shipping_address__last_name', 'default_shipping_address__city',
     'default_shipping_address__country')
 
 
+def resolve_user(info, id):
+    logged_user = info.context.user
+    if id == "":
```
Pacu2: Can we use `if not id`?

tests/api/test_account.py
```diff
@@ -143,6 +143,47 @@ def test_query_user(staff_api_client, customer_user, permission_manage_users):
...
+    }
+    """
+    id = graphene.Node.to_global_id('User', staff_user.id)
+    variables = {'id': id}
+    response = user_api_client.post_graphql(query, variables)
+    content = get_graphql_content(response)
+    data = content['data']['user']
+    assert data is None
+
+    # Ensure user can see own data
```
Pacu2: I have mixed feeling about using `ensure` in this context, at this point, you can only check if the user can see his data.

Pacu2: I'm also missing an explanation for the first case, maybe we should split this test in half?

tests/api/test_account.py
```diff
@@ -143,6 +143,47 @@ def test_query_user(staff_api_client, customer_user, permission_manage_users):
...
+
+def test_user_query_with_empty_id(user_api_client):
+    query = """
+    query User($id: ID!) {
+        user(id: $id) {
+            email
+        }
+    }
+    """
+    variables = {'id': ""}
```
Pacu2: Single quotes

tests/api/test_account.py
```diff
@@ -408,6 +452,180 @@ def test_customer_update(
     assert not data['user']['isActive']
 
 
+def test_logged_customer_update(user_api_client, address):
+    query = """
+    mutation UpdateCustomer($id: ID!, $billing: AddressInput,
```
Pacu2: Please extract this query to some sort of a constant and reuse it in each test. It's getting kind long here

tests/api/test_account.py
```diff
@@ -408,6 +452,180 @@ def test_customer_update(
...
+    new_street_address = 'Updated street address'
+    address_data['streetAddress1'] = new_street_address
+
+    variables = {
+        'id': id, 'billing': address_data,
+        'shipping': address_data}
+    response = user_api_client.post_graphql(query, variables)
+    content = get_graphql_content(response)
+
+    User = get_user_model()
```
Pacu2: Can't we just import `User` instance?

tests/api/test_account.py
```diff
@@ -408,6 +452,180 @@ def test_customer_update(
...
+    customer = User.objects.get(email=user.email)
```
Pacu2: I don't understand this query, can't we just use `user`? Or refresh it from db if needed?

tests/api/test_account.py
```diff
@@ -408,6 +452,180 @@ def test_customer_update(
...
+    }
+    """
+
+    # this test requires addresses to be set and checks whether new address
+    # instances weren't created, but the existing ones got updated
+    user = user_api_client.user
+    assert user.default_billing_address
+    assert user.default_shipping_address
+    billing_address_pk = user.default_billing_address.pk
+    shipping_address_pk = user.default_shipping_address.pk
```
Pacu2: We can just use `user.default_shipping_address_id`

tests/api/test_account.py
```diff
@@ -408,6 +452,180 @@ def test_customer_update(
...
+
+    # check that existing instances are updated
+    assert customer_billing_address.pk == billing_address_pk
+    assert customer_shipping_address.pk == shipping_address_pk
+
+    assert customer_billing_address.street_address_1 == new_street_address
+    assert customer_shipping_address.street_address_1 == new_street_address
+
+    data = content['data']['loggedCustomerUpdate']
+    assert data['errors'] == []
```
Pacu2: `assert not`?

tests/api/test_account.py
```diff
@@ -408,6 +452,180 @@ def test_customer_update(
...
+                }
+            }
+        }
+    }
+    """
+
+    assert customer_user.default_billing_address
+    assert customer_user.default_shipping_address
+    billing_address_pk = customer_user.default_billing_address.pk
+    shipping_address_pk = customer_user.default_shipping_address.pk
```
Pacu2: Some variables are unused in this test. Please clean it up and test only the minimal scope necessary.

tests/api/test_account.py
```diff
@@ -408,6 +452,180 @@ def test_customer_update(
...
+
+    id = graphene.Node.to_global_id('User', customer_user.id)
+    address_data = convert_dict_keys_to_camel_case(address.as_data())
```
Pacu2: We have `graphql_address_data` fixture

tests/api/test_account.py
```diff
@@ -408,6 +452,180 @@ def test_customer_update(
...
+    address_data['streetAddress1'] = new_street_address
+
+    variables = {
+        'id': id, 'billing': address_data,
+        'shipping': address_data}
+    response = staff_api_client.post_graphql(query, variables)
+    assert_no_permission(response)
+
+
+def test_logged_customer_update_with_empty_id(user_api_client, address):
```
Pacu2: Can we test the minimum scope necessary here as well? I don't see a point in recreating the whole flow just to check one `if data['id'] == ''` statement

tests/api/test_account.py
Pacu2: Also, we are testing the whole flow in `test_customer_update` test, is it really necessary to check it twice? Can't we just update a simple field, like name, to check if it's working properly?

saleor/graphql/account/mutations.py
```diff
@@ -148,6 +152,32 @@ class Meta:
         model = models.User
 
 
+class LoggedCustomerUpdate(CustomerCreate):
```
Pacu2: This can be used not only for customers but also for staff.

saleor/graphql/account/mutations.py
```diff
@@ -148,6 +152,32 @@ class Meta:
...
+    class Arguments:
+        id = graphene.ID(
+            description='ID of a customer to update.', required=True)
+        input = UserAddressInput(
+            description='Fields required to update logged in customer.',
+            required=True)
+
+    class Meta:
+        description = 'Updates an existing logged in customer.'
```
Pacu2: Not very helpful tho. It allows logged user (not only customer) to update his own data.

saleor/graphql/account/mutations.py
Pacu2: Have not seen that this mutation inherits from `CustomerCreate` which already validates the addresses

tests/api/test_account.py
```diff
@@ -143,6 +143,55 @@ def test_query_user(staff_api_client, customer_user, permission_manage_users):
     assert address['phone'] == user_address.phone.as_e164
 
 
+def test_non_staff_user_can_see_his_user_data(user_api_client):
```
Pacu2: `non_staff_user` is just a customer

saleor/graphql/account/mutations.py
```diff
@@ -148,6 +152,29 @@ class Meta:
         model = models.User
 
 
+class LoggedUserUpdate(CustomerCreate):
+    class Arguments:
+        input = UserAddressInput(
+            description='Fields required to update logged in customer.',
```
Pacu2: customer -> user

saleor/graphql/account/mutations.py
```diff
@@ -148,6 +152,29 @@ class Meta:
...
+
+    class Meta:
+        description = 'Updates data of the logged in user.'
+        exclude = ['password']
+        model = models.User
+
+    @classmethod
+    def user_is_allowed(cls, user, input):
+        id = graphene.Node.to_global_id('User', user.id)
+        return user.has_perm('account.manage_users') or id == input['id']
```
Pacu2: So I think this function is inaccurate at this moment. No matter of user's permission, one can update his own data. I think we should just check if user is logged in.

tests/api/test_account.py
```diff
@@ -143,6 +143,55 @@ def test_query_user(staff_api_client, customer_user, permission_manage_users):
     assert address['phone'] == user_address.phone.as_e164
 
 
+def test_customer_can_see_his_user_data(user_api_client):
```
Pacu2: I think `..his_own_data` works better

tests/api/test_account.py
```diff
@@ -408,6 +460,60 @@ def test_customer_update(
...
+    }
+    """
+
+    # this test requires addresses to be set and checks whether new address
+    # instances weren't created, but the existing ones got updated
+    user = user_api_client.user
+    assert user.default_billing_address
+    assert user.default_shipping_address
+
+    address_data = convert_dict_keys_to_camel_case(address.as_data())
```
Pacu2: We have `graphql_address_data` fixture

tests/api/conftest.py
```diff
@@ -82,3 +84,8 @@ def staff_api_client(staff_user):
 @pytest.fixture
 def user_api_client(customer_user):
     return ApiClient(user=customer_user)
+
+
+@pytest.fixture
+def anonymous_api_client():
```
Pacu2: To keep the convention, it can be just "api_client"

tests/api/test_account.py
```diff
@@ -15,6 +15,39 @@
...
+                    id
+                }
+                defaultShippingAddress {
+                    id
+                }
+            }
+        }
+    }
+"""
+USER_EMAIL_QUERY = """
```
Pacu2: I like the idea of keeping that query just above the tests using it, so going back and forward is way easier.

tests/api/conftest.py
Pacu2: Also, please add an issue about changing the user_api_client to api_client in tests

tests/api/test_account.py
```diff
@@ -143,6 +142,50 @@ def test_query_user(staff_api_client, customer_user, permission_manage_users):
     assert address['phone'] == user_address.phone.as_e164
 
 
+USER_EMAIL_QUERY = """
```
Pacu2: It could be just `USER_QUERY` tbh

PR #3205, opened 2018-10-31
Add the razorpay provider and add payment gateways documentations

NyanKiyoshi:
This re-implements the [django-payments-razorpay](https://github.com/NyanKiyoshi/django-payments-razorpay) provider to saleor-payment.

### Pull Request Checklist

### DIFF
--- a/docs/payment-gateways/braintree.rst
+++ b/docs/payment-gateways/braintree.rst
@@ -18,3 +18,6 @@ This gateway implements payments using `Braintree <https://www.braintreepayments
 .. note::
 
     This backend does not support fraud detection.
+    
+  .. warning::
+     Make sure that Braintree's currency is the same as your shop's, otherwise, customers will be charged the wrong amount.
--- a/docs/payment-gateways/braintree.rst
+++ b/docs/payment-gateways/braintree.rst
@@ -1,0 +1,20 @@
+Braintree (supports PayPal and Credit Cards)
+============================================
+
+This gateway implements payments using `Braintree <https://www.braintreepayments.com/>`_.
+
+.. table::
+
+    ========================== =================================================================================
+    Environment variable       Description
+    ========================== =================================================================================
+    ``BRAINTREE_SANDBOX_MODE`` Whether to use a sandbox environment for testing, ``True`` (default) or ``False``
+    ``BRAINTREE_MERCHANT_ID``  Merchant ID assigned by Braintree
+    ``BRAINTREE_PUBLIC_KEY``   Public key assigned by Braintree
+    ``BRAINTREE_PRIVATE_KEY``  Private key assigned by Braintree
+    ========================== =================================================================================
+
+
+.. note::
+
+    This backend does not support fraud detection.
--- a/saleor/payment/gateways/razorpay/__init__.py
+++ b/saleor/payment/gateways/razorpay/__init__.py
@@ -3,26 +3,15 @@ import uuid
 from decimal import Decimal
 from typing import Dict, Tuple
 
-from django.utils.translation import pgettext_lazy
-
 import razorpay
-from razorpay import errors
+import razorpay.errors
 
+from . import errors
 from ... import TransactionKind
 from ...models import Payment, Transaction
 from ...utils import create_transaction
 from .forms import RazorPaymentForm
 from .utils import get_amount_for_razorpay, get_error_response
-
-# Define the existing error messages as lazy `pgettext`.
-ERROR_MSG_ORDER_NOT_CHARGED = pgettext_lazy(
-    'Razorpay payment error', 'Order was not charged.')
-ERROR_MSG_INVALID_REQUEST = pgettext_lazy(
-    'Razorpay payment error', 'The payment data was invalid.')
-ERROR_MSG_SERVER_ERROR = pgettext_lazy(
-    'Razorpay payment error', 'The order couldn\'t be proceeded.')
-ERROR_UNSUPPORTED_CURRENCY = pgettext_lazy(
-    'Razorpay payment error', 'The %(currency)s currency is not supported.')
 
 # The list of currencies supported by razorpay
 SUPPORTED_CURRENCIES = 'INR',
@@ -30,7 +19,9 @@ SUPPORTED_CURRENCIES = 'INR',
 # Define what are the razorpay exceptions,
 # as the razorpay provider doesn't define a base exception as of now.
 RAZORPAY_EXCEPTIONS = (
-    errors.BadRequestError, errors.GatewayError, errors.ServerError)
+    razorpay.errors.BadRequestError,
+    razorpay.errors.GatewayError,
+    razorpay.errors.ServerError)
 
 # Get the logger for this file, it will allow us to log
 # error responses from razorpay.
@@ -61,7 +52,7 @@ def _generate_transaction(
 def check_payment_supported(payment: Payment):
     """Checks that a given payment is supported"""
     if payment.currency not in SUPPORTED_CURRENCIES:
-        return ERROR_UNSUPPORTED_CURRENCY % {
+        return errors.UNSUPPORTED_CURRENCY % {
             'currency': payment.currency}
 
 
@@ -69,15 +60,15 @@ def get_error_message_from_razorpay_error(exc: BaseException):
     """Convert a error razorpay error to a user friendly error message
     and log the exception to stderr."""
     logger.exception(exc)
-    if isinstance(exc, errors.BadRequestError):
-        return ERROR_MSG_INVALID_REQUEST
+    if isinstance(exc, razorpay.errors.BadRequestError):
+        return errors.INVALID_REQUEST
     else:
-        return ERROR_MSG_SERVER_ERROR
+        return errors.SERVER_ERROR
 
 
 def clean_razorpay_response(response: dict):
     """As the Razorpay response payload contains the final amount
-    as an integer, we converts it to a decimal object (by dividing by 100)."""
+    in Indian rupees, we convert the amount to paisa (by dividing by 100)."""
     response['amount'] = Decimal(response['amount']) / 100
 
 
@@ -158,7 +149,7 @@ def refund(payment: Payment, amount: Decimal, **connection_params):
             error = get_error_message_from_razorpay_error(exc)
             response = get_error_response(amount)
     else:
-        error = ERROR_MSG_ORDER_NOT_CHARGED
+        error = errors.ORDER_NOT_CHARGED
         response = get_error_response(amount)
 
     transaction = _generate_transaction(
--- a/saleor/payment/gateways/razorpay/__init__.py
+++ b/saleor/payment/gateways/razorpay/__init__.py
@@ -12,6 +12,7 @@ from ... import TransactionKind
 from ...models import Payment, Transaction
 from ...utils import create_transaction
 from .forms import RazorPaymentForm
+from .utils import get_amount_for_razorpay, get_error_response
 
 # Define the existing error messages as lazy `pgettext`.
 ERROR_MSG_ORDER_NOT_CHARGED = pgettext_lazy(
@@ -20,6 +21,11 @@ ERROR_MSG_INVALID_REQUEST = pgettext_lazy(
     'Razorpay payment error', 'The payment data was invalid.')
 ERROR_MSG_SERVER_ERROR = pgettext_lazy(
     'Razorpay payment error', 'The order couldn\'t be proceeded.')
+ERROR_UNSUPPORTED_CURRENCY = pgettext_lazy(
+    'Razorpay payment error', 'The %(currency)s currency is not supported.')
+
+# The list of currencies supported by razorpay
+SUPPORTED_CURRENCIES = 'INR',
 
 # Define what are the razorpay exceptions,
 # as the razorpay provider doesn't define a base exception as of now.
@@ -52,6 +58,13 @@ def _generate_transaction(
     return transaction
 
 
+def check_payment_supported(payment: Payment):
+    """Checks that a given payment is supported"""
+    if payment.currency not in SUPPORTED_CURRENCIES:
+        return ERROR_UNSUPPORTED_CURRENCY % {
+            'currency': payment.currency}
+
+
 def get_error_message_from_razorpay_error(exc: BaseException):
     """Convert a error razorpay error to a user friendly error message
     and log the exception to stderr."""
@@ -60,17 +73,6 @@ def get_error_message_from_razorpay_error(exc: BaseException):
         return ERROR_MSG_INVALID_REQUEST
     else:
         return ERROR_MSG_SERVER_ERROR
-
-
-def get_error_response(amount: Decimal) -> dict:
-    """Create a place holder response for invalid/ failed requests
-    for generated a failed transaction object."""
-    return {'is_success': False, 'amount': amount}
-
-
-def get_amount_for_razorpay(amount: Decimal) -> int:
-    """Convert a decimal amount to int, by multiplying the value by 100."""
-    return int(amount * 100)
 
 
 def clean_razorpay_response(response: dict):
@@ -101,22 +103,28 @@ def charge(
         amount: Decimal,
         **connection_params: Dict) -> Tuple[Transaction, str]:
     """Charge a authorized payment using the razorpay client.
+
+    But it first check if the given payment instance is supported
+    by the gateway.
+
     If an error from razorpay occurs,
     we flag the transaction as failed and return
     a short user friendly description of the error
     after logging the error to stderr."""
-    error = ''
+    error = check_payment_supported(payment=payment)
     razorpay_client = get_client(**connection_params)
     razorpay_amount = get_amount_for_razorpay(amount)
 
-    try:
-        response = razorpay_client.payment.capture(
-            payment_token, razorpay_amount)
-        clean_razorpay_response(response)
-    except RAZORPAY_EXCEPTIONS as exc:
-        error = get_error_message_from_razorpay_error(exc)
-        response = get_error_response(amount)
-        response['id'] = payment_token
+    if not error:
+        try:
+            response = razorpay_client.payment.capture(
+                payment_token, razorpay_amount)
+            clean_razorpay_response(response)
+        except RAZORPAY_EXCEPTIONS as exc:
+            error = get_error_message_from_razorpay_error(exc)
+            response = get_error_response(amount, id=payment_token)
+    else:
+        response = get_error_response(amount, id=payment_token)
 
     transaction = _generate_transaction(
         payment=payment, kind=TransactionKind.CHARGE, **response)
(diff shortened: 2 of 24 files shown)


NyanKiyoshi:
Note that I added a documentation for the payment gateways, including braintree, to tell users how to configure them. I did not write how to enable them though, as it is planned to automatically do it in `settings.py`.

docs/payment-gateways/braintree.rst
```diff
@@ -0,0 +1,20 @@
...
+    ``BRAINTREE_SANDBOX_MODE`` Whether to use a sandbox environment for testing, ``True`` (default) or ``False``
+    ``BRAINTREE_MERCHANT_ID``  Merchant ID assigned by Braintree
+    ``BRAINTREE_PUBLIC_KEY``   Public key assigned by Braintree
+    ``BRAINTREE_PRIVATE_KEY``  Private key assigned by Braintree
+    ========================== =================================================================================
+
+
+.. note::
+
+    This backend does not support fraud detection.
```
Pacu2:
```suggestion
    This backend does not support fraud detection.
    
  .. warning::
     Make sure that Braintree's currency is the same as your shop's, otherwise, customers will be charged the wrong amount.
```

NyanKiyoshi:
Just to let you know, it's available for reviewing. Let me know if there is any question, issue, concern or anything.

I don't plan on fixing the codeclimate issues, unless there are some other thoughts on the matters:
1. **`_generate_transaction` has 7 arguments**, it's for the sake of simplicity and to force an implicit decomposition of the gateway response rather than explicit through `pop(key)` which may be slower (to be proven) ;
2. **Redefining built-in '`id`'**, same as above, the gateway passes the transaction ID as `id` ;
3. **Parameters differ from overridden '`render`' method**, sounds like a false positive? ``*args`, and `**kwargs` are here to wrap Django's base signature and prevent any incompatibility ;
4. **Unused argument '`args`'**, false positive?

saleor/payment/gateways/razorpay/forms.py
```diff
@@ -0,0 +1,69 @@
...
+PRIMARY_BTN_TEXT = pgettext_lazy(
+    'Rozorpay payment gateway primary button', 'Pay now with Razorpay')
+SECONDARY_TITLE = pgettext_lazy(
+    'Rozorpay payment gateway secondary title', 'Total payment')
+
+
+class RazorPayCheckoutWidget(HiddenInput):
+    def __init__(
+            self,
+            *, payment,
```
Pacu2: We should pick some convention and stick to it, sometimes we put one argument per line, sometimes multiple ones. I'd go for one arg per line + type annotations whenever possible.

saleor/payment/gateways/razorpay/forms.py
```diff
@@ -0,0 +1,69 @@
...
+        base_attrs = kwargs['attrs'] = {
+            'src': CHECKOUT_SCRIPT_URL,
+            'data-key': public_key,
+            'data-buttontext': PRIMARY_BTN_TEXT,
+            'data-image': store_image,
+            'data-name': store_name,
+            'data-description': SECONDARY_TITLE,
+            'data-amount': int(payment.total * 100),
+            'data-currency': payment.currency
+        }
```
Pacu2: I think we are against the hanging brackets

saleor/payment/gateways/razorpay/forms.py
```diff
@@ -0,0 +1,69 @@
...
+            **kwargs):
+        override_attrs = kwargs.get('attrs', None)
+        base_attrs = kwargs['attrs'] = {
+            'src': CHECKOUT_SCRIPT_URL,
+            'data-key': public_key,
+            'data-buttontext': PRIMARY_BTN_TEXT,
+            'data-image': store_image,
+            'data-name': store_name,
+            'data-description': SECONDARY_TITLE,
+            'data-amount': int(payment.total * 100),
```
Pacu2: I see this operation all over the files, can we pack it into some kind of a function, like `get_amount_for_razorpay` and add a nice docstring why is that?

saleor/payment/gateways/razorpay/forms.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        widget = RazorPayCheckoutWidget(
+            payment=self.payment, **self.gateway_params)
+        self.fields['razorpay'] = forms.CharField(
+            widget=widget, required=False)
+
+    def process_payment(self):
+        data = super(RazorPaymentForm, self).clean()
+        transaction_id = data['razorpay_payment_id']
```
Pacu2: I think we should(can?) save it to `payment`, otherwise, it will be lost after an unsuccessful transaction.

tests/gateway/test_razorpay.py
```diff
@@ -0,0 +1,148 @@
...
+        yield mocked
+
+
+@pytest.fixture()
+def razorpay_payment(payment_dummy):
+    payment_dummy.gateway = settings.RAZORPAY
+    return payment_dummy
+
+
+def _get_gateway_params(*, prefill=True):
```
Pacu2: I'd remove the underscore as this function is not only for internal use

tests/gateway/test_razorpay.py
```diff
@@ -0,0 +1,148 @@
...
+
+from saleor.payment import ChargeStatus, PaymentError, TransactionKind
+from saleor.payment.gateways import razorpay
+from saleor.payment.gateways.razorpay.forms import (
+    RazorPayCheckoutWidget, RazorPaymentForm)
+
+
+@pytest.fixture()
+def mocked_gateway_client(razorpay_payment):
+    with patch('razorpay.Client') as mocked:
```
Pacu2: Not a fan of fixtures with side effects.  Are there any other ways we can implement that?

saleor/payment/gateways/razorpay/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+
+def refund(payment, amount: Decimal, **connection_params):
+    error = ''
+    capture_txn = payment.transactions.filter(
+        kind=TransactionKind.CHARGE, is_success=True).first()
+
+    if capture_txn is not None:
+        int_amount = int(amount * 100)
+        razorpay_client = get_client(**connection_params)
+        response = razorpay_client.payment.refund(
```
Pacu2: What will happen if we pass incorrect token? 
Braintree, for example, throws the NotFound error.

saleor/payment/gateways/razorpay/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+import razorpay
+
+from ... import TransactionKind
+from ...utils import create_transaction
+from ...models import Payment, Transaction
+
+from .forms import RazorPaymentForm
+
+
+E_ORDER_NOT_CHARGED = 'Order was not charged.'
```
Pacu2: It's not obvious what the `E` stands for in the first place, can we expand it?

saleor/payment/gateways/razorpay/__init__.py
```diff
@@ -0,0 +1,89 @@
+import uuid
+from decimal import Decimal
+from typing import Dict, Tuple
+
+import razorpay
+
+from ... import TransactionKind
+from ...utils import create_transaction
+from ...models import Payment, Transaction
+
```
Pacu2: H+Not sure if that blank space is intended

saleor/payment/gateways/razorpay/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+    int_amount = int(amount) * 100
+    razorpay_client = get_client(**connection_params)
+    response = razorpay_client.payment.capture(payment_token, int_amount)
+
+    transaction = _generate_transaction(
+        payment=payment, kind=TransactionKind.CHARGE, **response)
+    return transaction, ''
+
+
+def refund(payment, amount: Decimal, **connection_params):
```
Pacu2: Can we add types to all functions in this file?

saleor/payment/gateways/razorpay/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+def get_form_class():
+    return RazorPaymentForm
+
+
+def get_client(public_key, secret_key, **_):
+    razorpay_client = razorpay.Client(auth=(public_key, secret_key))
+    return razorpay_client
+
+
+def get_client_token(**_):
```
Pacu2: Missing tests, I mean there's nothing to break here, but let's just ping it.

saleor/payment/gateways/razorpay/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+        is_success=is_success)
+
+    return transaction
+
+
+def get_form_class():
+    return RazorPaymentForm
+
+
+def get_client(public_key, secret_key, **_):
```
Pacu2: Missing tests

saleor/payment/gateways/razorpay/forms.py
NyanKiyoshi: The logic here was
1. Passed parameters;
2. Configured parameters;
3. Any Django/ other parameters.

saleor/payment/gateways/razorpay/__init__.py
NyanKiyoshi: Fixed it, `isort` was ignoring the file.

saleor/payment/gateways/razorpay/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+        response = {'is_success': False}
+        error = E_ORDER_NOT_CHARGED
+
+    transaction = _generate_transaction(
+        payment=payment, kind=TransactionKind.REFUND, **response)
+    return transaction, error
+
+
+def void(payment, **params):
+    transaction = _generate_transaction(
```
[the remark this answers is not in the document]
NyanKiyoshi: It doesn't. Razorpay's workflow:
1. Auth is client side;
2. We capture through `charge()`;
3. We refund.

NyanKiyoshi:
> We have some checks like `can_capture` `can_void`, I assume that Razorpay cannot be captured or authorized, could we adjust those checks as well?

What do you mean? `charge()` already sets `can_void` and `can_refund` to `False`.

saleor/payment/gateways/razorpay/__init__.py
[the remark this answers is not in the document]
NyanKiyoshi: Is it safe to assume that one cannot, ever, manage to call `void()`? For now I removed it.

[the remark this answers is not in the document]
Pacu2: Gateway utils are not checking if method is implemented on the provider, that would need some adjustment.

Pacu2:
> What do you mean? charge() already sets can_void and can_refund to False.
You're right, there's another problem. We have not implemented the `can_charge()` method.
It should be implemented on the Payment, Order, and their relevant GraphQL types (plus expanding the `OrderAction`)

saleor/payment/gateways/razorpay/forms.py
Pacu2: Let's say the transaction failed. 
Do we want to keep the payment's token for debugging purposes?

NyanKiyoshi:
Sorry for the stagnant PR. I made all the requested changes a few days back, but I still want to rewrite the tests from the ground because I'm not satisfied by their flow. So I will commit the changes whenever the tests are rewritten (works fully and is fully covered as is, but I prefer to come back later with something that is much better).

I won't be available for a while because I'm swamped by work, but I could probably make room for it this Tuesday or Wednesday.

Thanks a lot @Pacu2 for your feedbacks.

*I wonder how many times I will have to rebase this branch because of `pipenv`?* :wink:

patrys:
@NyanKiyoshi I've marked some of the Code Climate issues as WONTFIX but the one about `*args` is valid and should be addressed.

saleor/payment/gateways/razorpay/forms.py
NyanKiyoshi: Nice catch, thanks for this.

NyanKiyoshi:
Alright, so:
- I updated the code with the suggested changes;
- The tests have been refactored;
- The codeclimate issue should be resolved;
- The conflicts have been resolved.

saleor/payment/gateways/razorpay/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+
+def get_error_response(amount: Decimal) -> dict:
+    """Create a place holder response for invalid/ failed requests
+    for generated a failed transaction object."""
+    return {'is_success': False, 'amount': amount}
+
+
+def get_amount_for_razorpay(amount: Decimal) -> int:
+    """Convert a decimal amount to int, by multiplying the value by 100."""
+    return int(amount * 100)
```
Pacu2: So actually I'm worried about this function. 

How can we ensure that amount will always have 2 decimal places? 
If the store's admin provides the wrong amount for a product, then his customers will be charged tens times more than they should be.

I think there are currencies which have more than 2 decimal or have no at all.
On the other hand, shifting the numbers will not work as well.
Eg.
"1" -> 1
"1.01" -> 101
"1.000001" -> ...

Does RazorPay allow other currencies?

NyanKiyoshi: I see what you mean.

Razorpay always gives us a value in integer that has two decimal places.

It is this instruction that will tell razorpay how much we are charging (it still has the `int() * 100`):
https://github.com/mirumee/saleor/blob/97bb5ebfd924c9d6068e78547a16aef815915c15/saleor/payment/gateways/razorpay/forms.py#L35

So, it will see the converted, let's guess, USD, so in the settings, we have `DEFAULT_DECIMAL_PLACES = get_currency_fraction('USD')  # -> 2`. But will fail for a store having `BTC` as default currency.

I will fix that.

For your question: RazorPay only support the INR (Indian) currency.

NyanKiyoshi: Would you be okay with doing like so:
```python
int(Decimal('3.14159').quantize(Decimal('1.01')) * 100)  # -> 3.15
```

Pacu2: I'd think of something like 
```
# One indian rupee consists of 100 paisa coins
PAISA = Decimal('0.01')

def get_amount_as_paisa(money):
    """Razorpay expects Indian Rupees to be sent in their subdivision - paisa.
    """
    if money.currency != `Indian rupee`:
        # handle error
    amount_in_paisa = money.amount // PAISA
    return amount_in_paisa
```

NyanKiyoshi: But, in my mind during the writing of this code, I assume that they *may* support other currencies. e.g. USD is almost supported but they deny you from paying. If you attempt to pay in EUR, they will show the error "Sorry, your currency is not supported.".

Should we really check that? I feel like it is not needed in our job, and then, if they happen to support USD, EUR for example, that would require us to change our internal gateway logic.

Pacu2: I assume that currency check is up to Razorpay's gateway settings?

Maybe in such case, we could try to get the number of decimal places for currency used to make it more error-proof? And then handle the amount accordingly...

patrys: Do their docs specify that the amount it so be given in the currency's lowest denomination or do they say that the amount is to be given in 1/100s of the main denomination? If it's the latter then I'd assume then to expect to multiply by 100 even when dealing with Japenese Yen that does not have a <1 denomination.

NyanKiyoshi: The docs doesn't tell anything else than
> The amount of payment in paisa. e.g. If amount = 100, it means Rs 1.
https://docs.razorpay.com/docs/return-objects#payment-entity

patrys: In that case I propose that we explicitly check that the currency is INR and add a comment that we need to revisit this logic if other currencies are going to be supported.

NyanKiyoshi: Alright, I will do that whenever I have a few minutes free. Thank you for your input, @patrys !

NyanKiyoshi: I just pushed 57338e1de14bfccd3ae532897d0b05dac0ceaa71, let me know what you think about doing this way.

Then, we could refactor later depending on how we will implement #3250.

NyanKiyoshi: I just still need to write a test for `check_payment_supported`, I totally forgot about this, until going off.

2m0op1

saleor/payment/gateways/razorpay/__init__.py
```diff
@@ -0,0 +1,166 @@
...
+from razorpay import errors
+
+from ... import TransactionKind
+from ...models import Payment, Transaction
+from ...utils import create_transaction
+from .forms import RazorPaymentForm
+from .utils import get_amount_for_razorpay, get_error_response
+
+# Define the existing error messages as lazy `pgettext`.
+ERROR_MSG_ORDER_NOT_CHARGED = pgettext_lazy(
```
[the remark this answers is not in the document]
NyanKiyoshi: That could clean that up. I was thinking about the messages into `errors.py` so we could do `[razorpay.]errors.ORDER_NOT_CHARGED`?

NyanKiyoshi:
Thanks again @Pacu2 ! I just pushed the suggested changes.

NyanKiyoshi:
@maarcingebala will you be able to drop a review on this? I have to rebase the changes, I will do that tonight.

maarcingebala:
Hey @NyanKiyoshi, I'll take a look at this PR tomorrow.

maarcingebala:
Looks really good! What about support in the API, is it compatible with it? We have `PaymentGatewayEnum` which provides available gateway types that can be used with checkout mutations.

NyanKiyoshi:
The checkout is a JS modal, we only need to have the resulting `razorpay_payment_id` to be passed to the gateway.

https://github.com/mirumee/saleor/blob/dc885fdacafdb1c8474307ccbe1a710b97a737d0/saleor/payment/gateways/razorpay/forms.py#L57-L59

maarcingebala:
I see, but if we want to support Razorpay in the Storefront 2.0, we would have to adjust the API, as those forms are used only in the Django storefront. I'm totally fine with adding support in the future.

NyanKiyoshi:
It would only require to call the gateway's `charge` method as done in the form:
https://github.com/mirumee/saleor/blob/dc885fdacafdb1c8474307ccbe1a710b97a737d0/saleor/payment/gateways/razorpay/forms.py#L71

Then, in, let's say, the storefront 2.0, we load the modal through a manual checkout, instead of this `<script>`:
https://github.com/mirumee/saleor/blob/dc885fdacafdb1c8474307ccbe1a710b97a737d0/saleor/payment/gateways/razorpay/forms.py#L11

For example, in the docs, it give this example of implementation:
```html
<button id="rzp-button1">Pay</button>
<script src="https://checkout.razorpay.com/v1/checkout.js"></script>
<script>
var options = {
    "key": "YOUR_KEY_ID",
    "amount": "2000", // 2000 paise = INR 20
    "name": "Merchant Name",
    "description": "Purchase Description",
    "image": "/your_logo.png",
    "handler": function (response){
        alert(response.razorpay_payment_id);
    },
    "prefill": {
        "name": "Gaurav Kumar",
        "email": "test@test.com"
    },
    "notes": {
        "address": "Hello World"
    },
    "theme": {
        "color": "#F37254"
    }
};
var rzp1 = new Razorpay(options);

document.getElementById('rzp-button1').onclick = function(e){
    rzp1.open();
    e.preventDefault();
}
</script>
```

Then, on the `click` event, we would pass the `razorpay_payment_id` to the gateway's `charge`. Unless I am missing a/ the point?

maarcingebala:
Yep, something like that - we would probably have to pass `razorpay_payment_id` to `CheckoutPaymentCreate`. Anyway, currently Storefront 2.0 and entire payment API is optimized mostly for Braintree's workflow, so we will get to that in the future.

PR #3223, opened 2018-11-01
Populatedb improvements

Pacu2:
- [x] Brought back option to run populatedb without creating images for the products
- [x] Adjusted the menu creation process so it should be recreated each time from the scratch, only with categories/collections which has some products


Pacu2:
@stephenmoloney It will work the same, except for the fact that all menu items are erased and recreated. 

Otherwise, there could be some categories without products left (as it happens on the demo store atm)

PR #3227, opened 2018-11-01
Skip installing chromium to make deployment faster

jxltom:
Storybooks's addon snapshots requires ```puppeteer``` which requires downloading chromium. The size is around 100MB, which make deployement process slower.

This PR disable installing chromium, since it is not required for webpack building.

I'm not so familar with storybook. But I think it looks like it not required by ```jest``` for tests or udpating snapshots by ```jest --updateSnapshot```.

I'm glad anyone could tell me when will be chromium really required in storybook's developing process.

### Pull Request Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 !.github
 !.gitignore
 !.nvmrc
+!.npmrc
 !.pylintrc
 !.travis*
 !.tx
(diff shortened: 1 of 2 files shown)


dominik-zeglen:
Hi,

Storybook uses chromium to render its stories as images. This is feature is not used in this project and probably won't be.

jxltom:
@dominik-zeglen Thanks for the helpful information!

PR #3254, opened 2018-11-08
Remove DJANGO_SETTINGS_MODULE env in CI settings but use the one in setup.cfg

jxltom:
```DJANGO_SETTINGS_MODULE ``` is set in ```setup.cfg``` for testing and then there is no need to setup this env var in CI settings.

### Pull Request Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,8 @@ script:
   - tox
 env:
   global:
-    - DATABASE_URL='postgres://postgres@localhost:5432/saleor'
+    - DATABASE_URL="postgres://postgres@localhost:5432/saleor"
+    - SECRET_KEY="irrelevant"
   matrix:
   - DJANGO="1.11"
   - DJANGO="2.1"
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,6 @@ script:
   - tox
 env:
   global:
-    - DJANGO_SETTINGS_MODULE=tests.settings
     - DATABASE_URL='postgres://postgres@localhost:5432/saleor'
   matrix:
   - DJANGO="1.11"
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,7 @@ skipsdist = True
 [testenv]
 passenv = 
     CODECOV_* TOXENV CI TRAVIS TRAVIS_*
-    DATABASE_URL
+    DATABASE_URL SECRET_KEY
 deps =
     -rrequirements.txt
     -rrequirements_dev.txt
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,7 @@ skipsdist = True
 [testenv]
 passenv = 
     CODECOV_* TOXENV CI TRAVIS TRAVIS_*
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by patrys** [comment]
Shouldn't we rather remove `-e DJANGO_SETTINGS_MODULE` from the `pytest` invocation two lines above?

**Comment by jxltom** [comment]
I'm sorry. You means?

Without setting ```DJANGO_SETTINGS_MODULE ``` env var inside docker container, pytest will rasie ```django.core.exceptions.ImproperlyConfigured: Requested setting INSTALLED_APPS, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.``` error.

**Comment by patrys** [comment]
Pytest will set the variable if it's not explicitly set (see `setup.cfg`). `-e DJANGO_SETTINGS_MODULE` overrides it.

**Comment by jxltom** [comment]
I can confirm https://github.com/mirumee/saleor/blob/5d95eaa3be3affe8d6d6226f4c4a248a616c2daa/setup.cfg#L18 is not working. There is a plugin for setting env vars in ```pytest.ini```.

Also, there is a [stackoverflow question](https://stackoverflow.com/questions/44958993/pytest-environment-variable-to-specify-pytest-ini-location) for that.

**Comment by jxltom** [comment]
Thanks for the clarification.

I got it, it is because ```pytest.ini``` is overriding pytest settings in ```setup.cfg```. The PR is updated.

File: .travis.yml [hunk 1]
```diff
@@ -21,8 +21,8 @@ script:
   - tox
 env:
   global:
-    - DJANGO_SETTINGS_MODULE=tests.settings
-    - DATABASE_URL='postgres://postgres@localhost:5432/saleor'
+    - DATABASE_URL="postgres://postgres@localhost:5432/saleor"
+    - SECRET_KEY="irrelevant"
```
NyanKiyoshi: Why are we setting the `SECRET_KEY` environment variable? It's already defined here:
https://github.com/mirumee/saleor/blob/cd74214e5fa916219517bc52ad4704754da5f553/tests/settings.py#L6

jxltom: Right. I removed the ```SECRET_KEY `` in circle CI.

But the one in travis is kept since ```collectstatic``` in tox requires it.

tox.ini
```diff
@@ -5,7 +5,7 @@ skipsdist = True
 [testenv]
 passenv = 
     CODECOV_* TOXENV CI TRAVIS TRAVIS_*
-    DJANGO_SETTINGS_MODULE DATABASE_URL
+    DATABASE_URL SECRET_KEY
```
NyanKiyoshi: Are you sure that just passing `DJANGO_SETTINGS_MODULE` wouldn't be enough? That sounds like it would work, since it would define the `SECRET_KEY`.

jxltom: Yeah that's also my first thought. But it is required because ```collectstatic``` won't use pytest's test settings. :smile:

maarcingebala:
Thanks a lot for taking care of this configuration @jxltom and @NyanKiyoshi!

PR #3255, opened 2018-11-08
Force postgresql to 9.4

jxltom:
This commit https://github.com/mirumee/saleor/commit/0191dd92e25230f0b1b9d0c20bb1c48fa56f53dc#diff-354f30a63fb0907d4ad57269548329e3 force postgresql version to 9.4. I don't know why. But it is better to keep same both in travis and circle CI.

### Pull Request Checklist


patrys:
I'm pretty sure we'd rather use PostgreSQL 11.x, @artursmet any idea why this was set to 9.4?

artursmet:
As far as I remember, travis default psql version was petty old and we needed something newer that contains Hstore extensions. There are no blockers to use more recent version of postgresql in tests. Probably they were not available at the time

PR #3258, opened 2018-11-09
Add default test runner

jxltom:
It will be nice to add pytest as the default test runner since pytest as well as its plugins are already used in this project. Then users can run test by django's default CLI ```python manage.py test```

### Pull Request Checklist

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -367,7 +367,7 @@ bootstrap4 = {
     'form_renderers': {
         'default': 'saleor.core.utils.form_renderer.FormRenderer'}}
 
-TEST_RUNNER = ''
+TEST_RUNNER = 'tests.runner.PytestTestRunner'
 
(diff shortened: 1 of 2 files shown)


patrys:
Do you think it's better to run tests using `manage.py`?

jxltom:
I think both are fine, no matter ```pytest``` or ```django-admin test```. But personally I prefer the CLI since it provides uniform way to calling different administrative tasks.

But any way, this PR provides support for ```django-admin test``` but calling ```pytest``` directly still works. People are free to choose any of them. :D

maarcingebala:
We used to use `pytest` and running it manually is common among our team members. However, newcomers who are familiar with Django may expect that running `python manage.py test` will work, while at this moment it doesn't and you have to find out that we're actually using `pytest`. Therefore I think these changes are useful and may lower the entry barrier.

PR #3263, opened 2018-11-09
Replace Graphene view with a custom one

### LINKED ISSUE #3256: Implement own GraphQLView with query batching feature
To be able to combine multiple queries into one request.

Add own view instead or relying on django-graphene's GraphQLView.

Django-graphene is able to resolve multiple queries coming from one request, but requires requests to always contain list of queries. This is unacceptable as it requires API consumer to always send list of queries, even if one does not use query batching feature. Also, it disables GraphiQL.

### Additional feature
If we are going to implement described view, I'd propose to drop GraphiQL and use GraphQL Playgrond
https://github.com/prisma/graphql-playground
as it provides support for custom headers, meaning that user would be able to test auth-protected queries and mutations.

patrys:
Custom view replaces GraphiQL with Playground, adds query batching in parallel to supporting unbatched queries and defaults to accepting all CORS origins.

Fixes #3256 

TODO:
* [x] Tests (help wanted)

### Pull Request Checklist


saleor/graphql/file_upload/views.py
```diff
@@ -1,58 +1,177 @@
...
+            data = self.parse_body(request)
+        except ValueError as e:
+            return HttpResponseBadRequest(
+                content_type='application/json',
+                content=json.dumps({
+                    'errors': [self.format_error('Unable to parse query.')]}))
+        if isinstance(data, list):
+            responses = [self.get_response(request, entry) for entry in data]
+            result = [response for response, code in responses]
+            status_code = max((code for response, code in responses),
```
NyanKiyoshi: Shouldn't you add a blank line after `max(`?

saleor/graphql/file_upload/views.py
```diff
@@ -1,58 +1,177 @@
...
+            return request.POST
+        return {}
+
+    def get_response(self, request: HttpRequest, data):
+        query, variables, operation_name = self.get_graphql_params(
+            request, data)
+        execution_result = self.execute_graphql_request(
+            request, data, query, variables, operation_name)
+        status_code = 200
+        if execution_result:
```
NyanKiyoshi: Is this condition ever False?

saleor/graphql/file_upload/views.py
patrys: That's the output of YAPF you're looking at 😄

saleor/graphql/file_upload/views.py
patrys: Not sure, that's the original logic from Django Graphene.

saleor/graphql/file_upload/views.py
```diff
@@ -1,58 +1,170 @@
...
+        try:
+            return document.execute(
+                root=self.get_root_value(request),
+                variables=variables,
+                operation_name=operation_name,
+                context=request,
+                middleware=self.middleware,
+                **extra_options)
+        except Exception as e:
+            return ExecutionResult(errors=[e], invalid=True)
```
NyanKiyoshi: That needs to be checked, but if there is an exception, you may be able to catch it here and log the stacktrace to stderr, instead of an unhelpful error message, if you can take a look, otherwise it's fine.

c.f.: https://github.com/graphql-python/graphql-core/issues/208

PR #3293, opened 2018-11-16
Fix resolving empty attributes list

### LINKED ISSUE #3264: `Attributes` field in product type should not be nullable
To get product's attribute list. 

1. Execute the following query on a product with no attributes
```
{
  product {
    attributes {
      id
    }
  }
}
```

### What I expected to happen
Instead of `null`, `[]` should be returned.

maarcingebala:
Fixes resolving attributes HStore into `SelectedAttribute` objects. Currently, when HStore was not correctly mapped to instances, such a structure might be returned:
```
attributes: [{attribute: null, value: null}]
```

This PR fixes issue be returning an empty `attributes` list instead:
```
attributes: []
```

This is why the product details page breaks for some products in the new dashboard.

Fixes #3264

### Pull Request Checklist


patrys:
Could we also change the schema to mark the fields as mandatory?

dominik-zeglen:
^ This way we'd close https://github.com/mirumee/saleor/issues/3264

maarcingebala:
I've changed the schema as you suggested and regenerated TS types. Haven't touched the components but we should be able now to remove some of the `null` checks related to `SelectedAttribute` type.

PR #3296, opened 2018-11-16
Modals refactor

### LINKED ISSUE #2884: Close modals after confirmation
We need a solution that is able to close modals when API returns no errors. Currently, this is impossible, because modals are integral part of `Page` component, so Apollo Mutations do not have any access to callbacks provided by `Toggle` components.
One solution is to place `Toggle` components above mutation providers - then we have access to close callbacks, but it kinda breaks policy that page is self-contained. It would also require to wrap every `Page` component story in `<Toggle />`s to maintain story usability.
The other one is to use context to manage modals' state. The pros is that pages would still be self-contained and instead of ladder of `<Toggle />`s, we'd only have to place one context consumer on top of `Page` component.
There had been proposed another solution based on context, but instead of toggling modal components, a callback with `content` argument of type `React.Component` would be passed.

dominik-zeglen:
I want to merge this change because it fixes opened dialog windows after successful operations.
Fixes #2884 

### Pull Request Checklist


saleor/static/dashboard-next/orders/components/OrderAddressEditDialog/OrderAddressEditDialog.tsx
```diff
@@ -6,63 +6,63 @@ import DialogTitle from "@material-ui/core/DialogTitle";
...
-        onClick={onConfirm}
-        type="submit"
-      >
-        {i18n.t("Confirm", { context: "button" })}
-      </Button>
-    </DialogActions>
-  </Dialog>
-);
+> = ({ address, open, errors, variant, countries, onClose, onConfirm }) => {
+  return (
```
patrys: I believe this could be a statement?

PR #3302, opened 2018-11-19
Update API schema

maarcingebala:
Recent changes in files structure must have affected the order of queries in mutations and it seems that schema wasn't updated. Ultimately I would consider if with the current files structure we can keep the schema alphabetically sorted.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/schema.graphql
+++ b/saleor/graphql/schema.graphql
@@ -833,6 +833,17 @@ type MoneyRange {
 }
 
 type Mutations {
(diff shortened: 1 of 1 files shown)


patrys:
Have you tried rebuilding the TS type definitions? Are there no changes required?

maarcingebala:
Yep, I did that and there are no changes, it seems that only the order of API fields has changed.

patrys:
The joy of auto-generated schemas 🐼

PR #3303, opened 2018-11-19
Add cookies notification and update privacy policy for HotJar on GetSaleor page

Bonifacy1:
### This PR will:

- [x] Add privacy policy bar for HotJar at the bottom of the screen
- [x] Create cookie if the policy is accepted
- [x] Update Privacy policy for HotJar
- [x] Connect with the HotJar

Issue: 
Close mirumee/mirumee#592
Close mirumee/mirumee#594


src/components/Header/Header.jsx
```diff
@@ -15,18 +17,19 @@ class Header extends Component {
...
-      visibleNewsBar: cookies.get('newsbar') ? false : true, 
-      sticky: false,
-      scrollDirection: 'bottom',
-      lastScrollPos: null
-    }
+    const { cookieValue } =
+      this.state = {
+        mobileMenu: false,
+        visibleNewsBar: cookies.get('newsbar') ? false : true,
+        visiblePrivacyPolicyBar: cookies.get('privacypolicybar') ? false : true,
```
NyanKiyoshi: What about putting `newsbar` and `privacypolicybar` into a `const` instead of having them hard-coded?

src/components/Header/Header.jsx
```diff
@@ -101,6 +111,15 @@ class Header extends Component {
             </nav>
           </div>
         </div>
+        {this.state.visiblePrivacyPolicyBar ?
+          <div className="notification">
+            <div className="content">
+              <img className="cookieImg" src={cookieImg} onClick={this.closePrivacyPolicyBar} />
+              <p className="privacyPolicyText">By accepting our usage of third-party software such as HotJar, you help us to deliver a better website experience to all our users. To see our full privacy policy, <a href="https://getsaleor.com/privacy-policy-terms-and-conditions/" target="_blank">click here.</a></p>
```
NyanKiyoshi: - It doesn't seems to invite the user to accept third-party tools, it doesn't sound as a prompt;
- I would also add that technical words, especially services names (HotJar) should be avoided to target a wider range of users. Maybe we could replace HotJar with "visitor analytics" or something else that doesn't sound scary.

Maybe something like (with some more rewording):
```suggestion
              <p className="privacyPolicyText">Our website is using third-party tools, such as a visitor analytics tool, that help us to deliver you a better experience. Are you agreeing?<a href="https://getsaleor.com/privacy-policy-terms-and-conditions/" target="_blank">More details here</a>.</p>
```

Also, will those services (e.g. HotJar) be loaded only if the user accepts? Or the user as to manually opt-out? If so, we probably should have a "Dismiss" button instead of "Accept", that would make more sense (and thus, we would not need to ask the user "Are you agreeing?").

NyanKiyoshi:
If the data is pseudonymized, it should be fine with the GDPR. I may be wrong though.

stephenmoloney:
For me, tooling like `google_analytics` and `hotjar` go beyond core functionality and additionally have privacy implications.

I think it would be better if they

- Can be optional
- Can be simply the task of the developer to include in his/her own fork.

NyanKiyoshi:
But here we are only talking about the `gh-pages` branch, the [getsaleor.com](https://getsaleor.com/) landing page, so. Not Saleor's code.

But I do agree that we could/ should make HotJar optional with a options modal allowing to toggle hotjar. But... Once again, it's the landing page, so it should be fine as is.

stephenmoloney:
@NyanKiyoshi My mistake, I had presumed this was a PR into master.

PR #3310, opened 2018-11-22
Remove runtime warning when product has discount

jxltom:
It is acutally a bug. It happens when product has discounts.

```
.local/share/virtualenvs/saleor-Ar4mgESs/lib/python3.6/site-packages/django/template/defaulttags.py:308: RuntimeWarning: `bool(taxed_money)` will always evaluate to True, consider replacing the test with explicit `if taxed_money is None` or `if taxed_money.gross`.
  if match:
```

It is because ```availability.discount``` can be ```None``` or ```TaxedMoney``` which is returned in ```get_availability```.

Anyway, this PR removes the annoying warning.

### Pull Request Checklist

### DIFF
--- a/templates/product/details.html
+++ b/templates/product/details.html
@@ -130,7 +130,7 @@
         {% else %}
           <h2 class="product__info__price">
             <span>{% price_range availability.price_range %}</span>
-            {% if availability.discount %}
+            {% if availability.discount != None %}
               <small class="product__info__price__undiscounted">
                 {% price availability.price_range_undiscounted %}
               </small>


templates/product/_items.html
```diff
@@ -21,7 +21,7 @@
         <div class="panel-footer">
           {% if availability.available %}
             {% price_range availability.price_range %}
-            {% if availability.discount %}
+            {% if availability.discount != None %}
```
maarcingebala: Such checks in templates always feel a little bit fishy to me and I was wondering if we shouldn't change the `availability.discount` to always return `TaxedMoney`. But it actually makes sense (e.g. in API) to return `None` if there is no discount. 
Anyway, there is also the `availability.on_sale` flag which seems to be meant exactly for that use-case. Could you test if we could use it here?

jxltom: Indeed. Thansk for the info. ```on_sale``` is perfect for this case. I've updated the PR.

PR #3311, opened 2018-11-22
Fix language code

jxltom:
Continue https://github.com/mirumee/saleor/pull/2635, Closes https://github.com/mirumee/saleor/issues/2634

The PR is made according to https://github.com/django/django/tree/2.1/django/conf/locale and https://github.com/django/django/blob/master/django/conf/global_settings.py#L51-L141

One thing I'm not sure is I've renamed and deleted some unused locale directories. How to sync with transifex without losing data?

### Pull Request Checklist


maarcingebala:
[This section](http://docs.getsaleor.com/en/latest/customization/i18n.html) in our docs explains how to sync with Transifex, you can try that out.

jxltom:
@maarcingebala Thanks for the info. It looks like I didn't do anything one. The issue is someone has transifex admin privilege should coordinate with this PR. Any changes in project's locale should sync with transifex's language settings.

For storefront and dashboard in transifex:

- ```zh``` should be renamed to ```zh_Hans```
- ```zh_TW``` should be renamed to ```zh_Hant```
- ```fa_IR``` should be renamed to ```fa```
- ```ru_RU``` should be deleted(No merge is required since ru is already 100% translated). Or just keep it in transifex since ```tx pull``` won't pull it which does not exist in local project.
- ```bg``` should be deleted and ```bg_BG``` should be renamed to ```bg```. This is because ```bg``` in transifex is actually 0%

Here is the ref for updating language in transifex, https://docs.transifex.com/projects/adding-and-removing-project-languages#changing-a-target-language

jxltom:
One more worth to do is rename ```Arabic (Egypt)``` in transifex to ```ar``` and add it to saleor's storefront and dashboard's ```locale``` and ```LANGUAGES``` setting. This is because the arabic language in transifex is translated 65.9% but is missing in saleor.

jxltom:
Still one more thing worth to consider to update transifex's language label to have same name as django's ```LANGUAGE``` settings https://github.com/django/django/blob/master/django/conf/global_settings.py#L51-L141. So users who translate won't get confused.

For example, rename ```Chinese``` language label in transifex to ```Simplified Chinese```, ```Portuguese (Brazil)``` to ```Brazilian Portuguese``` and etc.

jxltom:
If this PR can be merged, a followed PR which sync transifex's translation should be made, which runs ```tx pull```. And then everything will be up-to-date and perfect. :smile:

patrys:
I've updated the list of languages on Transifex to reflex Django's locale support. The only downside is that I'm currently marked as the author of a large chunk of the renamed translations.

jxltom:
@patrys Awesome. Thank you. It looks great to me.

PR #3316, opened 2018-11-22
Refactor logged in user query

### LINKED ISSUE #3285: Add viewer query
With PR #3202 we've added ability to get the data of currently logged in user with the `user` query. I would recommend refactoring it a little bit and introducing a separate query for that. The problem with the `user` query is that it expects the `ID!` argument, but it also accepts passing `""` as ID value to resolve the logged in user. This brakes the single-responsibility rule (which is a good practice for GraphQL queries), is a bit unintuitive and makes to code harder to maintain.

I would propose changing the schema to have the following queries:
`viewer: User` - returns currently authenticated user
`user(id: ID!): User` - resolves a user by ID, where ID is required

fowczarek:
I want to merge this change because resolve #3285 

We need to refactor the `user` queries to adapt it to single-responsibility rule. We can achieve this by moving logged in user data query from `user` query to `me` query.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/types.py
+++ b/saleor/graphql/account/types.py
@@ -48,7 +48,7 @@ class User(CountableDjangoObjectType):
         PrefetchingConnectionField(
             Address, description='List of all user\'s addresses.'),
         model_field='addresses')
-    note = graphene.String(description='')
+    note = graphene.String(description='A note about the customer')
 
     class Meta:
         exclude_fields = ['password', 'is_superuser', 'OrderEvent_set']
--- a/saleor/graphql/account/types.py
+++ b/saleor/graphql/account/types.py
@@ -2,6 +2,7 @@ import graphene
 import graphene_django_optimizer as gql_optimizer
 from django.contrib.auth import get_user_model
 from graphene import relay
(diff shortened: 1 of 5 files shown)


saleor/graphql/account/types.py
```diff
@@ -47,6 +48,7 @@ class User(CountableDjangoObjectType):
         PrefetchingConnectionField(
             Address, description='List of all user\'s addresses.'),
         model_field='addresses')
+    note = graphene.String(description='')
```
maarcingebala: We could add the description here, I think a suitable one would be "A note about the customer".

tests/api/test_account.py
```diff
@@ -278,6 +255,52 @@ def test_who_can_see_user(
     assert content['data']['customers']['totalCount'] == 1
 
 
+ME_QUERY = """
+    query Me{
```
NyanKiyoshi: The fact that it doesn't have a space before the bracket bugs me. It be more consistent if there was a space, if you can quickly do that change? Same for the query in `test_me_query_customer_can_not_see_note`'s body.
```suggestion
    query Me {
```

PR #3322, opened 2018-11-23
Sync translations with Transifex

patrys:
New languages:
* Arabic
* Armenian
* Bengali
* Danish
* Mongolian
* Swedish

### Pull Request Checklist

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -5,14 +5,23 @@ host = https://www.transifex.com
 file_filter = saleor/dashboard/locale/<lang>/LC_MESSAGES/django.po
 source_file = saleor/dashboard/locale/en/LC_MESSAGES/django.po
 source_lang = en
+minimum_perc = 1
+trans.zh-Hans = saleor/dashboard/locale/zh_Hans/LC_MESSAGES/django.po
+trans.zh-Hant = saleor/dashboard/locale/zh_Hant/LC_MESSAGES/django.po
 
 [saleor-1.javascriptpo]
 file_filter = locale/<lang>/LC_MESSAGES/djangojs.po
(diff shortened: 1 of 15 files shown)


.tx/config
```diff
@@ -5,14 +5,23 @@ host = https://www.transifex.com
 file_filter = saleor/dashboard/locale/<lang>/LC_MESSAGES/django.po
 source_file = saleor/dashboard/locale/en/LC_MESSAGES/django.po
 source_lang = en
+minimum_perc = 1
+trans.zh-Hans = saleor/dashboard/locale/zh_Hans/LC_MESSAGES/django.po
+trans.zh-Hant = saleor/dashboard/locale/zh_Hant/LC_MESSAGES/django.po
```
jxltom: I don't understand why only zh-hans/hanz are added here as a special case?

saleor/settings.py
```diff
@@ -58,24 +58,30 @@ def get_bool_from_env(name, default_value):
...
+    ('bn', _('Bengali')),
     ('cs', _('Czech')),
+    ('da', _('Danish')),
     ('de', _('German')),
     ('en', _('English')),
     ('es', _('Spanish')),
     ('fa', _('Persian')),
     ('fr', _('French')),
     ('hu', _('Hungarian')),
+    ('hy', _('Armenian')),
```
jxltom: It looks like Armenian is not valid language code in django

.tx/config
patrys: They have a dash in their name that is translated to an underscore on our side.

saleor/settings.py
patrys: Good catch!

PR #3327, opened 2018-11-26
Only enable graphql-playground in DEBUG mode

jxltom:
- Enable graphql-playground in production mode has security issues such as it has inline-scripts which breaks CSP security. This PR only enable graphql-playground in DEBUG mode
- Default graphene-django has support for GET method such as ```query``` and ```variable``` query parameters. This PR adds this kind of support
- Slash of graphql api url is removed as shown in the example of ```graphene-python```, then ```graphql``` and ```graphql/``` can both be accessed by POST method.

### Pull Request Checklist

### DIFF
--- a/tests/api/test_view.py
+++ b/tests/api/test_view.py
@@ -1,5 +1,6 @@
 import graphene
 import pytest
+from django.test import override_settings
 
 from .conftest import API_PATH
 from .utils import get_graphql_content, _get_graphql_content_from_response
@@ -39,7 +40,13 @@ def test_batch_queries(category, product, api_client):
     assert data['category']['name'] == category.name
 
 
-def test_graphql_view_get(client):
+def test_graphql_view_get_in_non_debug_mode(client):
+    response = client.get(API_PATH)
+    assert response.status_code == 405
+
+
+@override_settings(DEBUG=True)
+def test_graphql_view_get_in_debug_mode(client):
     response = client.get(API_PATH)
     assert response.status_code == 200
     assert response.templates[0].name == 'graphql/playground.html'


patrys:
GET is not part of the GraphQL transport spec. Until we try to support persisted queries, I don't see a reason to enable querying over GET. It will result in all sorts of issues related to caching at proxy level (GraphQL is not meant to be cached at transport level).

As for calling out to a CDN, we could serve that script from a local copy. I don't see a reason to disable Playground for non-DEBUG deployments.

saleor/urls.py
```diff
@@ -26,7 +26,7 @@
 non_translatable_urlpatterns = [
     url(r'^dashboard/',
         include((dashboard_urls, 'dashboard'), namespace='dashboard')),
-    url(r'^graphql/', csrf_exempt(GraphQLView.as_view(
+    url(r'^graphql', csrf_exempt(GraphQLView.as_view(
```
patrys: Why would you like to change that? IMHO there should be exactly one way to access an API and Saleor terminates all other resource names with a slash.

saleor/graphql/views.py
```diff
@@ -46,8 +47,10 @@ def __init__(
     def dispatch(self, request, *args, **kwargs):
         # Handle options method the GraphQlView restricts it.
         if request.method == 'GET':
-            return render_to_response('graphql/playground.html')
-        if request.method == 'OPTIONS':
+            if settings.DEBUG:
+                return render_to_response('graphql/playground.html')
+            response = self.handle_query(request, *args, **kwargs)
```
NyanKiyoshi: Shouldn't we throw a `405 - method not allowed`, instead?

saleor/urls.py
jxltom: It is because ```graphql``` url make graphql can be accessed by both ```graphql``` and ```graphql\```. With ```graphql\```, post to ```graphql``` will be lost even it is redirected by ```APPEND_SLASH``` settings.

Sometime it will confuse user who don't know Saloer's graphql api must be ```graphql\``` instead of ```graphql```

saleor/graphql/views.py
jxltom: You mean ```405``` should be raised when it is not in DEBUG mode?

NyanKiyoshi: Yup!

jxltom:
> As for calling out to a CDN, we could serve that script from a local copy. I don't see a reason to disable Playground for non-DEBUG deployments.

Probably one reason is that I don't want the Graphql API can be played by users but only to developers? A playground to normal users will make the server has more stress which is not neccessary since I may don't would like to expose the APIs as SDK at all.

saleor/graphql/views.py
jxltom: Sure, if Saleor has no plan for support GET method of GraphQL, it is reasonable to do that.

patrys:
@jxltom Hosted Playground is just a convenience tool, if someone wants to fiddle with the API, they can run either GraphiQL or Playground as a desktop app. The API has introspection enabled so I don't see hosted Playground as an attack vector. Large players like Shopify link to their GraphiQL interfaces in their platform docs.

jxltom:
@patrys  Sounds reasonable, it won't be a attack vector.

But what if I would like to hide my backend docs from normal user? With the playground, users can read the grpahql docs as well as the APIs, they can make their own mobile app or frontend.

Probably it would be nice to add a setting to enable/disable playground as well as CORS manually?

patrys:
If you want to expose a private API, I think I'd start by allowing the introspection to be disabled. I'm not sure that's part of Saleor's scope though, do we have a good use case?

jxltom:
The user case is that the shops which are using Saleor may just don't would like to publish their APIs. As for the reason, hmm, with introspection, duplicating ios/android/web client can be easily (though it can be prevented by CORS), or introspection may leak backend information accidentally. I don't know :D

jxltom:
Update: I've remove the support for GET of graphql, but still kept the part that only serving palyground in non-debug mode which is mainly for CSP security.

maarcingebala:
I think it is fine to disable Playground when `Debug=False` and we also used to do that with GraphiQL before.

PR #3340, opened 2018-11-27
Fix blank product type tax rate field

### LINKED ISSUE #3294: Error resolving ProductType with no TaxRateType
To resolve `TaxRateType` on `ProductType`, when it's set to null.

1. Execute query
```
{
  productType(id: "UHJvZHVjdFR5cGU6MTA=") {
    id
    taxRate
  }
}
```
2. Get an error
```
{
  "message": "Expected a value of type \"TaxRateType\" but received: ",
  "path": [
    "productType",
    "taxRate"
  ]
}
```

fowczarek:
I want to merge this change because resolve #3294 

If the product type has a blank tax rate, the GraphQL API returns errors.  We must ensure that this field cannot be blank.

### Pull Request Checklist


Pacu2:
We were talking about the possibility to set a "no tax" option for a product, what do you think @maarcingebala?

Also, please fix the closing keyword in the PR description

maarcingebala:
@Pacu2 I think having the `standard` tax rate as a default is fine. Tax calculations can be disabled at all so if don't need it, they won't be taken into account.

PR #3350, opened 2018-11-28
Split paths and urls

### LINKED ISSUE #3305: Embed `encodeURIComponent` in url generating functions
We can write wrapper functions to current url generators to embed `encodeURIComponent` there, eg. 

```export const orderUrl = (id: string) => urlJoin(orderSectionUrl, id);```

would become

```
export const orderPath = (id: string) => urlJoin(orderSectionUrl, id);
export const orderUrl = (id: string) => orderPath(encodeURIComponent(id));
```

This way we can omit repeating pattern `orderUrl(encodeURIComponent(id))` in favor to `orderUrl(id)`.
Also, this would require using `...Path() functions` in `<Route />` components, as they should not be excaped.

### PR DESCRIPTION
**Pull request by dominik-zeglen**
I want to merge this change because it introduces the pattern that eases router path creation and unburdens developer from encoding parameters each time URL function is used.
Resolves #3305 

### Pull Request Checklist

### DIFF
--- a/saleor/static/dashboard-next/productTypes/urls.ts
+++ b/saleor/static/dashboard-next/productTypes/urls.ts
@@ -1,6 +1,13 @@
 import * as urlJoin from "url-join";
 
 const productTypeSection = "/productTypes/";
-export const productTypeListUrl = productTypeSection;
-export const productTypeAddUrl = urlJoin(productTypeSection, "add");
-export const productTypeUrl = (id: string) => urlJoin(productTypeSection, id);
+
+export const productTypeListPath = productTypeSection;
+export const productTypeListUrl = productTypeListPath;
+
+export const productTypeAddPath = urlJoin(productTypeSection, "add");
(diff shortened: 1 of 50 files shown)

### CODE REVIEW

File: saleor/static/dashboard-next/productTypes/urls.ts [hunk 1]
```diff
@@ -1,6 +1,13 @@
 import * as urlJoin from "url-join";
 
 const productTypeSection = "/productTypes/";
```
maarcingebala: Let's also change this path to `/product-types/`.

saleor/static/dashboard-next/siteSettings/urls.ts
```diff
@@ -1,5 +1,9 @@
 import * as urlJoin from "url-join";
 
 const siteSettingsSection = "/siteSettings";
```
maarcingebala: Add this one to `/site-settings/`.

PR #3356, opened 2018-11-29
Add can_finalize flag in draft order

### LINKED ISSUE #3318: Return `canFinalize` flag in draft order
To get an information whether a draft order can be finalized or not. This way we can avoid duplicating logic in the dashboard.

fowczarek:
I want to merge this change because resolve #3318 

In GraphQL queries and mutations, we should be able to get an information whether a draft order can be finalized or not.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/types.py
+++ b/saleor/graphql/order/types.py
@@ -9,6 +9,7 @@ from ..core.types.common import CountableDjangoObjectType
 from ..core.types.money import Money, TaxedMoney
 from ..payment.types import OrderAction, Payment, PaymentChargeStatusEnum
 from ..shipping.types import ShippingMethod
+from .utils import can_finalize_order
 
 OrderEventsEnum = graphene.Enum.from_enum(OrderEvents)
 OrderEventsEmailsEnum = graphene.Enum.from_enum(OrderEventsEmails)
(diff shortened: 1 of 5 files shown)


saleor/graphql/order/utils.py
```diff
@@ -0,0 +1,41 @@
...
+    if not order.user and not order.user_email:
+        errors.append(
+            Error(
+                field=None,
+                message='Both user and user_email fields are null'))
+    return errors
+
+
+def can_finalize_order(order):
+    """Return `True` if an order draft can be complete.
```
maarcingebala: Have you checked this docstring with `pydocstyle`? I think it's not in correct format.

saleor/graphql/order/types.py
```diff
@@ -175,6 +176,8 @@ class Order(CountableDjangoObjectType):
         TaxedMoney,
         description='The sum of line prices not including shipping.')
     status_display = graphene.String(description='User-friendly order status.')
+    can_finalize = graphene.Boolean(
+        description='Informs if an order draft can be complete.')
```
maarcingebala: I'd rephrase a little bit: "Informs whether a draft order can be finalized (turned into a regular order)."

saleor/graphql/order/utils.py
maarcingebala: It seems that `check_for_draft_order_errors` function tells us whether a draft order can be finalized. If yes I'd suggest renaming `check_for_draft_order_errors` to `can_finalize_draft_order` and dropping `can_finalize_order` at all.

Issue #3364, opened 2018-11-30
Celery in Docker won't start


bykof:
### What I'm trying to achieve

Hi guys, 
I want to develop saleor with docker-compose

### Steps to reproduce the problem
1. I pulled the project via: `git clone https://github.com/mirumee/saleor.git`.
2. Then I run docker-compose up. 
3. Error shows up 

### What I expected to happen

That it should work? :D 

### Stacktrace 
```
celery_1  |  -------------- celery@1c13f3acc574 v4.2.1 (windowlicker)
celery_1  | ---- **** -----
celery_1  | --- * ***  * -- Linux-4.9.125-linuxkit-x86_64-with-debian-9.6 2018-11-30 07:28:26
celery_1  | -- * - **** ---
celery_1  | - ** ---------- [config]
celery_1  | - ** ---------- .> app:         saleor:0x7f40d790b438
celery_1  | - ** ---------- .> transport:   redis://redis:6379/1
celery_1  | - ** ---------- .> results:
celery_1  | - *** --- * --- .> concurrency: 4 (prefork)
celery_1  | -- ******* ---- .> task events: OFF (enable -E to monitor tasks in this worker)
celery_1  | --- ***** -----
celery_1  |  -------------- [queues]
celery_1  |                 .> celery           exchange=celery(direct) key=celery
celery_1  |
celery_1  |
celery_1  | [tasks]
celery_1  |   . saleor.account.emails.send_account_delete_confirmation_email
celery_1  |   . saleor.account.emails.send_password_reset_email
celery_1  |   . saleor.core.analytics.ga_report
celery_1  |   . saleor.dashboard.emails.send_promote_customer_to_staff_email
celery_1  |   . saleor.dashboard.emails.send_set_password_email
celery_1  |   . saleor.order.emails.send_fulfillment_confirmation
celery_1  |   . saleor.order.emails.send_fulfillment_update
celery_1  |   . saleor.order.emails.send_order_confirmation
celery_1  |   . saleor.order.emails.send_payment_confirmation
celery_1  |   . saleor.product.tasks.update_variants_names
celery_1  |   . saleor.product.thumbnails.create_category_background_image_thumbnails
celery_1  |   . saleor.product.thumbnails.create_collection_background_image_thumbnails
celery_1  |   . saleor.product.thumbnails.create_product_thumbnails
celery_1  |
celery_1  | [2018-11-30 07:28:27,403: INFO/MainProcess] Connected to redis://redis:6379/1
[...]

PR #3366, opened 2018-11-30
Fix celery in docker-compose startup (#3364)

bykof:
I want to merge this change because of this ticket

https://github.com/mirumee/saleor/issues/3364


bykof:
First it worked for me in the container, then the problem come again. I will close the pull request unless I find another solution

Issue #3125, opened 2018-10-17
Add publication date to collections


dominik-zeglen:
### What I'm trying to achieve
To publish collection with some delay.

### Describe a proposed solution
Add `published_at`field to collection model and check if the current date is greater than `published_at`. Basically the same as `available_on` in products.

Pacu2:
@lokesh005 go for it, there's similar mechanism already live on `Page` model, you might want to take some inspiration out of it 

Any questions- I'm here to help

gagantrivedi:
can i take this up @Pacu2

kika115:
Can I take this up?

Pacu2:
I think that someone already handles it in #3369

PR #3369, opened 2018-12-01
Add publication date to collections #3125

k-brk:


Pacu2:
Could you please use closing keywords in the PR description? It makes maintainers life easier https://help.github.com/articles/closing-issues-using-keywords/

templates/dashboard/includes/_collection_availability.html
```diff
@@ -1,9 +1,20 @@
...
-      {% trans "Not Published" context "Collection availability status" %}
-    {% endif %}
+    {% elif not collection.is_published %}
+    <div class="label">
+      {% trans "Hidden" context "Collection availability status" %}
+    </div>
+    {% elif collection.published_at %}
+  <div class="label">
+    {% blocktrans with date=collection.published_at context "Collection availability status" %}
+      Hidden (will become visible on {{ date }})
```
NyanKiyoshi: The `elif` logic is wrong here. You want this code to be executed whenever `is_published is True` and `published_at is not None`.

tests/conftest.py
```diff
@@ -598,6 +598,14 @@ def collection(db):
     return collection
 
 
+@pytest.fixture
+def not_yet_available_collection(db):
```
NyanKiyoshi: Once again, saying:
> is collection available

Is really odd for a collection (plus the test `test_collection_not_yet_published` is having a different name).

NyanKiyoshi: Do we really need this fixture? It's only used once.

tests/test_collection.py
```diff
@@ -38,3 +38,11 @@ def test_collection_not_published_404(client, draft_collection):
     url = reverse('product:collection', kwargs=url_kwargs)
     response = client.get(url)
     assert response.status_code == 404
+
+def test_collection_not_yet_published(admin_client, client, not_yet_available_collection):
```
NyanKiyoshi: We could add a multiple cases test here (parametrized test) instead of having two tests (`test_collection_not_published_404`). Also, we could have a more descriptive test name, as the test above this one: `test_collection_not[_yet]_published_returns_404`.

tests/test_collection.py
```diff
@@ -38,3 +38,11 @@ def test_collection_not_published_404(client, draft_collection):
...
+    url_kwargs = {'pk': not_yet_available_collection.pk, 'slug': not_yet_available_collection.slug}
+    url = reverse('product:collection', kwargs=url_kwargs)
+    response = client.get(url)
+    assert response.status_code == 404
+    response = admin_client.get(url)
```
NyanKiyoshi: I don't think the test will pass, you cannot have two client sessions.

templates/dashboard/includes/_collection_availability.html
```diff
@@ -1,9 +1,20 @@
...
+    {% endblocktrans %}
   </div>
+    <div class="label">
```
NyanKiyoshi: That block seems to be an accidental paste. Otherwise the logic seems wrong.

saleor/product/models.py
```diff
@@ -469,6 +472,12 @@ def get_absolute_url(self):
             'product:collection',
             kwargs={'pk': self.id, 'slug': self.slug})
 
+    @property
+    def is_available(self):
+        today = datetime.date.today()
```
NyanKiyoshi: Couldn't we have this instruction in a single condition line? That would save some CPU cycles when `is_published is True`.

saleor/menu/migrations/0011_auto_20181130_1540.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('menu', '0010_auto_20180913_0841'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='menu',
+            options={'ordering': ['pk'], 'permissions': (('manage_menus', 'Manage navigation.'),)},
```
NyanKiyoshi: That migration seems accidental.

saleor/product/migrations/0081_auto_20181202_1337.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.3 on 2018-12-02 19:37
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('product', '0080_collection_published_at'),
```
NyanKiyoshi: Could you merge the file?

saleor/product/models.py
```diff
@@ -435,10 +435,9 @@ class VariantImage(models.Model):
 class CollectionQuerySet(models.QuerySet):
 
     def public(self):
-        today = datetime.datetime.today()
         return self.filter(
             Q(is_published=True),
-            Q(published_at__lte=today) | Q(published_at__isnull=True))
+            Q(published_date__lte=datetime.date.today()) | Q(published_date__isnull=True))
```
NyanKiyoshi: Wouldn't be switching the two conditions be faster?

```diff
-published_date <= today or published_date is None
+published_date is None or published_date <= today
```

templates/collection/index.html
```diff
@@ -17,6 +17,13 @@
         You are previewing a collection that is not published.
       {% endblocktrans %}
     </div>
+  {% elif not is_available %}
+    <div class="alert alert-warning" role="alert">
+      {% blocktrans trimmed with date=object.published_at|date context "Unpublished collection details text" %}
+        <strong>Warning!</strong>
+        You are previewing a collection that will become visible on <strong>{{ date }}</strong>.
```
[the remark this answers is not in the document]
NyanKiyoshi: *resolved*

templates/collection/index.html
```diff
@@ -17,9 +17,9 @@
         You are previewing a collection that is not published.
       {% endblocktrans %}
     </div>
-  {% elif not is_available %}
+  {% elif object.is_published and object.published_date is not None %}
```
NyanKiyoshi: We already know the collection is published, no need to evaluate twice.

templates/dashboard/includes/_collection_availability.html
```diff
@@ -1,20 +1,16 @@
...
   <div class="label label-{{ label_cls }}">
-    {% if is_available %}
+    {% if is_visible %}
       {% trans "Published" context "Collection availability status" %}
     {% elif not collection.is_published %}
     <div class="label">
       {% trans "Hidden" context "Collection availability status" %}
     </div>
-    {% elif collection.published_at %}
+    {% elif collection.is_published and collection.published_date is not None %}
```
NyanKiyoshi: `collection.is_published` is `True`, no need to check twice.

The logic is still wrong though, you want to check `date is not None` first, then have a `else` block.

templates/dashboard/includes/_collection_availability.html
NyanKiyoshi: *partially resolved, see recent comments*

tests/test_collection.py
```diff
@@ -33,8 +33,17 @@ def test_collection_not_exists(client):
     assert response.status_code == 404
 
 
-def test_collection_not_published_404(client, draft_collection):
+def test_collection_not_published_404(admin_client, client, draft_collection):
```
NyanKiyoshi:
```suggestion
def test_collection_not_yet_published_returns_404(
    admin_client, client, draft_collection):
```

tests/test_collection.py
```diff
@@ -33,8 +33,17 @@ def test_collection_not_exists(client):
...
     url_kwargs = {'pk': draft_collection.pk, 'slug': draft_collection.slug}
     url = reverse('product:collection', kwargs=url_kwargs)
     response = client.get(url)
     assert response.status_code == 404
+
+    draft_collection.is_published = True
+    draft_collection.published_date = date.today() + timedelta(days=1)
+    draft_collection.save()
```
NyanKiyoshi: We could add a blank line here, and comment the two below sections, with:
> A non staff user should not have access to collections yet to be published

> A staff user should have access to collections yet to be published

templates/dashboard/includes/_collection_availability.html
```diff
@@ -1,16 +1,17 @@
...
-      {% trans "Published" context "Collection availability status" %}
-    {% elif not collection.is_published %}
-    <div class="label">
-      {% trans "Hidden" context "Collection availability status" %}
-    </div>
-    {% elif collection.is_published and collection.published_date is not None %}
+<div class="label label-{{ label_cls }}">
+  {% if is_visible %}
+    {% trans "Published" context "Collection availability status" %}
+  {% elif not collection.is_published %}
```
NyanKiyoshi: We could simplify the logic with:
```python
if is_visible:
    'published'
elif collection.published_date is not None: 
    '{date}'
else:
    'Hidden'
```

templates/dashboard/includes/_collection_availability.html
```diff
@@ -1,16 +1,17 @@
...
+    {% trans "Hidden" context "Collection availability status" %}
+  {% elif collection.published_date is not None %}
   <div class="label">
```
NyanKiyoshi: Note the indentation is wrong here.

tests/test_collection.py
```diff
@@ -33,12 +33,15 @@ def test_collection_not_exists(client):
...
-def test_collection_not_published_404(admin_client, client, draft_collection):
+def test_collection_not_yet_published_returns_404(
+    admin_client, client, draft_collection):
     url_kwargs = {'pk': draft_collection.pk, 'slug': draft_collection.slug}
     url = reverse('product:collection', kwargs=url_kwargs)
     response = client.get(url)
     assert response.status_code == 404
 
+    # A non staff user should not have access to collections yet to be published
+    # A staff user should have access to collections yet to be published
```
NyanKiyoshi: I meant above `client.get(...)` and `admin_client.get(...)`, that would make more sense.

NyanKiyoshi:
@k-brk could you flag the yellow outdated reviews as resolved? That would be awesome, because the page is really messy! 👍

templates/dashboard/includes/_collection_availability.html
k-brk: What about case:
```python
collection.is_published = False and collection.published_date is not None
```
with simplified logic it will say that is hidden and will become available on published_date, but I think its misleading as is_published = False then it will not be published on published_date.
```python
    @property
    def is_visible(self):
        return self.is_published and (
            self.published_date is None or
            self.published_date <= datetime.date.today())
```

NyanKiyoshi: > ```python
> collection.is_published = False and collection.published_date is not None
> ```

It's the same thing as the second case:

> ```python
> if is_visible: pass
> elif collection.published_date is not None: pass
> ```

I don't understand what you mean.

saleor/graphql/product/mutations/products.py
```diff
@@ -110,6 +110,8 @@ class CollectionInput(graphene.InputObjectType):
     description = graphene.String(description='Description of the collection.')
     background_image = Upload(description='Background image file.')
     seo = SeoInput(description='Search engine optimization fields.')
+    published_date = graphene.String(
```
NyanKiyoshi: Shouldn't this be of the type `Date`?

k-brk: Yeah It should. I got inspiration from PageInput (saleor/graphql/page/mutations.py) available_on field and set string instead of date.Also with this change I think that a change will be required in conftest.py as json module does not serialize date, probably djangoJSONEncoder will do the job. Tomorrow I will push another commit for review with fix.

templates/dashboard/includes/_collection_availability.html
k-brk: I apologise for not making myself clear, that's correct that its same thing as the second case, but at least for me the message which is displayed in dashboard for this case is misleading when logic is simplified.
Lets take a look at the example, I simplified this logic as you said then created four collections in which different publication options were used. Below you can see what publication options were used for given collection:
1) Published without publication date
2) Published with publication date
3) **Not published with publication date**
4) Not published without publication date

Now in dashboard their visibility column shows:
1) Published 
2) Hidden, will become visible on date
3) **Hidden, will become visible on date**
4) Hidden

1st, 2nd and 4th is correct, but is 3rd correct? Should it say in visibility column that it will become visible when published field is false and published_date is set? In my opinion it shouldn't as is_visible model property still would be false even when published_date is reached. It can be fixed with nested if inside elif but that would same logic as is now.

I hope this time is more clear what I mean.

saleor/graphql/product/mutations/products.py
k-brk: Pushed commit with changes, but I think that djangoJSONEncoder causes security/snyk to fail (I do not have access), there was nothing new except this change, so I think that's the reason. That's probably the reason why PageInput has date as string.

k-brk: Pushed commit with changes related to last review, but I think that djangoJSONEncoder causes security/snyk to fail (I do not have access to check), there was not much new except this change, so I think that's the reason. That's probably the reason why PageInput has date as string.

tests/test_collection.py
```diff
@@ -40,13 +40,13 @@ def test_collection_not_yet_published_returns_404(
...
-    # A staff user should have access to collections yet to be published
     draft_collection.is_published = True
     draft_collection.published_date = date.today() + timedelta(days=1)
     draft_collection.save()
+
+    # A non staff user should not have access to collections yet to be published
     response = client.get(url)
     assert response.status_code == 404
-
+    # A staff user should have access to collections yet to be published
```
NyanKiyoshi: Just add a blank line before, for better readability :+1:

k-brk:
Commit 04c5b22 introduced changes in type of CollectionInput(graphql/product/mutations/products.py) published_date field, from string to date. Had to change data dumper from json to djangoJsonEncoder(which is almost the same as json.JSONEncoder) in tests/api/conftest.py due to the fact that built in json module does not serialize dates. After this change security/snyk is failing. Should I roll back it to string as it was before? Similar string approach is used in PageInput (graphql/page/mutations.py) available_on field.

templates/collection/index.html
```diff
@@ -17,6 +17,13 @@
         You are previewing a collection that is not published.
       {% endblocktrans %}
     </div>
+  {% elif object.published_date is not None %}
```
Pacu2: can we just `elif object.published_date`? Is there any case where we explicitly need to check if published date is None?

Pacu2:
@k-brk Snyk is checking for dependency vulnerabilities and this error is not connected to your changes, as you've introduced no new packages. 

Probably package was flagged as vulnerable between your commits, that's why we've started seeing this message after certain push.

templates/collection/index.html
k-brk: I think that, there is no case where we need to check if its None, but on other hand is this condition valid? Let say we create published collection with publication date set to today + 1 day. It will say that we are previewing not published collection that will become visible on xyz which is fine, but the next day it will continue to show same information because this condition just checks if object published date is set. I'm assuming it should be additionally compared with today's date.

Pacu2: Right, I think that we can remove this `elif` statement and manipulate the message in the above's `if`. `object.is_published` already checks if 1. `published_date is not None` or `published_date < date.today()`.

In such case, I think that we could construct it like:

```
if not object.is_published:
    if object.published_date:
         You are previewing a collection that will become visible on...
    else:
          You are previewing a collection that is not published...
```

templates/collection/index.html
```diff
@@ -10,19 +10,19 @@
...
-    <div class="alert alert-warning" role="alert">
-      {% blocktrans context "Unavailable collection detail text" %}
-        <strong>Warning!</strong>
-        You are previewing a collection that is not published.
-      {% endblocktrans %}
-    </div>
-  {% elif object.published_date is not None %}
+  {% if not object.is_visible %}
     <div class="alert alert-warning" role="alert">
+    {% if object.published_date %}
```
NyanKiyoshi: Could you fix the indentation in the `<div>`'s body?

maarcingebala:
@k-brk We've added changelog file recently, would you mind adding a note describing your changes?

PR #3385, opened 2018-12-04
Fix range filters not being applied

mjelonek92:
Fixes https://github.com/mirumee/saleor/issues/3371.

`django-filters` requires suffixes instead of indexed fields (i.e. range_min/max instead of range_0/1). We use a different renderer (materialcss) effectively skipping Django widget render and thus never running `get_context` that django-filters relies on (it renames the fields to use the correct suffix). This fix kinda "brute-forces" the correct naming with the cycle in the template, but assuming that range will always be max (and min) two fields, which it wouldn't make sense if it was less or more (it's a range after all, right? ;') )

### Screenshots

image

### Pull Request Checklist

### DIFF
--- a/saleor/dashboard/widgets.py
+++ b/saleor/dashboard/widgets.py
@@ -1,25 +1,25 @@
 from django import forms
 from django.conf import settings
 from django.forms import Textarea, TextInput
-from django_filters.widgets import RangeWidget
+from django_filters import widgets
 from django_prices.widgets import MoneyInput
 
 from ..account.widgets import PhonePrefixWidget as StorefrontPhonePrefixWidget
 
 
-class DateRangeWidget(RangeWidget):
+class DateRangeWidget(widgets.DateRangeWidget):
     def __init__(self, attrs=None):
-        widgets = (forms.DateInput, forms.DateInput)
+        date_widgets = (forms.DateInput, forms.DateInput)
         # pylint: disable=bad-super-call
-        super(RangeWidget, self).__init__(widgets, attrs)
+        super(widgets.RangeWidget, self).__init__(date_widgets, attrs)
 
 
-class MoneyRangeWidget(RangeWidget):
+class MoneyRangeWidget(widgets.RangeWidget):
     def __init__(self, attrs=None):
         self.currency = getattr(settings, 'DEFAULT_CURRENCY')
-        widgets = (MoneyInput(self.currency), MoneyInput(self.currency))
+        money_widgets = (MoneyInput(self.currency), MoneyInput(self.currency))
         # pylint: disable=bad-super-call
-        super(RangeWidget, self).__init__(widgets, attrs)
+        super(widgets.RangeWidget, self).__init__(money_widgets, attrs)
 
 
 class PhonePrefixWidget(StorefrontPhonePrefixWidget):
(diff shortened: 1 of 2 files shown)


saleor/dashboard/widgets.py
```diff
@@ -1,25 +1,26 @@
 from django import forms
 from django.conf import settings
 from django.forms import Textarea, TextInput
-from django_filters.widgets import RangeWidget
+from django_filters.widgets import (
+    DateRangeWidget as df_DateRangeWidget, RangeWidget as df_RangeWidget)
```
maarcingebala: Just a small styling issue - I think `df_DateRangeWidget` and `df_RangeWidget` look bit "unpythonic". To avoid that we could as well do `from django_filters import widgets` and below use it like `widgets. DateRangeWidget` etc., what do you think?

templates/materializecssform/field.html
```diff
@@ -258,8 +258,9 @@
       {% endif %}
     </label>
     {% for value in field.value %}
-      <input type="date" id="{{ field.auto_id }}" class="datepicker col s5" name="{{ field.name }}_{{ forloop.counter0 }}"
-        value="{% if field.value %}{{ field.value|date:'Y-m-d' }}{% endif %}" data-value="{% if field.value %}{{ field.value|date:'Y-m-d' }}{% endif %}">
+      {% comment %} cycle after/before - this is assuming range will only iterate over 2 fields, required for django-filter instead of 0-1 indexes {% endcomment %}
+      <input type="date" id="{{ field.auto_id }}" class="datepicker col s5" name="{{ field.name }}_{% cycle 'after' 'before' %}"
```
maarcingebala: Just out of curiosity - have you tried to change these suffixes to ["0", "1"] in our widgets that override those from `django-filters`? It would be easier to control that from Python than from templates, but it's fine if that's the only way.

mjelonek92: That would work as well, the end result would be the same. Could make a base class that specifies the suffixes and the other widgets subclass it, but imo price_min in the url looks better than price_0 :P and holds more meaning

maarcingebala: True about the URLs. But all right, we can keep it in templates as well 👍

saleor/dashboard/widgets.py
```diff
@@ -1,25 +1,25 @@
...
 
-class MoneyRangeWidget(RangeWidget):
+class MoneyRangeWidget(widgets.RangeWidget):
     def __init__(self, attrs=None):
         self.currency = getattr(settings, 'DEFAULT_CURRENCY')
-        widgets = (MoneyInput(self.currency), MoneyInput(self.currency))
+        money_widgets = (MoneyInput(self.currency), MoneyInput(self.currency))
         # pylint: disable=bad-super-call
-        super(RangeWidget, self).__init__(widgets, attrs)
+        super(widgets.RangeWidget, self).__init__(money_widgets, attrs)
```
NyanKiyoshi: I wonder why we are using the old-style super call and disabled `bad-super-call`? From the `git blame`, it was written in the late 2017, and the the warning was disabled in Feb. 2018. Maybe Python 2 wasn't dropped yet?

patrys: We're explicitly skipping one level of inheritance here.

PR #3392, opened 2018-12-05
Add vat tax rates to country

michaljelonek:
Resolves https://github.com/mirumee/saleor/issues/3383.

Added vat tax rate information when querying for countries. Instead of proposed solution to change country from String! to object I decided it's better to put tax information next to it. Additionally added tax rate to default country. Other places where country is used didn't seem like needed this information. 

Example query:
```
{
  shop {
    countries {
      country
      vat {
        standardRate
        reducedRates {
          rate
          rateType
        }
      }
    }
  }
}
```

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/shop/types.py
+++ b/saleor/graphql/shop/types.py
@@ -4,6 +4,7 @@ import graphene
 import graphene_django_optimizer as gql_optimizer
 from django.conf import settings
 from django_countries import countries
+from django_prices_vatlayer.models import VAT
 from graphql_jwt.decorators import permission_required
 from phonenumbers import COUNTRY_CODE_TO_REGION_CODE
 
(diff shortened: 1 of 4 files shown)


saleor/graphql/shop/types.py
```diff
@@ -117,8 +118,10 @@ def resolve_authorization_keys(self, info):
         return site_models.AuthorizationKey.objects.all()
 
     def resolve_countries(self, info):
+        taxes = {vat.country_code: vat for vat in VAT.objects.all()}
```
akjanik: This looks like something that could be extracted outside of the resolver, to be calculated once, not every time query involving countries is being executed.

maarcingebala: We could leave it as it is - it's a single database query to fetch a constant number of items (28). Also building this relatively small dictionary is rather efficient in Python and shouldn't impact the performance. I'd instead debug the entire `shop` query in our client apps and see how fast it is or how many queries in total it performs.

PR #3398, opened 2018-12-06
Do not show Pay For Order if order is partly paid since partial payment is not supported

jxltom:
Since saleor does not support partial payment (from https://github.com/mirumee/saleor/blob/master/saleor/graphql/payment/mutations.py#L75 and https://github.com/mirumee/saleor/blob/master/saleor/payment/gateways/braintree/forms.py#L31), it's better to do not show ```Pay For Order``` in order's detail page if a order is partly paid. Othewise user may pay twice with total amount if a payment is partly paid such as when payment is partly refunded, though it will finally raise a ```Charged transactions cannot be authorized again``` error by ```clean_authorize```.

### Pull Request Checklist


maarcingebala:
@jxltom Does the issue affect the API / Dashboard 2.0 as well? If yes, we could fix it in this PR but if you don't have time and the issue occurs over there, just open a new issue so we don't forget about it.

We've recently added `CHANGELOG.md` file, could to you update it (and all your other PRs well)? Thanks!

PR #3408, opened 2018-12-07
Add stripe payment gateway

jxltom:
This PR adds stripe payment gateway to saleor.

- [x] The code is ready
- [x]  Tests and docs are added

### Screenshots

Here is the screenshot with stripe gateway's default configurations:
screenshot from 2018-12-07 20-51-19

### Pull Request Checklist


saleor/payment/gateways/stripe/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+def get_form_class():
+    return StripePaymentModalForm
+
+
+def _get_client(**connection_params):
+    stripe.api_key = connection_params.get('secret_key')
+    return stripe
+
+
+def _create_stripe_charge(client, payment, amount, payment_token, capture):
```
NyanKiyoshi: In `payment.gateways.stripe`, could you define the private functions at the top of the file? It's most of the time the convention (not sure about Python), but mostly because when reading `authorize`, `capture`, etc. we call private functions, and we expect to have read them already or else, to find them a bit above (or in the imports).

jxltom:
No, it is basically a porting from django-payments with several minor enhancements.

saleor/payment/gateways/stripe/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+    currency = response.get('currency', payment.currency)
+
+    # Get amount from response or payment
+    if 'amount' in response:
+        amount = get_amount_from_stripe(response.get('amount'), currency)
+
+    # Get token, which is an empty string for error responses
+    token = response.get('id', '')
+
+    # Check is_uccess from response
```
NyanKiyoshi: Typo, and maybe we could reword to this?
```suggestion
    # Check if the response's status is flagged as succeeded
```

saleor/payment/gateways/stripe/forms.py
```diff
@@ -0,0 +1,74 @@
...
+            'data-description': CHECKOUT_DESCRIPTION,
+            'data-currency': payment.currency,
+            'data-locale': gateway_params.get('local'),
+            'data-allow-remember-me': gateway_params.get('remember_me'),
+            'data-billing-address': 'true' if gateway_params.get(
+                'enable_billing_address') else 'false',
+            'data-zip-code': 'true' if gateway_params.get(
+                'enable_billing_address') else 'false',
+            'data-shipping-address': 'true' if gateway_params.get(
+                'enable_shipping_address') else 'false',
```
NyanKiyoshi: Note: there is a hanging comma here (against Saleor's code style).

saleor/payment/gateways/stripe/__init__.py
jxltom: Thanks, sounds better!

saleor/payment/gateways/stripe/forms.py
jxltom: Thanks, all the hanging comma in this PR are removed

saleor/payment/gateways/stripe/__init__.py
jxltom: Thanks for review, and thanks for your implementation on razorpay, from which I've extracted some patterns and used in stripe. :smile:

jxltom: As the the issue, for c/c++, these functions should definitely be placed at the top. But for python, IMHO it is only a matter of personal choice. There is no standard, and we only need to stick to a specific style.

As for me, I prefer these private functions at the bottom since I actually only would like to expose the public APIs. The readers don't need to know internal implementation (such as these private methods). If they really would like to examine the implementation, scroll down or IDE navigation would do that.

And for stripe, I didn't see any explicit stardand for that. So I would like to still keep this.

stephenmoloney:
Pretty happy to see a stripe payments PR. 

@jxltom 
Is this using `stripe-elements`, does the credit card data never touch the saleor server - are there some stripe docs I could read somewhere that show the flow that this PR follows?

jxltom:
@stephenmoloney The frontend is not using stripe-element currently, though you can implement it easiely no matter in saleor or saleor-storefront. The frontend is using stripe's ```checkout.js```, tokenize credit card information, and pass it to saleor's backend API via ```payment_token```. The sensitive card info will never touch the backend and then it is PCI compliant.

For more info about ```checkout.js```, you could refer to https://stripe.com/docs/checkout. Basically it is a convenient wrapper of ```stripe.js```.

stephenmoloney:
thanks @jxltom

jxltom:
Update: The tests and docs are added. The only issue in codeclimate can be marked as WONTFIX.

This PR is ready for review. @maarcingebala @patrys @Pacu2 @NyanKiyoshi or whoever, reviews or any comments/thoughts/concerns are appreciated, thanks!

docs/payment-gateways/stripe.rst
```diff
@@ -0,0 +1,20 @@
...
+    ================================== ======================================================================================================
+    Environment variable               Description
+    ================================== ======================================================================================================
+    ``STRIPE_PUBLIC_KEY``              Your Stripe's public key (test or live)
+    ``STRIPE_SECRET_KEY``              Your Stripe's secret key (test or live)
+    ``STRIPE_STORE_NAME``              Your store name to show in the checkout form
+    ``STRIPE_STORE_IMAGE``             An absolute or relative link of your store logo to show in the checkout form
+    ``STRIPE_PREFILL``                 Prefill the email adddress in the checkout form if set to ``True`` (default)
+    ``STRIPE_REMEMBER_ME``             Add "Remember Me" for future purchases in the checkout form if set to ``True`` (default)
+    ``STRIPE_LOCAL``                   Specify auto to display checkout form in the user's preferred language (default)
```
NyanKiyoshi: I would suggest to do this, to be more straightforward:
```suggestion
    ``STRIPE_LOCAL``                   Specify ``auto`` to display checkout form in the user's preferred language (default)
```

docs/payment-gateways/stripe.rst
```diff
@@ -0,0 +1,20 @@
...
+
+This gateway implements payments using `Stripe <https://stripe.com/>`_.
+
+.. table::
+
+    ================================== ======================================================================================================
+    Environment variable               Description
+    ================================== ======================================================================================================
+    ``STRIPE_PUBLIC_KEY``              Your Stripe's public key (test or live)
+    ``STRIPE_SECRET_KEY``              Your Stripe's secret key (test or live)
```
NyanKiyoshi: The `'s` are sounding wrong to me.

tests/gateway/test_stripe.py
```diff
@@ -0,0 +1,648 @@
...
+
+@pytest.fixture()
+def stripe_charge_success_response():
+    return {
+        'id': TRANSACTION_TOKEN,
+        'amount': get_amount_for_stripe(
+            TRANSACTION_AMOUNT, TRANSACTION_CURRENCY),
+        "amount_refunded": 0,
+        'currency': get_currency_for_stripe(TRANSACTION_CURRENCY),
+        "status": "succeeded"}
```
NyanKiyoshi: Accidental double quotes here (and a bit above)

tests/gateway/test_stripe.py
```diff
@@ -0,0 +1,648 @@
...
+        stripe.error.StripeError(json_body=dict(message=ERROR_MESSAGE))
+    invalid_request_error = \
+        stripe.error.InvalidRequestError(message=ERROR_MESSAGE, param=None)
+
+    assert _get_error_response_from_exc(stripe_error) == \
+        dict(message=ERROR_MESSAGE)
+    assert _get_error_response_from_exc(invalid_request_error) == dict()
+
+
+def test_create_transaction(
```
NyanKiyoshi: We probably should split this into 4 tests, it would more readable and comprehensible, especially as the fixtures are not heavy on CPU time.

tests/gateway/test_stripe.py
```diff
@@ -0,0 +1,648 @@
...
+
+
+def test_get_error_response_from_exc():
+    stripe_error = \
+        stripe.error.StripeError(json_body=dict(message=ERROR_MESSAGE))
+    invalid_request_error = \
+        stripe.error.InvalidRequestError(message=ERROR_MESSAGE, param=None)
+
+    assert _get_error_response_from_exc(stripe_error) == \
+        dict(message=ERROR_MESSAGE)
```
NyanKiyoshi: Maybe using dict literals would be more readable? (same below)
```python
    assert _get_error_response_from_exc(stripe_error) == {
        'message': ERROR_MESSAGE}
```

tests/settings.py
```diff
@@ -27,5 +27,6 @@
 CHECKOUT_PAYMENT_GATEWAYS = {
     DUMMY: 'Dummy gateway',
     BRAINTREE: 'Braintree',
-    RAZORPAY: 'Razorpay'
+    RAZORPAY: 'Razorpay',
+    STRIPE: 'Stripe',
```
NyanKiyoshi: We should remove the hanging comma.

docs/payment-gateways/stripe.rst
jxltom: Thanks!

tests/gateway/test_stripe.py
jxltom: Thanks for the careful review! :smile:

tests/gateway/test_stripe.py
jxltom: Agree, dict literal is also a little bit faster

tests/gateway/test_stripe.py
jxltom: Indeed!

maarcingebala:
Thanks @jxltom!

saleor/payment/gateways/stripe/forms.py
```diff
@@ -0,0 +1,75 @@
...
+            'data-amount': get_amount_for_stripe(
+                payment.total, payment.currency),
+            'data-name': gateway_params.get('store_name'),
+            'data-image': gateway_params.get('store_image'),
+            'data-description': CHECKOUT_DESCRIPTION,
+            'data-currency': payment.currency,
+            'data-locale': gateway_params.get('locale'),
+            'data-allow-remember-me': 'true' if gateway_params.get(
+                'remember_me') else 'false',
+            'data-billing-address': 'true' if gateway_params.get(
```
NyanKiyoshi: Shouldn't we provide the billing address by default to prevent fraud and facilitate the dispute process? (even if the client can alter this data since it is client side only).

jxltom: Sadly, stripe.js or checkout.js of stripe doesn't support pre-fill the billing or shipping address. Enable this will add an extra step for the user to input billing/shipping address during checkout. The user actually has already provide the address in saleor's checkout process, and re-input this feels very annoying. That's why these options are falsy by default.

PR #3417, opened 2018-12-10
Get authorized amount from succeeded auth transactions instead of payment.total

jxltom:
The ```Preauthorized amount``` used in dashboard's order detail page is using ```payment.total```. But this may be wrong such as when a payment is created but ```authorize``` method fails. In this case, a payment did in ```NOT_CHARGED``` status but it is not acutally authorized since the authorization process is failed.

So this authorized amount should be calcuated from the succeeded auth transactions.

### Pull Request Checklist

### DIFF
--- a/saleor/order/models.py
+++ b/saleor/order/models.py
@@ -248,7 +248,7 @@ class Order(models.Model):
     def total_authorized(self):
         payment = self.get_last_payment()
         if payment:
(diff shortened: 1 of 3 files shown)


jxltom:
@maarcingebala Sure, working on it!

jxltom:
@maarcingebala The test is added! Thanks for review.

PR #3418, opened 2018-12-10
Improve several payment validations

jxltom:
There are several improvements with payment's validations:

- ```Payment.can_void``` should check if there is a existed succeeded AUTH transaction.
- ```Payment.can_capture``` should check if there is a existed succeeded AUTH transaction.
- ```not_fully_charged``` validation is removed for ```Payment.can_capture```. This is because saleor and most payment gateways do not support **multiple partial capture**. Check here https://github.com/mirumee/saleor/blob/d68063606888913654c2ec68e6f58ae87d132f86/saleor/dashboard/order/forms.py#L286. Actually in saleor ```capture``` requires ```NOT_CHARGED``` status, and once a partial charge is made, the payment status will transform to ```CHARGED```, this means multiple partial payment is impossible.
- ```Payment.can_charge``` should have different logic with ```Payment.can_capture```. This is because ```Payment.can_capture``` requires existed succeeded AUTH transaction but ```Payment.can_charge``` does not requires it (A charge can be made without authorization)
- ```clean_charge``` replaces ```clean_authorize``` and ```clean_capture```. This is because ```charge``` may have ```authorize``` and ```capture``` transactions together or have a ```charge``` transaction only. Its validation shouldn't validate for ```authorize``` and ```capture``` separately. Some payment gateways may only have unified charge transaction.
- Since multiple partial capture is not supported, authorized amount should returns zero money when capture is once performed. ~~This improvement is blocked by https://github.com/mirumee/saleor/pull/3417 and that PR should be merged first.~~

### Pull Request Checklist

### DIFF
--- a/saleor/payment/models.py
+++ b/saleor/payment/models.py
@@ -98,7 +98,8 @@ class Payment(models.Model):
     def get_authorized_amount(self):
         money = zero_money()
 
-        # Query all the transactions
+        # Query all the transactions which should be prefetched
+        # to optimize db queries
         transactions = self.transactions.all()
 
         # There is no authorized amount anymore when capture is succeeded
@@ -107,11 +108,15 @@ class Payment(models.Model):
                 and txn.is_success for txn in transactions]):
             return money
 
+        # Filter the succeeded auth transactions
+        authorized_txns = [
+            txn for txn in transactions
+            if txn.kind == TransactionKind.AUTH and txn.is_success]
+
         # Calculate authorized amount from all succeeded auth transactions
-        for txn in transactions:
-            if txn.kind == TransactionKind.AUTH and txn.is_success:
-                money += Money(
-                    txn.amount, self.currency or settings.DEFAULT_CURRENCY)
+        for txn in authorized_txns:
+            money += Money(
+                txn.amount, self.currency or settings.DEFAULT_CURRENCY)
 
         # If multiple partial capture is supported later though it's unlikely,
         # the authorized amount should exclude the already captured amount here


maarcingebala:
#3417 is ready, you can proceed with this one @jxltom.

jxltom:
@maarcingebala Thanks for taking care of that. Now this PR is ready for review,!

saleor/payment/models.py
```diff
@@ -99,14 +99,24 @@ def get_total(self):
...
+
         # Calculate authorized amount from all succeeded auth transactions
         for transaction in self.transactions.filter(
                 kind=TransactionKind.AUTH, is_success=True).all():
             money += Money(
                 transaction.amount, self.currency or settings.DEFAULT_CURRENCY)
 
         # The authorized amount should exclude the already captured amount
-        money -= self.get_captured_amount()
+        for transaction in self.transactions.filter(
```
michaljelonek: Will this actually run? In 106 you exit the function if there's an existing capture, but here you iterate over them?

jxltom: You catch it. It won't run indeed. It is just put here in case later multiple partial capture is supported though it is unlikely. It shows the correctly logic: total_authorized = authorized - captured. I think it is fine to keep it here though it adds one more db query. I've updated the comment in the code so the developers won't got confused.

jxltom: I changed my mind, I remove this part of code since it effects the test coverage... I've only left the comment in case multiple partial payment is supported later though it is unlikely

saleor/payment/models.py
```diff
@@ -151,19 +157,25 @@ def can_authorize(self):
             self.is_active and self.charge_status == ChargeStatus.NOT_CHARGED)
 
     def can_capture(self):
-        # FIXME should also have an auth transaction
         not_charged = self.charge_status == ChargeStatus.NOT_CHARGED
+        is_authorized = self.transactions.filter(
```
maarcingebala: Have you tested how these changes affect database performance? Each call to `transactions.filter()` or `exists()` will perform a database query (this applies to other changed lines in this PR as well). This may result in duplicated queries and slow down our views or API calls.

We could prevent that if we ensured that the transactions are prefetched at some point earlier and iterate over them here to take advantage of prefetch cache:
```
is_authorized = any([txn.kind == TransactionKind.AUTH and txn.is_success for txn in self.transactions.all()])
```
Something like that should do the trick.

jxltom: Very good suggestions, I will think about where to pretech the transactions and how to update the cache when new transaction is added.

maarcingebala: Usually, we do the prefetching in the views as this is most often the "entry point" to a particular flow and it starts with fetching necessary data from DB. Every other function/method that performs some logic could safely assume that data was already prefetched at this point.

jxltom: I've updated the PR. The db query is optimized by cached transactions, and the cached transactions will be invalidated when new transactions are added or deleted. Tests are also added. I've tested locally and the db queries looks fine. How about this approach?

saleor/payment/models.py
```diff
@@ -101,15 +108,15 @@ def get_authorized_amount(self):
 
         # There is no authorized amount anymore when capture is succeeded
         # since capture can only be made once, even it is a partial capture
-        if self.transactions.filter(
-                kind=TransactionKind.CAPTURE, is_success=True).exists():
+        if any([txn.kind == TransactionKind.CAPTURE
+                and txn.is_success for txn in self.cached_transactions]):
```
maarcingebala: We do it a bit differently in Saleor. At this point, we should assume that the transactions are prefetched and we can safely iterate over `self.transactions.all()`. E.g. I've checked that `get_authorized_amount` is only used in `resolve_total_authorized` (indirectly). We have an optimizer there that already prefetches payments:

```
@gql_optimizer.resolver_hints(prefetch_related='payments')
```

This is probably not enough and we should also prefetch transactions, something like this:
```
@gql_optimizer.resolver_hints(prefetch_related='payments__transactions')
```

Maybe an easier example is with `can_capture` - this function is used by `order. can_capture` which in turn is used for example by `dashboard:order-details` view. It means that this view should prefetch transactions [here](https://github.com/mirumee/saleor/blob/master/saleor/dashboard/order/views.py#L119).

If you look around the codebase, we don't use Django signals at all and this is on purpose, as we find them difficult to debug and error-prone. I'd change your implementation to only iterate over `self.transactions.all()`. I'm also fine with adding the prefetches in a separate PR.

jxltom: I see I see, thanks a lot for the explanation. I'm just curious why there is no signals at all in Saleor and that's why. 😄

jxltom: I've updated the PR and going to submit another PR later to support prefetch for transactions

maarcingebala:
Thanks a lot for your effort 👍

saleor/payment/models.py
```diff
@@ -99,14 +98,28 @@ def get_total(self):
...
+
+        # Filter the succeeded auth transactions
+        authorized_txns = [
+            txn for txn in transactions
+            if txn.kind == TransactionKind.AUTH and txn.is_success]
+
         # Calculate authorized amount from all succeeded auth transactions
-        for transaction in self.transactions.filter(
-                kind=TransactionKind.AUTH, is_success=True).all():
+        for txn in authorized_txns:
```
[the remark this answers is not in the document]
NyanKiyoshi: Even if we prevent a nested code?
```python
for txn in transactions:
    if not (txn.kind == TransactionKind.AUTH and txn.is_success):
        continue
    money += Money(
        txn.amount, self.currency or settings.DEFAULT_CURRENCY)
```

saleor/payment/models.py
```diff
@@ -151,19 +164,29 @@ def can_authorize(self):
...
     def can_void(self):
+        is_authorized = any([
+            txn.kind == TransactionKind.AUTH
+            and txn.is_success for txn in self.transactions.all()])
+
         return (
-            self.is_active and self.charge_status == ChargeStatus.NOT_CHARGED)
+            self.is_active
+            and self.charge_status == ChargeStatus.NOT_CHARGED
+            and is_authorized)
```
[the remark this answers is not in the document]
jxltom: Will do :D

saleor/payment/models.py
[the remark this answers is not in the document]
jxltom: Looks good, will try!

jxltom:
Hey @NyanKiyoshi, I've made some changes according to your suggestions in https://github.com/mirumee/saleor/pull/3468, please have a look. :p

saleor/payment/models.py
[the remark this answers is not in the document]
jxltom: The codeclimate still has warning for this. I'm going to leave it since this list iteration is not heavy actually.

PR #3442, opened 2018-12-12
[WIP] Payments refactor

michaljelonek:
This is still in development. However it is working for Braintree.

Opening so the progress can be tracked and suggestions made. Discussion can be found here: https://github.com/mirumee/saleor/issues/3422

The overall logic _should_ stay the same, just moved out of gateways back to Saleor, so the gateways can just talk with providers, normalize data and return it back to Saleor in some unified form. 

Not sure how to handle forms yet since they inherit from Saleors `PaymentForm` form.

### TODOs:
In no particular order
- [x] Move model logic from gateway (Braintree) to Saleor
- [x] Fix gateway response serialisation 
- [ ] Adjust other gateways to fit
- [ ] Make adjustments to payments to send/receive additional data to/from gateways
- [ ] Move Braintree to a separate package
- [ ] Fix tests and other stuff that needs fixing
- [ ] Update docs
- [ ] Add some proper exceptions, per package? _could be added later on_

### Pull Request Checklist


saleor/payment/gateways/braintree/__init__.py
```diff
@@ -33,23 +30,23 @@
 }
 
 
-def get_customer_data(payment: Payment) -> Dict:
+def get_customer_data(payment: Dict) -> Dict:
```
patrys: Both `payment` and the output of this look like a good candidate for either a `dataclass` or a `namedtuple`? Notice how many hardcoded string constants are there in the following lines.

saleor/payment/gateways/braintree/__init__.py
```diff
@@ -33,23 +30,23 @@
...
-            'first_name': payment.billing_first_name,
-            'last_name': payment.billing_last_name,
-            'company': payment.billing_company_name,
-            'postal_code': payment.billing_postal_code,
-            'street_address': payment.billing_address_1,
-            'extended_address': payment.billing_address_2,
-            "locality": payment.billing_city,
-            'region': payment.billing_country_area,
-            'country_code_alpha2': payment.billing_country_code},
+            'first_name': payment['billing_first_name'],
```
patrys: Why can't the payment dict have a sub-dict named `billing`?

michaljelonek: `payment` currently is just `model_to_dict(Payment)`, nothing fancy and those are normal fields of that model

saleor/payment/gateways/braintree/__init__.py
```diff
@@ -137,127 +115,85 @@ def get_client_token(**connection_params: Dict) -> str:
...
                 'submit_for_settlement': CONFIRM_MANUALLY,
                 'three_d_secure': {
                     'required': THREE_D_SECURE_REQUIRED}},
             **get_customer_data(payment)})
     except braintree_sdk.exceptions.NotFoundError:
-        return transaction_and_incorrect_token_error(
-            payment, kind=TransactionKind.AUTH, token=payment_token)
+        # FIXME: make this a payment generic exception
+        raise Exception('Unable to process the transaction. '
+            'Transaction\'s token is incorrect or expired.')
```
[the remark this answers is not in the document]
michaljelonek: Currently not, I was thinking what's the best approach here if the provider package throws an exception and my thinking was that the Saleor gateway implementation would handle those and re-raise exception to Saleor if need be. I did not look into specifics when Braintree raises an exception and when it returns just errors with `is_success = False`, but I guess returning instead of re-raising would serve the same purpose..

saleor/payment/gateways/braintree/__init__.py
michaljelonek: `payment` being a custom type might be tricky, because then we'll need a separate package to just have the custom types. I believe the point is to have gateways separated from Saleor, but if we introduce a new type *in* Saleor (`payment` here is just a dict produced by `model_to_dict(Payment)`) to provide to gateways then they'll need that type also.

saleor/payment/utils.py
```diff
@@ -158,11 +176,43 @@ def gateway_authorize(payment: Payment, payment_token: str) -> Transaction:
     clean_authorize(payment)
 
     gateway, gateway_params = get_payment_gateway(payment.gateway)
-    with transaction.atomic():
-        txn, error = gateway.authorize(
-            payment=payment, payment_token=payment_token, **gateway_params)
+    try:
+        is_success, gateway_response, error = gateway.authorize(
+            payment=model_to_dict(payment), payment_token=payment_token,
```
jxltom: This ```model_to_dict``` should return information of a payment as more as possible. For example, some gateways may requires the shipping address, but this shipping address is not saved as a normal field of the Payment but in ```Payment.order.shipping_address.street_address_1|street_address_2|etc```. This method should make sure these kind of information is properly returned and then passed to gateways.

saleor/payment/utils.py
```diff
@@ -207,25 +306,51 @@ def gateway_refund(payment, amount: Decimal) -> Transaction:
     """Refunds the charged funds back to the customer.
     Refunds can be total or partial.
     """
+    if not payment.can_refund():
+        raise PaymentError('This payment cannot be refunded.')
+
     if amount <= 0:
         raise PaymentError('Amount should be a positive number.')
-    if amount > payment.captured_amount:
+    elif amount > payment.captured_amount:
```
jxltom: IMO, using ```if``` is easier to understand since each ```if``` clause is ended with a ```raise```

saleor/payment/utils.py
```diff
@@ -130,21 +131,38 @@ def gateway_charge(
     clean_capture(payment, amount)
 
     gateway, gateway_params = get_payment_gateway(payment.gateway)
-    with transaction.atomic():
-        txn, error = gateway.charge(
+    try:
+        is_success, gateway_response, error = gateway.charge(
```
jxltom: I have a few concerns/suggestions for this:
- IMHO, for this ```gateway.charge``` as well as others ```gateway.authorize|capture|etc```, a wrapper for them can be added. This wrapper or layer can check whether the gateway is returning the correct format. With Saleor's payment evolves, the expected format of a gateway could be changed, and the third-party gateway repo could be outdated. It will be nice to check the gateway response before passing the gateway response to saleor. If the gateway response format is out-of-date, raise exceptions or warnings. This can solve a lot of issues in advance.
- This layer could also used for deprecation cycles. For example, if saleor payment is going to deprecate certain field from the gateway response such as the ```is_success``` field, the deprecation warning can done in this wrapper. 
- Returning ```is_success, gateway_response, error``` probably is not the best way to get the gateway response. Maybe a dict or namedtuple? So extra result could be passed back to saleor without raising errors. 
 Otherwise one minor extra return may break the whole payment system.
- Since ```gateway_response``` is used directly to create transactions which means this is not the raw gateway (I mean paypal, stripe or whatever as gateway here) response. The transactions in saleor should used saleor's standard of ```amount``` or ```currency```, but the raw gateway response may not have these kind of information directly. Some gateway may not return ```currency``` at all. So this ```gateway_response``` is processed response from the providers (I means saleor-payment-paypal or saleor-payment-stripe and whatever here as providers), what is the standard of the gateway response? Which fields must it include? It will be nice to define it.
- It will be nice to add a new field such as ```raw_gateway_response``` to ```Transaction``` model. This field could save the raw gateway response. This helps debugging. Sometimes due to bugs, the postprocessed ```gateway_response``` could be the wrong data.

michaljelonek:
Closing this for now as I stumbled upon a problem that needs more consideration. Those payments are pretty complex how they are handled currently and it's not that simple to just take the gateway out, so there's no point for the PR to sit idly. Thanks for the review @patrys @jxltom, definitely will keep them in mind, so don't think that you're getting ignored :)

jxltom:
I'm curious what is the main issue? This PR seems fine.

michaljelonek:
@jxltom main issue was that this wasn't really thought through. I tried to maintain the current flows that the gateways did and that kinda didn't work as it led to problems that some gateways didn't support certain actions. It's fine right now, because the gateways can define their own flow in the form their provide, but that form was using Saleors methods, which we don't want. Gateway should be a separate being that exposes some functions that Saleor calls and gateway can handle what it does in itself, preferably without involving Saleor during the process, and then just returns some unified data that is required and sent for example to refund, so the gateway can complete that refund. Hope that makes sense?

jxltom:
I thought every gateway would have at least authorize+capture+void or charge+refund actions. If so, it is fine that we expose only these four APIs and the users could call them by any combinations from the frontend, as they used to do in ```process_payment``` in the forms.

PR #3478, opened 2018-12-18
Change filed type to choice field of tax rate in product

### LINKED ISSUE #2829: TaxRate on the Product model should be a choice field
We're missing `choices` on `tax_rate` field for `Product` model. Should be set the same way as on the ProductType.

fowczarek:
I want to merge this change because resolve #2829 

### Pull Request Checklist


saleor/graphql/schema.graphql
```diff
@@ -1448,6 +1448,33 @@ enum ProductOrderField {
   DATE
 }
 
+enum ProductTaxRate {
```
maarcingebala: We already have the `TaxRateType` in the schema. The difference is that `TaxRateType` is implemented explicitly by us, while this one is generated by Graphene after introducing `choices` in `tax_rate` field (I guess). We should only keep one of them in the schema. I think the `TaxRateType` would be easier to drop but the question is how can we use the `ProductTaxRate` in other places.

fowczarek: Is it good now?

maarcingebala: Looks good now! We're changing the type here and since we keep the same name of the field, we cannot deprecate it. This means that we have to pass this PR to @dominik-zeglen to fix references to `taxRate` field in the Dashboard 2.0.

dominik-zeglen:
Front end reports full okayness.

saleor/graphql/product/types.py
```diff
@@ -286,6 +286,7 @@ class Product(CountableDjangoObjectType):
         Money,
         description=dedent("""The product's base price (without any discounts
         applied)."""))
+    tax_rate = TaxRateType(description='A type of tax rate.')
```
NyanKiyoshi: The `a type of...` is sounding weird to me.
What do you think of: `The tax rate type.` ?

NyanKiyoshi: Welp @maarcingebala just merged it :laughing:

PR #3505, opened 2018-12-21
Add default variant create to Product mutations

### LINKED ISSUE #2724: Mutation to create products doesn't create the default variant

fowczarek:
I want to merge this change because resolve #2724

### Pull Request Checklist


saleor/graphql/product/mutations/products.py
```diff
@@ -312,6 +323,21 @@ def clean_input(cls, info, instance, input, errors):
...
+        if not instance.product_type.has_variants:
+            site_settings = info.context.site.settings
+            track_inventory = cleaned_input.get(
+                'track_inventory', site_settings.track_inventory_by_default)
+            quantity = cleaned_input.get('quantity', 0)
+            sku = cleaned_input.get('sku')
+            variant = models.ProductVariant(
+                product=instance, track_inventory=track_inventory,
+                sku=sku, quantity=quantity)
+            variant.save()
```
Pacu2: How about using `create()` instead?

saleor/graphql/product/mutations/products.py
```diff
@@ -312,6 +323,21 @@ def clean_input(cls, info, instance, input, errors):
...
+    @classmethod
+    @transaction.atomic
+    def save(cls, info, instance, cleaned_input):
+        instance.save()
+        if not instance.product_type.has_variants:
+            site_settings = info.context.site.settings
+            track_inventory = cleaned_input.get(
+                'track_inventory', site_settings.track_inventory_by_default)
+            quantity = cleaned_input.get('quantity', 0)
+            sku = cleaned_input.get('sku')
```
Pacu2: Is there a case when `sku` might be missing in the `cleaned_input`?

saleor/graphql/product/mutations/products.py
```diff
@@ -271,6 +272,16 @@ class ProductInput(graphene.InputObjectType):
     seo = SeoInput(description='Search engine optimization fields.')
     weight = WeightScalar(
         description='Weight of the Product.', required=False)
+    sku = graphene.String(description=dedent("""Stock keeping unit. Note: this
+    field is required only used for products without variants."""))
+    quantity = graphene.Int(
+        description=dedent("""The total quantity of this variant availablefor
```
Pacu2: Missing space between `availablefor`

saleor/graphql/product/mutations/products.py
```diff
@@ -271,6 +272,16 @@ class ProductInput(graphene.InputObjectType):
     seo = SeoInput(description='Search engine optimization fields.')
     weight = WeightScalar(
         description='Weight of the Product.', required=False)
+    sku = graphene.String(description=dedent("""Stock keeping unit. Note: this
```
Pacu2: I think we can work on this copy

saleor/graphql/product/mutations/products.py
```diff
@@ -334,6 +360,20 @@ class Meta:
...
+    @transaction.atomic
+    def save(cls, info, instance, cleaned_input):
+        instance.save()
+        if not instance.product_type.has_variants:
+            variant = instance.variants.first()
+            if 'track_inventory' in cleaned_input:
+                variant.track_inventory = cleaned_input.get('track_inventory')
+            if 'quantity' in cleaned_input:
+                variant.quantity = cleaned_input.get('quantity')
+            if 'sku' in cleaned_input:
```
Pacu2: I think that this whole logic should live in `clean_input` instead, ane here we should just assign the values to `variant` instance.

saleor/graphql/product/mutations/products.py
maarcingebala: Stock keeping unit of a product. Note: this field is only used if a product doesn't use variants.

saleor/graphql/product/mutations/products.py
maarcingebala: The total quantity of a product available for sale. Note: this field is only used if a product doesn't use variants.

saleor/graphql/product/mutations/products.py
```diff
@@ -271,6 +272,16 @@ class ProductInput(graphene.InputObjectType):
     seo = SeoInput(description='Search engine optimization fields.')
     weight = WeightScalar(
         description='Weight of the Product.', required=False)
+    sku = graphene.String(description=dedent("""Stock keeping unit. Note: this
+    field is required only used for products without variants."""))
+    quantity = graphene.Int(
+        description=dedent("""The total quantity of this variant availablefor
+        sale. Note: this field is only used for products without variants."""))
+    track_inventory = graphene.Boolean(
+        description=dedent("""Determines if the inventory of this variant
```
maarcingebala: Determines if the inventory of this **product** [...] - I think using the word "product" in this context will be more understandable.

saleor/graphql/product/mutations/products.py
```diff
@@ -271,6 +272,16 @@ class ProductInput(graphene.InputObjectType):
...
+        should be tracked. If false, the quantity won't change when customers
+        buy this item. Note: this field is only used for products
```
maarcingebala: Note: this field is only used if a product doesn't use variants.

saleor/graphql/product/mutations/products.py
```diff
@@ -334,6 +372,29 @@ class Meta:
         description = 'Updates an existing product.'
         model = models.Product
 
+    @classmethod
+    def clean_sku(cls, product_type, cleaned_input, errors):
+        pass
```
maarcingebala: Let's add a comment here as well to explain why we explicitly skip this logic. E.g: "SKU is an optional field in the product update mutation, so we explicitly skip the validation."

saleor/graphql/product/mutations/products.py
```diff
@@ -334,6 +372,29 @@ class Meta:
...
+
+    @classmethod
+    @transaction.atomic
+    def save(cls, info, instance, cleaned_input):
+        instance.save()
+        if not instance.product_type.has_variants:
+            variant = instance.variants.first()
+            update_fields = []
+            if 'track_inventory' in cleaned_input:
+                variant.track_inventory = cleaned_input.get('track_inventory')
```
maarcingebala: Here and similarly in the lines below we can just write:
```
variant.track_inventory = cleaned_input['track_inventory']
```
No need to do `get` since at this point we know that this key is present in the dict.

maarcingebala:
@fowczarek Please generate a new `schema.graphql` file.

jxltom:
@fowczarek @maarcingebala The pipenv part looks good to me except that ```kombu = "*"``` probably is unnecessary. Since kombu is only a dependency of celery and we don't rely on it directly. There is no need to pin it in Pipfile. There is an issue with upstream https://github.com/celery/kombu/issues/975 and we only need to regenerate the lockfile by ```pipenv lock``` or ```pipenv lock --clear``` if you would like to clean the caches.

jxltom:
@fowczarek Nice, now the pipenv part is great!

maarcingebala:
@jxltom Thanks for reviewing this part!

saleor/graphql/product/mutations/products.py
```diff
@@ -311,8 +324,35 @@ def clean_input(cls, info, instance, input, errors):
...
+    def clean_sku(cls, product_type, cleaned_input, errors):
+        """Validate SKU input field.
+
+        When creating products that don't use variants, SKU is required in
+        the input in order to create the default variant underneath.
+        See the documentation for `has_variants` field for details:
+        http://docs.getsaleor.com/en/latest/architecture/products.html#product-types
+        """
+        if not product_type.has_variants and not cleaned_input.get('sku'):
+            cls.add_error(errors, 'sku', 'This field cannot be blank.')
```
Pacu2: I think that we should validate if SKU is unique here as well, unless `clean_input` or `clean_instance` is already checking it.

saleor/graphql/product/mutations/products.py
Pacu2: I'd access the key directly. Sku is required and should be always present

Issue #3511, opened 2018-12-22
PopulateDB currency issue


stephenmoloney:
### What I'm trying to achieve

To run populatedb after running migrations.

### Steps to reproduce the problem

1. `python manage.py migrate` (WORKS - thanks to [fix](https://github.com/mirumee/saleor/pull/3506))
2. `python manage.py populatedb`

Then an error is generated as follows:

```text
Shipping Zone: Europe
Shipping Zone: Oceania
Shipping Zone: Asia
Shipping Zone: Americas
Shipping Zone: Africa
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 502, in update_or_create
    obj = self.select_for_update().get(**lookup)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 399, in get
    self.model._meta.object_name
saleor.product.models.DoesNotExist: Product matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 316, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 353, in execute
    output = self.handle(*args, **options)
  File "/app/saleor/core/management/commands/populatedb.py", line 83, in handle
    create_products_by_schema(self.placeholders_dir, create_images)
  File "/app/saleor/core/utils/random_data.py", line 227, in create_products_by_schema
    placeholder_dir=placeholder_dir, create_images=create_images)
  File "/app/saleor/core/utils/random_data.py", line 187, in create_products
    product, _ = Product.objects.update_or_create(pk=pk, defaults=defaults)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 506, in update_or_create
[...]

PR #3512, opened 2018-12-22
Fix for populatedb and currency issue

stephenmoloney:
***What does this commit/MR/PR do?***

- Changes the currency of the demo objects to the default currency

***Why is this commit/MR/PR needed?***

- So that populateDB works

***Relevant issues:***

#3511 

### Pull Request Checklist

### DIFF
--- a/saleor/core/utils/random_data.py
+++ b/saleor/core/utils/random_data.py
@@ -183,6 +183,8 @@ def create_products(products_data, placeholder_dir, create_images):
         defaults['weight'] = get_weight(defaults['weight'])
         defaults['category_id'] = defaults.pop('category')
         defaults['product_type_id'] = defaults.pop('product_type')
+        defaults['price'] = get_in_default_currency(
+            defaults, 'price', settings.DEFAULT_CURRENCY)
         defaults['attributes'] = json.loads(defaults['attributes'])
         product, _ = Product.objects.update_or_create(pk=pk, defaults=defaults)
 
@@ -203,7 +205,17 @@ def create_product_variants(variants_data):
             continue
         defaults['product_id'] = product_id
         defaults['attributes'] = json.loads(defaults['attributes'])
+        defaults['price_override'] = get_in_default_currency(
+            defaults, 'price_override', settings.DEFAULT_CURRENCY)
+        defaults['cost_price'] = get_in_default_currency(
+            defaults, 'cost_price', settings.DEFAULT_CURRENCY)
         ProductVariant.objects.update_or_create(pk=pk, defaults=defaults)
+
+
+def get_in_default_currency(defaults, field, currency):
+    if field in defaults and defaults[field] is not None:
+        return Money(defaults[field].amount, currency)
+    return None
 
 
 def create_products_by_schema(placeholder_dir, create_images):


stephenmoloney:
@Pacu2 could you take a look at this. I'm not a python developer.

I haven't tested this yet at all either. But something along these lines might fix #3511 .

saleor/core/utils/random_data.py
```diff
@@ -210,6 +210,9 @@ def create_products_by_schema(placeholder_dir, create_images):
     path = os.path.join(settings.PROJECT_ROOT, 'saleor', 'static', 'db.json')
     with open(path) as f:
         db_items = json.load(f, object_hook=object_hook)
+        for item in db_items:
+            if item["model"] in ["product.product", "product.productvariant"]:
```
Pacu2: Could we move this logic to `create_products` and `create_product_variants` functions?

stephenmoloney: Sure. good suggestion - gonna try and rebase

stephenmoloney:
Any ideas to overcome this error - 
https://travis-ci.org/mirumee/saleor/jobs/471641544#L1500

I'm not familiar with python types

saleor/core/utils/random_data.py
```diff
@@ -183,6 +183,7 @@ def create_products(products_data, placeholder_dir, create_images):
         defaults['weight'] = get_weight(defaults['weight'])
         defaults['category_id'] = defaults.pop('category')
         defaults['product_type_id'] = defaults.pop('product_type')
+        defaults['price'] = Money(defaults['price']['amount'], settings.DEFAULT_CURRENCY)
```
Pacu2: I believe that we have `Money` instance under the `defaults['price']` so it should look like:
```
defaults['price'] = Money(
    defaults['price'].amount, settings.DEFAULT_CURRENCY)
```

saleor/core/utils/random_data.py
```diff
@@ -203,6 +204,7 @@ def create_product_variants(variants_data):
             continue
         defaults['product_id'] = product_id
         defaults['attributes'] = json.loads(defaults['attributes'])
+        defaults['price'] = Money(defaults['price']['amount'], settings.DEFAULT_CURRENCY)
```
Pacu2: We have no `price` on the `ProductVariant`, instead we should handle `cost_price` and `price_override`. 
Both of them are nullable, so  we should handle it as well.
Something like:
```
if defaults['price_override']:
    defaults['price_override'] = Money(
        defaults['price_override'].amount, settings.DEFAULT_CURRENCY)
```
Same for the other field

stephenmoloney: @Pacu2 thanks!

saleor/core/utils/random_data.py
```diff
@@ -203,6 +205,9 @@ def create_product_variants(variants_data):
             continue
         defaults['product_id'] = product_id
         defaults['attributes'] = json.loads(defaults['attributes'])
+        if defaults['price_override']:
```
Pacu2: Do you mind adding same logic for the `cost_price` field?

stephenmoloney: i missed that - will do

stephenmoloney:
@Pacu2 
I ran into a code complexity error, so I've externalized the logic into an intermediate function.
Again, I'm not a python guy - maybe you can have a look.
https://codeclimate.com/github/mirumee/saleor/pull/3512#

image

saleor/core/utils/random_data.py
```diff
@@ -203,9 +204,17 @@ def create_product_variants(variants_data):
             continue
         defaults['product_id'] = product_id
         defaults['attributes'] = json.loads(defaults['attributes'])
+        reset_currency(defaults, 'price_override', settings.DEFAULT_CURRENCY)
```
Pacu2: I think I'd opt for more explicit version, like:
```
defaults['price_override'] = get_in_default_currency(
    defaults['price_override'])
```
I don't like the idea of modifying the dictionary under the hood

stephenmoloney: YEa, I had the same doubt at the time, will implement suggested changes later.

stephenmoloney: should be fixed now.

saleor/core/utils/random_data.py
```diff
@@ -203,9 +205,19 @@ def create_product_variants(variants_data):
...
         defaults['attributes'] = json.loads(defaults['attributes'])
+        defaults['price_override'] = get_in_default_currency(
+            defaults, 'price_override', settings.DEFAULT_CURRENCY)
+        defaults['cost_price'] = get_in_default_currency(
+            defaults, 'cost_price', settings.DEFAULT_CURRENCY)
         ProductVariant.objects.update_or_create(pk=pk, defaults=defaults)
 
 
+def get_in_default_currency(defaults, field, currency):
+    if dict.get(defaults, field):
```
maarcingebala: It's rather uncommon in Python to call the dict methods like that. Instead, you could simply write:
```
if field in dict:
```

saleor/core/utils/random_data.py
```diff
@@ -203,9 +205,19 @@ def create_product_variants(variants_data):
...
+        return Money(getattr(dict.get(defaults, field), 'amount'), currency)
```
maarcingebala: At this point, we know that `field` is present in `defaults`. Also, there's no need to do `getattr` as we will always have a `Money` type here. A Pythonic way to write this would be:
```
return Money(defaults[field].amount), currency)
```

stephenmoloney: @maarcingebala @NyanKiyoshi 

I have a syntactical error:

```text
E       return Money(defaults[field].amount), currency)
E                                                     ^
E   SyntaxError: invalid syntax
```

Can you please advise me how to correct it - thanks !

```python
def get_in_default_currency(defaults, field, currency):
    if field in defaults:
        return Money(defaults[field].amount), currency)
    return None
```

maarcingebala: Comma after `amount` should is unnecessary, it should be:
`return Money(defaults[field].amount, currency)`

stephenmoloney: oops, i missed that - you mean the parenthesis. thanks.

maarcingebala: Yep, sorry - I was thinking of that closing bracket 😄.

saleor/core/utils/random_data.py
```diff
@@ -183,6 +183,8 @@ def create_products(products_data, placeholder_dir, create_images):
         defaults['weight'] = get_weight(defaults['weight'])
         defaults['category_id'] = defaults.pop('category')
         defaults['product_type_id'] = defaults.pop('product_type')
+        defaults['price'] = get_in_default_currency(
```
patrys: Why do this workaround instead of fixing how defaults are generated?

stephenmoloney: It would be good if the defaults were generated with the right currency - they seem to be coming from a static json file. Right now I've only got time and sufficient python to offer this workaround.

PR #3518, opened 2018-12-27
Remove volume on celery compose

sergioisidoro:
In docker compose, `web` does not have a volume mounted (because the /app is in the container through Dockerfile)

However, `celery` has  a volume mounted, which caused me problems (I don't have the full codebase in the place I run docker compose, eg `docker-compose -f path/to/docker-compose.yml`) and seems to cause some other problems: https://github.com/mirumee/saleor/issues/3465 -- I imagine that windows has trouble with mounting that volume.

I don't know of any use-case for mounting that volume other than for development environments. But if that would be the case, `web` in the docker-compose file would also have it mounted.

So here's a proposal to drop that.
Alternative is to add that same volume in `web` and assume that the docker compose is a development environment.
 
--- 

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,8 +51,6 @@ services:
     restart: unless-stopped
     networks:
       - saleor-backend-tier
-    volumes:
-      - .:/app:Z
     env_file: common.env
     depends_on:
       - redis
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,8 +10,6 @@ services:
     restart: unless-stopped
     networks:
       - saleor-backend-tier
-    ports:
-      - 8000:8000
     env_file: common.env
     depends_on:
       - db
@@ -23,10 +21,6 @@ services:
     restart: unless-stopped
     networks:
       - saleor-backend-tier
-    volumes:
-      - saleor-db:/var/lib/postgresql
-    ports:
-      - 5432:5432
     environment:
       - POSTGRES_USER=saleor
       - POSTGRES_PASSWORD=saleor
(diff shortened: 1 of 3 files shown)


patrys:
It's a mess. For development both (`web` and `celery`) should have the volume mount, for production none of them should. I think we should work on getting #3093 to a mergeable state.

sergioisidoro:
Would the `docker-compose.yml` be the production with overrides for development or the other way around? 

I've seen somewhere that production settings should override the development ones, but on the other hand overriding volumes to be empty looks a bit clumsy. 

`docker-compose.override.yml`
```
...
    ports:
      - 8000:8000
    volumes:    <--- This looks clumsy
    env_file: common.env
```

Should we merge this one, and iterate the override? 
How can I help?

patrys:
I think production-by-default is the usual way to go.

sergioisidoro:
Shall we merge this one then and tackle the docker compose override in another PR?

sergioisidoro:
Ok, I've picked up the work done by @tiangolo and used the same principles to move the volume mounts to the override. Also updated the Docs. :)

Sorry @tiangolo for not cherry picking your commits, but since it had stale for 2 months it was easier to copy the changes you made to your docs. Hope you don't mind.

docker-compose.yml
```diff
@@ -23,10 +21,6 @@ services:
     restart: unless-stopped
     networks:
       - saleor-backend-tier
-    volumes:
-      - saleor-db:/var/lib/postgresql
-    ports:
-      - 5432:5432
     environment:
       - POSTGRES_USER=saleor
```
sergioisidoro: Maybe moving these also to override would not be a bad idea, since these are dev env variables.

maarcingebala:
Please see PR #3657 which was partially based on your work. Thanks!

PR #3523, opened 2018-12-28
Extract enums

maarcingebala:
GraphQL maintenance - this PR extracts all enums to separate `enums.py` files within each graphql module. Also, I've moved some scalars to their own files.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/discount/schema.py
+++ b/saleor/graphql/discount/schema.py
@@ -2,12 +2,14 @@ import graphene
 from graphql_jwt.decorators import permission_required
 
 from ..core.fields import PrefetchingConnectionField
-from ..descriptions import DESCRIPTIONS
+# FIXME: Types are imported before mutations on purpose. Otherwise these types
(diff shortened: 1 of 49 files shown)


saleor/graphql/discount/schema.py
```diff
@@ -2,27 +2,31 @@
 from graphql_jwt.decorators import permission_required
 
 from ..core.fields import PrefetchingConnectionField
-from ..descriptions import DESCRIPTIONS
+# FIXME: Types are imported before mutations on purpose. Otherwise these types
```
patrys: What if we imported these at the beginning of `mutations.py`?

maarcingebala: We could do that as well. Actually, I think there is no much difference - either here or there `isort` will complain about the wrong order of imports. I have no idea why this happens, I spent one hour trying to debug it but found nothing.

PR #3532, opened 2019-01-02
Add migrate container to run Saleor's Django migrations.

hairychris:
This makes it easier to get up and running quickly, it does mean that the migrations will be run on startup every time but I don't think this should cause any issues and in some ways is good.

I want to merge this change because running extra commands inside the docker containers at boot (e.g. docker-compose run web python manage.py migrate or whatever) is annoying.

Due to the restart policy of these containers, you also sometimes end up with multiple long running containers and it's a mess. It seems much cleaner to just automatically run the migrations in development by having a migrate container which the web container depends on with no restart set.

### Pull Request Checklist

2. [x] The code is documented (docstrings, project documentation).
3. [x] Changes are mentioned in the changelog.

This is WIP, I need to add changelog and documentation so please don't merge just yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,3 +41,4 @@ All notable, unreleased changes to this project will be documented in this file.
 - Filter attributes by collection in API - #3508 by @maarcingebala
 - Fixed hard-coded site name in order PDFs - #3526 by @NyanKiyoshi
 - Extract enums to separate files - #3523 by @maarcingebala
+- Add `migrate` docker container to automatically run migrations - #3532 by @hairychris
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,23 @@ services:
     depends_on:
       - db
       - redis
+      - migrate
     command: python manage.py runserver 0.0.0.0:8000
+
+  migrate:
+    build:
+      context: .
+      dockerfile: ./Dockerfile
+      args:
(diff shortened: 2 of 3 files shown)


hairychris:
Does anyone see any flaws to this approach?

jxltom:
In heroku's python buildpack, ```migrate``` will run in each deploy automatically. Collectstatic will also run automatically but it can be disabled by ```DISABLE_COLLECTSTATIC```. So I think this approach is good.

michaljelonek:
I don't think we should implement this. The problem is that `depends_on` doesn't guarantee that the database is "ready" and accepting connections. Incidentally docker swarm ignores `depends_on` altogether. I feel this might be more problematic than it is worth, which is to solve a minor inconvenience of typing a command.

Heroku has [release phase](https://devcenter.heroku.com/articles/release-phase), which I feel is a great place to run migrations before actual code deployment.

As for the problem with restarting containers because of the restart policy, the way to solve this is to use `--rm` flag along with the `run` command. That will destroy the container on exit and not keep it in an endless restarting loop. Docs will be updated if they aren't already.

hairychris:
That makes sense, I agree with the documentation fix solution and not merging this. From the perspective of new users running extra documented commands is fine as long as we don't list commands that will get the user into a mess.

hairychris:
Doc fixes https://github.com/mirumee/saleor/pull/3643

PR #3605, opened 2019-01-16
Support get correct payment status for order without any payments

jxltom:
Current storefront 1.0 supports order without payments. But the payment status of order is got from last payment. This means payment status of order without any payments will display as ```None```, this is unfriendly and confusing to users. This PR return default payment status ```NOT_CHARGED``` for order without any payments.

### Screenshots

Before:
screenshot from 2019-01-16 11-53-29

After:
screenshot from 2019-01-16 11-52-59

### Pull Request Checklist

### DIFF
--- a/saleor/order/models.py
+++ b/saleor/order/models.py
@@ -188,6 +188,18 @@ class Order(models.Model):
             return last_payment.get_charge_status_display()
         return None
 
+    def get_payment_status(self):
+        status = self.get_last_payment_status()
+        if status is None:
+            return ChargeStatus.NOT_CHARGED
+        return status
+
+    def get_payment_status_display(self):
+        status = self.get_last_payment_status_display()
+        if status is None:
+            return dict(ChargeStatus.CHOICES).get(ChargeStatus.NOT_CHARGED)
+        return status
+
     def is_pre_authorized(self):
         return self.payments.filter(
             is_active=True,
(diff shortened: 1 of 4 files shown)


saleor/order/models.py
```diff
@@ -188,6 +188,18 @@ def get_last_payment_status_display(self):
             return last_payment.get_charge_status_display()
         return None
 
+    def get_payment_status(self):
+        status = self.get_last_payment_status()
```
maarcingebala: Is `get_last_payment_status` and `get_payment_status_display` needed anymore, after introducing these two functions? If no, we could get rid of them to simplify the codebase. Now we would have 4 similar functions while maybe we could have only these 2.

jxltom: No, they are not used as more. But I'm thinking ```get_last_payment_status``` and ```get_payment_status``` are actually not same. Maybe there is a chance that ```get_last_payment_status``` would be used in further development? Combine them into single function will loose the ability to access last payment status.

How about make them as protected methods like ```_get_last_payment_status``` and ```_get_last_payment_status_display```, or just removing them, what do you think?

maarcingebala: I think that `get_last_payment` could be in fact called `get_payment`. If you look at the logic related to order management, we always take into account the last payment to determine whether the order was fully paid etc. I also think that it's better to remove the function now if they are not used, than to keep them just in case they were needed at some point.

jxltom: Yeah I noticed that. Makes sense. I've updated the PR.

Issue #3086, opened 2018-10-10
Drop M2M relationship between product types and attribtues


maarcingebala:
Currently, the relationship is M2M, but in the new dashboard and API we want to make attributes specific to a product type, instead of sharing them between various types.

- [ ] change relationship from M2M to foreign keys
- [ ] migrate existing (possibly reused withing multiple types) attributes to match the new model
- [ ] adjust logic in backend and API

maarcingebala:
Related: #3088

PR #3628, opened 2019-01-22
Support filter products with same attribute slug and value slug in storefront and GraphQL API

jxltom:
After https://github.com/mirumee/saleor/issues/3086, the attributes can not be shared by different product types. So it is very likely that there are multiple attributes with same slug. But currently Saleor does not support filtering products by same attribute slug and value slug. This is because in storefront the filters with same attribute slug in MultipleChoiceFilter will be overrided in https://github.com/mirumee/saleor/blob/7698e5f31d51d79520b4f05948d86d91d840dc5e/saleor/product/filters.py#L62 and in GraphQL the filters with same attribute slug is also overrided in https://github.com/mirumee/saleor/blob/7698e5f31d51d79520b4f05948d86d91d840dc5e/saleor/graphql/product/filters.py#L13.

This PR makes Saleor able to filter products by same attribute slug and value slug:

- Same logic is applied to GraphQL API and storefront 1.0. The products can be filtered by ```attribute_slug:value_slug```
- Since we filter the products by slugs, the values of attributes with same slug are merged into same attribute and then shown as one group in storefront 1.0.
- This PR merges product attributes and product variant attributes into one only if they have same slug. There is no need to display same attributes multiple times in storefront 1.0 if they have same slug, since the products are finally filtered by slug instead of specific pk.
- Fix a bug where attribute with emtpy values is unnecessary to be rendered in storefront 1.0

### Screenshots

For example, I have a product with ```red``` attribute value from product attribute (it has ```red```, ```blue``` values) and I have another product with ```red``` attribute value from another product variant attribute  (it has ```red```, ```purple```, ```black``` values) , the first image shows that storefront doesn't show all the attribute values correctly and it doesn't successfully fitler all products with ```color``` attribute slug and ```red``` value slug. After this PR, the second image shows Saleor did the work correctly.

Before:
screenshot from 2019-01-22 15-01-39

After:
screenshot from 2019-01-22 15-02-19

### Pull Request Checklist


maarcingebala:
Thanks for opening this @jxltom. It's pretty large PR and I still had no time to look into it more carefully, but I promise I will 🙂. For now, don't bother with resolving conflicts in changelog if there are any, I can always do that before I merge the PR.

jxltom:
@maarcingebala Sure please take your time 😃 The logic with attributes is a little bit tricky but I tried my best to make the logic in this PR clear, such as merging logic of storefront and API into a single one. The only flaw with this PR is merging attribute values may loosing some translatable names. But it is the issue with the design, and we can solve this after design of product attributes is refactored such as suggested in https://github.com/mirumee/saleor/issues/3088.

wulaaf:
So:

attribute 1: color, values: red, green, black, white
attribute 2: color, values: black, white

Expected result: red, green, black, white
Actual result: black, white

Any idea as to why this might be happening? I've set both attributes name and slug to the same thing which is "color".

wulaaf:
> So:
> 
> attribute 1: color, values: red, green, black, white
> attribute 2: color, values: black, white
> 
> Expected result: red, green, black, white
> Actual result: black, white
> 
> Any idea as to why this might be happening? I've set both attributes name and slug to the same thing which is "color".

Like, the filters on the left do not show all the possible values of the attribute, which is weird to say the least. I haven't changed anything.

wulaaf:
> > So:
> > attribute 1: color, values: red, green, black, white
> > attribute 2: color, values: black, white
> > Expected result: red, green, black, white
> > Actual result: black, white
> > Any idea as to why this might be happening? I've set both attributes name and slug to the same thing which is "color".
> 
> Like, the filters on the left do not show all the possible values of the attribute, which is weird to say the least. I haven't changed anything.

I did some testing and figured out that a newly created attribute takes precedence over an older attribute. The values of the newer attribute are rendered whilst the older values are not rendered even if they are present in the newer attribute.

I would like to know if this is only specific to my use case or is anyone else experiencing this.

jxltom:
@wulaaf All the cases reported above are tested with this PR, right? If so, there may be some bugs and I will do some investigations.

wulaaf:
After some more investigation, I found another thing that might be crucial:

Product Type = Case
Attribute 1 = Brand (pk=1)

Product Type = Screen Protector
Attribute 2 = Brand (pk=2)

When I filter brand names together, the URL should show something like this:
?brand=1&brand=2

However, it only shows
?brand=1

Which means that when the filter is applied, the 2nd product type is not shown.

I hope that helps.

wulaaf:
Guys, any updates on this?

jxltom:
I will have a look for the issue you mentioned ASAP.

jxltom:
> > > So:
> > > attribute 1: color, values: red, green, black, white
> > > attribute 2: color, values: black, white
> > > Expected result: red, green, black, white
> > > Actual result: black, white
> > > Any idea as to why this might be happening? I've set both attributes name and slug to the same thing which is "color".
> > 
> > 
> > Like, the filters on the left do not show all the possible values of the attribute, which is weird to say the least. I haven't changed anything.
> 
> I did some testing and figured out that a newly created attribute takes precedence over an older attribute. The values of the newer attribute are rendered whilst the older values are not rendered even if they are present in the newer attribute.
> 
> I would like to know if this is only specific to my use case or is anyone else experiencing this.

@wulaaf This issue has already been solved by this PR, you can merge this PR into your own fork. I have set same attributes as your suggested, and now the attributes are rendered correctly as below image.
screenshot from 2019-03-08 14-15-17

jxltom:
> Screen Protector

The second issue is also solved this issue, I have set case and screen protector which all have brand attribute with brand1 value as you suggested, now if you filter by ```brand1``` attribute value, all products (case and screen protector) can be filtered out as below image.

screenshot from 2019-03-08 14-22-54

jxltom:
@maarcingebala This PR is updated with upstream, any chance that this can be merged?

maarcingebala:
@jxltom We're currently re-evaluating our attributes structure in Saleor and it is very likely that we'll change it to allow standalone attributes that you'd be able to create independently from product types and reuse in different product types. We're also going to add some more fields to the Attribute model that will allow you to control whether an attribute is private/public, whether it should be included in filtering etc. I think these features would resolve all the problems that you've mentioned here.

maarcingebala:
That's because we want to be closer to the Magento's structure of attributes to enable easier migration to Saleor 😎

jxltom:
Nice, look forward to that big refactor!

petedermott:
Can I ask what the reasoning behind [#3086](https://github.com/mirumee/saleor/issues/3086) was? (or could someone point me towards the discussion behind it?)

jxltom:
@petedermott There is a open issue https://github.com/mirumee/saleor/issues/3554 for that.

maarcingebala:
@petedermott In the old dashboard, we had three separate sections to manage products, product types and attributes. We found it inconvenient to use for product management as it required jumping between these sections. We decided to move attributes closer to product types to make the UX better and that change in #3086 was implemented as a result. 

While this might work for a small shop, we found it problematic in larger ones when there is a lot of arguments, and it is a valid case to share them between types. That's why we're going to revert that behavior but still try to come up with efficient UX for management.

maarcingebala:
I'm gonna close this PR and within next few weeks, we'll start working on the new structure.

PR #3632, opened 2019-01-22
Fix bug where node order is not preserved in GraphQL API

jxltom:
For example in ```CheckoutCreate``` mutation, quantity is got from lines as list in https://github.com/mirumee/saleor/blob/3eb828dcd75e9ab7cba77d0d324253f7748f6000/saleor/graphql/checkout/mutations.py#L121, variants is obtained by ids from lines in https://github.com/mirumee/saleor/blob/3eb828dcd75e9ab7cba77d0d324253f7748f6000/saleor/graphql/checkout/mutations.py#L118-L120. But the variants order in ```get_nodes_or_error->get_nodes``` is not preserved as the input ```ids```. This causes the created checkout lines in https://github.com/mirumee/saleor/blob/3eb828dcd75e9ab7cba77d0d324253f7748f6000/saleor/graphql/checkout/mutations.py#L178-L179 may have swapped quantities.

### Pull Request Checklist


saleor/graphql/utils.py
```diff
@@ -62,6 +62,7 @@ def get_nodes(ids, graphene_type=None):
                 break
 
     nodes = list(graphene_type._meta.model.objects.filter(pk__in=pks))
+    nodes.sort(key=lambda e: pks.index(str(e.pk)))  # preserve order in pks
```
michaljelonek: Wouldn't using `order_by` on the query set achieve the same? I also think that this evaluation to list could be removed (it's iterated over under this if) and `nodes.exists()` should be much quicker than `bool(list(nodes))`? Unless I'm missing something here.

maarcingebala: Using `order_by` would just sort the instances by database ID, but I think the idea here was to reflect the order of keys that were passed as arguments to this function. 

I agree that we might not need the `list` evaluation.

michaljelonek: Technically you can use `Case` and `When` to achieve that as well in `order_by`, but how fast (or slow) that is, I have no idea, I didn't do any benchmarking.

jxltom: Sorting by the order of pks (ids) can be achieved by db queries actually, but that requires different implementation based on the database you are using, such as postgresql or mysql. I know saleor is using postgresql, but I prefer solving this issue in a generic way as nodes is converted to list anyway.

jxltom: After some digging, I agree that it can be acheive by ```Case``` and ```When``` in db side. If we can make sure all method calling ```get_nodes``` expects queryset instead of list, maybe we can do more optimization based on that.

PR #3639, opened 2019-01-23
Expose in API list of supported payment gateways

### LINKED ISSUE #3617: Expose in API list of supported payment gateways
In API we expose the `GatewaysEnum` enum which contains all the available payment gateways, but there is no way for the Storefront 2.0 to determine which of them are actually enabled (currently Braintree is hardcoded over there). We should expose a field that would provide this information, e.g:
```
type Shop {
  enabledGateways: [GatewaysEnum]
}
```

A resolver function should take into account gateways enabled in `CHECKOUT_PAYMENT_GATEWAYS` setting.

fowczarek:
I want to merge this change because resolve #3617 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/shop/types.py
+++ b/saleor/graphql/shop/types.py
@@ -78,7 +78,8 @@ class Shop(graphene.ObjectType):
     default_currency = graphene.String(
         description='Default shop\'s currency.', required=True)
     available_payment_gateways = graphene.List(
-        PaymentGatewayEnum, description='List of available payment gateways.')
+        PaymentGatewayEnum, description='List of available payment gateways.',
+        required=True)
     default_country = graphene.Field(
         CountryDisplay, description='Default shop\'s country')
     description = graphene.String(description='Shop\'s description.')
--- a/saleor/graphql/shop/types.py
+++ b/saleor/graphql/shop/types.py
@@ -77,7 +77,7 @@ class Shop(graphene.ObjectType):
         required=True)
     default_currency = graphene.String(
         description='Default shop\'s currency.', required=True)
-    enabled_gateways = graphene.List(
+    available_payment_gateways = graphene.List(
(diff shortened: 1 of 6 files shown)


saleor/graphql/shop/types.py
```diff
@@ -76,6 +77,8 @@ class Shop(graphene.ObjectType):
         required=True)
     default_currency = graphene.String(
         description='Default shop\'s currency.', required=True)
+    enabled_gateways = graphene.List(
```
Pacu2: It might not be obvious that we are talking about the payments gateways. I’d opt for payment_gateways or available_payment_gateways 

(Don’t feel like “enabled” fits well in this context, as the API is user facing we should talk about the gateways that are available to use, not enabled by the store’s admin)

saleor/graphql/shop/types.py
```diff
@@ -76,6 +77,8 @@ class Shop(graphene.ObjectType):
         required=True)
     default_currency = graphene.String(
         description='Default shop\'s currency.', required=True)
+    available_payment_gateways = graphene.List(
```
maarcingebala: This field will never be `null` so we should add `required=True`.

maarcingebala: Well, theoretically it would make sense, but did we really had any use case in any of our storefronts for that?

patrys: Each gateway has its limitations: some only accept cards issued in certain countries, most of them are not available in China, etc.

maarcingebala: OK, we could move this field to the `Checkout` type. We could keep the resolver to return gateways that have been configured in settings, but later this will be a good place to add some logic.

fowczarek: Done

PR #3643, opened 2019-01-23
Fix docs so docker manage.py containers don't restart infinitely

hairychris:
I want to merge this change because containers shouldn't restart indefinitely.

### Pull Request Checklist

1. [x ] The changes are tested.

### DIFF
--- a/docs/customization/docker.rst
+++ b/docs/customization/docker.rst
@@ -35,9 +35,9 @@ Usage
 
    .. code-block:: bash
 
-    $ docker-compose run web python3 manage.py migrate
-    $ docker-compose run web python3 manage.py collectstatic
-    $ docker-compose run web python3 manage.py populatedb --createsuperuser
+    $ docker-compose run --rm web python3 manage.py migrate
+    $ docker-compose run --rm web python3 manage.py collectstatic
+    $ docker-compose run --rm web python3 manage.py populatedb --createsuperuser
 
    The ``--createsuperuser`` argument creates an admin account for
    ``admin@example.com`` with the password set to ``admin``.


docs/customization/docker.rst
```diff
@@ -35,9 +35,9 @@ Usage
 
    .. code-block:: bash
 
-    $ docker-compose run web python3 manage.py migrate
-    $ docker-compose run web python3 manage.py collectstatic
-    $ docker-compose run web python3 manage.py populatedb --createsuperuser
+    $ docker-compose run --rm web python3 manage.py migrate
```
jxltom: Shouldn't we use https://docs.docker.com/engine/reference/commandline/run/#restart-policies---restart instead?

michaljelonek: We're using `docker-compose run` here though, we'd have to change the policy in `docker-compose.yml` and I personally don't have an argument whether we prefer `unless-stopped` or `on-failure`.

jxltom: One reason is that keeping these container makes the logs of these commands could be accessible after finishing running. The logs are kept as the stdout and stderr of these containers. Deleting these containers will make the logs unaccessible.

hairychris: Not sure that's a huge issue as the log would be spat out to the users console, it would be up to them to cat that to a file if they need to keep it around. What are you suggesting? Not sure that restart policies helps here.

jxltom: I don't have a strong preference actually, just sharing some potential concerns. :P

hairychris: Thanks I appreciate it, my only concern is a new developer cloning the project, running docker as per the docs and having containers wasting resources boot looping. Any way we can fix that is good by me. Currently I'm not seeing a viable second solution.

PR #3645, opened 2019-01-24
Remove unused imports in saleor and tests

jxltom:
This PR removes the unused imports in saleor and tests directory

### Pull Request Checklist

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -31,6 +31,7 @@ exclude_patterns:
 - "**/__pycache__/"
 - "**/migrations/"
 - "node_modules/"
+- "tests/"
 ratings:
   paths:
   - saleor/**
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -31,8 +31,6 @@ exclude_patterns:
 - "**/__pycache__/"
 - "**/migrations/"
 - "node_modules/"
-- "tests/"
-- "**/test_*"
 ratings:
   paths:
   - saleor/**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -247,3 +247,14 @@ All notable, unreleased changes to this project will be documented in this file.
 ### Docs
 - Add createdb note to getting started for Windows - #3106 by @ajostergaard
 - Update docs on pipenv - #3045 by @jxltom
+- Validate files uploaded in API in a unified way - #3633 by @fowczarek
+- Add ShopFetchTaxRates mutation - #3622 by @fowczarek
+- Add taxes section - #3622 by @dominik-zeglen
+- Expose in API list of supported payment gateways - #3639 by @fowczarek
+- Display payment status in account order list page and account order detail page - #3637 by @jxltom
+- Fix bug where node order is not preserved in GraphQL API - #3632 by @jxltom
+- Support set arbitary charge status for dummy gateway in storefront 1.0 - #3648 by @jxltom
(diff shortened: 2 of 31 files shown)


michaljelonek:
Looks good to me 👍 

I wonder though whether we want to include tests back in codeclimate. [This PR](https://github.com/mirumee/saleor/pull/1384) excluded them, so we're never assuring quality of code in our tests, but the question is.. should we? 

Also I just realised we never run isort on `__init__.py` files, as isort by default skips them. I think we might want to include those as well? 

cc: @maarcingebala

jxltom:
It is weird that the tests failed...

maarcingebala:
Tests are an integral part of the project so I think we could actually re-enable checking code style issues for them. I also agree about enabling `isort` for `__init__.py` files. [This comment](https://github.com/timothycrosley/isort/issues/139#issuecomment-451616157) explains how to do that. @jxltom Do you think you could do that in this PR so we'll have it all sorted out in one go?

maarcingebala:
Tests that failed were using dates, maybe we have indeterministic tests? But I don't recall to have seen them failing before.

jxltom:
@maarcingebala No problem, I will handle it. :D

jxltom:
@maarcingebala @maciekjablonski This PR is ready to review. Besides enabling tests in codeclimate and ```__init__``` for isorts, I also update threhold of codeclimate: updating max argument to 5 which is same as pylint's default value and max lines to 1000 since it is quite common to have a lot of lines in single file in saleor.

maarcingebala:
But only now I saw `codeclimate - 606 issues to fix` 🙂There are some issues about arguments and lines count, are you sure the config is good? What were the limits before this change?

jxltom:
@maarcingebala The original config is codeclimate's defaults: 4 maximum arguments and 250 max lines. The new settings deduces the warnings but not gets rid of all of them. I can restore to original settings or make the threshold even larger if you prefer.

michaljelonek:
@jxltom @maarcingebala 

A lot of those issues, while valid, make absolutely no sense to fix in tests. Those would explain why the tests were excluded in the first place. Looking at the code climate documentation it looks like there's no way to disable certain checks using exclude patters -> https://docs.codeclimate.com/docs/advanced-configuration#section-exclude-patterns. Only global or plugin (or engines in our case) exclusions are available, so we could disable for example pylint and duplication plugins (engines) and only have pep8 and isort run there. Additionally code climate by default excludes tests -> https://docs.codeclimate.com/docs/default-analysis-configuration, perhaps it's not the best tool to assure quality in tests.

But I don't want more time and this PR in its original form wasted, so what I propose is to exclude tests how they were, just merge what we have here without the `codeclimate.yml` change. We would ignore most of those issues anyway, so kinda beats the point of checking for them. We could definitely come back to this topic in a separate issue (I did not expect it would bring that many issues to be fair).

Thoughts?

maarcingebala:
I fully agree with @michaljelonek on this, let's revert the codeclimate config changes.

jxltom:
@maarcingebala @michaljelonek I've reverted the changes in codeclimate configurations.

maarcingebala:
@jxltom Thanks! We also agreed that we want to keep the tests excluded as it was before. Could you revert the `.codeclimate.yml` file?

jxltom:
@maarcingebala No problem, done.

maarcingebala:
Thanks!

PR #3670, opened 2019-01-30
Add component generator

dominik-zeglen:
I want to merge this change because it adds a simple component generator to stop copying boilerplate.

Usage: run `npm run generate`.

### Pull Request Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,6 +129,7 @@
     "mini-css-extract-plugin": "^0.4.0",
     "mjml": "^4.0.5",
     "node-sass": "^4.10.0",
+    "plop": "^2.2.0",
     "postcss": "^6.0.21",
     "postcss-loader": "^2.1.3",
     "react-test-renderer": "^16.6.3",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -187,7 +188,8 @@
   "scripts": {
     "build-assets": "webpack -p",
     "build-schema": "python manage.py get_graphql_schema > saleor/graphql/schema.graphql",
-    "generate-types": "apollo client:codegen --target=typescript types --globalTypesFile=saleor/static/dashboard-next/types/globalTypes.ts",
+    "build-types": "apollo client:codegen --target=typescript types --globalTypesFile=saleor/static/dashboard-next/types/globalTypes.ts",
+    "generate": "plop --plopfile .plop/plopfile.js",
```
maarcingebala: I think `npm run generate` sounds too enigmatic. I'd propose using a more explicit command like `generate-component` for that.

dominik-zeglen: At this moment it generates only components, but in the future, it can be used to generate other things.

maarcingebala: So how will we tell the generator what files to generate then? Still, I think `generate` doesn't give any guesses about what it actually does.

maarcingebala: Another idea: `npm run new-component`

piotrgrundas: I agree with @maarcingebala with renaming it sth like `generate-component`.
OR u could leave generate, but change the prompt in such way, that ATM u could only choose creation of a new component.

PR #3696, opened 2019-02-08
Checkout shipping address form and logo fix

### LINKED ISSUE #3673: Storefront 1.0 visual issues
After recent changes to Storefront 1.0 templates I've found two issues:
- [ ] Checkout layout is broken on large screens. This happens only for guest customers for whom we render the "email" input which I suppose breaks the layout. 

image

- [ ] Logo displayed at the top of the storefront seems too small. In checkout it renders properly.

Storefront:
image

Checkout:
image

piotrgrundas:
Resolves #3673 

image
image
image

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -78,7 +78,7 @@ module.exports = (env, argv) => {
         {
           test: /\.scss$/,
           use: [
-            MiniCssExtractPlugin.loader,
+            devMode ? "style-loader" : MiniCssExtractPlugin.loader,
             {
               loader: 'css-loader',
               options: {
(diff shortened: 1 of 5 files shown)


webpack.config.js
```diff
@@ -78,7 +78,7 @@ module.exports = (env, argv) => {
         {
           test: /\.scss$/,
           use: [
-            MiniCssExtractPlugin.loader,
+            devMode ? "style-loader" : MiniCssExtractPlugin.loader,
```
piotrgrundas: For source maps to work properly.

PR #3706, opened 2019-02-11
Unify fields on product, collection and page models

### LINKED ISSUE #3479: Unify field naming related to publication date
There is naming inconsistency in the fields related to publication date and availability:
    - `Product.available_on` + `Product.is_published`
    - `Page.available_on` + `Page.is_visible`
    - `Collection.published_date` + `Collection.is_published` (will be merged soon, #3369)

Also fields `is_visible`/`is_published` are problematic since they can be set to `True` but if the future publication date is set, an item won't be available publicly.

What I would propose to do about it:
- introduce `publication_date = models.DateTimeField()` field in all of the above-mentioned models (currently we have `DateField` but I think some may want to specify exact time as well)
- ~rename `is_published`/`is_visible` fields to `is_draft` - I think that what those fields do is telling if as an admin I've already finished editing an object or I still have something to add, but I'd like to hear some feedback about this one.~ Let's consistently use `is_published` for this kind of fields.

michaljelonek:
Resolves #3479.

* changes `Product.available_on` to `Product.publication_date`
* changes `Page.available_on` to `Page.publication_date`
* changes `Page.is_visible` to `Page.is_published`
* changes `Collection.published_date` to `Collection.publication_date`
* changes `Product.is_available` to property
* use `is_available` consistently on all models
* use a common queryset that exposes `available` (all published objects) and `visible_to_user` (all if uses has access otherwise published objects)

This breaks mutations as there's no easy way that I can see to deprecate the fields. Queries have old fields added and resolved properly with a deprecation flag.

### Pull Request Checklist


saleor/page/models.py
```diff
@@ -1,40 +1,31 @@
...
 from django.utils.translation import pgettext_lazy
 
 from ..core.utils import build_absolute_uri
 from ..core.utils.translations import TranslationProxy
+from ..product.models import PublishedQuerySet
 from ..seo.models import SeoModel, SeoModelTranslation
 
 
-class PageQuerySet(models.QuerySet):
+class PagePublishedQuerySet(PublishedQuerySet):
```
maarcingebala: I'm not a big fan of inheritance but I feel tempted to move `PublishedQuerySet` to `saleor.core.models` an abstracting `is_published` and `publication_date` fields to some kind of `PubilshableModel` abstract model. What do you think?

michaljelonek: That sounds good, I didn't do it to keep it simple, but that works as well 💃

PR #3707, opened 2019-02-11
Improve discounts api, fix decimal empty string

michaljelonek:
Improvements to discounts api.

Adds `Voucher.all_countries` that indicates whether it applies to all countries (`Voucher.countries` is then ignored).

### Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable, unreleased changes to this project will be documented in this file.
 
 ## [Unreleased]
 - Fix problem with l10n in Braintree payment gateway template - #3691 by @Kwaidan00
+- Improve discounts api, fix decimal empty string - #3707 by @michaljelonek
 
 
 ## 2.3.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable, unreleased changes to this project will be documented in this file.
 
 ## [Unreleased]
 - Fix problem with l10n in Braintree payment gateway template - #3691 by @Kwaidan00
-- Improve discounts api, fix decimal empty string - #3707 by @michaljelonek
+- Improve vouchers country limiting  - #3707 by @michaljelonek
 
 
 ## 2.3.0
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -16,7 +16,6 @@ from ..account.forms import get_address_form
 from ..account.models import Address
 from ..account.utils import store_user_address
 from ..core.exceptions import InsufficientStock
-from ..core.i18n import ANY_COUNTRY
 from ..core.utils import to_local_currency
 from ..core.utils.taxes import ZERO_MONEY, get_taxes_for_country
 from ..discount import VoucherType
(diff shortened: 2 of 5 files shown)


saleor/discount/models.py
```diff
@@ -56,6 +56,7 @@ class Voucher(models.Model):
         decimal_places=settings.DEFAULT_DECIMAL_PLACES)
     # not mandatory fields, usage depends on type
     countries = CountryField(multiple=True, blank=True)
+    all_countries = models.BooleanField(default=False)
```
maarcingebala: Do we really need a new database field for that? Looking at the current backend implementation I see that we would only have to allow setting `""` value as a valid choice to `countries` input field in the API. This is how it works for the old dashboard.

saleor/graphql/core/scalars.py
```diff
@@ -28,7 +28,7 @@ def parse_value(value):
             value = str(value)
             return decimal.Decimal(value)
         except decimal.DecimalException:
-            return None
+            return 0
```
maarcingebala: I'm not 100% about that. E.g. `ProductVariantInput` has `priceOverride` Decimal field and I think that we can pass `null` in API in order to "reset" the field. Setting `0` could override the price to zero. But that would have to be checked.

saleor/discount/models.py
maarcingebala: In other words, there should be one way to handle that logic. By adding a new database we're introducing a new way to solve a problem which was already somehow implemented in the existing code.

michaljelonek: I based this off `shippingZone` that has similar solution, but introduces an additional property. Also setting `""` sounds counter-intuitive that it enables all countries to be fair.

saleor/graphql/core/scalars.py
michaljelonek: That's how it has to be though. Decimal does not accept `None` or `""`. It will throw an exception.

saleor/discount/models.py
maarcingebala: So we have three ways - field `default` for shipping zones, `all_countries` for vouchers and `ANY_COUNTRY = ''` as a valid country choice (not available in API though). We want to have one common way to handle that behavior.

saleor/graphql/core/scalars.py
maarcingebala: We need to make sure that this change won't introduce bugs in other parts of our codebase. I would double check how it works with `ProductVariantInput` and other occurrences of `Decimal` in API.

saleor/discount/models.py
michaljelonek: I modified the `shop` query to additionally return `('', 'Any Country')`, but that probably means some bits need to be updated on PWA, as I believe the same query is used for the checkout addresses (probably an easy filter to remove countries without code).

maarcingebala: 👍 Makes sense.

saleor/graphql/core/scalars.py
michaljelonek: I tested it with `ProductVariantInput`, if `null` is sent then the field is correctly nulled. If `""` then it sets it to 0 (before this change it would explode with `""`). 

In my opinion using `""` and expecting that the field is going to be nulled/zeroed is kinda too optimistic, but this change will at least ensure the API won't break. We should make sure that if we want to zero a field, then we send either `0` or `"0"`, not `""`. Ditto with `null`.

maarcingebala: So maybe we should raise an error if `""` is passed as Decimal? I think we could do that at the scalar implementation level.

michaljelonek: How is that different than not fixing this case? That's how it worked. If it was `""` it said it expected a Decimal and got an empty string.

```
    parse_result = type.parse_literal(value_ast)
    if parse_result is None:
        return [
            u'Expected type "{}", found {}.'.format(type.name, print_ast(value_ast))
        ]
```
from `graphql.utils` and then it reported an error:
`"message": "Variable \"$value\" got invalid value \"\".\nExpected type \"Decimal\", found \"\".",`

From the discussion we had I thought we want to support `""`, which resulted in this fix.

Sending `null` or `"0"` or `0` would alleviate this issue and assign `0` to the value (or `null` in case of a nullable field).

michaljelonek: I feel this is the same kind of a problem if one tried to assign a totally random number to a date field and expecting it to work.

maarcingebala: Right. I'm switching between different issues and I lost the context.

Issue #3679, opened 2019-02-01
Merge "charge" and "auth" payment operations


maarcingebala:
Saleor used to have two operations related to charging payments: authorization and capture. In this context, "charge" is simply an operation which immediately performs "capture" after "authorization". If that's possible, we should merge those two operations to simplify the payments architecture.

Related discussion can be found here: https://github.com/mirumee/saleor/pull/3519#discussion_r247864471. 

As far as I remember, "charging" as a separate operation was introduced with Stripe, but this gateway supports charing in two steps: https://stripe.com/docs/charges#auth-capture

NyanKiyoshi:
We should add a flag to gateways to say whether the gateway supports authorization or not. And then check for the authorization's presence or not depending on that during `gateway_capture(...)`.

NyanKiyoshi:
Some work have been done at https://github.com/mirumee/saleor/commit/c7f93b3054fdf231b7576385f705466b9758d9c2 to introduce a boolean constant `SUPPORTS_AUTHORIZATION` in gateways that is used to check or not for the authorization payment when capturing.

maarcingebala:
Resolved in #4098

PR #3715, opened 2019-02-13
Checkout api fixes

### LINKED ISSUE #3652: Reuse Storefront 1.0 payment logic in API
We need to unify the way of creating payments between Storefront 1.0 and API (and Storefront 2.0 consequently). Storefront 1.0 uses `start_payment` view which creates a new `Payment` object and calls the `process_payment` function. This function knows which gateway operations to perform for each gateway.

Similar flow should happen in API. Currently, in the API we have `checkoutPaymentCreate` which is used by Storefront 2.0. The mutation underneath only implements creating a new `Payment` instance but it seems to be missing the logic that is implemented in `start_payment`.

From that perspective this mutation is no different from the storefront's view - they both should trigger the payment processing flow which should be identical in both cases. 

@michaljelonek / @jxltom - You both had a significant impact on the shape of payments in Saleor, so I'd like to have your feedback on that.

maarcingebala:
Fixes #3652 

We should keep the business logic away from views, forms, mutations or resolvers and keep it in smaller utility functions, that can be composed in different parts of our codebase. We used to do that in Saleor, but during the rapid development of checkout API and payments refactoring we bent those rules a bit. This PR is the first attempt to clean up the implementation of checkout/payments logic in our API.

Functions that I refactored/extracted + other significant changes:
- `create_payment` - a common function to create a new payment instance, now used both in Storefront 1.0 and in `checkoutPaymentCreate ` mutation.
- `mark_order_as_paid` - extracted function to create an "empty" payment
- `clean_mark_order_as_paid` - common function to check for errors before marking order as paid
- `is_full_paid` - I found quite serious bug in this function, as it wasn't taking discounts and taxes into account to calculate the prices - fixed.
- Added `checkout_token` field in the `Order` model - in order to prevent creating multiple orders from a single checkout instance in `checkoutComplete` mutation.
- Changed signature of `checkoutPaymentCreate` mutation to follow the style of other mutations.

What needs to be done yet:
- I'd like to remove `paymentCharge` mutation - I'm don't see a use case for that. Storefront shouldn't call it manually - charge would happen during checkout completion. Also, the dashboard doesn't need that operation. I could do that in this PR as the tests are failing currently for this mutation. Also, we're going to drop `charge` operation in the future (#3679).
- There is a lot more logic to unify between API and current storefront but I'd like to investigate it in a different PR as this is already big.

### Pull Request Checklist


saleor/checkout/utils.py
```diff
@@ -879,20 +883,22 @@ def create_order(cart, tracking_code, discounts, taxes):
...
-def is_fully_paid(cart: Cart):
-    payments = cart.payments.filter(is_active=True)
-    total_paid = sum(
-        [p.total for p in payments])
-    return total_paid >= cart.get_total().gross.amount
+def is_fully_paid(cart: Cart, taxes, discounts):
+    """Check if checkout is fully paid."""
+    payments = [
+        payment for payment in cart.payments.all() if payment.is_active]
+    total_paid = sum([p.total for p in payments])
```
michaljelonek: Same as the previous one, also could use `aggregate` here, since payments aren't needed to be loaded in memory. Something along the lines of: `filter(is_active=True).aggregate(Sum('total'))['total__sum']`

saleor/graphql/checkout/mutations.py
```diff
@@ -448,57 +448,45 @@ class CheckoutComplete(BaseMutation):
...
     class Arguments:
-        checkout_id = graphene.ID(description='Checkout ID')
+        checkout_id = graphene.ID(description='Checkout ID', required=True)
 
     class Meta:
         description = (
-            'Completes the checkout, creates an order from it and '
-            'charges the customer\'s funding source.')
+            'Completes the checkout. As a result a new order is created and '
+            'a payment charge is made. This action requires a succesfull '
```
michaljelonek: Typo, should be `successful`.

tests/api/test_payment.py
```diff
@@ -265,6 +264,7 @@ def test_payment_capture_gateway_error(
 """
 
 
+@pytest.mark.xfail
```
michaljelonek: A reason would be nice :P

tests/test_payment.py
```diff
@@ -153,15 +133,38 @@ def test_function(payment, *args, **kwargs):
...
     assert payment == same_payment
 
 
+def test_mark_as_paid(admin_user, draft_order):
+    mark_order_as_paid(draft_order, admin_user)
+    payment = draft_order.payments.last()
+    assert payment.charge_status == ChargeStatus.CHARGED
+    assert payment.captured_amount == draft_order.total.gross.amount
+    assert (
+        draft_order.events.last().type == OrderEvents.ORDER_MARKED_AS_PAID.
```
michaljelonek: Breaking on a dot looks sooo ugly :(

Maybe something more like:
```
assert draft_order.events.last().type == (
    OrderEvents.ORDER_MARKED_AS_PAID.value)
```

saleor/checkout/models.py
```diff
@@ -111,6 +113,11 @@ def get_line(self, variant):
         matching_lines = (line for line in self if line.variant == variant)
         return next(matching_lines, None)
 
+    def get_last_active_payment(self):
+        payments = [
+            payment for payment in self.payments.all() if payment.is_active]
```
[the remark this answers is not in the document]
maarcingebala: I did this on purpose to save some database queries (assuming that `payments` are prefetched). We do the same thing in Order model.

tests/api/test_payment.py
maarcingebala: Right, but I wanted to remove this test at all once I get rid of the "charge" mutation :)

saleor/checkout/utils.py
maarcingebala: Same as above, is someone e.g. lists orders in API which might call this method we will have database queries duplication.

PR #3717, opened 2019-02-13
Generate voucher code if it wasn't provided in mutation

### LINKED ISSUE #3704: Generate voucher code if it wasn't provided
When creating a voucher through API, a user has to provide the voucher code (it is required by the database model). In Dashboard 1.0 we used to generate the code for users and here we should bring back the same behavior.
Once this feature is implemented we'll add a help text in Dashboard 2.0 saying that this field would be generated is users leave it empty (that's a separate issue that we'll open later).

**Required changes**
Within the `voucherCreate` mutation generate a new code if an empty string was passed in the `VoucherInput.code` field. We should use the same function for generating codes as in Dashboard 1.0 (if this logic sits within a Django form, it should be abstracted to a utility function in the `discounts` module).

Kwaidan00:
I want to merge this change because it implements generating voucher code if it wasn't provided in `voucherCreate` mutation. `code` argument is still required in the mutation; voucher code is generated only if empty string was passed as `code` value.

It closes #3704. 

### Pull Request Checklist


saleor/discount/utils.py
```diff
@@ -86,3 +88,11 @@ def get_products_voucher_discount(voucher, prices):
         voucher.get_discount_amount_for(price) for price in prices)
     total_amount = sum(discounts, ZERO_MONEY)
     return total_amount
+
+
+def generate_voucher_code():
+    """Generate new unique voucher code."""
+    while True:
```
maarcingebala: 😱 I just realized that we've had this ugly `while True` for so long. Can we get rid of this line? Chances of collision in UUID are so low that we could ignore it (we do for generating order tokens for example).

Kwaidan00: In addition, if an impossible was happen, we wouldn't store the duplicate in database (due to `unique=True`). 
I've removed the loop.

PR #3724, opened 2019-02-14
Fix access to unpublished objects via API

### LINKED ISSUE #3522: Access to unpublished objects in API
Some of the API fields leak access to unpublished objects. The issue affects the following fields:
- [x] `collection(id: ID!)`
- [x] `product(id: ID!)`
- [x] `productVariant(id: ID!)`

All above fields use `graphene.Node.get_node_from_global_id` in their resolvers which operates on queryset of all objects. We should add some logic that checks if the `request.user` has access to the given object. This is properly implemented for example in the [page resolver](https://github.com/mirumee/saleor/blob/7109b84e50eb46113d141dc07ab99e9f99da0e05/saleor/graphql/page/resolvers.py#L10). Access rules in summary:
- anonymous users and customers should have only access to objects that are published
- unpublished objects can be viewed by staff users with proper permissions

We can fix that by implementing simlar logic as in the page resolver (link above). This should do for now, but in the long run I think it would be nice to have the ability to determine the queryset that `get_node_from_global_id` uses underneath. There is [an open PR in graphene-django](https://github.com/graphql-python/graphene-django/pull/528) for that.

Kwaidan00:
I want to merge this change because it fixes access to unpublished objects via API. When user send query to `Product` or `Collection` node, server checks user's permissions with `PublishedQuerySet.visible_to_user` method. When user send query to `ProductVariant` node, server checks visibility of variant's related product.

In addition, that PR fixes similar access leak in `graphql.product.resolvers.resolve_product_variants` method.

Please note that permission mechanizm requires `user` object from `info.context`. It is always provided by GraphQL engine when you executing query from web. However, when you use `schema.execute`, please provide `context_value` to ensure proper working of permission mechanizm.

Fixes #3522 .

### Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add instructions for using local assets in Docker - #3723 by @michaljelonek
 - Remove unused imports - #3645 by @jxltom
 - Add discount section - #3654 by @dominik-zeglen
+- Fix access to unpublished objects via API - #3724 by @Kwaidan00
 
 
 ## 2.3.0
--- a/saleor/graphql/product/resolvers.py
+++ b/saleor/graphql/product/resolvers.py
@@ -113,7 +113,11 @@ def resolve_product_types(info):
 
 
 def resolve_product_variants(info, ids=None):
-    qs = models.ProductVariant.objects.all()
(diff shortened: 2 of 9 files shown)


maarcingebala:
@Kwaidan00 Please resolve conflicting files.

PR #3725, opened 2019-02-15
Get or create emtpy checkout when accessing checkout of a user in GraphQL API

jxltom:
This PR is opinionated but I still would like to propose it since it will simplify the frontend logic when creating/updating checkouts for authenticated users.

If an anonymous user who already has a cart (stored in local storage) tries to login, the frontend needs to request ```me->checkout``` to see if a user has already a checkout. If the user doesn't have a checkout, the frontend needs to call ```checkoutCreate``` instead of ```checkoutLinesUpdate``` or ```checkoutLinesAdd``` to create a checkout first. If any user have an empty checkout by default, the frontend doesn't need to call ```checkoutCreate``` if checkout doesn't exist. This PR simplifies the frontend logic.

Thoughts?

### Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add instructions for using local assets in Docker - #3723 by @michaljelonek
 - Remove unused imports - #3645 by @jxltom
 - Add discount section - #3654 by @dominik-zeglen
+- Get or create checkout when accesing checkout of a user in GraphQL API - #3725 by @jxltom
 
 
 ## 2.3.0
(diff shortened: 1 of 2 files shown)


maarcingebala:
Our intention was to avoid creating empty checkout instances for every user. Also, a potential flaw is when fetching a list of customers and their checkouts, just to see if they have any (likely case in Dashboard 2.0):
```
{
  customers(first: $first) {
    edges {
      node {
        checkout {
          token
        }
      }
    }
  }
}
```
This would create empty checkouts for every user. I think that for now I'd like to leave it as it is.

PR #3727, opened 2019-02-15
Fix bug where cart can not be got or created when user has multiple carts

jxltom:
Closes https://github.com/mirumee/saleor/issues/3728

The ```checkoutCreate``` or ```checkoutCustomerAttach``` could create multiple checkouts for a same user. Then ```get_or_create_user_cart``` will raise ```saleor.apps.checkout.models.Cart.MultipleObjectsReturned: get() returned more than one Cart``` error since it can not handle this case. This PR fixes this issue.

### Pull Request Checklist

### DIFF
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -175,7 +175,11 @@ def get_or_create_user_cart(user, cart_queryset=Cart.objects.all()):
     defaults = {
         'shipping_address': user.default_shipping_address,
         'billing_address': user.default_billing_address}
-    return cart_queryset.get_or_create(user=user, defaults=defaults)[0]
+
+    cart = cart_queryset.filter(user=user).first()
+    if cart is None:
+        cart = Cart.objects.create(user=user, **defaults)
+    return cart
 
 
 def get_anonymous_cart_from_token(token, cart_queryset=Cart.objects.all()):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add instructions for using local assets in Docker - #3723 by @michaljelonek
 - Remove unused imports - #3645 by @jxltom
 - Add discount section - #3654 by @dominik-zeglen
+- Fix bug where cart can not be got or created when user has multiple carts - #3727 by @jxltom
 
 
 ## 2.3.0


saleor/checkout/utils.py
```diff
@@ -175,7 +175,11 @@ def get_or_create_user_cart(user, cart_queryset=Cart.objects.all()):
     defaults = {
         'shipping_address': user.default_shipping_address,
         'billing_address': user.default_billing_address}
-    return cart_queryset.get_or_create(user=user, defaults=defaults)[0]
+
+    cart = cart_queryset.filter(user=user).first()
```
maarcingebala: Wouldn't we achieve the same with passing `user` to `defaults` in the original implementation?

jxltom: ```get_or_create``` can not be used since it will raise ```MultipleObjectsReturned``` error. We should use ```filter``` instead.

maarcingebala:
@jxltom Could you provide steps to reproduce the issue?

jxltom:
Sure, I created an issue for tracing that https://github.com/mirumee/saleor/issues/3728.

saleor/checkout/utils.py
Pacu2: I feel like that's a workaround, and we should just check all the places where one can create a cart, and make sure that there's only one present at all the time.

jxltom: I think there is nothing wrong if a user has multiple carts? We only need to make sure in storefront or API only the latest changed cart is returned when accessing user cart.

maarcingebala: I've looked into it and this is the right place to fix the issue, but the problem is that we're using this function only in Storefront 1.0 and in API we duplicate logic for creating cart instances. We have to try to reuse the core logic as much as possible. Based on your implementation @jxltom I'll try to refactor it a bit more since I'm already into it.

maarcingebala: Also I don't think that we shuold allow creating more than one cart instance for a user at once. Is there any use-case for that? It would be much simpler to assume that user can have only one and optimize the code aginst that case.

jxltom: No problem for that.

As for multiple cart for single user, I can't think of any user cases for now but it may require some work to make sure every user only has one cart since ```checkoutCreate``` or ```CheckoutCustomerAttach``` or storefront 1.0 can all create checkout for users. We need to adjust all of these APIs.

On the other way, I can't think of any harm if a user has multiple carts. We only need to make sure we always use the last changed cart. It makes the life easier.

maarcingebala: About your first remark - that's the current problem. There are multiple places in our code where a new cart instance is created but we're not reusing the same logic which may lead to bugs and code duplication.

jxltom: Yes, that requires some enhancement if we only allow user has one cart.

maarcingebala:
Continued in #3761.

PR #3735, opened 2019-02-18
Support partially charged and partially refunded payment status

jxltom:
Fixes https://github.com/mirumee/saleor/blob/68c3b67e4118384c423d8cc1aa579a341a4a0a5a/saleor/payment/__init__.py#L85-L89

If a customer authorizes 10USD but the shop manager only captures 5USD, the payment status should be partially-charged. If 10USD is captured from the customer, the shop manager refunds only 5USD, the payment status should be partially refunded.

Introducing partially-charged and partiallly refunded can simplify many order/payment logic, and they can also be used in further development for other purposes. They also desscribes payment status more clearly and granularly.

This PR addes partially charged and partially refunded support.

### Pull Request Checklist


saleor/dashboard/order/forms.py
```diff
@@ -252,18 +252,25 @@ class OrderNoteForm(forms.Form):
 
 
 class ManagePaymentForm(forms.Form):
```
NyanKiyoshi: Wouldn't it be better to call it:
```suggestion
class BasePaymentForm(forms.Form):
```

So the user directly knows it's an abstract class?

saleor/dashboard/order/forms.py
```diff
@@ -252,18 +252,25 @@ class OrderNoteForm(forms.Form):
...
 
+    clean_status = None
+
     def __init__(self, *args, **kwargs):
         self.payment = kwargs.pop('payment')
         super().__init__(*args, **kwargs)
 
     def clean(self):
-        if self.payment.charge_status != self.clean_status:
+        # Convert clean_status to a tuple if it is not a tuple or list yet
```
NyanKiyoshi: Is it really needed? Why not just making it mandatory to put the values into an iterable? Those instructions are executed during the `clean()` step, it performs checks that we could prevent from being false.

saleor/dashboard/order/forms.py
```diff
@@ -309,23 +319,16 @@ def refund(self):
         return self.try_payment_action(gateway_refund)
 
 
-class VoidPaymentForm(forms.Form):
+class VoidPaymentForm(ManagePaymentForm):
+
+    clean_status = ChargeStatus.NOT_CHARGED
```
NyanKiyoshi: From above suggestion:
```suggestion
    clean_status = ChargeStatus.NOT_CHARGED,
```

saleor/dashboard/order/forms.py
```diff
@@ -309,23 +319,16 @@ def refund(self):
...
-                pgettext_lazy(
-                    'Payment form error',
-                    'Only pre-authorized payments can be voided'))
-
-    def payment_error(self, message):
-        self.add_error(
-            None, pgettext_lazy(
-                'Payment form error', 'Payment gateway error: %s') % message)
+        self.payment = kwargs.pop('payment')
+        self.fields.pop('amount')
```
NyanKiyoshi: What is the reason for that?

saleor/payment/models.py
```diff
@@ -149,16 +149,19 @@ def can_capture(self):
...
-            self.charge_status == ChargeStatus.CHARGED
-            and self.get_total() > self.get_captured_amount())
+            self.charge_status == ChargeStatus.PARTIALLY_CHARGED)
         return self.is_active and (self.not_charged or not_fully_charged)
 
     def can_void(self):
         return self.is_active and self.not_charged and self.is_authorized
 
     def can_refund(self):
+        can_refund_charge_status = (
```
NyanKiyoshi: We could make it constant in the class (`self.CAN_REFUND_CHARGE_STATUS`). I'm not sure Python  make those local variables static.

saleor/payment/utils.py
```diff
@@ -351,8 +351,14 @@ def _gateway_postprocess(transaction, payment):
     transaction_kind = transaction.kind
 
     if transaction_kind in [TransactionKind.CHARGE, TransactionKind.CAPTURE]:
-        payment.charge_status = ChargeStatus.CHARGED
         payment.captured_amount += transaction.amount
+
+        # Set payment charge status to fully charged
+        # only if there is no more amount needs to charge
+        payment.charge_status = ChargeStatus.PARTIALLY_CHARGED
```
NyanKiyoshi: Why not putting it into the `else` block?

jxltom: I feels that It's more condense, which means ```PARTIALLY_CHARGED``` is by default, only set to ```FULLY_CHARGED``` under some circumstances.

saleor/dashboard/order/forms.py
jxltom: ```VoidPaymentForm``` form does not require amount, so we have to pop amount field otherwise it will be rendered in frontend.

saleor/payment/utils.py
NyanKiyoshi: But we get to write into a variable first, then, we may override it again. That's bad as we write twice in memory, even though we will hit the CPU cache. It's wasting CPU cycles and I/O time for nothing :confused:

saleor/dashboard/order/forms.py
NyanKiyoshi: Maybe we could add a comment so we know why it's there? :+1:

saleor/payment/utils.py
jxltom: It is just a simple string variable assignment instead of db operations or looping or large data structure assignment. I assume it is not heavy at all and I prefer a better readability.

saleor/dashboard/order/forms.py
jxltom: Will do. :D

saleor/payment/utils.py
NyanKiyoshi: Yes, but the issues is that if `payment.get_charge_amount() <= 0` is `True`, we will have written twice in the variable. The issue with that is that putting the instruction in a simple `else` block would only be a simple jump CPU instruction, where only one write happen. Doing two write to maybe discard the first one is wasting important processing time.

jxltom:
```
a = init_value
if condition:
    a = other_value
```

You could consider the first assignment is like variable initialization. It has better readability. A string assignment assignment probably only consumes one cpu cycle which is super fast. It is not like doing any I/O operations, db operations or looping. It is not writing embedded system and there is no need spending time in optimizing these cases.

saleor/payment/models.py
jxltom: After a second thought, I decide not to refactor it since ```can_refund_charge_status``` is not used somewhere else. Besids, even it is defined as ```CAN_REFUND_CHARGE_STATUS```, there is no way to prevent changing it since there is actually no constant (no ```const``` as other language) in Python.

jxltom:
@NyanKiyoshi Thanks for the valuable comments! I've already updated the PR accordingly.

saleor/payment/__init__.py
```diff
@@ -74,22 +74,27 @@ class TransactionKind:
 
 class ChargeStatus:
     """
-    - Charged: Funds were taken off the customer founding source, partly or
-               completely covering the payment amount.
     - Not charged: No funds were take off the customer founding source yet.
+    - Partially charged: Funds were taken off the customer founding source,
```
michaljelonek: Shouldn't this be `customer's funding source`?

saleor/payment/migrations/0008_merge_20190214_0447.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('payment', '0007_auto_20190206_0938'),
+        ('payment', '0007_auto_20190125_0242'),
+    ]
+
+    operations = [
```
michaljelonek: Empty migration?

saleor/payment/migrations/0009_convert_to_partially_charged_and_partially_refunded.py
```diff
@@ -0,0 +1,53 @@
...
+            payment.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('payment', '0008_merge_20190214_0447'),
+    ]
+
+    operations = [
```
michaljelonek: Think you might need to switch 0009 and 0010 around, no? As to me it looks like you operate on status that the next migration adds.

saleor/payment/__init__.py
jxltom: Thansk, will update, sounds better

saleor/payment/migrations/0008_merge_20190214_0447.py
jxltom: It is because after this PR is submitted, there is migration with same number ```007``` is made in saleor. So I have to merge these two migrations as django reminds me.

saleor/payment/migrations/0009_convert_to_partially_charged_and_partially_refunded.py
jxltom: It should fine since this migrations only update the choices which actually doesn't effect the database.

saleor/payment/__init__.py
jxltom: Oh damn I just noticed ```funding``` is misspelled. :crying_cat_face:

jxltom:
@michaljelonek Thanks for reviewing this, the PR is updated. :D

jxltom:
@maarcingebala Do you mind having a look for this? I feel this PR is ready now. It describes the payment status better. With this, we may save a lot of logic related with order status such as get order status by calculating payment's captured amount as well as other logic.

maarcingebala:
Sure, I'll review the PR tomorrow!

saleor/dashboard/order/forms.py
```diff
@@ -251,19 +251,22 @@ class OrderNoteForm(forms.Form):
...
-class ManagePaymentForm(forms.Form):
+class BasePaymentForm(forms.Form):
+
     amount = forms.DecimalField(
         label=pgettext_lazy(
             'Payment management form (capture, refund, void)', 'Amount'),
         max_digits=settings.DEFAULT_MAX_DIGITS,
         decimal_places=settings.DEFAULT_DECIMAL_PLACES)
 
+    clean_status = tuple()
```
maarcingebala: As I understand this tuple holds statuses that are allowed to perform an action that the form performs, right? We could rename it to something like `precondition_statuses` to make it more clear. `clean_status` could be interpreted in different ways. 

Side note - maybe that's outside the scope of this PR but shouldn't this validation live outside of forms? That would make it easier to reuse it the API.

saleor/dashboard/order/forms.py
```diff
@@ -251,19 +251,22 @@ class OrderNoteForm(forms.Form):
...
+
     def __init__(self, *args, **kwargs):
         self.payment = kwargs.pop('payment')
         super().__init__(*args, **kwargs)
 
     def clean(self):
-        if self.payment.charge_status != self.clean_status:
+        if self.payment.charge_status not in self.clean_status:
             raise forms.ValidationError(self.clean_error)
```
maarcingebala: Since we're changing this form I think we could also define some base `clean_error` message. I know the subclasses implement it but it would be elegant if we provided a default one, what do you think?

saleor/order/models.py
```diff
@@ -49,11 +49,12 @@ def ready_to_capture(self):
...
         Orders ready to capture are those which are not draft or canceled and
-        have a preauthorized payment.
+        have a preauthorized payment. The preauthorized payment can not
+        already be partially or fully captured.
         """
         qs = self.filter(
-            payments__is_active=True, payments__charge_status__in=[
-                ChargeStatus.NOT_CHARGED, ChargeStatus.CHARGED])
+            payments__is_active=True,
+            payments__charge_status=ChargeStatus.NOT_CHARGED)
```
maarcingebala: Shouldn't that include partially charged orders as well? They have still some amount to capture.

saleor/order/utils.py
```diff
@@ -124,7 +124,11 @@ def cancel_order(order, restock):
 
     payments = order.payments.filter(
         is_active=True,
-        charge_status__in=[ChargeStatus.NOT_CHARGED, ChargeStatus.CHARGED])
+        charge_status__in=[
```
maarcingebala: Since we're including most of the statuses, maybe it would be more readable if we excluded full refunded ones?

saleor/payment/gateways/dummy/__init__.py
```diff
@@ -105,7 +106,13 @@ def process_payment(payment_information: Dict, connection_params) -> Dict:
     responses = [authorize(payment_information, connection_params)]
     if charge_status == ChargeStatus.NOT_CHARGED:
         return responses
+
+    # Only capture half of the authorized amount when set to partially charged
+    if charge_status == ChargeStatus.PARTIALLY_CHARGED:
```
maarcingebala: I'm not sure about this - we shouldn't know exactly what amount to capture, as partial captures should be only allowed through the dashboard where we should render an input to provide amount.

saleor/dashboard/order/forms.py
maarcingebala: I think we should use methods that are available in the Payment model, such as `can_capture`, `can_refund` etc. There should be only one place that defined whether a particular action can be performed.

maarcingebala:
@jxltom Have you actually tested this PR with any gateway? I looked at the docs of gateways that we support and actually, each of them handles partial operations differently:
- in Braintree we would have to implement [submit_for_partial_settlement](https://developers.braintreepayments.com/reference/request/transaction/submit-for-partial-settlement/python) function in order to have partial captures
- Stripe automatically releases rest of authorized amount if you perform partial capture (I think it's not a problem for us but it would be nice to check if that works).

saleor/order/models.py
jxltom: I'm not sure about braintree, in stripe if you authorize 10USD but the staff only capture 8USD, the rest 2USD will be automatically released and can not be captured anymore. So in this case, the payment which has already been captured partially can not be captured anymore...

What about braintree?

saleor/payment/gateways/dummy/__init__.py
jxltom: I just would like to make it easier to test since it is dummy gateway anyway. I assume decimal / 2 always works as partial charged.

But yeah maybe it is better to do it in dashboard.

jxltom:
@maarcingebala Thanks for the review! Yeah, I've used this PR with Stripe and dummy gateway in my own fork, it works well so far. Sadly I don't have environment to test braintree and razorpay.

saleor/order/models.py
jxltom: Update: If we decide not to support **multple partial capture** as like implementing ```submit_for_partial_settlement``` in braintree, partially charged orders won't have any amount to capture since the rest amount is released automatically.

saleor/dashboard/order/forms.py
jxltom: Indeed! I've refactored it and now payment's built-in methods are used for validation.

jxltom:
@maarcingebala Thanks for the insightful review, the PR is already updated accordingly.

maarcingebala:
I've read through the code again and I think it already provides a good value and we can merge it to master. If any gateway needs further adjustments, we can do that in follow-up PRs.

jxltom:
That's great! Thanks for taking care of it.

PR #3745, opened 2019-02-19
Create emtpy checkout when accessing checkout if user has no available checkouts in me API

jxltom:
Continue https://github.com/mirumee/saleor/pull/3725

@maarcingebala Thanks for the comments. It is not good to create empty checkouts for each user especially in dashboard 2.0 indeed. I did some update on the PR which will only create emtpy checkout for ```me->checkout``` API if this user doesn't have any available checkouts. The remaining API such as ```customers->checkout``` or ```user->checkout``` which is used in dashboard is still kept same which won't create empty checkouts.

Free feel to close it if it's still not suitable for Saleor. :d

### Pull Request Checklist


maarcingebala:
Thanks for opening this, I appreciate your effort but our original assumption was to create checkouts when they're needed and I'd like to stick to that implementation right now and see how it works in the longer run ;)

jxltom:
No problem at all. 🤝

PR #3754, opened 2019-02-20
Support fulfill order with 0 quantity only if total quantity is larger than 0 in dashboard and API

jxltom:
If an order has purchased 1 product A and 1 product B, it makes sense to only fulfill 1 product A or only 1 product B. This PR fixes this issue in dashboard as well as in API

Update: closes https://github.com/mirumee/saleor/issues/3758

### Screenshots

Before:
screenshot from 2019-02-20 14-23-02

After:
screenshot from 2019-02-20 17-18-28

### Pull Request Checklist


saleor/order/models.py
```diff
@@ -315,6 +315,8 @@ class Meta:
     def __str__(self):
         return self.product_name
 
+    __repr__ = __str__
```
maarcingebala: This change seems unrelated, I'd remove it.

saleor/dashboard/order/forms.py
```diff
@@ -614,6 +614,16 @@ def __init__(self, *args, **kwargs):
         for form in self.forms:
             form.empty_permitted = False
 
+    def clean(self):
+        if any(self.errors):
```
maarcingebala: Why do we need that check? It's rather uncommon in clean methods and in our codebase as well I think. I've tested this code and if we removed it and below used `form.cleaned_data.get('quantity', 0)`, all errors would be returned properly.

saleor/dashboard/order/forms.py
```diff
@@ -614,6 +614,16 @@ def __init__(self, *args, **kwargs):
...
+            return
+
+        total_quantities = sum(
+            form.cleaned_data['quantity'] for form in self.forms)
+        if total_quantities <= 0:
```
maarcingebala: I'm not 100% if we need this validation. If we didn't have it, fulfilling 0 quantities for all lines would redirect the user to order details view with a message "No lines fulfilled" which quite makes sense.

saleor/dashboard/order/views.py
```diff
@@ -591,7 +591,7 @@ def fulfill_order_lines(request, order_pk):
         request.POST or None, queryset=FulfillmentLine.objects.none(),
         initial=initial)
     all_forms_valid = all([line_form.is_valid() for line_form in formset])
-    if all_forms_valid and form.is_valid():
+    if all_forms_valid and formset.is_valid() and form.is_valid():
```
maarcingebala: Maybe we should rename `all_forms_valid` to sth like `are_line_forms_valid` - now this naming suggests that literally all forms used here are valid while in the `if` check we still check `formset` and `form`.

saleor/order/models.py
```diff
@@ -367,7 +369,7 @@ class FulfillmentLine(models.Model):
         OrderLine, related_name='+', on_delete=models.CASCADE)
     fulfillment = models.ForeignKey(
         Fulfillment, related_name='lines', on_delete=models.CASCADE)
-    quantity = models.IntegerField(validators=[MinValueValidator(1)])
+    quantity = models.IntegerField(validators=[MinValueValidator(0)])
```
maarcingebala: With this change, this field basically becomes a `PositiveIntegerField`.

saleor/dashboard/order/forms.py
jxltom: It's just for bypassing all following validations in advance if there are errors already, which makes the clean method faster. I will remove it if it's not suitable for saleor codebase.

saleor/dashboard/order/forms.py
jxltom: But that gives wrong impression to users that it looks like the fulfillment is succeeded. Actually it has errors and the current form should reflect the errors. The user would like to fulfill the order indeed and so we shouldn't redirect to other pages if the form has errors.

saleor/dashboard/order/views.py
jxltom: Will do. :D

saleor/order/models.py
jxltom: Sounds more cleaner.

saleor/dashboard/order/forms.py
maarcingebala: OK, let's leave this validation then 👍

jxltom:
@maarcingebala The PR is revised accordingly.

PR #3755, opened 2019-02-20
Fix order weight mechanism

### LINKED ISSUE #3710: Order.get_total_weight fails if related product was removed from the database
Orders and order lines were designed to work even if some of the ordered products were removed from the database. The `Order.get_total_weight` method operates on a relationship with product variants and after a product is removed it will fail with an error: `'NoneType' object has no attribute 'get_weight'`. 

1. Place an order with one product.
2. Remove this product from the database.
3. Run `order.get_total_weight` method. This may be also observed in the Dashboard 2.0 order details view.

### What I expected to happen
We have to investigate how and when this method is used. There is also a `weight` field in the Order model, but I also don't know when it's used. Maybe it is the API that uses this method, while it should use the field.

Kwaidan00:
I want to merge this change because it fixes using `weight` field in `Order`.

We would like to store order data even if product related with `orderLine` was deleted. To accomplish that, we cannot use an old implementation of `Order.get_total_weight`, as it was mention in #3710. On the other hand, `Order.weight` was totally unused.

I could:
- add `weight` field to `orderLine` mode and copy value from `ProductVariant`;
- keep updated `weight` field in `Order` model.

I propose the second solution.

~~In addition, this PR resolves problem with creating order from draft with lines related to non-existing product. In that case I assumed that user cannot create order from draft.~~

EDIT (@maarcingebala): I've removed the above part from PR, we should handle that separately.

I didn't remove `Order.get_total_weight` method to keep common protocol with `Cart`. It is used when available shipping method is calculated. 

It fixes #3710 

### Pull Request Checklist

### DIFF
--- a/saleor/dashboard/order/forms.py
+++ b/saleor/dashboard/order/forms.py
@@ -362,12 +362,11 @@ class CancelOrderLineForm(forms.Form):
         super().__init__(*args, **kwargs)
 
     def cancel_line(self):
-        if self.line.variant:
-            if self.line.variant.track_inventory:
-                deallocate_stock(self.line.variant, self.line.quantity)
-            order = self.line.order
-            delete_order_line(self.line)
-            recalculate_order(order)
+        if self.line.variant and self.line.variant.track_inventory:
+            deallocate_stock(self.line.variant, self.line.quantity)
+        order = self.line.order
+        delete_order_line(self.line)
+        recalculate_order(order)
 
 
 class ChangeQuantityForm(forms.ModelForm):
@@ -402,13 +401,12 @@ class ChangeQuantityForm(forms.ModelForm):
     def save(self):
         quantity = self.cleaned_data['quantity']
         variant = self.instance.variant
-        if variant:
-            if variant.track_inventory:
-                # update stock allocation
-                delta = quantity - self.initial_quantity
-                allocate_stock(variant, delta)
-            change_order_line_quantity(self.instance, quantity)
-            recalculate_order(self.instance.order)
+        if variant and variant.track_inventory:
+            # update stock allocation
+            delta = quantity - self.initial_quantity
+            allocate_stock(variant, delta)
+        change_order_line_quantity(self.instance, quantity)
+        recalculate_order(self.instance.order)
         return self.instance
 
 
--- a/saleor/dashboard/order/forms.py
+++ b/saleor/dashboard/order/forms.py
@@ -16,7 +16,7 @@ from ...order import OrderStatus
 from ...order.models import Fulfillment, FulfillmentLine, Order, OrderLine
 from ...order.utils import (
     add_variant_to_order, cancel_fulfillment, cancel_order,
-    change_order_line_quantity, recalculate_order)
+    change_order_line_quantity, delete_order_line, recalculate_order)
 from ...payment import ChargeStatus, CustomPaymentChoices, PaymentError
 from ...payment.utils import (
     clean_mark_order_as_paid, gateway_capture, gateway_refund, gateway_void,
@@ -362,11 +362,12 @@ class CancelOrderLineForm(forms.Form):
         super().__init__(*args, **kwargs)
 
     def cancel_line(self):
-        if self.line.variant and self.line.variant.track_inventory:
-            deallocate_stock(self.line.variant, self.line.quantity)
-        order = self.line.order
-        self.line.delete()
-        recalculate_order(order)
+        if self.line.variant:
+            if self.line.variant.track_inventory:
+                deallocate_stock(self.line.variant, self.line.quantity)
+            order = self.line.order
+            delete_order_line(self.line)
+            recalculate_order(order)
 
 
 class ChangeQuantityForm(forms.ModelForm):
@@ -401,12 +402,13 @@ class ChangeQuantityForm(forms.ModelForm):
     def save(self):
         quantity = self.cleaned_data['quantity']
         variant = self.instance.variant
-        if variant and variant.track_inventory:
-            # update stock allocation
-            delta = quantity - self.initial_quantity
-            allocate_stock(variant, delta)
-        change_order_line_quantity(self.instance, quantity)
-        recalculate_order(self.instance.order)
+        if variant:
+            if variant.track_inventory:
+                # update stock allocation
+                delta = quantity - self.initial_quantity
+                allocate_stock(variant, delta)
+            change_order_line_quantity(self.instance, quantity)
+            recalculate_order(self.instance.order)
         return self.instance
 
 
--- a/saleor/order/models.py
+++ b/saleor/order/models.py
@@ -312,12 +312,12 @@ class OrderLine(models.Model):
         return self.product_name
 
     def save(self, *args, **kwargs):
-        print(kwargs)
         if not self.pk:
             if ('order' in kwargs and 'variant' in kwargs
                     and 'quantity' in kwargs):
                 order = kwargs['order']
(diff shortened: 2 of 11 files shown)


saleor/order/models.py
```diff
@@ -315,6 +311,16 @@ class Meta:
     def __str__(self):
         return self.product_name
 
+    def save(self, *args, **kwargs):
+        print(kwargs)
```
jxltom: Addtional print :D

Kwaidan00: Shame on me... :(

saleor/dashboard/order/forms.py
```diff
@@ -362,11 +362,12 @@ def __init__(self, *args, **kwargs):
...
     def cancel_line(self):
-        if self.line.variant and self.line.variant.track_inventory:
-            deallocate_stock(self.line.variant, self.line.quantity)
-        order = self.line.order
-        self.line.delete()
-        recalculate_order(order)
+        if self.line.variant:
+            if self.line.variant.track_inventory:
+                deallocate_stock(self.line.variant, self.line.quantity)
+            order = self.line.order
```
maarcingebala: How does canceling order lines work, when there is no variant then? Shouldn't we trigger `delete_order_line` and `recalculate_order` always?

saleor/order/models.py
```diff
@@ -315,6 +311,16 @@ class Meta:
     def __str__(self):
         return self.product_name
 
+    def save(self, *args, **kwargs):
+        if not self.pk:
```
maarcingebala: Every additional logic in `model.save()` adds magic to the behavior and here we're also saving a different model under the hood. From what I see we already have the logic to update the weight in `add_variant_to_order`, so why do we need it here as well?

saleor/dashboard/order/forms.py
Kwaidan00: I assumed that you can not cancel that line. If we allow that, we could not calculate new `order.weight` (due to lack of variant weight). Calling `delete_order_line` in that case will raise an exception.
Maybe we should have an ability to cancel **all** lines related with non-existing product variants? Then we could calculate `order.weight`.

saleor/order/models.py
Kwaidan00: Ok, I'll change that place. And `conftest` and `random_data`, where `order.lines.create` is used.

saleor/dashboard/order/forms.py
maarcingebala: So if we had this case:
- as an admin, I've created a draft order with two lines,
- later I've removed a product variant that was related to one of those lines,
- since I cannot finalize my draft order now, I need to remove the line related to the non-existing variant

It feels totally valid to be able to cancel the line. After removing it, you'd calculate new order weight by iterating over existing lines and summing their variant weights.

Kwaidan00: You're right, but let's consider that case:

- I've created a draft order with three lines,
- I've removed two different product variant that was related to different two lines,
- I would like to remove one of that line (call it *line A*)

I can not calculate weight by iterating over existing lines, because order still contains line without variant. I can not calculate weight by subtraction, because I have no variant related with *line A*.

I see one solution now. Let's assume that invalid draft order (I mean: *draft order with lines related to non-existing variants*) **could** have wrong weight. I will calculate order weight by iterating existing variants - in your case it results proper weight, in my case it results wrong weight.
But I don't like that solution. ;)

maarcingebala: If two lines are broken you would have to remove both of them anyway and eventually you'll end up with the weight equal the weight of existing variants.

maarcingebala: Because of that I also think that we could always calculate weight by iterating over lines instead of doing substractions.

tests/api/test_order.py
```diff
@@ -416,6 +416,21 @@ def test_can_finalize_draft_order_no_order_lines(order):
     assert errors[0].message == 'Could not create order without any products.'
 
 
+def test_can_finalize_draft_order_non_existing_variant(order_with_lines):
+    order = order_with_lines
+    line = order.lines.first()
+    variant = line.variant
+    variant.delete()
+    line.refresh_from_db()
```
NyanKiyoshi: We probably could tell the ORM that we are only looking for the `variant` field.

tests/api/test_order.py
```diff
@@ -416,6 +416,21 @@ def test_can_finalize_draft_order_no_order_lines(order):
...
+    assert line.variant is None
+
+    order.refresh_from_db()
```
NyanKiyoshi: Is this really changing anything to the order's data?

tests/api/test_order.py
```diff
@@ -658,6 +673,25 @@ def test_draft_order_line_update(
     assert data['errors'][0]['field'] == 'quantity'
 
 
+def test_draft_order_update_with_non_existing_product(
```
NyanKiyoshi: We could use a little bit of blank lines to differentiate the groups in there.

tests/test_order.py
```diff
@@ -515,3 +517,70 @@ def test_add_order_note_view(order, authorized_client, customer_user):
     assert get_redirect_location(response) == redirect_url
     order.refresh_from_db()
     assert order.customer_note == customer_note
+
+
+def _calculate_order_weight_from_lines(order):
+    weight = Weight(kg=0)
```
NyanKiyoshi: It would probably be better to use `zero_weight` there.

tests/test_order.py
```diff
@@ -515,3 +517,70 @@ def test_add_order_note_view(order, authorized_client, customer_user):
...
+    for line in order:
+        weight += line.variant.get_weight() * line.quantity
+    return weight
+
+
+def test_order_weight(order_with_lines):
```
NyanKiyoshi:
```suggestion
def test_order_weight_is_correct(order_with_lines):
```

tests/test_order.py
```diff
@@ -515,3 +517,70 @@ def test_add_order_note_view(order, authorized_client, customer_user):
...
+    order_with_lines.refresh_from_db()
+
+    assert order_with_lines.weight == _calculate_order_weight_from_lines(
+        order_with_lines)
+
+
+def test_order_weight_change_line_quantity(order_with_lines):
+    line = order_with_lines.lines.first()
+    new_quantity = line.quantity + 2
+    change_order_line_quantity(line, new_quantity)
```
NyanKiyoshi:
```suggestion

    change_order_line_quantity(line, new_quantity)
```

tests/test_order.py
```diff
@@ -515,3 +517,70 @@ def test_add_order_note_view(order, authorized_client, customer_user):
...
+def test_order_weight_delete_line(order_with_lines):
+    line = order_with_lines.lines.first()
+    delete_order_line(line)
+
+    assert order_with_lines.weight == _calculate_order_weight_from_lines(
+        order_with_lines)
+
+
+def test_get_order_weight_non_existing_product(order_with_lines, product):
+    # Removing product should not affect order's weight
```
NyanKiyoshi:
```suggestion
    """Removing product should not affect the order's weight."""
```

saleor/order/utils.py
```diff
@@ -163,8 +182,13 @@ def attach_order_to_user(order, user):
 
 
 def add_variant_to_order(
```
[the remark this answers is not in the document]
maarcingebala: We could consider it, but it's beyond the scope of this PR.

tests/test_order.py
maarcingebala: In tests, we rather use line comments like this; what you suggest is more suitable for docstrings.

PR #3756, opened 2019-02-20
Add json fields to models for content/description

### LINKED ISSUE #3744: Store rich-text content in JSON fields
We're introducing a new rich-text editor for Dashboard 2.0 (#3067). This editor stores output as JSON and we need to also extend our backend to support that. We need to extend models that would use the editor with an additional field for storing JSON. Those fields will be used by Dashboard 2.0 and Storefront 2.0. The static storefront and dashboard would continue using existing text/HTML fields.

**Required changes**
1. Add the following model fields:
- `Page.content_json`
- `Product.description_json`
- `Collection.description_json`
- `Category.description_json`

Not sure if we need to use `JSONField` or we could simply use `TextField` with serialized JSON as content - that is to be agreed with @dominik-zeglen about which type works better with Draftail and can be easily rendered in the storefront.

2. Adjust GraphQL API to save content coming from Dashboard 2.0 into the above fields. As for the queries, I think we could keep the original fields returning text/HTML as they are and just extend the types with JSON counterparts.

Last thing to consider is how to properly handle SEO fields. I think that in Dashboard 2.0 we should extract plain text from the editor (@dominik-zeglen has to check if this is possible in Draftail) and use that as SEO description, if it wasn't explicitely overriden by user.

michaljelonek:
Resolves #3744. ~We agreed on a `TextField` in the end, since that's how it will be sent through API.~

### Pull Request Checklist


patrys:
Out of curiosity: why use a TextField over a JSONField? How do we validate that the values are valid JSON?

As for the schema, could we explicitly introduce a custom `JSON` scalar instead of returning JSON as `String`? There's a great guide at https://github.com/Shopify/graphql-design-tutorial/blob/master/TUTORIAL.md#naming-and-scalars

michaljelonek:
@patrys do we really feel the need to operate on json in the backend? Cause I don't think so. From a backend perspective I'd say that we don't really care, perhaps naming could be better, as it strongly suggests json while in the end treating it as text. Also I'm not sure how easy/hard it would be to search through a jsonb type instead of just text if we consider that different implementation might have different keys that store actual content. 

Also no need for custom scalars. Graphene already has a [JSONString](https://github.com/graphql-python/graphene/blob/master/graphene/types/json.py). All it does is encode and decode it.

patrys:
@michaljelonek I would say yes. Otherwise, it becomes "junk in, junk out". This is an API and we want to prevent buggy clients from breaking all other clients. Ideally, we'd check that it's not just valid JSON but also the correct structure for Draft content.

We'll likely want to update the old storefront to render the same content using https://github.com/springload/draftjs_exporter.

PR #3772, opened 2019-02-25
Customer address management

### LINKED ISSUE #3403: Address management for authenticated customers
Currently, all address-related mutations are available only to staff users (we require specific permissions to use them). We should expose mutations for regular registered customers to change their addresses.

Kwaidan00:
I want to merge this change because it adds an ability to manage addresses by authorized customers via API. Customers can add, delete and update their addresses and can set one of them as default one.

It closes #3403 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -4,7 +4,7 @@ from django.contrib.auth.tokens import default_token_generator
 from django.core.exceptions import ObjectDoesNotExist
 from django.utils.encoding import force_bytes
 from django.utils.http import urlsafe_base64_encode
-from graphql_jwt.decorators import permission_required
+from graphql_jwt.decorators import login_required, permission_required
 from graphql_jwt.exceptions import PermissionDenied
 
 from ...account import emails, models, utils
@@ -21,6 +21,10 @@ from ..core.mutations import BaseMutation, ModelDeleteMutation, ModelMutation
 from ..core.types import Error
 
 
+BILLING_ADDRESS_FIELD = 'default_billing_address'
+SHIPPING_ADDRESS_FIELD = 'default_shipping_address'
+
+
 def send_user_password_reset_email(user, site):
     context = {
         'email': user.email,
@@ -32,8 +36,16 @@ def send_user_password_reset_email(user, site):
     emails.send_password_reset_email.delay(context, user.email)
 
 
-BILLING_ADDRESS_FIELD = 'default_billing_address'
-SHIPPING_ADDRESS_FIELD = 'default_shipping_address'
+def can_edit_address(user, address):
+    """Determine whether the user can edit the given address.
+
+    This method assumes that an address can be edited by:
+    - users with proper permission (staff)
+    - customers who "own" the given address.
+    """
+    has_perm = user.has_perm('account.manage_users')
+    belongs_to_user = address in user.addresses.all()
+    return has_perm or belongs_to_user
 
 
 class CustomerRegisterInput(graphene.InputObjectType):
@@ -461,7 +473,7 @@ class CustomerAddressCreate(ModelMutation):
             description='Fields required to create address', required=True)
 
     class Meta:
-        description = 'Creates address for user'
+        description = 'Create a new address for the customer.'
         model = models.Address
         exclude = ['user_addresses']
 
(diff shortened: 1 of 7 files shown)


saleor/graphql/account/mutations.py
```diff
@@ -453,10 +455,38 @@ def user_is_allowed(cls, user, input):
...
+        exclude = ['user_addresses']
+
+    @classmethod
+    def user_is_allowed(cls, user, input):
+        return user.is_authenticated
+
+    @classmethod
+    def save(cls, info, instance, cleaned_input):
+        user = info.context.user
+        super().save(info, instance, cleaned_input)
```
NyanKiyoshi: It would make more sense to put the user variable after the super call. Or even fully remove it as it is redundant.
Unless I'm missing something out.

saleor/graphql/account/mutations.py
```diff
@@ -453,10 +455,38 @@ def user_is_allowed(cls, user, input):
...
+        instance.user_addresses.add(user)
+        instance.save()
+
+
+def check_is_stuff_or_user_edit_own_address(info, instance):
```
NyanKiyoshi: > check_is_stuff

I'm not sure about what you mean?

saleor/graphql/account/mutations.py
```diff
@@ -453,10 +455,38 @@ def user_is_allowed(cls, user, input):
...
+    user = info.context.user
+    if not (user.has_perm('account.manage_users')
+            or user in instance.user_addresses.all()):
```
NyanKiyoshi: I'm not sure getting all the user addresses is a good solution. It's pretty heavy to do that check such way.

saleor/graphql/account/mutations.py
```diff
@@ -453,10 +455,38 @@ def user_is_allowed(cls, user, input):
...
+        raise PermissionDenied()
+
+
 class AddressUpdate(ModelMutation):
     class Arguments:
-        id = graphene.ID(
-            description='ID of address to update', required=True)
+        id = graphene.ID(description='ID of address to update', required=True)
```
NyanKiyoshi:
```suggestion
        id = graphene.ID(description='ID of the address to update', required=True)
```

saleor/graphql/account/mutations.py
```diff
@@ -466,19 +496,55 @@ class Meta:
...
-    def user_is_allowed(cls, user, input):
-        return user.has_perm('account.manage_users')
+    def clean_instance(cls, info, instance, errors):
+        check_is_stuff_or_user_edit_own_address(info, instance)
+
+
+class CustomerSetDefaultAddress(BaseMutation):
+    class Arguments:
+        id = graphene.ID(
+            required=True, description='ID of address to set as default.')
```
NyanKiyoshi:
```suggestion
            required=True, description='ID of the address to set as default.')
```

saleor/graphql/account/mutations.py
```diff
@@ -466,19 +496,55 @@ class Meta:
...
+
+    @classmethod
+    def mutate(cls, root, info, id, type):
+        errors = []
+        address = cls.get_node_or_error(info, id, errors, 'id', Address)
+        if not address:
+            return CustomerSetDefaultAddress(errors=errors)
+
+        user = info.context.user
+        if not (user.is_authenticated and address in user.addresses.all()):
```
NyanKiyoshi: `all()` is too much and heavy.

tests/api/test_account.py
```diff
@@ -924,17 +928,54 @@ def test_address_update_mutation(
...
+    assert address_data['city'] != address_obj.city
+
+    variables = {
+        'addressId': graphene.Node.to_global_id('Address', address_obj.id),
+        'address': address_data}
+    response = user_api_client.post_graphql(query, variables)
+    content = get_graphql_content(response)
+    data = content['data']['addressUpdate']
+    assert data['address']['city'] == address_data['city']
+    address_obj.refresh_from_db()
```
NyanKiyoshi: Refreshing the `city` field should be enough?

tests/api/test_account.py
```diff
@@ -1029,3 +1093,125 @@ def test_customer_reset_password(
...
+}
+"""
+
+
+def test_customer_set_address_as_default(user_api_client, address):
+    user = user_api_client.user
+    user.default_billing_address = None
+    user.default_shipping_address = None
+    user.save()
+    user.refresh_from_db()
```
NyanKiyoshi: Seems unneeded?

saleor/graphql/account/mutations.py
maarcingebala: Why? I don't see an issue here. We're saving the address instance and assigning it to the user. Only the `instance.save()` below seems superficial, `add()` already saves data in the database.

NyanKiyoshi: The reason that I'm thinking is that the super call is being done after getting the user from the context, which by reading the code line per line makes it seem needed by the super call.
Then, for the second point, it seems that doing `instance.user_addresses.add(info.context.user)` would be pretty straightforward, as the user directly see that we are passing the user. So I suggested that maybe we could simplify to that.

saleor/graphql/account/mutations.py
maarcingebala: I've refactored it a little bit but I'd say that this is acceptable.

tests/api/test_account.py
maarcingebala: Yeah, but again, this solution is acceptable.

PR #3778, opened 2019-02-26
Do not return cancelled fulfillments in me API

jxltom:
In storefront 1.0, the cancelled fulfillment is not listed in user's account page. It will be nice to applly same logic to ```me->orders->fulfillments``` GraphQL API

### Pull Request Checklist


saleor/graphql/order/types.py
```diff
@@ -265,7 +266,12 @@ def resolve_total_balance(self, info):
 
     @staticmethod
     def resolve_fulfillments(self, info):
-        return self.fulfillments.all().order_by('pk')
+        context_kwargs = getattr(info.context, 'kwargs', {})
```
maarcingebala: I'm not 100% convinced about assigning arbitrary variables to `info.context` which in Graphene is a Django request object, but honestly, I also don't see a better idea how to customize the resolver behavior for `me` query.

What about a different approach - would it make sense if we excluded canceled fulfillments for unauthenticated users (customers)?

jxltom: I feel the same that it is not the best approach. It will nice if graphene supports passing context to next levels natively.

jxltom: The ```me``` API requires authentication, and ```me->orders->fulfillments``` which is used by frontend should exclude cancelled orders.

So I think it won't achieve this result if we only exclude cancelled fulfillments for unauthenticated users?

maarcingebala: Sorry, my bad- I was thinking about users who are not staff, so regular customers.

jxltom: This sounds quite reasonable :+1:

jxltom:
@maarcingebala Thanks for the better approach! The PR has been updated.

saleor/graphql/order/types.py
```diff
@@ -265,7 +266,12 @@ def resolve_total_balance(self, info):
 
     @staticmethod
     def resolve_fulfillments(self, info):
-        return self.fulfillments.all().order_by('pk')
+        user = info.context.user
+        if user.is_authenticated and user.is_active and user.is_staff:
```
maarcingebala: It's enough to check only `user.is_staff` here - if it's true, it implies that the user is authenticated which in turn implies that they are active.

jxltom: ```is_staff``` means user is authenticated.

But I think a staff can be not active anymore?

jxltom: Oh, yes, an inactive user can be authenticated at all...

jxltom: But I see there are a lot of code using ```is_active``` in Saleor, maybe let's keep this in case we miss anything?

maarcingebala: `is_active` flag determines if the user is able to log in at all, so first you have to be "active" in order to determine whether you're staff or not.

jxltom: OK, PR is updated.

PR #3781, opened 2019-02-27
Improve user and staff management in dashboard 1.0

jxltom:
This PR addes several improvements for user and staff management in dashboard 1.0

- Fix bug where ```user``` is not passed into customer form and staff form in creation.
- Fix bug where page is not redirecting to staff list after staff detail is updated.
- ~The super user has all permissions by default but it is not shown in staff form. This PR exposes ```is_superuser``` in staff form but disable editing it. Otherwise it is hard to tell a user is superuser or not.~
- ~Add ```is_staff``` to customer form. The users with managing staff permission and superuser are able to assign/remove ```is_staff``` to customers. So there is no need to always create new users when creating staff.~
- Remove all permissions if user is not staff anymore.
- Disable editing superuser's permissions/email/is_superuser/etc if current user is not superuser.
- Only users with managing staff permission and superuser are able to assign/remove ```is_staff``` to customers.
- Closes https://github.com/mirumee/saleor/issues/2000. Staff shouldn't be able to edit other staff's email address.

### Pull Request Checklist


saleor/dashboard/staff/forms.py
```diff
@@ -17,7 +17,7 @@ class StaffForm(forms.ModelForm):
     class Meta:
         model = User
         fields = ['first_name', 'last_name', 'email',
-                  'user_permissions', 'is_active']
+                  'user_permissions', 'is_active', 'is_staff', 'is_superuser']
```
maarcingebala: "Superuser" functionality is hidden in UI on purpose and we don't want to reveal that. We think that superusers are convenient for development but normally, regular staff accounts and permission systems should be used.

maarcingebala: Also, it's a rare case to turn your customers into staff - that's why this is hidden as well. However, if you really wanted to do that, go to the "staff create" view and try to add a staff user with an email of existing customer - this action will succeed and the user will receive "Promote to staff" email.

maarcingebala:
We have two related issues, does this PR by any chance solve any of them?

https://github.com/mirumee/saleor/issues/1966
https://github.com/mirumee/saleor/issues/2000

saleor/dashboard/staff/forms.py
jxltom: > Also, it's a rare case to turn your customers into staff - that's why this is hidden as well. However, if you really wanted to do that, go to the "staff create" view and try to add a staff user with an email of existing customer - this action will succeed and the user will receive "Promote to staff" email.

OK, sounds good, I will remove is_staff in customer form

jxltom: > "Superuser" functionality is hidden in UI on purpose and we don't want to reveal that. We think that superusers are convenient for development but normally, regular staff accounts and permission systems should be used.

Maybe at least we should reveal which user is superuser? I found it's confusing in some API that I have permission but I shouldn't have. At last I found I'm a superuser. :D

jxltom:
I thinks this PR is able to solve both of them. I will do some digging.

jxltom:
After some digging, I think https://github.com/mirumee/saleor/issues/1966 may not be the best approach to avoid privilege elevetion. Staff users should be able to assign any privilges to other staff users. Otherwise we need superuser to do that or a user with all privilges. At this point, I think it is better to keep as it is. Maybe I will do some more digging later, but this PR won't try to sovle this issue.

On the other side, I agree with https://github.com/mirumee/saleor/issues/2000 that staff shouldn't be able to edit other staff's email. One reason for that is this can make other staff can not login anymore.

@maarcingebala I've updated the PR and description, please have a look.

saleor/dashboard/staff/forms.py
maarcingebala: We had a discussion about that in the past (I don't remember if I have it somewhere, it might have been outside of Github), but we decided to hide that information in the UI. If you need to display superusers somehow and manage them through UI I'd recommend doing that in your fork. In Saleor I'd like to leave it as it was before and rely on permissions only.

jxltom: Sure, I've updated the PR.

I will expose superuser but disable editing it in my own fork. The reason for that is hiding the superuser information maks hard to find out why some staff can not edit other staff's profile who is superuser actually.

jxltom:
This PR is ready, I think it's better to fix https://github.com/mirumee/saleor/issues/1966 in another PR.

saleor/dashboard/staff/forms.py
```diff
@@ -44,3 +63,12 @@ def __init__(self, *args, **kwargs):
         if address.last_name:
             placeholder = get_name_placeholder(address.last_name)
             self.fields['last_name'].widget.attrs['placeholder'] = placeholder
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        # Remove all permissions if user is not staff
+        if not cleaned_data['is_staff']:
+            cleaned_data['user_permissions'] = []
```
maarcingebala: It would be nice to add tests for this logic.

saleor/dashboard/staff/forms.py
```diff
@@ -26,14 +26,33 @@ class Meta:
             'email': pgettext_lazy(
                 'Email', 'Email'),
             'is_active': pgettext_lazy(
-                'User active toggle', 'User is active')}
+                'User active toggle', 'User is active'),
+            'is_staff': pgettext_lazy(
+                'User staff toggle', 'User is staff')}
 
     def __init__(self, *args, **kwargs):
         self.user = kwargs.pop('user', None)
```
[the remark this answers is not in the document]
jxltom: Makes sense. But the user argument could still be passed as None, so we may still need to check user is not None in the permission checking logic.

saleor/dashboard/staff/forms.py
jxltom: Sure, test is added

saleor/dashboard/staff/forms.py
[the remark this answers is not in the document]
maarcingebala: Passing `None` is just a mistake in code that would fail immediately since this is a `__init__` method. I don't think that `request.user` is ever `None` in Saleor, it's either AnonymousUser or just a User and this form is intended to be used by authenticated users. We could safely assume that `user` is a required argument and it won't be `None`.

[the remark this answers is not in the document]
jxltom: I agree with request.user should be safe. But I think None can be passed to ```user``` argument during form initialization such as ```StaffForm(request.POST or None, instance=staff_member, user=None)```?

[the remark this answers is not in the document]
jxltom:
```
class Example:
    def __init__(self, *arg, **kwargs):
        pass


Example(somearg=None)
```

I think this code won't raise errors in Python?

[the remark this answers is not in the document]
maarcingebala: Of course, you could pass `None` here but we would you do that? From what I see all `StaffForm` occurrences could use `request.user`. We don't have to check that all arguments that a function takes are non-null if you expect them to be always non-null. E.g. we have the `create_order(cart, **something)` function which expects cart instance. We don't check if you accidentally passed `None` because that's would be your mistake and the code would fail immediately.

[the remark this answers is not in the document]
jxltom: I got what you mean. So we just assume correct types of argument are passed without checking it. It's Python's limitation comparing with other languages with type checking. The business logic shouldn't check the argument types.

saleor/dashboard/staff/forms.py
```diff
@@ -26,14 +26,32 @@ class Meta:
...
+                'User staff toggle', 'User is staff')}
 
     def __init__(self, *args, **kwargs):
-        self.user = kwargs.pop('user', None)
+        # The user argument is required
+        self.user = kwargs.pop('user')
         super().__init__(*args, **kwargs)
-        self.instance.is_staff = True
+
+        # Non-superuser couldn't edit superuser's profile
```
NyanKiyoshi:
```suggestion
        # Non-superusers shouldn't be able to edit a superuser's profile
```

jxltom:
@maarcingebala I think this PR is OK? Let's me know if anything more I should improve?

maarcingebala:
Yes, it's fine now. I'll merge it when the tests pass.

PR #3783, opened 2019-02-27
Support setting as default address directly when creating customer address

jxltom:
Closes https://github.com/mirumee/saleor/issues/3782

### Pull Request Checklist


saleor/graphql/schema.graphql
```diff
@@ -965,7 +980,7 @@ type Mutations {
   customerPasswordReset(input: CustomerPasswordResetInput!): CustomerPasswordReset
   customerRegister(input: CustomerRegisterInput!): CustomerRegister
   customerUpdate(id: ID!, input: CustomerInput!): CustomerUpdate
-  customerAddressCreate(input: AddressInput!): CustomerAddressCreate
+  customerAddressCreate(input: DefaultAddressInput!): CustomerAddressCreate
```
maarcingebala: Instead of introducing a new type which is almost the same as `AddressInput`, we could change the signature of the mutation to: `customerAddressCreate(input: AddressInput, type: AddressTypeEnum): CustomerAddressCreate `. Additionally, we should provide a description for the `type` argument e.g. "A type of address. If provided, the new address will be automatically assigned as the customer's default address of that type."

jxltom: Sounds better, thanks, will do!

jxltom: I remembered why I design like this now, it looks like ```ModelMutation``` only supports ID and input argument, adding more arguments requires override some methods...

jxltom:
@maarcingebala I've updated the PR with the new approach.

PR #3787, opened 2019-02-28
Support is_default_shipping/billing_address for address API

jxltom:
It will be nice to display if an address is user's default shipping/billing address in address API. This PR adds this support. It will be easier for the frontend to display whether an address is default shipping/billing address in address list page.

### Pull Request Checklist


saleor/account/models.py
```diff
@@ -160,6 +160,21 @@ def get_ajax_label(self):
                 address.first_name, address.last_name, self.email)
         return self.email
 
+    def get_addresses(self):
```
maarcingebala: This looks like a good candidate for a custom queryset for the Address model that we could use like: `user.addresses.annotate_default(user)`.

saleor/account/models.py
```diff
@@ -160,6 +160,21 @@ def get_ajax_label(self):
...
+        # Set default shipping/billing address pk to -1
+        # if default shipping/billing address doesn't exist
+        default_shipping_address_pk, default_billing_address_pk = -1, -1
```
maarcingebala: Is it possible to use `None` instead of `-1` here?

saleor/graphql/account/types.py
```diff
@@ -42,6 +48,38 @@ def resolve_country(self, info):
         return CountryDisplay(
             code=self.country.code, country=self.country.name)
 
+    def resolve_is_default_shipping_address(self, info):
+        """
```
maarcingebala: This could be an inline-style comment rather than a docstring, as it is a more internal comment, not really documentation.

saleor/account/models.py
jxltom: Yeah it can be set to None in grammar, but then it will be harder to tell it's because user doesn't have default address or backend doesn't annotate. For now, only ```resolve_addresses``` have ```is_default_shipping/billing_address``` rather than ```resolve_default_shipping/billing_address```. I think it's better to set ```-1``` (```is_default_shipping/billing_address``` will be False instead of None) to indicate the addresses are annotated but user doesn't have default address.

saleor/account/models.py
jxltom: Sorry I didn't get it. Could you give further explanation please?

maarcingebala: I'd create something like this:
```
# account/models.py

class AddressQueryset(models.QuerySet):
    def annotate_default(self, user):
        # logic from `get_addresses`
        return user.addresses.annotate(...)

class Address(models.Model):
    ...
    objects = AddressQueryset.as_manager()
```

And then in the `User` type in addresses resolver you could use `return self.addresses.annotate_default(self)` (where `self` is the `User` instance).

saleor/account/models.py
maarcingebala: I don't know if is a problem if we returned `True/False` when we have information whether an address is the default one or `null` if we don't know. Frontend client would care about the `True` value anyway.

saleor/account/models.py
jxltom: Got it! Yeah it's better to move to queryset manager. Will refactor it tomorrow :D

saleor/account/models.py
jxltom: That's true for the frontend. But I think we could still keep None and False both as falsy cases since it could tell more information? If we treat unknown state as False, technically it could be True in some cases but we just don't know the information.

maarcingebala: So maybe I misunderstood you - I'd prefer to have here three possible values: true/false/null. I only don't like setting `-1` because it's kind of a "magic number" which I'd expected in pieces of code where we have some iteration or operate on counters etc. In terms of Django and model related stuff `null` just seems to me more appropriate.

jxltom: Yeah I agree. I just can't find a better approach. Setting it as ```None``` is just same as the query is not annotated. That's why I set it to ```-1``` as a special case.

Maybe I could define -1 as const with meaningful name?

jxltom: Never mind, i'm silly. There is a difference between variable as None and variable doesn't exist for ```getattr```. I will refactor it tomorrow.

jxltom:
@maarcingebala This PR is updated :handshake:

saleor/graphql/account/types.py
```diff
@@ -67,7 +105,7 @@ class Meta:
         model = get_user_model()
 
     def resolve_addresses(self, info, **kwargs):
-        return self.addresses.all()
+        return self.get_addresses()
```
maarcingebala: Here we could use `self.addresses.annotate_default(self)`, couldn't we? And then we won't need `get_addresses` function anymore.

jxltom: Yeah we could, I'm not sure which one is better, exposing ```get_addresses``` in model level or just accessing queryset manager in graphql API level.

Maybe let's keep it simple, without so many abstract layers.

maarcingebala: Resolvers/views are a piece of code where you usually fetch necessary data from the DB so using querysets there is totally fine in my opinion. Also, querysets are a clear and readable method of encapsulating the logic that you want to perform on database tables. So I think we ended up with a very nice implementation 👍

jxltom: That's true. Thanks for the detailed review.

maarcingebala:
@jxltom Could you resolve changelog conflicts? You could as well rebase this branch with master.

jxltom:
Sure, I merged the upstream master as it is easier to solve the conflicts.

PR #3795, opened 2019-03-04
Several improvements on checkout and payment API

jxltom:
- Remove additional FIXME since it is already fixed
- The billing address of checkout should be required in placing order
- Fix bug where ```cls.add_error``` is incorrectly used
- The amount in order payment create shouldn't be required but checkout's total should be used as default.
- The available payment gateways API should only return enabled gateways as defined in ```CHECKOUT_PAYMENT_GATEWAYS```. If the shop hasn't' enabled a specific gateway, it shouldn't be exposed in API.
- I feel ```can_be_fully_paid``` is more proper than ```is_fully_paid``` in ```ready_to_place_order``` since this method only checks the payment's total but the payment can be not charged at all.

### Pull Request Checklist


saleor/payment/__init__.py
```diff
@@ -108,7 +108,7 @@ class ChargeStatus:
 GATEWAYS_ENUM = Enum(
     'GatewaysEnum',
     {key.upper(): key.lower()
-     for key in settings.PAYMENT_GATEWAYS})
+     for key in settings.CHECKOUT_PAYMENT_GATEWAYS})
```
maarcingebala: That's already handled at the resolver level, see `Checkout.availablePaymentGateways` field.

saleor/graphql/payment/mutations.py
```diff
@@ -28,7 +28,7 @@ class PaymentInput(graphene.InputObjectType):
             'Client-side generated payment token, representing customer\'s '
             'billing data in a secure manner.'))
     amount = Decimal(
-        required=True,
+        required=False,
```
maarcingebala: In that case, let's also change the description of that field: "Total amount of the transaction, including taxes and discounts. If no amount is provided, the checkout total will be used."

saleor/graphql/order/mutations/orders.py
```diff
@@ -290,7 +289,8 @@ class Meta:
     def mutate(cls, root, info, id, amount):
         errors = []
         if amount <= 0:
-            cls.add_error('Amount should be a positive number.')
+            cls.add_error(
+                errors, amount, 'Amount should be a positive number.')
```
maarcingebala: `amount` is a field name so it should put in quotes: `'amount'`

saleor/graphql/order/mutations/orders.py
```diff
@@ -356,7 +356,8 @@ class Meta:
...
```
maarcingebala: Ditto

saleor/checkout/utils.py
```diff
@@ -880,8 +880,8 @@ def create_order(cart: Cart, tracking_code: str, discounts, taxes):
     return order
 
 
-def is_fully_paid(cart: Cart, taxes, discounts):
-    """Check if checkout is fully paid."""
+def can_be_fully_paid(cart: Cart, taxes, discounts):
```
maarcingebala: I think the original name was fine, I don't see a reason why we should change it.

saleor/payment/__init__.py
jxltom: I mean the ```availablePaymentGateways``` will only return the available gateway indeed, but from the docs, you can still find other payment gateways which is from ```GATEWAYS_ENUM```. Since these gateways are not enabled, should we expose it in the API even only in the docs?

saleor/checkout/utils.py
jxltom: Because this method only checks ```payment.total``` instead of ```payement.captured``` which means the order may not be charged at all but only have ability to be charged later. With this reason , I think ```**can_be**_fully_paid``` is more approprate than ```**is**_fully_paid```, what do you think?

saleor/payment/__init__.py
maarcingebala: OK, I think we can change it that way and only return those that are enabled in the enum as well.

saleor/checkout/utils.py
maarcingebala: Ok, I got what you mean. We don't check `captured` amount at this point on purpose - at the point when this method is run we only require the customer to have provided a valid payment method for the required amount.

saleor/checkout/utils.py
```diff
@@ -880,8 +880,8 @@ def create_order(cart: Cart, tracking_code: str, discounts, taxes):
...
+    """Check if checkout can be fully paid with proper payments."""
```
maarcingebala: I'd only change this comment to be more accurate. Maybe something like "Check if provided payment methods cover the checkout's total amount."

saleor/checkout/utils.py
```diff
@@ -902,7 +902,11 @@ def ready_to_place_order(cart: Cart, taxes, discounts):
...
                 'Shipping method is not valid for your shipping address')
-    if not is_fully_paid(cart, taxes, discounts):
+    if not cart.billing_address:
         return False, pgettext_lazy(
-            'order placement error', 'Checkout is not fully paid')
+            'order placement_error', 'Billing address is not set')
+    if not can_be_fully_paid(cart, taxes, discounts):
+        return False, pgettext_lazy(
+            'order placement error',
+            'Checkout can not be fully paid with proper payments')
```
maarcingebala: Similarly, for this error message, I'd say something like: "Provided amount is smaller than checkout's total". Does it sound right?

saleor/checkout/utils.py
jxltom: OK let's only change the comments then.

saleor/checkout/utils.py
jxltom: It's more detailed. It's fine if only we don't add more other validations except for checking payment.total in ```is_fully_paid```.

jxltom: Maybe we could use same message as like in the comments: ```Provided payment methods can not cover the checkout's total amount.```

jxltom:
@maarcingebala I've updated the comments and error message related with ```is_fully_paid``` and kept using this name, though I still feel it's a more appropriate name since it reflects what the method really do.

Please let me know if there is anythinge else improper.

CHANGELOG.md
```diff
@@ -64,13 +65,6 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fix incorrect cart badge location - #3786 by @jxltom
 - Fix storefront styles after bootstrap is updated to 4.3.1 - #3753 by @jxltom
 - Fix logo size in different browser and devices with different sizes - #3722 by @jxltom
-- Rename dumpdata file `db.json` to `populatedb_data.json` - #3810 by @maarcingebala
```
jxltom: @maarcingebala Are these changelog deleted accidentally?

maarcingebala: Oops, yes, I was trying to fix changelog conflict but something went wrong

maarcingebala: Could you fix the Changelog and rebase this branch with master?

jxltom: Sure, done

PR #3808, opened 2019-03-06
Update checkout addresses

jxltom:
- Current ```checkoutShipping/BillingAddressUpdate``` API only supports modify checkout's addresses directly. This PR supports that other addresses can be assigned to checkout by address ID.
- This PR also make the checkout ID in several checkout mutations as required.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/mutations.py
+++ b/saleor/graphql/checkout/mutations.py
@@ -428,7 +428,7 @@ class CheckoutBillingAddressUpdate(CheckoutShippingAddressUpdate):
 
         with transaction.atomic():
             billing_address.save()
-            change_billing_address_in_cart(checkout, billing_address)
+            change_billing_address_in_checkout(checkout, billing_address)
         return CheckoutBillingAddressUpdate(checkout=checkout)
 
 
--- a/saleor/graphql/checkout/mutations.py
+++ b/saleor/graphql/checkout/mutations.py
@@ -353,7 +353,7 @@ class CheckoutShippingAddressUpdate(BaseMutation, I18nMixin):
         arguments = (shipping_address, shipping_address_id)
         if not any(arguments) or all(arguments):
             raise ValidationError(
-                'One and only one of shipping address or shipping address ID '
+                'One and only one shipping address or shipping address ID '
                 'should be provided.')
 
         if shipping_address is not None:
@@ -411,7 +411,7 @@ class CheckoutBillingAddressUpdate(CheckoutShippingAddressUpdate):
         arguments = (billing_address, billing_address_id)
         if not any(arguments) or all(arguments):
             raise ValidationError(
-                'One and only one of billing address or billing address ID '
+                'One and only one billing address or billing address ID '
                 'should be provided.')
 
         if billing_address is not None:
--- a/saleor/graphql/checkout/mutations.py
+++ b/saleor/graphql/checkout/mutations.py
@@ -331,25 +331,42 @@ class CheckoutShippingAddressUpdate(BaseMutation, I18nMixin):
         shipping_address = AddressInput(
             required=False,
             description=(
-                'The mailling address to where the checkout will be shipped.'))
+                'The mailing address to where the checkout will be shipped.'))
         shipping_address_id = graphene.ID(
             required=False,
             description=(
                 'ID of the mailing address '
                 'to where the checkout will be shipped.'))
-                'The mailing address to where the checkout will be shipped.'))
 
     class Meta:
         description = 'Update shipping address in the existing Checkout.'
 
     @classmethod
-    def perform_mutation(cls, _root, info, checkout_id,
-            shipping_address=None, shipping_address_id=Nonev):
-    def perform_mutation(cls, _root, info, checkout_id, shipping_address):
+    def perform_mutation(
+            cls, _root, info, checkout_id,
+            shipping_address=None, shipping_address_id=None):
         checkout = cls.get_node_or_error(
             info, checkout_id, only_type=Checkout, field='checkout_id')
-        shipping_address = cls.validate_address(
-            shipping_address, instance=checkout.shipping_address)
+
+        # Validate that one and only one of
+        # shipping address or shipping address ID is provided
+        arguments = (shipping_address, shipping_address_id)
+        if not any(arguments) or all(arguments):
+            raise ValidationError(
+                'One and only one of shipping address or shipping address ID '
+                'should be provided.')
+
+        if shipping_address is not None:
+            # Validate shipping address
+            # if shipping address is provided directly
+            shipping_address = cls.validate_address(
+                shipping_address, instance=checkout.shipping_address)
+        elif shipping_address_id is not None:
+            # Query and validate shipping address
+            # if shipping address ID is provided
+            shipping_address = cls.get_node_or_error(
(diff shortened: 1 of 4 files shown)


jxltom:
@maarcingebala Do you mind spending some time on these PR https://github.com/mirumee/saleor/pull/3808 and https://github.com/mirumee/saleor/pull/3812? They all related wtih GraphQL's addresses. I hope these two PR could sort it out.

maarcingebala:
@jxltom Sorry for lack of input from me in both PRs - I remember about them but I've been busy with some other things recently. I'll discuss with the team both PR when we have time and we'll provide some feedback.

jxltom:
Update: this PR is rebased with latest master since https://github.com/mirumee/saleor/pull/3891 brings conflicts.

saleor/graphql/checkout/mutations.py
```diff
@@ -325,21 +325,47 @@ class CheckoutShippingAddressUpdate(BaseMutation, I18nMixin):
...
             info, checkout_id, only_type=Checkout, field='checkout_id')
-        shipping_address = cls.validate_address(
-            shipping_address, instance=checkout.shipping_address)
+
+        # Validate that one and only one of
+        # shipping address or shipping address ID is provided
+        arguments = (shipping_address, shipping_address_id)
+        if not any(arguments) or all(arguments):
+            raise ValidationError(
+                'One and only one of shipping address or shipping address ID '
```
NyanKiyoshi:
```suggestion
                'One and only one shipping address or shipping address ID '
```

saleor/graphql/checkout/mutations.py
```diff
@@ -360,32 +386,56 @@ class CheckoutBillingAddressUpdate(CheckoutShippingAddressUpdate):
...
             info, checkout_id, only_type=Checkout, field='checkout_id')
 
-        billing_address = cls.validate_address(
-            billing_address, instance=checkout.billing_address)
+        # Validate that one and only one of
+        # billing address or billing address ID is provided
+        arguments = (billing_address, billing_address_id)
+        if not any(arguments) or all(arguments):
+            raise ValidationError(
+                'One and only one of billing address or billing address ID '
```
NyanKiyoshi:
```suggestion
                'One and only one billing address or billing address ID '
```

tests/api/test_checkout.py
```diff
@@ -723,20 +728,72 @@ def test_checkout_shipping_address_update_invalid_country_code(
...
+        'shippingAddress': shipping_address,
+        'shippingAddressId': address_id}
+
+    response = user_api_client.post_graphql(
+        MUTATION_CHECKOUT_SHIPPING_ADDRESS_UPDATE, variables)
+    content = get_graphql_content(response)
+    data = content['data']['checkoutShippingAddressUpdate']
+    assert len(data['errors']) == 1
+    assert data['errors'][0]['message'] == (
+        'One and only one of shipping address or shipping address ID '
```
NyanKiyoshi:
```suggestion
        'One and only one shipping address or shipping address ID '
```

tests/api/test_checkout.py
```diff
@@ -766,6 +831,56 @@ def test_checkout_billing_address_update(
...
+        'billingAddress': billing_address,
+        'billingAddressId': address_id}
+
+    response = user_api_client.post_graphql(
+        MUTATION_CHECKOUT_BILLING_ADDRESS_UPDATE, variables)
+    content = get_graphql_content(response)
+    data = content['data']['checkoutBillingAddressUpdate']
+    assert len(data['errors']) == 1
+    assert data['errors'][0]['message'] == (
+        'One and only one of billing address or billing address ID '
```
NyanKiyoshi:
```suggestion
        'One and only one billing address or billing address ID '
```

saleor/graphql/checkout/mutations.py
jxltom: Sounds better!

akjanik:
I am doubtful about introducing maybe-optional parameters into the API (`shipping_address_id`). We are trying to be as consistent as possible; in my opinion, it may cause confusion when frontend client sees both fields - `shipping_address_id` and `shipping_address` - as _optional_, decides to not fill any and then gets an error message that, notwithstanding one is required.
AFAIK GraphQL does not specify such a situation when _one_of_ (not explicitly named) parameter is required.

@jxltom What is the use-case for this? I am working mainly on the backend side, but maybe there is a case when it's worth stretching our tolerance of level of inconsistency.

jxltom:
Current API only suport setting default shipping/billing address with new created addresses. With ```shipping_address_id``` or ```billing_address_id```, the API could update these default address with existed addresses.

akjanik:
Ok, I understand. But in my opinion, we should then create a separate mutation, that takes `checkout_id`, `address_id` and enum `address_type` and results in assigning an existing `Address` to given `Checkout` as `BILLING` or `SHIPPING`.
It seems more readable for me.

jxltom:
So we keep ```checkoutBillingAddressUpdate``` and ```checkoutShippingAddressUpdate```, and create new ```checkoutAddressUpdateById``` mutation? It feels more readable in code, but feels a little bit redundent in API?

akjanik:
Yes, it's kinda tradeoff as you mentioned. I once read that GraphQL is loved by frontend devs but can be sometimes tricky to implement, it may be one of such cases. It never gonna be perfect, but we want to avoid introducing something that causes confusion.

jxltom:
OK, I will do some adjustments.

maarcingebala:
We should first discuss and agree on the shape of the API before we start implementing anything, but personally, I don't have any idea yet how should it look like.

jxltom:
OK, I will postphone it until we have an agreement on the best shape of API.

maarcingebala:
I'm looking at two other e-commerce APIs and none of them allows passing address IDs in the checkout; instead, they require you to unpack the address and pass all the fields separately:
- https://docs.moltin.com/api/carts-and-checkout/checkout
- https://help.shopify.com/en/api/custom-storefronts/storefront-api/reference/mutation/checkoutshippingaddressupdatev2

Since our API follows the same pattern, I'd recommend sticking to it, unless we find valid reasons to change that.

jxltom:
If so, that means the frontend needs to store all information (such as each field name and value) of users' addresses if we want to update shipping/billing address from existed addresses. Othewise, the frontend couldn't unpack the address to correct form. One more thing is that with such kind of design, if the backend changes or adds more address fields, the frontend is very likey to fail to unpack the address. This maybe unfriendly to frontend implementation.

May it is better for the frontend to only store the id of address and use these ids to perform operations such as updating default address?

maarcingebala:
I don't think it's a problem for the frontend app to store the addresses. E.g. in the PWA Storefront all the addresses are fetched from the API when users log in and then they're stored in the Apollo cache. Also, the address model is very unlikely to change and if that was supposed to happen, frontend code would have to be adjusted anyway.

jxltom:
Emm, from PWA point of view, saving all information of addresses sounds reasonable.

maarcingebala:
Alright, I'm closing this PR then. Thank you for your effort anyway and sorry that it took us so long to sort it out; there are so many changes going on in Saleor and sometimes it's hard to keep up with them 😉

jxltom:
No problem man. :handshake:

PR #3814, opened 2019-03-07
Bump django-graphql-jwt

michaljelonek:
Updates to new version and fixes one issue that it breaks 🙈.

### Pull Request Checklist


Pipfile.lock
```diff
@@ -282,11 +282,10 @@
         },
         "django-graphql-jwt": {
             "hashes": [
-                "sha256:d18085c6bb7bf20701a0cc6ee556436d9558c929aabdcecace52252a43408ab7",
-                "sha256:da4c4c4a7c7c75f6c7ed117589076e430614248f155cfe343561fe6bcd10e574"
+                "sha256:6dcf9dd4990098e289d76ab4220edbf436a2e26dc6af1d4a526ac699ad93bf14"
```
jxltom: Did you update this file manually? It should be updated by ```pipenv lock``` and using ```pipenv lock -r > requirements.txt``` and ```pipenv lock -r -d > requirements_dev.txt``` to update requirements.txt and requirements_dev.txt.

jxltom: One more thing is we should pin ```django-recaptcha = "<=1.5.0"``` if we need to update the Pipfile.lock since Saleor doesn't support recaptcha 2.x.

michaljelonek: Partly, I did run `pipenv lock`, but just removed all the other changes that I did not want to make at this stage. `pipenv lock` updated around 20-30 packages, some are even major version bumps and there's no way to speciify that you want to update just one package + its dependencies, so I gave up on that idea and just removed everything I did not care about.

jxltom: That's one issue with pipenv that we cannot update specific package.

But I think it is fine to also update other packages as well since they meet the specification in Pipfile. If we really need to pin other packages, we should pin their version in Pipfile, as I suggested for ```django-recaptcha```.

Update Pipfile.lock manually will make the Pipfile.lock outdated or broken. You can try ```pipenv install --dev```, you will find pipenv will think the Pipfile.lock is outdated and re-lock again anyway.

michaljelonek: That's one of many issues of pipenv. 

I don't want to do this in this PR as it might introduce breaking changes that are not covered by tests. Sure, `Pipfile` has versions, but I do believe that it should be looked over (best in a different issue) and correctly use identifiers to specify versions, instead of using `>=` everywhere.

`Pipfile.lock` was not edited manually in a sense that I added or removed anything. I simply reverted (or checked out if you prefer in git terms) all the other changes that `pipenv update` did, leaving only the change in this PR. `pipenv install --dev` works fine for me and it doesn't raise any issues about `Pipfile.lock` being outdated and doesn't re-lock, neither does `pipenv sync`.

However if you try to install a new package using `pipenv install <package>` then indeed it will try to re-lock all the packages, which is another issue of pipenv.

Perhaps it is worth to consider an alternative (poetry) and see how it fares.

maarcingebala:
Let's merge it as it is. If Pipfile breaks we'll update other packages as well. For now, we only wanted to update one of them.

Pipfile.lock
jxltom: It doesn't have version identifier like npm which is not good indeed. But maybe we can lock them by ```>=1.1, <1.2``` in Pipfile.

Issue #3830, opened 2019-03-13
Version mismatch between celery and redis-py


Rohja:
### What I'm trying to achieve
Deploy Saleor using the `docker-compose.yml` file.

### Steps to reproduce the problem
You need to have git, docker and docker-compose installed on your machine
1. run `git clone https://github.com/mirumee/saleor`
2. **optional:** run `git checkout 2.4.0`
3. run `docker-compose up`

### What I expected to happen
After the build of the docker image, a viable environment is supposed to run, instead, the celery instance is crashing with the following message:

```celery_1  | [2019-03-13 09:10:01,799: CRITICAL/MainProcess] Unrecoverable error: VersionMismatch('Redis transport requires redis-py versions 3.2.0 or later. You have 2.10.6',)
celery_1  | Traceback (most recent call last):
celery_1  |   File "/usr/local/lib/python3.6/site-packages/kombu/transport/virtual/base.py", line 921, in create_channel
celery_1  |     return self._avail_channels.pop()
celery_1  | IndexError: pop from empty list
celery_1  | 
celery_1  | During handling of the above exception, another exception occurred:
celery_1  | 
celery_1  | Traceback (most recent call last):
celery_1  |   File "/usr/local/lib/python3.6/site-packages/celery/worker/worker.py", line 205, in start
celery_1  |     self.blueprint.start(self)
celery_1  |   File "/usr/local/lib/python3.6/site-packages/celery/bootsteps.py", line 119, in start
celery_1  |     step.start(parent)
celery_1  |   File "/usr/local/lib/python3.6/site-packages/celery/bootsteps.py", line 369, in start
celery_1  |     return self.obj.start()
celery_1  |   File "/usr/local/lib/python3.6/site-packages/celery/worker/consumer/consumer.py", line 317, in start
celery_1  |     blueprint.start(self)
celery_1  |   File "/usr/local/lib/python3.6/site-packages/celery/bootsteps.py", line 119, in start
celery_1  |     step.start(parent)
celery_1  |   File "/usr/local/lib/python3.6/site-packages/celery/worker/consumer/connection.py", line 23, in start
celery_1  |     c.connection = c.connect()
celery_1  |   File "/usr/local/lib/python3.6/site-packages/celery/worker/consumer/consumer.py", line 404, in connect
celery_1  |     conn = self.connection_for_read(heartbeat=self.amqheartbeat)
[...]

PR #3831, opened 2019-03-13
Fix version mismatch between celery and redis-py

Rohja:

It fix the issue details in #3830 : Celery can't start because the version of redis-py in Pipenv/Pipenv.lock/requirements.txt is too old.

### Pull Request Checklist

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -58,7 +58,7 @@ social-auth-app-django = ">=2.0.0"
 text-unidecode = "*"
 uwsgi = {version = ">=2.0.0", platform_system = "!= 'Windows'"}
 celery = {version = "*", extras = ["redis"]}
-redis = "<3.0.0"  # this can be safely removed after Celery v4.2.2 is released
+redis = ">=3.2.0"
 weasyprint = ">=0.42.2"
 graphene-django-optimizer = "*"
 braintree = "==3.49.0"
--- a/Pipfile
+++ b/Pipfile
@@ -58,7 +58,8 @@ social-auth-app-django = ">=2.0.0"
 text-unidecode = "*"
 uwsgi = {version = ">=2.0.0", platform_system = "!= 'Windows'"}
 celery = {version = "*", extras = ["redis"]}
-redis = ">=3.2.0"
+redis = "<3.0.0"  # this can be safely removed after Celery v4.2.2 is released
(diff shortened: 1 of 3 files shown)


maarcingebala:
Hey, thanks for opening this PR! Have you tested it locally and can confirm that after these changes Celery works correctly? I'm asking because we're bumping major version of `redis` package from 2.x to 3.x which has some backward incompatibilities.

jxltom:
Which version of celery are you using? I think the problem is Celery instead of redis. As mentioned in https://github.com/mirumee/saleor/pull/3374, redis 3.x is not compatible with celery 4.2.1.

Rohja:
Hi, 
I tested it locally, running `docker-compose up --build` after this change allow celery to run without issue.
I'm running celery 4.2.1 as specified in the original `Pipenv.lock` and `redis:5.0-alpine` from the Docker Hub as specifier in the file `docker-compose.yml`.

jxltom:
As reported in https://github.com/celery/celery/issues/5369, the problem is with celery and kombo. It looks like we should pin kombo < 4.4.0 in Pipfile.

requirements.txt
```diff
@@ -90,7 +90,7 @@ python3-openid==3.1.0 ; python_version >= '3.0'
 pytz==2018.9
 raven==6.9.0
 razorpay==1.2.0
-redis==2.10.6
+redis==3.2.0
```
jxltom: One more thing is that I noticed that botocore and some other packages are also upgraded in Pipfile.lock, could you please update the requirements files by ```pipenv lock -r > requirements.txt``` and ```pipenv lock -r -d > requirements_dev.txt``` as well?

Pipfile
```diff
@@ -58,7 +58,7 @@ social-auth-app-django = ">=2.0.0"
 text-unidecode = "*"
 uwsgi = {version = ">=2.0.0", platform_system = "!= 'Windows'"}
 celery = {version = "*", extras = ["redis"]}
-redis = "<3.0.0"  # this can be safely removed after Celery v4.2.2 is released
+redis = ">=3.2.0"
```
jxltom: I would recommend revert this but pin ```kombu = "<=4.3.0"``` to fix this issue.

jxltom:
LGTM 👍

maarcingebala:
@Rohja Thanks for contributing to Saleor!

PR #3838, opened 2019-03-14
Add bulk delete mutations

### LINKED ISSUE #3799: Add support for bulk deleting in API
We're implementing bulk actions in Dashboard 2.0 (#3537) and we need to support them in API as well. I'd look at our Figma mockups in order to determine which models need that, but I'd guess that all of them.

A first action that we'll add is bulk delete. The easiest and quickest way (1) to support that is to modify our existing "delete mutations" to receive a list of IDs instead of a single ID e.g.:
`categoryDelete(id: ID!)` -> `categoryDelete(ids: [ID!]!)`

Another way (2) is to introduce dedicated bulk delete mutations:
`categoryBulkDelete(ids: [ID!]!)`.

Originally I wanted to choose the first option but now I'm wondering if it wasn't more elegant to have dedicated mutations for that?

michaljelonek:
Resolves #3799.

Adds bulk delete where delete was already present.

#### Example
image

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/bulk_mutations.py
+++ b/saleor/graphql/account/bulk_mutations.py
@@ -1,0 +1,58 @@
+import graphene
+
+from ...account import models
+from ..core.mutations import ModelBulkDeleteMutation
+
+
+class UserBulkDelete(ModelBulkDeleteMutation):
+    class Arguments:
+        ids = graphene.List(
+            graphene.ID,
+            required=True,
+            description='List of sale IDs to delete.')
+
+    class Meta:
+        abstract = True
+
+    @classmethod
+    def clean_instance(cls, info, instance, errors):
+        user = info.context.user
+        if instance == user:
+            cls.add_error(
+                errors, 'id', 'You cannot delete your own account.')
+        elif instance.is_superuser:
+            cls.add_error(
+                errors, 'id', 'Only superuser can delete his own account.')
+
+
+class CustomerBulkDelete(UserBulkDelete):
+    class Meta:
+        description = 'Deletes customers.'
+        model = models.User
+
+    @classmethod
+    def user_is_allowed(cls, user, input):
+        return user.has_perm('account.manage_users')
+
+    @classmethod
+    def clean_instance(cls, info, instance, errors):
+        super().clean_instance(info, instance, errors)
+        if instance.is_staff:
+            cls.add_error(errors, 'id', 'Cannot delete a staff account.')
+
+
+class StaffBulkDelete(ModelBulkDeleteMutation):
+    class Meta:
+        description = 'Deletes staff users.'
+        model = models.User
+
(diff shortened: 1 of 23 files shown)


saleor/graphql/account/bulk_mutations.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @classmethod
+    def clean_instance(cls, info, instance, errors):
+        user = info.context.user
+        if instance == user:
+            cls.add_error(
+                errors, 'id', 'You cannot delete your own account.')
+        elif instance.is_superuser:
+            cls.add_error(
+                errors, 'id', 'Only superuser can delete his own account.')
```
jxltom: Should it be ```Superuser's account cannot be delelted``` since only ```instance.is_superuser``` is checked?

saleor/graphql/core/mutations.py
```diff
@@ -355,6 +355,70 @@ def mutate(cls, root, info, **data):
...
+    def mutate(cls, root, info, ids):
+        """Perform a mutation that deletes a list of model instances."""
+        if not cls.user_is_allowed(info.context.user, ids):
+            raise PermissionDenied()
+
+        model_type = registry.get_type_for_model(cls._meta.model)
+        count, all_errors = 0, []
+        for id in ids:
+            errors = []
+            instance = cls.get_node_or_error(
```
maarcingebala: We have the `get_nodes_or_error` function as well, would it make sense to use it here?

saleor/graphql/account/bulk_mutations.py
maarcingebala: We could also say "Cannot delete this account" or something like that to not reveal that this ID represents a superuser account, does it make sense? It would actually apply also to deleting single users.

saleor/graphql/order/bulk_mutations/draft_orders.py
```diff
@@ -0,0 +1,20 @@
...
+class DraftOrderBulkDelete(ModelBulkDeleteMutation):
+    class Arguments:
+        ids = graphene.List(
+            graphene.ID,
+            required=True,
+            description='List of draft order IDs to delete.')
+
+    class Meta:
+        description = 'Deletes draft orders.'
+        model = models.Order
```
maarcingebala: Since this takes just the list of IDs it would only allow deleting regular orders, is that right? If so, we should add some checks to prevent that. Only draft orders can be deleted in Saleor, while regular orders can be at most canceled.

saleor/graphql/account/bulk_mutations.py
michaljelonek: I admit this has been copied from the normal (non bulk) delete. I planned to extract that validation to a separate function, so it can be used in both without copying like this.

Issue #3846, opened 2019-03-18
Migrate to Black


maarcingebala:
Currently, in Saleor we use Yapf code formatter for Python. The problem is that we've never actually formatted the whole codebase with it which results in mixed code styles across the project. Another problem is that we don't have any automatic formatting enabled so we cannot enforce a unified style in PRs from the community.

We could stick to Yapf, but in the meantime, a new tool appeared - [Black](https://github.com/ambv/black) - and it's becoming adopted by many other Python projects. A few teams also use it at Mirumee and they recommend it. After internal discussion, we decided to migrate to Black in Saleor as well. This change will affect every open PR and will probably cause a lot of merge conflicts for those how regularly update Saleor, but this step is necessary to improve the readability and consistency of Saleor's codebase.

What needs to be done:
- add `black` to dev requirements
- apply the new formatting
- configure automatic formatting - I'd like to hear some feedback on that as I don't know what might be the best way to achieve that. One solution that I found is [pre-commit](https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/) and it looks quite promising.

jxltom:
Maybe we could also consider adding isort/db-migration/graphql-schema/ts-types checks in pre-commit.

We may need following part in Pipfile since black still doesn't have a stable release.

```
[pipenv]
allow_prereleases = true
```

maarcingebala:
For now, I would rather only one tool and if it works for us, reconsider adding automating more in the future.

krzysztofwolski:
Also, there should be proper tests launched on CI.

Re isort - I like this tool, but sometimes I've got problems when using with black. And `pre-commit` can't work with it - theres issue with python environment separation.

krzysztofwolski:
@jxltom thanks for code review! 💪  
As you noticed, black has a nonstandard line length of 88 chars. Personally, I like the reasoning and approach of its authors: https://github.com/ambv/black#line-length
88 Chars give extra breathing room when using types and clear (I mean long 😆) variable names. And still can open 2 files on one screen without horizontal scrolling.

Do we embrace 88?

krzysztofwolski:
Update: We are waiting for version 1.0 - ATM black is preparing new version which will for sure affect our codebase. To Avoid making massive formatting twice, we will wait. 

Related tweet: https://twitter.com/llanga/status/1106247623802060802

NyanKiyoshi:
[...]

PR #3852, opened 2019-03-19
Use black code formatter

krzysztofwolski:
I want to merge this change because... ref #3846

### Screenshots

```
All done! 💥 💔 💥
644 files would be reformatted, 89 files would be left unchanged.
```

### Todo
- [x] Rebase on latest master
- [x] Reapply black
- [x] Add black CI
- [x] Add pre-commit to the development dependencies
- [x] Fix conflicts between isort and black
- [x] Fix issues with isort unable to pass in pre-commit
- [x] Update docs
  - [x] We use black
  - [x] We use a 88 character limit

### Pull Request Checklist


pyproject.toml
```diff
@@ -0,0 +1,26 @@
+[tool.black]
+target_version = ['py36', 'py37', 'py38']
+include = '\.pyi?$'
+exclude = '''
```
krzysztofwolski: I've extended default exclude list for faster file discovery. We don't need to format cache dirs.

saleor/checkout/__init__.py
```diff
@@ -6,15 +6,10 @@
 
 
 class AddressType:
-    BILLING = 'billing'
-    SHIPPING = 'shipping'
+    BILLING = "billing"
```
jxltom: It looks like black supports ignoring format strings as in https://github.com/ambv/black/issues/118#issuecomment-393298377. Since Saleor is using single quotes but black is using double qutoes, should we apply that?

Pipfile
```diff
@@ -56,8 +56,8 @@ raven = "==6.9.0"
 requests = ">=1.2.0"
 social-auth-app-django = ">=2.0.0"
 text-unidecode = "*"
-uwsgi = {version = ">=2.0.0", platform_system = "!= 'Windows'"}
-celery = {version = "*", extras = ["redis"]}
+uwsgi = {version = ">=2.0.0",platform_system = "!= 'Windows'"}
```
jxltom: It looks like the official docs is using spaces as in https://pipenv.readthedocs.io/en/latest/advanced/#specifying-basically-anything, could we add the spaces back? It has cleaner styles.

Pipfile
```diff
@@ -87,3 +87,4 @@ yapf = "*"
 django-extensions = "*"
 pylint-plugin-utils = "*"
 transifex-client = "*"
+black = "==19.3b0"
```
jxltom: Do you mind updating the requirements files as well by ```pipenv lock -r > requirements.txt``` and ```pipenv lock -r -d -> requirements_dev.txt```?

saleor/account/forms.py
```diff
@@ -40,66 +39,66 @@ def get_address_form(
         address_form = address_form_class(data, instance=instance, **kwargs)
     else:
         initial_address = (
-            initial if not preview
-            else data.dict() if data is not None else data)
+            initial if not preview else data.dict() if data is not None else data
```
jxltom: Black supports pass max line length by ```--line-length``` as mentioned in https://github.com/ambv/black#line-length. Saleor is using flake8's default 79, should we pass this argument as 79 as well?

tests/api/conftest.py
```diff
@@ -38,23 +38,28 @@ def post(self, data=None, **kwargs):
...
 
     def post_graphql(
-            self, query, variables=None, permissions=None,
-            check_no_permissions=True, **kwargs):
+        self,
+        query,
+        variables=None,
+        permissions=None,
+        check_no_permissions=True,
+        **kwargs,
```
jxltom: The ```**kwargs,)``` is not supported in python35 and is fixed in https://bugs.python.org/issue9232 for python36.

I'm not sure if this is a bug with black since ```target_version = ['py35', 'py36', 'py37', 'py38']``` is already set. I think black should take py35 as the minimal?

Pipfile
krzysztofwolski: Change was done by pipenv, did not modified manually. (pipenv, version 2018.11.26)

saleor/checkout/__init__.py
krzysztofwolski: Had my doubts in the last project if I should use `--skip-string-normalization`. Wanted to have as 'standardized' config as possible to avoid discussions about what to choose.
Finally used default black setting (double quotes) and that turned out ok - I personally use single quotes but my editor change that without my attention. 
@maarcingebala what are your thoughts?

maarcingebala: I'd go with the default Black settings and use double quotes.

saleor/account/forms.py
maarcingebala: I'm also fine with using Black's default settings here.

tests/api/conftest.py
krzysztofwolski: Right, that's not good. According to docs formatted code should be valid with all provided target versions.

Pipfile
jxltom: Oh never mind, it is bug with pipenv and is fixed in master. It can be fixed manually by removing spaces though. I think it's fine since it's so minor.

tests/api/conftest.py
maarcingebala: Yep, we have to make sure that the code outputted from Black works under 3.5. I don't know what might be the issue though.

maarcingebala:
With this PR we're adding `.toml` file which is yet another configuration Python file in our repo. Since `.toml` files are supposed to become a standard in Python, it would be nice to migrate to `toml` entirely. This is obviously beyond the scope of this PR but it's something to consider. I don't know if `pipenv` supports `toml`, maybe having only one configuration file would be only possible if we migrated to `Poetry`.

jxltom:
Pipfile is actually toml format, but pipenv iteself doesn't support pyproject.toml yet.

tests/api/conftest.py
krzysztofwolski: Its a bug https://github.com/ambv/black/issues/759 :/

krzysztofwolski:
Depending on https://github.com/ambv/black/issues/759

jxltom:
Plus, https://github.com/ambv/black/pull/763 should fix this issue.

pyproject.toml
NyanKiyoshi: Is it really needed to include `.pyi` stub files?

maarcingebala:
We have to put this PR on hold until the next version of Black is released. This is supposed to happen in April (no exact date yet).

jxltom:
As https://github.com/ambv/black/pull/763 is merged, and Pipenv supports pin any package via git commit, if we don't want to wait for black's release, ```black = {git = "https://github.com/ambv/black@cea13f4", editable = true}``` can be used in Pipfile for patched black version.

maarcingebala:
@jxltom It would be great to migrate to Black as soon as possible. What's your opinion on relying on an unreleased version of a library @patrys @krzysztofwolski ?

patrys:
I'd rather wait for it to be released as I fear people would still use released versions and we'd end up changing the formats back and forth.

jxltom:
It won't be a problem if users install packages via our ```requirements_dev.txt```by ```pip install -r requirements_dev.txt``` or ```Pipfile.lock``` by ```pipenv install --dev``` since Pipenv has already locked the version of black.

Besides, I waited for a released version of black for half a year via https://github.com/ambv/black/issues/517, but there still has no release. Hope it can be released in early April.

krzysztofwolski:
I would rather not use an unreleased version of the software which calls itself `This is a beta product` ;)
I agree with @patrys

krzysztofwolski:
re migration to toml: our tools slowly migrate to "new" format https://gitlab.com/pycqa/flake8/issues/428 and toml libs seem to be an issue. Shame it's not JSON.

maarcingebala:
I just realized that in PR #4028 we're removing support for Python 3.5 which means we'll be no longer blocked by the issue in Black 🎉 .

krzysztofwolski:
@maarcingebala this week I'll update PR 🎉

NyanKiyoshi:
I updated the docs to mention pre-commit & black and added a black badge to the readme.

docs/contributing/coding-style.rst
```diff
@@ -1,43 +1,55 @@
 Coding Style
 ============
 
+Saleor uses the `pre-commit <https://pre-commit.com/#install>`_ tool to check and
```
maarcingebala: When users read this section they'll first read how to use `pre-commit` but we don't say how to install it so I'm worried we'll be getting questions like "pre-commit command doesn't exist, how do I fix it".

At the end of this section, we say:

``python -m pip install -r requirements_dev.txt`` can install these tools. For Pipenv users, they can also be installed by ``pipenv install --dev``.

What about putting it at the beginning like so:

Saleor uses various tools to maintain a common coding style and help with development. To install all the development tools do:

``python -m pip install -r requirements_dev.txt`` 

If you're using `Pipenv`: 

``pipenv install --dev``.

And then we would go with "Saleor uses pre-commit tool to check [...]".

docs/contributing/coding-style.rst
```diff
@@ -1,43 +1,55 @@
...
+automatically fix any formatting issue before creating a git commit.
+
+You can easily install it after having setup saleor by running the below instruction.
```
maarcingebala: It's a good practice to avoid words like "easily" in docs like this, as they're highly subjective (though I know it's easy ;)). I would be more specific here: "Run the following command to install pre-commit into your git hooks and have it run on every commit:"

It would be nice to also say below something like: "See ".pre-commit-config.yaml" for pre-commit configuration.

PR #3861, opened 2019-03-21
Resolve global orders query only for staff user

pawelzar:
Resolves https://github.com/mirumee/saleor/issues/3833

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/resolvers.py
+++ b/saleor/graphql/order/resolvers.py
@@ -30,11 +30,7 @@ def filter_orders(qs, info, created, status, query):
 
 
 def resolve_orders(info, created, status, query):
-    user = info.context.user
-    if user.has_perm('order.manage_orders'):
-        qs = models.Order.objects.confirmed()
-    else:
-        qs = user.orders.confirmed()
+    qs = models.Order.objects.confirmed()
     return filter_orders(qs, info, created, status, query)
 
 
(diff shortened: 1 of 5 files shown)


saleor/graphql/order/resolvers.py
```diff
@@ -31,10 +31,11 @@ def filter_orders(qs, info, created, status, query):
 
 def resolve_orders(info, created, status, query):
     user = info.context.user
-    if user.has_perm('order.manage_orders'):
-        qs = models.Order.objects.confirmed()
-    else:
-        qs = user.orders.confirmed()
+
+    if not user.has_perm('order.manage_orders'):
```
maarcingebala: I think we could as well wrap the order resolver with `@permission_required('order.manage_orders')` just as we do for any other query that requires particular permission. Usually, we do it in the `schema.py` file, in this case, it would have to be changed in the `OrderQueries` class.

tests/api/test_order.py
```diff
@@ -174,6 +174,23 @@ def test_order_query(
     assert expected_method.type.upper() == method['type']
 
 
+def test_order_query_customer(api_client):
```
jxltom: Will ```test_order_query_by_customer``` be more accurate?

maarcingebala: That's not a big deal, the current name is also fine.

maarcingebala:
Congrats on your first contribution to Saleor @pawelzar!

Issue #2098, opened 2018-04-18
Fulfillment of digital products


maarcingebala:
Saleor should provide ways to fulfill digital products and send the digital content to the customer. Ideas of how we could implement it:
- Allow defining the digital content at the variant level. It would become available to customers once the order is fulfilled.
- Send an email with a generated, unique link to access the digital content.
- "Automatic fulfillment" - if enabled, order lines with digital products should be automatically fulfilled, without interaction from the staff. Question - should this be a global setting or product-specific?

saqibkhan2523:
I have downloaded saleor for digital file downloads. But I find that there is no template for that. If anyone here have any kind of tutorials or guide for saleor please refer it here.

maarcingebala:
I cannot give yet any estimates on where we'll have it, but I'll put it as one of the priorities on our roadmap. It's an essential feature that a lot of users is asking about.

blitzd:
In addition to this (or possibly as a separate request), the ability to have products with no deliverable (aside from an email confirmation) like memberships and donations would attract a lot of non-profits. The donation one is often tricky due to the variable nature of the price, and possible maximums for political entities. Memberships can get complex too with optional family members being included at different price points.

creveco:
It would be really nice to have a unique serial number to sell tickets to events. 
This is a like product keys proposed  by ByteX7, but  with a unique auto-generated key.

korycins:
Hi Guys

We discussed it with @maarcingebala.
At the beginning, we would like to add base logic and model that we will develop in future
We were thinking about the two DB models.
DigitalContent which will store the digital content for each product variant.
DigitalContentUrl will be responsible for URLs pointing to digital content. The second model will give a possibility to manage URLs shared with customers, limit number of downloads, set expiration time for URL, create a unique link for each customer.

maarcingebala:
Resolved by #3868

R5n3h:
First of all, thank you guys for a great product! (Saleor)
I personally actually waited for this PR to be merged.
I have just a small question, maybe I'm missing something here...

The field 'is_digital' is missing in dashboard.product.forms.ProductTypeForm?
I can see any way to enable this product type from the dashboard...

Thanks again,
Ron

maarcingebala:
[...]

PR #3868, opened 2019-03-25
2098/support for digital product

korycins:
I want to merge this change because... #2098 

### Pull Request Checklist


saleor/graphql/product/mutations/products.py
```diff
@@ -846,3 +851,161 @@ def mutate(cls, root, info, image_id, variant_id):
                     variant_image.delete()
         return VariantImageUnassign(
             product_variant=variant, image=image, errors=errors)
+
+class ProductVariantDigitalInput(graphene.InputObjectType):
```
maarcingebala: I think we don't have to reveal in names of this type that it is related to variants underneath. The API would look cleaner if we had simply `DigitalContentInput`.

saleor/graphql/product/mutations/products.py
```diff
@@ -846,3 +851,161 @@ def mutate(cls, root, info, image_id, variant_id):
...
+    )
+
+
+class ProductVariantDigitalUploadInput(ProductVariantDigitalInput):
+    content_file = Upload(
+        required=True,
+        description='Represents an file in a multipart request.')
+
+
+class ProductVariantDigitalUpload(BaseMutation):
```
maarcingebala: And here I would use the name `DigitalContentCreate` - should be quite simple and easy to understand what it does just by looking at the schema.

saleor/graphql/product/mutations/products.py
```diff
@@ -846,3 +851,161 @@ def mutate(cls, root, info, image_id, variant_id):
...
+        https://github.com/jaydenseric/graphql-multipart-request-spec''')
+
+    @classmethod
+    @permission_required('product.manage_products')
+    def mutate(cls, root, info, id, input):
+        errors = []
+        variant = cls.get_node_or_error(
+            info, id, errors, 'id', only_type=ProductVariant)
+        content_data = info.context.FILES.get(input['content_file'])
+        if hasattr(variant, 'digital_content'):
```
maarcingebala: As we spoke - in this case, we could simply override existing content instead of raising an error.

saleor/graphql/product/mutations/products.py
```diff
@@ -846,3 +851,161 @@ def mutate(cls, root, info, image_id, variant_id):
...
+            'Remove digital content assigned to given variant')
+
+    @classmethod
+    @permission_required('product.manage_products')
+    def mutate(cls, root, info, id):
+        errors = []
+        variant = cls.get_node_or_error(
+            info, id, errors, 'id', only_type=ProductVariant)
+
+        if not hasattr(variant, 'digital_content'):
```
maarcingebala: I think here as well we can simply do nothing if there is no content. Mutation result will anyway return a variant without any digital content.

saleor/graphql/product/mutations/products.py
```diff
@@ -846,3 +851,161 @@ def mutate(cls, root, info, image_id, variant_id):
...
+            if automatic_fulfillment:
+                digital_content.automatic_fulfillment = automatic_fulfillment
+
+            variant.digital_content = digital_content
+            variant.digital_content.save()
+        return ProductVariantDigitalUpload(
+            content=digital_content, errors=errors)
+
+
+class ProductVariantDigitalDelete(BaseMutation):
```
maarcingebala: We have the `ModelDeleteMutation` class, why can't we use it here?

saleor/graphql/product/mutations/products.py
maarcingebala: Have you tried using `ModelMutation` here?

saleor/graphql/product/mutations/products.py
maarcingebala: Also, in GraphQL types, we have the `DigitalContent` type and mutations should also use that naming pattern for better consistency.

saleor/graphql/product/schema.py
```diff
@@ -123,6 +130,14 @@ def resolve_collection(self, info, id):
     def resolve_collections(self, info, query=None, **kwargs):
         return resolve_collections(info, query)
 
+    @permission_required(['product.manage_products',])
```
maarcingebala: If there is single permission, it can be passed as a string: `@permission_required('product.manage_products')`.

saleor/graphql/product/types.py
```diff
@@ -202,6 +228,12 @@ class Meta:
         interfaces = [relay.Node]
         model = models.ProductVariant
 
+    @permission_required('product.manage_products')
+    def resolve_digital_content(self, info):
+        if hasattr(self, 'digital_content'):
```
maarcingebala: You would achieve the same result with: `return getattr(self, 'digital_content', None)`.

saleor/graphql/product/mutations/products.py
maarcingebala: What do you think about moving all digital content related mutations to a separate file?

saleor/product/models.py
```diff
@@ -304,6 +309,58 @@ def __str__(self):
...
+                 update_fields=None):
+        if not self.token:
+            self.token = str(uuid4()).replace('-', '')
+        super().save(force_insert, force_update, using, update_fields)
+
+    def get_absolute_url(self) -> str:
+        url = reverse('product:digital-product', kwargs={'token': str(self.token)})
+        return build_absolute_uri(url)
+
+    def ready_to_share(self) -> bool:
```
maarcingebala: I'd suggest renaming this function to the short and simple `is_valid`, what do you think?

saleor/product/utils/__init__.py
```diff
@@ -12,6 +13,21 @@
 from .availability import products_with_availability
 
 
+def get_default_automatic_fulfillment() -> bool:
```
maarcingebala: Would it make sense to have a single function instead of those three, that would return a dictionary of all digital download defaults? You would have to fetch `site` only once. Actually, one thing to consider here is that we have a middleware that stores the current site in `request.site` so that you don't have to fetch it every time, maybe we could take advantage of that somehow?

saleor/product/views.py
```diff
@@ -84,6 +87,19 @@ def product_details(request, slug, product_id, form=None):
     return TemplateResponse(request, 'product/details.html', ctx)
 
 
+def digital_product(request, token: str) -> Union[FileResponse, HttpResponseNotFound]:
```
maarcingebala: This will work in the old storefront, but we'll need something similar in API/new storefront as well. We have to discuss what's the best way to achieve that.

saleor/product/models.py
```diff
@@ -304,6 +309,58 @@ def __str__(self):
...
+        url_valid_days = self.content.url_valid_days
+        if url_valid_days is not None:
+            valid_days = datetime.timedelta(days=url_valid_days)
+            valid_until = self.created + valid_days
+            if now() > valid_until:
+                return False
+
+        max_downloads = self.content.max_downloads
+        if max_downloads is not None and max_downloads <= self.download_num:
```
maarcingebala: Instead of explicitly returning booleans we could have:
```
return True if max_downloads is None else self.download_num < max_downloads
```

saleor/graphql/product/mutations/products.py
korycins: I checked `ModelMutation` previously but I wasn't sure what I can achieve by using it. As you see I have to overwrite `clean_input` method as well as `mutate` method. AFAIK `ModelMutation` will not give me anything

saleor/graphql/product/mutations/digital_contents.py
```diff
@@ -0,0 +1,222 @@
...
+        if use_default_settings:
+            return input
+
+        required_fields = [
+            'max_downloads', 'url_valid_days', 'automatic_fulfillment']
+
+        if not all(field in input for field in required_fields):
+            msg = ('Use default settings is disabled. Provide all '
+                   'configuration fields')
+            missing_field = set(required_fields).difference(set(input))
```
akjanik: nit: `missing_fields`

saleor/order/models.py
```diff
@@ -327,6 +327,14 @@ def get_total(self):
     def quantity_unfulfilled(self):
         return self.quantity - self.quantity_fulfilled
 
+    @property
+    def is_digital(self) -> bool:
```
akjanik: Not sure if we should combine two checks (`is_digital` and `is_empty`) into one, but if it's reasonable, then ok.

saleor/order/utils.py
```diff
@@ -13,10 +13,68 @@
...
+        return True
+    if content.automatic_fulfillment:
+        return True
+    return False
+
+
+def order_needs_automatic_fullfilment(order: Order) -> bool:
+    """Check if order has digital products which should be automatically
+    fulfilled"""
+    digital_lines = [line for line in order if line.is_digital]
```
akjanik: Did you mean `(...) in lines (...)`?

akjanik: I meant, `order.lines`.

korycins: Order has overwriten __iter__ method. It returns `self.lines.all()`

saleor/order/utils.py
```diff
@@ -13,10 +13,68 @@
...
+    if not digital_lines:
+        return False
+    for line in digital_lines:
```
akjanik: I am not sure if I like this loop. I would consider to make a `property` or `manager` for an `OrderLine` model and then one could use it as a `filter` condition.
But it seems to work, so if there are no big benefits from change, it's fine.

akjanik:
I do not see the way we can update number of downloads (`download_num`) via graphql API.

saleor/order/utils.py
korycins: I don't want to put it inside the `models` file, keeping in mind how fast `models` file can grow. 
We can ask Marcin also. Is it fine for you?

saleor/order/models.py
korycins: What do you propose? 
OrderLine is digital only when a variant is digital and variant has already attached a file.

akjanik: As I said, if there is a valid reason, it may stay as it is. Looks like there is, so no changes I guess!

saleor/order/utils.py
akjanik: Yes. Maybe manager could do the job then.

PR #3891, opened 2019-04-01
Refactor error handling in mutations

maarcingebala:
This PR is an attempt to simplify the mutations code by introducing the following changes:
- added `perform_mutation` function which is fired by `mutate` - this allows to catch validation errors that occur at any time during `perform_mutation` execution:
- removed `if errors` checks 
- removed accumulating errors in `errors` lists which were passed between various functions during a mutation execution

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -621,8 +621,6 @@ class CustomerSetDefaultAddress(BaseMutation):
             address_type = AddressType.BILLING
         elif type == AddressTypeEnum.SHIPPING.value:
             address_type = AddressType.SHIPPING
-        else:
-            raise ValueError('Unknown value of AddressTypeEnum: %s' % type)
 
         utils.change_user_default_address(user, address, address_type)
         return cls(user=user)
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -661,7 +661,7 @@ class UserAvatarUpdate(BaseMutation):
         user.save()
         create_user_avatar_thumbnails.delay(user_id=user.pk)
 
-        return UserAvatarUpdate(user=user, errors=errors)
+        return UserAvatarUpdate(user=user)
 
 
 class UserAvatarDelete(BaseMutation):
@@ -674,7 +674,6 @@ class UserAvatarDelete(BaseMutation):
     @staff_member_required
     def perform_mutation(cls, root, info):
         user = info.context.user
-        errors = []
         user.avatar.delete_sized_images()
         user.avatar.delete()
         return UserAvatarDelete(user=user)
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -522,7 +522,8 @@ class AddressSetDefault(BaseMutation):
         address_id = graphene.ID(
             required=True, description='ID of the address.')
         user_id = graphene.ID(
-            required=True, description='ID of the user to change the address for.')
+            required=True,
+            description='ID of the user to change the address for.')
         type = AddressTypeEnum(
             required=True, description='The type of address.')
 
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -269,10 +269,7 @@ class StaffUpdate(StaffCreate):
         model = models.User
 
     @classmethod
-    def clean_is_active(cls, is_active, instance, user, errors):
-        if is_active is None:
-            return errors
-
+    def clean_is_active(cls, is_active, instance, user):
         if not is_active:
             if user == instance:
                 raise ValidationError({
@@ -280,14 +277,13 @@ class StaffUpdate(StaffCreate):
             elif instance.is_superuser:
                 raise ValidationError({
                     'is_active': 'Cannot deactivate superuser\'s account.'})
-        return errors
 
     @classmethod
     def clean_input(cls, info, instance, input, errors):
         cleaned_input = super().clean_input(info, instance, input, errors)
-        cls.clean_is_active(
-            cleaned_input.get('is_active'), instance, info.context.user,
-            errors)
+        is_active = cleaned_input.get('is_active')
+        if is_active is not None:
+            cls.clean_is_active(is_active, instance, info.context.user)
         return cleaned_input
 
 
@@ -376,6 +372,7 @@ class PasswordReset(BaseMutation):
                 'email': 'User with this email doesn\'t exist'})
         site = info.context.site
         send_user_password_reset_email(user, site)
+        return PasswordReset()
 
 
 class CustomerPasswordResetInput(graphene.InputObjectType):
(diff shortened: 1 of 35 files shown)


saleor/graphql/account/mutations.py
```diff
@@ -649,7 +625,7 @@ def mutate(cls, root, info, id, type):
             raise ValueError('Unknown value of AddressTypeEnum: %s' % type)
```
akjanik: It's not changed in this PR, but isn't it validated by GraphQL itself? When you try to type query/mutation with enum values not defined in schema (in this case from here https://github.com/mirumee/saleor/blob/23208ede820d720645fe6bf2e0a5e4cda0b6f794/saleor/checkout/__init__.py) graphql should disallow process such request.

saleor/graphql/core/mutations.py
```diff
@@ -383,38 +388,60 @@ def user_is_allowed(cls, user, ids):
         """
         return True
 
+    @classmethod
+    def mutate(cls, root, info, **data):
+        if not cls.user_is_allowed(info.context.user, data):
+            raise PermissionDenied()
+
+        count, errors = cls.perform_mutation(root, info, **data)
```
akjanik: I (maybe overreacting) see some danger here. In base case this method return one argument (`None`) and it's not overwritten in this class to return more, so if someone forgets implementing this, may get confused by `TypeError: cannot unpack non-iterable NoneType object`

saleor/graphql/core/mutations.py
```diff
@@ -383,38 +388,60 @@ def user_is_allowed(cls, user, ids):
...
+            # catch individual validation errors to raise them later as
+            # a single error
+            try:
+                cls.clean_instance(info, instance)
+            except ValidationError as e:
+                msg = '. '.join(e.messages)
+                instance_errors.append(msg)
 
             if not instance_errors:
                 instance.delete()
```
akjanik: Is this desired by design? If I have instances list, say `[1,2,3,4,5]`, where `3` raises error, this code still deletes `1` and `2`].
Then on `3`,  it adds one entry to empty list `instance_errors` (which is instantiated to empty for every iteration ¯\_(ツ)_/¯), which is used to create message of initialized `ValidationError` (not sure if it's hold in some dynamic scope to preserve data, need to dig about it) and then `4` and `5` still gets deleted.

akjanik: I think I see how it work, not sure if we should allow to delete instances if not all are validated.

saleor/graphql/menu/mutations.py
```diff
@@ -213,21 +209,24 @@ class Meta:
...
+            menu = cls.get_node_or_error(info, menu, field='menu')
+
+        if navigation_type == NavigationType.MAIN:
+            site_settings.top_menu = menu
+            site_settings.save(update_fields=['top_menu'])
+        elif navigation_type == NavigationType.SECONDARY:
+            site_settings.bottom_menu = menu
+            site_settings.save(update_fields=['bottom_menu'])
+        else:
+            raise AssertionError(
```
akjanik: Why `AssertionError`? And again I see we use enums, so technically this code should never be executed.

saleor/graphql/shop/mutations.py
```diff
@@ -92,16 +90,18 @@ class Meta:
...
 
     @classmethod
-    @permission_required('site.manage_settings')
-    def mutate(cls, root, info):
-        errors = []
+    def user_is_allowed(cls, user, input):
+        return user.has_perm('site.manage_settings')
+
+    @classmethod
+    def perform_mutation(cls, root, info):
```
korycins: It is only visual change but personaly I prefer this way:
```
    @classmethod
    def perform_mutation(cls, root, info):
        if not settings.VATLAYER_ACCESS_KEY:
            raise ValidationError(
                'Could not fetch tax rates. Make sure you have supplied a '
                'valid API Access Key.')
        call_command('get_vat_rates')
        return ShopFetchTaxRates(shop=Shop())
```

saleor/graphql/core/mutations.py
```diff
@@ -64,66 +81,52 @@ def _update_mutation_arguments_and_fields(cls, arguments, fields):
...
             if node is None:
-                message = "Couldn't resolve to a node: %s" % global_id
-                cls.add_error(errors, field, message)
+                raise ValidationError({
+                    field: "Couldn't resolve to a node: %s" % node_id})
         return node
 
     @classmethod
-    def get_nodes_or_error(cls, ids, errors, field, only_type=None):
+    def get_nodes_or_error(cls, ids, field, only_type=None):
```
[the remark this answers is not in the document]
akjanik: I would stick with the name till we find better one - I am out of ideas (`get_nodes_or_raise_validation_error` does not look pleased to type everywhere) and we already have `get_nodes` util.

saleor/graphql/order/mutations/orders.py
```diff
@@ -140,17 +114,19 @@ class Meta:
...
-        errors = []
-        order = cls.get_node_or_error(info, id, errors, 'id', Order)
+    def user_is_allowed(cls, user, input):
+        return user.has_perm('order.manage_orders')
+
+    @classmethod
+    def perform_mutation(cls, root, info, id, input):
+        order = cls.get_node_or_error(info, id, only_type=Order)
 
         if not input['shipping_method']:
```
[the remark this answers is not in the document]
akjanik: I think that can be the case, indeed.

saleor/graphql/core/mutations.py
```diff
@@ -431,18 +411,48 @@ def perform_mutation(cls, root, info, ids):
...
+    @classmethod
+    def mutate(cls, root, info, **data):
+        if not cls.user_is_allowed(info.context.user, data):
+            raise PermissionDenied()
+
+        count, errors = cls.perform_mutation(root, info, **data)
+        if errors:
+            errors = validation_error_to_error_type(errors)
+            return cls(errors=errors, count=count)
+        else:
```
korycins: don't need `else`

korycins: Will it work if you leave only one `return` statement? 
```
        if errors:
             errors = validation_error_to_error_type(errors)
        return cls(errors=errors, count=count)
```
What do you think?

saleor/graphql/checkout/mutations.py
```diff
@@ -372,34 +325,33 @@ class CheckoutShippingAddressUpdate(BaseMutation, I18nMixin):
     class Arguments:
         checkout_id = graphene.ID(description='ID of the Checkout.')
         shipping_address = AddressInput(
+            required=True,
```
korycins: is it expected?

saleor/graphql/core/mutations.py
```diff
@@ -154,6 +155,35 @@ def construct_instance(cls, instance, cleaned_data):
...
+        return True
+
+    @classmethod
+    def mutate(cls, root, info, **data):
+        if not cls.user_is_allowed(info.context.user, data):
+            raise PermissionDenied()
+
+        try:
+            response = cls.perform_mutation(root, info, **data)
+            if response.errors is None:
```
korycins: So... maybe `perform_mutation` should return empty errors as a `[]`?

PR #3665, opened 2019-01-29
Set python version in docker from 3.6 to 3.7

bykof:
I want to merge this change because...

### Pull Request Checklist

1. [ ] Privileged views and APIs are guarded by proper permission checks.
1. [ ] All visible strings are translated with proper context.
1. [ ] All data-formatting is locale-aware (dates, numbers, and so on).
1. [ ] Database queries are optimized and the number of queries is constant.
1. [ ] Database migration files are up to date.
1. [ ] The changes are tested.
1. [ ] The code is documented (docstrings, project documentation).
1. [ ] GraphQL schema and type definitions are up to date.
1. [ ] Changes are mentioned in the changelog.

PR #3898, opened 2019-04-03
chore(heroku): upgrading to python-3.6.8

duanhongyi:

Python has released a security update! Please consider upgrading to python-3.6.8
Learn More: https://devcenter.heroku.com/articles/python-runtimes

### DIFF
--- a/runtime.txt
+++ b/runtime.txt
@@ -1,1 +1,1 @@
-python-3.6.6
+python-3.6.8


maarcingebala:
Thanks for opening this PR! I'll merge it, but from what I see, Heroku supports now Python 3.7.2 and we could migrate straight to that version. Since we have PR #3665 which bumps default Python version in Docker to 3.7, we could do it there.

PR #3942, opened 2019-04-08
Clean up graphql API code

### LINKED ISSUE #3909: Cleanup and maintenance in GraphQL API code
There are a few problems with the code style that we should fix in order to make the codebase cleaner and stop repeating problematic patterns in the future. `pylint` and codeclimate might be helpful to detect most of the problems. Few that I found so far are:

- `Unused argument 'info'` - get rid of unused arguments and use `*args` or `**kwargs` instead (or no arguments at all, if that's possible). E.g. we have a lot of resolvers that look like this:
```
def resolve_payments(self, info):
        return self.payments.all()
```
`info` argument is not used which raises `pylint` warning. Here we could use `*args` so that this functions is able to take `info` but not use it.

- `graphql/order/types.py` - get rid of `staticmethod` decorators - they're not required in field resolvers 

- `Redefining built-in 'input'` / `Redefining built-in 'id'` - we should rename those variables, maybe better names would be `data` / `node_id`. I think we should keep those names in the public schema untouched so that we don't have to update our client apps now; just do the change in Python functions that consume those arguments.

NyanKiyoshi:
Closes #3909 

### Fixes (total of 390)
```
- views.py:112:4: R1260: 'execute_graphql_request' is too complex. The McCabe rating is 7 (too-complex)
- utils.py:23:0: R1260: 'get_nodes' is too complex. The McCabe rating is 12 (too-complex)
- utils.py:56:8: W0612: Unused variable 'model' (unused-variable)
- (many) shipping/mutations.py:47:41: W0622: Redefining built-in 'input' (redefined-builtin)
- (many) shipping/mutations.py:208:32: W0622: Redefining built-in 'id' (redefined-builtin)
- (a few) shipping/mutations.py:117:8: W0622: Redefining built-in 'type' (redefined-builtin)
- (many) shipping/mutations.py:208:20: W0613: Unused argument 'root' (unused-argument)
- (a few) order/types.py:249:4: W0211: Static method with 'self' as first argument (bad-staticmethod-argument)
- some missing optimizations in the gql types
- some typos
- some PEP8 issues
```

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -124,7 +124,7 @@ class CustomerCreate(ModelMutation, I18nMixin):
         model = models.User
 
     @classmethod
-    def user_is_allowed(cls, user, _data):
+    def user_is_allowed(cls, user):
         return user.has_perm('account.manage_users')
 
     @classmethod
@@ -189,7 +189,7 @@ class LoggedUserUpdate(CustomerCreate):
         model = models.User
 
     @classmethod
-    def user_is_allowed(cls, user, _data):
+    def user_is_allowed(cls, user):
         return user.is_authenticated
 
     @classmethod
@@ -214,7 +214,7 @@ class CustomerDelete(CustomerDeleteMixin, UserDelete):
             required=True, description='ID of a customer to delete.')
 
     @classmethod
-    def user_is_allowed(cls, user, _data):
+    def user_is_allowed(cls, user):
         return user.has_perm('account.manage_users')
 
 
@@ -230,7 +230,7 @@ class StaffCreate(ModelMutation):
         model = models.User
 
     @classmethod
-    def user_is_allowed(cls, user, _data):
+    def user_is_allowed(cls, user):
         return user.has_perm('account.manage_staff')
 
     @classmethod
@@ -297,12 +297,12 @@ class StaffDelete(StaffDeleteMixin, UserDelete):
             required=True, description='ID of a staff user to delete.')
 
     @classmethod
-    def user_is_allowed(cls, user, _data):
+    def user_is_allowed(cls, user):
         return user.has_perm('account.manage_staff')
 
     @classmethod
     def perform_mutation(cls, root, info, **data):
-        if not cls.user_is_allowed(info.context.user, data):
+        if not cls.user_is_allowed(info.context.user):
             raise PermissionDenied()
 
         user_id = data.get('id')
@@ -428,7 +428,7 @@ class AddressCreate(ModelMutation):
         return response
 
     @classmethod
-    def user_is_allowed(cls, user, _data):
+    def user_is_allowed(cls, user):
         return user.has_perm('account.manage_users')
 
 
@@ -485,7 +485,7 @@ class AddressDelete(ModelDeleteMutation):
 
     @classmethod
     def perform_mutation(cls, root, info, **data):
-        if not cls.user_is_allowed(info.context.user, data):
+        if not cls.user_is_allowed(info.context.user):
             raise PermissionDenied()
 
         node_id = data.get('id')
@@ -576,7 +576,7 @@ class CustomerAddressCreate(ModelMutation):
         exclude = ['user_addresses']
 
     @classmethod
-    def user_is_allowed(cls, user, _data):
+    def user_is_allowed(cls, user):
         return user.is_authenticated
 
     @classmethod
--- a/saleor/graphql/account/mutations.py
+++ b/saleor/graphql/account/mutations.py
@@ -507,7 +507,7 @@ class AddressDelete(ModelDeleteMutation):
 
         # Refresh the user instance to clear the default addresses. If the
         # deleted address was used as default, it would stay cached in the
-        # user instance and the invalid ID returned in the response migt cause
+        # user instance and the invalid ID returned in the response might cause
         # an error.
         user.refresh_from_db()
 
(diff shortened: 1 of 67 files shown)


saleor/graphql/core/mutations.py
```diff
@@ -259,7 +259,7 @@ def _save_m2m(cls, info, instance, cleaned_data):
                 f.save_form_data(instance, cleaned_data[f.name])
 
     @classmethod
-    def user_is_allowed(cls, user, input):
+    def user_is_allowed(cls, user):
```
maarcingebala: I was almost sure that we used `input` argument in one of our mutations to determine user's permission based on the input data. But that might have changed in the meantime so if we are sure that it's unused, we can drop it.

maarcingebala:
@NyanKiyoshi After merging #3891 we have a lot of conflicts in this PR, but this was unavoidable in this or the other PR anyway.

saleor/graphql/core/mutations.py
NyanKiyoshi: Yes, definitively sure about that.

NyanKiyoshi:
@maarcingebala will do that tomorrow :+1: 
<div align='right'><sup><i>Wow. That's a lot of conflicts.</i></sup></div>

NyanKiyoshi:
Alright, it's been rebased.

saleor/graphql/account/mutations.py
```diff
@@ -545,12 +547,14 @@ def perform_mutation(cls, root, info, address_id, user_id, type):
             raise ValidationError({
                 'address_id': 'The address doesn\'t belong to that user.'})
 
-        if type == AddressTypeEnum.BILLING.value:
+        raw_address_type = data.get('type')
```
akjanik: I would stick with `address_type` name, `raw` does not give (for me) any useful information.

saleor/graphql/account/mutations.py
```diff
@@ -545,12 +547,14 @@ def perform_mutation(cls, root, info, address_id, user_id, type):
...
+        if raw_address_type == AddressTypeEnum.BILLING.value:
             address_type = AddressType.BILLING
-        elif type == AddressTypeEnum.SHIPPING.value:
+        elif raw_address_type == AddressTypeEnum.SHIPPING.value:
             address_type = AddressType.SHIPPING
         else:
-            raise ValueError('Unknown value of AddressTypeEnum: %s' % type)
+            raise ValueError(
```
akjanik: Consider dropping this `else`, as an execution should never reach there because enums are validated by GraphQL itself.

saleor/graphql/checkout/mutations.py
```diff
@@ -127,21 +127,23 @@ def clean_input(cls, info, instance, input):
             default_billing_address = user.default_billing_address
             default_shipping_address = user.default_shipping_address
 
-        if 'shipping_address' in input:
-            shipping_address = cls.validate_address(input['shipping_address'])
+        if 'shipping_address' in raw_input:
```
korycins: I know that we are py3 but it seems that GH highlights `raw_input` same as `input`. Maybe we should use a different name? 
@akjanik, @NyanKiyoshi What do you think?

maarcingebala: Maybe `data`? Django forms use such name to initialize form instances.

NyanKiyoshi: `data` would be good!

saleor/graphql/utils.py
```diff
@@ -20,6 +24,49 @@ def get_database_id(info, node_id, only_type):
...
+
+
+def _resolve_graphene_type(types: List):
+    assert len(set(types)) == 1, 'Received IDs of more than one type.'
+    # get type by name
+    type_name = types[0]
+    for _, _type in registry._registry.items():
+        if _type._meta.name == type_name:
+            return _type
+    return None
```
akjanik: I can't recall a case when we may want to get `None` here - so maybe raising an ValidationError should be present here.

saleor/graphql/utils.py
```diff
@@ -20,6 +24,49 @@ def get_database_id(info, node_id, only_type):
...
+
+        try:
+            _type, _id = from_global_id(graphql_id)
+        except Exception:
+            invalid_ids.append(graphql_id)
+            continue
+
+        _check_graphene_type(graphene_type, _type)
+        pks.append(_id)
+        types.append(_type)
```
akjanik: @maarcingebala I spoke with @dominik-zeglen and both of us can't recall a case when client uses list with items of different types. This list seems to be legacy of idea that someone one day would.
Do you think we can drop it in favor of something like
```
# initally
type = None
# in the loop
_type, _id = from_global_id(graphql_id)
if _type is None:
    type = _type
else:
    assert _type == type, 'Received IDs of more than one type.' # Maybe raise?
```
It is not a big deal but we could avoid maintaining list that we expect to be one element one anyway.
Unless of course there is a case, when we want to use more types in one query.
What do you think?

maarcingebala: I agree - this function's purpose should be returning objects of a single type; the presence of a type different than the expected one should be considered an error.

saleor/graphql/utils.py
NyanKiyoshi: I decided to raise a `AssertionError`, is that okay? It seems to fit better in the context.

saleor/graphql/utils.py
NyanKiyoshi: Let me know if that was what you were thinking https://github.com/mirumee/saleor/pull/3942/commits/e147063275686d481113db040c51648f234f6188

akjanik: Yes, it's ok!

tests/api/test_graphql.py
```diff
@@ -240,7 +240,7 @@ def test_filter_by_query_param(qs):
     for q in test_kwargs:
         q_objects |= Q(**{q: test_kwargs[q]})
     # FIXME: django 1.11 fails on called_once_with(q_objects)
```
akjanik: Can you check if it's still an issue with 2.2?

NyanKiyoshi: It's still an issue. I wasn't sure about it being accidental.
Wouldn't be better to do a `@pytest.mark.xfail` or something? So we could provide a reason and see whenever it gets fixed? I'm unsure how we could do that cleanly... It seems odd to let that condition alone, it will get forgotten 😕

PR #3948, opened 2019-04-09
Implement variant availability, introducing discounts in variants

### LINKED ISSUE #3765: Add Discounted Price Field on GraphQL Products
It is normal to display the discounted price for a product on any listings. Ideally, we should have both the original price and the discounted price so we can show them side-by-side to the user. This is currently not available when the product is retrieved from graphql.

Add a sale price field that holds the discounted price on the product representation in graphql. Additionally, having a list of the vouchers applicable to that product would be very useful to further entice shoppers.

A roundabout way would be to request all the sales, all the vouchers, and then manually try to figure out if any of them apply to the product. When listing hundreds of products, this is just not efficient or practical.

NyanKiyoshi:
Fixes #3765. Proposed solution is to add a `availability` field of type `VariantAvailability` based on `ProductAvailability` for the `ProductType` variant as both variants and products have the same availability data fields, except price ranges.

### Changelist
- [x] Refactor `get_availability` to `get_product_availability`;
- [x] Implement `variants.availability`;
- [x] Write tests;
- [x] Write test for the local currency case;
- [x] Add missing gql descriptions;
- [x] Move availability to parent type;
- [x] Update the gql schema;
- [x] Ensure it doesn't break the storefront 2.0.;

### Pull Request Checklist


NyanKiyoshi:
@maarcingebala I like and agree with your suggestions.

korycins:
> Ideally, we should have only one way to access the product/variant price within the type and now we have two: price and availability, both for products and for variants. I'm thinking about dropping (deprecating) the price field and having only availability as the only and preferred way of accessing price info about a product.

What data is stored in `product.price` when each variant can have a different price?
+1 to depreacating the price field

> I'm still trying to think of a better name than "availability", e.g. "ProductPricing", "PricingInfo" etc. but I'm not convinced about any of them. If we don't come up with anything better maybe we should simply leave availability for now.

Personally I prefer more `PricingInfo`  than `availability`

Changes look really good. 👍

tests/api/test_variant_availability.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # ensure the availability was correctly retrieved and sent
+    assert availability
+
+    # check availability
+    assert availability['available'] is True
+    assert availability['onSale'] is True
+
+    # check the discount
+    assert availability['discount']['currency'] == 'USD'
```
akjanik: I suggest to use proper o object properties (`product.price.currency` or whatever it is) rather than hardcoded values - price in the fixture can change, same with currency, etc., it should not break tests in such case.

akjanik:
About naming, if we drop `availability` name, should we move `available` information somewhere else or keep it there?

NyanKiyoshi:
I deprecated `availability` and renamed them to `pricing`.

maarcingebala:
I spoke with @patrys about this and we have few more thoughts on this.
- Price field at the product level is useful from the dashboard/mutations perspective, as it is a single field for modifying the base product price. We concluded that we could actually keep it in the schema, but rename it to `basePrice` and make it protected and accessible only by staff members. So let's deprecate the `Product.price` and instead add `Product.basePrice` in our schema that will underneath use the `price` model field. This should be made clear in the field's description.
- `PricingInfo` in turn should be meant only for read-only usage e.g. by our storefront. Similarly here we should add a clear description so that any consumer of API will understand that they should use it for displaying price in their storefronts.

NyanKiyoshi:
I have doubts about that because we would have to refactor Product's price field to `basePrice`. Otherwise, the mutations `productCreate` and `productUpdate` will have a `price` field and in the meantime, `Product` will have a the `price` attribute resolved to `basePrice` instead of `price`.

NyanKiyoshi:
The fact that I restricted the product base price may break the storefront 2.0, if someone could make sure it doesn't or if we need to provide a fix for it.

NyanKiyoshi:
@auvipy don't worry, I was doing that 😉

NyanKiyoshi:
@maarcingebala the requested changes and compatibility checks have been made, it should be ready for review.

saleor/product/models.py
```diff
@@ -227,6 +231,14 @@ def __str__(self):
     def quantity_available(self):
         return max(self.quantity - self.quantity_allocated, 0)
 
+    @property
+    def is_visible(self):
+        return self.product.is_visible
+
+    @property
+    def is_available(self):
+        return self.product.is_in_stock()
```
maarcingebala: Shouldn't this use `product.is_available`?
```suggestion
        return self.product.is_available()
```

NyanKiyoshi: Thanks!

Issue #2588, opened 2018-08-16
Implement filtering interface


dominik-zeglen:
We should be able to filter objects given certain parameters.

maarcingebala:
Before we move on with implementing proper filtering in Dashboard 2.0, we need to figure out how the implementation should look like in the backend. First I'd like to analyze few other GraphQL APIs and propose a design of universal filtering interface in our API. 

Requirements:
- ideally filtering should be enclosed within a single object, similarly to how we handle sorting in `products` query:

```
type ProductOrder {
  field: ProductOrderField!
  direction: OrderDirection!
}
```

- it should be relatively easy to reuse the code that handles filtering between different queries/resolvers, so we can use the same patterns across the API

Some examples that I found:

- Github v4 API - filtering issues: `issues(filterBy: {states: OPEN, assignee: ""})` - seems very simple but I think the pattern it's not really re-used across different queries
- Strapi - [see the "Filters" section in the documentation](https://strapi.io/documentation/3.x.x/guides/graphql.html#configurations) - seems like quite powerful filtering capabilities which may be a good idea in an open API
- [GraphCMS](https://docs.graphcms.com/developers/api#filtering-entries)

We need to do some research and decide what shape of filtering API we need and can implement. Maybe we'll have to take a look at built-in support for `django-filters` in Graphene.

PR #3952, opened 2019-04-10
Filtering interface implementation

korycins:
Backend part - #2588

This feature introduces base filter logic that can be used in any graphQL queries.
 
How it works:
I've created `EnumFilter` which handle graphQL enums, it was needed to serve enum fields over api.  Based on `django-filters` and `EnumFilter` we create new `FilterSet` object with all fields that we want to use in filters. Our custom `FilterSet` should be assigned in a derived class from  `FilterInputObjectType`. `FilterInputObjectType` is responsible for converting all django-filter's fields to graphene fields (also enum fields). An object of a class that derives from `FilterInputObjectType` should be assigned as a new param in `FilterInputConnectionField` (by default it should be filters=FilterInputObjectType(), but you can also pass custom filters name - All filters will be available under this name in API).
You can also check the test:` test_filter_input which`. The test contains configuration step by step.

What do you think? If you like it we can think about the implementation of the filters for the rest of the queries. For now, I've added some filters only for products view.  
We can also think about some small documentation chapter about filtration.

### Pull Request Checklist


korycins:
@NyanKiyoshi No worries :)

tests/api/test_product.py
```diff
@@ -160,6 +177,89 @@ def test_product_query(staff_api_client, product, permission_manage_products):
     assert margin[1] == product_data['margin']['stop']
 
 
+def test_products_query_with_filters_product_type(
+        query_products_with_filters, staff_api_client, product,
+        permission_manage_products):
+    product_type = ProductType.objects.create(
+        name='Custom Type', has_variants=True, is_shipping_required=True,)
```
NyanKiyoshi:
```suggestion
        name='Custom Type', has_variants=True, is_shipping_required=True)
```

saleor/graphql/core/types/filter_input.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @classmethod
+    def get_filtering_args_from_filterset(cls):
+        """ Inspect a FilterSet and produce the arguments to pass to
+            a Graphene Field. These arguments will be available to
+            filter against in the GraphQL
+        """
+        if not cls.custom_filterset_class:
+            assert cls.model and cls.fields, (
+                "Provide filterset class or model and fields requested to "
```
NyanKiyoshi: Should be single quotes here

saleor/graphql/core/filters.py
```diff
@@ -0,0 +1,12 @@
+import django_filters
+
+
+class EnumFilter(django_filters.CharFilter):
+    """
+    Filter for GraphQL's enum objects. enum_class stores graphQL enum
+    needed to generated schema. method needs to be always pass explicitly"""
+    def __init__(self, enum_class, *args, **kwargs):
+        assert kwargs.get('method'), ("Providing exact filter method is "
```
NyanKiyoshi: Should be single quotes here as well

saleor/graphql/core/filters.py
```diff
@@ -0,0 +1,12 @@
+import django_filters
+
+
+class EnumFilter(django_filters.CharFilter):
+    """
+    Filter for GraphQL's enum objects. enum_class stores graphQL enum
```
maarcingebala: There are a few grammar issues in this docstring, maybe @timcla help with this (or Grammarly if he's busy)?

saleor/graphql/schema.graphql
```diff
@@ -1717,6 +1717,17 @@ type ProductDelete {
   product: Product
 }
 
+input ProductFilterInput {
+  name: String
+  name__icontains: String
+  category: ID
+  product_type: ID
+  price: Float
+  price__lt: Float
```
maarcingebala: Currently, we have a few filters that we've added manually to the `products` query and I think we should move them to the `filters` object so we don't duplicate them and have a single place to actually perform filtering. These are:
- `priceLte` & `priceGte` - these could be deprecated, as we already have them in in `filters`
- `collections` - allows you to get products that belong to given collections - I think it shouldn't be a problem to move it to have that handled by `ProductFilter`,
- `attributes` - this one is tricky because it uses very specific format, but if you go back in the Github history you'd find an implementation of that filter with django-filters.
- `stockAvailability` - should be moved to `ProductFilter`
- `query` - this one is used in a few more resolvers and I know that our Dashboard 2.0 uses it for searching objects. I think that I'd rather drop `name` and `name__icontains` for now in favor of `query`. Question is if we can move `query` to the `filters` object easily.

saleor/graphql/core/fields.py
```diff
@@ -66,3 +68,55 @@ def resolve_connection(cls, connection, default_manager, args, iterable):
...
+            enforce_first_or_last, filterset_class, root, info, **args):
+
+        # Disable `enforce_first_or_last` if not querying for `edges`.
+        values = [
+            field.name.value
+            for field in info.field_asts[0].selection_set.selections]
+        if 'edges' not in values:
+            enforce_first_or_last = False
+
+        filters_input = args.get('filters')
```
maarcingebala: I might be wrong, but shouldn't this actually use `filters_name` from the `__init__`? Isn't the `filters` name hardcoded?

korycins: @maarcingebala  you are right

saleor/graphql/schema.graphql
korycins: Do we have our own solutions for deprecation argument/input fields? 
If I understand it correctly, this feature is not supported by graphene.
https://github.com/graphql-python/graphene/issues/799
https://github.com/graphql/graphql-spec/issues/197

maarcingebala: Right, I guess we cannot deprecate an argument in any way. It won't look very elegant but I think  we should keep the old ones for now and try to implement as many as we can within the `filters` object. Once frontend apps migrate to the new approach we'll drop the old ones.

maarcingebala: @korycins I have a few more thoughts about this particular filter, after discussion with @patrys. It doesn't make much sense to have the exact `price` filter - usually, you'd want to filter by prices ranges. I think that we could replace all three price filter with something like:
```
type PriceFilter {
  min: Float
  max: Float
}

input ProductFilterInput {
  price: PriceFilter
}
```

If we do that, I guess we won't need to use any of "__lt", "__gt" or "__icontains" suffixes in our filter names. That in turns means that maybe we could use camel case for filter fields names.

korycins: @maarcingebala Sounds good.

saleor/graphql/core/types/common.py
```diff
@@ -77,3 +77,15 @@ def get_adjusted(image, alt, size, rendition_key_set, info):
             url = image.url
         url = info.context.build_absolute_uri(url)
         return Image(url, alt)
+
+
+class PriceInput(graphene.InputObjectType):
```
akjanik: If we call below `DateRangeInput`, maybe this one could be `PriceRangeInout`? Or the previous one just `DateInput`? Would be more consistent IMO.

saleor/graphql/core/types/filter_input.py
```diff
@@ -0,0 +1,63 @@
...
+        else:
+            _meta.fields = fields
+
+        super().__init_subclass_with_meta__(_meta=_meta, **options)
+
+    @classmethod
+    def get_filtering_args_from_filterset(cls):
+        """ Inspect a FilterSet and produce the arguments to pass to
+            a Graphene Field. These arguments will be available to
+            filter against in the GraphQL
```
akjanik: I am not sure if I understand this sentence, what do you mean with `in GraphQL`?

korycins: It was taken directly from graphene_django.
I assume that it means that it takes all discovered filterset's fields and convert them to graphQL's types. Then, all these arguments will be visible in the schema.

saleor/graphql/order/enums.py
```diff
@@ -7,5 +7,9 @@
 
 
 class OrderStatusFilter(graphene.Enum):
-    READY_TO_FULFILL = 'READY_TO_FULFILL'
-    READY_TO_CAPTURE = 'READY_TO_CAPTURE'
+    READY_TO_FULFILL = 'ready_to_fulfill'
+    READY_TO_CAPTURE = 'ready_to_capture'
+    UNFULFILLED = 'unfulfilled'
+    PARTIALLY_FULFILLED = 'partially fulfilled'
```
akjanik: Do we want space (` `) here?

saleor/graphql/core/fields.py
```diff
@@ -66,3 +68,57 @@ def resolve_connection(cls, connection, default_manager, args, iterable):
...
+            qs,
+            max_limit,
+            enforce_first_or_last,
+            root,
+            info,
+            **args
+        )
+
+    def get_resolver(self, parent_resolver):
+        return partial(
```
akjanik: If I understand correctly, you wanted to get the same `partial` as base `DjangoConnectionField`'s `get_resolver` returns extended by two arguments. In this case, I think you can do
```
returns partial(super().get_resolver(parent_resolver), self.filterset_class, self.filter_field_name)
```
as partial of partial, returns another partial. I am not 100% sure however, but if tests/usage won't fail, we could save some lines of code.

korycins: Will test it. Thanks!

saleor/graphql/product/types/products.py
```diff
@@ -8,28 +7,40 @@
...
+from .attributes import Attribute, SelectedAttribute
+from .digital_contents import DigitalContent
+
+
+def prefetch_products(info, *_args, **_kwargs):
+    """Prefetch products visible to the current user.
+
+    Can be used with models that have the `products` relationship. Queryset of
+    products being prefetched is filtered based on permissions of the viewing
+    user, to restrict access to unpublished products to non-staff users.
```
akjanik:
```suggestion
    user, to restrict access to unpublished products for non-staff users.
```

saleor/graphql/product/types/products.py
```diff
@@ -8,28 +7,40 @@
...
+from ...utils import get_database_id, reporting_period_to_date
+from ..enums import OrderDirection, ProductOrderField
+from .attributes import Attribute, SelectedAttribute
+from .digital_contents import DigitalContent
+
+
+def prefetch_products(info, *_args, **_kwargs):
+    """Prefetch products visible to the current user.
+
+    Can be used with models that have the `products` relationship. Queryset of
```
NyanKiyoshi:
```suggestion
    Can be used with models that have the `products` relationship. The queryset of
```

saleor/graphql/product/types/products.py
```diff
@@ -8,28 +7,40 @@
...
+    products being prefetched is filtered based on permissions of the viewing
```
NyanKiyoshi: Sounds weird to me, what do you think?
```suggestion
    products being prefetched is filtered based on permissions of the requesting
```

saleor/graphql/product/types/products.py
NyanKiyoshi: I would rather say `from` instead.

saleor/graphql/account/filters.py
```diff
@@ -0,0 +1,109 @@
+import django_filters
+from django.db.models import Count, Sum
+
+from saleor.graphql.utils import filter_by_query_param
```
NyanKiyoshi: Absolute import here

saleor/graphql/product/filters.py
```diff
@@ -2,9 +2,21 @@
 import operator
 from collections import defaultdict
 
-from django.db.models import Q
+import django_filters
+from django.db.models import Q, Sum
+from graphene_django.filter import GlobalIDFilter, GlobalIDMultipleChoiceFilter
 
-from ...product.models import Attribute
+from saleor.search.backends import picker
```
NyanKiyoshi: Absolute import

saleor/graphql/product/filters.py
```diff
@@ -59,3 +71,126 @@ def sort_qs(qs, sort_by_product_order):
...
                          + sort_by_product_order['field'])
     return qs
+
+
+def filter_products_by_stock_availability(qs, stock_availability):
+    qs = qs.annotate(total_quantity=Sum('variants__quantity'))
+    if stock_availability == StockAvailability.IN_STOCK:
+        qs = qs.filter(total_quantity__gt=0)
+    elif stock_availability == StockAvailability.OUT_OF_STOCK:
+        qs = qs.filter(total_quantity__lte=0)
```
NyanKiyoshi: Wouldn't it be faster to only look for quantities equal to 0?

saleor/graphql/account/filters.py
```diff
@@ -0,0 +1,109 @@
...
+        qs = qs.filter(money_spent__lte=money_spent_lte)
+    if money_spent_gte:
+        qs = qs.filter(money_spent__gte=money_spent_gte)
+    return qs
+
+
+def filter_number_of_orders(qs, _, value):
+    qs = qs.annotate(total_orders=Count('orders'))
+    gte = value.get('gte')
+    lte = value.get('lte')
```
NyanKiyoshi: I believe it's a bit faster for python to do
```python
gte, lte = value.get('gte'), value.get('lte')
```

But I'm not 100% sure. Let me know what you think.

patrys: I don't think it's faster but may be more readable if the two are related.

saleor/graphql/core/types/common.py
```diff
@@ -77,3 +77,22 @@ def get_adjusted(image, alt, size, rendition_key_set, info):
...
+
+class PriceRangeInput(graphene.InputObjectType):
+    gte = graphene.Float(
+        description='Price greater than or equal', required=False)
+    lte = graphene.Float(
+        description='Price less than or equal', required=False)
+
+
+class DateRangeInput(graphene.InputObjectType):
+    from_date = graphene.Date(description='Starting date from', required=False)
```
maarcingebala: I guess we don't have to duplicate the word "date" in the field name since it's included in the type name, and probably it will be understandable from the filter name as well. "From" and "To" should be fine.

maarcingebala: And for the descriptions maybe simply: "Start date" and "End date".

saleor/graphql/product/types/attributes.py
```diff
@@ -0,0 +1,103 @@
...
+        description='Value of an attribute.', required=True)
+
+    class Meta:
+        description = 'Represents a custom attribute.'
+
+
+class AttributeInput(graphene.InputObjectType):
+    slug = graphene.String(
+        required=True, description=AttributeDescriptions.SLUG)
+    attribute_value = graphene.String(
```
maarcingebala: I'd change it to "value". I'm only afraid that people using this might pass the display value here instead of slug and wonder why it doesn't work, but I don't have any better idea how to name this field.

maarcingebala:
@korycins Looks good! I have few notes to change later but I'd really like to merge it as it is and make some improvements later. Remember to update the changelog ;)

PR #3958, opened 2019-04-11
Implement the menu items reorder mutation

### LINKED ISSUE #3912: API for navigation management in Dashboard 2.0
Shortly we'll be implementing a new Dashboard 2.0 section for managing storefront navigation menus. We need to make sure that the API that we have currently allow implementing this UI:

image

We already have some mutations to add menus and menu items, but I'm not sure if they allow convenient support for all the reordering and drag-drop features that this UI assumes (shape of the API should be discussed with @dominik-zeglen). 

@patrys once suggested if we shouldn't keep the entire menu structure as JSON in the database, instead of MPTT models. In the old storefront, we serialize menus to JSON anyway and cache it in a request to save some database queries. We should do some research whether it is more convenient for us to fully migrate to JSON for menus or keep the current model. We have to remember that menus and menu items are translatable and we should support that.

NyanKiyoshi:
Closes #3912.

### Description
This adds the `menuItemMove` mutation, which its role is to allow partial (or fully) reordering of menu's (or menu**s**') items.

### Definition
It's defined as following.

```gql
input MenuItemMoveInput {
  itemId: ID!
  parentId: ID
  sortOrder: Int
}

mutation menuItemMove(move: [MenuItemMoveInput]!) { ... }
```

### How each field works
| field | description |
| --- | -- |
| ``` parentId = MenuItem! ``` | It moves the given menu item to a given parent, so the given menu item will become the child of `parentId`. |
| `parentId = null` | It moves the given menu item to the top level menu. |
| `sortOrder` | Moves the menu item to a given position (`0` to `x`) |

### Example Usage
```gql
mutation menuItemMove($move: [MenuItemMoveInput]!) {
  menuItemMove(move: $move) {
    errors {
      field
      message
    }

    menuItem {
      sortOrder
   }
  }
}
```

```
{
  "move": [
    {"itemId": "TWVudUl0ZW06Mw==", "parentId": null, "sortOrder": 1}
  ]
}
```

### Pull Request Checklist


tests/conftest.py
```diff
@@ -774,6 +774,14 @@ def menu_item(menu):
         url='http://example.com/')
 
 
+@pytest.fixture
+def menu_item_list(menu):
+    menu_item_1 = MenuItem.objects.create(menu=menu, name='Link 1')
```
korycins: https://docs.djangoproject.com/en/dev/ref/models/querysets/#bulk-create

NyanKiyoshi: Nice. There is probably quite a few places that could use that change! Thanks for letting me know about this new feature!

NyanKiyoshi: The issue is that the MPTT tree is not yet compatible with bulk creation, it will fail 😕

saleor/graphql/menu/mutations.py
```diff
@@ -196,6 +198,103 @@ def user_is_allowed(cls, user, input):
...
+_MenuMoveOperation = namedtuple(
+    '_MenuMoveOperation', ('menu_item', 'parent', 'sort_order'))
+
+
+class MenuItemMove(BaseMutation):
+    menu_item = graphene.List(
+        MenuItem, description='Assigned menu to move within.')
+
+    class Arguments:
+        move = graphene.List(
```
maarcingebala: Maybe we could rename it to `moves`? This is a list and we also iterate over its items later.

saleor/graphql/menu/mutations.py
```diff
@@ -196,6 +198,103 @@ def user_is_allowed(cls, user, input):
         return user.has_perm('menu.manage_menus')
 
 
+_MenuMoveOperation = namedtuple(
+    '_MenuMoveOperation', ('menu_item', 'parent', 'sort_order'))
+
+
+class MenuItemMove(BaseMutation):
+    menu_item = graphene.List(
```
maarcingebala: I think we would make @dominik-zeglen's life easier, if we also returned the whole `menu` that this item belongs to, is that right Dominik? Apollo Client would be able to refresh the whole menu in its cache based on that.

saleor/graphql/menu/mutations.py
```diff
@@ -196,6 +198,103 @@ def user_is_allowed(cls, user, input):
...
+
+        # Move the menu item
+        if item.sort_order is not None:
+            menu_item.sort_order = item.sort_order
+
+        menu_item.save()
+
+    @classmethod
+    def perform_mutation(cls, root, info, move):
+        move = cls.clean_move(info, move)
```
maarcingebala: There is a lot of things going on in this mutation and I'm a little confused with switching between "moves" and "operations". I don't know how the others feel about it, but maybe some renaming would make it easier to understand. What do you think about:

```
operations = cls.clean_moves(moves)
```

and later:

```
for operation in operations:
    cls.perform_operation(operation)
    menu_items.append(operation.menu_item)
```

When looking at `process_items` I get confused what items are we actually operating on: menu items, moves or operations. This little renaming might help.

saleor/graphql/menu/mutations.py
NyanKiyoshi: Let me know.

dominik-zeglen: Definitely!

saleor/graphql/menu/mutations.py
```diff
@@ -196,6 +198,103 @@ def user_is_allowed(cls, user, input):
...
+
+        if operation.parent:
+            if operation.menu_item.is_ancestor_of(operation.parent):
+                raise ValidationError({
+                    'parent': (
+                        'Cannot assign a node as child of '
+                        'one of its descendants.')})
+
+    @classmethod
+    def move_to_operation(cls, info, move) -> _MenuMoveOperation:
```
maarcingebala: What about renaming this as well to `get_operation`? When reading "move_to_operation"  it sounds like this function was literally "moving" something to some operation. What it does, is extracting your internal "operation" object from input item. I guess "get_operation" would sound good.

saleor/graphql/menu/mutations.py
```diff
@@ -196,6 +198,103 @@ def user_is_allowed(cls, user, input):
...
+    @staticmethod
+    def pre_validate_move(move):
+        """Validate if the given move could be possibly possible."""
+        if move.parent_id:
+            if move.item_id == move.parent_id:
+                raise ValidationError({
+                    'parent': 'Cannot assign a node to itself.'})
+
+    @staticmethod
+    def post_validate_move(operation: _MenuMoveOperation):
```
maarcingebala: Since this takes operation as an argument, I'd suggest renaming it to `clean_operation`.

saleor/graphql/menu/mutations.py
```diff
@@ -196,6 +198,103 @@ def user_is_allowed(cls, user, input):
...
+    class Arguments:
+        moves = graphene.List(
+            MenuItemMoveInput,
+            required=True, description='The menu position data')
+
+    class Meta:
+        description = 'Moves items of menus'
+
+    @staticmethod
+    def pre_validate_move(move):
```
maarcingebala: To not introduce new wording I'd stick with good ol' "clean" and have here `clean_move`. It's easier to read.

PR #3963, opened 2019-04-12
Rename Cart to Checkout

### LINKED ISSUE #3903: Rename Cart to Checkout
We should unify naming related to checkout functionality, as currently, we're mixing "checkout" with "cart" in many places, which leads to confusion when reading the code. I recommend that we settle on the name "checkout" and rename the `Cart` model and all other occurrences of `cart`.

michaljelonek:
Resolves #3903. 

There are no logic fixes in here. Some comments/functions might need to be renamed to make more sense or remove bloat. 

Todos: 
- [x] Fix urls as there are conflicts, e.g. `cart:index` and `checkout:index` both exist, so can't rename to `checkout:index` easily (including views).
- [ ] *optional* Change css to use `checkout` instead of `cart`.
- [ ] *optional* Rename Cart to Checkout on storefront _where_ it makes sense. Eventually use the same naming like in PWA (bag and basket). Discuss?

CSS and naming changes are very minor and not that important, so both optional steps can be skipped or changed at a later date. 

### Pull Request Checklist


saleor/checkout/urls.py
```diff
@@ -5,21 +5,18 @@
 
 checkout_urlpatterns = [
     url(r'^$', views.checkout_index, name='index'),
+    url(r'^details/$', views.checkout_details, name='details'),
```
maarcingebala: Before we had `cart_index`; I think it would be fine to have `checkout_index` here, to not introduce new wording like "details" which is kind of confusing when you also have "summary" and "order summary" views. I know "index" is also not perfect but I don't have any better idea.

saleor/checkout/urls.py
```diff
@@ -5,21 +5,18 @@
...
+    url(r'^shipping-options/$', views.checkout_shipping_options,
+        name='shipping-options'),
     url(r'^shipping-address/', views.checkout_shipping_address,
         name='shipping-address'),
     url(r'^shipping-method/', views.checkout_shipping_method,
         name='shipping-method'),
-    url(r'^summary/', views.checkout_summary, name='summary'),
-    url(r'^remove_voucher/', remove_voucher_view,
+    url(r'^order-summary/', views.checkout_order_summary, name='order-summary'),
+    url(r'^summary/$', views.cart_summary, name='summary'),
```
maarcingebala: What this view does is rendering a "cart dropdown" HTML snippet, called via AJAX in one of storefront's JS files and injected to regular template. Let's change it to:

```suggestion
    url(r'^dropdown/$', views.checkout_dropdown, name='dropdown'),
```

BTW what an ugly pattern when looking at it from a perspective. I'm so happy that we're migrating to GraphQL/SPA and the old strorefront will be deprecated at some point.

saleor/checkout/urls.py
```diff
@@ -5,21 +5,18 @@
...
+    url(r'^clear/$', views.clear_checkout, name='clear'),
+    url(r'^shipping-options/$', views.checkout_shipping_options,
+        name='shipping-options'),
     url(r'^shipping-address/', views.checkout_shipping_address,
         name='shipping-address'),
     url(r'^shipping-method/', views.checkout_shipping_method,
         name='shipping-method'),
-    url(r'^summary/', views.checkout_summary, name='summary'),
-    url(r'^remove_voucher/', remove_voucher_view,
+    url(r'^order-summary/', views.checkout_order_summary, name='order-summary'),
```
maarcingebala: If we introduce the change I suggested below, we'll be able to change this to:

```suggestion
    url(r'^summary/', views.checkout_summary, name='summary'),
```

saleor/checkout/utils.py
```diff
@@ -575,92 +575,92 @@ def _check_new_cart_address(cart, address, address_type):
...
-            cart.shipping_address.delete()
-        cart.shipping_address = address
-        cart.save(update_fields=['shipping_address'])
+            checkout.shipping_address.delete()
+        checkout.shipping_address = address
+        checkout.save(update_fields=['shipping_address'])
 
 
-def get_cart_data_for_checkout(cart, discounts, taxes):
+def get_checkout_data_for_checkout(checkout, discounts, taxes):
```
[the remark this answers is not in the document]
maarcingebala: I investigated a bit more and I found out that above we have `get_checkout_data` function which does almost the same thing as this one 🤦‍♂️. Maybe we could we drop `get_checkout_data` and keep this one only?

saleor/checkout/urls.py
michaljelonek: Well to be fair, I based this off the docstring which does say "details". And having two index urls I think might even be more confusing as it would end up being something along the lines of `checkout:index` and `checkout:checkout-index`.

maarcingebala: Ah right, we have the other "index" above, I missed it. Alright, so let's keep the "details" name.

michaljelonek: Eventually rename the first `index` to `start` as all it does is redirect to `checkout:shipping-address`, so all it does is start the checkout process it seems.

maarcingebala: Sounds good.

PR #3968, opened 2019-04-15
Overall improvement of the storefront 2.0 performances

### LINKED ISSUE #3631: Optimize Storefront 2.0 queries and mutations
We should analyze GraphQL queries sent by Storefront 2.0 and try to optimize the number of database queries made underneath. Currently, I'm getting the following counts:

- product details page: 40 queries
- mutation `updateCheckoutLine` with single variant: 34 queries
- mutation `checkoutShippingAddressUpdate`: 38 queries

I suspect that the `Checkout` type is not optimized well. It has relationships with a few other models and every mutation that returns it seems to be performing too many DB queries.

NyanKiyoshi:
This fixes #3631.

### Optimization of Single Nodes
This pull enables optimizations for single nodes. Which optimizes a lot of requests. See https://github.com/tfoxy/graphene-django-optimizer/pull/18 for more information about this.

### Overriding the Retrieval of a Single Node
Following the https://github.com/tfoxy/graphene-django-optimizer/pull/18 changes, getting a single node whether optimized or not is now simpler. You should always call `maybe_optimize(info, qs, pk)` and not directly call `get_optimized_node(info, qs, pk)` when overriding `get_node(info, pk)` as it would result to issues when the node is not supposed to actually get optimized.

This sums up of doing this:

Screenshot 2019-04-18 at 09 54 42

### During The Payment Capture
- Prefetch order lines that then get proceeded;
- Prefetch the user that will later get used to get the email address.

## Comparison

| Page | Before | After |
|------|--------|------|
| Product Listing | 29 queries | 11 queries |
| Product Details | 40 queries | 14 queries |
| Checkout Shipping Method Update | 82 queries | 38 queries |
| Placing an Order | 135 queries | 64 queries |
| ... | ... | ... |

## Pull Request Checklist

### DIFF
--- a/saleor/graphql/payment/mutations.py
+++ b/saleor/graphql/payment/mutations.py
@@ -14,6 +14,7 @@ from ..account.types import AddressInput
 from ..checkout.types import Checkout
 from ..core.mutations import BaseMutation
 from ..core.scalars import Decimal
+from ..core.utils import from_global_id_strict_type
 from .enums import PaymentGatewayEnum
 from .types import Payment
 
@@ -54,7 +55,7 @@ class CheckoutPaymentCreate(BaseMutation, I18nMixin):
 
     @classmethod
     def perform_mutation(cls, _root, info, checkout_id, **data):
-        checkout_id = cls.from_global_id_strict_type(
+        checkout_id = from_global_id_strict_type(
             info, checkout_id, only_type=Checkout, field='checkout_id')
         checkout = models.Checkout.objects.prefetch_related(
             'lines__variant__product__collections').get(pk=checkout_id)
--- a/saleor/graphql/payment/mutations.py
+++ b/saleor/graphql/payment/mutations.py
@@ -3,11 +3,10 @@ from textwrap import dedent
 import graphene
 from django.conf import settings
 from django.core.exceptions import ValidationError
-from graphql_jwt.decorators import permission_required
 
 from ...core.utils import get_client_ip
 from ...core.utils.taxes import get_taxes_for_address
-from ...payment import PaymentError
+from ...payment import PaymentError, models
 from ...payment.utils import (
     create_payment, gateway_capture, gateway_refund, gateway_void)
 from ..account.i18n import I18nMixin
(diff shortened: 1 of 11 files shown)


NyanKiyoshi:
There are bugs/issues around the optimizer library, so this PR will sadly (brings a lot of great optimizations) not be ready until the root of this issue is figured out. Hopefully I can dig more tomorrow around the troubles (relates to https://github.com/mirumee/saleor-storefront/issues/295).

NyanKiyoshi:
@auvipy please don't review drafts, that generating useless notifications on my side (thus making me lose precious time). Just leave a little 👍 reaction on the PR if you like it, that's enough to say you approve it!

NyanKiyoshi:
I added `from_global_id_strict_type` that has the aim to get easily a node from a given ID while passing a given queryset instead of using a simple `.get(pk)` that doesn't prefetch anything.

Let me know what you think, or if you see a better approach to this matter.

saleor/graphql/payment/mutations.py
```diff
@@ -55,8 +54,10 @@ class Meta:
 
     @classmethod
     def perform_mutation(cls, _root, info, checkout_id, **data):
-        checkout = cls.get_node_or_error(
-            info, checkout_id, field='checkout_id', only_type=Checkout)
+        checkout_id = cls.from_global_id_strict_type(
+            info, checkout_id, only_type=Checkout, field='checkout_id')
+        checkout = models.Checkout.objects.prefetch_related(
+            'lines__variant__product__collections').get(pk=checkout_id)
```
maarcingebala: Is there any way that we could use the optimizer here, maybe somehow with `Checkout.get_node`? I don't know if that's possible but it would be nice to be able to prefetch these fields only when they are requested in a query.

NyanKiyoshi: I'm not sure of what you want there. It's a mutation, the lines collections are needed to get the discounted prices.

maarcingebala: OK, I thought this was added here to optimize the mutation response, not the actual operation. In this case, this is fine.

PR #3979, opened 2019-04-17
Remove comment and set up explicit __hash__ function

akjanik:
It turns out that in Django 2.2 bug we were unknowingly relying on was fixed. Indeed, from https://docs.python.org/3/reference/datamodel.html#object.__hash__

> A class that overrides __eq__() and does not define __hash__() will have its __hash__() implicitly set to None.

> If a class that overrides __eq__() needs to retain the implementation of __hash__() from a parent class, the interpreter must be told this explicitly by setting __hash__ = <ParentClass>.__hash__.

Additional explanation of the issue:
https://code.djangoproject.com/ticket/30333

### Pull Request Checklist

### DIFF
--- a/saleor/account/models.py
+++ b/saleor/account/models.py
@@ -68,9 +68,7 @@ class Address(models.Model):
     def __eq__(self, other):
         return self.as_data() == other.as_data()
 
-    def __hash__(self):
-        # FIXME: in Django 2.2 this is not present if __eq__ is defined
-        return super().__hash__()
+    __hash__ = models.Model.__hash__
 
     def as_data(self):
         """Return the address as a dict suitable for passing as kwargs.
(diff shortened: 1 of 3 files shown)


saleor/account/models.py
```diff
@@ -68,9 +68,7 @@ def __str__(self):
     def __eq__(self, other):
         return self.as_data() == other.as_data()
 
-    def __hash__(self):
-        # FIXME: in Django 2.2 this is not present if __eq__ is defined
-        return super().__hash__()
+    __hash__ = models.Model.__hash__
```
patrys: Are there cases where `__eq__` would return `False` while `__hash__` remains `True`? Because this will break Python.

PR #3980, opened 2019-04-17
 Simplify permission management in API through meta classes 

### LINKED ISSUE #3949: Define mutation permissions in Meta class
Now that PR #3891 is merged we can think of further improvements to our mutations code. One thing that I noticed while doing the refactoring is that almost every mutation defines `user_is_allowed` method and in most cases, they perform the same logic of checking if a user has a particular permission. What do you think about refactoring it so that we could define permissions at the Meta class level, e.g.:

```
class CollectionRemoveProducts(BaseMutation):
    class Meta:
        description = 'Remove products from a collection.'
        permissions = ['product.manage_products', ]
```

At the same time, we could keep the `user_is_allowed` as an internal method used in `BaseMutations`. It could be overridden if you needed to perform some additional checks e.g.:

```
@classmethod
def user_is_allowed(cls, user, input):
    return user.is_authenticated
```
(example taken from `LoggedUserUpdate` mutation).

The default behavior of `user_is_allowed` would simply do `user.has_perms([permissions])` where `permissions` would come from the Meta class.

Such changes would allow us to reduce the number of lines of code per mutation even more.

What do you think @akjanik  @korycins @NyanKiyoshi?

NyanKiyoshi:
Closes #3949.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/core/mutations.py
+++ b/saleor/graphql/core/mutations.py
@@ -179,9 +179,9 @@ class BaseMutation(graphene.Mutation):
 
         Default implementation assumes that user is allowed to perform any
         mutation. By overriding this method or defining required permissions
-        in the meta-class, you can restrict access to it. `user` is
-        the User instance associated with the request and `input` is
-        the input data provided as mutation arguments.
+        in the meta-class, you can restrict access to it.
+
+        The `user` parameter is the User instance associated with the request.
         """
         if cls._meta.permissions:
             return user.has_perms(cls._meta.permissions)
--- a/saleor/graphql/core/mutations.py
+++ b/saleor/graphql/core/mutations.py
@@ -83,7 +83,11 @@ class BaseMutation(graphene.Mutation):
         description = dedent(description)
 
         if isinstance(permissions, str):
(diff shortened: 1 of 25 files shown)


saleor/graphql/core/mutations.py
```diff
@@ -157,19 +174,22 @@ def construct_instance(cls, instance, cleaned_data):
...
-    def user_is_allowed(cls, user):
+    def check_permissions(cls, user):
         """Determine whether user has rights to perform this mutation.
 
         Default implementation assumes that user is allowed to perform any
-        mutation. By overriding this method, you can restrict access to it.
-        `user` is the User instance associated with the request and `input` is
+        mutation. By overriding this method or defining required permissions
+        in the meta-class, you can restrict access to it. `user` is
+        the User instance associated with the request and `input` is
```
maarcingebala: The "input" part is outdated as we've changed the signature of this method recently.

maarcingebala:
@NyanKiyoshi I left one comment, but overall - this looks great! 🎉

Issue #3978, opened 2019-04-17
Do not enable Debug Toolbar by default with DEBUG mode


krzysztofwolski:
### What I'm trying to achieve
Django Debug Toolbar combined with debugger causes on my env timeouts and unfortunately, it's turned on by default with DEBUG mode:
https://github.com/mirumee/saleor/blob/master/saleor/settings.py#L254

This tool should be enabled on demand, like Silk:
https://github.com/mirumee/saleor/blob/master/saleor/settings.py#L278

### Describe a proposed solution
- add separate setting for enabling DDT (new environment variable)
- the default setting should be discussed

jxltom:
Is that a bug with debugtoolbar? It works fine for me locally.

koradon:
@jxltom for me loading products list in dashboard take a lot of time.

dushr:
@koradon are you sure its the debug toolbar? Have you tried loading the product list by explicitly removing debug toolbar from the installed apps?

If you're using dev server, the images are also served from the dev server - for me that was the problem - as it is one worker serving images (and if the size is not available, it is also resizing them)

koradon:
@dushr yes I tried and without debug toolbar, it is very fast.

krzysztofwolski:
As I wrote on @koradon s PR - DDT does not help at all when front uses GQL API. Which is our recommended way of doing new projects. Silk does not have this limitation.

@dushr How do you handle images on dev srv?

PR #3983, opened 2019-04-17
Add ENABLE_DEBUG_TOOLBAR to settings

koradon:
#3978 

### Pull Request Checklist

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -252,7 +252,7 @@ INSTALLED_APPS = [
 
 
 ENABLE_DEBUG_TOOLBAR = get_bool_from_env('ENABLE_DEBUG_TOOLBAR', False)
-if DEBUG and ENABLE_DEBUG_TOOLBAR:
+if ENABLE_DEBUG_TOOLBAR:
     MIDDLEWARE.append(
         'debug_toolbar.middleware.DebugToolbarMiddleware')
     INSTALLED_APPS.append('debug_toolbar')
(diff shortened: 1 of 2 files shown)


NyanKiyoshi:
Since it's a PR for the debug toolbar, could you take a minute to bump `django-debug-toolbar-request-history` to `0.0.10`? I made a minor change to the toolbar's request history, it now shows the requests methods, which is quite useful for backend developers.

saleor/settings.py
```diff
@@ -250,7 +250,9 @@ def get_bool_from_env(name, default_value):
     'phonenumber_field',
     'captcha']
 
-if DEBUG:
+
+ENABLE_DEBUG_TOOLBAR = get_bool_from_env('ENABLE_DEBUG_TOOLBAR', False)
```
NyanKiyoshi: Should we really disable it by default?

saleor/settings.py
```diff
@@ -276,7 +278,7 @@ def get_bool_from_env(name, default_value):
         'RESULTS_CACHE_SIZE': 100}
 
 ENABLE_SILK = get_bool_from_env('ENABLE_SILK', False)
-if ENABLE_SILK:
+if DEBUG and ENABLE_SILK:
```
jxltom: Sometimes I use silk or debugtoolbar to debug bugs, but these bugs may not be reproduces in ```DEBUG``` mode, should we only enable them in DEBUG mode?

koradon: Yea I was wondering if this next level of protection is required. Probably not.

saleor/settings.py
koradon: If we don't want the next line
`if DEBUG and ENABLE_DEBUG_TOOLBAR:` then yes. Otherwise, it will be added on production servers by default and we for sure don't want that.

saleor/settings.py
koradon: On the second thoughts, right now debug toolbar is added only in DEBUG mode so no changes to that.
I don't know probably this needs discussion :)

koradon:
@NyanKiyoshi lets bump dependencies in another PR

PR #4018, opened 2019-04-25
Implement order events

### LINKED ISSUE #2443: Log order and customer actions and events
Keep track of whatever changes and actions were made by the staff.

This is useful to catch errors and check data changes.

#### 1 — Models
Introduce a new model containing log entries. A possible model (based on django-admin's `LogEntry` model) would be the following:

| field name | content |
| ---------- | ---- |
| action_date | Datetime of the action |
| action_type | Action type from the enum: `CREATE\|EDIT\|DELETE` |
| user | FK to `User` |
| object | Object type name or verbose name if available (e.g.: `Product Variant`) |
| object_info | The object representation through `__str__` |

This model would introduce a new listing view: action logs.

#### 2 — Event trapping
To only log dashboard action, we cannot trap at the model level efficiently (unless we want to log the whole website).

##### 2.1 — Object creation and update
Object creation and update could be trapped by making a base form receiving the user request object (to catch the requester) and handling calls to `save(...)`.

##### 2.2 — Object deletion
We could implement a method `delete_object(request, object)` to replace `object.delete()` to trigger a log creation.

#### 3 — Log saving
As the dashboard doesn't depend on logging, we probably could make log entry creation as an asynchronous task.

NyanKiyoshi:
- [x] For the gql type of OrderEvent we have to include:
	- [x] New fields for getting data from the JSON parameters. [2]
	- [x] A payment method field taken from the parameters.
	- [x] Tests.

- [x] The OrderEvent model should have:
	- [x] A field `user` containing who triggered the event. [4]
	- [x] (outside) should contain a bunch of functions taking data to generate event(s)

- [x] Update dashboard 1.0 to be compatible with the latest changes (only order side)
- [x] Update docs definitions
- [x] Document the different parameters generated by the order events (docstrings + sphinx-apidoc)
- [x] Update definitions (migrations & GQL)

- Introduce new order event types (OrderEventsEnum):
	- [x] the order was placed
	- [x] **new:** the order payment failed [failure event]
		- [x] Implement it
	- [x] the payment was captured (+ amount)
	- [x] a note was added to the order
	- [x] the order confirmation has been set
	- [x] **new:** digital link was sent to the customer -> `saleor/order/emails.py:46`
		- [x] Implement it
	- [x] Manual staff actions events:
		- [x] the tracking email was sent -> `saleor/dashboard/order/views.py:688`
		- [x] **new:** a draft was created
			- [x] Implement it
		- [x] a draft was placed
		- [x] **new:** added products (store a strings)
			- [x] Implement it
		- [x] **new:** removed products (idem)
			- [x] Implement it
		- [x] order was manually fulfilled
		- [x] order was manually marked as paid

- [x] An event should be called to be created and may create other events.
- [x] We need to migrate the old events to the new model.
- [x] Refactor logic to commit on the go instead of waiting on the user's green light.

[2] Allows to retrieve additional data from an event (e.g., an amount, a quantity, a message, etc.)
[4] We want to know who triggered the event, even if it's the customer itself or a staff user.

[!] Failed payment expected data (payment id, etc. for user support) https://i.imgur.com/ayYrCjr.png

### Pull Request Checklist

### Magic Final Words
Closes #2443.


NyanKiyoshi:
I decided to move the customer events to #4039 as it would make this PR way too big, thus even harder to review.

saleor/events/models/OrderEvent.py
```diff
@@ -0,0 +1,274 @@
...
+        } for quantity, line in quantities_per_order_line]
+
+    def get_event_display(self):
+        return display_order_event(self)
+
+    @classmethod
+    def email_sent_event(
+            cls, *,
+            order: Order, email_type: OrderEventsEmails,
+            source: Optional[User]) -> models.Model:
```
maarcingebala: I would consider renaming all `source` arguments to `user`, it would be much more readable.

saleor/events/models/OrderEvent.py
```diff
@@ -0,0 +1,274 @@
...
+
+        return cls(
+            order=order, type=OrderEvents.DRAFT_REMOVED_PRODUCTS.value,
+            user=source,
+            parameters={
+                'lines': cls._lines_per_quantity_to_str_line_list(
+                    order_lines)})
+
+    @classmethod
+    def placed_event(
```
maarcingebala: I'd suggest `order_created_event`, it's a bit more explicit.

saleor/events/models/OrderEvent.py
```diff
@@ -0,0 +1,274 @@
...
+            order: Order, source: User,
+            oversold_items: List[str]) -> models.Model:
+        return cls(
+            order=order, type=OrderEvents.OVERSOLD_ITEMS.value,
+            user=source,
+            parameters={
+                'oversold_items': oversold_items})
+
+    @classmethod
+    def cancelled_event(
```
maarcingebala: We once discussed whether to use British or American spelling in Saleor, and we agreed on American. In American spelling this should be "canceled_event", also the event type inside is called `CANCELED`.

saleor/events/models/OrderEvent.py
```diff
@@ -0,0 +1,274 @@
...
+    @classmethod
+    def manually_marked_as_paid_event(
+            cls, *,
+            order: Order, source: User) -> models.Model:
+        return cls(
+            order=order, type=OrderEvents.ORDER_MARKED_AS_PAID.value,
+            user=source)
+
+    @classmethod
+    def fully_paid_event_event(cls, *, order: Order) -> models.Model:
```
maarcingebala: The word `event` is repeated here.

Would it make sense if we had a convention of naming events like `model_action_event`? For payment events below we actually use it: "payment_captured_event, "payment_refunded_event" and for draft-related actions "draft_something_event" etc. Here we would then have "order_fully_paid_event",  "order_marked_as_paid_event", "order_email_sent" etc. What do you think?

saleor/order/emails.py
```diff
@@ -65,6 +67,23 @@ def send_fulfillment_confirmation(order_pk, fulfillment_pk):
...
+        email_type=OrderEventsEmails.FULFILLMENT)]
+
+    # If digital lines were sent in the fulfillment email,
+    # trigger the event
+    if any((line for line in order if line.variant.is_digital())):
+        events.append(OrderEvent.email_sent_event(
+            order=order, source=user,
+            email_type=OrderEventsEmails.DIGITAL_LINKS))
+
+    return events
```
maarcingebala: I'm not sure about returning events here, none of the email function does that. If you wanted to encapsulate and reuse creating these two events I'd add some simple helper function only for that.

saleor/events/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .OrderEvent import OrderEvent
```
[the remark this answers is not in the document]
NyanKiyoshi: The issue is that the file would get quite huge as we add more event models.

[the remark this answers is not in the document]
maarcingebala: Then I'd suggest having this structure:

```
saleor/events
  models/
    order.py
    account.py
    ...
```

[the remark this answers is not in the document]
maarcingebala: ⬆️ This would match the structure of our modules in `/saleor` directory. Similar convention is used for modules in `/graphql`.

saleor/events/models/OrderEvent.py
```diff
@@ -0,0 +1,274 @@
...
+        return 'OrderEvent(type=%r, user=%r)' % (self.type, self.user)
+
+    @staticmethod
+    def _lines_per_quantity_to_str_line_list(quantities_per_order_line):
+        return [{
+            'quantity': quantity,
+            'item': str(line)
+        } for quantity, line in quantities_per_order_line]
+
+    def get_event_display(self):
```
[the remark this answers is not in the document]
NyanKiyoshi: I wouldn't think we should drop the order events from the dashboard 1.0 as this was a feature for quite some time in the order details view.

saleor/events/models/OrderEvent.py
NyanKiyoshi: Great suggestion. I refactored that and put a note about the naming convention in the docs.

saleor/checkout/utils.py
```diff
@@ -902,3 +906,33 @@ def clean_checkout(checkout: Checkout, taxes, discounts):
         raise ValidationError(
             'Provided payment methods can not cover the checkout\'s total '
             'amount')
+
+
+def place_checkout_to_order(request, checkout) -> Order:
```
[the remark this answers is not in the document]
NyanKiyoshi: This gives more flexibility. e.g. an order event generate a customer event. Simplicity may cost an additional query when writing to the DB, writes are not happening that often in e-commerce sites, so fair enough.

There is a case where we generate events from a for loop though. When we are bulk canceling orders. In that case, we could give it a special method, unless you see a better solution.

docs/architecture/events.rst
```diff
@@ -44,3 +44,46 @@ Order events
...
+
+Events Design
+-------------
+
+The ``OrderEvent`` class contains a bunch of static methods suffixed by
+``_event``. They take care of all the logic and required fields. It whether
+returns an instance of ``OrderEvent``. This design has for objective to allow
+delayed (or not) event creation in the database.
+
+The main power of this design comes when one wants to create multiple
```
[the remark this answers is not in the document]
NyanKiyoshi: This gives more flexibility. e.g. an order event generates a customer event. Dropping this would mean that simplicity may cost an additional query when writing to the DB but writes are not happening that often in e-commerce sites. Fair enough.

There is a case where we generate events from a for loop though. When we are bulk canceling orders. In that case, we could give it a special method, unless you see a better solution.

saleor/checkout/utils.py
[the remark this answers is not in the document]
NyanKiyoshi: Is the below diff what you are suggesting?

```diff
Index: saleor/checkout/utils.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- saleor/checkout/utils.py	(date 1556627593000)
+++ saleor/checkout/utils.py	(date 1556629757278)
@@ -839,7 +839,9 @@
 
 
 @transaction.atomic
-def create_order(checkout: Checkout, tracking_code: str, discounts, taxes):
+def create_order(
+        checkout: Checkout, tracking_code: str, discounts, taxes,
+        user: User):
     """Create an order from the checkout.
 
     Each order will get a private copy of both the billing and the shipping
@@ -875,6 +877,18 @@
 
     # assign checkout payments to the order
     checkout.payments.update(order=order)
+
+    # remove checkout after order is created
+    checkout.delete()
+
+    # Create the order placed and email confirmation sent events
+    OrderEvent.objects.bulk_create([
+        OrderEvent.order_created_event(order, user),
+        OrderEvent.email_sent_event(
+            order=order, email_type=OrderEventsEmails.ORDER, user=user)])
+
+    # Send the order confirmation email
+    send_order_confirmation.delay(order.pk)
     return order
 
 
@@ -906,33 +920,3 @@
         raise ValidationError(
             'Provided payment methods can not cover the checkout\'s total '
             'amount')
-
-
-def place_checkout_to_order(request, checkout) -> Order:
-    """Attempts to create an order from checkout.
-
-    This function creates an order from checkout and performs post-create
-    actions such as removing the checkout instance, sending order notification
-    email and creating order history events.
-
-    It returns the created order.
-    """
-    order = create_order(
-        checkout=checkout,
-        tracking_code=analytics.get_client_id(request),
-        discounts=request.discounts,
-        taxes=get_taxes_for_checkout(checkout, request.taxes))
-
-    # remove checkout after order is created
-    checkout.delete()
-
-    # Create the order placed and email confirmation sent events
-    OrderEvent.objects.bulk_create([
-        OrderEvent.order_created_event(order, request.user),
-        OrderEvent.email_sent_event(
-            order=order, email_type=OrderEventsEmails.ORDER,
-            user=request.user)])
-
-    # Send the order confirmation email
-    send_order_confirmation.delay(order.pk)
-    return order
Index: tests/test_checkout.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- tests/test_checkout.py	(date 1556627593000)
+++ tests/test_checkout.py	(date 1556629864789)
@@ -529,7 +529,8 @@
 
     with pytest.raises(InsufficientStock):
         create_order(
-            request_checkout, 'tracking_code', discounts=None, taxes=None)
+            request_checkout, 'tracking_code', discounts=None, taxes=None,
+            user=customer_user)
 
 
 def test_create_order_doesnt_duplicate_order(
@@ -541,18 +542,24 @@
     checkout.shipping_method = shipping_method
     checkout.save()
 
-    order_1 = create_order(checkout, tracking_code='', discounts=None, taxes=None)
+    order_1 = create_order(
+        checkout, tracking_code='', discounts=None, taxes=None,
+        user=customer_user)
     assert order_1.checkout_token == checkout_with_item.token
-    order_2 = create_order(checkout, tracking_code='', discounts=None, taxes=None)
+    order_2 = create_order(
+        checkout, tracking_code='', discounts=None, taxes=None,
+        user=customer_user)
     assert order_1.pk == order_2.pk
 
 
-def test_note_in_created_order(request_checkout_with_item, address):
+def test_note_in_created_order(
+        request_checkout_with_item, address, customer_user):
     request_checkout_with_item.shipping_address = address
     request_checkout_with_item.note = 'test_note'
     request_checkout_with_item.save()
     order = create_order(
-        request_checkout_with_item, 'tracking_code', discounts=None, taxes=None)
+        request_checkout_with_item, 'tracking_code', discounts=None,
+        taxes=None, user=customer_user)
     assert order.customer_note == request_checkout_with_item.note
 
 
Index: saleor/checkout/views/summary.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- saleor/checkout/views/summary.py	(date 1556627593000)
+++ saleor/checkout/views/summary.py	(date 1556629894789)
@@ -3,15 +3,16 @@
 from django.template.response import TemplateResponse
 from django.utils.translation import pgettext
 
+from ...core import analytics
 from ...account.models import Address
 from ...core.exceptions import InsufficientStock
 from ...discount.models import NotApplicable
 from ..forms import CheckoutNoteForm
 from ..utils import (
-    get_checkout_context, get_taxes_for_checkout, place_checkout_to_order,
+    get_checkout_context, get_taxes_for_checkout,
     update_billing_address_in_anonymous_checkout,
     update_billing_address_in_checkout,
-    update_billing_address_in_checkout_with_shipping)
+    update_billing_address_in_checkout_with_shipping, create_order)
 
 
 def handle_order_placement(request, checkout):
@@ -22,7 +23,12 @@
     and creating order history events.
     """
     try:
-        order = place_checkout_to_order(request, checkout)
+        order = create_order(
+            checkout=checkout,
+            tracking_code=analytics.get_client_id(request),
+            discounts=request.discounts,
+            taxes=get_taxes_for_checkout(checkout, request.taxes),
+            user=request.user)
     except InsufficientStock:
         return redirect('checkout:index')
     except NotApplicable:
Index: saleor/graphql/checkout/mutations.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- saleor/graphql/checkout/mutations.py	(date 1556627593000)
+++ saleor/graphql/checkout/mutations.py	(date 1556629751475)
@@ -5,13 +5,14 @@
 from django.core.exceptions import ValidationError
 from django.db import transaction
 
+from ...core import analytics
 from ...checkout import models
 from ...checkout.utils import (
     add_variant_to_checkout, add_voucher_to_checkout,
     change_billing_address_in_checkout, change_shipping_address_in_checkout,
     clean_checkout, get_or_create_user_checkout, get_taxes_for_checkout,
-    get_voucher_for_checkout, place_checkout_to_order,
-    recalculate_checkout_discount, remove_voucher_from_checkout)
+    get_voucher_for_checkout,
+    recalculate_checkout_discount, remove_voucher_from_checkout, create_order)
 from ...core.exceptions import InsufficientStock
 from ...core.utils.taxes import get_taxes_for_address
 from ...discount import models as voucher_model
@@ -461,7 +462,11 @@
 
         try:
             # FIXME: we shouldn't create the order, but only do checks
-            order = place_checkout_to_order(info.context, checkout)
+            order = create_order(
+                checkout=checkout,
+                tracking_code=analytics.get_client_id(info.context),
+                discounts=info.context.discounts,
+                taxes=get_taxes_for_checkout(checkout, info.context.taxes))
         except InsufficientStock:
             raise ValidationError('Insufficient product stock.')
         except voucher_model.NotApplicable:
Index: tests/test_order.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- tests/test_order.py	(date 1556627593000)
+++ tests/test_order.py	(date 1556629864794)
@@ -440,7 +440,8 @@
 
 
 def test_order_payment_flow(
-        request_checkout_with_item, client, address, shipping_zone, settings):
+        request_checkout_with_item, client, address, customer_user,
+        shipping_zone, settings):
     request_checkout_with_item.shipping_address = address
     request_checkout_with_item.billing_address = address.get_copy()
     request_checkout_with_item.email = 'test@example.com'
@@ -449,7 +450,8 @@
     request_checkout_with_item.save()
 
     order = create_order(
-        request_checkout_with_item, 'tracking_code', discounts=None, taxes=None)
+        request_checkout_with_item, 'tracking_code', discounts=None,
+        taxes=None, user=customer_user)
 
     # Select payment
     url = reverse('order:payment', kwargs={'token': order.token})
```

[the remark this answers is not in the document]
maarcingebala: Yep, I guess we need something like that. Do you think we could also refactor `handle_order_placement` in this PR?

docs/architecture/events.rst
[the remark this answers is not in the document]
maarcingebala: I can imagine this might be a problem if the number of records is large and in fact, bulk actions are executed one by one. But I'm always for going with the simple approach first and optimizing when we identify the performance issue. I guess we'll need to soon do some performance tests to find out.

auvipy:
Should I expect this feature on April release?

NyanKiyoshi:
@maarcingebala could you double check the OrderEvent model definitions? I'm confused by how the choices are supposed to be populated.

For example, `OrderStatus` populates the choices with `[('status-type', 'status-type-human-readable-through-gettext')]`, in the database the human readable value is stored instead of the "key". So I'm confused how we should actually proceed with choices, especially as the value can be different depending on what `gettext()` returned.

But it may work just fine, and seems to be the proper django way. I'm unsure about how django is performing the queries and returning the results, I can't seem to find any piece of documentation on that matter.

auvipy:
seems another rebase needed?

saleor/events/migrations/0067_replace_old_enum_namings.py
```diff
@@ -0,0 +1,49 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('events', '0066_auto_20190502_0910'),
+    ]
+
+    operations = [
+        migrations.RunPython(_replace_old_namings)
+    ]
```
auvipy: curious, can we reduce the number of new migrations for this PR?

NyanKiyoshi: This migration is a two-step migration. First we alter the field, then we update the database data.

If the python code fails for whatever reason, we want to be able to safely re-run it without editing the table again. That's why we need two migrations there.

auvipy: cool

docs/architecture/events.rst
```diff
@@ -6,41 +6,121 @@ Events
...
+order events in one go. Instead of generating multiple create query, one
+can simply pass the objects to django's |QuerySet.bulk_create|_ method.
+
+If one simply does not want to bulk create, delay the creation, one can
+directly run ``save()`` for the creation.
+
+.. |QuerySet.bulk_create| replace:: ``QuerySet.bulk_create``
+.. _QuerySet.bulk_create: https://docs.djangoproject.com/en/1.10/ref/models/querysets/#django.db.models.query.QuerySet.bulk_create
+
+Sending Events
```
maarcingebala: I wouldn't use the word "send" here as this suggests that we were using signals or some other event-based architecture underneath. We simply create objects so let's say "Creating events" and reword other parts of the docs accordingly.

saleor/checkout/utils.py
```diff
@@ -871,6 +876,19 @@ def create_order(checkout: Checkout, tracking_code: str, discounts, taxes):
 
     # assign checkout payments to the order
     checkout.payments.update(order=order)
+
+    # remove checkout after order is created
+    checkout.delete()
+
+    # Create the order placed and email confirmation sent events
+    OrderEventManager()\
```
maarcingebala: A minor thing but we rather don't use backslashes to break long lines. Instead of chaining method calls here we could also just call them one by one.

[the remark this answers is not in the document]
maarcingebala: I agree about the last part, this might be more helpful it the event was created from the same task so we are actually sure that the task was executed.

saleor/events/bases.py
```diff
@@ -0,0 +1,36 @@
+from django.core.exceptions import ImproperlyConfigured
+from django.db import models
+
+
+class EventManager:
```
maarcingebala: I really appreciate the design but to be honest, the whole thing seems over-engineered. The whole "events" design could be implemented as a single model class (in our case `OrderEvent`) plus a bunch of module functions to create specific events. What we have now is a manager object which the only purpose is to nicely create multiple objects at once. I still think that this is not worth the effort as we never create that many of them at once.

saleor/graphql/core/enums.py
```diff
@@ -11,6 +11,35 @@ class ReportingPeriod(graphene.Enum):
     THIS_MONTH = 'THIS_MONTH'
 
 
+def to_enum(enum_cls, *, type_name=None, **options) -> graphene.Enum:
```
maarcingebala: Why do we need this function and what were the consequences of not having it before? Are there any problems with enums currently?

NyanKiyoshi: It avoids to repetitively have
```python
graphene.Enum('something', 
    str_to_enum(code.upper()), code) for code, name in enum_cls.CHOICES)
```

It would make it more readable and easier to maintain.

saleor/events/__init__.py
```diff
@@ -0,0 +1,148 @@
+from django.utils.translation import pgettext_lazy
```
korycins: Do we really need a special module only for keeping up to date history of changes in order? I understand that in further steps we could have CustomerEvent etc, but even though I prefer to keep this logic inside a given module. Personally, it is more intuitive.
1) For now, module events operates only on Order's events.  So it can be easily moved to Order module.
2) If you really need base class: Event Manager, you can put it inside the core module.
3) If you put OrderEvents inside Order module, there is no need to modify so many old migrations, which personally I don't like

docs/architecture/events.rst
```diff
@@ -6,41 +6,110 @@ Events
...
+| ``fulfillment_confirmation`` | ``FULFILLMENT``      | One or more of the order's items have been fulfilled (both physical or digital)  |
++------------------------------+----------------------+----------------------------------------------------------------------------------+
+| ``digital_links``            | ``DIGITAL_LINKS``    | The links to the order's digital goods have been sent.                           |
++------------------------------+----------------------+----------------------------------------------------------------------------------+
+
+
+Events Design
+-------------
+
+The ``OrderEventManager`` class contains a bunch of static methods suffixed by
```
maarcingebala: We need to update this section accordingly to current design.

saleor/dashboard/order/views.py
```diff
@@ -15,10 +15,16 @@
 from ...core.exceptions import InsufficientStock
 from ...core.utils import get_paginator_items
 from ...core.utils.taxes import get_taxes_for_address
-from ...order import OrderEvents, OrderEventsEmails, OrderStatus
+from ...order import OrderStatus
 from ...order.emails import (
-    send_fulfillment_confirmation, send_fulfillment_update,
+    send_fulfillment_confirmation_to_customer, send_fulfillment_update,
     send_order_confirmation)
+from ...order.events import (
```
maarcingebala: We could also do `from ...order import events` to not have to worry about formatting and ordering of these all functions here.

saleor/dashboard/order/views.py
```diff
@@ -77,17 +87,15 @@ def create_order_from_draft(request, order_pk):
...
+
         if form.cleaned_data.get('notify_customer'):
-            send_order_confirmation.delay(order.pk)
-            order.events.create(
-                parameters={
-                    'email': order.get_user_current_email(),
-                    'email_type': OrderEventsEmails.ORDER.value},
-                type=OrderEvents.EMAIL_SENT.value)
+            send_order_confirmation.delay(order.pk, request.user.pk)
+            email_sent_event(
```
maarcingebala: This event is already created within `send_order_confirmation` task.

saleor/graphql/checkout/mutations.py
```diff
@@ -465,27 +463,22 @@ def perform_mutation(cls, _root, info, checkout_id):
         taxes = get_taxes_for_checkout(checkout, info.context.taxes)
         clean_checkout(checkout, taxes, info.context.discounts)
 
+        payment = checkout.get_last_active_payment()
+
         try:
+            # FIXME: we shouldn't create the order, but only do checks
```
maarcingebala: I think this is fine as it is - the same happens in the old storefront.

NyanKiyoshi: The issue is that if the payment fails, the order is placed anyway. Is this really the expected behavior?

maarcingebala: We have an issue for that: https://github.com/mirumee/saleor/issues/3877

maarcingebala: So we'll handle that anyway but we can do it in a separate PR.

maarcingebala:
🎉 Thanks @NyanKiyoshi, great effort!

Issue #3678, opened 2019-02-01
Use better data structure for exchanging data with payment gateways


maarcingebala:
Currently, all payment gateways return responses to Saleor using dicts. Since we know exactly what data Saleor expects from a gateway, we could use `dataclasses` or `namedtuples` for that.

Example of a dict response from Braintree: https://github.com/mirumee/saleor/blob/master/saleor/payment/gateways/braintree/__init__.py#L143-L153

This would make easier to implement new gateways if we had a common, typed structure of such response. Two things to discuss are:
- Which of those structures should we use? I haven't use dataclasses yet, they seem fine but are supported only in Python 3.7, while we support as well older versions.
- Where should we store the definitions of those types? Considering we want to extract gateways to separate libraries (#3422), maybe the best place would be an independent repository that every gateway could import from?

maarcingebala:
Also, after introducing those structures, `validate_gateway_response` function might be not needed anymore.

maarcingebala:
Related discussion in comments in a PR: https://github.com/mirumee/saleor/pull/3519#discussion_r247880724

maarcingebala:
We decided that we'd like to go with data classes as a data exchange format between payment gateways and Saleor. However for that to happen, we have to drop support for Python 3.5. Data classes are supported in Python 3.7 and for 3.6 there's a library that we can use: https://github.com/ericvsmith/dataclasses

We've considered various aspects that dropping Python 3.5 brings, but we agreed that the benefits of adding data classes are more important for Saleor. Also, most Python hosting platforms seem to support both Python 3.6 and 3.7.

PR #4028, opened 2019-04-25
Use dataclasses as a structure for payment's gateways

korycins:
#3678


saleor/payment/gateways/braintree/__init__.py
```diff
@@ -214,23 +221,26 @@ def refund(payment_information: Dict, connection_params: Dict) -> Dict:
...
 def process_payment(
-        payment_information: Dict, connection_params: Dict) -> Dict:
+        payment_information: Dict, connection_params: Dict) -> GatewayResponse:
     auth_resp = authorize(payment_information, connection_params)
-    if auth_resp['is_success']:
-        payment_information['token'] = auth_resp['transaction_id']
-        return [auth_resp, capture(payment_information, connection_params)]
-    return [auth_resp, void(payment_information, connection_params)]
+    if auth_resp.is_success:
+        print("SUKCESSS")
```
maarcingebala: 🎉

Pipfile
```diff
@@ -66,6 +66,7 @@ braintree = "==3.49.0"
 razorpay = ">=1.1.1"
 stripe = "*"
 "urllib3" = "*"
+dataclasses = "*"
```
jxltom: Kindly remind that requirements files should also be updated by ```pipenv lock -r > requirements.txt```

NyanKiyoshi: Travis kindly reminded him by failing 😉

saleor/payment/gateways/braintree/__init__.py
```diff
@@ -5,6 +5,7 @@
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.translation import pgettext_lazy
 
+from ...gateway_dataclasses import GatewayResponse
```
maarcingebala: I would think of some other name, I mean, we don't have to say in the filename that we use dataclasses dor that ;) What about `interface.py`? It would be a good name if we're going to extend it with some other common types in the future. But I'm open to suggestions.

korycins: I was going to ask for a better name 👍  Thanks @maarcingebala

NyanKiyoshi:
Don't forget that we have to update the README file and the Linux installation guide saying that the minimum version is 3.6.

Pipfile
```diff
@@ -65,6 +65,7 @@ braintree = "==3.49.0"
...
```
michaljelonek: Would it make sense to limit this to install only on 3.6? This package is not needed on 3.7 after all.

maarcingebala: Good idea

michaljelonek:
Also I would personally add removing support for 3.5 as a separate changelog entry. I realise you can deduce that by the word `dataclass`, but imo it's implied and not really that obvious, I think being explicit here for such change would be better.

Pipfile
NyanKiyoshi: Done during rebasing, thanks!

saleor/payment/interface.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@dataclass
+class PaymentInformation:
+    """Dataclass for storing all payment information"""
+
+    token: str
+    amount: Decimal
+    currency: str
+    billing: Optional[Dict[str, str]]
```
maarcingebala: Shouldn't we also have a data class to represent addresses? They're required by most gateways and even now we're using them as dicts, while we exactly know what fields are there.

saleor/payment/gateways/braintree/__init__.py
```diff
@@ -136,101 +139,111 @@ def authorize(payment_information: Dict, connection_params: Dict) -> Dict:
...
+) -> GatewayResponse:
     auth_resp = authorize(payment_information, connection_params)
-    if auth_resp['is_success']:
-        payment_information['token'] = auth_resp['transaction_id']
-        return [auth_resp, capture(payment_information, connection_params)]
-    return [auth_resp, void(payment_information, connection_params)]
+    if auth_resp.is_success:
+        payment_information.token = auth_resp.transaction_id
+        return capture(payment_information, connection_params)
+    return auth_resp
```
maarcingebala: We have to update our docs since the API has changed and now it returns a single object instead of a list: https://docs.getsaleor.com/en/latest/guides/payments.html?highlight=process_payment#process-payment-payment-information-connection-params

saleor/payment/interface.py
```diff
@@ -0,0 +1,30 @@
...
+    amount: Decimal
+    currency: str
+    transaction_id: str
+    error: Optional[str]
+    raw_response: Optional[Dict[str, str]] = None
+
+
+@dataclass
+class PaymentInformation:
+    """Dataclass for storing all payment information"""
```
maarcingebala: While this sentence is true, is also a bit vague in my opinion. I guess we should at least say what purpose this dataclass serves. From what I understand, it's used to represent a `Payment` model in a unified manner. Also, I'm wondering about the name as PaymentInformation is quite long. Maybe `PaymentData` would be better? It would somehow fit the "data" naming convention that is used in forms and now in our API as well. I'd like to hear some thoughts on this from others.

docs/guides/payments.rst
```diff
@@ -36,6 +36,14 @@ Example
         client_token = gateway.client_token.generate()
         return client_token
 
+
+.. note::
+
+    All the below methods receive payment_information as a dataclass - PaymentData.
```
NyanKiyoshi:
```suggestion
    All the below methods receive ``payment_information`` as a dataclass: ``PaymentData``. 
```

docs/guides/payments.rst
```diff
@@ -36,6 +36,14 @@ Example
         client_token = gateway.client_token.generate()
         return client_token
 
+
+.. note::
+
+    All the below methods receive payment_information as a dataclass - PaymentData. 
+    Methods should return a response as a dataclass - GatewayResponse.
```
NyanKiyoshi:
```suggestion
    Methods should return a response as a dataclass: ``GatewayResponse``. 
```

docs/guides/payments.rst
```diff
@@ -36,6 +36,14 @@ Example
         client_token = gateway.client_token.generate()
         return client_token
 
+
+.. note::
+
+    All the below methods receive payment_information as a dataclass - PaymentData. 
+    Methods should return a response as a dataclass - GatewayResponse. 
+    Description of the given structures can be found below.
```
NyanKiyoshi:
```suggestion
    The description of the given structures can be found below.
```

docs/guides/payments.rst
```diff
@@ -48,23 +56,23 @@ Example
...
-            'kind': 'auth',
-            'amount': response.amount,
-            'currency': response.currency,
-            'error': get_error(response),
-            'raw_response': get_payment_gateway_response(response),
-        }
+        return GatewayResponse(
+            is_success=response.is_success,
+            transaction_id=response.transaction.id,
+            kind='auth',
```
NyanKiyoshi: Should this be taken from an enum?

docs/guides/payments.rst
```diff
@@ -77,23 +85,23 @@ Example
...
-            'kind': 'refund',
-            'amount': response.amount,
-            'currency': response.currency,
-            'error': get_error(response),
-            'raw_response': get_payment_gateway_response(response),
-        }
+        return GatewayResponse(
+            is_success=response.is_success,
+            transaction_id=response.transaction.id,
+            kind='refund',
```
NyanKiyoshi: Same, enum?

docs/guides/payments.rst
```diff
@@ -197,83 +205,95 @@ Example
...
++-------------------------+-----------------+-----------------------------------------------------------------------------+
+| ``payment_information`` | ``PaymentData`` | Payment information, containing the token, amount, currency and billing.    |
++-------------------------+-----------------+-----------------------------------------------------------------------------+
+| ``connection_params``   | ``dict``        | List of parameters used for connecting to the payment's gateway.            |
++-------------------------+-----------------+-----------------------------------------------------------------------------+
 
-Example
-"""""""
+PaymentData
+"""""""""""""""""""""""
```
NyanKiyoshi:
```suggestion
"""""""""""
```

docs/guides/payments.rst
```diff
@@ -197,83 +205,95 @@ Example
...
++---------------------+-----------------+-----------------------------------------------------------------+
+| amount              | ``Decimal``     | Amount to be authorized/captured/charged/refunded.              |
++---------------------+-----------------+-----------------------------------------------------------------+
+| billing             | ``AddressData`` | Billing information.                                            |
++---------------------+-----------------+-----------------------------------------------------------------+
+| shipping            | ``AddressData`` | Shipping information.                                           |
++---------------------+-----------------+-----------------------------------------------------------------+
+| order_id            | ``int``         | Order id.                                                       |
++---------------------+-----------------+-----------------------------------------------------------------+
+| customer_ip_address | ``str``         | Ip address of the cusomter                                      |
```
NyanKiyoshi:
```suggestion
| customer_ip_address | ``str``         | IP address of the customer                                      |
```

docs/guides/payments.rst
```diff
@@ -197,83 +205,95 @@ Example
...
++---------------------+-----------------+-----------------------------------------------------------------+
+| customer_email      | ``str``         | Email address of the cusomter.                                  |
```
NyanKiyoshi:
```suggestion
| customer_email      | ``str``         | Email address of the customer.                                  |
```

docs/guides/payments.rst
```diff
@@ -197,83 +205,95 @@ Example
...
++---------------------+-----------------+-----------------------------------------------------------------+
+
+
+AddressData
+"""""""""""""""""""""""
```
NyanKiyoshi:
```suggestion
"""""""""""
```

docs/guides/payments.rst
```diff
@@ -197,83 +205,95 @@ Example
...
-Example
-"""""""
+PaymentData
+"""""""""""""""""""""""
 
-.. code-block:: python
++---------------------+-----------------+-----------------------------------------------------------------+
+| name                | type            | description                                                     |
++---------------------+-----------------+-----------------------------------------------------------------+
+| token               | ``str``         | Token used for transaction, provided by gateway.                |
```
NyanKiyoshi:
```suggestion
| token               | ``str``         | Token used for transaction, provided by the gateway.                |
```

docs/guides/payments.rst
```diff
@@ -197,83 +205,95 @@ Example
...
++----------------------+----------------------------+------------------------------------------------------------------------------------------------------------------------+
+| ``gateway_response`` | ``GatewayResponse``        | GatewayResponse containing details about every transaction, with ``is_success`` set to ``True`` if no error occurred.  |
++----------------------+----------------------------+------------------------------------------------------------------------------------------------------------------------+
+| ``client_token``     | ``str``                    | Unique client's token that will be used as his indentifier in the payment process.                                     |
++----------------------+----------------------------+------------------------------------------------------------------------------------------------------------------------+
 
 
-Gateway response fields
+GatewayResponse
 """""""""""""""""""""""
```
NyanKiyoshi:
```suggestion
"""""""""""""""
```

docs/guides/payments.rst
maarcingebala: About this change, let's use Grammarly to make sure about the grammar, it's available in the free plan.

korycins: I always check it with Grammarly

PR #4045, opened 2019-04-30
Remove all dedents

### LINKED ISSUE #4003: Problems with the newest Playground version
Recently we've bumped the Playground explorer to the newest version (#3971), but there are a few issues with it:
- new Playground has a polling feature which periodically sends a request to the server, I guess to keep the schema up-to-date. This makes the whole Playground run significantly slower. Polling can be disabled in Playground's settings: `schema.polling.enable: False`. I guess this should be disabled by default.
- Trying to view the "Schema" tab breaks the whole Playground for me. In the JS console I'm getting the following error when I just click the "Schema" button after opening the explorer, without any query:

```
Uncaught SyntaxError: Syntax Error: Unexpected Name "currency" (5050:3)
  5048 | 
  5049 |   
> 5050 |   currency: String!
       |   ^
  5051 | 
  5052 |   
  5053 |   gross: Money!
    at r (parser-graphql.js:1)
    at Object.parse (parser-graphql.js:1)
    at Object.parse$2 [as parse] (standalone.js:7310)
    at coreFormat (standalone.js:10511)
    at format (standalone.js:10684)
    at formatWithCursor (standalone.js:10696)
    at standalone.js:19300
    at Object.format (standalone.js:19308)
    at Object.i [as prettify] (utils.ts:22)
    at Object.a [as getSDL] (createSDL.ts:94)
```

salwator:
This PR removes unnecessary _dedent_ from schema docs, since we are using playground now.
Thanks to those changes, new playground stops crashing on schema polling.

Fixes #4003 

### Pull Request Checklist


saleor/graphql/account/mutations.py
```diff
@@ -617,7 +616,7 @@ class Arguments:
         )
 
     class Meta:
-        description = dedent(
+        description = (
```
maarcingebala: Brackets were left here on purpose? With multiline strings, we don't actually need them.

salwator: Good point, I was trying to check _fast_ if this fixes the problem.

PR #4094, opened 2019-05-09
Implement customer events

### LINKED ISSUE #4039: Implement customer events
Once we get #4018 merged, we will want to implement the following as customer events:

	Model changes
		- [ ] Add a CustomerEvent model and types

	New Customer Events
		- [ ] the order was placed
		- [ ] a note was added to the order
		- [ ] downloaded digital link

	New Staff Events
		- [ ] A customer was deleted

	GraphQL Changes
		- [ ] A countable object from the CustomerEvent model
		- [ ] A `events` field in `customer`

NyanKiyoshi:
Closes #4039.

### Checklist
- **Documentation**
  - [x] Document changes in a sub section "Customer Events"

- **Model changes**
	- [x] Add a CustomerEvent model and types

- **New Customer Events**
	- [x] the order was placed
	- [x] a note was added to the order
	- [x] downloaded digital link
	- [x] account was created
	- [x] password reset link sent
	- [x] email assigned
	- [x] note added
	- [x] name added => `first_name: str|none; last_name: str|none`

- **New Staff Events**
	- [x] A customer or customers were deleted (event with a `count` parameter)

- **GraphQL Changes**
	- [x] A countable object from the CustomerEvent model
	- [x] A `events` field in `customer`

- **Misc**
  - [x] Move incremental to product.utils.digital_download

### Pull Request Checklist


saleor/account/models.py
```diff
@@ -198,3 +202,26 @@ class CustomerNote(models.Model):
 
     class Meta:
         ordering = ("date",)
+
+
+class CustomerEvent(models.Model):
+    """Model used to store events that happened during the customer lifecycle."""
+
+    date = models.DateTimeField(default=now, editable=False)
+    type = models.CharField(
```
korycins: Maybe we could figure out a different name than built-in function name?

saleor/account/models.py
```diff
@@ -6,15 +6,19 @@
     BaseUserManager,
     PermissionsMixin,
 )
+from django.contrib.postgres.fields import JSONField
 from django.db import models
 from django.db.models import Q, Value
 from django.forms.models import model_to_dict
 from django.utils import timezone
+from django.utils.timezone import now
```
korycins: Maybe you can use `timezone.now` instead of direct `now` call, then this import will be not needed

saleor/core/utils/__init__.py
```diff
@@ -136,6 +135,8 @@ def serialize_decimal(obj):
 
 
 def create_superuser(credentials):
+    from ...account.models import User
```
korycins: Is it expected?

NyanKiyoshi: Yes, otherwise their would be a circular import. Utils files in saleor are always a cause of circular imports as they contain mostly core/heavy/important logic, we should look forward to refactoring them into proper files (utils files are often a bad practice).

saleor/account/models.py
NyanKiyoshi: Django's `timezone.now` looks at django settings for proper timezone info, whether it should use UTC or the local timezone.

saleor/account/models.py
NyanKiyoshi: We could refactor it to `event_type` but then we would lose consistency with `OrderEvent`. I believe it's fine to define `type` as column name here as it doesn't shadow with the builtin `type` but instead, is only caused an issue with GitHub's Python syntax highlight.

saleor/account/models.py
korycins: Yes, I know but I meant that you have two similar imports here.
```
from django.utils.timezone import now
```

NyanKiyoshi: Oh good catch! I may be blind! Thanks!

NyanKiyoshi:
Note it's still a WiP 😉 It was a draft at first because I didn't want travis to report because I didn't test the changes yet, then removed the draft flag as the changes were tested 👍

tests/test_checkout.py
```diff
@@ -551,6 +555,67 @@ def test_view_checkout_summary_remove_voucher(
...
+        discounts=None,
+        taxes=None,
+        user=customer_user if not is_anonymous_user else AnonymousUser(),
+    )
+
+    # Ensure only two events were created, and retrieve them
+    placement_event, email_sent_event = order.events.all()  # type: OrderEvent
+
+    # Ensure the correct order event was created
+    assert placement_event.type == OrderEvents.PLACED  # is the event the expected type
```
korycins: There is no need to create such big tests and so many asserts. You can create small unit tests for event functions - like `order_created_event`, `customer_added_to_note_order_event`, etc. Then you will be sure of the output from the function events. Then you can simply mock given event function and do `assert call`

tests/checks/events.py
```diff
@@ -0,0 +1,112 @@
+from typing import Dict, List, Optional
```
korycins: When you will create tests for each event function this file will be not needed anymore. Instead of doing asserts like in `was_digital_good_download_event_properly_generated` you can easily use mock and `assert with given arguments`

saleor/graphql/account/types.py
```diff
@@ -96,6 +100,54 @@ def resolve_is_default_billing_address(self, _info):
...
+    )
+    order_line = graphene.Field(
+        lazy_resolve("order.types:OrderLine"), description="The concerned order line."
+    )
+
+    class Meta:
+        description = "History log of the order."
+        model = models.CustomerEvent
+        interfaces = [relay.Node]
+        only_fields = ["id"]
```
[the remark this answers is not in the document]
NyanKiyoshi: They are already all defined above. So it should be fine.

Also, you made me notice an error on the description 👍

saleor/core/utils/__init__.py
```diff
@@ -136,6 +135,8 @@ def serialize_decimal(obj):
 
 
 def create_superuser(credentials):
```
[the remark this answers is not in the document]
NyanKiyoshi: Definitively.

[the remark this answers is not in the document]
NyanKiyoshi: Opened https://github.com/mirumee/saleor/issues/4106.

tests/checks/events.py
NyanKiyoshi: I kept the file, renamed it to `customer`, it checks the consistency between the API and the dashboard as we have two totally different logic for deleting customers. They don't process the same way, so they can't call a wrapped logic.

Only thing we can do, is mock them and assert the call.

saleor/graphql/account/types.py
```diff
@@ -107,14 +159,20 @@ class User(CountableDjangoObjectType):
     note = graphene.String(description="A note about the customer")
     orders = gql_optimizer.field(
         PrefetchingConnectionField(
-            "saleor.graphql.order.types.Order", description="List of user's orders."
+            lazy_resolve("order.types:Order"), description="List of user's orders."
```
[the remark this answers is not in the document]
NyanKiyoshi: It's easier to maintain if we ever get the graphql API into its own package. The idea is to isolate the graphql definitions from the rest of saleor. It just seems better to use relative imports (just as we do in django model definitions for lazy relations).

tests/api/test_account_events.py
```diff
@@ -0,0 +1,223 @@
...
+        elif isinstance(value, dict) and len(value) == 1:
+            _, value = value.popitem()
+        return super().__setattr__(key, value)
+
+
+@pytest.mark.parametrize(
+    "_test_name, event, expected_event_data",
+    [
+        (
+            "customer_placed_order_event",
```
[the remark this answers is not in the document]
NyanKiyoshi: Not easy to make something simple and easy to maintain... I have an idea on how to make it KISS, I will do that tomorrow.

My only trouble is how to make sure that every event is getting tested. What do you think about something like this:
```python

@pytest.fixture(scope="module", autouse=True)
def are_all_events_are_tested():
    tested_events = set()
    yield lambda event: tested_events.add(event)
    assert tested_events == CustomerEvents.CHOICES.keys()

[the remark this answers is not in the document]
Comment (pawelzar) [reply tests/api/test_account_events.py]: I don't have any nice solution right now but I'm not sure if there is a need for such verification. It's rather a matter of code review to check if there are proper tests for each necessary scenario. This fixture looks very cool and I guess it could work but it's a bit complex concept and may be hard to understand/maintain. Anyways, these tests seemed a bit too complicated at a first glance, so I just wanted to suggest a simplification as they especially should be easy to read. But I have to say your ideas are really awesome! I haven't seen such usage of a fixture before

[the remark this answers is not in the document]
Comment (maarcingebala) [reply tests/api/test_account_events.py]: I agree with @pawelzar - we don't want to introduce new patterns in code suddenly in one file, while all of the others use simple functions. If you're adding any new code and there are already some established patterns around that in the codebase, we have to follow them. Otherwise, we're just introducing inconsistency. Here we want to have just a simple unit functions for every case.

Some patterns might not work and we can think of improvements, but that has to be always discussed first and introduced consistently.

File: saleor/graphql/account/types.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (maarcingebala) [reply saleor/graphql/account/types.py]: Just to clarify - we're not going to make GraphQL API its own package. If anything, we'll separate the frontend parts to separate repos.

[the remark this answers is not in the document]
Comment (maarcingebala) [reply saleor/graphql/account/types.py]: Why can't we just have `"saleor.graphql.order.types.Order"`?

File: saleor/graphql/order/types.py [hunk 10]
```diff
@@ -52,7 +52,7 @@ class OrderEvent(CountableDjangoObjectType):
     )
 
     class Meta:
-        description = "History log of the order."
+        description = "History log of the customer."
```
NyanKiyoshi:
```suggestion
        description = "History log of the order."
```

saleor/graphql/account/types.py
```diff
@@ -96,6 +99,54 @@ def resolve_is_default_billing_address(self, _info):
...
+            "saleor.graphql.order.types.Order", description="The concerned order."
+        ),
+        model_field="order",
+    )
+    order_line = graphene.Field(
+        "saleor.graphql.order.types.OrderLine", description="The concerned order line."
+    )
+
+    class Meta:
+        description = "History log of the order."
```
NyanKiyoshi:
```suggestion
        description = "History log of the customer."
```

saleor/product/models.py
```diff
@@ -399,6 +400,18 @@ def save(
             self.token = str(uuid4()).replace("-", "")
         super().save(force_insert, force_update, using, update_fields)
 
+    def increment_download_count(self):
```
maarcingebala: I'd move it to a utility function to keep our models thin. It's okay for a model method to perform simple logic on model's fields but if it requires saving additional models, then we usually keep that in a function outside.

NyanKiyoshi: It's planned in the checklist! 👍  Thanks!

saleor/account/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+            pgettext_lazy(
+                "Event from a customer that got their account created",
+                "The account account was created",
+            ),
+        ),
+        (
+            PASSWORD_RESET_LINK_SENT,
+            pgettext_lazy(
+                "Event from a customer that got the password reset link sent by email",
+                "The password email reset link was sent",
```
maarcingebala: I guess on the designs it is more like:

```suggestion
                "Password reset link was sent to the customer",
```

docs/architecture/events.rst
```diff
@@ -6,6 +6,49 @@ Events
     when certain actions are completed, such us creating the order,
     cancelling fulfillment or completing a payment.
 
+
+Events Design
```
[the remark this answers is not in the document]
NyanKiyoshi: Thanks! A rebase that gone wrong probably.

docs/architecture/events.rst
```diff
@@ -6,6 +6,49 @@ Events
...
+
+Each package representing an entity (order, account, etc.) that generate events
+define both a model and a ``events.py`` file containing a set of functions
+having ``_event`` for suffix.
+
+Those functions are taking care of any logic and required fields for
+creating given events.
+
+
+Creating Events
```
maarcingebala: This section seems also duplicated.

PR #4096, opened 2019-05-09
Add black and flake8 linters to travis (tox)

NyanKiyoshi:
This adds CI checks for black and flake8 event when a PR is coming from a fork.

### Pull Request Checklist


patrys:
Not sure why we choose Python 3.6 over 3.7 for running linters.

NyanKiyoshi:
Thanks for the review 🎉 

I believe it would be better to run black as a python 3.6 language as it is the lowest version we support. Not sure if anything would change between python 3.6 and 3.7 code style though.

I just feel like we should we may want to use lowest version every time to refer it as the "base language".

.github/main.workflow
```diff
@@ -1,30 +0,0 @@
...
-
-action "Build" {
-  uses = "jefftriplett/python-actions@ba888adab5b57956dce82b7dfc06e87b65090a6d"
-  args = "pip install flake8 black"
-  env = {
-    VENV_DIR = "/opt/"
-  }
-}
-
-action "Black Linter" {
```
jxltom: I'm curious why this is removed, is this github action not good enough?

NyanKiyoshi: In another PR I said the following:
>After tests it doesn't work properly. GitHub is building on the latest base branch commit instead of the latest fork's branch commit.

So what happened is that GH Action does build on push/pull_request events but when it's coming from forks, it will build the base branch (master of the mirumee repo).

For now, I did not find any solution for that, so we will be using tox instead.

jxltom: Thanks for reply. It's working for PR but PR is from forks right? What do you mean by ```coming from forks```.

NyanKiyoshi: PR can also come from Mirumee members that create a PR through a branch in `mirumee/saleor`, those would get checked by the GH-Action. But not the people that create a PR with a branch coming from another repository.

jxltom: That's sad. That makes gh-actions quite unusable. 😢

NyanKiyoshi: There should be a way! Just didn't have time to dig it :) I am working on other actions and waiting from the GitHub support to give me access to it. So I will see from them.

PR #4102, opened 2019-05-13
Separate the legacy middleware from the GQL API middleware

### LINKED ISSUE #4042: Disable unneeded middlewares in the GraphQL API
Following #3631, we would like to disable any unneeded middleware. Some middlewares are actually needed sometimes but not everytime though, those should become lazy middlewares.

NyanKiyoshi:
Closes #4042.

- Disables unneeded middleware on the GQL API endpoint and for the 1.0 views;
- Makes heavy request attributes lazy.

### Pull Request Checklist


NyanKiyoshi:
The coverage loss is coming from disabling the Google Analytics middleware by default (when the API key is not provided). As it performs quite a few calls to get a value that end up not never being used (checked it).

saleor/settings.py
```diff
@@ -194,24 +194,64 @@ def get_bool_from_env(name, default_value):
...
+# Only used to dispatch requests to correct middlewares, please don't directly use it.
+MIDDLEWARE = ["saleor.core.middleware.middleware_dispatcher"]
+
+# Middleware for Saleor 1.0 (storefront and dashboard)
+DJANGO_MIDDLEWARE = [
+    # SessionMiddleware stores and manages session data from/to a cookie.
     "django.contrib.sessions.middleware.SessionMiddleware",
+    # SecurityMiddleware redirect the users to HTTPS when required/requested.
     "django.middleware.security.SecurityMiddleware",
+    # CommonMiddleware has for job to:
```
maarcingebala: I think most of these comments are unnecessary, they're documenting parts of Django or third-party libraries and should be already documented their docs. When it comes to comments related to our middlewares, they should be added as docstrings inside middlewares' definitions.

saleor/core/middleware.py
```diff
@@ -38,10 +104,13 @@ def discounts(get_response):
     """Assign active discounts to `request.discounts`."""
 
     def middleware(request):
-        discounts = Sale.objects.active(date.today()).prefetch_related(
-            "products", "categories", "collections"
+        request.discounts = SimpleLazyObject(
```
patrys: Since that's a queryset, what do we gain from making it even lazier?

saleor/core/middleware.py
```diff
@@ -15,9 +28,62 @@
...
 
 
+def django_only_middleware(middleware):
+    @wraps(middleware)
+    def wrapped(get_response):
+        api_path = reverse("api")
+        handler = middleware(get_response)
+
+        @wraps(handler)
+        def conditional(request):
```
patrys: Can we split this to a separate decorator and return `something(handler)`?

PR #4114, opened 2019-05-15
Storing the order lines PKs in events for clients to retrieve more data

### LINKED ISSUE #4112: Order events should return the line id
The order events should be able to return order and line objects in the resolvers.

NyanKiyoshi:
Closes #4112.

We are still storing the product name because in the case removing products from a draft, if the quantity is 0, it will delete the line. Thus, it won't be retrievable.

### Todo
- [x] Update docs
- [x] Add changelog entry

### Pull Request Checklist


saleor/graphql/order/types.py
```diff
@@ -46,9 +47,9 @@ class OrderEvent(CountableDjangoObjectType):
     oversold_items = graphene.List(
         graphene.String, description="List of oversold lines names."
     )
-    lines = graphene.List(OrderEventLineObject, description="The concerned lines.")
+    lines = graphene.List(OrderEventOrderLineObject, description="The concerned lines.")
     fulfilled_items = graphene.List(
-        OrderEventLineObject, description="The lines fulfilled."
+        lambda: FulfillmentLine, description="The lines fulfilled."
```
maarcingebala: Shouldn't this field be wrapped in `gql_optimizer`? I don't know if it would help us since in the resolver we do queryset filtering.

NyanKiyoshi: I have a hard time to see how it will optimize it without knowing what to look for?

maarcingebala: I mean this type is able to return something that requires a database query underneath (to fetch the fulfillment lines). I'm wondering if it wouldn't introduce N+1 problem in `line` or `fulfilledItems` fields when I run a query like: 
```
{
  order(id: "") {
    events {
      lines
      fulfilledItems
    }
  }
}
```

NyanKiyoshi: Oh okay, I see what you mean! 👍

PR #4125, opened 2019-05-17
Change the way address validation rules are being obtained

### LINKED ISSUE #3642: Expose address validation rules in API
API should expose address validation rules so that Storefront 2.0 could fetch them and perform first validation in JS. We could use that also to validate addresses in the offline mode, once the validation rules are fetched and cached.

https://github.com/mirumee/google-i18n-address#validation-rules

The exact shape of the API is to be discussed once we start working on that.

akjanik:
Closes #3642

### Pull Request Checklist


saleor/graphql/account/schema.py
```diff
@@ -44,7 +45,9 @@ class Meta:
 class AccountQueries(graphene.ObjectType):
     address_validator = graphene.Field(
         AddressValidationData,
-        input=graphene.Argument(AddressValidationInput, required=True),
+        country_code=graphene.Argument(CountryCodeEnum, required=False),
```
maarcingebala: Isn't this one actually required?

saleor/graphql/schema.graphql
```diff
@@ -2151,7 +2398,7 @@ type Query {
   checkouts(before: String, after: String, first: Int, last: Int): CheckoutCountableConnection
   checkoutLine(id: ID): CheckoutLine
   checkoutLines(before: String, after: String, first: Int, last: Int): CheckoutLineCountableConnection
-  addressValidator(input: AddressValidationInput!): AddressValidationData
+  addressValidator(countryCode: CountryCode, countryArea: String, cityArea: String): AddressValidationData
```
maarcingebala: I would rename it to `addressValidationRules` so it sounds more natural as a query name.

saleor/graphql/account/schema.py
akjanik: I thought the same, but looking at one of the tests, no, it is not
https://github.com/mirumee/saleor/pull/4125/files#diff-14900e464e1e3ba1cd78619e9d93f1e9R1348
Right now we are trying to resolve by IP if the country code is not provided.

saleor/graphql/schema.graphql
```diff
@@ -2151,7 +2398,7 @@ type Query {
   checkouts(before: String, after: String, first: Int, last: Int): CheckoutCountableConnection
   checkoutLine(id: ID): CheckoutLine
   checkoutLines(before: String, after: String, first: Int, last: Int): CheckoutLineCountableConnection
-  addressValidator(input: AddressValidationInput!): AddressValidationData
```
jxltom: Should it be deprecated instead of removed?

maarcingebala: I decided to just change the query in the schema since it was not used either in Dashboard 2.0 or the PWA storefront. In the long run, we'll be more strict about using deprecations when our API reaches a stable level. Right now we consider it still the beta version.

PR #4148, opened 2019-05-22
GraphQL now prints exceptions to stderr as well as returning them or not

NyanKiyoshi:
As a sysadmin/maintainer, I would like to be able to log exceptions occurring in production when executing the GraphQL API.

As a backend developer, I would like to be able to easily read in my console the errors occurring when working with GraphQL.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/views.py
+++ b/saleor/graphql/views.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import traceback
 
 from django.conf import settings
@@ -14,6 +15,8 @@ from graphql.error import (
     format_error as format_graphql_error,
 )
 from graphql.execution import ExecutionResult
+
+logger = logging.getLogger(__name__)
 
 
 class GraphQLView(View):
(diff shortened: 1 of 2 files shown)


saleor/graphql/views.py
```diff
@@ -190,10 +190,14 @@ def format_error(error):
             result = format_graphql_error(error)
         else:
             result = {"message": str(error)}
+
+        exc = error
+        while isinstance(exc, GraphQLError) and hasattr(exc, "original_error"):
+            exc = exc.original_error
+
+        traceback.print_exception(type(exc), exc, exc.__traceback__)
```
patrys: This should really be `logging.error(..., exc_info=exc)`. In production environments, no one can hear you scream.

NyanKiyoshi: Thanks! That's what I wanted to do, but it didn't work this time. I always though it was `logger.exception(exc)`.

It works in my other projects somehow... I will have to update that.

Thanks again.

patrys: `logger.exception` accepts a message as its first argument, it will cast whatever you pass it to a string. That's not where it gets the exception info from. Underneath it does something like this:

```python
import sys


def exception(self, msg, **kwargs):
    exc_info = sys.exc_info()
    return self.error(msg, **kwargs, exc_info=exc_info)
```

However `sys.exc_info()` only works if you're currently inside an `except:` block.

PR #4154, opened 2019-05-24
Checkout payment can fail without creating an order and deleting checkout

### LINKED ISSUE #3877: Alternative order creation flow
Current flow:

- user add products to cart
- fills shipment data
- clicks on 'create and pay' 
- order is created 
- user can pay for order

This flow is good in some applications (That's what for example Allegro doing), but in my projects it's not the best one - it creates many stale orders, reserve stock and block user from changing which products to buy.

Flow I implement in my projects:

- user add products to cart
- fills shipment data
- fills payment data
- if payment was successful - create order

Discussion:
- do we change default flow? Or if we even should
- can we make it confgurable?

NyanKiyoshi:
Closes #3877.

### Pull Request Checklist

### DIFF
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -22,7 +22,7 @@ from ..core.utils.promo_code import (
     promo_code_is_gift_card,
     promo_code_is_voucher,
 )
-from ..core.utils.taxes import ZERO_MONEY, get_taxes_for_country
+from ..core.utils.taxes import ZERO_MONEY, get_tax_rate_by_name, get_taxes_for_country
 from ..discount import VoucherType
 from ..discount.models import NotApplicable, Voucher
 from ..discount.utils import (
@@ -37,7 +37,8 @@ from ..giftcard.utils import (
 )
 from ..order import events
 from ..order.emails import send_order_confirmation
-from ..order.models import Order
+from ..order.models import Order, OrderLine
+from ..product.models import ProductVariant
 from ..shipping.models import ShippingMethod
 from . import AddressType, logger
 from .forms import (
@@ -46,7 +47,7 @@ from .forms import (
     AnonymousUserShippingForm,
     BillingAddressChoiceForm,
 )
-from .models import Checkout
+from .models import Checkout, CheckoutLine
 
 COOKIE_NAME = "checkout"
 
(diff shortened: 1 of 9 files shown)


maarcingebala:
The idea here was basically to change the flow of creating orders and payments in API to follow this logic in `checkoutComplete` mutation:
1. process payment,
2. if payment is successful, create an order.

But after looking at this code now, I see a few problems. `create_order` function runs some validations and can raise errors: `InsufficientStock` or `voucher_model.NotApplicable`. So before we can start processing the payment, we have to make sure that these validations were already run and we can safely charge the amount. After that, we should create order and delete checkout instance. Whole logic should be performed in one transaction.

saleor/checkout/utils.py
```diff
@@ -943,18 +967,51 @@ def create_order(checkout: Checkout, tracking_code: str, discounts, taxes, user:
...
+    Each order will get a private copy of both the billing and the shipping
+    address (if shipping).
+
+    If any of the addresses is new and the user is logged in the address
+    will also get saved to that user's address book.
+
+    Current user's language is saved in the order so we can later determine
+    which language to use when sending email.
+    """
+    from saleor.product.utils import allocate_stock
```
maarcingebala: Absolute import.

django-queries:
Here is the report for 1a9308bb295720fa7a90d596dbe2f2889539beb8 (NyanKiyoshi/saleor @ gql/checkout/payment)
Missing base report (5b2a3f93419aa3b704a7f85c963ca0df7e778a4f). The results couldn't be compared.

```diff
# api.benchmark checkout
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  add billing address to checkout     	         41	         41
  add shipping to checkout            	          7	          7
  checkout payment charge             	         16	         16
  complete checkout                   	          6	          6
  create checkout                     	         45	         45

# api.benchmark homepage
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve main menu                  	          5	          5
  retrieve product list               	          4	          4
  retrieve secondary menu             	          5	          5
  retrieve shop                       	          2	          2

# api.benchmark product
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  product details                     	         16	         16

# api.benchmark variant
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve variant list               	         13	         13
```

Issue #4134, opened 2019-05-21
Setting "price" or "price override" field raises an error


maarcingebala:
### Steps to reproduce the problem
1. Set a product price or variant's "price override" field to zero (either in Dashboard 1.0 or 2.0).
2. Go to product details page (both dashboards).
3. An error is raised:

```
"Traceback (most recent call last):",
            "  File \"/Users/marcin/.pyenv/versions/saleor3.7/lib/python3.7/site-packages/promise/promise.py\", line 487, in _resolve_from_executor",
            "    executor(resolve, reject)",
            "  File \"/Users/marcin/.pyenv/versions/saleor3.7/lib/python3.7/site-packages/promise/promise.py\", line 754, in executor",
            "    return resolve(f(*args, **kwargs))",
            "  File \"/Users/marcin/.pyenv/versions/saleor3.7/lib/python3.7/site-packages/graphql/execution/middleware.py\", line 76, in make_it_promise",
            "    return next(*args, **kwargs)",
            "  File \"/Users/marcin/.pyenv/versions/saleor3.7/lib/python3.7/site-packages/graphql_jwt/decorators.py\", line 31, in wrapper",
            "    return func(info.context, *args, **kwargs)",
            "  File \"/Users/marcin/.pyenv/versions/saleor3.7/lib/python3.7/site-packages/graphql_jwt/decorators.py\", line 42, in wrapper",
            "    return f(*args, **kwargs)",
            "  File \"/Users/marcin/mirumee/saleor/saleor/graphql/product/types/products.py\", line 480, in resolve_margin",
            "    _, margin = get_product_costs_data(self)",
            "  File \"/Users/marcin/mirumee/saleor/saleor/product/utils/costs.py\", line 14, in get_product_costs_data",
            "    costs_data = get_cost_data_from_variants(variants)",
            "  File \"/Users/marcin/mirumee/saleor/saleor/product/utils/costs.py\", line 34, in get_cost_data_from_variants",
            "    costs_data = get_variant_costs_data(variant)",
            "  File \"/Users/marcin/mirumee/saleor/saleor/product/utils/costs.py\", line 44, in get_variant_costs_data",
            "    margin = get_margin_for_variant(variant)",
            "  File \"/Users/marcin/mirumee/saleor/saleor/product/utils/costs.py\", line 61, in get_margin_for_variant",
            "    percent = round((margin / base_price) * 100, 0)",
[...]

PR #4170, opened 2019-05-28
Fixed division by zero when setting "price" or "price override"

MahmoudRizk:
I want to merge this change because fix bug #4134

### Pull Request Checklist

### DIFF
--- a/saleor/product/utils/costs.py
+++ b/saleor/product/utils/costs.py
@@ -57,10 +57,8 @@ def get_margin_for_variant(variant):
     if variant.cost_price is None:
         return None
     base_price = variant.base_price
+    if not base_price:
+        return None
     margin = base_price - variant.cost_price
-    try:
-        percent = round((margin / base_price) * 100, 0)
-    except ZeroDivisionError:
-        sign = lambda margin: (1, -1)[margin.amount < 0]
-        percent = 9999999999999999999999 * sign(margin)
+    percent = round((margin / base_price) * 100, 0)
     return percent
--- a/saleor/product/utils/costs.py
+++ b/saleor/product/utils/costs.py
@@ -58,5 +58,9 @@ def get_margin_for_variant(variant):
         return None
     base_price = variant.base_price
     margin = base_price - variant.cost_price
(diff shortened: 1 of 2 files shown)


saleor/product/utils/costs.py
```diff
@@ -58,5 +58,9 @@ def get_margin_for_variant(variant):
         return None
     base_price = variant.base_price
     margin = base_price - variant.cost_price
-    percent = round((margin / base_price) * 100, 0)
+    try:
+        percent = round((margin / base_price) * 100, 0)
+    except ZeroDivisionError:
+        sign = lambda margin: (1, -1)[margin.amount < 0]
+        percent = 9999999999999999999999 * sign(margin)
```
MahmoudRizk: since division any number by zero equals infinity, so 99999..99 here is to simulate infinity.

saleor/product/utils/costs.py
```diff
@@ -58,5 +58,9 @@ def get_margin_for_variant(variant):
         return None
     base_price = variant.base_price
     margin = base_price - variant.cost_price
-    percent = round((margin / base_price) * 100, 0)
+    try:
```
maarcingebala: Hey @MahmoudRizk, thanks for opening the PR! I think it would make the most sense if we returned `None` in that case, similarly as we do if `cost_price` is also undefined. So the only change I would make is this:

```
base_price = variant.base_price
if not base_price:
    return None
```

maarcingebala:
Looks good now! There is one last thing to take care of - could you add a test for that function, that would check that `None` is returned when product's or variant's price is zero? Unfortunately, I don't see any test for that function so you could add it in `tests/test_product.py`. Let me know if you can take care of that!

MahmoudRizk:
> Looks good now! There is one last thing to take care of - could you add a test for that function, that would check that `None` is returned when product's or variant's price is zero? Unfortunately, I don't see any test for that function so you could add it in `tests/test_product.py`. Let me know if you can take care of that!

Ok, I will make it.

maarcingebala:
Thanks for your contribution to Saleor @MahmoudRizk!

PR #4188, opened 2019-05-30
Disable polling in Playgroud by default

maarcingebala:
The "polling" feature of new Playground makes it really slow on a local machine, so I suggest that we disable it by default and users will re-enable it manually if they want to use it.


maarcingebala:
@korycins I think there is no need, similarly as we don't say that you can switch between dark/light modes in Playground ;)

PR #4192, opened 2019-05-30
Fix using discounts in subtotal calculation in API

maarcingebala:
This PR fixes checkout's subtotal calculation in API as it wasn't using discounts properly.

### Pull Request Checklist


tests/api/test_checkout.py
```diff
@@ -1452,3 +1452,58 @@ def test_checkout_shipping_address_update_with_not_applicable_voucher(
 
     assert checkout_with_item.shipping_address.country == new_address["country"]
     assert checkout_with_item.voucher_code is None
+
+
+def test_checkout_totals_use_discounts(api_client, checkout_with_item, sale):
+    query = """
+    query getCheckout($token: UUID) {
```
korycins: I think that you should test one thing per one unit test. 
If this test fails you will get information that something is wrong with `total` or `subtotal` or `line_total` but you won't get explicit information what went wrong. And before you will start fixing the changes that have broken this test you will need to investigate what price was calculated wrong.

tests/api/test_checkout.py
```diff
@@ -1452,3 +1452,58 @@ def test_checkout_shipping_address_update_with_not_applicable_voucher(
...
+    checkout = checkout_with_item
+    variables = {"token": str(checkout.token)}
+    response = api_client.post_graphql(query, variables)
+    content = get_graphql_content(response)
+    data = content["data"]["checkout"]
+
+    discounts = [sale]
+
+    # make sure that we're testing a variant that is actually on sale
+    variant = checkout.lines.first().variant
```
korycins: It works because sale fixture gets the same category as product fixture, and `checkout_with_item` fixture gets product fixture. All relations are creating inside a fixture. maybe we can explicitly show the most important connection?. Instead of doing assert ` == variant.get_price(discounts=discounts).gross.amount + sale.value` you can simply add the product to `Sale` object inside a test.

maarcingebala: Right, that's a better way to do it, I'll change that.

tests/api/test_checkout.py
maarcingebala: The intent of this test is to check if discounts were correctly included in all checkout price fields. I think it's clear enough to now which part failed, e.g.: 
```
>       assert (
            data["subtotalPrice"]["gross"]["amount"]
            == checkout.get_subtotal(discounts=discounts).gross.amount
        )
E       AssertionError: assert 30.0 == Decimal('15.00')
```

Issue #4076, opened 2019-05-07
Data for search engine feeds is missing brand attribute


koradon:
### What I'm trying to achieve
Generate proper structured data for search engines.

### Steps to reproduce the problem
1. Add Attribute `Brand` to product type.
2. Add value to some product.
3. Inspect generated data by `product_json_ld` on the product detail page.

### What I expected to happen
Structured data should contain data like this `"brand": {"@type": "Thing", "name": "Some Brand Name"}`

Right now we have a bug in `get_brand_from_attributes` method. Right there we check if the name of an attribute is `brand` or `publisher`. Not only those values are hardcoded and hard to find but also they are checking if the name is lowercase which is specific to the implementation.
So we could change that logic to check attribute slug (which is more robust) and also add the ability to configure attributes that are describing this field in some settings. For example, I could use the attribute `Manufacturer` or `made by`. 
Also adding this attribute to populatedb could be very useful.

PR #4194, opened 2019-05-31
fixing issue 4076

MahmoudRizk:
Suggested solution for issue #4076.

I suggest to create a separate model linked to **Attributes** called **AttributesCategories** which will be used in <code>get_brand_from_attributes</code> function. The **AttributesCategories** will have the values from [schema.org](https://schema.org) types and will be used instead of the attributes name. I think this will make the <code>get_brand_from_attributes</code> function more generic and will return other attributes not only **brand** and **publisher**.

### Screenshots
aaa

### Pull Request Checklist


maarcingebala:
Hey @MahmoudRizk, thanks for opening the PR! I think there should be more straightforward ways to solve the issue .e.g we could add a new field to the product model or create a "system attribute" that is protected from removing. We haven't decided yet how to tackle it, but a good practice is first to discuss the potential solution in the issue. We need really good reasons to introduce a new model and here I think it's not necessary.

PR #4195, opened 2019-05-31
Add ability to store payment source

### LINKED ISSUE #2700: Ability to store credit card for further usage
Most of the payment providers allow storing tokenized credit card data in a secure way for further transactions.
So after the first transaction, the customer doesn't have to enter credit card details again but just selects stored card from a list.

- [ ] Adapt payment gateways interface to support storing credit cards
- [ ] Implement flag for storing card data in payment method endpoint
- [ ] Add an endpoint that will return list of stored payment methods for the customer

We should be able to store the credit card data in such way we can retrieve them later (e.g. token) but never as plain.

salwator:
Adds ability to store credit card information on payment.

Closes #2700 

This PR introduces:
- new private meta field for user to store external data, ex. gateway information
- new interface for extending parameters for client token generation, ex. braintree token for existing customer
- new flag in CheckoutComplete mutation to rerquest saving user payment source in gateway (needs to be enabled in gateway settings too)
- new flow for charging existing customer instead of one-time-token each time, implemented in Braintree gateway
- improvement in testing, most sensitive tests in braintree gateway test suite are now gone and replaced with VCR-based containing real life sandbox communication
- new gateway method for retrieving customer stored payment sources (credit cards)
- fixes bug with billing and shipping addresses causing _call_gateway_ to fail

### Pull Request Checklist


NyanKiyoshi:
Closes #2700?

django-queries:
Here is the report for 41051fcaa4038856e1c3c76463d952a7f7b67848 (mirumee/saleor @ store-credit-card-info)
Missing base report (bb6813c24c07b2f6004467d0f3bbf6108ab4a38a). The results couldn't be compared.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  add billing address to checkout     	         41	         41
  add shipping to checkout            	          7	          7
  checkout payment charge             	         16	         16
  complete checkout                   	          6	          6
  create checkout                     	         45	         45

# api.benchmark homepage
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve main menu                  	          5	          5
  retrieve product list               	          4	          4
  retrieve secondary menu             	          5	          5
  retrieve shop                       	          2	          2

# api.benchmark product
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  product details                     	         16	         16

# api.benchmark variant
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve variant list               	         13	         13
```

</p>
</details>

saleor/payment/gateways/braintree/__init__.py
```diff
@@ -32,6 +32,7 @@
 
 
 def get_customer_data(payment_information: PaymentData) -> Dict:
+    ''' Provides customer info, use only for new customer creation '''
```
NyanKiyoshi: Our style is like this:

```suggestion
    """Provides customer info, use only for new customer creation."""
```

saleor/payment/gateways/braintree/__init__.py
```diff
@@ -2,11 +2,19 @@
 from typing import Dict, List
 
 import braintree as braintree_sdk
+from autopep8 import extract_code_from_function
```
NyanKiyoshi: Leftover?

saleor/payment/utils.py
```diff
@@ -495,3 +511,25 @@ def gateway_refund(payment, amount: Decimal = None) -> Transaction:
 
     _gateway_postprocess(transaction, payment)
     return transaction
+
+
+GATEWAYS_LABEL = "gateways"
+
+
+def extract_id_for_payment_gateway(user, gateway):
+    gateway_meta = user.get_private_meta(label=GATEWAYS_LABEL)
```
korycins: How about:
```
    def get_private_meta(self, label):
         if label not in self.private_meta:
             return {}
         return self.private_meta[label]
OR 
    def get_private_meta(self, label):
         return self.private_meta.get(label, {})
```
And then you don't need to use so many `if`s
```
    gateway_config = gateway_meta.get(gateway)
    if not gateway_config;
        None
    return gateway_config.get("customer_id")

File: saleor/settings.py [hunk 4]
```diff
@@ -599,6 +600,7 @@ def get_host():
         "module": "saleor.payment.gateways.braintree",
         "config": {
             "auto_capture": True,
+            "store_card": False,
```
korycins: I assume that this value should be taken from env first. (Same as `auto_capture` btw)

saleor/graphql/payment/resolvers.py
```diff
@@ -1,8 +1,16 @@
+from itertools import chain
+
 import graphene_django_optimizer as gql_optimizer
 
+from saleor.payment import get_payment_gateway
```
NyanKiyoshi: Absolute import 👍

saleor/graphql/payment/resolvers.py
```diff
@@ -16,6 +24,26 @@ def resolve_payments(info, query):
...
 def resolve_payment_client_token(gateway=None, user=None):
     token_config = TokenConfig()
-    if user:
+    if user is not None:
         token_config.customer_id = extract_id_for_payment_gateway(user, gateway)
     return gateway_get_client_token(gateway, token_config)
+
+
+def resolve_payment_sources(user):
+    def retrieve_sources(gateway_name, customer_id):
```
NyanKiyoshi: I would think it would be better to put it outside of the function, with a more meaningful name. It seems like the logic could be used by other functions. Maybe but it private?

saleor/payment/utils.py
```diff
@@ -495,3 +515,28 @@ def gateway_refund(payment, amount: Decimal = None) -> Transaction:
...
+    return gateway_config["customer_id"]
+
+
+def store_id_for_payment_gateway(user, gateway, customer_id):
+    user.store_private_meta(
+        label=GATEWAYS_LABEL,
+        key=gateway.strip().upper(),
+        value={"customer_id": customer_id},
+    )
+    user.save()
```
NyanKiyoshi: Maybe we should only save the `private_meta` field?

tests/api/test_payment.py
```diff
@@ -1,15 +1,21 @@
 from unittest.mock import patch
 
 import graphene
+import pytest
+from tests.api.utils import get_graphql_content
```
NyanKiyoshi: Relative import needed here

tests/api/test_payment.py
```diff
@@ -454,19 +460,120 @@ def test_query_payments(payment_dummy, permission_manage_orders, staff_api_clien
...
+    assert content["data"]["paymentClientToken"] == TOKEN
+
+
+@pytest.fixture
+def enable_braintree(settings):
+    settings.CHECKOUT_PAYMENT_GATEWAYS = {"braintree": None}
+    return settings
+
+
+def test_list_payment_sources(
```
NyanKiyoshi: You might want to check that it doesn't work for anonymous users and that a user cannot access other users' data.

tests/gateway/test_braintree.py
```diff
@@ -263,250 +255,162 @@ def test_authorize_error_response(
...
-
-    payment_info = create_payment_information(payment, payment_token)
-    with pytest.raises(BraintreeException) as e:
-        authorize(payment_info, gateway_config)
-    assert str(e.value) == DEFAULT_ERROR_MESSAGE
+@pytest.fixture
+def sandbox_braintree_gateway_config(gateway_config):
+    """ set up your environment variables to record sandbox """
+    gateway_config.connection_params = {
+        "merchant_id": os.getenv("BRAINTREE_MERCHANT_ID", "9m6qhfxsqzm3cgzw"),
```
NyanKiyoshi: Why trying to get them from the environment?

salwator: To integrate gateway you need to set those up either way so it is probably best way to leave it to env for next person to record cassettes. We *surely* do not want those in code.

tests/api/test_payment.py
salwator: Separate test for resolver decorator seems excessive, query fetches data for logged user so checking for other users data is pointless and even is possible does not provide enough information to do anything useful with it.

tests/api/test_payment.py
salwator: Those are tests and this util is imported everywhere like this.

saleor/graphql/payment/resolvers.py
salwator: Moved to utils

saleor/account/models.py
```diff
@@ -187,6 +189,16 @@ def get_ajax_label(self):
             return "%s %s (%s)" % (address.first_name, address.last_name, self.email)
         return self.email
 
+    def get_private_meta(self, label):
+        if label not in self.private_meta:
+            return None
+        return self.private_meta[label]
+
+    def store_private_meta(self, label, key, value):
+        if label not in self.private_meta:
```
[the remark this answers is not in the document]
salwator: Sure, but that won't store anything if label is not present 😄

saleor/payment/utils.py
salwator: I kind of dislike passing valid but empty structures when no value is really found, but in this case it probably makes more sense and is more practical.
Changed.

saleor/account/models.py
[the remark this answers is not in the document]
NyanKiyoshi: True... `.setdefault(label, {})` then 😆

saleor/settings.py
```diff
@@ -620,8 +620,8 @@ def get_host():
     RAZORPAY: {
         "module": "saleor.payment.gateways.razorpay",
         "config": {
-            "store_card": False,
-            "auto_capture": None,
+            "store_card": get_bool_from_env("RAZORAY_STORE_CARD", False),
+            "auto_capture": get_bool_from_env("RAZORPAY_AUTO_CAPTURE", None),
```
NyanKiyoshi: I don't think we should provide it for razorpay as it doesn't support authorization. It will always auto capture.

saleor/settings.py
```diff
@@ -620,8 +620,8 @@ def get_host():
     RAZORPAY: {
         "module": "saleor.payment.gateways.razorpay",
         "config": {
-            "store_card": False,
-            "auto_capture": None,
+            "store_card": get_bool_from_env("RAZORAY_STORE_CARD", False),
```
NyanKiyoshi: I'm not sure about this. We have no way of telling through the API or any other way the credit card information.

tests/gateway/test_braintree.py
```diff
@@ -263,250 +255,164 @@ def test_authorize_error_response(
...
+        "public_key": os.getenv("BRAINTREE_PUBLIC_KEY", "fake_public_key"),
+        "private_key": os.getenv("BRAINTREE_PRIVATE_KEY", "fake_private_key"),
```
korycins: I don't think that taking them from `env` is the best solution.  Someone can have these variables already set in his env and tests will fail or will generate strange behavior.

tests/api/test_payment.py
```diff
@@ -459,19 +464,121 @@ def test_query_payments(payment_dummy, permission_manage_orders, staff_api_clien
     assert payment_ids == [payment_id]
 
 
-@patch("saleor.graphql.payment.resolvers.gateway_get_client_token")
-def test_query_payment_client_token(mock_get_client_token, user_api_client):
+@pytest.fixture
```
korycins: Shouldn't be at the beginning of the file?

tests/api/test_payment.py
```diff
@@ -459,19 +464,121 @@ def test_query_payments(payment_dummy, permission_manage_orders, staff_api_clien
...
     token = content["data"]["paymentClientToken"]
     assert token == example_token
+
+
+@pytest.fixture
+def token_config_with_customer(braintree_customer_id):
+    return TokenConfig(customer_id=braintree_customer_id)
+
+
+@pytest.fixture
```
korycins: Shouldn't be at the beginning of the file?

tests/api/test_payment.py
```diff
@@ -459,19 +464,121 @@ def test_query_payments(payment_dummy, permission_manage_orders, staff_api_clien
...
     response = user_api_client.post_graphql(query, variables)
 
     content = get_graphql_content(response)
-    assert mock_get_client_token.called_once_with(PaymentGatewayEnum.BRAINTREE.name)
+    mock_get_client_token.assert_called_once_with("braintree")
     token = content["data"]["paymentClientToken"]
     assert token == example_token
+
+
+@pytest.fixture
```
korycins: Shouldn't be at the beginning of the file?

saleor/payment/gateways/braintree/__init__.py
```diff
@@ -234,3 +276,27 @@ def process_payment(
 ) -> GatewayResponse:
     auth_resp = authorize(payment_information, config)
     return auth_resp
+
+
+def list_client_sources(
+    config: GatewayConfig, customer_id: str
```
korycins: You should keep consistency between functions' params. If you introduce new dataclass then you should use it to pass `customer_id`, instead of direct passing `customer_id`. This will help us with keeping all generic method in similar design.

saleor/graphql/payment/resolvers.py
```diff
@@ -13,5 +21,42 @@ def resolve_payments(info, query):
...
     return gateway_get_client_token(gateway)
+
+
+def resolve_payment_sources(user):
+    stored_customer_accounts = {
+        gateway: extract_id_for_payment_gateway(user, gateway)
+        for gateway in list_enabled_gateways()
+    }
+    return list(
+        chain(
```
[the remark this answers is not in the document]
salwator: One additional level added for type conversion added unintentional complexity lately.
Moved it and unnested comprehensions.

saleor/graphql/payment/schema.py
```diff
@@ -23,7 +30,13 @@ def resolve_payment(self, info, **data):
...
         return resolve_payments(info, query)
 
-    def resolve_payment_client_token(self, _info, gateway=None):
+    @login_required
+    def resolve_payment_stored_sources(self, info):
+        return resolve_payment_sources(info.context.user)
+
+    def resolve_payment_client_token(self, info, gateway=None):
+        if info.context.user.is_authenticated:
+            return resolve_payment_client_token(gateway, user=info.context.user)
```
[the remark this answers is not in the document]
salwator: Sure, fixed.

saleor/graphql/payment/resolvers.py
```diff
@@ -13,5 +21,42 @@ def resolve_payments(info, query):
     return gql_optimizer.query(queryset, info)
 
 
-def resolve_payment_client_token(gateway=None):
+def resolve_payment_client_token(gateway=None, user=None):
+    if user is not None:
```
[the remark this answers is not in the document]
salwator: Is it really a problem? Resulting loc count is the same. I know it is minor issue, not worth the trouble, but maybe we can have some clarifying discussion here, cause I see some common thread in your review 😉 
It is probably the matter of programming taste, but I virtually always refactor in the opposite direction! I consider mutating *anything* a penalty in itself and try to remove it so things are more declarative and less sequenced. There is incredibly long list of reasons for that but you could just boil it down to: functional style is more future proof, easier to reason, refactor and debug.
There is also typing issue - after refactoring _token_config_ is _de facto_ union of NoneType and TokenConfig, not simple dataclass.
It is better to have one return call instead of two, obviously, not sure if it something to micro-optimize for.

saleor/graphql/payment/resolvers.py
```diff
@@ -13,5 +21,42 @@ def resolve_payments(info, query):
...
+        customer_id = extract_id_for_payment_gateway(user, gateway)
```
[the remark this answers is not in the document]
salwator: In _call_gateway_ we fill payment information:
`store_source = (
        extra_params.pop("store_source", False) and gateway_config.store_customer
    )
    payment_information = create_payment_information(
        payment, payment_token, store_source=store_source, **extra_params
    )`

saleor/graphql/payment/resolvers.py
```diff
@@ -13,5 +21,42 @@ def resolve_payments(info, query):
...
+        if customer_id:
+            token_config = TokenConfig(customer_id=customer_id)
+            return gateway_get_client_token(gateway, token_config)
```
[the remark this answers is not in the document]
salwator: Not yet, as it requires separate gateway method to exist.
@mateuszkula is creating issue for that and we should make it happen soon.

saleor/payment/utils.py
```diff
@@ -495,3 +516,31 @@ def gateway_refund(payment, amount: Decimal = None) -> Transaction:
 
     _gateway_postprocess(transaction, payment)
     return transaction
+
+
+GATEWAYS_LABEL = "gateways"
```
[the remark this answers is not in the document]
salwator: Is it used by code above?

tests/api/test_payment.py
salwator: Is it used by test above?

saleor/settings.py
salwator: We should keep consistent configuration across gateways.

saleor/settings.py
salwator: Could you please elaborate?

saleor/payment/gateways/braintree/__init__.py
salwator: There is no new dataclass here, GatewayConfig is not a place for customer_id.
Id for customer is precisely what every gateway will need to fetch stored sources, not sure what is more generic than that?

tests/gateway/test_braintree.py
salwator: I tried to prevent people (including myself) from committing credentials by accident but your scenario is troublesome indeed. I'll change it.

saleor/settings.py
NyanKiyoshi: There razorpay payment provider is JS based and will not give us any information about the payment apart from the payment method, e.g. "card"

saleor/payment/utils.py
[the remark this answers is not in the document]
korycins: No, it isn't.  But I think that all constants variable should be at least in one place. The first place where I would look for constants variables would be the beginning of the file.
Also - We should keep the consistency with all saleor's modules. and I've never seen the constants variables declaring in the middle of the file. Even if the constant variable is using at the end of the file.
I 've made grep for all saleor's sources and I haven't seen any file with that approach. If you dislike it then we can wait for Marcin to discuss it also.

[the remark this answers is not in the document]
salwator: A general rule of clean code is to keep things close to usage so there is as little need to jump around in files. It is both easier to read and easier to extract from module all together if needed. I really do dislike huge piles of fixtures and constants at the top of a test suite module.

saleor/payment/models.py
```diff
@@ -210,6 +210,7 @@ class Transaction(models.Model):
         max_length=256,
         null=True,
     )
+    customer_id = models.CharField(max_length=256, null=True)
```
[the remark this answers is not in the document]
salwator: Good catch, it is a leftover. 😳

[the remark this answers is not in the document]
salwator: In the end, it appears to me essential for pushing data back from gateway to business logic in saleor, can't remove it without major changes in call_gateway.

saleor/settings.py
salwator: We only need _customer_id_ and I found sth like [this](https://razorpay.com/docs/payment-methods/cards/saved-cards/).
It seems that we might reuse source from razorpay from my very quick research.

saleor/payment/interface.py
```diff
@@ -39,14 +40,23 @@ class PaymentData:
...
+    order_id: int
     customer_ip_address: str
     customer_email: str
+    token: Optional[str] = None
+    customer_id: Optional[str] = None
+    reuse_source: bool = False
+
+
+@dataclass
+class TokenConfig:
```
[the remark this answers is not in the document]
salwator: I will try to generalize this interface while working on new stripe integration.

saleor/payment/utils.py
```diff
@@ -227,17 +237,26 @@ def create_transaction(
...
     """Gets client token, that will be used as a customer's identificator for
     client-side tokenization of the chosen payment method.
     """
+    if not token_config:
+        token_config = TokenConfig()
+    gateway, gateway_config = get_payment_gateway(gateway_name)
+    return gateway.get_client_token(config=gateway_config, token_config=token_config)
+
+
+def gateway_list_client_sources(gateway_name: str, customer_id: str):
```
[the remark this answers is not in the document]
salwator: One is proxy method to avoid retrieving gateway in resolver, good for testing and for no magic machinery in business logic.

saleor/payment/utils.py
```diff
@@ -495,3 +517,28 @@ def gateway_refund(payment, amount: Decimal = None) -> Transaction:
 
     _gateway_postprocess(transaction, payment)
     return transaction
+
+
+def extract_id_for_payment_gateway(user, gateway):
+    key = prepare_label_name(gateway)
```
maarcingebala: Shouldn't this live inside getters/setters for meta? At this level, I only want to get a value or null and not care about how keys are internally stored.

maarcingebala: Question is whether we have any requirements for the format of the keys, e.g. that they have to be always stripped and uppercase.

saleor/payment/utils.py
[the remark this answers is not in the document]
maarcingebala: Right, seems like we should have only one function for that.

saleor/payment/interface.py
```diff
@@ -60,3 +70,23 @@ class GatewayConfig:
...
     # a unified structure
     connection_params: Dict[str, Any]
+    store_customer: bool = False
+
+
+@dataclass
+class CreditCardInfo:
+    """Uniform way to represent Credit Card information"""
+
+    last4: str
```
maarcingebala: Not `last_4`? Braintree also seems to write it like that.

saleor/payment/interface.py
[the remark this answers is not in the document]
maarcingebala: Maybe it's too early to introduce something like the entire config object for tokens since it only uses `customer_id`?

saleor/payment/utils.py
```diff
@@ -495,3 +517,28 @@ def gateway_refund(payment, amount: Decimal = None) -> Transaction:
...
+    gateway_config = user.get_private_meta(label=GATEWAYS_LABEL).get(key, {})
+    return gateway_config.get("customer_id", None)
+
+
+def store_id_for_payment_gateway(user, gateway, customer_id):
```
maarcingebala: This function could use some docstring - when you look at the usage of it in a mutation it's hard to guess what ID it stores. Or shouldn't it be called `store_customer_id`?

saleor/account/models.py
```diff
@@ -151,6 +151,8 @@ class User(PermissionsMixin, AbstractBaseUser):
     )
     avatar = VersatileImageField(upload_to="user-avatars", blank=True, null=True)
 
+    private_meta = JSONField(blank=True, default=dict, encoder=CustomJsonEncoder)
```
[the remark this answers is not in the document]
salwator: @maarcingebala fully agree, that's my plan for future to refactor models into some useful mixin.
Better to do in separate PR, I think.

saleor/payment/gateways/braintree/__init__.py
```diff
@@ -234,3 +276,27 @@ def process_payment(
...
+def extract_credit_card_data(card):
+    credit_card = CreditCardInfo(
+        exp_year=int(card.expiration_year),
+        exp_month=int(card.expiration_month),
+        last4=card.last_4,
+        name_on_card=card.cardholder_name,
+    )
+    return CustomerSource(
+        id=card.unique_number_identifier,
+        gateway="braintree",
```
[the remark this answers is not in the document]
salwator: W really shouldn't depend on saleor settings in gateway. Perfect place for this is in GatewayConfig, I will extend dataclass with gateway name, could easily see future use for such input.

saleor/payment/utils.py
salwator: I just added this for safety in gateway customer-id related stuff, not really sure if this should apply to all _meta_?

saleor/graphql/checkout/mutations.py
```diff
@@ -469,6 +469,12 @@ class CheckoutComplete(BaseMutation):
 
     class Arguments:
         checkout_id = graphene.ID(description="Checkout ID", required=True)
+        store_source = graphene.Boolean(
+            default_value=False,
+            description=(
+                "Determines whether to stor" "the payment source for future usage."
```
NyanKiyoshi:
```suggestion
                "Determines whether to store the payment source for future usage."
```

saleor/graphql/payment/types.py
```diff
@@ -52,6 +52,19 @@ class CreditCard(graphene.ObjectType):
     )
 
 
+class PaymentSource(graphene.ObjectType):
+    class Meta:
+        description = (
+            "Represents a payment source stored"
+            " for user in payment gateway, such as credit card."
```
NyanKiyoshi: I believe it looks less weird/better if you put the trailing space in the above line instead at beginning. 

```python
x = (
    "hello "
    "world."
)

File: saleor/account/models.py (hunk 31 above)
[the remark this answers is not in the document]
Comment (salwator) [reply saleor/account/models.py]: Created issue for that: #4425

File: saleor/payment/interface.py (hunk 29 above)
Comment (salwator) [reply saleor/payment/interface.py]: Changed.

File: saleor/payment/utils.py (hunk 28 above)
Comment (salwator) [reply saleor/payment/utils.py]: Probably discussion should continue in #4425

File: saleor/graphql/payment/schema.py [hunk 35]
```diff
@@ -1,16 +1,23 @@
...
+    resolve_payment_sources,
+    resolve_payments,
+)
+from .types import Payment, PaymentSource
 
 
 class PaymentQueries(graphene.ObjectType):
     payment = graphene.Field(Payment, id=graphene.Argument(graphene.ID))
     payments = PrefetchingConnectionField(Payment, description="List of payments")
+    payment_stored_sources = graphene.List(
```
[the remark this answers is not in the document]
salwator: You're absolutely right, moved!

saleor/payment/gateways/braintree/__init__.py
```diff
@@ -234,3 +276,27 @@ def process_payment(
 ) -> GatewayResponse:
     auth_resp = authorize(payment_information, config)
     return auth_resp
+
+
+def list_client_sources(
+    config: GatewayConfig, customer_id: str
+) -> List[CustomerSource]:
+    gateway = get_braintree_gateway(**config.connection_params)
+    customer = gateway.customer.find(customer_id)
```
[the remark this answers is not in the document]
korycins: @salwator What about this? I've checked the code and AFAIK API will return client sources even if gateway has disabled storing credit cards

saleor/payment/gateways/braintree/__init__.py
```diff
@@ -109,29 +117,30 @@ def get_braintree_gateway(sandbox_mode, merchant_id, public_key, private_key):
...
-                "payment_method_nonce": payment_information.token,
-                "options": {
-                    "submit_for_settlement": config.auto_capture,
-                    "three_d_secure": {"required": THREE_D_SECURE_REQUIRED},
-                },
-                **get_customer_data(payment_information),
-            }
-        )
+        if not payment_information.customer_id:
+            result = transaction_for_new_customer(payment_information, config)
```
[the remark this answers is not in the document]
korycins: @salwator Do you agree?

saleor/graphql/account/types.py
```diff
@@ -252,6 +256,13 @@ def resolve_avatar(root: models.User, info, size=None, **_kwargs):
                 info=info,
             )
 
+    @staticmethod
+    @login_required
+    def resolve_stored_payment_sources(root: models.User, _):
```
NyanKiyoshi: By convention, we use `_info`

```suggestion
    def resolve_stored_payment_sources(root: models.User, _info):
```

PR #4204, opened 2019-06-04
Add database query count tests for API

### LINKED ISSUE #3946: Tests for counting db queries in GraphQL queries and mutations
I want to have some tests that will give me a better perspective for usage of DB queries under graphql API.  Now we don't have any information about numbers of queries hitting DB. Most of the fields have optimization but even this I would like to have explicit logic to validate that it works as expected. Also, it would be helpful during any refactor of any graphQL logic.

It would be great to have some basic tests to validate number of DB queries executed under a given graphql view.

NyanKiyoshi:
Closes #3946.

### Pull Request Checklist


tests/api/test_checkout_bench.py
```diff
@@ -0,0 +1,602 @@
...
+
+@pytest.mark.django_db
+@pytest.mark.count_queries
+def test_create_checkout(api_client, graphql_address_data, variant):
+    query = """
+        fragment Price on TaxedMoney {
+          gross {
+            amount
+            localized
+            __typename
```
maarcingebala: We can safely drop all `__typename` occurrences - they're used by Apollo internally but don't impact the database performance.

NyanKiyoshi:
This performs the most important queries that are used in storefront (checkout process, retrieve products/variants).

I believe it is ready for merge. Let me know what you think about the changes.

There is still do document this and add a note on our "convention" (having a sub-test file with the `_bench` suffix under the API tests, *à la golang* so we can clearly have the "benchmarks" under the feature tests themselves.)

The documentation itself won't explain the full workflow, only what we actually use and do inside saleor. The rest, I leave it to the user to learn more at the plugin documentation ([here](https://pytest-django-queries.readthedocs.io/en/latest/)).

Before being able to merge those, I still have to work on adding support for `pytest-xdist` which is used by saleor to run tests into multiple workers (c.f. https://github.com/NyanKiyoshi/pytest-django-queries/issues/36).

This should be the last ticket to close. So if that PR is looking good to you, I will decide that the plugin is good enough and shouldn't change in any major way, thus I will release the stable release for adding it to `pipenv` (as it is currently a release candidate).

The bot should be available tomorrow. I just need to talk with Patryk so he can set-up the webhook secret key into the repository.

*Note: the tests are failing because I didn't add the plugin yet to the requirements.*

<div align='center'>bench results</div>

NyanKiyoshi:
Ready once I get the webhook set-up into the repository.

django-queries:
Here is the report for 435f864ce095d03bf4fed36a066e7e4ab5ec1505 (NyanKiyoshi/saleor @ gql/bench/count-db-queries)
Missing base report (926df6efaae2d9d51f8905edfbb2e404d051f224). The results couldn't be compared.

```diff
# api checkout bench
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  add billing address to checkout     	         40	         40
  add shipping to checkout            	          7	          7
  checkout payment charge             	         15	         15
  complete checkout                   	          6	          6
  create checkout                     	         44	         44

# api homepage bench
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve main menu                  	          5	          5
  retrieve product list               	          4	          4
  retrieve secondary menu             	          5	          5
  retrieve shop                       	          2	          2

# api product bench
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  product details                     	         16	         16

# api variant bench
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve variant list               	         13	         13
```

maarcingebala:
@NyanKiyoshi I think we could have a bit cleaner structure of tests if we moved newly added benchmark tests to a separate directory e.g. `api/benchmark`, what do you think?

NyanKiyoshi:
@maarcingebala thanks for the suggestion! It might be better that way, so we don't have a huge amount of files in a single directory. I moved them.

PR #4242, opened 2019-06-10
Fix for ajax request to update checkout line items

korycins:
Displaying prices in (`gross` or `net`) should use site settings. 
Refresh checkout and checkout_line object after update quantity.

### DIFF
--- a/saleor/checkout/views/__init__.py
+++ b/saleor/checkout/views/__init__.py
@@ -3,6 +3,8 @@ from django.http import JsonResponse
 from django.shortcuts import get_object_or_404, redirect, render
 from django.template.response import TemplateResponse
 
+from saleor.core.utils.taxes import price
+
 from ...account.forms import LoginForm
 from ...core.utils import format_money, get_user_shipping_country, to_local_currency
 from ...shipping.utils import get_shipping_price_estimate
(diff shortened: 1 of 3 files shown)


saleor/checkout/views/__init__.py
```diff
@@ -3,6 +3,8 @@
 from django.shortcuts import get_object_or_404, redirect, render
 from django.template.response import TemplateResponse
 
+from saleor.core.utils.taxes import price
```
NyanKiyoshi: Absolute import 👀

saleor/core/utils/taxes.py
```diff
@@ -104,3 +106,18 @@ def get_taxed_shipping_price(shipping_price, taxes):
     if not charge_taxes_on_shipping():
         taxes = None
     return apply_tax_to_price(taxes, DEFAULT_TAX_RATE_NAME, shipping_price)
+
+
+def price(base: Union[TaxedMoney, TaxedMoneyRange], display_gross=None) -> Money:
```
NyanKiyoshi: As it is not a type itself but a convertion, wouldn't that be better to add at least a verb?

> get_display_price

Or something like this?

PR #4292, opened 2019-06-14
Fixed internal error when creating a checkout with a voucher code

### LINKED ISSUE #4291: User cannot checkout without voucher
As side effect from #4154, it was noticed that users would not be able to use a checkout code without encountering a server error when placing their order.

NyanKiyoshi:
Closes #4291

This is a fix that will need further rework, because we have issues with managing the voucher usage count:
  1. If we were to do a transaction on the whole mutation, and the charge API call to the payment gateway was to be slow–or a lot of people were to be using the same voucher around the same time,  they we would be getting locked by the row, and thus would have a long loading time.
  1. UserB could use the same voucher code when UserA was getting charged for his voucher. UserB gets the error "Voucher is expired" but after that, if charging UserA fails, UserC will then use the voucher code and it will no longer be expired because it was updated.

### Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,8 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fix quantity height in draft order edit - #4273 by @benekex2
 - Fix checkbox clickable area size - #4280 by @dominik-zeglen
 - Fix exploding object select in menu section - #4282 by @dominik-zeglen
-- Add default state to rich text editor = #4281 by @dominik-zeglen
+- Add default state to rich text editor - #4281 by @dominik-zeglen
+- Fixed internal error when creating a checkout with a voucher code - #4292 by @NyanKiyoshi
 
 
 ## 2.6.0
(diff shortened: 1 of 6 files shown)


django-queries:
Here is the report for 90ca684c760d7c3c5fd7240e7e80202119f0932c (NyanKiyoshi/saleor @ fix/checkout-with-voucher)
Base comparison is 15ae49e1f9664395b3f53c50b3775dfab9b05104.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  add billing address to checkout     	         41	         41
  add shipping to checkout            	          7	          7
  checkout payment charge             	         16	         16
  complete checkout                   	          6	          6
  create checkout                     	         45	         45

# api.benchmark homepage
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve main menu                  	          5	          5
  retrieve product list               	          4	          4
  retrieve secondary menu             	          5	          5
  retrieve shop                       	          2	          2

# api.benchmark product
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  product details                     	         16	         16

# api.benchmark variant
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve variant list               	         13	         13
```

</p>
</details>

PR #4294, opened 2019-06-14
Fix ajax request for removing item from bag

korycins:
It was raising an exception when checkout_line had been removed from cart

### DIFF
--- a/saleor/checkout/views/__init__.py
+++ b/saleor/checkout/views/__init__.py
@@ -5,7 +5,7 @@ from django.template.response import TemplateResponse
 
 from ...account.forms import LoginForm
 from ...core.utils import format_money, get_user_shipping_country, to_local_currency
-from ...core.utils.taxes import get_display_price
+from ...core.utils.taxes import ZERO_TAXED_MONEY, get_display_price
 from ...shipping.utils import get_shipping_price_estimate
 from ..forms import CheckoutShippingMethodForm, CountryForm, ReplaceCheckoutLineForm
 from ..models import Checkout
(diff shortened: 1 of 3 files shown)


django-queries:
Here is the report for 24691cec4be8b1aefb023c1f6a21a5b16b84f2a3 (korycins/saleor @ fix_for_removing_checkout_line_from_bag)
Base comparison is 15ae49e1f9664395b3f53c50b3775dfab9b05104.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  add billing address to checkout     	         41	         41
  add shipping to checkout            	          7	          7
  checkout payment charge             	         16	         16
  complete checkout                   	          6	          6
  create checkout                     	         45	         45

# api.benchmark homepage
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve main menu                  	          5	          5
  retrieve product list               	          4	          4
  retrieve secondary menu             	          5	          5
  retrieve shop                       	          2	          2

# api.benchmark product
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  product details                     	         16	         16

# api.benchmark variant
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve variant list               	         13	         13
```

</p>
</details>

saleor/checkout/views/__init__.py
```diff
@@ -218,8 +218,11 @@ def update_checkout_line(request, checkout, variant_id):
     if form.is_valid():
         form.save()
         checkout.refresh_from_db()
-        checkout_line.refresh_from_db()
-        subtotal = get_display_price(checkout_line.get_total(discounts, taxes))
+        checkout_line = checkout.lines.filter(variant_id=variant_id).first()
```
NyanKiyoshi: I would add a comment explain why we need this. If I understand correctly, it's because if the quantity is set to 0, it will delete the line.

korycins: Exactly

PR #4338, opened 2019-06-24
Staff users can now flag attributes as to be hidden in the storefront

### LINKED ISSUE #4256: Add ability to determine visibility of an attribute in the storefront
Add a boolean flag to the `Attribute` model to determine whether an attribute should be visible in the storefront (e.g. on the product details page). Suggested field name: `visible_in_storefront`. Include the field in the API as well - in the `Attribute` type and mutations to create/update attributes.

NyanKiyoshi:
Closes #4256.

### Pull Request Checklist

### DIFF
--- a/saleor/product/models.py
+++ b/saleor/product/models.py
@@ -515,8 +515,8 @@ class Attribute(models.Model):
         through=AttributeVariant,
         through_fields=["attribute", "product_type"],
     )
-    visible_in_storefront = models.BooleanField(default=True)
 
+    visible_in_storefront = models.BooleanField(default=True)
     is_variant_only = models.BooleanField(default=False)
 
     objects = AttributeQuerySet.as_manager()
--- a/saleor/product/models.py
+++ b/saleor/product/models.py
@@ -478,6 +478,18 @@ class AttributeQuerySet(models.QuerySet):
             | Q(attributevariant__product_type_id=product_type_pk)
         )
 
+    @staticmethod
+    def user_has_access_to_all(user):
+        return user.is_active and user.has_perm("product.manage_products")
+
+    def get_public_attributes(self):
+        return self.filter(visible_in_storefront=True)
+
+    def get_visible_to_user(self, user):
+        if self.user_has_access_to_all(user):
+            return self.all()
+        return self.get_public_attributes()
+
 
 class Attribute(models.Model):
     slug = models.SlugField(max_length=50, unique=True)
@@ -503,6 +515,7 @@ class Attribute(models.Model):
         through=AttributeVariant,
         through_fields=["attribute", "product_type"],
     )
+    visible_in_storefront = models.BooleanField(default=True)
 
     is_variant_only = models.BooleanField(default=False)
 
--- a/tests/api/test_attributes.py
+++ b/tests/api/test_attributes.py
@@ -5,6 +5,7 @@ from django.template.defaultfilters import slugify
 
 from saleor.graphql.product.enums import AttributeTypeEnum, AttributeValueType
 from saleor.graphql.product.types.attributes import resolve_attribute_value_type
+from saleor.graphql.product.types.products import resolve_attribute_list
 from saleor.graphql.product.utils import attributes_to_json
 from saleor.product import AttributeInputType
 from saleor.product.models import Attribute, AttributeValue, Category, ProductType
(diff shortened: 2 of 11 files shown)


saleor/product/models.py
```diff
@@ -478,6 +478,18 @@ def get_assigned_attributes(self, product_type_pk: int):
...
         )
 
+    @staticmethod
+    def user_has_access_to_all(user):
+        return user.is_active and user.has_perm("product.manage_products")
+
+    def get_public_attributes(self):
+        return self.filter(visible_in_storefront=True)
+
+    def get_visible_to_user(self, user):
```
korycins: Are you going to use `get_public_attributes`, `user_has_access_to_all`  somewhere else? If not I think that you can leave only `get_visible_to_user` method.
And another important question is: Are you sure that we want to show different attributes for admin users and for the rest of customers? It could lead to a strange situation like - admin user see newly added attributes, he checks storefront and everything looks good, but normal customers don't see newly added attributes.

tests/api/test_attributes.py
```diff
@@ -93,13 +92,171 @@ def test_attributes_query(user_api_client, product):
...
-    """
+"""
+
+
+def test_attributes_query(user_api_client, product):
+    attributes = Attribute.objects
+    query = QUERY_ATTRIBUTES
     response = user_api_client.post_graphql(query)
     content = get_graphql_content(response)
     attributes_data = content["data"]["attributes"]["edges"]
```
korycins: you could add here assert to confirm that `attributes_data` has `len`> 0

tests/api/test_attributes.py
```diff
@@ -93,13 +92,171 @@ def test_attributes_query(user_api_client, product):
...
+    query = QUERY_ATTRIBUTES
+
+    # hide the attribute
+    color_attribute.visible_in_storefront = False
+    color_attribute.save(update_fields=["visible_in_storefront"])
+
+    attribute_count = Attribute.objects.get_visible_to_user(
+        user_api_client.user
+    ).count()
+    assert attribute_count == Attribute.objects.all().count() - 1
```
korycins: Why ` Attribute.objects.all().count() - 1`?  How about  `Attribute.objects.filter(visible_in_storefront=True).count()`

tests/api/test_attributes.py
```diff
@@ -93,13 +92,171 @@ def test_attributes_query(user_api_client, product):
...
+    color_attribute.visible_in_storefront = False
+    color_attribute.save(update_fields=["visible_in_storefront"])
+
+    attribute_count = Attribute.objects.all().count()
+
+    # The user doesn't have the permission yet to manage products,
+    # the user shouldn't be able to see the hidden attributes
+    assert (
+        Attribute.objects.get_visible_to_user(staff_api_client.user).count()
+        == attribute_count - 1
```
korycins: Same as above

tests/api/test_attributes.py
NyanKiyoshi: Because there is only one attribute that should be hidden, so it just seems better to hard-code the one and be sure that we have what we expect (where in backend it does the proper filter).

tests/api/test_attributes.py
NyanKiyoshi:
```suggestion
    attributes_data = content["data"]["attributes"]["edges"]
    assert attributes_data
```

That should be enough 👍

tests/api/test_attributes.py
korycins: So if you always have only one attribute, you don't need to do `Attribute.objects.all().count() - 1` or `Attribute.objects.filter(visible_in_storefront=True).count()`

```
attribute_count = Attribute.objects.get_visible_to_user(
         user_api_client.user
).count()
assert attribute_count == 1
```

saleor/product/models.py
NyanKiyoshi: Yes, it will be used in another task

NyanKiyoshi: > Are you sure that we want to show different attributes for admin users and for the rest of customers? It could lead to a strange situation like - admin user see newly added attributes, he checks storefront and everything looks good, but normal customers don't see newly added attributes.

In another task we will add filtering. The storefront will take care of displaying the correct information depending on the context.

PR #4339, opened 2019-06-24
Fix voucher apply once per order

### LINKED ISSUE #4314: Voucher one per order doesn't work
The once per user option doesn't work on percentage discounts. It always gives percentage of the whole order value. But it works for dollar amounts.

fowczarek:
I want to merge this change because, resolve #4314

### Pull Request Checklist

### DIFF
--- a/tests/api/test_discount.py
+++ b/tests/api/test_discount.py
@@ -3,11 +3,11 @@ from datetime import date, timedelta
 import graphene
 import pytest
 from django_countries import countries
-from tests.api.utils import get_graphql_content
 
 from saleor.discount import DiscountValueType, VoucherType
 from saleor.discount.models import Sale, Voucher
 from saleor.graphql.discount.enums import DiscountValueTypeEnum, VoucherTypeEnum
+from tests.api.utils import get_graphql_content
 
 
 @pytest.fixture
@@ -139,13 +139,14 @@ mutation  voucherCreate(
     $type: VoucherTypeEnum, $name: String, $code: String,
     $discountValueType: DiscountValueTypeEnum,
     $discountValue: Decimal, $minAmountSpent: Decimal,
-    $startDate: Date, $endDate: Date) {
+    $startDate: Date, $endDate: Date, $applyOncePerOrder: Boolean) {
         voucherCreate(input: {
                 name: $name, type: $type, code: $code,
                 discountValueType: $discountValueType,
                 discountValue: $discountValue,
                 minAmountSpent: $minAmountSpent,
-                startDate: $startDate, endDate: $endDate}) {
+                startDate: $startDate, endDate: $endDate,
+                applyOncePerOrder: $applyOncePerOrder}) {
             errors {
                 field
                 message
@@ -160,6 +161,7 @@ mutation  voucherCreate(
                 discountValueType
                 startDate
                 endDate
+                applyOncePerOrder
             }
         }
     }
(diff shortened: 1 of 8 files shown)


django-queries:
Here is the report for c24bfa53c45c1d233d11182b928a9c0a5587a38f (mirumee/saleor @ fix_voucher_one_per_order)
Base comparison is f7897c7a9b67086b2963153cb848b63d57120fde.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  add billing address to checkout     	         41	         41
  add shipping to checkout            	          7	          7
  checkout payment charge             	         16	         16
  complete checkout                   	          6	          6
  create checkout                     	         45	         45

# api.benchmark homepage
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve main menu                  	          5	          5
  retrieve product list               	          4	          4
  retrieve secondary menu             	          5	          5
  retrieve shop                       	          2	          2

# api.benchmark product
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  product details                     	         16	         16

# api.benchmark variant
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve variant list               	         13	         13
```

</p>
</details>

tests/api/test_discount.py
```diff
@@ -3,11 +3,11 @@
 import graphene
 import pytest
 from django_countries import countries
-from tests.api.utils import get_graphql_content
 
 from saleor.discount import DiscountValueType, VoucherType
 from saleor.discount.models import Sale, Voucher
 from saleor.graphql.discount.enums import DiscountValueTypeEnum, VoucherTypeEnum
+from tests.api.utils import get_graphql_content
```
korycins: Relative

tests/api/test_discount.py
```diff
@@ -271,28 +275,34 @@ def test_create_voucher_with_existing_voucher_code(
...
                 voucher {
                     code
                     discountValueType
+                    applyOncePerOrder
                 }
             }
         }
     """
     # Set discount value type to 'fixed' and change it in mutation
+    apply_once_per_order = not voucher.apply_once_per_order
```
korycins: You put this line between line's comment and line. You should keep them together or remove the comment if it is not needed

tests/conftest.py
```diff
@@ -508,6 +519,12 @@ def product_list(product_type, category):
             ),
         ]
     )
+    sku = 0
+    for product in products:
+        ProductVariant.objects.create(
+            product=product, sku=sku, track_inventory=True, quantity=100
+        )
+        sku = sku + 1
```
korycins: This could generate a problem with db because sku is unique

tests/conftest.py
```diff
@@ -93,6 +93,17 @@ def checkout_with_item(checkout, product):
     return checkout
 
 
+@pytest.fixture
+def checkout_with_items(checkout, product_list, product):
```
korycins: Isn't `checkout_with_item` enough for you? This fixture generates a huge amount of unneeded data and it is time and db expensive.

tests/test_checkout.py
```diff
@@ -849,6 +849,38 @@ def test_get_discount_for_checkout_value_voucher(
...
+    discount_amount,
+):
+    voucher = Voucher(
+        code="unique",
+        type=VoucherType.PRODUCT,
+        discount_value_type=discount_type,
+        discount_value=discount_value,
+        apply_once_per_order=apply_once_per_order,
+    )
+    voucher.save()
```
korycins: Why not `Voucher.objects.create`?

tests/conftest.py
NyanKiyoshi: Are you sure? We have something like `123` or `555` in other places.

korycins: Try to use fixtures `product` and `unavailable_product_with_variant` in the same test ;)
If we don't use tools like faker we should keep in mind that unique fields shouldn't be duplicated (like `uuid4()`)

PR #4351, opened 2019-06-26
Enterprise grade attributes

### LINKED ISSUE #3554: Shared attribute between different product types

The attribute couldn't be shared between different product type. Each product type must have its own attributes. How about making same attribute able to be shared by different product types?

I can create a ```productTypeA``` with ```size``` attribute (S, M, L). But if I would like to create another ```productTypeB``` with same ```size``` attribute, I have to create another ```size``` attribute, even they have same values.

Will there have any issue if we change the ```product_type``` and  ```product_variant_type``` of ```Attribute``` to ManyToMany fields?

NyanKiyoshi:
This PR introduces enhanced product attributes management:
- attributes independent from product types,
- ability to assign attributes to multiple product types,
- assigning different input types to attribute values: single-select or multi-select,
- set of attribute properties determining their behavior in the storefront and in the dashboard, e.g. visibility in the storefront or ability to filter by an attribute.

Fixes #3554

### Before Merging: checklist
- [x] New migrations are squashed
- [x] Definitions are up to date
- [x] Requirements files are updated
- [ ] Clean up the changelogs
- [x] The travis changes' commits are **REMOVED** (by doing a rebase)
- [x] The branch is fully rebased onto master

### Pull Request Checklist


django-queries:
Here is the report for b1993e461f01976075686b468eb5c749c1aa0383 (mirumee/saleor @ next/enterprise-grade-attributes)
Base comparison is 482ba001bf3cbb8a6e765649e546b33bb58d00f7.



<details><summary>**Found 2 differences!** (click me)</summary>
<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  add billing address to checkout     	         41	         41
  add shipping to checkout            	          7	          7
  checkout payment charge             	         16	         16
  complete checkout                   	          6	          6
  create checkout                     	         45	         45

# api.benchmark homepage
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
  retrieve main menu                  	          5	          5
  retrieve product list               	          4	          4
  retrieve secondary menu             	          5	          5
  retrieve shop                       	          2	          2

# api.benchmark product
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
- product details                     	         13	         15

# api.benchmark variant
  test name                           	left count 	right count
  ------------------------------------	-----------	-----------
- retrieve variant list               	          9	         15
```

</p>
</details>

saleor/product/models.py
```diff
@@ -421,28 +449,89 @@ def get_absolute_url(self) -> str:
...
+    def user_has_access_to_all(user):
+        return user.is_active and user.has_perm("product.manage_products")
+
+    def get_public_attributes(self):
+        return self.filter(visible_in_storefront=True)
+
+    def get_visible_to_user(self, user):
+        if self.user_has_access_to_all(user):
+            return self.all()
+        return self.get_public_attributes()
```
NyanKiyoshi: The mutation prefetching doesn't work. And is triggering a query for each attribute value (`prefetch_related("values")`).

This is also triggering a double queries (`prefetch` and `get_visible_to_user`).

We should have a `Prefetch` instead.

maarcingebala:
UI review:

- Fix alignment of columns in the attributes tables (product type details view):
image

- Adding an attribute value with a name that already exists breaks the UI (this validation error is unhandled: `0: {field: null, message: "Attribute value with this Name and Attribute already exists."}`

- Assigning attributes to product types breaks when trying to search attributes by name.

khalibloo:
Hi. What changes/features can we expect from this PR?

NyanKiyoshi:
@dominik-zeglen maybe we would want to set `Visible on Product Page in Storefront` when creating a new attribute to `true` by default?

maarcingebala:
@NyanKiyoshi @dominik-zeglen Actually I think we could set all of those switches to "true" by default, I think in most cases we want to to be enabled.

NyanKiyoshi:
There is an unhandled error as well when trying to assign a multi value attribute as a variant attribute: `Attributes having for input types ['multiselect'] cannot be assigned as variant attributes`

KyleMcDonell:
Some thoughts I wanted to share on attributes and filtering in general:

There is a great article on product filtering functionality: 
https://www.bigcommerce.com/blog/7-product-filtering-implementations-that-make-macys-best-in-class/#undefined

To me the biggest feature that I think is missing here is limiting/specifically choosing which filters appear in a category view. For example a _Sleeping Bag_ category might show filters for attributes like temperature rating, but the parent category _Camping Gear_, which might also include camping stoves, should not.

I would propose adding a "filter builder" to the category editor. Just as an admin can view a list of products in a category, they should be able to view the attributes of the products in the category. They should be able to choose which product attributes will be available as a filter. Ideally an admin will also be able to control the *order* of the filters.

maarcingebala:
Thanks for your input @KyleMcDonell. We'll take a look at the article and think about the problem you mentioned. It's definitely not the final shape of the new attributes management yet, but the core functionality is ready and we want to merge it.

PR #4433, opened 2019-07-10
Add extensions manager

korycins:
- Add settings variables for storing the path to Extension Manager and list of enabled plugins.
- Add check logic to validate extension_manager and plugins imports in runtime
- Add abstract base plugin class.
- Add default Manager
- Add `extensions` to middleware.
- Add unittests for the manager, and runtime checks


saleor/settings.py
```diff
@@ -658,3 +658,7 @@ def get_host():
     "RELAY_CONNECTION_ENFORCE_FIRST_OR_LAST": True,
     "RELAY_CONNECTION_MAX_LIMIT": 100,
 }
+
+EXTENSIONS_MANAGER = "saleor.core.extensions.manager.BaseManager"
+
+PLUGINS = []
```
NyanKiyoshi: Note for the documentation: we should tell the users how to set `DJANGO_SETTINGS` to override `PLUGINS` without touching `settings.py`

django-queries:
Here is the report for ce05ea8deb2f67560f10d9d35c92b33dd140bf92 (korycins/saleor @ extensions_manager)
Base comparison is d5282637b518869d24c8b54a5179fb531500e92d.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         43	         43	             21

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	          9	          9	              2
```

</p>
</details>

saleor/core/extensions/manager.py
```diff
@@ -0,0 +1,58 @@
...
+
+        return super().__getattribute__(item)
+
+    def __get_plugin_method(self, name: str):
+        def run_plugin_method(*args, **kwargs):
+            value = None
+            for p in self.plugins:
+                try:
+                    plugin_value = getattr(p, name)(*args, **kwargs)
+                    # FIXME
```
korycins: This will be solved after I convert avalara and vatlayer to plugins

saleor/core/extensions/dummy_plugin.py
```diff
@@ -0,0 +1,98 @@
...
+if TYPE_CHECKING:
+    from ...checkout.models import Checkout, CheckoutLine
+    from ...product.models import Product
+    from ...account.models import Address
+    from ...order.models import OrderLine, Order
+
+
+class DummyPlugin(BasePlugin):
+    """Dummy plugin class which overwrites all base plugin methods. It should be used
+    when any other plugins are disabled"""
```
NyanKiyoshi: "any other"

Did you mean when "no plugin are enabled"? Any other sounds like it is enabled when some plugins are disabled.

saleor/core/extensions/dummy_plugin.py
```diff
@@ -0,0 +1,98 @@
...
+from .plugin import BasePlugin
+
+if TYPE_CHECKING:
+    from ...checkout.models import Checkout, CheckoutLine
+    from ...product.models import Product
+    from ...account.models import Address
+    from ...order.models import OrderLine, Order
+
+
+class DummyPlugin(BasePlugin):
```
maarcingebala: Maybe `StandardPlugin` as this is the "standard" implementation of all calculations without any 3rd party integration?

saleor/core/extensions/plugin.py
```diff
@@ -0,0 +1,84 @@
...
+    ):
+        raise NotImplementedError()
+
+    def show_taxes_on_storefront(self) -> bool:
+        raise NotImplementedError()
+
+    def taxes_are_enabled(self) -> bool:
+        raise NotImplementedError()
+
+    def apply_taxes_to_shipping_price_range(
```
maarcingebala: I would group related methods together e.g. all "apply_taxes_" should be together, "calculate_checkout_" together, etc.

saleor/core/extensions/plugin.py
```diff
@@ -0,0 +1,84 @@
...
+    from ...checkout.models import Checkout, CheckoutLine
+    from ...product.models import Product
+    from ...account.models import Address
+    from ...order.models import OrderLine, Order
+
+
+class BasePlugin:
+    """Abstract class for storing all methods available for any plugin."""
+
+    def calculate_checkout_total(
```
maarcingebala: We have four "calculate_checkout_" functions that calculate some amounts related to checkout. Do we prefer this granularity over single "calculate_checkout_totals" function?

korycins: I was thinking about this. Multiple methods will be better than single. We have graphQL API which can query only single value. We calculate only the required value instead of all values.
After I will move vatlayer and avalara to the manager we can discuss it one more time.

saleor/core/extensions/manager.py
```diff
@@ -0,0 +1,181 @@
...
+from ..taxes import TaxType, quantize_price
+
+if TYPE_CHECKING:
+    from ...checkout.models import Checkout, CheckoutLine
+    from ...product.models import Product
+    from ...account.models import Address
+    from ...order.models import OrderLine, Order
+
+
+class BaseManager:
```
salwator: Why is it BaseManager if it is not an abstract base class?

saleor/core/extensions/manager.py
```diff
@@ -0,0 +1,181 @@
...
+        for plugin_path in plugins:
+            plugin_path, _, plugin_name = plugin_path.rpartition(".")
+            plugin_module = importlib.import_module(plugin_path)
+            plugin_class = getattr(plugin_module, plugin_name)
+            self.plugins.append(plugin_class())
+
+    def run_plugin_method(self, name: str, default_value: Any, *args, **kwargs):
+        value = default_value
+        for p in self.plugins:
+            plugin_value = getattr(p, name, NotImplemented)(
```
salwator: This is extremely unclear what is happening here. Can we have some kind of function extracted from here with proper name or at least other kind of explanation? Impossible to quickly reason about types being passed around here.

saleor/core/extensions/manager.py
```diff
@@ -0,0 +1,181 @@
...
+        self.plugins = []
+        for plugin_path in plugins:
+            plugin_path, _, plugin_name = plugin_path.rpartition(".")
+            plugin_module = importlib.import_module(plugin_path)
+            plugin_class = getattr(plugin_module, plugin_name)
+            self.plugins.append(plugin_class())
+
+    def run_plugin_method(self, name: str, default_value: Any, *args, **kwargs):
+        value = default_value
+        for p in self.plugins:
```
salwator: Is it just plain reduce here?
We could handle it a little more clearly in few ways, with reduce or simpler loop:
```python
methods = (getattr(p, name) for p in self.plugins if hasattr(p, name))
value = default_value
for method in methods:
    value = method(*args, **kwargs, previous_value=value)
return value
```

saleor/core/extensions/plugin.py
salwator: I was thinking: shouldn't we extract ex. _checkout_ related methods to subclass or other separate thing that residues in relevant place?
Same for other domain parts.

tests/extensions/test_manager.py
```diff
@@ -0,0 +1,14 @@
+from saleor.core.extensions.manager import BaseManager, get_extensions_manager
+from saleor.core.extensions.plugin import BasePlugin
+
+
+class TestPlugin(BasePlugin):
+    """"""
+
+
+def test_get_extensions_manager():
```
salwator: Do we have any plans for testing if calling any plugin methods actually works?

saleor/core/extensions/plugin.py
korycins: Yeap, we were discussing it. We will do that in further steps. For now, we want to introduce a single manager but I am still keeping that in mind ;)

tests/extensions/test_manager.py
korycins: Yeap. that was my fault. After I've changed architecture for Plugin and Manager I've removed previous tests but I didn't add any new one.

saleor/core/extensions/manager.py
korycins: Logic has been split to improve readability

PR #4454, opened 2019-07-12
DebugToolbar 2.0 doesn't work with RequestHistory 0.0.11 plugin

derenio:
Recent update of the `django-debug-toolbar==2.0` isn't compatible with the latest 'djangoe-debug-toolbar-request-history==0.0.11` plugin.

https://github.com/djsutho/django-debug-toolbar-request-history/issues/28

Btw. why do we keep `debug-toolbar` in the standard requirements instead of the dev one?


django-queries:
Here is the report for 0ab4b6ce8bfd39ff4e4fa563850990b38c1e0008 (mirumee/saleor @ fix/downgrade_django_debug_toolbar)
Base comparison is 33da5754e16c89e65af31e4b646231f35880be99.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         43	         43	             21

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	          9	          9	              2
```

</p>
</details>

maarcingebala:
Since we're touching it, can we move it to the right requirements in this PR? Also, we should update the Pipfile.

Pipfile
```diff
@@ -84,6 +82,8 @@ pylint = "*"
 pylint_celery = "*"
 isort = "*"
 yapf = "*"
+django-debug-toolbar = "==1.11" # The Request History plugin isn't compatible with 2.0
```
NyanKiyoshi: We could just tell pipenv to install `<2`. So we don't block any `1.x` patch (even if it is very unlikely to happen).

jxltom:
There is a discussion about why debugtoolbar should stay in regular requirements in PR https://github.com/mirumee/saleor/pull/3076.

NyanKiyoshi:
Why though? The related issue was just showing an error, not the error stack... Maybe it was an issue because they used `DEBUG=True` without having the development dependencies?

jxltom:
How about this explanation? https://github.com/mirumee/saleor/issues/3042#issuecomment-427298303

NyanKiyoshi:
Thanks for the link! After discussion, it is better to have them as a base requirement as it is something that is directly importer from the code. It wouldn't crash in production but the issue is that some users might not read properly or at all the documentation, and thus open issues about the error because they didn't install the dev dependencies.

Thanks for the catch and remark @jxltom!

PR #4510, opened 2019-07-23
Refactor account mutations

### LINKED ISSUE #4480: New naming convention for logged-in user
We should use the naming convention for all mutation related to the logged-in user. 
New naming for this mutation is account<mutationName>(eg. "CustomerAddressCreate" => "AccountAddressCreate")

fowczarek:
I want to merge this change because resolve #4480 

### To do: 
- [x] Split account mutations for files
- [x] Allow using PasswordReset mutation for accounts
- [x] Refactor address mutations for dashboard and storefront.
- [x] Unify returned values by address mutations (User + address)
- [x] Move deprecated mutations to separate files
- [x] use "Account" prefix for all mutation related to the logged-in user

### Pull Request Checklist


django-queries:
Here is the report for e96e9e51d3fbf049ee2308f465a5b9005950cbb1 (mirumee/saleor @ 4480/refactor_account_mutations)
Base comparison is 82531e12ac3a2fedb5c6aa80c81365d6a8b1538a.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         41	         41	             20

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	          9	          9	              2
```

</p>
</details>

tests/api/test_account.py
```diff
@@ -1258,9 +1258,7 @@ def test_password_reset_email(
     """
     email = customer_user.email
     variables = {"email": email}
-    response = staff_api_client.post_graphql(
-        query, variables, permissions=[permission_manage_users]
-    )
+    response = staff_api_client.post_graphql(query, variables)
```
NyanKiyoshi: Why did we drop the permission checks?

fowczarek: We have 2 mutations for resetting password.  The only difference between them is permission. I remove the permission form `PasswordReset` and set `CustomerPasswordReset` as deprecated.

saleor/graphql/account/mutations/account.py
```diff
@@ -0,0 +1,217 @@
...
+
+class CustomerPasswordReset(BaseMutation):
+    class Arguments:
+        input = CustomerPasswordResetInput(
+            description="Fields required to reset customer's password", required=True
+        )
+
+    class Meta:
+        description = (
+            "DEPRECATED: Use SetPassword instead. Resets the customer's password."
```
NyanKiyoshi: How does one get the reset token?

tests/api/test_account.py
NyanKiyoshi: Then, shouldn't we use `customer_api_client` instead of `staff_client`?
Same thing for `test_password_reset_email_non_existing_user`

saleor/graphql/account/mutations/staff.py
```diff
@@ -369,6 +370,14 @@ def perform_mutation(cls, _root, info, address_id, user_id, **data):
         return cls(user=user)
 
 
+class AddressUpdate(BaseAddressUpdate):
```
NyanKiyoshi: Maybe we could name it `CustomerAddressUpdate`? So users can more easily assume this is a staff mutation.

tests/api/test_account.py
```diff
@@ -1369,10 +1369,24 @@ def test_address_update_mutation(
     assert address_obj.city == graphql_address_data["city"]
 
 
+ACCOUNT_ADDRESS_UPDATE_MUTATION = """
+    mutation updateUserAddress($addressId: ID!, $address: AddressInput!) {
```
NyanKiyoshi:
```suggestion
    mutation updateAccountAddress($addressId: ID!, $address: AddressInput!) {
```

saleor/graphql/account/mutations/staff.py
fowczarek: This is a good idea for the future. Now we can't use prefix "Customer" for all address mutation for staff user because in current API we have `CustomerAddressCreate` as a deprecated mutation for creating address by a logged-in user.

saleor/graphql/account/mutations/account.py
fowczarek: I added this information to the token description.

saleor/graphql/account/mutations/account.py
```diff
@@ -0,0 +1,209 @@
...
+class AccountUpdate(CustomerCreate):
+    class Arguments:
+        input = AccountInput(
+            description="Fields required to update the account of the logged-in user.",
+            required=True,
+        )
+
+    class Meta:
+        description = "Updates the account of the logged-in user."
+        exclude = ["password"]
```
korycins: Instead of excluding fields, safer is to explicitly pass a list of included fields

tests/api/test_account.py
```diff
@@ -1288,7 +1312,39 @@ def test_set_password(user_api_client, customer_user):
...
+    user_id = graphene.Node.to_global_id("User", customer_user.id)
+    variables = {"id": user_id}
+    response = staff_api_client.post_graphql(
+        query, variables, permissions=[permission_manage_users]
+    )
+    content = get_graphql_content(response)
+    data = content["data"]["customerRequestPasswordReset"]
+    assert data == {"errors": []}
+    assert send_password_reset_mock.call_count == 1
+    args, kwargs = send_password_reset_mock.call_args
```
korycins: Maybe you can use - https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with

saleor/graphql/account/mutations/account.py
```diff
@@ -0,0 +1,256 @@
...
+class AccountRequestPasswordResetInput(graphene.InputObjectType):
+    email = graphene.String(
+        required=True,
+        description=("Email of the user that will be used for password recovery."),
+    )
+
+
+class AccountRequestPasswordReset(BaseMutation):
+    class Arguments:
+        input = AccountRequestPasswordResetInput(
```
maarcingebala: Since there is only `email` field in this input, we could use it directly here and omit the input object.

saleor/graphql/account/mutations/base.py
```diff
@@ -0,0 +1,181 @@
...
+    has_perm = user.has_perm("account.manage_users")
+    belongs_to_user = address in user.addresses.all()
+    return has_perm or belongs_to_user
+
+
+class SetPasswordInput(graphene.InputObjectType):
+    token = graphene.String(
+        description=(
+            "A one-time token required to set the password. "
+            "Sent by email using PasswordReset mutation for if staff user "
```
maarcingebala: Here I would only say "A one-time token required to set the password.". Reset of the description would be moved to meta description.

saleor/graphql/account/mutations/staff.py
```diff
@@ -341,6 +214,28 @@ def perform_mutation(cls, root, info, **data):
         return results
 
 
+class CustomerRequestPasswordReset(BaseMutation):
```
maarcingebala: Let's have only one mutation in API to request the password reset: `requestPasswordReset(email: String): RequestPasswordReset`.

saleor/graphql/account/mutations/account.py
maarcingebala: As we spoke this excludes fields from validation, not from the output. @fowczarek I would go through all `exclude` that this PR adds and see if they're actually needed. They mostly try to exclude password but that should be already excluded at the `User` type level.

saleor/graphql/account/mutations/staff.py
```diff
@@ -0,0 +1,379 @@
...
+        permissions = ("account.manage_users",)
+
+    @classmethod
+    def perform_mutation(cls, _root, info, address_id, user_id, **data):
+        address = cls.get_node_or_error(
+            info, address_id, field="address_id", only_type=Address
+        )
+        user = cls.get_node_or_error(info, user_id, field="user_id", only_type=User)
+
+        if address not in user.addresses.all():
```
korycins: It loads all user's addresses to memory. I think that a better idea is to use `.exists()` or normal `filter()`

korycins: Let's change this in all modified files

tests/api/test_account.py
```diff
@@ -967,11 +991,80 @@ def test_account_request_deletion(
...
+
+
+def test_account_delete_staff_user(staff_api_client):
+    user = staff_api_client.user
+    variables = {"token": "invalid"}
+
+    response = staff_api_client.post_graphql(ACCOUNT_DELETE_MUTATION, variables)
+    content = get_graphql_content(response)
+    data = content["data"]["accountDelete"]
+    assert not data["errors"] == []
```
korycins: `assert len(data["errors"]) == 1`  would be more meaningful

saleor/graphql/account/mutations/deprecated_account.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+class CustomerPasswordReset(BaseMutation):
+    class Arguments:
+        input = CustomerPasswordResetInput(
+            description="Fields required to reset customer's password", required=True
+        )
+
+    class Meta:
+        description = "Resets the customer's password."
```
maarcingebala: If this mutation is deprecated we should say in the description which one to use instead.

saleor/graphql/account/mutations/deprecated_staff.py
```diff
@@ -0,0 +1,28 @@
...
+from .base import send_user_password_reset_email
+
+
+class PasswordReset(BaseMutation):
+    class Arguments:
+        email = graphene.String(description="Email", required=True)
+
+    class Meta:
+        description = (
+            "DEPRECATED: Use CustomerRequestPasswordReset instead."
```
maarcingebala: We don't have `CustomerRequestPasswordReset` mutation.

saleor/graphql/account/mutations/account.py
```diff
@@ -0,0 +1,220 @@
...
+    class Meta:
+        description = "Register a new user."
+        exclude = ["password"]
+        model = models.User
+
+    @classmethod
+    def save(cls, info, user, cleaned_input):
+        password = cleaned_input["password"]
+        user.set_password(password)
+        user.save()
```
NyanKiyoshi:
```suggestion
        user.save(update_fields=['password'])
```

saleor/graphql/account/mutations/base.py
```diff
@@ -0,0 +1,289 @@
...
+def can_edit_address(user, address):
+    """Determine whether the user can edit the given address.
+
+    This method assumes that an address can be edited by:
+    - users with proper permissions (staff)
+    - customers associated to the given address.
+    """
+    has_perm = user.has_perm("account.manage_users")
+    belongs_to_user = user.addresses.filter(pk=address.pk).exists()
+    return has_perm or belongs_to_user
```
NyanKiyoshi: If the user has the permission, there is no need to call the database.

```python

    has_perm = user.has_perm("account.manage_users")
    belongs_to_user_qs = user.addresses.filter(pk=address.pk)
    return has_perm or belongs_to_user.exists()
```

saleor/graphql/account/mutations/base.py
```diff
@@ -0,0 +1,289 @@
...
+            if default_billing_address:
+                default_billing_address.save()
+                instance.default_billing_address = default_billing_address
+
+            is_creation = instance.pk is None
+            super().save(info, instance, cleaned_input)
+
+        # The instance is a new object in db, create an event
+        if is_creation:
+            account_events.customer_account_created_event(user=instance)
```
NyanKiyoshi: We might want this inside of the transaction as we are doing an insert here.

saleor/graphql/account/mutations/deprecated_account.py
```diff
@@ -0,0 +1,175 @@
...
+    class Meta:
+        description = "DEPRECATED: Use AccountRegister instead. Register a new user."
+        exclude = ["password"]
+        model = models.User
+
+    @classmethod
+    def save(cls, info, user, cleaned_input):
+        password = cleaned_input["password"]
+        user.set_password(password)
+        user.save()
```
NyanKiyoshi:
```suggestion
        user.save(update_fields=["password"])
```

saleor/graphql/account/mutations/staff.py
```diff
@@ -0,0 +1,379 @@
...
+
+        user_id = data.get("id")
+        instance = cls.get_node_or_error(info, user_id, only_type=User)
+        cls.clean_instance(info, instance)
+
+        db_id = instance.id
+        remove_staff_member(instance)
+        # After the instance is deleted, set its ID to the original database's
+        # ID so that the success response contains ID of the deleted object.
+        instance.id = db_id
```
NyanKiyoshi: We don't delete the user here, so we don't need this

saleor/graphql/account/mutations/deprecated_account.py
fowczarek: We can't use update_fields for an object without primary key.

saleor/graphql/account/mutations/account.py
fowczarek: We can't use update_fields for an object without primary key.

saleor/graphql/account/mutations/staff.py
fowczarek: We delete the user if they have no orders placed.

PR #4573, opened 2019-07-31
Disable shipping on checkout process when checkout is not shippable

NyanKiyoshi:
- If shipping is not allowed/available on the current checkout:
  - `checkoutCreate` doesn't save the shipping address (no error);
  - `checkoutShippingMethodUpdate` now raises a validation error on the `shipping_method` field;
  - `checkoutShippingAddressUpdate` now raises a validation error on the `shipping_address` field.
- If there is an active checkout `checkoutCreate` will now return `created: False` as well as simply returning the active checkout. So now, the client is aware when the checkout was not created. In addition, the field adds some details on how to merge a cart with an active checkout.
- If the `checkoutId` is not existing when invoking `checkoutShippingAddressUpdate` it will raise an error instead of crashing.

---

Note: tests are still yet to be written.

### Pull Request Checklist


django-queries:
Here is the report for d19ee23a9546c8beb8f4caef3209d0687824fb96 (NyanKiyoshi/saleor @ checkout/no-shipping)
Base comparison is 328ddb09eefcce68fca187d8188468eaf33f99f0.



<details><summary>**Found 1 differences!** (click me)</summary>
<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
- create checkout                     	         41	         46	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	          9	          9	              2
```

</p>
</details>

django-queries:
Here is the report for d19ee23a9546c8beb8f4caef3209d0687824fb96 (NyanKiyoshi/saleor @ checkout/no-shipping)
Base comparison is 328ddb09eefcce68fca187d8188468eaf33f99f0.



<details><summary>**Found 1 differences!** (click me)</summary>
<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
- create checkout                     	         41	         46	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	          9	          9	              2
```

</p>
</details>

saleor/graphql/checkout/mutations.py
```diff
@@ -193,23 +215,36 @@ def clean_input(cls, info, instance, data):
...
+        instance: models.Checkout,
+        cleaned_input: dict,
+        variants: List[product_models.ProductVariant],
+    ):
         shipping_address = cleaned_input.get("shipping_address")
         billing_address = cleaned_input.get("billing_address")
 
-        if shipping_address:
+        if shipping_address and any(
+            variant.is_shipping_required() for variant in variants
```
maarcingebala: What about `checkout.is_shipping_required()`? Could we use it here?

NyanKiyoshi: The issue with this is that there are no lines yet in the checkout object. We could add the lines first, but then it would add an additional query as we would have to:
1. create checkout
1. add the lines
1. add the addresses
1. save checkout

instead of
1. add the addresses
1. create the checkout
1. add the lines

maarcingebala: I'm only worried that if we changed the logic of `checkout.is_shipping_required` function for some reason, this is one more place in the code that needs to be updated as well. I think one additional query when saving is not a big deal, it's still O(1).

saleor/graphql/checkout/mutations.py
```diff
@@ -193,31 +215,46 @@ def clean_input(cls, info, instance, data):
...
+        if shipping_address and instance.is_shipping_required():
             shipping_address.save()
             instance.shipping_address = shipping_address.get_copy()
+            updated_fields.append("shipping_address")
         if billing_address:
             billing_address.save()
             instance.billing_address = billing_address.get_copy()
+            updated_fields.append("billing_address")
+
+        # Save the updated fields
```
korycins: Do we need these comments? I think that code is enough readable and we don't need additional comments..

saleor/graphql/checkout/mutations.py
```diff
@@ -384,9 +421,22 @@ class Meta:
...
+            checkout = models.Checkout.objects.prefetch_related(
+                "lines__variant__product__product_type"
+            ).get(pk=pk)
+        except ObjectDoesNotExist:
+            raise ValidationError(
+                {"checkout_id": f"Couldn't resolve to a node: {checkout_id}"}
+            )
+
+        if not checkout.is_shipping_required():
+            raise ValidationError({"shipping_address": ERROR_DOES_NOT_SHIP})
```
korycins: `ERROR_DOES_NOT_SHIP = "This checkout doesn't have shipping"` -> Wouldn't the `"This checkout doesn't need shipping"` be better?

saleor/graphql/checkout/mutations.py
```diff
@@ -538,13 +599,19 @@ def perform_mutation(cls, _root, info, checkout_id, store_source):
                 )
 
         try:
-            billing_address = order_data["billing_address"]  # type: models.Address
-            shipping_address = order_data["shipping_address"]  # type: models.Address
+            billing_address = order_data["billing_address"]
```
korycins: I don't think that this part of the code should be in `try`, `except`. Let's keep them as tight as possible 
`perform_mutation` is growing and growing, maybe we can extract this part to code to separate method.

saleor/graphql/checkout/mutations.py
```diff
@@ -538,13 +599,19 @@ def perform_mutation(cls, _root, info, checkout_id, store_source):
...
             txn = gateway_process_payment(
                 payment=payment,
                 payment_token=payment.token,
-                billing_address=AddressData(**billing_address.as_data()),
-                shipping_address=AddressData(**shipping_address.as_data()),
+                billing_address=billing_address,
+                shipping_address=shipping_address,
                 store_source=store_source,
             )
             if txn.is_success and txn.customer_id and user.is_authenticated:
```
korycins: I know that this is not part of this PR but maybe we could move this outside the try except? Then we don't need to check `txn.is_success`

NyanKiyoshi: We definitively should almost move everything from it. Will do that 👍

saleor/graphql/checkout/mutations.py
NyanKiyoshi: I added those, in case someone goes through this and is wondering about the case where `updated_fields` is empty.

I could remove `Save [...]` for sure

PR #4576, opened 2019-07-31
Add dynamic settings api for plugins

korycins:
- Add new module to saleor `extensions`
- Add new model `PluginConfiguration` for storing the configuration of plugins
- Add queries - `pluginConfigurations`, `pluginConfiguration` for serving all mandatory data about plugins.
- Add mutation `pluginConfigurationUpdate` which allows user to update configuration of the plugin.
- Add logic to save and retrieve configuration to `BasePlugin`
- Add new permision `manage_extensions`
- `Avalara` and `Vatlayer` can be enabled by API or by settings.
- Tests for new logic
- Update documentation


django-queries:
Here is the report for 439adf2a46050908a06fa49c2c34c93ca8becb53 (korycins/saleor @ dynamic_settings_for_plugins)
Base comparison is 7818fa14feadd8f9deb746ce45b5278de804c18e.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         41	         41	             20

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	          9	          9	              2
```

</p>
</details>

setup.cfg
```diff
@@ -14,7 +14,6 @@ exclude_lines =
     return NotImplemented
 
 [tool:pytest]
-addopts = -n auto --vcr-record-mode=none
```
NyanKiyoshi: Don't forget to update tox with `-n auto`

NyanKiyoshi:
@korycins what about if the plugin owner wants to change their default configuration or add entries to it? Wouldn't it be better to just have a configuration to `{}` by default and then merge the defaults and the db one?

saleor/graphql/schema.graphql
```diff
@@ -1309,8 +1327,6 @@ input FulfillmentUpdateTrackingInput {
 enum GatewaysEnum {
   DUMMY
   BRAINTREE
-  RAZORPAY
-  STRIPE
```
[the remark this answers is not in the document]
korycins: It was added by mistake

setup.cfg
korycins: It was added by mistake. `git stash` is guilty!

saleor/extensions/plugins/avatax/plugin.py
```diff
@@ -37,12 +39,48 @@ class AvataxPlugin(BasePlugin):
...
-        self._enabled = (
-            settings.AVATAX_USERNAME_OR_ACCOUNT and settings.AVATAX_PASSWORD_OR_LICENSE
-        )
+        self._cached_config_from_db = None
+        self.config = None
+        self.active = None
+
+    def _initialize_plugin_configuration(self):
+        plugin_config_qs = PluginConfiguration.objects.filter(name=self.PLUGIN_NAME)
+        plugin_config = self._cached_config_from_db or plugin_config_qs.first()
```
NyanKiyoshi: Couldn't we abstract the logic to `BasePlugin`?

korycins:
@NyanKiyoshi 
Each plugin can have a different configuration structure. It can have 2 or 5 or 7 different fields and because of that, the configuration is stored in DB. See model `PluginConfiguration`.
When someone integrates new plugin, like Avalara or Vatlayer, the active plugin needs to save its default configuration into DB. For most of them, it will be only a template of configuration structure that API will send to the Dashboard.  Shop owner can modify his plugin configuration directly from the Dashboard.
So to answer your question.  Plugin owner creates own configuration with as many config fields as he needs to have. He creates own default configuration that should be saved in DB. This config will be served by API.

saleor/extensions/models.py
```diff
@@ -0,0 +1,22 @@
...
+
+from saleor.core.utils.json_serializer import CustomJsonEncoder
+
+
+class PluginConfiguration(models.Model):
+    name = models.CharField(max_length=128, unique=True)
+    description = models.TextField(blank=True)
+    active = models.BooleanField(default=True)
+    configuration = JSONField(
+        blank=True, null=True, default=dict, encoder=CustomJsonEncoder
```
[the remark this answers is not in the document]
korycins: I would prefer to leave this field with `blank=True, null=True`. Some plugins can have configuration == `None`. It is good to explicitly add it. You can check a vatlayer's config.

saleor/extensions/plugins/avatax/plugin.py
```diff
@@ -292,17 +350,73 @@ def assign_tax_code_to_object_meta(
...
+        defaults = {
+            "name": cls.PLUGIN_NAME,
+            "description": "",
+            "active": False,
+            "configuration": [
+                {
+                    "name": "Username or account",
+                    "value": "",
+                    "type": ConfigurationTypeField.STRING,
+                    "help_text": "Provide user or account details",
```
maarcingebala: This should be also marked for translation I guess. Btw how are we going to switch the language when returning content like this in the API? The dashboard would be available in different languages. Should [we manually activate a language ](https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.translation.activate) in the resolver? cc @patrys

patrys: If we want the client to pass the language to the resolver, we can temporarily activate a language using `with translation.override(...):`.

saleor/graphql/extensions/types.py
```diff
@@ -0,0 +1,23 @@
+import graphene
+
+from ...extensions import models
+from ..core.connection import CountableDjangoObjectType
+from .enums import ConfigurationTypeFieldEnum
+
+
+class ConfigurationItem(graphene.ObjectType):
```
maarcingebala: Would be nice to add `class Meta` with a `description` field that explains briefly what this object represents.

PR #4597, opened 2019-08-05
No longer accept invalid IP addresses in core's get_client_ip

### LINKED ISSUE #4595: Which IP to retrieve from the environment variables
Currently we have this code to retrieve an IP:
https://github.com/mirumee/saleor/blob/35d8750ae07bddb57222f5fec3512c2788b56f5c/saleor/core/utils/__init__.py#L54-L58

Which is fine–apart the fact someone can inject any else than an IP.

The question is: should we really trust the `X-Forwarded-For` header? Shouldn't we let the web server handle this part and provide us the IP it believe is correct through `REMOTE_ADDR`?

For example, in nginx, we can get the environment variable `REALIP_REMOVE_ADDR` and thus leave the choice to the user of which IP(s) and header(s) are to be **trusted** and handle the IP **validation**. With the above snippet, we don't leave the choice to the user.

Another case, is when the user has their saleor workers behind a proxy, e.g. CloudFlare. The IP would be set from `CF-Connecting-IP` to `REALIP_REMOVE_ADDR` (nginx).

I believe we could do something like this:
```python
ip_headers = ["REALIP_REMOVE_ADDR", "WHATEVER_APACHE_USES", "REMOTE_ADDR"]

for header in ip_headers:
    ip = request.META.get(header, None)
    if ip and not valid_ip(ip):
        return ip
return None
```

NyanKiyoshi:
Closes #4595.

This fixes the error `ValueError: 'Oopsie?' does not appear to be an IPv4 or IPv6 address` when trying to get the user's country.

### Pull Request Checklist

### DIFF
--- a/saleor/core/utils/__init__.py
+++ b/saleor/core/utils/__init__.py
@@ -1,5 +1,6 @@
 import decimal
 import logging
+import socket
 from json import JSONEncoder
 from urllib.parse import urljoin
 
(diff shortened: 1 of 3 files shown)


django-queries:
Here is the report for 83efa146f3da3f06a01bb6214929365ad7a5fbc8 (NyanKiyoshi/saleor @ core/ip/check-if-valid)
Base comparison is 73135af4ce8362e53661eba2e0f6203e5760c2e1.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         48	         48	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	          9	          9	              2
```

</p>
</details>

saleor/core/utils/__init__.py
```diff
@@ -51,10 +52,37 @@ def build_absolute_uri(location):
...
+    return True
+
+
 def get_client_ip(request):
+    """Retrieve the IP address from the request data.
+
+    Tries to get the IP address from X-Forwarded-For, if the user is hiding behind
+    a transparent proxy. Otherwise, retrieve the requester IP.
+
+    If the forwarded IP is invalid, it fallback to the requester IP.
```
patrys: Most of the time this will return the IP address of our own reverse proxy (eg. CloudFront).

NyanKiyoshi: Are you asking about changing the comment or?

patrys: I'm talking about this fallback not being something we want. We should fall back to the first valid IP in the "forwarded for" chain, not discard the entire chain.

PR #4663, opened 2019-08-20
Drop JSONB mapping of attributes for M2M relations

### LINKED ISSUE #4643: Migrate the product attributes mapping to M2M
At beginning of Saleor (https://github.com/mirumee/saleor/pull/280) we implemented a JSONB mapping to dynamically store attribute data and/or attribute relations. It then became a HStore field for better indexing and as of #4351 was migrated back to JSONB for our new needs: multiple values per attribute.

A year ago, we dropped the support of having dynamic values assigned to a product/variant without having to create a new attribute value, e.g. `12 (features) => It's blue and shiny`. Instead we have implemented flags for the attributes and always create an attribute value for it into the database for consistency and to make life easier.

Now is the time to drop this custom M2M mapping for a proper and fully reliable relation through pSQL checks, e.g. the user could enter invalid data or relation through raw SQL and deleting an attribute or a value doesn't trigger a cascade deletion.

Overall, the performances shouldn't change as in both cases, the queries are meant to always it the pSQL indexes for optimal sorting. But now it will be more straightforward and prevent human errors when manually changing data into the database.

This migration should come in the same time as we release the new features for heavy filtering and heavy sorting of attributes to make sure the new relations are fully responding to our next needs before it hitting master.

---

Related pull requests to base the refactoring on:
- https://github.com/mirumee/saleor/pull/4255, migrated the logic from HStore to JSONB

---

Definition of done:
- [ ] The migrations are automatically tested against all cases
- [ ] Dashboard 1.0, and GraphQL support
- [ ] The old attribute logic in SEO is updated
- [ ] Storefront 1.0 support
- [ ] Populatedb is updated to assign attributes and values to the products/variants

### LINKED ISSUE #4674: Populatedb doesn't trigger the task to update the minimal prices after creating sales
1. Start a fresh DB
1. Run populatedb
1. The minimal price sorting is invalid (minimal prices are out of date)

### LINKED ISSUE #4685: Variant filters not working
Filters by vairant attribute

When i selected a variant attribute it gives the following message: 

"Sorry, no matches found for your request.
Try again or shop new arrivals."

The product matches all the variant types but it is not showing in the search result's page.

### What I expected to happen
Show the products that matches the selected variant attributes.

### Screenshots

Captura de tela de 2019-08-26 20-09-25

**System information**
Operating system:  Ubuntu 
Browser: Firefox and Google Chorme

NyanKiyoshi:
Closes #4643, closes #4674, fixes #4685.

- [x] The migrations are done
- [ ] ~~The migrations are automatically tested against all cases~~ 

- [x] Dashboard 1.0 support
- [x] GraphQL: mutations
- [x] GraphQL filtering
- [x] GraphQL tests are updated adequately 

- **GraphQL test cases checklist**
  - [x] Existing attribute value doesn't create a new one
  - [x] Attributes flagged as required returns a validation error if no value supplied
  - [x] Passing a non existing attribute ID and slug
  - [x] When no slug or ID was provided in input, raises a validation error (deprecated)
  - [x] The deprecated field `SelectAttribute.value` is tested
  - [x] Blank value triggers a validation error
  - [x] All variant attributes are required

- [x] The old attribute logic in SEO is updated
- [x] Storefront 1.0 support
- [x] unit tested?

- [x] Storefront 1.0 filtering support
- [x] unit tested?

- [x] Populatedb is updated to assign attributes and values to the products/variants

### Pull Request Checklist


templates/product/details.html
```diff
@@ -198,11 +198,13 @@ <h3>{% trans "Description" context "Product details title" %}</h3>
...
-        {% for attribute, value in product_attributes.items %}
-          <tr>
-            <td>{{ attribute.translated }}:</td>
-            <td><strong>{{ value }}</strong></td>
-          </tr>
+        {% for attribute_rel in product.attributes.all %}
+          {% with attribute=attribute_rel.attribute %}
+            <tr>
+              <td>{{ attribute.translated }}:</td>
+              <td><strong>{% for value in attribute_rel.values.all %}{{ value.translated }}{% endfor %}</strong></td>
```
NyanKiyoshi: Should be joined with a comma

django-queries:
Here is the report for 0ac7aa2baf0c599288c47a5883b439c29d91cbfb (NyanKiyoshi/saleor @ product/attributes/mapping)
Base comparison is 65e00b0d1497d329320b33c8bcdbe1fa8febfb9d.



<details><summary>**Found 2 differences!** (click me)</summary>
<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         48	         48	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
- product details                     	         13	         15	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
- retrieve variant list               	         15	         18	              8
```

</p>
</details>

saleor/product/utils/attributes.py
```diff
@@ -1,70 +1,66 @@
...
+    assigned will be overridden by this call.
     """
-    return " / ".join(
-        str(attribute_value)
-        for attribute_pk, attribute_value in sorted(
-            attributes_dict.items(), key=lambda x: x[0]
-        )
-    )
+    assignment = _associate_attribute_to_instance(instance, attribute.pk)
+    assignment.values.set(values)
```
maarcingebala: Does it allow to assign values of a different attribute than the one that was provided or does the relationship protects us from it?

saleor/graphql/core/utils/__init__.py
```diff
@@ -30,12 +34,17 @@ def validate_image_file(file, field_name):
...
+) -> str:
     """Resolve a node global id with a strict given type required."""
-    _type, _id = graphene.Node.from_global_id(global_id)
-    graphene_type = info.schema.get_type(_type).graphene_type
-    if graphene_type != only_type:
-        raise ValidationError({field: "Couldn't resolve to a node: %s" % global_id})
+    try:
+        _type, _id = graphene.Node.from_global_id(global_id)
+    except (binascii.Error, UnicodeDecodeError) as exc:
+        raise ValidationError({field: f"Invalid base64 global ID"}) from exc
```
maarcingebala: Since these errors might be returned to API users, I think it would be enough to tell them that the API failed to find a node of a given ID. Revealing the base64 decoding failed sounds like a low-level implementation detail.

saleor/graphql/product/mutations/products.py
```diff
@@ -469,6 +476,189 @@ class ProductCreateInput(ProductInput):
...
+            key = pks.get(attribute.pk, None)
+
+            if key is None:
+                key = slugs[attribute.slug]
+
+            cleaned_input.append((attribute, key))
+        return cleaned_input
+
+    @classmethod
+    def validate_input(cls, cleaned_input: T_INPUT_MAP, attribute_qs, is_variant: bool):
```
maarcingebala: Not sure how useful this method is - since you have to pass a right flag to trigger to flow you want, you could also just call the right method that is called underneath.

saleor/graphql/product/mutations/products.py
```diff
@@ -469,6 +476,189 @@ class ProductCreateInput(ProductInput):
...
+T_INPUT_MAP = List[Tuple[models.Attribute, List[str]]]
+T_INSTANCE = Union[models.Product, models.ProductVariant]
+
+
+class AttributeAssignmentMixin:
+    """Handles cleaning of the attribute input and creating the proper relations.
+
+    1. You should first call ``clean_input``, to transform and attempt to resolve
+       the provided input into actual objects.
+    2. You should then call ``is_valid`` with the cleaned input to perform a few
```
maarcingebala: Do we have the `is_valid` function here? If that's possible, I would try to stick to the convention that we have in model mutations, where `clean_input` is the only function you need to call to run input validation.

saleor/graphql/core/mutations.py
```diff
@@ -104,12 +106,40 @@ def _update_mutation_arguments_and_fields(cls, arguments, fields):
         cls._meta.fields.update(fields)
 
     @classmethod
-    def get_node_or_error(cls, info, node_id, field="id", only_type=None):
+    def get_node_by_id(
```
NyanKiyoshi:
```suggestion
    def get_node_by_pk(
```

saleor/graphql/core/mutations.py
```diff
@@ -104,12 +106,40 @@ def _update_mutation_arguments_and_fields(cls, arguments, fields):
...
+        cls, info, graphene_type: ObjectType, node_id: Union[int, str], qs=None
```
NyanKiyoshi:
```suggestion
        cls, info, graphene_type: ObjectType, pk: Union[int, str], qs=None
```

saleor/graphql/core/mutations.py
```diff
@@ -104,12 +106,40 @@ def _update_mutation_arguments_and_fields(cls, arguments, fields):
...
 
         try:
-            node = graphene.Node.get_node_from_global_id(info, node_id, only_type)
+            if only_type is not None:
+                internal_id = from_global_id_strict_type(
+                    node_id, only_type, field=field
+                )
+            else:
+                # FIXME: warn when supplied only_type is None?
+                only_type, internal_id = graphene.Node.from_global_id(node_id)
```
NyanKiyoshi:
```suggestion
                only_type, pk = graphene.Node.from_global_id(node_id)
```

saleor/product/utils/attributes.py
NyanKiyoshi: Yes, if one calls this directly with wrong values, it will still assign them. But it is the job of `AttributeAssignmentMixin` to check whether the value is from this attribute or not, otherwise it would really be heavy to check here.

saleor/graphql/product/mutations/products.py
```diff
@@ -469,6 +476,194 @@ class ProductCreateInput(ProductInput):
...
+                slugs[slug] = values
+            else:
+                raise ValidationError("You must whether supply an ID or a slug")
+
+        attributes = cls._get_attribute_nodes(
+            attributes_qs, global_ids=global_ids, pks=pks.keys(), slugs=slugs.keys()
+        )
+        cleaned_input = []
+        for attribute in attributes:
+            key = pks.get(attribute.pk, None)
```
korycins: maybe - 
`key = pks.get(attribute.pk, slugs[attribute.slug])`
?

saleor/graphql/product/types/attributes.py
```diff
@@ -6,6 +6,7 @@
 from graphql_jwt.decorators import permission_required
 
 from ....product import models
+from ....product.utils.attributes import T_ASSIGNMENT_REL
```
korycins: Maybe we could have own `typing` file for such structures? @maarcingebala  What do you think?

saleor/graphql/product/mutations/products.py
```diff
@@ -35,15 +40,17 @@
...
     Category,
     Collection,
     MoveProductInput,
     Product,
     ProductImage,
     ProductVariant,
 )
-from ..utils import attributes_to_json
+from ..utils import (
+    validate_attribute_input_for_product,
```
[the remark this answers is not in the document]
maarcingebala: If we wanted to introduce separate files for validators, I would rather do that once for the whole codebase.

saleor/product/utils/attributes.py
maarcingebala: I guess we could just iterate over `values` and see if `value.attribute` is the same as the one passed as an argument. That could work if the `values` are saved to the database at this point, not sure if they are.

maarcingebala: I'm still thinking what's the best way to keep the validation rules in Saleor, which are part of the business logic. I think that in general API layer should just call the right methods and validation should happen in some lower layer, like this. It's something to discuss.

saleor/graphql/core/mutations.py
```diff
@@ -330,11 +356,17 @@ def save(cls, info, instance, cleaned_input):
         instance.save()
 
     @classmethod
-    def get_instance(cls, info, **data):
+    def get_instance(cls, info, __qs: QuerySet = None, **data):
```
maarcingebala: Why the argument name `__qs`. Couldn't we have `queryset` or `qs`?

saleor/graphql/product/mutations/products.py
```diff
@@ -539,6 +744,18 @@ def clean_sku(cls, product_type, cleaned_input):
             elif models.ProductVariant.objects.filter(sku=input_sku).exists():
                 raise ValidationError({"sku": "Product with this SKU already exists."})
 
+    @classmethod
+    def get_instance(cls, info, __qs: QuerySet = None, **data):
+        """Prefetch related fields that are needed to process the mutation."""
+        # If we are updating an instance and want to update its attributes,
+        # prefetch them.
+        if data.get("id") and data.get("attributes"):
```
korycins: I am not sure that we really want to complicate whole mutation logic only to pass `qs` to the lowest layer of the `ModelMutation`. I think that maintaining it and understanding will cost us a lot.
Product with Attributes is a special case so maybe we should focus on modifying `get_instance` on the Product level without the need to applying changes to core Mutations also.
Maybe something similar to this?
```
    @classmethod
    def get_instance(cls, info, **data):
        """Prefetch related fields that are needed to process the mutation."""
        # If we are updating an instance and want to update its attributes,
        # prefetch them.
        if data.get("id") and data.get("attributes"):
            qs = Product.objects.prefetch_related(
                "product_type__product_attributes",
                "attributes__assignment__attribute__values",
            )
            object = qs.object(pk=data.get("id")).first()
            if object:
                return object
        return super().get_instance(info, **data)
```

NyanKiyoshi: Agreed.

saleor/graphql/product/mutations/products.py
NyanKiyoshi: Won't work because we don't know yet if the attribute was resolved through a global ID (base64 PK) or by a slug.

saleor/graphql/core/mutations.py
```diff
@@ -331,6 +356,10 @@ def save(cls, info, instance, cleaned_input):
 
     @classmethod
     def get_instance(cls, info, **data):
+        """Retrieve an instance from the supplied global id.
+
+        Optionally take a qs parameter to resolve the node using a custom query set.
```
maarcingebala: This is no longer valid.

saleor/graphql/product/types/attributes.py
maarcingebala: Not sure if we have enough things to put into such file. I'm also not sure what is the best place to keep this one. But what I am sure, is that this name is a little bit scary as for our codebase ;) Maybe let's rename it to something like `AttributeAssignmentType`?

PR #4669, opened 2019-08-21
Extract payment plugins

salwator:
This PRs purpose is to use the recently introduced plugin system to saleor payment gateways. All payment implementations will be turned into saleor plugins and configured through API instead of the environment.   

Current _Payment_ uses lazy imports and checks for implementation when processing payment. This kind of dispatch and parametrization is no longer useful since we can use _PluginManager_ interface and plugin mechanism for dispatching and establishing contract for payment flow process. This is why I introduce new simpler way of calling payment methods and pre/post transaction processing.

Business logic for payments will be encapsulated in payment.gateway module, which will execute actions on payment gateways using plugins. All payment related plugin hooks will be grouped into _PaymentInterface_ so it easy to test against and divide _PluginManager_ in the future.

We should make payment methods interface more verbose and probably since we're no longer forced to reuse same data structure as an argument for different actions, creating action-specific parameters and structure types makes more sense.

This PR provides both storefronts (old and new) with dynamic gateway list during checkout process, gateways can be enabled/disabled and entirely configured in saleor dashboard or directly through GraphQL plugin mutation.

NOTE: GraphQL schema is static, for now it is necessary to add new payment gateway plugin directly in settings and extend gateway list there to generate proper enum type in schema. Will be changed soon in next PR.

---------------
Roadmap:
- [x] Extend BasePlugin with payment methods
- [x] Introduce DummyGatewayPlugin
- [x] Implement BraintreeGatewayPlugin
- [x] Implement StripeGatewayPlugin
- [x] Implement RazorpayGatewayPlugin
- [x] Use properly gateway plugins in plugin manager
- [x] Introduce new _PaymentInterface_ abstract class so we can isolate payment stuff from the rest of manager methods.
- [x] Create _PaymentGateway_ and test with mocked _PaymentInterface_ 
- [x] Switch to _PaymentGateway_ in API:
  - [x] _process_payment_
  - [x] _authorize_
  - [x] _capture_
  - [x] _confirm_
  - [x] _refund_
  - [x] _void_
  - [x] _list_payment_sources_ 
- [x] Add _create_form_ to payment plugins for backward compatibility
- [x] Remove dead code from old payment architecture
- [x] Serve available gateway plugin list in checkout field
- [x] Serve available gateway plugin list in old storefront view

### Pull Request Checklist


django-queries:
Here is the report for 8c792059cdc868f5c508e364cbc031e33a0ba263 (mirumee/saleor @ extract-payment-plugins)
Base comparison is 246c788bbb9464d19235244c62e88744fe87c7e7.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         48	         48	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	         15	         15	              8
```

</p>
</details>

saleor/payment/gateways/braintree/plugin.py
```diff
@@ -0,0 +1,165 @@
...
+                store_customer=gateway_config["store_card"],
+            )
+            self.active = GATEWAY_NAME in settings.CHECKOUT_PAYMENT_GATEWAYS
+
+    @classmethod
+    def _get_default_configuration(cls):
+        defaults = {
+            "name": cls.PLUGIN_NAME,
+            "description": "",
+            "active": False,
```
korycins: Dynamic settings allow a user to enable or disable the plugin. Where do you handle it on the plugin side?

saleor/payment/utils.py
```diff
@@ -373,11 +373,14 @@ def validate_gateway_response(response: GatewayResponse):
...
-def _gateway_postprocess(transaction, payment):
+def gateway_postprocess(transaction, payment):
+    if not transaction.is_success:
+        return
+
     transaction_kind = transaction.kind
 
     if transaction_kind in {TransactionKind.CAPTURE, TransactionKind.CONFIRM}:
-        payment.captured_amount += transaction.amount
+        payment.captured_amount += Decimal(transaction.amount)
```
korycins: Based on `GatewayResponse` `amount` should be in `Decimal` format. If it has a different format then it should be changed during the creation of transaction. Do you agree?
And the second thing. Decimal without quantizing can be tricky.
```
In [18]: Decimal(1.1)
Out[18]: Decimal('1.100000000000000088817841970012523233890533447265625')
```

saleor/settings.py
```diff
@@ -671,7 +671,13 @@ def get_host():
 
 EXTENSIONS_MANAGER = "saleor.extensions.manager.ExtensionsManager"
 
+PAYMENT_PLUGINS = [
```
korycins: This should be taken from env

saleor/settings.py
```diff
@@ -671,7 +671,13 @@ def get_host():
...
+    "saleor.payment.gateways.dummy.plugin.DummyGatewayPlugin",
+    "saleor.payment.gateways.stripe.plugin.StripeGatewayPlugin",
+    "saleor.payment.gateways.braintree.plugin.BraintreeGatewayPlugin",
+]
 PLUGINS = os.environ.get("PLUGINS", [])
+PLUGINS += PAYMENT_PLUGINS
```
korycins: This can be problematic when someone by mistake adds payment plugins to `PLUGIN` from env.
What do you think?

tests/api/test_checkout.py
```diff
@@ -1058,32 +1068,61 @@ def test_checkout_complete(
...
+    error_side_effect,
     user_api_client,
     checkout_with_voucher,
     voucher,
     payment_dummy,
     address,
     shipping_method,
-    side_effect,
 ):
+    expected_error = "Oops! Something went wrong."
```
korycins: unused variable

tests/api/test_checkout.py
```diff
@@ -1107,12 +1146,12 @@ def test_checkout_complete_does_not_delete_checkout_after_unsuccessful_payment(
     response = user_api_client.post_graphql(MUTATION_CHECKOUT_COMPLETE, variables)
     content = get_graphql_content(response)
     data = content["data"]["checkoutComplete"]
```
korycins: Data is not used anymore

saleor/settings.py
salwator: I think we could filter duplicates here but the best way to that would probably be to prevent same plugin from being loaded more than once in manager.

saleor/core/payments.py
```diff
@@ -0,0 +1,60 @@
...
+    TODO: Create this in runtime based on available plugins
+    """
+
+    DUMMY = "dummy"
+    BRAINTREE = "braintree"
+    RAZORPAY = "razorpay"
+    STRIPE = "stripe"
+
+
+class PaymentInterface(ABC):
```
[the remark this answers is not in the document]
salwator: This is an interface, among other things forces client to use only relevant part of manager functionality, which is a step into dividing it into separate plugin managers in the future.

korycins:
@salwator FYI  `populate_db` is failing.

saleor/core/payments.py
```diff
@@ -0,0 +1,76 @@
...
+    @abstractmethod
+    def list_payment_gateways(self) -> List[Gateway]:
```
maarcingebala: It would be nice to add some docstrings about the purpose of each function. Not sure if we should have it here or in the extensions manager?

salwator: I would put docstring in Gateway since it is entry point for clients code.
Additionally some explanation might be put into interface.

saleor/core/payments.py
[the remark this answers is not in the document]
NyanKiyoshi: One thing for sure is that we need to make sure the other base plugins are consistent with this way of doing, i.e. not have in one base plugin an abstract class using `ABC` and in another one `raise NotImplementedError`

NyanKiyoshi: I wonder if we shouldn't have it called `Abstract` or `Base` instead of `Interface`? If we ever have a need of a non-abstract method.

salwator: This is not a _base_ for Payment, it serves as an interface for payment business logic.

saleor/core/payments.py
```diff
@@ -0,0 +1,72 @@
+from abc import ABC, abstractmethod
+from enum import Enum
+from typing import TYPE_CHECKING, List
+
+if TYPE_CHECKING:
+    from saleor.payment.interface import PaymentData, GatewayResponse, CustomerSource
+
+
+class Gateway(Enum):
```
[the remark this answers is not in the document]
salwator: I didn't want core logic (extensions manager) to depend on payment application. Both should depend on shared core types/interfaces.

saleor/core/payments.py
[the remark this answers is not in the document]
salwator: This is not _abc_ for plugin, those could be created independently. Right now we do not force payment gateways to implement all payment methods, _process_payment_ could be enough for some integration, that's why I did not create PaymentPluginInterface, at least for now 😉
Plugin system still needs some attention and rethinking way of grouping, recognizing types and enforcing method overrides for plugin types. I still consider creating separate plugin manager for payment gateways in near future.

saleor/payment/gateways/braintree/plugin.py
salwator: Now all payment plugins have decorators to avoid running payment methods when disabled.

saleor/settings.py
salwator: As I spoke with @maarcingebala , we're not convinced it is the best idea to inject python module paths through PATH, since now we still need to set thing up in saleor settings, will leave it manually created here.
After we implement automatic plugin discovery we might consider configuring additional plugin paths through env.

saleor/settings.py
NyanKiyoshi: Also, it is getting a string, not a list 😉

saleor/payment/utils.py
korycins: @salwator  waht about this?

saleor/extensions/base_plugin.py
```diff
@@ -136,6 +137,47 @@ def get_tax_rate_percentage_value(
     ) -> Decimal:
         return NotImplemented
 
+    def authorize_payment(
+        self, payment_information: "PaymentData", previous_value
+    ) -> "GatewayResponse":
+        pass
```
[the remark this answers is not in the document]
korycins: And the second thing. All payment methods use `__run_method_on_single_plugin` which uses `NotImplemented` to discover if plugin has given method. I would like to use `NotImplemented` to fit in the convention. Especially when we use the logic that operates on it.

saleor/extensions/manager.py
```diff
@@ -178,7 +185,101 @@ def postprocess_order_creation(self, order: "Order"):
...
+        gateway_name = gateway.value
+        gtw = self.get_plugin(gateway_name)
+        if gtw is not None:
+            return self.__run_method_on_single_plugin(
+                gtw, "list_payment_sources", default_value, customer_id=customer_id
+            )
+        raise Exception(f"Payment plugin {gateway_name} is inaccessible!")
+
+    def list_payment_gateways(self) -> List[Gateway]:
+        PAYMENT_METHOD = "process_payment"
```
korycins: It is local method variable. Shouldn't be a `payment_method`?

saleor/extensions/base_plugin.py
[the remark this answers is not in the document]
korycins: And the third thing. Returning `NotImplemented` tells manager that this method is not implemented on the plugin side.

saleor/extensions/manager.py
```diff
@@ -178,7 +185,101 @@ def postprocess_order_creation(self, order: "Order"):
...
+        gateway: Gateway,
+        method_name: str,
+        payment_information: "PaymentData",
+        **kwargs,
+    ) -> Optional["GatewayResposne"]:
+        default_value = None
+        gateway_name = gateway.value
+        gtw = self.get_plugin(gateway_name)
+        if gtw is not None:
+            return self.__run_method_on_single_plugin(
```
korycins: What about the scenario when someone creates PaymentPlugin but without mandatory method? We don't have any validation here. It will just do nothing. Shouldn't we have some validation here?

saleor/extensions/manager.py
```diff
@@ -178,7 +185,101 @@ def postprocess_order_creation(self, order: "Order"):
...
+                gtw,
+                method_name,
+                default_value,
+                payment_information=payment_information,
+                **kwargs,
+            )
+
+        raise Exception(f"Payment plugin {gateway_name} is inaccessible!")
```
korycins: I think that we should have some `PaymentException` class. We have a couple of places where it could be used.

saleor/payment/__init__.py
```diff
@@ -1,19 +1,20 @@
...
-
 
 class PaymentError(Exception):
     def __init__(self, message):
         super(PaymentError, self).__init__(message)
         self.message = message
 
 
+GATEWAYS_ENUM = Enum(
+    "GatewaysEnum", {key.upper(): key.lower() for key in settings.PAYMENT_GATEWAYS}
```
korycins: @salwator Maybe we could iterate over a list of payment plugins to get their names and generate this ENUM based on this? Thanks to it we will not need `PAYMENT_GATEWAYS` anymore.

saleor/payment/gateways/braintree/__init__.py
```diff
@@ -53,7 +53,9 @@ def get_customer_data(payment_information: PaymentData) -> Dict:
             "locality": billing.city,
             "region": billing.country_area,
             "country_code_alpha2": billing.country,
-        },
+        }
+        if billing
```
korycins: Lets create a billing structure in separate function or create billing before `return` section

saleor/settings.py
```diff
@@ -615,61 +615,12 @@ def get_host():
...
-            "store_card": get_bool_from_env("STRIPE_STORE_CARD", False),
-            "auto_capture": get_bool_from_env("STRIPE_AUTO_CAPTURE", True),
-            "template_path": "order/payment/stripe.html",
-            "connection_params": {
-                "public_key": os.environ.get("STRIPE_PUBLIC_KEY"),
-                "secret_key": os.environ.get("STRIPE_SECRET_KEY"),
-            },
-        },
-    },
+    DUMMY: {"template_path": "order/payment/dummy.html"},
```
korycins: Don't you think that `template_path` should be inside a plugin?

saleor/settings.py
```diff
@@ -615,61 +615,12 @@ def get_host():
...
+    BRAINTREE: {"template_path": "order/payment/braintree.html"},
+    RAZORPAY: {"template_path": "implement_me"},
```
korycins: `implement_me`? Previously we had templates for `RAZORPAY`

saleor/settings.py
```diff
@@ -615,61 +615,12 @@ def get_host():
...
+    STRIPE: {"template_path": "implement_me"},
```
korycins: `STRIPE` doesn't work for storefront1.0 so I assume that we should have `template_path` as a dynamic variable dedicated to plugin. What do you think?

saleor/settings.py
salwator: Will do that in next PR when removing static settings.

saleor/settings.py
salwator: Braintree does work for old storefront and uses this setting, will need to use dynamic setting anyway after removing this dict drom settings.

saleor/payment/utils.py
salwator: temporary hack, removed

saleor/payment/__init__.py
salwator: We can't easily do that right now and I don't really think it is ultimately the way to go, we might want to enable plugins in runtime in the future. I don't think API spec should depend on plugin discovery process 🤷‍♂️

saleor/extensions/manager.py
salwator: Sure, attempt to run _authorize_ or other payment method that we expect side effects should not fail silently. Changed.

saleor/extensions/manager.py
salwator: Good idea, there is a need to rethink exception handling in payment architecture after changes.

PR #4676, opened 2019-08-22
Add error codes to API

### LINKED ISSUE #4119: Return error codes in API
Investigate how to extend our `Error` type to include the error code. Currently, our client apps rely on the error message from the `Error` type and render it in the UI, but there's an idea that this message should be actually assembled client-side. My only concern is how to handle the multitude of specific errors that the server might return. 

A helpful resource about what error codes to provide might be this documentation:
https://cloud.google.com/apis/design/errors#handling_errors

(By using the word "error code" I'm thinking about the general concept, we could actually use an enum to represent various error types.)

Kwaidan00:
I want to merge this change because it adds error codes for mutation errors.

It closes #4119.

### Pull Request Checklist


saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,177 @@
...
+
+DJANGO_PASSWORD_VALIDATION_ERROR_CODES = [
+    "password_entirely_numeric",
+    "password_too_common",
+    "password_too_short",
+    "password_too_similar",
+]
+
+
+class AccountErrorCode(Enum):
```
NyanKiyoshi: I think it would be better to put them to the concerned packages, i.e. `AccountErrorCode` into `graphql.account`.

Not sure where. The current convention for enums is to put them in the `__init__.py` but I'm not sure if that's the best place, especially as we don't always expect to have other stuff than imports inside this file.

NyanKiyoshi: cc. @maarcingebala

Kwaidan00: I would be glad to hear your suggestion. I do not insist with that solution. :)

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,182 @@
...
+    NOT_PRODUCTS_IMAGE = "not_products_image"
+    PRODUCT_ALREADY_EXISTS = "product_already_exists"
+
+
+class ShippingErrorCode(Enum):
+    DEFAULT_SHIPPING_ZONE_ALREADY_EXISTS = "default_shipping_zone_already_exists"
+    MAX_LESS_THAN_MIN = "max_less_than_min"
+
+
+ERROR_CODE_UNKNOWN = "unknown"
```
maarcingebala: I'm not sure what I think about returning "unknown" code. First of all of shouldn't have such errors. Secondly, maybe instead we should just return `null` in that case?

saleor/graphql/menu/mutations.py
```diff
@@ -254,9 +280,12 @@ def clean_operation(operation: _MenuMoveOperation):
...
                     {
-                        "parent": (
-                            "Cannot assign a node as child of "
-                            "one of its descendants."
+                        "parent": ValidationError(
+                            (
+                                "Cannot assign a node as child of "
+                                "one of its descendants."
+                            ),
+                            code=MenuErrorCode.CANNOT_ASSIGN_NOTE,
```
maarcingebala: Looks like a typo, should be `CANNOT_ASSIGN_NODE`? But I think this one and `ASSIGN_MENU_ITEM_TO_ITSELF` are in fact the same error which tells you that this particular assignment is not allowed.

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,182 @@
...
+
+class AccountErrorCode(Enum):
+    ACTIVATE_OWN_ACCOUNT = "activate_own_account"
+    ACTIVATE_SUPERUSER_ACCOUNT = "activate_superuser_account"
+    DEACTIVATE_OWN_ACCOUNT = "deactivate_own_account"
+    DEACTIVATE_SUPERUSER_ACCOUNT = "deactivate_superuser_account"
+    DELETE_NON_STAFF_USER = "delete_non_staff_user"
+    DELETE_OWN_ACCOUNT = "delete_own_account"
+    DELETE_STAFF_ACCOUNT = "delete_staff_account"
+    DELETE_STUFF_ACCOUNT = "delete_stuff_account"
```
maarcingebala: `DELETE_STAFF_ACCOUNT ` vs `DELETE_STUFF_ACCOUNT `

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -302,7 +310,12 @@ def perform_mutation(cls, _root, info, **data):
                     lines_to_add.append((quantity, variant))
             else:
                 raise ValidationError(
-                    {"quantity": "Ensure this value is greater than or equal to 1."}
+                    {
+                        "quantity": ValidationError(
+                            "Ensure this value is greater than or equal to 1.",
+                            code=OrderErrorCode.QUANTITY_LESS_THAN_ONE,
```
maarcingebala: We have at least these codes that are similar: `POSITIVE_NUMBER_REQUIRED`, QUANTITY_LESS_THAN_ONE`. I think that instead, we should have a few generic ones such as `LESS_THAN`, `LESS_THAN_OR_EQUAL_TO` and the same for "greater" checks.

saleor/graphql/product/mutations/attributes.py
```diff
@@ -160,15 +171,29 @@ def clean_attribute(cls, instance, cleaned_input):
         if input_slug is None:
             cleaned_input["slug"] = slugify(cleaned_input["name"])
         elif input_slug == "":
-            raise ValidationError({"slug": "The attribute's slug cannot be blank."})
+            raise ValidationError(
+                {
+                    "slug": ValidationError(
+                        "The attribute's slug cannot be blank.",
+                        code=AttributeErrorCode.ATTRIBUTE_SLUG_BLANK,
```
maarcingebala: This is simply a "value required" type of error.

maarcingebala:
I did the first review and I think we have to look through the codes and think whether we can unify/generalize some of them. Another question is whether we need to start exposing them all at once. Maybe a better idea would be to have dedicated enums for each group e.g. `CheckoutErrorCodes`.

saleor/core/utils/promo_code.py
```diff
@@ -4,19 +4,29 @@
...
 
 
 class PromoCodeAlreadyExists(ValidationError):
     def __init__(self, message=None, **kwargs):
         if message is None:
-            message = {"promo_code": "Promo code already exists."}
+            message = {
+                "promo_code": ValidationError(
+                    "Promo code already exists.",
+                    code=CheckoutErrorCode.PROMO_CODE_ALREADY_EXISTS,
```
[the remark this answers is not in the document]
Kwaidan00: I don't think there is any benefit. In some context it could be more readable, but I think simplicity is better.
I've unified it.

saleor/graphql/menu/mutations.py
Kwaidan00: Yes, it was a typo.
I've unified those two error codes.

saleor/graphql/core/utils/error_codes.py
Kwaidan00: We should not, but something must be returned if some bad error code was used. 
For simplicity I've changed it to `null`.

saleor/graphql/account/mutations/account.py
```diff
@@ -120,7 +124,13 @@ def perform_mutation(cls, _root, info, **data):
 
         token = data.pop("token")
         if str(user.token) != token:
-            raise ValidationError({"token": INVALID_TOKEN})
+            raise ValidationError(
+                {
+                    "token": ValidationError(
+                        INVALID_TOKEN, code=AccountErrorCode.INVALID_USER_TOKEN
```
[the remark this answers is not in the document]
Kwaidan00: I've added generic invalid token code.

saleor/graphql/product/mutations/attributes.py
Kwaidan00: I've changed it to Django's `blank` error code.

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,182 @@
+from enum import Enum
+
+DJANGO_VALIDATORS_ERROR_CODES = [
```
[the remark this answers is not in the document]
Kwaidan00: I would be glad if we were able to do it prettier. Unfortunately, Django doesn't store any list of errors or something. Every code that throws `ValidationError` uses string literals as the value of code parameter.
On the other hand, we can not drop those codes, because there are returned by model and forms validation.

django-queries:
Here is the report for 43625e6f8eb5f6fe80ab829ab7061a08c18db103 (mirumee/saleor @ add-error-codes-in-api)
Base comparison is 1ac9086f38296b2986c2a884e780b20ff5be3355.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         48	         48	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	         15	         15	              8
```

</p>
</details>

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,169 @@
...
+    CAPTURE_INACTIVE_PAYMENT = "capture_inactive_payment"
+    DELETE_LINE_NON_DRAFT_ORDER = "delete_line_non_draft_order"
+    DELETE_NON_DRAFT_ORDER = "delete_non_draft_order"
+    EDIT_NON_DRAFT_ORDER = "edit_non_draft_order"
+    FULFILL_ORDER_LINE = "fulfill_order_line"
+    ORDER_INVALID_SHIPPING_METHOD = "order_invalid_shipping_method"
+    ORDER_NO_PAYMENT = "order_no_payment"
+    ORDER_NO_PRODUCTS = "order_without_products"
+    ORDER_NO_SHIPPING_ADDRESS = "order_no_shipping_address"
+    QUANTITY_LESS_THAN_ONE = "quantity_less_than_one"
```
patrys: This is not true. What we require is a non-zero quantity. We might as well allow decimals like 0.2 in the future. I propose that we name it `ZERO_QUANTITY` and fix the error message.

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,169 @@
...
+    REFUND_MANUAL_PAYMENT = "refund_manual_payment"
```
patrys: Does it need to be specific like that? How about `CANNOT_REFUND` to match the first two?

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,169 @@
...
+class OrderErrorCode(Enum):
+    CANNOT_CANCEL_FULFILLMENT = "cannot_cancel_fulfillment"
+    CANNOT_CANCEL_ORDER = "cannot_cancel_order"
+    CAPTURE_INACTIVE_PAYMENT = "capture_inactive_payment"
+    DELETE_LINE_NON_DRAFT_ORDER = "delete_line_non_draft_order"
+    DELETE_NON_DRAFT_ORDER = "delete_non_draft_order"
+    EDIT_NON_DRAFT_ORDER = "edit_non_draft_order"
+    FULFILL_ORDER_LINE = "fulfill_order_line"
+    ORDER_INVALID_SHIPPING_METHOD = "order_invalid_shipping_method"
+    ORDER_NO_PAYMENT = "order_no_payment"
```
patrys: `PAYMENT_MISSING`? `ORDER_NO_PAYMENT` sounds like a music group from Tokyo.

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,169 @@
...
+    NO_MENU_ITEM_PROVIDED = "no_item_provided"
+    TOO_MANY_MENU_ITEMS = "too_many_items"
+
+
+class OrderErrorCode(Enum):
+    CANNOT_CANCEL_FULFILLMENT = "cannot_cancel_fulfillment"
+    CANNOT_CANCEL_ORDER = "cannot_cancel_order"
+    CAPTURE_INACTIVE_PAYMENT = "capture_inactive_payment"
+    DELETE_LINE_NON_DRAFT_ORDER = "delete_line_non_draft_order"
+    DELETE_NON_DRAFT_ORDER = "delete_non_draft_order"
```
patrys: `CANNOT_DELETE`?

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,169 @@
...
+    INVALID_MENU_ITEM = "invalid_menu_item"
+    NO_MENU_ITEM_PROVIDED = "no_item_provided"
+    TOO_MANY_MENU_ITEMS = "too_many_items"
+
+
+class OrderErrorCode(Enum):
+    CANNOT_CANCEL_FULFILLMENT = "cannot_cancel_fulfillment"
+    CANNOT_CANCEL_ORDER = "cannot_cancel_order"
+    CAPTURE_INACTIVE_PAYMENT = "capture_inactive_payment"
+    DELETE_LINE_NON_DRAFT_ORDER = "delete_line_non_draft_order"
```
patrys: `CANNOT_DELETE`? Do we need a separate error code for lines and orders?

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,169 @@
...
+    DEACTIVATE_OWN_ACCOUNT = "deactivate_own_account"
+    DEACTIVATE_SUPERUSER_ACCOUNT = "deactivate_superuser_account"
+    DELETE_NON_STAFF_USER = "delete_non_staff_user"
+    DELETE_OWN_ACCOUNT = "delete_own_account"
+    DELETE_STAFF_ACCOUNT = "delete_staff_account"
+    DELETE_SUPERUSER_ACCOUNT = "delete_superuser_account"
+    INVALID_COUNTRY = "invalid_country"
+    INVALID_PASSWORD = "invalid_password"
+    INVALID_PHONE_NUMBER = "invalid_phone_number"
+    NOT_USERS_ADDRESS = "not_users_address"
```
patrys: Why is this useful? Do we ever need the UI to know that an address ID is correct but belongs to a different user?

patrys:
Can we plan a meeting where we go through all of these codes and determine whether they are correct or even needed?

Kwaidan00:
@patrys @maarcingebala I've created typed versions of `Error` type and separated enums. Please take a look.

saleor/graphql/product/mutations/digital_contents.py
```diff
@@ -165,7 +170,10 @@ def clean_input(cls, info, data):
             missing_fields = set(required_fields).difference(set(data))
             if missing_fields:
                 msg += "{}, " * len(missing_fields)
-                raise ValidationError(msg.format(*missing_fields))
+                raise ValidationError(
+                    msg.format(*missing_fields),
+                    code=ProductErrorCode.MISSING_CONFIGURATION_FIELDS,
```
maarcingebala: `REQUIRED` would work here I guess.

saleor/graphql/schema.graphql
```diff
@@ -2573,9 +2858,38 @@ input ProductCreateInput {
...
+
+enum ProductErrorCode {
+  ATTRIBUTE_ALREADY_ASSIGNED
+  ATTRIBUTE_BAD_VALUE
+  ATTRIBUTE_CANNOT_BE_ASSIGNED
+  ATTRIBUTE_DISABLED_VARIANTS
+  ATTRIBUTE_NON_ASSIGNABLE
+  ATTRIBUTE_VALUES_NOT_UNIQUE
+  GRAPHQL_ERROR
+  INCORRECT_VALUE
```
maarcingebala: Can't see where this is used. Maybe we could replace it with `INVALID`?

saleor/graphql/account/mutations/base.py
```diff
@@ -39,7 +41,29 @@ def can_edit_address(user, address):
...
 
-class SetPassword(CreateToken):
+class AccountErrorMixin:
+    account_errors = graphene.List(
+        graphene.NonNull(AccountError),
+        description="List of errors that occurred executing the mutation.",
+    )
+
+    @classmethod
+    def handle_typed_errors(cls, errors: list, **extra):
```
maarcingebala: For every mixing like this, we need to override the entire method and the implementation is the same in every case. What if we kept this implementation in the BaseMutation only and in the mixins just provided a constant like `ERROR_TYPE_CLASS=AccountError` etc that we could use in `BaseMutation`? I would experiment with something like this to avoid code duplication.

saleor/graphql/schema.graphql
```diff
@@ -5,27 +5,63 @@ schema {
...
+  DELETE_OWN_ACCOUNT
+  DELETE_STAFF_ACCOUNT
+  DELETE_SUPERUSER_ACCOUNT
+  GRAPHQL_ERROR
+  INVALID
+  INVALID_COUNTRY
+  INVALID_PASSWORD
+  INVALID_PHONE_NUMBER
+  INVALID_STOREFRONT_URL
+  INVALID_TOKEN
```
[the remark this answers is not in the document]
patrys: Not sure about the password: should we separate the cases of a password being invalid and a password being too weak? I'm thinking about the registration flow, where the user might want to know whether to add more random characters or to remove a character that is not allowed.

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,225 @@
...
+    VOID_INACTIVE_PAYMENT = "void_inactive_payment"
+    ZERO_QUANTITY = "zero_quantity"
+
+
+class ProductErrorCode(Enum):
+    ATTRIBUTE_ALREADY_ASSIGNED = "attribute_already_assigned"
+    ATTRIBUTE_BAD_VALUE = "attribute_bad_value"
+    ATTRIBUTE_CANNOT_BE_ASSIGNED = "attribute_cannot_be_assigned"
+    ATTRIBUTE_DISABLED_VARIANTS = "attribute_disabled_variants"
+    ATTRIBUTE_NON_ASSIGNABLE = "attribute_non_assignable"
```
patrys: How is this different from "cannot be assigned"?

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,225 @@
...
+    ATTRIBUTE_VALUES_NOT_UNIQUE = "attribute_values_not_unique"
+    GRAPHQL_ERROR = "graphql_error"
+    INCORRECT_VALUE = "incorrect_value"
+    INVALID = "invalid"
+    MISSING_CONFIGURATION_FIELDS = "missing_configuration_fields"
```
patrys: Shouldn't those fields have a `REQUIRED` error?

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,225 @@
...
+    MISSING_VALUE = "missing_value"
+    NOT_PRODUCTS_IMAGE = "not_products_image"
+    NULL = "null"
```
patrys: How is `NULL` different from `REQUIRED`?

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,225 @@
...
+    OBJECT_ALREADY_EXISTS = "object_already_exists"
+    OBJECT_DOES_NOT_EXIST = "object_does_not_exist"
+    REQUIRED = "required"
+    UNIQUE = "unique"
+    VALUE_ERROR = "value_error"
```
patrys: Is this not the same as `INVALID`?

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,225 @@
...
+    VARIANT_NO_DIGITAL_CONTENT = "variant_no_digital_content"
+
+
+class ShopErrorCode(Enum):
+    CANNOT_FETCH_TAX_RATES = "cannot_fetch_tax_rates"
+    GRAPHQL_ERROR = "graphql_error"
+    INVALID = "invalid"
+    NULL = "null"
+    OBJECT_ALREADY_EXISTS = "object_already_exists"
+    OBJECT_DOES_NOT_EXIST = "object_does_not_exist"
```
patrys: Those two look strangely generic.

saleor/graphql/account/mutations/base.py
Kwaidan00: I wanted to do it that way, but I've stuck with that at some point. I'll refactor it.

saleor/graphql/schema.graphql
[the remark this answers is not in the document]
Kwaidan00: I've kept invalid password error code and I've merged all the others.

saleor/graphql/account/mutations/base.py
Kwaidan00: I've added the second field to the `cls`, for typed error field name. I think it is more readable than e..g searching class attribute with some special name.

saleor/graphql/core/utils/error_codes.py
Kwaidan00: Is it bad?

patrys: When would we return these? Not just for this enum, for all enums. There's usually a better way to indicate what went wrong like returning `CUSTOMER_ALREADY_EXISTS` or returning `ALREADY_EXISTS` accompanied by a proper field name.

saleor/graphql/account/mutations/base.py
```diff
@@ -39,7 +41,21 @@ def can_edit_address(user, address):
     )
 
 
-class SetPassword(CreateToken):
+class AccountErrorMixin:
+    account_errors = graphene.List(
+        graphene.NonNull(AccountError),
+        description="List of errors that occurred executing the mutation.",
+    )
+    ERROR_TYPE_CLASS = AccountError
```
maarcingebala: Maybe let's just add one empty line above, to visually separate fields declaration from those mixin settings.

Kwaidan00: I've added a line :)

saleor/graphql/core/utils/error_codes.py
Kwaidan00: I think all necessary information is given with that error code by attaching it to the correct field. However, I've replaced `OBJECT_ALREADY_EXISTS` with more specific error codes, as you suggested.

Refactoring of `OBJECT_DOES_NOT_EXIST ` is more complicated. I think e.g. `get_node_or_error` should be rewritten and should create field-specific codes - or we have to remain some generic error code for that. I think it could be done via separated PR.

patrys: Yeah, I guess `get_node_or_error` should start accepting error codes.

maarcingebala: But is the current implementation acceptable or is this something that we definitely want to change before merging?

saleor/graphql/core/utils/error_codes.py
maarcingebala: Fixed, now they sit next to models.

maarcingebala:
@patrys @NyanKiyoshi If you could take a look - I've made few changes: moved error codes definitions to separate files and replaced:
- `OBJECT_DOES_NOT_EXIST` with `NOT_FOUND`
- `OBJECT_ALREADY_EXISTS` with `ALREADY_EXISTS` (plus unified a few other "already exists" errors)

saleor/core/exceptions.py
```diff
@@ -1,4 +1,8 @@
+from ..checkout.error_codes import CheckoutErrorCode
```
korycins: Core module should not take any dependencies from other Saleor's module.

saleor/core/utils/promo_code.py
```diff
@@ -9,14 +9,19 @@
 class InvalidPromoCode(ValidationError):
     def __init__(self, message=None, **kwargs):
         if message is None:
-            message = {"promo_code": "Promo code is invalid"}
+            message = {
+                "promo_code": ValidationError("Promo code is invalid", code="invalid")
```
korycins: Shouldn't we use `ErrorCode` class for `code` here?

saleor/core/utils/promo_code.py
```diff
@@ -9,14 +9,19 @@
...
+            }
         super().__init__(message, **kwargs)
 
 
 class PromoCodeAlreadyExists(ValidationError):
     def __init__(self, message=None, **kwargs):
+        code = kwargs.get("code", "promo_code_already_exists")
```
korycins: Shouldn't we use `ErrorCode` class here?

saleor/core/utils/url.py
```diff
@@ -15,10 +17,14 @@ def validate_storefront_url(url):
         parsed_url = urlparse(url)
         domain, _ = split_domain_port(parsed_url.netloc)
     except ValueError as error:
-        raise ValidationError({"redirectUrl": str(error)})
+        raise ValidationError(
+            {"redirectUrl": str(error)}, code=AccountErrorCode.INVALID
```
korycins: Same as above. We shouldn't import any saleor's modules in core. This is a perfect wat to get more circular imports.
If we need to import `AccountErrorCode` it means that this class or `validate_storefront_url` is declared in wrong place.

saleor/core/weight.py
```diff
@@ -98,7 +98,8 @@ def validate(self, weight):
                 raise Exception("%r is not a valid weight." % (weight,))
             if weight.unit != unit:
                 raise forms.ValidationError(
-                    "Invalid unit: %r (expected %r)." % (weight.unit, unit)
+                    "Invalid unit: %r (expected %r)." % (weight.unit, unit),
+                    code="invalid",
```
korycins: Shouldn't we use `ErrorCode` class here?

saleor/graphql/account/mutations/base.py
```diff
@@ -137,19 +177,26 @@ def perform_mutation(cls, _root, info, **data):
...
+                        "Old password isn't valid.",
+                        code=AccountErrorCode.INVALID_PASSWORD,
+                    )
+                }
+            )
         try:
             password_validation.validate_password(new_password, user)
         except ValidationError as error:
-            raise ValidationError({"new_password": error.messages})
+            raise ValidationError({"new_password": error})
```
korycins: Shouldn't return error code here?

saleor/graphql/giftcard/mutations.py
```diff
@@ -8,12 +8,29 @@
...
 from .types import GiftCard
 
 
+class GiftcardErrorMixin:
+    gift_card_errors = graphene.List(
+        graphene.NonNull(GiftcardError),
+        description="List of errors that occurred executing the mutation.",
+    )
+
+    ERROR_TYPE_CLASS = GiftcardError
```
korycins: `GiftcardError` -> `GiftCardError`

saleor/graphql/giftcard/mutations.py
```diff
@@ -8,12 +8,29 @@
...
 from ...giftcard import models
+from ...giftcard.error_codes import GiftcardErrorCode
 from ...giftcard.utils import activate_gift_card, deactivate_gift_card
 from ..core.mutations import BaseMutation, ModelMutation
 from ..core.scalars import Decimal
+from ..core.types.common import GiftcardError
 from .types import GiftCard
 
 
+class GiftcardErrorMixin:
```
korycins: `GiftcardError` -> `GiftCardError`

saleor/graphql/menu/mutations.py
```diff
@@ -7,16 +7,33 @@
...
 from ...product import models as product_models
 from ..core.mutations import BaseMutation, ModelDeleteMutation, ModelMutation
+from ..core.types.common import MenuError
 from ..page.types import Page
 from ..product.types import Category, Collection
 from .enums import NavigationType
 from .types import Menu, MenuItem, MenuItemMoveInput
 
 
+class MenuErrorMixin:
```
korycins: Do we really want to declare and maintain base mutation for each module?
It seems scary for me. 
Why don't we use `__init_subclass_with_meta__` in BaseMutation to dynamically declare expected fields for the given class?.
```
class MenuUpdate(ModelMutation):
    class Meta:
        ERROR_TYPE_CLASS = MenuError
        ERROR_TYPE_FIELD = "menu_errors"
        error_type = MenuError
```
`__init_subclass_with_meta__` could use this fields to generate error structure. This will add some magic to base mutations but even though I think that this is better than declaring new base mutation per module.

saleor/graphql/account/mutations/base.py
Kwaidan00: `error` here is a `ValidationError` instance and it should already have error code.

saleor/account/error_codes.py
```diff
@@ -0,0 +1,22 @@
+from enum import Enum
+
+
+class AccountErrorCode(Enum):
```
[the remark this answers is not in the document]
maarcingebala: If some of them are common to ALL enums, then we could have one base enum for that.

saleor/core/exceptions.py
maarcingebala: Ideally yes; in this case, the exception is either used in `product` / `checkout` or `order` modules. Maybe instead of `core`, it should be defined inside one of them (not sure which one though). We will anyway end up importing it between those modules. @korycins @k-brk  If you have better idea then we can move it, if not I would leave it here.

saleor/graphql/product/mutations/attributes.py
```diff
@@ -239,7 +263,13 @@ def clean_remove_values(cls, cleaned_input, instance):
         for value in remove_values:
             if value.attribute != instance:
                 msg = "Value %s does not belong to this attribute." % value
-                raise ValidationError({"remove_values": msg})
+                raise ValidationError(
+                    {
+                        "remove_values": ValidationError(
+                            msg, code=ProductErrorCode.ATTRIBUTE_BAD_VALUE
```
NyanKiyoshi:
```suggestion
                            msg, code=ProductErrorCode.INVALID
```

saleor/graphql/product/mutations/attributes.py
```diff
@@ -387,7 +426,12 @@ def perform_mutation(
 
         if variant_attrs_pks and not product_type.has_variants:
             raise ValidationError(
-                {"operations": "Variants are disabled in this product type."}
+                {
+                    "operations": ValidationError(
+                        "Variants are disabled in this product type.",
+                        code=ProductErrorCode.ATTRIBUTE_DISABLED_VARIANTS,
```
NyanKiyoshi:
```suggestion
                        code=ProductErrorCode.ATTRIBUTE_VARIANTS_DISABLED,
```

saleor/graphql/product/mutations/products.py
```diff
@@ -1157,7 +1229,12 @@ def perform_mutation(cls, _root, info, image_id, variant_id):
             )
         except models.VariantImage.DoesNotExist:
             raise ValidationError(
-                {"image_id": "Image is not assigned to this variant."}
+                {
+                    "image_id": ValidationError(
+                        "Image is not assigned to this variant.",
+                        code=ProductErrorCode.NOT_PRODUCTS_IMAGE,
```
NyanKiyoshi:
```suggestion
                        code=ProductErrorCode.NOT_FOUND,
```

saleor/graphql/account/mutations/staff.py
```diff
@@ -210,11 +212,21 @@ def clean_is_active(cls, is_active, instance, user):
...
+                        )
+                    }
                 )
             elif instance.is_superuser:
                 raise ValidationError(
-                    {"is_active": "Cannot deactivate superuser's account."}
+                    {
+                        "is_active": ValidationError(
+                            "Cannot deactivate superuser's account.",
+                            code=AccountErrorCode.DEACTIVATE_SUPERUSER_ACCOUNT,
```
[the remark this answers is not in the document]
patrys: 🤔 Wouldn't that mean that most if not all error codes would start with CANNOT? I think I'd prefer to avoid negatives and use errors like `ACCOUNT_PROTECTED`. Ideally, the error code could be used to finish the following sentence: "the operation failed because...".

saleor/account/error_codes.py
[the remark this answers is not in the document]
Kwaidan00: @maarcingebala We discussed with @korycins and we decided to change enums to classes. The problem is that you cannot extend enum with fields (https://docs.python.org/3/library/enum.html#restricted-enum-subclassing). We don't use any enums special logic, so class-based approach would resolve that problem.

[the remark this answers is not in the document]
patrys: This sound counterintuitive. Subclassing means it's hard to see all members without executing code and custom classes cannot be used for typing.

saleor/graphql/product/mutations/products.py
Kwaidan00: @NyanKiyoshi are you sure it's the same? There is a different between non-existing image and existing image that not belongs to that product.

saleor/graphql/order/utils.py
```diff
@@ -16,15 +25,25 @@ def validate_shipping_method(order):
     )  # noqa
     if shipping_not_valid:
         raise ValidationError(
-            {"shipping": "Shipping method is not valid for chosen shipping address"}
+            {
+                "shipping": ValidationError(
+                    "Shipping method is not valid for chosen shipping address",
+                    code=OrderErrorCode.ORDER_INVALID_SHIPPING_METHOD,
```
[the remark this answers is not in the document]
Kwaidan00: I would prefer to have a separated error code here. It's an important part of checkout process and I think it should have that code.

saleor/graphql/core/utils/__init__.py
```diff
@@ -27,15 +27,24 @@ def str_to_enum(name):
...
     """Resolve a node global id with a strict given type required."""
     _type, _id = graphene.Node.from_global_id(global_id)
     graphene_type = info.schema.get_type(_type).graphene_type
     if graphene_type != only_type:
-        raise ValidationError({field: "Couldn't resolve to a node: %s" % global_id})
+        raise ValidationError(
+            {
+                field: ValidationError(
+                    "Couldn't resolve to a node: %s" % global_id,
+                    code="object_does_not_exist",
```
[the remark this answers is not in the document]
Kwaidan00: Yes, it's hardcoded - moreover, with bad value. 
But I think it must be hardcoded - we cannot extend some base enum, so I cannot decide which `SthErrorCode` should I import to that module.
The alternative way is passing correct enum when that function is called - but I'm not sure it's good idea.

saleor/graphql/menu/mutations.py
Kwaidan00: The last commit brings that implementation. ;)

saleor/account/error_codes.py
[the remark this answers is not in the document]
Kwaidan00: Ok, I won't change it.

saleor/core/utils/url.py
Kwaidan00: It will stay as it is for now; I've created a ticket for that: #4744

saleor/core/utils/promo_code.py
Kwaidan00: I've added an import from `GiftCardError` here; PR for changing that: #4744

saleor/core/utils/promo_code.py
Kwaidan00: As above.

saleor/core/weight.py
Kwaidan00: I cannot see any way to avoid it for now; I've created a ticket for that, as we discussed: #4745

saleor/graphql/core/utils/__init__.py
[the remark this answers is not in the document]
Kwaidan00: @NyanKiyoshi I've created a ticket for later refactoring: #4745

saleor/core/exceptions.py
Kwaidan00: I've created a ticket for that: #4744 
We can decide it later.

saleor/graphql/core/utils/error_codes.py
```diff
@@ -0,0 +1,73 @@
...
+    ShopErrorCode,
+]
+
+saleor_error_codes = []
+for enum in SALEOR_ERROR_CODE_ENUMS:
+    saleor_error_codes.extend([code.value for code in enum])
+
+
+def get_error_code_from_error(error):
+    """Return valid error code."""
```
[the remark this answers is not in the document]
Kwaidan00: I've added it.

Kwaidan00:
@maarcingebala @korycins @NyanKiyoshi I've changed error code mechanism to use `Meta` class instead of inheritance, for some of your comments I've created separated issues, I've rebased branch and added error codes for new parts of the codebase.

saleor/graphql/order/utils.py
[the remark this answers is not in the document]
NyanKiyoshi: Fair enough. Maybe we could say `SHIPPING_METHOD_NOT_APPLICABLE` or something like this? As it is not really an invalid shipping method but it's simply no applying to the active checkout.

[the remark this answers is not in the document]
Kwaidan00: You're right; I've changed it to `SHIPPING_METHOD_NOT_APPLICABLE`.

Issue #4683, opened 2019-08-26
Product attributes not visible on product details page


likewhoa:
### What I'm trying to achieve
See product attributes values on the product details page.

### Steps to reproduce the problem
1. Fresh master clone
2. Add two shirts with various size and color attributes
3. Go to product details page
4. No attributes under product description

### What I expected to happen
An html table as shown in https://github.com/mirumee/saleor/blob/65e00b0d1497d329320b33c8bcdbe1fa8febfb9d/templates/product/details.html#L200-L207 showing the product attribute name and value.

Python-3.7.4
Gentoo Linux kv 5.0.13
Chromium-77.0.3865.35

maarcingebala:
I'm on newest master after running `populatedb` on fresh DB. Here is my product details page:
![image](https://user-images.githubusercontent.com/5421321/63835974-e662bd80-c978-11e9-8102-972216aecefe.png)

"Material: Cotton" is a product attribute in this case. If I understand correctly, it isn't rendered in your case?

likewhoa:
> I'm on newest master after running `populatedb` on fresh DB. Here is my product details page:
> ![image](https://user-images.githubusercontent.com/5421321/63835974-e662bd80-c978-11e9-8102-972216aecefe.png)
> 
> "Material: Cotton" is a product attribute in this case. If I understand correctly, it isn't rendered in your case?

Try adding a new product with a custom attribute though. I only had issues with new products not existing ones.

NyanKiyoshi:
Hi, @likewhoa thanks for the finding. I created a patch in https://github.com/mirumee/saleor/pull/4705. The issue was that the field `visible_in_storefront` of the `Attribute` model (`product_attribute`) was being set to `False` by default by the dashboard 1.0 due to mistake that was on the form.

To make your affected attributes appear, you have multiple possibilities:
- You could delete and recreate the attribute, the values, reassign the values to the products, etc. (hardest solution)
- You could update all attributes to be visible through SQL (`./manage.py dbshell`):
```sql
UPDATE product_attribute SET visible_in_storefront = TRUE WHERE visible_in_storefront is FALSE;
```
- You could use the [saleor-dashboard](https://github.com/mirumee/saleor-dashboard), go to the "Configuration" section, and in the attributes page, toggle the field `Visible on Product Page in Storefront` of the affected attributes and hit save.
[...]

PR #4687, opened 2019-08-27
Fix to prevent saving attributes from clearing selected product type with 1.0 Dashboard

petedermott:

Currently if you save an attribute using the 1.0 Dashboard it will clear out that attributes product type with blank data.

This PR will exclude any product type related fields ("product_types", "product_variant_types") from the AttributeForm to ensure that this information is not overridden and the selected product types are retained.

### Pull Request Checklist

### DIFF
--- a/saleor/dashboard/product/forms.py
+++ b/saleor/dashboard/product/forms.py
@@ -527,7 +527,7 @@ class VariantImagesSelectForm(forms.Form):
 class AttributeForm(forms.ModelForm):
     class Meta:
         model = Attribute
-        exclude = ["input_type"]
+        exclude = ["input_type", "product_types", "product_variant_types"]
         labels = {
             "name": pgettext_lazy("Product display name", "Display name"),
             "slug": pgettext_lazy("Product internal name", "Internal name"),


django-queries:
Here is the report for ec249714d99f6ecfbf6944862f78df538a76d3dc (fermorltd/saleor @ master)
Base comparison is 65e00b0d1497d329320b33c8bcdbe1fa8febfb9d.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         48	         48	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	         15	         15	              8
```

</p>
</details>

saleor/dashboard/product/forms.py
```diff
@@ -527,7 +527,7 @@ def save(self):
 class AttributeForm(forms.ModelForm):
     class Meta:
         model = Attribute
-        exclude = ["input_type"]
+        exclude = ["input_type", "product_types", "product_variant_types"]
```
NyanKiyoshi: `exclude` doesn't make sense here as we only want to allow to change two fields. Instead we should only select the two fields we actually use and expect:

```suggestion
        fields = ["name", "slug"]
```

petedermott: Agreed if those are the only two fields you need. In our project we are also including the new value_required, visible_in_storefront, filterable_in_storefront fields in our Dashboard. If we also want to expose these to Dash 1.0 users I'm happy to put together a simple PR for it.

maarcingebala: I agree with @NyanKiyoshi. For now, I would go with allowing only those two fields.

NyanKiyoshi:
Hi! Thank you for your finding and your pull request! I will open a new pull request with the suggested change it is quite major (#4683).

PR #4689, opened 2019-08-27
Add logic to grant access to api through a token

korycins:
This PR introduces service account authentication. Staff user with service_account permission can create new service_account with the given name. `Create` mutation returns the auth_token, which should be passed to external service that requires access to our resources.  Staff user specifies permissions for a service_account. 
Service_account has access only to specified resources (based on permissions). 
All service accounts requests are identifying by `HTTP_AUTHORIZATION` header. If a request contains `bearer` token, the middleware will try to assign service_account obj to request. If service_account has permission to given resources,  it will return the data. 

----

- Add `Bot` model with django permissions
- Add query `bot`
- Add query `bots`
- Add mutation `botCreate`
- Add mutation `botUpdate`
- Add mutation `botDelete`
- Add mutation `botUpdatePrivateMeta`
- Add mutation `botClearStoredPrivateMeta`
- Add middleware to append bot based on token `saleor.graphql.middleware.bot_middleware`
- Modify default permision to grant access to bot `saleor.graphql.decorators.permission_required`


jyoost:
Hi Maciej,

I have been working on the "core django" saleor code for about 2 months
now.  I have totally rearchitected it for my own purposes using
materializecss (like dashboard 1.0) . while in the process I ripped out ALL
the bootstrap, nodejs, and webpack loader. My current storefront and
dashboard have no dependencies on any  of them.

To this end I have also developed a "plug-in" methodology to allow my model
to work with the 2.0 stuff moving forward.

I have changed the dashboard 1.0 code to work with this plug-in methodology.

I have also developed a helpdesk module and a shipping module with shipping
labels that works with my dashboard and the existing order/fulfillment
models.

I would like to do a pull on the current shipping module and add one model
called "Shipment" as the only physical tie in to the current system. It is
related to Order, Fulfillment and User.

All other tie ins are logical depending on what plug-in you have loaded in
url "namespace".

I have 35 years experience in the computer industry, and am really enjoying
working on your code. It is the best written system I have ever worked on
and have seen your comments and agree with them. My favorite is the
phonenumberfield where you commented "this should be a multiwidget we are
back to this!" .  I just went through all the i18naddress stuff and
unraveled it. (Still unraveling) needs fixed Bad.

Sincerely,

John

saleor/account/models.py
```diff
@@ -189,6 +191,48 @@ def get_ajax_label(self):
         return self.email
 
 
+class Bot(models.Model):
+    name = models.CharField(max_length=128)
```
NyanKiyoshi: That's a very long bot name

saleor/account/models.py
```diff
@@ -189,6 +191,48 @@ def get_ajax_label(self):
         return self.email
 
 
+class Bot(models.Model):
```
NyanKiyoshi: We might want to allow metadata for bots? That would be useful to add some data such as urls, descriptions, etc.

saleor/account/models.py
```diff
@@ -189,6 +191,48 @@ def get_ajax_label(self):
...
+
+    def _get_permissions(self):
+        """Return the permissions of the bot."""
+        if not self.is_active:
+            return set()
+        perm_cache_name = "_bot_perm_cache"
+        if not hasattr(self, perm_cache_name):
+            perms = self.permissions.all()
+            perms = perms.values_list("content_type__app_label", "codename").order_by()
+            setattr(self, perm_cache_name, {"%s.%s" % (ct, name) for ct, name in perms})
```
NyanKiyoshi:
```suggestion
            setattr(self, perm_cache_name, {f"{ct}.{name}" for ct, name in perms})
```

saleor/account/models.py
```diff
@@ -189,6 +191,48 @@ def get_ajax_label(self):
...
+        related_name="bot_set",
+        related_query_name="bot",
+    )
+
+    class Meta:
+        permissions = (
+            ("manage_bots", pgettext_lazy("Permission description", "Manage bots")),
+        )
+
+    def _get_permissions(self):
```
NyanKiyoshi: Does it return `Set[str]`? Would be useful to have a type hint.

saleor/account/models.py
```diff
@@ -189,6 +191,48 @@ def get_ajax_label(self):
...
+        perm_cache_name = "_bot_perm_cache"
+        if not hasattr(self, perm_cache_name):
+            perms = self.permissions.all()
+            perms = perms.values_list("content_type__app_label", "codename").order_by()
+            setattr(self, perm_cache_name, {"%s.%s" % (ct, name) for ct, name in perms})
+        return getattr(self, perm_cache_name)
+
+    def has_perms(self, perm_list):
+        """Return True if the bot has each of the specified permissions."""
+        return all(self.has_perm(perm) for perm in perm_list)
```
NyanKiyoshi: Very heavy... What an operation over a set?

```python
wanted_perms = set(perm_list)
actual_perms = self._get_permissions()

if not self.is_active:
    return False

return (wanted_perms & actual_perms) == wanted_perms

File: saleor/graphql/account/filters.py [hunk 6]
```diff
@@ -69,6 +69,13 @@ def filter_search(qs, _, value):
     return qs
 
 
+def filter_bot_search(qs, _, value):
+    search_fields = ("name", "is_active")
```
NyanKiyoshi: Do we really want to match on `is_active`? Wouldn't it be better to have a boolean filter for it?

korycins:
Hi @jyoost!
We are really happy that you like Saleor so much!

saleor/account/models.py
korycins: It is not heavy. It operates on cached perms. I will change it anyway

django-queries:
Here is the report for cd837bb49134e0af38082bdb3a8391846d443518 (korycins/saleor @ feature/bot_authentication)
Base comparison is 62b3f6ab29003ea1fa018c2b2ad1a81ae48ed4da.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         48	         48	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	         15	         15	              8
```

</p>
</details>

tests/api/test_middleware.py
```diff
@@ -0,0 +1,24 @@
+from unittest.mock import Mock
+
+import pytest
+from django.urls import reverse
+
+from saleor.graphql.middleware import service_account_middleware
+
+
+@pytest.mark.parametrize("path, should_accept", [("api", True), ("home", False)])
```
salwator: If one of test cases parameters is bool and we have condition based on it inside, there is strong sign that those are two separate test cases in reality.
This is not really huge scenario so maybe we could have to, well named tests for _api_ and _home_ ?

saleor/account/models.py
```diff
@@ -182,6 +186,57 @@ def get_ajax_label(self):
...
+        )
+
+    def _get_permissions(self) -> Set[str]:
+        """Return the permissions of the service."""
+        if not self.is_active:
+            return set()
+        perm_cache_name = "_service_perm_cache"
+        if not hasattr(self, perm_cache_name):
+            perms = self.permissions.all()
+            perms = perms.values_list("content_type__app_label", "codename").order_by()
```
NyanKiyoshi: Is the empty `order_by` accidental?

saleor/account/models.py
```diff
@@ -182,6 +186,57 @@ def get_ajax_label(self):
...
+            setattr(self, perm_cache_name, {f"{ct}.{name}" for ct, name in perms})
+        return getattr(self, perm_cache_name)
+
+    def has_perms(self, perm_list):
+        """Return True if the service has each of the specified permissions."""
+        wanted_perms = set(perm_list)
+        actual_perms = self._get_permissions()
```
NyanKiyoshi: We don't need to retrieve the data and do operations on the input if the bot is not active 👍

saleor/graphql/account/mutations/service_account.py
```diff
@@ -0,0 +1,115 @@
...
+    ModelDeleteMutation,
+    ModelMutation,
+    UpdateMetaBaseMutation,
+)
+
+
+class ServiceAccountInput(graphene.InputObjectType):
+    name = graphene.types.String(description="Name of the service account")
+    is_active = graphene.types.Boolean(
+        description="Determine if service account should be enabled"
```
NyanKiyoshi:
```suggestion
        description="Determine if this service account should be enabled"
```

saleor/graphql/account/mutations/service_account.py
```diff
@@ -0,0 +1,115 @@
...
+        # clean and prepare permissions
+        if "permissions" in cleaned_input:
+            permissions = cleaned_input.pop("permissions")
+            cleaned_input["permissions"] = get_permissions(permissions)
+        return cleaned_input
+
+    @classmethod
+    def perform_mutation(cls, root, info, **data):
+        instance = cls.get_instance(info, **data)
+        data = data.get("input")
```
NyanKiyoshi: `data["input"]` instead as this field is required thus non-nullable?

_Interesting fact: apparently the cost difference is omissible (https://wiki.python.org/moin/PythonSpeed/PerformanceTips#Initializing_Dictionary_Elements)_

saleor/graphql/account/mutations/service_account.py
```diff
@@ -0,0 +1,115 @@
...
+        description = "Updates an existing service account"
+        model = models.ServiceAccount
+        permissions = ("account.manage_service_account",)
+
+    @classmethod
+    def clean_input(cls, info, instance, data):
+        cleaned_input = super().clean_input(info, instance, data)
+        # clean and prepare permissions
+        if "permissions" in cleaned_input:
+            permissions = cleaned_input.pop("permissions")
```
NyanKiyoshi: We don't need to `pop` it as we will replace the value.

saleor/graphql/account/mutations/service_account.py
```diff
@@ -0,0 +1,115 @@
...
+
+    class Meta:
+        description = "Deletes a service account"
+        model = models.ServiceAccount
+        permissions = ("account.manage_service_account",)
+
+
+class ServiceAccountUpdatePrivateMeta(UpdateMetaBaseMutation):
+    class Meta:
+        description = "Updates private metadata for service account."
```
NyanKiyoshi:
```suggestion
        description = "Updates private metadata for a service account."
```

saleor/graphql/account/schema.py
```diff
@@ -87,10 +101,24 @@ class AccountQueries(graphene.ObjectType):
...
+        filter=ServiceAccountFilterInput(),
+        description="List of the service accounts",
+    )
+    service_account = graphene.Field(
+        ServiceAccount,
+        id=graphene.Argument(graphene.ID, required=True),
+        description="Lookup a service account by ID.",
+    )
+    service_account_valid_token = graphene.Field(
+        graphene.Boolean, description="Confirm token for service account"
```
NyanKiyoshi:
```suggestion
        graphene.Boolean, description="Confirm token for a service account"
```

saleor/graphql/account/types.py
```diff
@@ -153,6 +154,37 @@ def resolve_order_line(root: models.CustomerEvent, info):
         return None
 
 
+class ServiceAccount(MetadataObjectType, CountableDjangoObjectType):
+    permissions = graphene.List(
+        PermissionDisplay, description="List of service's permissions."
```
NyanKiyoshi:
```suggestion
        PermissionDisplay, description="List of the service's permissions."
```

saleor/graphql/account/types.py
```diff
@@ -153,6 +154,37 @@ def resolve_order_line(root: models.CustomerEvent, info):
...
+    @staticmethod
+    def resolve_permissions(root: models.ServiceAccount, _info, **_kwargs):
+        permissions = root.permissions.prefetch_related("content_type").order_by(
+            "codename"
+        )
+        return format_permissions_for_display(permissions)
+
+    @staticmethod
+    def resolve_auth_token(root: models.ServiceAccount, _info, **_kwargs):
+        return "*" * 6 + root.auth_token[-4:]
```
NyanKiyoshi: Do we want to expose the length of it?

saleor/graphql/account/types.py
```diff
@@ -153,6 +154,37 @@ def resolve_order_line(root: models.CustomerEvent, info):
...
+
+    @staticmethod
+    def resolve_meta(root, _info):
+        return resolve_meta(root, _info)
```
NyanKiyoshi:
```suggestion
        return resolve_meta(root, info)
```

saleor/graphql/account/types.py
```diff
@@ -153,6 +154,37 @@ def resolve_order_line(root: models.CustomerEvent, info):
...
+            "codename"
+        )
+        return format_permissions_for_display(permissions)
+
+    @staticmethod
+    def resolve_auth_token(root: models.ServiceAccount, _info, **_kwargs):
+        return "*" * 6 + root.auth_token[-4:]
+
+    @staticmethod
+    def resolve_meta(root, _info):
```
NyanKiyoshi:
```suggestion
    def resolve_meta(root, info):
```

saleor/graphql/decorators.py
```diff
@@ -0,0 +1,35 @@
+from functools import wraps
+
+import six
+from graphql_jwt import exceptions
+from graphql_jwt.decorators import context
+
+
+def user_passes_test(test_func):
```
NyanKiyoshi: I think it could use a short docstring to explain what it does, the name doesn't tell us much 👍

saleor/graphql/middleware.py
```diff
@@ -1,8 +1,11 @@
 from django.contrib.auth.models import AnonymousUser
 from django.shortcuts import reverse
+from django.utils.functional import SimpleLazyObject
 from graphene_django.settings import graphene_settings
 from graphql_jwt.middleware import JSONWebTokenMiddleware
 
+from saleor.account.models import ServiceAccount
```
NyanKiyoshi:
```suggestion
from ..account.models import ServiceAccount
```

saleor/graphql/middleware.py
```diff
@@ -27,3 +30,27 @@ def middleware(request):
         return get_response(request)
 
     return middleware
+
+
+def get_service_account(auth_token):
```
NyanKiyoshi: I think it would be useful to hint `Optional[ServiceAccount]` so we directly know it returns `None` if it's not found 👍

saleor/graphql/middleware.py
```diff
@@ -27,3 +30,27 @@ def middleware(request):
...
+
+    service_account_auth_header = "HTTP_AUTHORIZATION"
+    prefix = "Bearer"
+
+    def middleware(request):
+        if request.path == reverse("api"):
+            auth = request.META.get(service_account_auth_header, "").split()
+            if len(auth) == 2:
+                auth_prefix, auth_token = auth
+                if auth_prefix == prefix:
```
NyanKiyoshi: I wonder if the prefix should really be case sensitive?

saleor/account/models.py
korycins: To keep consistency between User and ServiceAccount I took permission logic from django.
`django.contrib.auth.backends.ModelBackend#_get_permissions` Here you can find original implementation. I've just changed some part of the flow.

saleor/graphql/account/mutations/service_account.py
korycins: I don't get it. We don't have such convention to use `['input']` when the field is required. 
Almost all of our mutations use `.get('input")` and I did the same to fit in our convention.

saleor/graphql/account/types.py
korycins: I don't think so. This is the reason why I show only 10 characters. 

patrys: I don't think we should be adding the `*` server-side. UI may prefer to use other characters like (`●` is a popular choice).

saleor/graphql/account/mutations/service_account.py
maarcingebala: I think both ways are fine; `data["input"]` may be more Pythonic if we now that `input` is always present, and this is the assumption for `ModelMutation`. If it's a problem, it's not here but in `ModelMutation` where this code was copied from. This mutation shouldn't override `perform_mutation` but `success_response` do assign the `auth_token` to the response.

tests/api/test_account_bot.py
```diff
@@ -0,0 +1,335 @@
+import graphene
```
maarcingebala: Let's rename this file to `test_service_account.py` to match the naming used in the models/API.

patrys:
@korycins Could you include a short high-level overview of how it works? I think that would make it much easier to review the changes and to document them properly.

korycins:
@patrys  Yes, sure

korycins:
@patrys Done

PR #4700, opened 2019-08-29
fix: attributes resolver inCategory and inCollection

### LINKED ISSUE #4698: Attributes filter "InCategory" does not work (graphql)
I'm trying to fetch attributes that are relevant to a category or collection as filters on the storefront.

1. Query `attributes` using the `InCategory` filter.

### What I expected to happen
I expected only categories used in the specified category to be returned. Instead, it returned all attributes in the database. I haven't checked the `InCollection` filter, but I guess it is likely to have the same issue.

### Sample
```gql
{
  attributes(inCategory: "UHJvZHVjdFR5cGU6MTE=", first: 10){
    edges{
      node{
        id
        name
        values{
          name
          slug
        }
      }
    }
  }
}
```

**System information**
Operating system: Windows 10 x64
Browser: Firefox 68

khalibloo:
It fixes the `InCategory` and `InCollection` filters in the `attributes` resolver. Fixes #4698 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/resolvers.py
+++ b/saleor/graphql/product/resolvers.py
@@ -1,8 +1,7 @@
 from typing import TYPE_CHECKING, Optional
 
-import graphene
 import graphene_django_optimizer as gql_optimizer
-from django.db.models import Q, Sum
+from django.db.models import Sum
 from graphql import GraphQLError
 from graphql_relay import from_global_id
 
@@ -12,6 +11,7 @@ from ...search.backends import picker
 from ..utils import filter_by_period, filter_by_query_param, get_database_id, get_nodes
 from .enums import AttributeSortField, OrderDirection
 from .filters import (
+    filter_attributes_by_product_types,
     filter_products_by_attributes,
     filter_products_by_categories,
     filter_products_by_collections,
@@ -31,18 +31,11 @@ COLLECTION_SEARCH_FIELDS = ("name", "slug")
 ATTRIBUTES_SEARCH_FIELDS = ("name", "slug")
 
 
-def _filter_attributes_by_product_types(attribute_qs, product_qs):
-    product_types = set(product_qs.values_list("product_type_id", flat=True))
-    return attribute_qs.filter(
-        Q(product_types__in=product_types) | Q(product_variant_types__in=product_types)
-    )
-
-
 def resolve_attributes(
     info,
     qs=None,
-    category_id=None,
-    collection_id=None,
+    in_category=None,
+    in_collection=None,
     query=None,
     sort_by=None,
     **_kwargs,
@@ -50,26 +43,11 @@ def resolve_attributes(
     qs = qs or models.Attribute.objects.get_visible_to_user(info.context.user)
     qs = filter_by_query_param(qs, query, ATTRIBUTES_SEARCH_FIELDS)
 
-    if category_id:
-        # Filter attributes by product types belonging to the given category.
-        category = graphene.Node.get_node_from_global_id(info, category_id, "Category")
-        if category:
-            tree = category.get_descendants(include_self=True)
-            product_qs = models.Product.objects.filter(category__in=tree)
-            qs = _filter_attributes_by_product_types(qs, product_qs)
-        else:
-            qs = qs.none()
+    if in_category:
+        qs = filter_attributes_by_product_types(qs, "in_category", in_category)
 
-    if collection_id:
-        # Filter attributes by product types belonging to the given collection.
-        collection = graphene.Node.get_node_from_global_id(
-            info, collection_id, "Collection"
-        )
-        if collection:
-            product_qs = collection.products.all()
-            qs = _filter_attributes_by_product_types(qs, product_qs)
-        else:
-            qs = qs.none()
+    if in_collection:
+        qs = filter_attributes_by_product_types(qs, "in_collection", in_collection)
 
     if sort_by:
         is_asc = sort_by["direction"] == OrderDirection.ASC.value
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - PaymentGatewayEnum removed from GraphQL schema as gateways now are dynamic plugins. Gateway names changed. - #4756 by @salwator
 - Add support for webhooks - #4731 by @korycins
 - Fixed the random failure of `populatedb` trying to create a new user with an existing email - #4769 by @NyanKiyoshi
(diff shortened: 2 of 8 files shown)


saleor/graphql/product/resolvers.py
```diff
@@ -16,6 +16,7 @@
     filter_products_by_stock_availability,
     sort_qs,
 )
+from .types import Category, Collection
```
NyanKiyoshi: Please keep them lazy. Just as it was before, i.e. `"Category"`. This prevents infinite loops when we want to invoke a product resolver from another type's resolver.

saleor/graphql/product/types/products.py
```diff
@@ -43,7 +43,6 @@
 from ...utils import get_database_id, reporting_period_to_date
 from ..enums import OrderDirection, ProductOrderField
 from ..filters import AttributeFilterInput
-from ..resolvers import resolve_attributes
```
khalibloo: Missed this part. Looks like the import is being used. Adding it back results in circular imports. And using lazy imports in the resolver file gives this error.
```
ERROR saleor.graphql.views Exception information: [PID:7088:Thread-1]
  File "C:\Users\khali\.virtualenvs\ksaleor-oqf0lJ8X\lib\site-packages\promise\promise.py", line 487, in _resolve_from_executor
    executor(resolve, reject)
  File "C:\Users\khali\.virtualenvs\ksaleor-oqf0lJ8X\lib\site-packages\promise\promise.py", line 754, in executor
    return resolve(f(*args, **kwargs))
  File "C:\Users\khali\.virtualenvs\ksaleor-oqf0lJ8X\lib\site-packages\graphql\execution\middleware.py", line 75, in make_it_promise
    return next(*args, **kwargs)
  File "D:\Workspace\I-Novanet\Manjoo\Backend\ksaleor\saleor\graphql\core\fields.py", line 144, in connection_resolver
    iterable = resolver(root, info, **args)
  File "D:\Workspace\I-Novanet\Manjoo\Backend\ksaleor\saleor\graphql\product\schema.py", line 247, in resolve_attributes
    return resolve_attributes(info, **kwargs)
  File "D:\Workspace\I-Novanet\Manjoo\Backend\ksaleor\saleor\graphql\product\resolvers.py", line 47, in resolve_attributes
    category = graphene.Node.get_node_from_global_id(info, in_category, "Category")
  File "C:\Users\khali\.virtualenvs\ksaleor-oqf0lJ8X\lib\site-packages\graphene\relay\node.py", line 102, in get_node_from_global_id
    only_type._meta.name
AttributeError: 'str' object has no attribute '_meta'
```
Any clue how to work around this?

NyanKiyoshi: You could use `from_global_id_strict_type` which returns a `pk` (as str, don't need to cast it). You can provides it with the keyword argument `only_type` which supports lazy types. Then you can manually get the node through `category = models.Category(pk=pk)` but it will raise `ObjectDoesNotExist`, which thus needs to be wrapped into a try-catch.

But that would make the code a bit dirty. You could add a django-like shortcut: `get_or_error(field, model or queryset, **kwargs)` function into `saleor.graphql.core.utils`. Note: `field` would be used for `raise ValidationError({field: "message here"})`.

I think it would be better as it give us more control over how to resolve the node into a model instance (prefetch, etc.).

So you can do this:
```python
category = get_node_or_error("in_category", models.Category, pk=pk)
```

Which in other places would be useful for things like:
```python
qs = MyRelatedModel.objects.only("pk").prefetch("pizzas")
obj = get_node_or_error("model_id", qs, user_id=user_id, checkout_id=checkout_id)
```

NyanKiyoshi: Actually, we do have one in the core mutation `get_node_or_error`. You can invoke it through `BaseMutation.get_node_or_error`. For now it should be fine, but we should extend it a bit and extract it to a simple function instead of `classmethod` but that's for another PR.

khalibloo: `BaseMutation.get_node_or_error` makes a call to `get_node_from_global_id`, passing `only_type` as a str. Using it produces the same error as before.

I tried adding a `get_node_or_error` in `graphql.core.utils`. Could you take a look at https://github.com/mirumee/saleor/blob/62b3f6ab29003ea1fa018c2b2ad1a81ae48ed4da/saleor/graphql/core/utils/__init__.py#L37
At this point, I have `only_type` as a str and `graphene_type` as a `<class 'graphene.utils.subclass_with_meta.SubclassWithMeta_Meta'>`, so the check fails and the validation error is raised. Is it fine to cast graphene_type to str here?

Not very familiar with the graphql code in saleor still.

django-queries:
Here is the report for b5160df32886e3291163a877a714c33cc0a60a60 (khalibloo/saleor @ fix-attrs-resolver)
Base comparison is 62b3f6ab29003ea1fa018c2b2ad1a81ae48ed4da.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         48	         48	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         13	         13	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	         15	         15	              8
```

</p>
</details>

saleor/graphql/core/utils/__init__.py
```diff
@@ -34,7 +46,7 @@ def from_global_id_strict_type(info, global_id, only_type, field="id"):
     """Resolve a node global id with a strict given type required."""
     _type, _id = graphene.Node.from_global_id(global_id)
     graphene_type = info.schema.get_type(_type).graphene_type
-    if graphene_type != only_type:
+    if str(graphene_type) != only_type:
```
NyanKiyoshi:
```suggestion
    if str(graphene_type) != str(only_type):
```

maarcingebala:
Thanks for your contribution @khalibloo and @NyanKiyoshi!

PR #4721, opened 2019-09-05
Validate shipping address on reused checkout

bogdal:
This PR fixes the case when we try to save invalid shipping address using the `checkoutCreate` mutation for the user who already has existing checkout with a valid address.

### Pull Request Checklist


bogdal:
The current solution is correct. To avoid problems of mixed checkout data for the logged user, first of all, we should get the checkout object and then make operations on it.

PR #4731, opened 2019-09-10
Support for webhooks

korycins:
Resolves https://github.com/mirumee/saleor/issues/2084

It adds support for Webhooks. 
It adds mutation for subscribing, deleting and listing Webhook subscription.  Service Account can assign to all events but permissions to given resources are required to trigger webhook event. Webhook logic is written as a Saleor's plugin.


django-queries:
Here is the report for 98de23d30268f92e30af26afc9f3eb8a249d95c6 (korycins/saleor @ feature/webhooks)
Base comparison is 2a6f85bacbb420ef53a4e05b3f88ca948d910436.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         48	         48	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         15	         15	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	         18	         18	              8
```

</p>
</details>

saleor/graphql/webhook/mutations.py
```diff
@@ -0,0 +1,125 @@
...
+from django.core.exceptions import ValidationError
+
+from ...webhook import models
+from ..core.mutations import ModelDeleteMutation, ModelMutation
+from .types import Webhook
+
+
+class WebhookSubscribeInput(graphene.InputObjectType):
+    target_url = graphene.String(description="The url to receive the payload")
+    event = graphene.String(description="The event that webhook wants to subscribe")
```
NyanKiyoshi: Most of the time we want to subscribe to multiple events

saleor/graphql/webhook/mutations.py
```diff
@@ -0,0 +1,125 @@
+import graphene
+from django.core.exceptions import ValidationError
+
+from ...webhook import models
+from ..core.mutations import ModelDeleteMutation, ModelMutation
+from .types import Webhook
+
+
+class WebhookSubscribeInput(graphene.InputObjectType):
+    target_url = graphene.String(description="The url to receive the payload")
```
NyanKiyoshi: One would want to `secretKey` field (255 chars), blank by default to allow Saleor to send a HMAC signature header of the payload when sending the webhook event, c.f. the related issue + GH repositories are good examples.

This allows to verify the payload is actually coming from Saleor and not from somebody else by calculating the HMAC of the payload against the shared secret key.

Good reference https://developer.github.com/webhooks/securing/
Example by me https://github.com/NyanKiyoshi/pytest-django-queries-bot/blob/master/github/security/github.go

korycins: @NyanKiyoshi I know about this and about multiple events. And this is a reason why this PR is in the draft :) I just wanted to have a draft of API structure.

saleor/extensions/plugins/webhook/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+from ....webhook import WebhookEventType
+from ....webhook.models import Webhook
+from . import create_webhook_headers
+
+logger = logging.getLogger(__name__)
+
+WEBHOOK_TIMEOUT = 10
+
+
+@shared_task
```
maarcingebala: I think we should use `saleor.celeryconf.app.task` here. Look for other Celery tasks, where it is used. BTW we should change Avatax task similarly.

saleor/graphql/order/bulk_mutations/orders.py
```diff
@@ -40,5 +42,7 @@ def bulk_action(cls, queryset, user, restock):
                 events.fulfillment_restocked_items_event(
                     order=order, user=user, fulfillment=order
                 )
-
+            manager = get_extensions_manager()
+            manager.order_cancelled(order)
+            manager.order_updated(order)
```
maarcingebala: Why running two webhooks here? Isn't "cancelled" enough? I'm wondering whether we shouldn't have one webhook like "order/updated" for all orders' state changes to make things simpler. I think Github does something like that e.g. for issues.

korycins: Shopify + Magento have separate webhooks for that.  If you want to receive only cancel events you will subscribe to `order_cancelled`. 
Another service can be subscribed to receive all updates - `order_update`.
Those two services are assigned to separate events and both of them will recieve payload for different events.

saleor/extensions/plugins/webhook/tasks.py
korycins: I am not sure of that. I wanted to limit the dependencies of Saleor in Plugin code. This could be helpful in case of extracting Plugin to different repositories.

korycins:
@salwator @NyanKiyoshi @maarcingebala ready for review

tests/api/test_fulfillment.py
```diff
@@ -35,10 +35,7 @@
 """
 
 
-@patch(
-    "saleor.graphql.order.mutations.fulfillments."
-    "send_fulfillment_confirmation_to_customer"
-)
+@patch("saleor.order.actions.send_fulfillment_confirmation_to_customer")
```
salwator: We should use _spec_ (autospec) when creating mocks.

tests/test_order_actions.py
```diff
@@ -0,0 +1,140 @@
...
+    clean_mark_order_as_paid,
+    handle_fully_paid_order,
+    mark_order_as_paid,
+)
+from saleor.payment import ChargeStatus, PaymentError
+
+
+@patch("saleor.order.emails.send_fulfillment_confirmation.delay")
+@patch("saleor.order.emails.send_payment_confirmation.delay")
+def test_handle_fully_paid_order_digital_lines(
```
salwator: Test body is really complex, could we have test setup in the fixture(s)?

saleor/order/actions.py
```diff
@@ -0,0 +1,209 @@
...
+def order_created(order: "Order", user: "User", from_draft: bool = False):
+    events.order_created_event(order=order, user=user, from_draft=from_draft)
+    manager = get_extensions_manager()
+    manager.order_created(order)
+
+
+def handle_fully_paid_order(order: "Order"):
+    events.order_fully_paid_event(order=order)
+
+    if order.get_customer_email():
```
salwator: Maybe we could extract function out of this? _notify_...

saleor/extensions/plugins/webhook/tasks.py
korycins: Done

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -2,10 +2,14 @@
...
-from ....order.emails import send_fulfillment_confirmation_to_customer
+from ....order import models
+from ....order.actions import (
+    cancel_fulfillment,
+    fulfillment_tracking_updated,
+    order_fulfilled,
+)
 from ....order.error_codes import OrderErrorCode
-from ....order.utils import cancel_fulfillment, fulfill_order_line, update_order_status
+from ....order.utils import fulfill_order_line
```
maarcingebala: Isn't `fulfill_order_line` an *action* as well?

saleor/webhook/models.py
```diff
@@ -0,0 +1,31 @@
...
+from django.utils.translation import pgettext_lazy
+
+from ..account.models import ServiceAccount
+
+
+class Webhook(models.Model):
+    service_account = models.ForeignKey(
+        ServiceAccount, related_name="webhooks", on_delete=models.CASCADE
+    )
+    target_url = models.URLField("Target URL", max_length=255)
```
maarcingebala: Is the "Target URL" verbose name used on purpose? We rather don't use it in other model fields.

saleor/graphql/order/mutations/fulfillments.py
maarcingebala: I'm not sure what are the criteria for an *action* now, in other words, how do we decide whether a function should go to `actions.py`. From the function name `fulfill_order_line`, it sounds as an action.

PR #4734, opened 2019-09-12
Unified MenuItemMove to other reordering mutations

### LINKED ISSUE #4353: Unify the sorting logic
We decided to adopt a new logic, sorting by relative order in #4123. With #4257 coming, #4123's logic will be extracted to abstract objects.

In #3958 we have implemented an absolute ordering.

In a few other places we also have absolute (or implicit?) ordering. We should unify the sorting logic all over the API. But it might break dashboard 1.0–it might be better if we do this after dropping it.

## Tasks
- [ ] Unify menu ordering -> `MenuItemMove`
- [x] ~~Unify product image ordering -> `ProductImageReorder`~~ not needed.
- [ ] Make sure the fields are no longer available through the API
- [ ] Make sure the resolvers always return the correct order
- [ ] Update dashboard 2.0

NyanKiyoshi:
Closes #4353.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/menu/types.py
+++ b/saleor/graphql/menu/types.py
@@ -54,6 +54,14 @@ class MenuItem(CountableDjangoObjectType):
         resolver=resolve_translation,
     )
 
+    sort_order = graphene.Field(
+        graphene.Int,
+        deprecation_reason=(
+            "Use the position in list instead and relative "
+            "calculus to determine shift position."
+        ),
+    )
+
     class Meta:
         description = """Represents a single item of the related menu.
         Can store categories, collection or pages."""
--- a/saleor/graphql/menu/types.py
+++ b/saleor/graphql/menu/types.py
@@ -57,7 +57,8 @@ class MenuItem(CountableDjangoObjectType):
     sort_order = graphene.Field(
         graphene.Int,
         deprecation_reason=(
-            "Use the position in list instead and relative "
+            "Will be dropped in 2.10 and is deprecated since 2.9: "
+            "use the position in list instead and relative "
             "calculus to determine shift position."
         ),
     )
--- a/saleor/graphql/views.py
+++ b/saleor/graphql/views.py
@@ -202,8 +202,14 @@ class GraphQLView(View):
 
         if settings.DEBUG:
             lines = []
-            for line in traceback.format_exception(type(exc), exc, exc.__traceback__):
-                lines.extend(line.rstrip().splitlines())
+
+            # FIXME: test :80
+            if isinstance(exc, BaseException):
+                for line in traceback.format_exception(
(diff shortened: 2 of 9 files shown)


django-queries:
Here is the report for fe3764c44a6ae83b0f96f40d125e7583ae0d5aa3 (NyanKiyoshi/saleor @ unify/sorting/menu-items)
Base comparison is c42b24fe991e99cbe27406652d3300a912dc0cad.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         34	         34	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         48	         48	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         15	         15	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	         18	         18	              8
```

</p>
</details>

saleor/graphql/menu/types.py
```diff
@@ -54,6 +54,14 @@ class MenuItem(CountableDjangoObjectType):
         resolver=resolve_translation,
     )
 
+    sort_order = graphene.Field(
+        graphene.Int,
+        deprecation_reason=(
+            "Use the position in list instead and relative "
```
maarcingebala: We need to also say in which version it will be removed. I suggest 2.10 since 2.9 is about to be released.

saleor/graphql/views.py
```diff
@@ -202,8 +202,14 @@ def format_error(error):
 
         if settings.DEBUG:
             lines = []
-            for line in traceback.format_exception(type(exc), exc, exc.__traceback__):
-                lines.extend(line.rstrip().splitlines())
+
+            # FIXME: test :80
```
maarcingebala: What do you mean? :)

NyanKiyoshi: It means: "@NyanKiyoshi please add test runs the line 80". It was raising an internal error (`DEBUG = True` only) when input was invalid because there was no trace. I forgot to add the test... That's why there was a fixme because... I knew I would forget :wink:

PR #4753, opened 2019-09-23
Provide capped stock quantity instead of exact one and restrict the costPrice field productVariant

### LINKED ISSUE #4724: What is the difference between a product's "purchaseCost" and its variants' "cost price"?

I am working with saleor release 2.8.0.

I am trying to build up my products list in a consistent way. In particular, I need a clear relationship between what a product cost me and how much it will cost my buyers.

In the GraphQL API, I can find the field `purchaseCost` in the `product` query. That field is protected by the `manage_products` permission. Is that `purchaseCost` the amount I paid as a vendor to get that product in my shop?  To be more specific, let's say I have an e-commerce platform grouping the products I get from some people. I want to gain money on those products, so that I sell them on my platform a bit more expensive than what those people sold them to me. Is the `purchaseCost` the price those people sold me their products? 

I can also find a field `costPrice` in the `variants` field of the `product` query. That field is not protected with the `manage_products` permission. I expect this field to be related to the product's `purchaseCost`. Therefore I expect that field to be protected by the same permission.

But I might've misunderstood the purpose of those fields. What is the difference between a product's `purchaseCost` and a its variant's `costPrice`?

My assumption until now is that `purchaseCost` has the same relationship to the variants' `costPrice` as the product `priceRange` has to the variants' `pricing`. If that holds true, however, I don't understand why the variants `costPrice` isn't protected with the `product.manage_products` permission.

NyanKiyoshi:
Closes #4724.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/types/products.py
+++ b/saleor/graphql/product/types/products.py
@@ -310,6 +310,7 @@ class ProductVariant(CountableDjangoObjectType, MetadataObjectType):
         return getattr(root, "digital_content", None)
 
     @staticmethod
+    @permission_required("product.manage_products")
     def resolve_stock_quantity(root: models.ProductVariant, *_args):
         return root.quantity_available
 
@@ -324,6 +325,11 @@ class ProductVariant(CountableDjangoObjectType, MetadataObjectType):
     @permission_required("product.manage_products")
     def resolve_margin(root: models.ProductVariant, *_args):
         return get_margin_for_variant(root)
+
+    @staticmethod
+    @permission_required("product.manage_products")
+    def resolve_cost_price(root: models.ProductVariant, *_args):
+        return root.cost_price
 
     @staticmethod
     def resolve_price(root: models.ProductVariant, *_args):
--- a/saleor/graphql/product/types/products.py
+++ b/saleor/graphql/product/types/products.py
@@ -315,10 +315,8 @@ class ProductVariant(CountableDjangoObjectType, MetadataObjectType):
         return getattr(root, "digital_content", None)
 
     @staticmethod
-    def resolve_stock_quantity(root: models.ProductVariant, info):
+    def resolve_stock_quantity(root: models.ProductVariant, _info):
         exact_quantity_available = root.quantity_available
-        if info.context.user.has_perm("product.manage_products"):
-            return exact_quantity_available
         return min(exact_quantity_available, settings.MAX_CHECKOUT_LINE_QUANTITY)
 
     @staticmethod
--- a/saleor/graphql/product/types/products.py
+++ b/saleor/graphql/product/types/products.py
@@ -2,6 +2,7 @@ from typing import List, Union
 
 import graphene
 import graphene_django_optimizer as gql_optimizer
+from django.conf import settings
 from django.db.models import Prefetch
 from graphene import relay
 from graphql.error import GraphQLError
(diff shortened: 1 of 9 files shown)


django-queries:
Here is the report for 4a0653007236ef002561cdf5e497c36aac538d58 (NyanKiyoshi/saleor @ fix/variants/stock-permissions)
Base comparison is 9398d4f949d6f650deedcaaf784ba94cf87db8a7.



<p>

```diff
# api.benchmark checkout
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  add billing address to checkout     	         38	         38	             20
  add shipping to checkout            	          7	          7	              0
  checkout payment charge             	         14	         14	              0
  complete checkout                   	          6	          6	              0
  create checkout                     	         52	         52	             24

# api.benchmark homepage
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve main menu                  	          5	          5	              0
  retrieve product list               	          4	          4	              0
  retrieve secondary menu             	          5	          5	              0
  retrieve shop                       	          2	          2	              0

# api.benchmark product
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  product details                     	         15	         15	              3

# api.benchmark variant
  test name                           	left count 	right count	duplicate count
  ------------------------------------	-----------	-----------	---------------
  retrieve variant list               	         18	         18	              8
```

</p>
</details>

maarcingebala:
I think there should be some way to determine available quantity in the public API. We spoke with @patrys and came up with an idea to:
- have the `stockQuantity` restricted for admins and if they have rights to return the total amount of items in stock
- extend the `pricing` field for variants with `quantity` field that would be public and would return number of items, but not greater than `settings.MAX_CHECKOUT_LINE_QUANTITY` value

Also, I was checking if we could use the `isAvailable` field for variants and found out that underlying `ProductVariant.is_available` function is broken. Instead of checking that a particular variant is in stock, it uses `self.product.is_available`, which will always return `True` if at least one variant is in stock.

saleor/graphql/product/types/products.py
```diff
@@ -307,8 +312,11 @@ def resolve_digital_content(root: models.ProductVariant, *_args):
         return getattr(root, "digital_content", None)
 
     @staticmethod
-    def resolve_stock_quantity(root: models.ProductVariant, *_args):
-        return root.quantity_available
+    def resolve_stock_quantity(root: models.ProductVariant, info):
+        exact_quantity_available = root.quantity_available
+        if info.context.user.has_perm("product.manage_products"):
```
maarcingebala: We want this field to work the same for everyone. Admins have the `quantity` fields that they can rely on. `stockQuantity` however is meant for public API and it should behave in the same way for admins and for customers. So let's always return `min(exact_quantity_available, settings.MAX_CHECKOUT_LINE_QUANTITY)`.

NyanKiyoshi: `stockQuantity` returns the stock available right now and `quantity` returns the stock available + allocated quantity. So there would be no way of knowing apart from `quantity - quantityAllocated`. We might want to introduce another field? `realStockQuantity` maybe as suggested previously?

maarcingebala: But we do have `quantityAllocated` which gives you this information don't we?

PR #4780, opened 2019-09-28
Fixed display of tax rate (forms & details) + update of product tax rate

NyanKiyoshi:
Closes https://github.com/mirumee/saleor/issues/4635

- No longer display the tax rate field if there is no tax support
- User is now able to change a product's tax rate
- User is now able to see the tax rate of in the product details page

### Pull Request Checklist

### DIFF
--- a/saleor/dashboard/product/forms.py
+++ b/saleor/dashboard/product/forms.py
@@ -253,7 +253,15 @@ class AttributesMixin:
 
 class ProductForm(MoneyModelForm, AttributesMixin):
     tax_rate = forms.ChoiceField(
-        required=False, label=pgettext_lazy("Product tax rate type", "Tax rate")
+        required=False,
+        label=pgettext_lazy("Product tax rate type", "Tax rate"),
+        help_text=pgettext_lazy(
+            "Help text for the tax rate field over the product update/create form",
+            (
+                "Make sure you have enabled a VAT provider and fetched the rates if "
+                "needed by the plugin."
+            ),
+        ),
     )
     category = TreeNodeChoiceField(
         queryset=Category.objects.all(), label=pgettext_lazy("Category", "Category")
@@ -301,9 +309,9 @@ class ProductForm(MoneyModelForm, AttributesMixin):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.manager = manager = get_extensions_manager()
+        self.manager = get_extensions_manager()
         product_type = self.instance.product_type
-        self.initial["tax_rate"] = get_product_tax_rate(self.instance, manager)
+        self.initial["tax_rate"] = get_product_tax_rate(self.instance, self.manager)
         self.available_attributes = product_type.product_attributes.prefetch_related(
             "values"
         ).product_attributes_sorted()
@@ -320,9 +328,15 @@ class ProductForm(MoneyModelForm, AttributesMixin):
         self.fields["seo_title"] = SeoTitleField(
             extra_attrs={"data-bind": self["name"].auto_id}
         )
-        self.fields["tax_rate"].choices = [
-            (tax.code, tax.description) for tax in manager.get_tax_rate_type_choices()
+        tax_rate_field = self.fields["tax_rate"]
+        tax_rate_field.choices = [
+            (tax.code, tax.description)
+            for tax in self.manager.get_tax_rate_type_choices()
         ]
+
+        if not tax_rate_field.choices:
+            tax_rate_field.disabled = True
+
         if include_taxes_in_prices():
             self.fields["price"].label = pgettext_lazy(
                 "Currency gross amount", "Gross price"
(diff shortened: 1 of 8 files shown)


django-queries:
Here is the report for b2946f0c5a1ac60f4a37499474d29408e6e93646 (NyanKiyoshi/saleor @ fix/dashboard-1.0/products/tax-update-and-details)
Base comparison is e44fcd583a61a5d094f16eccbf46932edc501181.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         38	         38	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          6	          6	              0
  create checkout                            	         52	         52	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	         13	         13	              2

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve variant list                      	         18	         18	              8

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/dashboard/product/forms.py
```diff
@@ -300,12 +301,9 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        manager = get_extensions_manager()
+        self.manager = manager = get_extensions_manager()
```
maarcingebala: I think we can make it simpler:
```suggestion
        self.manager = get_extensions_manager()
```

saleor/dashboard/product/forms.py
```diff
@@ -300,12 +301,9 @@ class Meta:
...
         product_type = self.instance.product_type
-        product_tax_rate = manager.get_tax_code_from_object_meta(self.instance).code
-        self.initial["tax_rate"] = (
-            product_tax_rate or manager.get_tax_code_from_object_meta(product_type).code
-        )
+        self.initial["tax_rate"] = get_product_tax_rate(self.instance, manager)
```
maarcingebala:
```suggestion
        self.initial["tax_rate"] = get_product_tax_rate(self.instance, self.manager)
```

saleor/dashboard/product/utils.py
```diff
@@ -0,0 +1,11 @@
+from ...extensions.manager import ExtensionsManager, get_extensions_manager
+from ...product.models import Product
+
+
+def get_product_tax_rate(product: Product, manager: ExtensionsManager = None) -> str:
```
maarcingebala: Shouldn't we use the same logic in the API as well? If this was the case, I think we could have this logic in the manager's `get_tax_code_from_object_meta` method. @korycins what do you think?

NyanKiyoshi: In the API we only retrieve the value once (resolver). I am not sure if there would be any other use case. That's why I put it into the dashboard utilities for now, so it can be removed safely with the dashboard.

maarcingebala: In the dashboard, we're getting the tax rate from product and if it's unknown,  then we're getting it from the product type. Don't we have to use the same logic over there as well? If not then fine, but I thought it should be reused.

templates/dashboard/product/form.html
```diff
@@ -87,7 +87,9 @@
                   {{ product_form.price|materializecss }}
                 </div>
                 <div class="row">
-                  {{ product_form.tax_rate|materializecss }}
+                  {% if product_form.tax_rate.field.choices %}
```
korycins: Do we want to hide this field?  This will change the structure of the form and can be misleading for users. We should show some popup with information that tax support is disabled and freeze the input.

tests/dashboard/test_product.py
```diff
@@ -255,6 +261,56 @@ def test_view_product_edit(admin_client, product):
...
+        mocked_manager = mocked_manager()
+        mocked_manager.assign_tax_code_to_object_meta.side_effect = (
+            manager.assign_tax_code_to_object_meta
+        )
+        mocked_manager.get_tax_rate_type_choices.return_value = [
+            TaxType(code=tax_rate_code, description="")
+        ]
+
+        # Ensure the available tax rate codes are properly returned
+        response = admin_client.get(url)
```
korycins: You should have two separate tests here. You are testing two different things. 1) retrieving tax data from the form. 2) updating tax data for product

saleor/dashboard/product/forms.py
NyanKiyoshi: When line length is prefered over simplicity 👀

PR #4791, opened 2019-10-02
Add Prettier

### LINKED ISSUE #3404: Adopt prettier for older JS files
We currently have two contradicting styles in the same repository. Typescript uses `prettier` but the old storefront code follows a different set of rules. Let's make `prettier` the only accepted style and update our JS files to match it.

Julian-B90:
I want to merge this change because i add prettier to formate the js files.

Fixes #3404


django-queries:
Here is the report for f7606d9536df1f6bf2965eaa5ab2dbd3dbd2c80b (Julian-B90/saleor @ master)
Base comparison is 5f2dde844f806219e385175c58366da1c157090b.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          6	          6	              0
  create checkout                            	         48	         48	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	         13	         13	              2

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve variant list                      	         18	         18	              8

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

maarcingebala:
We should also fix formatting in other JS files in `saleor/static/js` (e.g. in `components` or `stores`) subdirectories.

.prettierrc.json
```diff
@@ -0,0 +1,17 @@
...
+  "printWidth": 80,
+  "proseWrap": "preserve",
+  "quoteProps": "as-needed",
+  "requirePragma": false,
+  "semi": true,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "none",
+  "useTabs": false
+}
```
maarcingebala: I would try to use the default Prettier's config as much as possible here. I think we don't have any specific rules, e.g. commonly using single quotes in those files is rather by accident than by our strict internal rules or preferences.

Julian-B90: Default was this:
```json
{
  "arrowParens": "avoid",
  "bracketSpacing": true,
  "htmlWhitespaceSensitivity": "css",
  "insertPragma": false,
  "jsxBracketSameLine": false,
  "jsxSingleQuote": false,
  "printWidth": 80,
  "proseWrap": "preserve",
  "quoteProps": "as-needed",
  "requirePragma": false,
  "semi": true,
  "singleQuote": false,
  "tabWidth": 2,
  "trailingComma": "none",
  "useTabs": false
}
```

Julian-B90: The only thing that is different is the `singleQuote`. Do we want to adjust that?

maarcingebala: I'd go with the default double-quotes. We use it it Python as well and it works well because of the fewer number of escapes (see https://prettier.io/docs/en/rationale.html#strings). If others are the default, do we need to include them in the config file?

Julian-B90: When we you the defaults we don´t need a config file.

maarcingebala:
This looks really good now! I'm waiting for the CI to pass and it will be ready to merge.

This issue is marked with `hacktoberfest` label and [we have a special gift for people who take part in it - ultra-fast Saleor socks](https://twitter.com/getsaleor/status/1179370205895634946) 😄. Email me at marcin@mirumee.com with your address and we'll send you a package! 🎉

PR #4817, opened 2019-10-07
Feature/add sample payload to webhook event

korycins:
Add sample webhook payload to WebhookEvent query.


django-queries:
Here is the report for 113fadcddf5a3700cff48d63ef9931d024123e18 (korycins/saleor @ feature/add_sample_payload_to_webhook_event)
Base comparison is 2fdc874ae131516e1bb0186e0cf2a2bfc48d178e.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          6	          6	              0
  create checkout                            	         48	         48	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	         13	         13	              2

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve variant list                      	         18	         18	              8

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/webhook/payloads.py
```diff
@@ -151,3 +155,37 @@ def generate_product_payload(product: "Product"):
...
+        orders = order_qs.filter(status=OrderStatus.UNFULFILLED)[:2]
+    elif event_name == WebhookEventType.ORDER_FULLYPAID:
+        orders = order_qs.filter(payments__status=ChargeStatus.FULLY_CHARGED)[:2]
+    elif event_name == WebhookEventType.ORDER_FULFILLED:
+        orders = order_qs.filter(fulfillments__status=FulfillmentStatus.FULFILLED)[:2]
+    elif event_name in [
+        WebhookEventType.ORDER_CANCELLED,
+        WebhookEventType.ORDER_UPDATED,
+    ]:
+        orders = order_qs.filter(status=OrderStatus.CANCELED)[:2]
```
NyanKiyoshi: We could add a test to ensure that an unknown event is returning `[]` so we are actually expecting a given result and thus is actually a feature to return `[]` when we don't have any data

saleor/webhook/payloads.py
```diff
@@ -151,3 +155,37 @@ def generate_product_payload(product: "Product"):
         },
     )
     return product_payload
+
+
+def generate_sample_payload(event_name):
+    if event_name == WebhookEventType.CUSTOMER_CREATED:
+        users = User.objects.filter(is_staff=False, is_active=True)[:2]
```
maarcingebala: Is there a reason to always return `2` objects from the queryset in this function? Since this is a list you could always return one object and pack it into a list, unless there is some requirement from Zapier.

korycins: It can be a sinlge object in list. I will change it

saleor/webhook/payloads.py
```diff
@@ -151,3 +150,46 @@ def generate_product_payload(product: "Product"):
...
+    elif event_name in [
+        WebhookEventType.ORDER_CANCELLED,
+        WebhookEventType.ORDER_UPDATED,
+    ]:
+        order = order_qs.filter(status=OrderStatus.CANCELED).first()
+
+    return generate_order_payload(order) if order else []
+
+
+def generate_sample_payload(event_name):
```
NyanKiyoshi: Maybe we could hint this is always returning a list?

saleor/graphql/webhook/schema.py
```diff
@@ -15,6 +16,17 @@ class WebhookQueries(graphene.ObjectType):
         description="Lookup a webhook by ID.",
     )
     webhooks = PrefetchingConnectionField(Webhook, description="List of webhooks")
+    webhook_sample_payload = graphene.Field(
+        graphene.JSONString,
+        event_type=graphene.Argument(
+            WebhookEventTypeEnum,
+            required=True,
+            description="Name of the reguested event type.",
```
NyanKiyoshi: Please don't try to guess the event names.
```suggestion
            description="Name of the requested event type.",
```

saleor/webhook/payloads.py
```diff
@@ -155,6 +158,16 @@ def generate_product_payload(product: "Product"):
     return product_payload
 
 
+def _get_sample_object(qs: QuerySet) -> Optional[Model]:
+    """Return random object from query."""
+    count = qs.count()
```
NyanKiyoshi: Why not the first one every time? Doing a count then getting the `x`th one has a heavy cost

tests/extensions/plugins/test_webhook.py
```diff
@@ -40,6 +46,66 @@ def test_trigger_webhooks_for_event(
...
+    event_name,
+    total_webhook_calls,
+    expected_target_urls,
+    service_account,
+    order_with_lines,
+    permission_manage_orders,
+    permission_manage_users,
+    permission_manage_products,
+):
+    service_account.permissions.add(permission_manage_orders)
```
NyanKiyoshi: It tests the permissions, right? Could we precise it? In the docstring or somewhere so we know what we are testing.

saleor/webhook/payloads.py
korycins: What do you mean by saying heavy cost? 
I need to have different results for the sample.

NyanKiyoshi: What about `order_by("?").first()` then?

korycins: I found some old tests which point that order_by is heavier than count but I've checked that and it seems that is not relevant anymore. I will change this. Thanks

tests/extensions/plugins/test_webhook.py
korycins: As the name says - `trigger_webhooks_for_event` calls expected events. I don't know what else should I add to docstring

saleor/graphql/webhook/schema.py
```diff
@@ -21,6 +22,19 @@ class WebhookQueries(graphene.ObjectType):
         filter=WebhookFilterInput(description="Filtering options for webhooks."),
     )
 
+    webhook_sample_payload = graphene.Field(
```
maarcingebala: Missing description for this field. I would explain that sample payloads are used by some integrations and it returns real data.

korycins: 🚓 👮

PR #4820, opened 2019-10-07
Add backend logic for dynamic default email sender settings

### LINKED ISSUE #4815: Dynamic configuration for DEFAULT_FROM_EMAIL setting
We would like to be able to dynamically set the value of `DEFAULT_FROM_EMAIL` setting. A good place to store it is the `SiteSettings` model. We would need to extend the API to allow managing and retrieving it.

There is also `ORDER_FROM_EMAIL` setting which acts identically, except it's only used for sending order-related emails. I'm wondering whether it is actually needed or we could simplify it to have only one configuration. I'd like to hear some feedback on that.

NyanKiyoshi:
Closes #4815 

### Definition of Done
- [x] Mutations and types are implemented
- [x] Dashboard 1.0 is implemented
- [x] We ensure one cannot inject headers
- [x] The expected header value is tested
- [x] It is tested one cannot inject a header even if the validation is bypassed (`send_mail`'s own double check)
- [x] Changes are implemented into the email tasks

### Pull Request Checklist


django-queries:
Here is the report for c9916b70045d774b96eae0b2e2b7644d628c2c90 (NyanKiyoshi/saleor @ dynamic-settings/email-sender)
Base comparison is 0bde7f7685f1d820a64f54b0b9abd3a5c478595d.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          6	          6	              0
  create checkout                            	         48	         48	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	         13	         13	              2

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve variant list                      	         18	         18	              8

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/core/emails.py
```diff
@@ -4,7 +4,13 @@
 from ..core.utils import build_absolute_uri
 
 
-def get_email_base_context():
-    site = Site.objects.get_current()
+def get_email_bases():
```
maarcingebala: Let's add a docstring here.

maarcingebala: I think `get_email_context` would be a quite good name for this function, although it returns a tuple. But maybe with a good docstring, it would be good enough.

saleor/graphql/shop/mutations.py
```diff
@@ -34,7 +34,13 @@ class ShopSettingsInput(graphene.InputObjectType):
         description="Default number of max downloads per digital content url"
     )
     default_digital_url_valid_days = graphene.Int(
-        description=("Default number of days which digital content url will be valid")
+        description="Default number of days which digital content url will be valid"
```
maarcingebala: Let's add a period at the end of each description that we're touching here 😉 👮‍♂️

saleor/site/models.py
```diff
@@ -57,6 +82,27 @@ class Meta:
...
+    @property
+    def default_from_email(self) -> Optional[str]:
+        sender_name: Optional[str] = self.default_mail_sender_name or ""
+        sender_address: Optional[str] = self.default_mail_sender_address
+
+        if not sender_address:
+            sender_address = settings.DEFAULT_FROM_EMAIL
+
+            if not sender_address:
+                return None
```
maarcingebala: Since this configuration is becoming dynamic, I think we should throw a warning when this property returns `None`. I'm even wondering if it shouldn't be thrown right away if `settings.DEFAULT_FROM_EMAIL` is used here. Site settings should be configured as early as possible.

saleor/graphql/shop/types.py
```diff
@@ -270,6 +276,19 @@ def resolve_default_country(_, _info):
...
+    @permission_required("site.manage_settings")
+    def resolve_default_mail_sender_name(_, info):
+        return info.context.site.settings.default_mail_sender_name
+
+    @staticmethod
+    @permission_required("site.manage_settings")
+    def resolve_default_mail_sender_address(_, info):
+        return (
+            info.context.site.settings.default_mail_sender_address
+            or settings.DEFAULT_FROM_EMAIL
```
maarcingebala: Is this required? `SiteSettings.default_from_email` property seems to already handle this `or`.

saleor/site/models.py
NyanKiyoshi: We are currently using `console:///` by default. Thus users would get an error when trying saleor. Maybe we could add the default `demo@example.com` in settings when `console:///` is used?

PR #4835, opened 2019-10-10
Fix fetching staff user without manage_users permission

### LINKED ISSUE #4831: Can't fetch staff user without manage_users permission
To fetch staff user using `user(id: "")` query.

1. 
```
staffUser: user(id: "VXNlcjoyMQ==") {
  id
}
```
2. Get `PermissionDenied` error

fowczarek:
I want to merge this change because resolve #4831

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/resolvers.py
+++ b/saleor/graphql/account/resolvers.py
@@ -3,7 +3,6 @@ from typing import Optional
 
 import graphene
 import graphene_django_optimizer as gql_optimizer
-from django.db.models import Q
 from graphql_jwt.exceptions import PermissionDenied
 from i18naddress import get_validation_rules
 
@@ -26,9 +25,7 @@ USER_SEARCH_FIELDS = (
 
 
 def resolve_customers(info, query):
-    qs = models.User.objects.filter(
-        Q(is_staff=False) | (Q(is_staff=True) & Q(orders__isnull=False))
-    )
+    qs = models.User.objects.customers()
     qs = filter_by_query_param(
         queryset=qs, query=query, search_fields=USER_SEARCH_FIELDS
     )
@@ -38,7 +35,7 @@ def resolve_customers(info, query):
 
 
 def resolve_staff_users(info, query):
-    qs = models.User.objects.filter(is_staff=True)
+    qs = models.User.objects.staff()
     qs = filter_by_query_param(
         queryset=qs, query=query, search_fields=USER_SEARCH_FIELDS
     )
@@ -48,15 +45,15 @@ def resolve_staff_users(info, query):
 
 
 def resolve_user(info, id):
-    api_account = info.context.user or info.context.service_account
-    if api_account:
+    requester = info.context.user or info.context.service_account
+    if requester:
         _model, user_pk = graphene.Node.from_global_id(id)
-        if api_account.has_perms(["account.manage_staff", "account.manage_users"]):
+        if requester.has_perms(["account.manage_staff", "account.manage_users"]):
             return models.User.objects.filter(pk=user_pk).first()
-        if api_account.has_perm("account.manage_staff"):
-            return models.User.objects.filter(pk=user_pk, is_staff=True).first()
-        if api_account.has_perm("account.manage_users"):
-            return models.User.objects.filter(pk=user_pk, is_staff=False).first()
+        if requester.has_perm("account.manage_staff"):
+            return models.User.objects.staff().filter(pk=user_pk).first()
+        if requester.has_perm("account.manage_users"):
+            return models.User.objects.customers().filter(pk=user_pk).first()
     return PermissionDenied()
 
 
--- a/saleor/graphql/account/resolvers.py
+++ b/saleor/graphql/account/resolvers.py
@@ -1,8 +1,10 @@
 from itertools import chain
 from typing import Optional
 
+import graphene
 import graphene_django_optimizer as gql_optimizer
 from django.db.models import Q
+from graphql_jwt.exceptions import PermissionDenied
 from i18naddress import get_validation_rules
 
 from ...account import models
@@ -43,6 +45,19 @@ def resolve_staff_users(info, query):
     qs = qs.order_by("email")
     qs = qs.distinct()
     return gql_optimizer.query(qs, info)
+
+
+def resolve_user(info, id):
+    api_account = info.context.user or info.context.service_account
+    if api_account:
+        _model, user_pk = graphene.Node.from_global_id(id)
+        if api_account.has_perms(["account.manage_staff", "account.manage_users"]):
+            return models.User.objects.filter(pk=user_pk).first()
+        if api_account.has_perm("account.manage_staff"):
+            return models.User.objects.filter(pk=user_pk, is_staff=True).first()
+        if api_account.has_perm("account.manage_users"):
+            return models.User.objects.filter(pk=user_pk, is_staff=False).first()
+    return PermissionDenied()
 
 
 def resolve_service_accounts(info):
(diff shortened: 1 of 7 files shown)


django-queries:
Here is the report for e9ae24e5cab14d8142e5ca0c13eb97cd73e7b75b (mirumee/saleor @ 4831/fetch_staff_user_without_manage_users_permission)
Base comparison is b92e705d631746c89756cb269b392f96dbd6ffe2.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          6	          6	              0
  create checkout                            	         48	         48	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	         13	         13	              2

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve variant list                      	         15	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/graphql/account/schema.py
```diff
@@ -164,9 +165,9 @@ def resolve_me(self, info):
     def resolve_staff_users(self, info, query=None, **_kwargs):
         return resolve_staff_users(info, query=query)
 
-    @permission_required("account.manage_users")
+    @one_of_permissions_required(["account.manage_staff", "account.manage_users"])
```
NyanKiyoshi: I don't think we need to check twice the permissions? Through this wrap and then we check again in `resolve_user`

saleor/graphql/account/resolvers.py
```diff
@@ -45,6 +47,19 @@ def resolve_staff_users(info, query):
     return gql_optimizer.query(qs, info)
 
 
+def resolve_user(info, id):
+    api_account = info.context.user or info.context.service_account
+    if api_account:
```
NyanKiyoshi: `api_account` seems confusing? Why `api` and not simply `current_user`, `requester` or something of this kind?

saleor/graphql/decorators.py
```diff
@@ -23,15 +23,30 @@ def wrapper(context, *args, **kwargs):
...
+        for perm in perms:
+            has_perm = _permission_required(perm, context)
+            if has_perm:
+                return True
         return False
 
     return account_passes_test(check_perms)
+
+
+def _permission_required(perm, context):
```
NyanKiyoshi: I would put it at the top as it would make the code easier to read as we use it in all the above functions, we expect to read from top to bottom, and not to have to scroll down to find the reference (most of the time).

saleor/graphql/decorators.py
```diff
@@ -23,15 +23,30 @@ def wrapper(context, *args, **kwargs):
 
 def permission_required(perm):
```
NyanKiyoshi: As we expect as list or a single item that will then becoming a list, I think it would be more friendly to directly hint it supports multiple values.
```suggestion
def permission_required(perms: Union[str, Iterable[str]]):
```

saleor/graphql/decorators.py
```diff
@@ -23,15 +23,30 @@ def wrapper(context, *args, **kwargs):
...
+            has_perm = _permission_required(perm, context)
+            if has_perm:
+                return True
         return False
 
     return account_passes_test(check_perms)
+
+
+def _permission_required(perm, context):
+    if isinstance(perm, six.string_types):
```
NyanKiyoshi: I think we should only do this in `permission_required` and `one_of_permissions_required` should expect a list (with proper type hinting)

saleor/graphql/decorators.py
```diff
@@ -23,15 +23,30 @@ def wrapper(context, *args, **kwargs):
...
+        perms = (perm,)
+    else:
+        perms = perm
+    if context.user.has_perms(perms):
+        return True
+    service_account = getattr(context, "service_account", None)
+    if service_account and service_account.has_perms(perms):
+        return True
+    return False
```
NyanKiyoshi:
```python
    user = context.user or getattr(context, "service_account", None)

    if user and has_perm: ...
```

tests/api/test_account.py
```diff
@@ -297,6 +297,52 @@ def test_user_query_anonymous_user(api_client):
     assert_no_permission(response)
 
 
+def test_user_query_permission_manage_users_get_customer(
+    staff_api_client, customer_user, permission_manage_users
+):
+    staff_api_client.user.user_permissions.add(permission_manage_users)
+    customer_id = graphene.Node.to_global_id("User", customer_user.pk)
+    variables = {"id": customer_id}
+    response = staff_api_client.post_graphql(USER_QUERY, variables)
```
NyanKiyoshi: I think it would be best if we passed the permission directly to `post_graphql` so we also check it raises permission denied if the permission are missing for each case (get customer with X perm, get staff with X perm).

Same for below.

saleor/graphql/account/schema.py
fowczarek: We need this because if an anonymous user tries to get this query that can cause different error than permission denied.

saleor/graphql/decorators.py
fowczarek: We use to check it in this order because `context.user` have `AnonymousUser` for a service account. To make your solution to work, we need to add a condition for checking `AnonymousUser`. I think the current solution is better.

saleor/graphql/account/resolvers.py
```diff
@@ -45,6 +47,19 @@ def resolve_staff_users(info, query):
...
 
 
+def resolve_user(info, id):
+    requester = info.context.user or info.context.service_account
+    if requester:
+        _model, user_pk = graphene.Node.from_global_id(id)
+        if requester.has_perms(["account.manage_staff", "account.manage_users"]):
+            return models.User.objects.filter(pk=user_pk).first()
+        if requester.has_perm("account.manage_staff"):
+            return models.User.objects.filter(pk=user_pk, is_staff=True).first()
```
maarcingebala: We have a `staff` queryset which you could use here. Ideally, we should try to reuse those querysets across the codebase to have one logic for getting staff users.

maarcingebala: It seems that `resolve_staff_users ` function above could also use that.

saleor/graphql/account/resolvers.py
```diff
@@ -45,6 +47,19 @@ def resolve_staff_users(info, query):
...
+        if requester.has_perm("account.manage_users"):
+            return models.User.objects.filter(pk=user_pk, is_staff=False).first()
```
maarcingebala: For returning customers we have the `customers` queryset. Bonus: it also returns staff users who are customers (they have any orders - this is allowed in Saleor). We need to make sure that if a staff users is at the same time a customer, and we're returning their data here (because I as requester have `manage_users` permission), no restricted staff-related fields are leaking.

maarcingebala: Again, `resolve_customers ` above should use the same queryset as this resolver.

saleor/graphql/decorators.py
NyanKiyoshi: True, true, true. Forgot about `user` not being `None` in Django 👀

PR #4871, opened 2019-10-21
Remove mutations from validate_address

### LINKED ISSUE #4493: Unclear CheckoutShippingAddressUpdate mutation needs refactoring

I was tried to add some additional logic when checkout's shipping address is changed.
Unfortunately, mutation `CheckoutShippingAddressUpdate` is unclear and brings me to the buggy solution.

It's the mutation: https://github.com/mirumee/saleor/blob/7a21736a3363aa8f1bacac7218c4a78e44e09652/saleor/graphql/checkout/mutations.py#L342-L376
When I read it, what steps I see:
1. get the checkout
2. validate new shipping address
3. clean shipping method
4. new shipping address is saved and assigned to the checkout.
The following steps are described in #4321 too. But, in fact, it does something different in some cases.

Method `validate_address` not only validates address, but it modifies given instance. If checkout had no shipping address - everything works great. But if checkout has shipping address and we would like to change it, it would be changed inside `validate_address` method. Function `change_shipping_address_in_checkout` will do nothing in that case. 

I think that function should be refactored, it could be done by one of the following:
- change `validate_address` function name and docstring
- move saving valid address instance outside `validate_address` to make it clear

I think some additional tests should be written for the case that checkout has a shipping address and user wants change it - all tests I've found covers the case that `checkout.shipping_address` is `None` before sending mutation.

…

KevBelisle:
I want to merge this change because it fixes #4493 .

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/mutations.py
+++ b/saleor/graphql/checkout/mutations.py
@@ -6,6 +6,7 @@ from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.db import transaction
 from django.utils import timezone
 
+from ...account.models import Address
 from ...checkout import models
 from ...checkout.error_codes import CheckoutErrorCode
 from ...checkout.utils import (
@@ -34,7 +35,6 @@ from ...payment.interface import AddressData
 from ...payment.utils import store_customer_id
 from ...product import models as product_models
 from ..account.i18n import I18nMixin
-from ..account.models import Address
 from ..account.types import AddressInput, User
 from ..core.mutations import (
     BaseMutation,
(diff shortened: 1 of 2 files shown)


KevBelisle:
If someone could add a DRAFT label, that'd be great :)

saleor/graphql/checkout/mutations.py
```diff
@@ -480,17 +480,17 @@ def perform_mutation(cls, _root, info, checkout_id, shipping_address):
 
         cleaned_address_data = cls.validate_address(shipping_address)
 
-        checkout.shipping_address = cls.construct_instance(
-            checkout.shipping_address or Address(),
+        shipping_address = cls.construct_instance(
+            shipping_address,
```
Kwaidan00: `construct_instance` gets `instance` of type `Address` here as a first parameter, but you provides `AddressInput`.

PR #4866, opened 2019-10-18
Add fields to keep metadata for Fulfillment model

szewczykmira:
Closes #4865 

Also, I'm not sure if I've used proper permissions - any comments are welcome.

### Pull Request Checklist

1. [ ] Privileged views and APIs are guarded by proper permission checks.
1. [ ] All visible strings are translated with proper context.
1. [ ] All data-formatting is locale-aware (dates, numbers, and so on).
1. [x] Database queries are optimized and the number of queries is constant.
1. [x] Database migration files are up to date.
1. [x] The changes are tested.
1. [x] GraphQL schema and type definitions are up to date.
1. [x] Changes are mentioned in the changelog.

PR #4876, opened 2019-10-22
Fix permissions on updating/clearing Order metadata

szewczykmira:
During work on #4866 I thought that I've might have introduced some bugs to our codebase.
This PR will add tests on handling permissions on updating Order metadata and some small fixes to the logic.

### Pull Request Checklist


django-queries:
Here is the report for 06338ce4e1e68bb2c11cb74916df1ef2d4d894d6 (mirumee/saleor @ fix/order-metadata-permissions)
Base comparison is 861b282d291890d79d68da8a1c41c6d6688a6f2a.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          6	          6	              0
  create checkout                            	         48	         48	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	         13	         13	              2

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         15	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/graphql/order/mutations/orders.py
```diff
@@ -424,29 +424,29 @@ class OrderUpdateMeta(UpdateMetaBaseMutation):
     class Meta:
         description = "Updates meta for order."
         model = models.Order
-        permissions = "orders.manage_orders"
```
korycins: Maybe it is a good moment to create some permissions validator in `BaseMutation`? We just need some assert between expected permission and our list of available permission.

tests/api/test_order.py
```diff
@@ -2246,3 +2246,204 @@ def test_draft_orders_query_with_filter_search_by_id(
...
+    order_id = graphene.Node.to_global_id("Order", order.id)
+    return {
+        "id": order_id,
+        "input": {"namespace": "test", "clientName": "client1", "key": "foo"},
+    }
+
+
+@pytest.fixture
+def order_with_meta(order):
+    order.store_meta(namespace="test", client="client1", item={"foo": "bar"})
```
korycins: I recommend to use different data between `meta` and `private_meta`. If someone uses this fixture, it will be hard to confirm that the testing function/method returns expected data from meta/private_meta.

saleor/graphql/order/mutations/orders.py
maarcingebala: Not sure if I understand correctly, how is `BaseMutation` supposed to validate permissions? I think the issue is that we're relying on strings to identify permissions across the entire codebase, while we should have one enum for all permissions and use it instead.

maarcingebala: Yes, it would be a good time to change that now or very soon.

korycins: > Not sure if I understand correctly, how is `BaseMutation` supposed to validate permissions? I think the issue is that we're relying on strings to identify permissions across the entire codebase, while we should have one enum for all permissions and use it instead.

it can be a simple assert - `meta_permission in MODELS_PERMISSIONS` 🤷‍♂ 
I agree that, the better way would be not using `strings`.

saleor/graphql/order/mutations/orders.py
```diff
@@ -424,29 +424,29 @@ class OrderUpdateMeta(UpdateMetaBaseMutation):
...
+        permissions = ("order.manage_orders",)
```
korycins: @salwator  Should we protect order public meta? Personally, I think that PWA client should be able to store some data dedicated to order.

korycins: And the second question. Should we use the order's ID here instead of `order.token`? 
cc @maarcingebala

szewczykmira: I assume this also apply to clearing public meta? @salwator @korycins

szewczykmira:
@korycins could you review again?

saleor/graphql/order/mutations/orders.py
```diff
@@ -420,33 +421,68 @@ def perform_mutation(cls, _root, info, amount, **data):
...
+            required=True,
+        )
+
+    @classmethod
+    def get_instance(cls, info, **data):
+        token = data["token"]
+        return models.Order.objects.get(token=token)
+
+
+class OrderClearMetaBaseMutation(ClearMetaBaseMutation):
```
korycins: We don't want to offer a mutation to clear the whole meta for the user. It should be allowed only for the staff user with expected permission

saleor/graphql/order/mutations/orders.py
```diff
@@ -420,33 +421,68 @@ def perform_mutation(cls, _root, info, amount, **data):
...
+    class Arguments:
+        token = graphene.UUID(description="Token of an object to clear.", required=True)
+        input = MetaPath(
+            description="Fields required to update new or stored metadata item.",
+            required=True,
+        )
+
+    @classmethod
+    def get_instance(cls, info, **data):
+        token = data["token"]
```
korycins: @maarcingebala public meta will use `token` as ID. Should we do the same for private meta? Private is protected by `order.manage_orders` so we could also use a normal ID here.

szewczykmira: Aaaand done ;)

PR #4891, opened 2019-10-24
Plugins config should always keep Boolean as Boolean

### LINKED ISSUE #4879: URGENT: Braintree gateway config not working for live
We found a bug in the plugin braintree configuration that is leading to complete non-functional braintree gateway for production. The error is coming from the sandbox_mode configuration parameter.
It works great for sandbox, as this is the standard config:
https://github.com/mirumee/saleor/blob/master/saleor/payment/gateways/braintree/__init__.py#L100
But it seems like sandbox_mode gets stored in saleor as string!! not as python boolean. We checked it by introducing a print(type)) like this, directly after line 100:
```
    print('Sandbox ? ', sandbox_mode)
    print(type(sandbox_mode))
    if not sandbox_mode:
        environment = braintree_sdk.Environment.Production
```
and it returns: 
```
Sandbox ? false
<class 'str'>
``` 
So here we have the core-issue .. 
We wanted to propose a merge request to fix this issue, but honestly I don't know where to fix it the best. This field should never be allowed to be a string, so it seems to be a core issue. 

## Steps to reproduce the problem
1. Use braintree with sandbox_mode disabled
2. try to create an order

szewczykmira:
I want to merge this change because currently, requests from graphQL are saving boolean values as strings in the plugin configuration and this has caused us already some issues.

I would suggest creating a separate issue for adding tests and checks that will make sure that similar error will not happen again.

Closes #4879

### Pull Request Checklist

### DIFF
--- a/saleor/extensions/base_plugin.py
+++ b/saleor/extensions/base_plugin.py
@@ -322,17 +322,20 @@ class BasePlugin:
     def _update_config_items(
         cls, configuration_to_update: List[dict], current_config: List[dict]
     ):
-    
+        config_structure = (
+            cls.CONFIG_STRUCTURE if cls.CONFIG_STRUCTURE is not None else {}
+        )
         for config_item in current_config:
             for config_item_to_update in configuration_to_update:
-                if config_item["name"] == config_item_to_update.get("name"):
+                config_item_name = config_item_to_update.get("name")
+                if config_item["name"] == config_item_name:
                     new_value = config_item_to_update.get("value")
                     # The frontend can send all configuration fields with
                     # all secrets as **...**, we have to omit all secrets which
                     # weren't modified
                     if new_value == cls.REDACTED_FORM:
                         continue
-                    item_type = config_item.get("type")
+                    item_type = config_structure.get(config_item_name, {}).get("type")
                     if item_type == ConfigurationTypeField.BOOLEAN and not isinstance(
                         new_value, bool
                     ):
--- a/saleor/extensions/base_plugin.py
+++ b/saleor/extensions/base_plugin.py
@@ -6,6 +6,7 @@ from django.db.models import QuerySet
 from django_countries.fields import Country
 from prices import Money, MoneyRange, TaxedMoney, TaxedMoneyRange
 
+from . import ConfigurationTypeField
 from .models import PluginConfiguration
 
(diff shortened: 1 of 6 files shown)


saleor/extensions/base_plugin.py
```diff
@@ -206,6 +207,11 @@ def _update_config_items(
             for config_item_to_update in configuration_to_update:
                 if config_item["name"] == config_item_to_update.get("name"):
                     new_value = config_item_to_update.get("value")
+                    item_type = config_item.get("type")
```
korycins: What do you think about a separate function like - `clean_configuration` which should be run before the update of configuration items? This would be helpful for future plugins.

django-queries:
Here is the report for 145d2160c8b84478c98b9687ab1ebe0299cc27db (mirumee/saleor @ fix/sandbox-production-mode)
Base comparison is d090ddba7ebfca48c8bb1320c920ca7c5a8354dd.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          6	          6	              0
  create checkout                            	         48	         48	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	         13	         13	              2

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         15	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

PR #4939, opened 2019-11-06
PayU integration

Boquete:
Suggestion for https://github.com/mirumee/saleor/issues/4908
This pr is discussion purpose only, it's not working/tested solution yet. Just simple architecture concept.

### Pull Request Checklist


korycins:
Hi @Boquete,

Your PR looks very promising 👏 👏 👏. 
Could check the `saleor/graphql/schema.graphql `? Github treats `schema.graphql` as a fully changed file which complicates the review and hides new changes in the schema. I assume that this can be connected with the different character at the end of line

Boquete:
@korycins i will take a look why github treated it like that, meanwhile, the only change in schema is line 648, redirect_url inside CheckoutComplete type

korycins:
@Boquete 👋 
The architecture for plugins looks good. 
There are two things that we should change:

- url - `r'^%s/payu-notification/$' % (TOKEN_PATTERN,),`
We don't want to have gateway dependencies inside Saleor. The better way would be to create something like - `/integrations/<slug:plugin>/`, BasePlugin should have a method which ExtensionManager will call inside this view. In that case, each plugin can handle the request on his own, thanks to it - we don't have custom logic in Saleor's URLs, and we have a more customizable approach that can be used by any plugin.
- Migration of Payment to Plugin architecture was the first step to extract payment gateways into separate packages. We want to do that for our payment gateways. Thanks to this, the community will have a better possibility to create and share their own payment gateway implementation. So, for now, we don't want to add to Saleor's repository any new payment gateway as we don't have enough resources to maintain it. We are going to create a list (in docs or readme) of all implemented gateways (with URL) which community can use. 
You can prepare  PayU gateway on your own and we will append it to the gateways list. Of course, if you need any support we are ready to help you.

PR #4944, opened 2019-11-06
Hide all secret fields in plugin configuration

### LINKED ISSUE #4937: Plugins query doesn't have all fields
```
class ExtensionsManager(PaymentInterface):
    def get_plugin_configurations(self) -> List["PluginConfiguration"]:
        plugin_configuration_ids = []
        plugin_configurations_qs = PluginConfiguration.objects.all()
        for plugin in self.plugins:
            plugin_configuration = plugin.get_plugin_configuration(
                plugin_configurations_qs
            )
            plugin_configuration_ids.append(plugin_configuration.pk)
        return PluginConfiguration.objects.filter(pk__in=plugin_configuration_ids)
```
`get_plugin_configurations` returns configuration directly from database and skips all static fields that we attached in real time in `plugin.get_plugin_configuration`. 
`filter` input field in query requires QuerySet as a returned value from resolvers (previosuly it was `list`)

korycins:
- Adds new config fields - `Secret`, `Password`
   - `Secret`  dedicated for fields like `Public Key`, `Private Key`
   - `Password` dedicated to fields that should be fully hidden.
- Remove `hide_secret_fields`logic  from `BasePlugin`
- Hide secret logic on graphene type level - closes #4937
For API:
    - `Secret returns last 4 or 1 digits for token. Null if secret is empty
    - `Password` returns empty string if password exists. Null if password is empty


django-queries:
Here is the report for 03f8b4f1360f2311a2ba6c707ab3d9bc65a040be (korycins/saleor @ fix/hide_secret_plugin_configuration_fields)
Base comparison is ed8d664a7d1bf61843aa68d7f79bb6a6b63d750d.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          6	          6	              0
  create checkout                            	         48	         48	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	         13	         13	              2

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         15	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/graphql/extensions/types.py
```diff
@@ -24,3 +24,25 @@ class Meta:
...
+            for field in configuration:
+                name = field["name"]
+                value = field["value"]
+                if value is None:
+                    continue
+                field_type = plugin.CONFIG_STRUCTURE.get(name, {}).get("type")
+                if field_type == ConfigurationTypeField.PASSWORD:
+                    field["value"] = "" if value else None
+
+                if field_type == ConfigurationTypeField.SECRET:
```
szewczykmira: Maybe moving this to some separate function would be a good idea? I'm pretty sure that this will be needed somewhere else :)

PR #4970, opened 2019-11-14
Allow to delete category and leave products

### LINKED ISSUE #1928: Deleting a category deletes all subcategories and their products
Since there's no way to undo deleting one of the main categories is a great way to accidentally get rid of most if not all of your products. The only remedy is restoring the entire store from a recent backup.

IKarbowiak:
Allow to create product without category. After removing category, related products become unpublished and minimal variant price is recalculated.

Closes #1928 

### Pull Request Checklist


django-queries:
Here is the report for 9e21ed5fcd4982cacc531be6668b6871616cf761 (mirumee/saleor @ allow-to-delete-category-and-leave-products)
Base comparison is 92f5ba113426bf18595bb069c9cb2f739f5e7e88.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          6	          6	              0
  create checkout                            	         50	         50	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	         13	         13	              2

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         15	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/dashboard/product/forms.py
```diff
@@ -363,6 +366,13 @@ def clean_seo_description(self):
         )
         return seo_description
 
+    def clean_is_published(self):
+        is_published = self.cleaned_data["is_published"]
+        category = self.data["category"]
+        if not category and is_published:
+            raise ValidationError("You must add cateogry to publish product")
```
maarcingebala: We also need to add this rule in API. Two inputs that use this fields are `ProductCreateInput` and `ProductInput` and mutations that use them should behave similarly.

saleor/dashboard/category/views.py
```diff
@@ -110,7 +112,10 @@ def category_delete(request, pk):
     if request.method == "POST":
         descendants = category.get_descendants()
         menus = get_menus_that_need_update(categories=descendants)
+        products = unpublished_products_before_category_delete(category)
+        product_ids = [product.pk for product in products]
         category.delete()
+        update_products_minimal_variant_prices_task.delay(product_ids)
```
NyanKiyoshi: Could we add a test for this?

saleor/dashboard/category/views.py
```diff
@@ -110,7 +112,10 @@ def category_delete(request, pk):
     if request.method == "POST":
         descendants = category.get_descendants()
         menus = get_menus_that_need_update(categories=descendants)
+        products = unpublished_products_before_category_delete(category)
+        product_ids = [product.pk for product in products]
```
NyanKiyoshi:
```suggestion
        product_ids = list(products.values_list("id", flat=True))
```

Otherwise it is very heavy.

saleor/dashboard/product/forms.py
NyanKiyoshi:
```suggestion
            raise ValidationError("You must select a category to be able to publish")
```

saleor/dashboard/product/forms.py
```diff
@@ -363,6 +366,13 @@ def clean_seo_description(self):
         )
         return seo_description
 
+    def clean_is_published(self):
+        is_published = self.cleaned_data["is_published"]
+        category = self.data["category"]
```
NyanKiyoshi:
```suggestion
        category = self.data.get("category")
```

saleor/dashboard/product/forms.py
```diff
@@ -363,6 +366,13 @@ def clean_seo_description(self):
         )
         return seo_description
 
+    def clean_is_published(self):
+        is_published = self.cleaned_data["is_published"]
```
NyanKiyoshi:
```suggestion
        is_published = self.cleaned_data.get("is_published")
```

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -38,6 +44,15 @@ class Meta:
         error_type_class = ProductError
         error_type_field = "product_errors"
 
+    @classmethod
+    def bulk_action(cls, queryset):
+        products = models.Product.objects.none()
+        for category in queryset:
+            products = products | unpublished_products_before_category_delete(category)
+        product_ids = [product.pk for product in products]
```
NyanKiyoshi:
```suggestion
        product_ids = list(products.values_list("id", flat=True))
```

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -38,6 +44,15 @@ class Meta:
...
+        super().bulk_action(queryset)
+        update_products_minimal_variant_prices_task.delay(product_ids=product_ids)
```
NyanKiyoshi: Could you add a test? 👍

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -38,6 +44,15 @@ class Meta:
         error_type_class = ProductError
         error_type_field = "product_errors"
 
+    @classmethod
+    def bulk_action(cls, queryset):
+        products = models.Product.objects.none()
+        for category in queryset:
+            products = products | unpublished_products_before_category_delete(category)
+        product_ids = [product.pk for product in products]
+        super().bulk_action(queryset)
```
NyanKiyoshi: Should be inside a transaction

saleor/dashboard/category/views.py
```diff
@@ -110,7 +112,10 @@ def category_delete(request, pk):
...
         category.delete()
```
NyanKiyoshi: We need a transaction here

saleor/dashboard/category/views.py
maarcingebala: This four lines are executed every time when we delete a category. The ideal solution would be to put them into one function. I actually think that `unpublished_products_before_category_delete` could become the `delete_category` function that does all of that. It should also have a docstring which would tell you that this function removes a category plus it performs some additional logic related to unpublishing the products.

saleor/graphql/product/bulk_mutations/products.py
NyanKiyoshi: A bit heavy, but we can leave it like that for now. Though it would be nice to add a test in the benchmark package even though with https://github.com/mirumee/saleor/issues/4359 it's still not very useful.

saleor/graphql/product/mutations/products.py
```diff
@@ -143,6 +145,16 @@ class Meta:
         error_type_class = ProductError
         error_type_field = "product_errors"
 
+    @classmethod
+    def perform_mutation(cls, _root, info, **data):
+        node_id = data.get("id")
+        instance = cls.get_node_or_error(info, node_id, only_type=Category)
+        products = unpublished_products_before_category_delete(instance)
+        product_ids = [product.pk for product in products]
+        response = super().perform_mutation(_root, info, **data)
```
NyanKiyoshi: Transaction

saleor/product/models.py
```diff
@@ -315,6 +319,10 @@ def save(
         # Make sure the "minimal_variant_price_amount" is set
         if self.minimal_variant_price_amount is None:
             self.minimal_variant_price_amount = self.price_amount
+
+        # Product can't be published if doesn't have category
+        if not self.category and self.is_published:
```
maarcingebala: If we have the right checks in mutations and views that modify the category model, this logic here won't be necessary. We usually don't put much constraints in the `save` method in models.

saleor/graphql/product/mutations/products.py
```diff
@@ -143,6 +145,16 @@ class Meta:
         error_type_class = ProductError
         error_type_field = "product_errors"
 
+    @classmethod
+    def perform_mutation(cls, _root, info, **data):
+        node_id = data.get("id")
+        instance = cls.get_node_or_error(info, node_id, only_type=Category)
+        products = unpublished_products_before_category_delete(instance)
+        product_ids = [product.pk for product in products]
```
NyanKiyoshi: same, use `values_list("pk", flat=True)`

saleor/product/utils/attributes.py
```diff
@@ -81,3 +81,12 @@ def associate_attribute_values_to_instance(
...
     return assignment
+
+
+def unpublished_products_before_category_delete(category):
+    products = category.products.all()
+    descendants = category.get_descendants()
+    for descendant in descendants:
+        products = products | descendant.products.all()
+    products.update(is_published=False, publication_date=None)
+    return products
```
NyanKiyoshi: That's not a utility for attributes (cf. the filename).

Also, you could add it as a manager method on the category model, would be cleaner 😉

saleor/dashboard/product/forms.py
maarcingebala: @NyanKiyoshi While there is a typo in "cateogry" and it should say "a category", let's not comment on minor language changes which are doesn't really change the meaning or fix any real problem. There is no difference here between wording "add" vs "select" or "publish" or "be able to publish".

saleor/graphql/product/bulk_mutations/products.py
maarcingebala: Not sure if this is part of this PR - it's `BaseBulkMutation`'s reponsibility to ensure that there is a transaction.

saleor/product/utils/attributes.py
maarcingebala: It'd fine to just move it to `saleor/product/__init__.py` which contains some general utility functions related to products.

NyanKiyoshi: Better to avoid `__init__` it is not supposed to contain >10% of the logic, we expect mostly imports in it.

maarcingebala: But it does at the moment. It's better to put the function there now and do the refactoring in a separate PR, since we would need to probably move more functions.

saleor/dashboard/category/views.py
IKarbowiak: New function: `delete_categories` in saleor/product/utils/\_\_init\__.py

PR #4980, opened 2019-11-15
Checkout plugins rework

### LINKED ISSUE #4965: Shipping calculation, totals and plugins
Shipping price calculation plugin

1. I’ve introduced plugin with overwritten calculate_checkout_shipping method
2. Now I can call the manager to get shipping prices, cool!
3. Now checkout total will be invalid: `manager.calculate_checkout_total` takes total from Checkout.get_gotal()
4. Checkout.get_gotal() takes shipping from Checkout.get_shipping_price(), which ignores plugin manager existence

### What I expected to happen
Using a plugin to modify shipping price also modify checkout total

@korycins Am I missing something?

krzysztofwolski:

- Fix #4965
- want to remove methods from models
- want to address way of using extensions (via importing manager vs helper functions)
- type annotations matter

### Pull Request Checklist

### DIFF
--- a/saleor/checkout/base_calculations.py
+++ b/saleor/checkout/base_calculations.py
@@ -1,0 +1,63 @@
+"""Contain functions which are base for calculating checkout properties.
+
+It's recommended to use functions from calculations.py module to take in account plugin
+manager.
+"""
+
+from typing import TYPE_CHECKING
+
+from prices import TaxedMoney
+
+from ..core.taxes import quantize_price, zero_taxed_money
+from ..extensions.manager import get_extensions_manager
+
+if TYPE_CHECKING:
+    from .models import Checkout, CheckoutLine
+    from ..discount.types import DiscountsListType
+
+
+def get_base_checkout_shipping_price(
+    checkout: "Checkout", discounts: "DiscountsListType" = None
+) -> TaxedMoney:
+    """Return checkout shipping price."""
+    # FIXME: should it take discounts in account?
+    if not checkout.shipping_method or not checkout.is_shipping_required():
+        return zero_taxed_money(checkout.currency)
+
+    shipping_price = checkout.shipping_method.get_total()
+    return quantize_price(
+        TaxedMoney(net=shipping_price, gross=shipping_price), shipping_price.currency
+    )
+
+
+def get_base_checkout_subtotal(
+    checkout: "Checkout", discounts: "DiscountsListType" = None
+) -> TaxedMoney:
+    """Return the total cost of all checkout lines."""
+    extensions = get_extensions_manager()
+    subtotals = (
+        extensions.calculate_checkout_line_total(line, discounts) for line in checkout
+    )
+    return sum(subtotals, zero_taxed_money(checkout.currency))
+
+
+def get_base_checkout_total(
+    checkout: "Checkout", discounts: "DiscountsListType" = None
+) -> TaxedMoney:
+    """Return the total cost of the checkout."""
+    extensions = get_extensions_manager()
+    total = (
+        extensions.calculate_checkout_subtotal(checkout, discounts)
+        + extensions.calculate_checkout_shipping(checkout, discounts)
+        - checkout.discount
+    )
+    return max(total, zero_taxed_money(checkout.currency))
+
+
+def get_base_checkout_line_total(
+    line: "CheckoutLine", discounts: "DiscountsListType" = None
+) -> TaxedMoney:
+    """Return the total price of this line."""
+    amount = line.quantity * line.variant.get_price(discounts)
+    price = quantize_price(amount, amount.currency)
+    return TaxedMoney(net=price, gross=price)
--- a/saleor/checkout/base_calculations.py
+++ b/saleor/checkout/base_calculations.py
@@ -11,6 +11,7 @@ from prices import TaxedMoney
 from ..core.taxes import quantize_price, zero_taxed_money
 
 if TYPE_CHECKING:
+    # flake8: noqa
     from .models import Checkout, CheckoutLine
     from ..discount.types import DiscountsListType
 
--- a/saleor/checkout/base_calculations.py
+++ b/saleor/checkout/base_calculations.py
@@ -27,19 +27,20 @@ def base_checkout_shipping_price(checkout: "Checkout") -> TaxedMoney:
     )
 
 
-def base_checkout_subtotal(
-    checkout: "Checkout", line_totals: List[TaxedMoney]
-) -> TaxedMoney:
+def base_checkout_subtotal(line_totals: List[TaxedMoney], currency: str) -> TaxedMoney:
     """Return the total cost of all checkout lines."""
-    return sum(line_totals, zero_taxed_money(checkout.currency))
+    return sum(line_totals, zero_taxed_money(currency))
 
 
 def base_checkout_total(
-    checkout: "Checkout", subtotal: TaxedMoney, shipping_price: TaxedMoney
+    subtotal: TaxedMoney,
+    shipping_price: TaxedMoney,
+    discount: TaxedMoney,
+    currency: str,
 ) -> TaxedMoney:
     """Return the total cost of the checkout."""
-    total = subtotal + shipping_price - checkout.discount
-    return max(total, zero_taxed_money(checkout.currency))
+    total = subtotal + shipping_price - discount
+    return max(total, zero_taxed_money(currency))
 
 
 def base_checkout_line_total(
--- a/saleor/checkout/base_calculations.py
+++ b/saleor/checkout/base_calculations.py
@@ -15,11 +15,8 @@ if TYPE_CHECKING:
     from ..discount.types import DiscountsListType
 
 
-def get_base_checkout_shipping_price(
-    checkout: "Checkout", discounts: "DiscountsListType" = None
-) -> TaxedMoney:
+def base_checkout_shipping_price(checkout: "Checkout") -> TaxedMoney:
     """Return checkout shipping price."""
-    # FIXME: should it take discounts in account?
     if not checkout.shipping_method or not checkout.is_shipping_required():
         return zero_taxed_money(checkout.currency)
 
@@ -29,14 +26,14 @@ def get_base_checkout_shipping_price(
     )
 
 
-def get_base_checkout_subtotal(
+def base_checkout_subtotal(
     checkout: "Checkout", line_totals: List[TaxedMoney]
 ) -> TaxedMoney:
     """Return the total cost of all checkout lines."""
     return sum(line_totals, zero_taxed_money(checkout.currency))
 
 
-def get_base_checkout_total(
+def base_checkout_total(
     checkout: "Checkout", subtotal: TaxedMoney, shipping_price: TaxedMoney
 ) -> TaxedMoney:
     """Return the total cost of the checkout."""
@@ -44,7 +41,7 @@ def get_base_checkout_total(
     return max(total, zero_taxed_money(checkout.currency))
 
 
-def get_base_checkout_line_total(
+def base_checkout_line_total(
     line: "CheckoutLine", discounts: "DiscountsListType" = None
 ) -> TaxedMoney:
     """Return the total price of this line."""
--- a/saleor/checkout/base_calculations.py
+++ b/saleor/checkout/base_calculations.py
@@ -4,12 +4,11 @@ It's recommended to use functions from calculations.py module to take in account
 manager.
 """
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, List
 
 from prices import TaxedMoney
 
 from ..core.taxes import quantize_price, zero_taxed_money
-from ..extensions.manager import get_extensions_manager
 
 if TYPE_CHECKING:
     from .models import Checkout, CheckoutLine
@@ -31,26 +30,17 @@ def get_base_checkout_shipping_price(
 
 
 def get_base_checkout_subtotal(
-    checkout: "Checkout", discounts: "DiscountsListType" = None
+    checkout: "Checkout", line_totals: List[TaxedMoney]
 ) -> TaxedMoney:
     """Return the total cost of all checkout lines."""
-    extensions = get_extensions_manager()
-    subtotals = (
-        extensions.calculate_checkout_line_total(line, discounts) for line in checkout
-    )
-    return sum(subtotals, zero_taxed_money(checkout.currency))
+    return sum(line_totals, zero_taxed_money(checkout.currency))
 
 
 def get_base_checkout_total(
-    checkout: "Checkout", discounts: "DiscountsListType" = None
+    checkout: "Checkout", subtotal: TaxedMoney, shipping_price: TaxedMoney
 ) -> TaxedMoney:
     """Return the total cost of the checkout."""
-    extensions = get_extensions_manager()
-    total = (
-        extensions.calculate_checkout_subtotal(checkout, discounts)
-        + extensions.calculate_checkout_shipping(checkout, discounts)
-        - checkout.discount
-    )
+    total = subtotal + shipping_price - checkout.discount
     return max(total, zero_taxed_money(checkout.currency))
 
 
--- a/saleor/checkout/calculations.py
+++ b/saleor/checkout/calculations.py
@@ -1,0 +1,48 @@
+from typing import TYPE_CHECKING
+
+from ..extensions.manager import get_extensions_manager
+
+if TYPE_CHECKING:
+    from prices import Money
+    from .models import Checkout, CheckoutLine
+    from ..discount.types import DiscountsListType
+
+
+def get_checkout_shipping_price(
+    checkout: "Checkout", discounts: "DiscountsListType" = None
+) -> "Money":
+    """Return checkout price without taking in account aby plugins and extensions.
+
+    It takes in account all extensions.
+    """
+    return get_extensions_manager().calculate_checkout_shipping(checkout, discounts)
+
+
+def get_checkout_subtotal(
+    checkout: "Checkout", discounts: "DiscountsListType" = None
+) -> "Money":
+    """Return the total cost of the checkout prior to shipping.
+
+    It takes in account all extensions.
+    """
+    return get_extensions_manager().calculate_checkout_subtotal(checkout, discounts)
+
+
+def get_checkout_total(
+    checkout: "Checkout", discounts: "DiscountsListType" = None
+) -> "Money":
+    """Return the total cost of the checkout.
+
+    It takes in account all extensions.
+    """
+    return get_extensions_manager().calculate_checkout_total(checkout, discounts)
+
+
+def get_checkout_line_total(
+    line: "CheckoutLine", discounts: "DiscountsListType" = None
+) -> "Money":
+    """Return the total price of this line.
+
+    It takes in account all extensions.
+    """
+    return get_extensions_manager().calculate_checkout_line_total(line, discounts)
--- a/saleor/checkout/calculations.py
+++ b/saleor/checkout/calculations.py
@@ -3,14 +3,14 @@ from typing import TYPE_CHECKING
 from ..extensions.manager import get_extensions_manager
 
 if TYPE_CHECKING:
-    from prices import Money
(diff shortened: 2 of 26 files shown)


saleor/checkout/base_calculations.py
```diff
@@ -0,0 +1,63 @@
...
+if TYPE_CHECKING:
+    from .models import Checkout, CheckoutLine
+    from ..discount.types import DiscountsListType
+
+
+def get_base_checkout_shipping_price(
+    checkout: "Checkout", discounts: "DiscountsListType" = None
+) -> TaxedMoney:
+    """Return checkout shipping price."""
+    # FIXME: should it take discounts in account?
```
krzysztofwolski: @korycins wdyt?

saleor/extensions/plugins/avatax/plugin.py
```diff
@@ -182,30 +189,31 @@ def _calculate_checkout_subtotal(
...
+        discounts: "DiscountsListType",
         previous_value: TaxedMoney,
     ) -> TaxedMoney:
         self._initialize_plugin_configuration()
 
         if self._skip_plugin(previous_value):
             return previous_value
 
-        sub_total = checkout.get_subtotal(discounts)
+        # TODO: Isn't previous value for that?
```
krzysztofwolski: @korycins wdyt?

saleor/checkout/base_calculations.py
```diff
@@ -0,0 +1,63 @@
...
+
+from typing import TYPE_CHECKING
+
+from prices import TaxedMoney
+
+from ..core.taxes import quantize_price, zero_taxed_money
+from ..extensions.manager import get_extensions_manager
+
+if TYPE_CHECKING:
+    from .models import Checkout, CheckoutLine
```
krzysztofwolski: It's complete mystery for me when flake8 thinks those imports are used and when not.

krzysztofwolski:
There are conflicts to be fixed - but I would like to hear what do you think bout those changes

saleor/checkout/base_calculations.py
NyanKiyoshi: Yes... Same trouble with pylint in other projects. I think @korycins was putting `# noqa` for flake if those are errors and not install errors from the linter itself.

Maybe we should open and check for an issue to add support of `exclude_lines` setting that `coverage.py` provides. So we could exclude those type checking blocks.

That would be an easier temporary fix as as far as I remember, the issue has been there for a long time and no fix is in sight.

But maybe that's not possible for flake8 to have such filtering.

saleor/checkout/models.py
```diff
@@ -132,19 +118,19 @@ def get_total_gift_cards_balance(self):
...
         return weights
 
-    def get_line(self, variant):
+    def get_line(self, variant: "ProductVariant") -> Optional["CheckoutLine"]:
         """Return a line matching the given variant and data if any."""
         matching_lines = (line for line in self if line.variant.pk == variant.pk)
         return next(matching_lines, None)
 
-    def get_last_active_payment(self):
+    def get_last_active_payment(self) -> "Payment":
```
NyanKiyoshi: Looks like it is actually `Optional["Payment"]`

saleor/core/taxes.py
```diff
@@ -58,8 +55,8 @@ def get_display_price(
 
 
 def quantize_price(
-    price: Union["TaxedMoney", "Money", "Decimal", "TaxedMoneyRange"], currency
-):
+    price: Union["TaxedMoney", "Money", "Decimal", "TaxedMoneyRange"], currency: str
+) -> Union["TaxedMoney", "Money", "Decimal", "TaxedMoneyRange"]:
```
NyanKiyoshi: We could define a type for this case. What about `T_MoneyLike = Union["TaxedMoney", "Money", "Decimal", "TaxedMoneyRange"]`

saleor/discount/utils.py
```diff
@@ -115,7 +131,7 @@ def get_products_voucher_discount(voucher, prices):
     return total_amount
 
 
-def _fetch_categories(sale_pks):
+def _fetch_categories(sale_pks: Iterable[str]) -> dict:
```
NyanKiyoshi:
```suggestion
def _fetch_categories(sale_pks: Iterable[str]) -> Dict[str, Set[str]]:
```

saleor/checkout/base_calculations.py
korycins: I don't see any usage of discount for shipping, so I assume that we can skip them for base but plugin should still recieve a discounts for `calculate_checkout_shipping`

saleor/checkout/base_calculations.py
```diff
@@ -0,0 +1,63 @@
...
+    if not checkout.shipping_method or not checkout.is_shipping_required():
+        return zero_taxed_money(checkout.currency)
+
+    shipping_price = checkout.shipping_method.get_total()
+    return quantize_price(
+        TaxedMoney(net=shipping_price, gross=shipping_price), shipping_price.currency
+    )
+
+
+def get_base_checkout_subtotal(
```
korycins: I am not sure of these methods. Technically, they just trigger an `extension_manager`. They don't do any `base_` logic. Maybe it should be directly done in the `Manager` methods?
We import manager here and inside manager file we import `base_calulcation` file, which means that we could wrongly place some methods.

saleor/checkout/calculations.py
```diff
@@ -0,0 +1,48 @@
...
+
+if TYPE_CHECKING:
+    from prices import Money
+    from .models import Checkout, CheckoutLine
+    from ..discount.types import DiscountsListType
+
+
+def get_checkout_shipping_price(
+    checkout: "Checkout", discounts: "DiscountsListType" = None
+) -> "Money":
```
korycins: `TaxedMoney`?

saleor/checkout/models.py
```diff
@@ -18,7 +18,10 @@
 from ..giftcard.models import GiftCard
 from ..shipping.models import ShippingMethod
 
-CENTS = Decimal("0.01")
+if TYPE_CHECKING:
+    from ..product.models import ProductVariant
+    from django_measurement import Weight
```
korycins: I am not sure if `isort` handled that. Maybe we could separate 3rd packages from our modules? Or just put it above the `if`

saleor/checkout/utils.py
```diff
@@ -16,6 +16,12 @@
 from ..account.forms import get_address_form
 from ..account.models import Address, User
 from ..account.utils import store_user_address
+from ..checkout.calculations import (
+    get_checkout_line_total,
```
korycins: Mr. Wolski(@krzysztofwolski) - What do you think about removing `get_`?

saleor/checkout/utils.py
```diff
@@ -656,29 +667,36 @@ def change_shipping_address_in_checkout(checkout, address):
...
+    currency: Optional[str] = None,
+    shipping_range=None,
+) -> dict:
     """Retrieve the data shared between views in checkout process."""
-    manager = get_extensions_manager()
-    checkout_total = (
-        manager.calculate_checkout_total(checkout=checkout, discounts=discounts)
-        - checkout.get_total_gift_cards_balance()
-    )
+    cards_balance = checkout.get_total_gift_cards_balance()
```
korycins: Shouldn't be done in `get_checkout_total`?

saleor/checkout/utils.py
```diff
@@ -929,6 +962,7 @@ def get_shipping_price_estimate(checkout: Checkout, discounts, country_code):
     if shipping_methods is None:
         return None
 
+    # TODO: extension manager should be able to have impact on shipping price estimates
```
korycins: Maybe we could create a method for that or gh issue instead of `TODO`?

saleor/discount/utils.py
korycins: `Set[int]`?

saleor/extensions/plugins/avatax/plugin.py
korycins: I think that we should use `previous_value` here. But it doesn't solve our problem. Avatax calculates prices on its own. It doesn't take into account values from the previous plugins. It will be hard to change as Avatax requires a lot of information about single products...

saleor/checkout/base_calculations.py
krzysztofwolski: They are `base` not because they are not using mgr, but because that's most basic algorithm for calculating those values - not taking manager in the account will produce wrong results. Since we introduced plugins there is no such thing as 'subtotal without plugins'.

The idea behind file `calculations.py` is having basic checkout utils inside the checkout module and using those, not having to go through tens of manager methods. 

>

saleor/checkout/utils.py
krzysztofwolski: No. Gift cards are payment method. `checkout_total` is `price of items and handling, minus discounts`, not `how much is left to pay`.
IMO we need to discuss how we calculate 'totals' for checkout and how we handle discounts for them

saleor/checkout/base_calculations.py
maarcingebala: I like the fact that we move calculations from manager. I think manager's reponsibility should be to call the right function instead of containing any business logic. I'm not sure about introducing two files though - `base_calculations` and `calculations`. I'm worried that it may be confusing in the future what's the difference between them and where should I add my new calculation. Couldn't the "base calculations" sit inside `Checkout` model?

saleor/checkout/base_calculations.py
patrys: It does not consider them used if you refer to them as a string.

saleor/checkout/base_calculations.py
krzysztofwolski: Probably if you want new calculation, then you should write plugin :)

saleor/checkout/models.py
NyanKiyoshi: Related issue: https://github.com/timothycrosley/isort/issues/818

saleor/core/taxes.py
krzysztofwolski: I dont really like idea of having both Money and Decimal available here. It could lead to strange bugs.

saleor/discount/utils.py
krzysztofwolski: set is still iterable

saleor/checkout/utils.py
krzysztofwolski: I kind of like the idea, lets roll with it

krzysztofwolski:
I don't understand codeclimate. According to http://pylint.pycqa.org/en/latest/whatsnew/2.0.html TYPE_CHECKING should be handled by pylint :/

saleor/checkout/utils.py
krzysztofwolski: @koradon let me know what do you think about the change

saleor/checkout/models.py
krzysztofwolski: Maybe laziness is speaking thru me, but unless any tool does not manage that, I don't want to go back to pre isort/black time and deal with it manually :/ Type checking sections tend to be way smaller than usual imports so maybe it's not an issue?

django-queries:
Here is the report for ceeacd63f2ff72ac9ad8b9d397ec14e586c81ee6 (krzysztofwolski/saleor @ checkout-plugins-rework)
Base comparison is aec8d5cc933b1614601115091a69b7eb5af9814d.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	         50	         50	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         15	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/checkout/base_calculations.py
```diff
@@ -0,0 +1,51 @@
...
+    if not checkout.shipping_method or not checkout.is_shipping_required():
+        return zero_taxed_money(checkout.currency)
+
+    shipping_price = checkout.shipping_method.get_total()
+    return quantize_price(
+        TaxedMoney(net=shipping_price, gross=shipping_price), shipping_price.currency
+    )
+
+
+def base_checkout_subtotal(
```
korycins: This function can be more generic as It doesn't execute any checkout logic. It just sums up a list of taxed money.
`def sum_prices(prices: List[TaxedMoney], currency)`

saleor/checkout/calculations.py
```diff
@@ -0,0 +1,48 @@
...
+if TYPE_CHECKING:
+    from prices import TaxedMoney
+    from .models import Checkout, CheckoutLine
+    from ..discount.types import DiscountsListType
+
+
+def checkout_shipping_price(
+    checkout: "Checkout", discounts: "DiscountsListType" = None
+) -> "TaxedMoney":
+    """Return checkout price without taking in account aby plugins and extensions.
```
korycins: `aby` -> `any` ?

korycins: I am not sure. In the first sentence, you are writing `without taking in account aby plugins and extensions` but the second sentence says - `It takes in account all extensions.`. Am I missing something?

saleor/extensions/plugins/avatax/plugin.py
```diff
@@ -219,29 +223,29 @@ def _calculate_checkout_shipping(
...
+        discounts: "DiscountsListType",
         previous_value: TaxedMoney,
     ) -> TaxedMoney:
         self._initialize_plugin_configuration()
 
         if self._skip_plugin(previous_value):
             return previous_value
 
-        shipping_price = checkout.get_shipping_price()
+        base_shipping_price = base_calculations.base_checkout_shipping_price(checkout)
```
korycins: Shouldn't we take `previous_value` here? 
In case when the plugin is disabled we should return `previous_value` instead of the default value.

saleor/extensions/plugins/avatax/plugin.py
```diff
@@ -302,9 +306,12 @@ def calculate_checkout_line_total(
             return previous_value
 
         checkout = checkout_line.checkout
-        total = checkout_line.get_total(discounts)
+        base_total = base_calculations.base_checkout_line_total(
```
korycins: `previous_value`?

saleor/extensions/plugins/vatlayer/plugin.py
```diff
@@ -81,30 +81,6 @@ def calculate_checkout_total(
             - checkout.discount
         )
 
-    def calculate_checkout_subtotal(
```
korycins: Is it expected? Here you are calling this method: https://github.com/mirumee/saleor/pull/4980/files#diff-de9a58978d5e65a9854361c28fb83e80R80

saleor/checkout/utils.py
krzysztofwolski: Done #5001

saleor/core/taxes.py
krzysztofwolski: I'll modify it in different PR later on, Products module first. Probably I'll handle this util with it too.

saleor/checkout/calculations.py
krzysztofwolski: Typo

saleor/extensions/plugins/vatlayer/plugin.py
krzysztofwolski: Yes. Taxes are calculated by calculate_checkout_line_total and gathered inside base_calculated_subtotal. No need for overloading this method.

saleor/checkout/utils.py
```diff
@@ -919,7 +946,9 @@ def is_valid_shipping_method(checkout, discounts):
     return True
 
 
-def get_shipping_price_estimate(checkout: Checkout, discounts, country_code):
+def get_shipping_price_estimate(
+    checkout: Checkout, discounts: "DiscountsListType", country_code: str
+):
```
NyanKiyoshi: It returns `Money | TaxedMoney` I think? We could hint it 😄

saleor/checkout/utils.py
```diff
@@ -754,7 +772,9 @@ def get_voucher_discount_for_checkout(voucher, checkout, discounts=None) -> Mone
     raise NotImplementedError("Unknown discount type")
 
 
-def get_voucher_for_checkout(checkout, vouchers=None, with_lock=False):
+def get_voucher_for_checkout(
+    checkout: Checkout, vouchers=None, with_lock: bool = False
+):
```
NyanKiyoshi: `-> Optional[Voucher]`

saleor/discount/utils.py
```diff
@@ -38,15 +47,17 @@ def add_voucher_usage_by_customer(voucher, customer_email):
...
         voucher=voucher, customer_email=customer_email
     )
     if voucher_customer:
         voucher_customer.delete()
 
 
-def get_product_discount_on_sale(product, product_collections, discount: DiscountInfo):
+def get_product_discount_on_sale(
+    product: "Product", product_collections: Set["Collection"], discount: DiscountInfo
+):
```
NyanKiyoshi: `-> Money` AFAIK

saleor/checkout/utils.py
korycins: AFAIK `Optional[TaxedMoneyRange]`

saleor/discount/utils.py
krzysztofwolski: Thats why we need typing ;)

 ```
    product_discount = get_product_discount_on_sale(variant.product, set(), discount)
    discounted_price = product_discount(product.price)
```

Comment (krzysztofwolski) [reply saleor/discount/utils.py]: Honestly I dont know whats the type of this right now. Covering this module is out of scope of this PR anyway.

File: saleor/checkout/utils.py (hunk 10 above)
Comment (koradon) [reply saleor/checkout/utils.py]: A little bit late but I agree with @korycins

File: saleor/extensions/plugins/avatax/__init__.py [hunk 21]
```diff
@@ -107,7 +111,7 @@ def _validate_adddress_details(
 
 def _validate_order(order: "Order") -> bool:
     """Validate the order object if it is ready to generate a request to avatax."""
-    if not order.lines.count():
+    if not len(order.lines.all()):
```
maarcingebala: Microptimization: isn't `exists()` the optimal way to do this?

saleor/checkout/utils.py
```diff
@@ -656,28 +661,36 @@ def change_shipping_address_in_checkout(checkout, address):
...
+) -> dict:
     """Retrieve the data shared between views in checkout process."""
-    manager = get_extensions_manager()
-    checkout_total = (
-        manager.calculate_checkout_total(checkout=checkout, discounts=discounts)
-        - checkout.get_total_gift_cards_balance()
-    )
+    cards_balance = checkout.get_total_gift_cards_balance()
+
+    # FIXME: We are having here parts of checkout total calculation logic
```
maarcingebala: Maybe we should put the substractions below to the `checkout_total` function? It could take gift card or gift card balance as an argument. Or create another calculation metod to substract it. It would be really nice to have this piece in a atomic and tested function.

krzysztofwolski: This method is used for storefront views context. So we will remove it soon.

To handle this logic in one place: https://github.com/mirumee/saleor/issues/5009

PR #4986, opened 2019-11-19
Add implementation for warehouses

### LINKED ISSUE #4403: Provide an interface to configure businesses addresses for each region for taxing purposes
When a business operates in multiple countries it's not uncommon for them to be registered as a taxpayer in more than one country or region.

We should make it possible to manage a list of business addresses and to assign them as taxpayers for a list of regions.

Then Saleor should determine the appropriate address based on the shipping (or billing) address when calculating taxes.

A new configuration sheet in settings for addresses.

Not sure how to assign addresses to regions, would be most useful if it allowed region groups to be created (this could potentially also allow us to choose which tax backend to use for each region group).

N/A

szewczykmira:
A lot of international stores has more than one warehouse and It would be good to keep information about them in the database. Also:
- they are required by Avalara to calculate taxes
- Address of the warehouse might be required during calculating shipping costs.

Closes #4403

### Pull Request Checklist


saleor/dashboard/warehouse/views.py
```diff
@@ -0,0 +1,91 @@
+from typing import TYPE_CHECKING
+
+from django.conf import settings
+from django.contrib import messages
+from django.contrib.auth.decorators import permission_required
+from django.shortcuts import get_object_or_404, redirect
+from django.template.response import TemplateResponse
+from django.utils.translation import pgettext_lazy
+
+from saleor.core.utils import get_paginator_items
```
korycins: AFAIK  you should use relative imports for all Saleor's paths

saleor/graphql/schema.graphql
```diff
@@ -3441,6 +3444,8 @@ type Query {
   webhooks(filter: WebhookFilterInput, before: String, after: String, first: Int, last: Int): WebhookCountableConnection
   webhookEvents: [WebhookEvent]
   webhookSamplePayload(eventType: WebhookEventTypeEnum!): JSONString
+  warehouse(id: ID!): Warehouse
+  warehouses(before: String, after: String, first: Int, last: Int): WarehouseCountableConnection
```
korycins: I assume that `filter` would be useful here

saleor/graphql/warehouse/mutations.py
```diff
@@ -0,0 +1,84 @@
+import graphene
+
+from saleor.account.models import Address
```
korycins: relative Imports

saleor/graphql/warehouse/mutations.py
```diff
@@ -0,0 +1,84 @@
...
+    "city",
+    "city_area",
+    "postal_code",
+    "country",
+    "country_area",
+    "phone",
+]
+
+
+class WarehouseCreate(ModelMutation):
```
korycins: It would be good to extend the Warehouse model by `Meta`.

saleor/graphql/warehouse/mutations.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+
+    class Meta:
+        description = "Creates new warehouse."
+        model = models.Warehouse
+        permissions = ("warehouse.manage_warehouses",)
+
+    @classmethod
+    def create_address(cls, clean_data):
+        address_data = clean_data["address"]
```
korycins: Do we validate `address` somehow? 
I see that some address mutations use address validation from `I18nMixin`.

saleor/graphql/warehouse/mutations.py
```diff
@@ -0,0 +1,84 @@
...
+        address = Address.objects.create(**address_data)
+        return address
+
+    @classmethod
+    def clean_input(cls, info, instance, data, input_cls=None):
+        cleaned_data = super().clean_input(info, instance, data, input_cls=input_cls)
+        cleaned_data["address"] = cls.create_address(cleaned_data)
```
korycins: I am not sure that we should create an instance of an address inside a cleaning method. 
Maybe `construct_instance`,  `save` or `_save_m2m` would be a better place for that?

saleor/graphql/warehouse/mutations.py
```diff
@@ -0,0 +1,84 @@
...
+
+class WarehouseUpdate(ModelMutation):
+    class Meta:
+        model = models.Warehouse
+        permissions = ("warehouse.manage_warehouses",)
+        description = "Update given warehouse."
+
+    class Arguments:
+        id = graphene.ID(description="ID of a warehouse to update.", required=True)
+        input = WarehouseUpdateInput()
```
korycins: Shouldn't be `required=True`?

saleor/graphql/warehouse/mutations.py
```diff
@@ -0,0 +1,84 @@
...
+            field_value = address_data.get(field)
+            if field_value is not None:
+                setattr(address, field, field_value)
+        address.save()
+        return address
+
+    @classmethod
+    def clean_input(cls, info, instance, data, input_cls=None):
+        cleaned_data = super().clean_input(info, instance, data, input_cls=input_cls)
+        cleaned_data["address"] = cls.update_address(instance, cleaned_data)
```
korycins: Same question as above for `clean_method` and saving

saleor/graphql/warehouse/mutations.py
```diff
@@ -0,0 +1,84 @@
...
+        return cleaned_data
+
+
+class WarehouseDelete(ModelDeleteMutation):
+    class Meta:
+        model = models.Warehouse
+        permissions = ("warehouse.manage_warehouses",)
+        description = "Deletes selected warehouse."
+        error_type_class = WarehouseError
```
korycins: AFAIK you should also attach these classes also to above mutations 🤔

saleor/graphql/warehouse/schema.py
```diff
@@ -0,0 +1,47 @@
+import graphene
+import graphene_django_optimizer as gql_optimizer
+from graphql_jwt.decorators import login_required
+
+from saleor.graphql.core.fields import PrefetchingConnectionField
```
korycins: relative imports

saleor/graphql/warehouse/schema.py
```diff
@@ -0,0 +1,47 @@
...
+        description="Look up an order by ID.",
+        id=graphene.Argument(
+            graphene.ID, description="ID of an warehouse", required=True
+        ),
+    )
+    warehouses = PrefetchingConnectionField(
+        Warehouse, description="List of warehouses."
+    )
+
+    @login_required
```
korycins: Do we need this? I assume that `permission_required` already requires logged in users.

saleor/graphql/warehouse/schema.py
```diff
@@ -0,0 +1,47 @@
...
+    @permission_required("warehouse.manage_warehouses")
+    def resolve_warehouse(self, _info, **data):
+        warehouse_pk = data.get("id")
+        warehouse = graphene.Node.get_node_from_global_id(
+            _info, warehouse_pk, Warehouse
+        )
+        return warehouse
+
+    @login_required
```
korycins: ☝️

saleor/graphql/warehouse/schema.py
```diff
@@ -0,0 +1,47 @@
...
+            graphene.ID, description="ID of an warehouse", required=True
+        ),
+    )
+    warehouses = PrefetchingConnectionField(
+        Warehouse, description="List of warehouses."
+    )
+
+    @login_required
+    @permission_required("warehouse.manage_warehouses")
+    def resolve_warehouse(self, _info, **data):
```
korycins: no need to attach `_` to `info` as you are using it in `get_node_from_global_id`. `_` is useful when you don't use the given parameter in the body of a method.

saleor/graphql/warehouse/types.py
```diff
@@ -0,0 +1,44 @@
+import graphene
+
+from saleor.graphql.account.enums import CountryCodeEnum
```
korycins: relative improts

tests/api/test_warehouse.py
```diff
@@ -0,0 +1,388 @@
...
+            ],
+        }
+    }
+
+    response = staff_api_client.post_graphql(
+        MUTATION_CREATE_WAREHOUSE, variables=variables
+    )
+    content = get_graphql_content(response)
+    assert Warehouse.objects.count() == 1
+    warehouse = Warehouse.objects.prefetch_data().first()
```
korycins: no need for `.prefetch_data()` as you are checking only `id` and `name`

tests/api/test_warehouse.py
```diff
@@ -0,0 +1,388 @@
...
+    assert not staff_api_client.user.has_perm("warehouse.manage_warehouses")
+    warehouse_id = graphene.Node.to_global_id("Warehouse", warehouse.pk)
+    variables = {"input": {"name": "New test name"}, "id": warehouse_id}
+    response = staff_api_client.post_graphql(
+        MUTATION_UPDATE_WAREHOUSE, variables=variables
+    )
+    content = get_graphql_content(response, ignore_errors=True)
+    errors = content["errors"]
+    assert len(errors) == 1
+    assert errors[0]["message"] == "You do not have permission to perform this action"
```
korycins: `tests.api.utils.assert_no_permission`

tests/api/test_warehouse.py
```diff
@@ -0,0 +1,388 @@
...
+    staff_api_client.user.user_permissions.add(permission_manage_warehouses)
+    warehouse_id = graphene.Node.to_global_id("Warehouse", warehouse.id)
+    warehouse_old_name = warehouse.name
+    warehouse_old_company_name = warehouse.company_name
+    variables = {
+        "id": warehouse_id,
+        "input": {
+            "name": "New name",
+            "companyName": "New name for company",
+            "shippingZones": [],
```
korycins: It would be  good to have test for validating adding/ removing shippingZones

szewczykmira:
TODO:
- [ ] shipping zone can be attached only to one Warehouse
- [ ] every country can be operated only by one Warehouse

saleor/graphql/schema.graphql
maarcingebala: Yep, for consistency with other top-level queries we shuold also have a `filter`.

saleor/graphql/warehouse/mutations.py
maarcingebala: I would hold on with that since I'd like to first think about refactoring our meta mutations to make them more generic.

tests/api/test_warehouse.py
szewczykmira: I always forgot about this one! :(

saleor/dashboard/urls.py
```diff
@@ -32,5 +33,6 @@
     url(r"^shipping/", include(shipping_urls)),
     url(r"^style-guide/", core_views.styleguide, name="styleguide"),
     url(r"^search/", include(search_urls)),
+    url(r"warehouses/", include(warehouse_urls)),
```
krzysztofwolski: missing `^` intended?

tests/dashboard/test_permissions.py
```diff
@@ -1032,3 +1033,61 @@ def test_only_staff_user_with_permission_can_update_payment_gateway(
     staff_user.user_permissions.add(permission_manage_plugins)
     response = staff_client.get(url)
     assert response.status_code == 200
+
+
+@pytest.mark.parametrize(
+    "url_name", ["dashboard:warehouse-list", "dashboard:warehouse-create"]
```
krzysztofwolski: We are dropping dashboard soon - not sure if we should add new tests and views in the meantime.

saleor/dashboard/urls.py
szewczykmira: ofc not ;( good catch!

szewczykmira:
@maarcingebala @korycins 
Quick recap:
From now on stock will be assigned per product variant and warehouse so we can say that every country might have different stock. I'm planning to get country from request and by passing it to checkout I will be able to select proper stock and allocate it. After choosing shipping address we can also revalidate stocks.

I have two questions about stock management that came to my mind today:
- what can we do with digital products? They don't have a shipping address so the order cannot allocate proper stock? Or do we assume that every digital product has infinite stock?
- In saleor there is feeds generation (for eg. google merchant) in which we inform if the product is in stock. As I've said before - there are many stocks it's not quite easy to define if variant is available - do you have some idea how to handle that?

@krzysztofwolski you're opinion would be helpful as well ;)

korycins:
> I'm planning to get country from request and by passing it to checkout 

Does it work for API? @maarcingebala 

> They don't have a shipping address so the order cannot allocate proper stock?

Shouldn't we just try to do something like in other places -  `address = checkout.shipping_address or checkout.billing_address` ?

>Or do we assume that every digital product has infinite stock?

We can't assume that as some shops can have limited quantities. Some shop can sell a limited number of software keys etc

> In saleor there is feeds generation (for eg. google merchant)

I am not sure if it sitll works as expected. This PR is too big to extend google_merchant also.  Probably this chanage will require some small modification - AFAIK we will need to create single file per warehouse - https://support.google.com/merchants/answer/7439058 -> `Create a primary feed` -> `Country of sale`. My proposition is to use for now avaiablity of the default Warehouse, then create a gh ticket for support feed per warehouse and someone will handle it before next release.

krzysztofwolski:
> From now on stock will be assigned per product variant and warehouse so we can say that every country might have different stock.

Awesome!

> I'm planning to get country from request and by passing it to checkout I will be able to select proper stock and allocate it. After choosing shipping address we can also revalidate stocks.

Country in the Checkout is way to go. My dream setup would be extending extension mgr for methods which will allow custom address validation (for example - dont allow shipping to the PO boxes) and for changing country (often when you change country you have to check for example checkout vouchers or other metadata). But IMO it's out of scope of this particular PR but shoud be keept in mind 

>  order cannot allocate proper stock? Or do we assume that every digital product has infinite stock?

The stock should be chosen based on checkout country. Same as currency should be.

> In saleor there is feeds generation (for eg. google merchant) in which we inform if the product is in stock. As I've said before - there are many stocks it's not quite easy to define if variant is available - do you have some idea how to handle that?

Let's go with default_country in settings.py

saleor/warehouse/models.py
```diff
@@ -0,0 +1,61 @@
...
+    def countries(self) -> Set[str]:
+        countries_zone = ",".join(
+            ShippingZone.objects.prefetch_related("warehouse_set")
+            .filter(warehouse=self)
+            .values_list("countries", flat=True)
+        )
+        return set(countries_zone.split(","))
+
+    def delete(self, *args, **kwargs):
+        self.address.delete()
```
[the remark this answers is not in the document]
szewczykmira: Unfortunately no, on_delete works another way around - it defines what happens when you delete the address. Just to be sure I've double-checked it ;)
But as I thought about that - the current solution is not perfect - when someone will delete address "by mistake" then warehouse will be deleted and stock will be deleted.
So we have two choices on how to solve it - we can either allow the address to be null or protecting address if it's associated with the warehouse. I prefer the latter as the address is the most important part of warehouse right now ;)

saleor/warehouse/migrations/0001_initial.py
```diff
@@ -0,0 +1,64 @@
+# Generated by Django 2.2.6 on 2019-11-27 11:22
```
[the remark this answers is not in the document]
szewczykmira: I thought about that, but I didn't find any address (something like company address) that I could use to create default warehouse. If I missed it then please let know so I would be happy to add it :)

saleor/stock/utils/availability.py
```diff
@@ -0,0 +1,60 @@
...
+
+from ...core.exceptions import InsufficientStock
+from ...product.models import Product
+from ..models import Stock
+
+if TYPE_CHECKING:
+    from ...product.models import ProductVariant
+
+
+def check_stock_quantity(variant: "ProductVariant", country_code: str, quantity: int):
```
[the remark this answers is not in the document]
szewczykmira: Yes, this was its original logic ;)

saleor/stock/utils/availability.py
```diff
@@ -0,0 +1,60 @@
...
+def is_variant_in_stock(variant: "ProductVariant", country_code: str) -> bool:
+    stock = Stock.objects.for_country(country_code).get(product_variant__pk=variant.pk)
+    return stock.is_available
+
+
+def stocks_for_product(product: "Product", country_code: str):
+    return (
+        Stock.objects.annotate_available_quantity()
+        .for_country(country_code)
+        .filter(product_variant__product_id=product.pk)
```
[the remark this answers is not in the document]
szewczykmira: > Don't you forget about.first()?

No, a variant has only one stock per warehouse, but a product has many (as it has many variants)

> Don't you forget about prefetch_realted?

No worries, in `for_country()` there is `self.select_related("product_variant", "warehouse")` :)

saleor/stock/utils/availability.py
```diff
@@ -0,0 +1,60 @@
...
+
+def is_product_in_stock(product: "Product", country_code: str) -> bool:
+    return any(
+        stocks_for_product(product, country_code).values_list(
+            "available_quantity", flat=True
+        )
+    )
+
+
+def are_all_product_variants_in_stock(product: "Product", country_code: str) -> bool:
```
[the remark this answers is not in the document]
szewczykmira: I've started to use it today in `product.utils.availability.get_product_availability_status`

saleor/graphql/stock/schema.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+    stocks = FilterInputConnectionField(
+        Stock,
+        description="List of stocks.",
+        filter=StockFilterInput(),
+        query=graphene.String(),
+    )
+
+    @permission_required("stock.manage_stocks")
+    def resolve_stock(self, info, **kwargs):
```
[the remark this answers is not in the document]
szewczykmira: Good to know, as this code is based on https://github.com/mirumee/saleor/blob/master/saleor/graphql/shipping/schema.py#L31 and that means we might have issues in some part of API.

saleor/graphql/stock/schema.py
```diff
@@ -0,0 +1,45 @@
...
+from ..decorators import permission_required
+from .filters import StockFilterInput
+from .mutations import StockBulkDelete, StockCreate, StockDelete, StockUpdate
+from .types import Stock
+
+
+class StockQueries(graphene.ObjectType):
+    stock = graphene.Field(
+        Stock,
+        description="Look up a stok by ID",
```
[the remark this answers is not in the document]
szewczykmira: It's still not my best typo - once I've forgotten to put `f` in `shift` :) And this was in commit message ;)

saleor/checkout/models.py
```diff
@@ -148,6 +148,17 @@ def get_last_active_payment(self):
         payments = [payment for payment in self.payments.all() if payment.is_active]
         return max(payments, default=None, key=attrgetter("pk"))
 
+    def set_country(self, country_code):
+        if self.country is None:
+            self.store_private_meta("Shipping_info", "", {"country_code": country_code})
+            # self.save(update_fields=["private_meta"])
```
krzysztofwolski: In progress?

saleor/graphql/warehouse/filters.py
```diff
@@ -0,0 +1,37 @@
...
+        | Q(address__street_address_2__icontains=value)
+        | Q(address__city__icontains=value)
+        | Q(address__postal_code=value)
+        | Q(address__phone__contains=value)
+    )
+    return qs
+
+
+class WarehouseFilter(django_filters.FilterSet):
+    name = django_filters.CharFilter(method=filter_string)
```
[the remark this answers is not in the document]
krzysztofwolski: I would suggest adding some kind of slug or `external_id` field - warehouses will be most likely synced with externa;l systems so it would be nice to assign remote resource id.

saleor/checkout/models.py
szewczykmira: yep

saleor/graphql/warehouse/forms.py
```diff
@@ -0,0 +1,37 @@
...
+        warehouses = set(
+            shipping_zones.filter(warehouse__isnull=False).values_list(
+                "warehouse", flat=True
+            )
+        )
+        if len(warehouses) == 0:
+            return True
+        if len(warehouses) > 1:
+            return False
+        instance_id = self.instance.id
```
krzysztofwolski: Can `len(warehouses)` return <0?

saleor/graphql/warehouse/forms.py
```diff
@@ -0,0 +1,37 @@
...
+def get_available_shipping_zones():
+    return ShippingZone.objects.filter(warehouse__isnull=True)
+
+
+class WarehouseForm(forms.ModelForm):
+    class Meta:
+        model = Warehouse
+        fields = ["name", "company_name", "shipping_zones", "email"]
+
+    def validate_warehouse_count(self, shipping_zones) -> bool:
```
krzysztofwolski: Docstring on that would be really helpfull

saleor/stock/availability.py
```diff
@@ -0,0 +1,74 @@
...
+        .all()
+    )
+    are_all_available = all([elem[0] for elem in product_stocks])
+    variants_with_stocks = [elem[1] for elem in product_stocks]
+
+    product_variants = product.variants.exclude(id__in=variants_with_stocks).exists()
+    return are_all_available and not product_variants
+
+
+def products_with_low_stock(threshold: Optional[int] = None):
```
krzysztofwolski: Module is missing a docstrings - will be helpfull especially explaining whats low stock means or where its taken from. BTW. Is getattrnew recommended way of getting settings? Because in current form  it's having default value in the code instead of settings - IMO it's making it hidden from the user.

saleor/stock/error_codes.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
+
+
+class StockErorrCode(Enum):
```
krzysztofwolski: It's recommended to use inheritance from (str, Enum)

saleor/warehouse/models.py
```diff
@@ -0,0 +1,62 @@
...
+        return self.name
+
+    @property
+    def countries(self) -> Set[str]:
+        countries_zone = ",".join(
+            ShippingZone.objects.prefetch_related("warehouse_set")
+            .filter(warehouse=self)
+            .values_list("countries", flat=True)
+        )
+        return set(countries_zone.split(","))
```
krzysztofwolski: Join and split is some kind of force `str`? Or is there different purpose?

saleor/warehouse/models.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class WarehouseQueryset(models.QuerySet):
+    def prefetch_data(self):
+        return self.select_related("address").prefetch_related("shipping_zones")
+
+
+class Warehouse(models.Model):
+    id = models.UUIDField(default=uuid.uuid4, primary_key=True)
+    name = models.CharField(
```
krzysztofwolski: Will the name be presented?  If so, do we need translations for its name?

szewczykmira:
@korycins @maarcingebala @dominik-zeglen 
I have a question about https://github.com/mirumee/saleor/blob/master/saleor/graphql/product/filters.py#L101 - where is this used? As you remember stock is set by warehouse/country so if this is a query for storefront then I guess I need to filter it also by country. If this is for dashboard the product should be filtered by... all stocks? stocks for the selected country?
I'm not sure how to handle that so all suggestions are welcome :)

tests/dashboard/test_permissions.py
szewczykmira: This PR will be also added to legacy dashboard

saleor/warehouse/models.py
szewczykmira: Okay, I found that getting `shipping_zones.values_list("countries", flat=True)` returns a string with country code separated by commas and `shipping_zone.countries` returns a list of object :) I will add simplified version (that uses python objects) with next bach of commits :)

saleor/graphql/stock/schema.py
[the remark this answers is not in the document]
szewczykmira: As we talked few days ago on daily - this is query not type definition ;)

krzysztofwolski:
@szewczykmira one more thing - we are overall aiming at moving into 'thin models' and the Repository Pattern in upcoming months. It would be a good idea to not add any model methods if we know those will be altered. For inspiration 

Can't wait to put my hands on warehouses and add extensions to it, great feature! :)

saleor/graphql/warehouse/forms.py
```diff
@@ -0,0 +1,37 @@
...
+
+from ...shipping.models import ShippingZone
+from ...warehouse.models import Warehouse
+
+
+def get_available_shipping_zones():
+    return ShippingZone.objects.filter(warehouse__isnull=True)
+
+
+class WarehouseForm(forms.ModelForm):
```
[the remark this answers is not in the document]
szewczykmira: Finally, I've decided to delete `ModelForm` altogether.

saleor/graphql/stock/schema.py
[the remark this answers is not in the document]
szewczykmira: Same as above, this is query not type

saleor/graphql/warehouse/filters.py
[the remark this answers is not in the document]
szewczykmira: @krzysztofwolski I think this PR is huge enough and we can skip this part for now ;) Especially that we don't have any specific requirements about that.

szewczykmira:
@maarcingebala @korycins @krzysztofwolski 
Hi, it would be awesome if you will find time to review this PR. Currently, my focus is mostly to get it ready for legacy views (and merge PR in the legacy repository) and after that to make sure that everything works via API and all changes introduced in the main repo.

saleor/graphql/product/types/products.py
```diff
@@ -311,16 +318,25 @@ class Meta:
...
+    @permission_required("stock.manage_stocks")
+    def resolve_stocks(root: models.ProductVariant, *_args):
+        stocks = stock_models.Stock.objects.select_related("warehouse").filter(
+            product_variant=root
+        )
+        return stocks
+
+    @staticmethod
+    @permission_required("stock.manage_stocks")
+    def resolve_stock_for_country(root: models.ProductVariant, info):
```
korycins: It would be better if we will not assume that we have single stock for country (it can be changed in the future).  You have permission for both resolvers which means that customer will not be able to fetch current stock. Am I right?
Maybe we could have single resolver and  always return all avaiable stocks and storefront would decide how to handle the data?

korycins: cc @szewczykmira  @maarcingebala

saleor/graphql/product/types/products.py
```diff
@@ -567,8 +578,10 @@ def resolve_pricing(root: models.Product, info):
 
     @staticmethod
     @gql_optimizer.resolver_hints(prefetch_related=("variants"))
-    def resolve_is_available(root: models.Product, _info):
-        return root.is_available
+    def resolve_is_available_in_country(root: models.Product, _info):
```
korycins: Same question as above ☝️

saleor/graphql/schema.graphql
```diff
@@ -3300,12 +3307,9 @@ type ProductVariant implements Node {
   sku: String!
   product: Product!
   trackInventory: Boolean!
-  quantityAllocated: Int!
```
korycins: AFAIK We should not drop fields in API in that way. We should mark them as a deprecated with the info that we will drop them in the next release

saleor/graphql/stock/types.py
```diff
@@ -0,0 +1,31 @@
...
+
+class StockInput(graphene.InputObjectType):
+    product_variant = graphene.ID(
+        required=True, description="Product variant assiociated with stock."
+    )
+    warehouse = graphene.ID(
+        required=True, description="Warehouse in which stock is lockated."
+    )
+    quantity = graphene.Int(description="Quantity of items available for sell.")
+    quantity_allocated = graphene.Int()
```
korycins: Don't you need default value here? like `0`?

saleor/graphql/warehouse/filters.py
```diff
@@ -0,0 +1,40 @@
...
+        "name",
+        "company_name",
+        "email",
+        "address__street_address_1",
+        "address__street_address_2",
+        "address__city",
+        "address__postal_code",
+        "address__phone",
+    ]
+    qs = prefech_qs_for_filter(qs)
```
korycins: Do we wnat to `prefetch_qs_for_filter` event value can be null?

korycins:
```
if value:
    prefetch_qs_for_filter(qs)
    qs = filter_by_query_param(qs, value, search_fields)
```

saleor/graphql/warehouse/mutations.py
```diff
@@ -0,0 +1,113 @@
...
+            return False
+        return warehouses == {instance.id}
+
+    @classmethod
+    def clean_input(cls, info, instance, data):
+        cleaned_input = super().clean_input(info, instance, data)
+        shipping_zones = cleaned_input.get("shipping_zones", [])
+        if not cls.validate_warehouse_count(shipping_zones, instance):
+            msg = "Shipping zone can be assigned only to one warehouse."
+            raise ValidationError({"shipping_zones": msg})
```
korycins: AFAIK You should provide Error code here

saleor/graphql/warehouse/mutations.py
```diff
@@ -0,0 +1,113 @@
...
+
+class WarehouseMixin:
+    @classmethod
+    def validate_warehouse_count(cls, shipping_zones, instance):
+        """Every ShippingZone can be assigned to only one warehouse.
+
+        If not there would be issue with automatically selecting stock for operation.
+        """
+
+        warehouses = set(
```
korycins: I am watching at it over 10 minutes and I am not able to understand how it works. Can you provide some description? HEre or in the doc string?

saleor/graphql/warehouse/mutations.py
```diff
@@ -0,0 +1,113 @@
...
+    def prepare_address(cls, cleaned_data, *args):
+        address_form = cls.validate_address_form(cleaned_data["address"])
+        return address_form.save()
+
+
+class WarehouseUpdate(WarehouseMixin, ModelMutation, I18nMixin):
+    class Meta:
+        model = models.Warehouse
+        permissions = ("warehouse.manage_warehouses",)
+        description = "Update given warehouse."
```
korycins: `Updates`

saleor/product/migrations/0111_auto_20191209_0349.py
```diff
@@ -0,0 +1,78 @@
...
+
+    warehouse = get_or_create_warehouse(apps)
+    stocks = [
+        Stock(
+            product_variant=variant,
+            warehouse=warehouse,
+            quantity=variant.quantity,
+            quantity_allocated=variant.quantity_allocated,
+        )
+        for variant in ProductVariant.objects.all()
```
korycins: I would say that `.iterator` will be safer.

saleor/product/migrations/0111_auto_20191209_0349.py
```diff
@@ -0,0 +1,78 @@
+# Generated by Django 2.2.7 on 2019-12-09 09:49
+from django.db import migrations
+from django.db.models import Count, Sum
+
+
+def get_or_create_warehouse(apps):
+    Warehouse = apps.get_model("warehouse", "Warehouse")
```
korycins: I think that this logic should be as a second step inside a `warehouse.init_migration`

saleor/product/migrations/0111_auto_20191209_0349.py
```diff
@@ -0,0 +1,78 @@
...
+    if warehouses.first() is not None:
+        return warehouses.first()
+
+    warehouse = Warehouse.objects.create(address=address)
+    warehouse.shipping_zones.add(*ShippingZone.objects.all())
+    return warehouse
+
+
+def forward(apps, schema_editor):
+    ProductVariant = apps.get_model("product", "ProductVariant")
```
korycins: IMHO filling all stock records should be done inside a `stock.init_migration`

saleor/product/migrations/0112_auto_20191209_0437.py
```diff
@@ -0,0 +1,13 @@
+# Generated by Django 2.2.7 on 2019-12-09 10:37
```
korycins: If you move warehouse and stock logic from 0111 to warehouse.init_migration and stock.Init_migration, then you should be able to easily create one migration for product(instead of 111 and 112)

saleor/warehouse/migrations/0002_auto_20191205_0753.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.7 on 2019-12-05 13:53
```
korycins: Let's merge these migrations into one

saleor/graphql/warehouse/mutations.py
krzysztofwolski: BTW - is that method called automatically by graphene or have to be called explicit?

saleor/graphql/product/types/products.py
```diff
@@ -86,13 +88,16 @@ def resolve_attribute_list(
...
     resolved_attributes = []
+    attributes_qs = None
 
     # Retrieve the product type
     if isinstance(instance, models.Product):
         product_type = instance.product_type
         product_type_attributes_assoc_field = "attributeproduct"
         assigned_attribute_instance_field = "productassignments"
         assigned_attribute_instance_filters = {"product_id": instance.pk}
+        if hasattr(product_type, "storefront_attributes"):
```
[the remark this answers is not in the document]
szewczykmira: I'm pretty sure I know how - this branch is currently rebased on the lovel of legacy - I didn't touch it :D

saleor/graphql/product/types/products.py
szewczykmira: Hmm, that seems ok for me, I'll just wait for @maarcingebala to confirm

saleor/graphql/schema.graphql
szewczykmira: Ok, we can discuss during the daily how to handle this properly

saleor/graphql/stock/types.py
szewczykmira: Default value is set in DB (to 0)

saleor/graphql/warehouse/mutations.py
szewczykmira: @krzysztofwolski it must be called explicit (see `clean_input`)

krzysztofwolski: @szewczykmira lets move it outside of the mixin:
- much easier testing 
- it will be possible to use it outside of the mutations (what if I want to make some operations in management command?)

korycins: @krzysztofwolski doesn't want to have any business logic inside mutations ;) 👍

saleor/product/migrations/0111_auto_20191209_0349.py
szewczykmira: In the beginning, I was thinking the same. But I don't want to force-create someone object in DB if I don't need it. Besides - someone can delete it between migrations.

korycins: I am not sure that I understand it correctly. `But I don't want to force-create someone object in DB if I don't need it.` -> the result will be the same for both approaches.
`Besides - someone can delete it between migrations.` -> We could solve the proble by using `dependencies` between given migrations

szewczykmira: `the result will be the same for both approaches.` - not exactly. In your approach, we will create a warehouse no matter if we really need it or not. In my approach - we will create warehouse if a) there are some product variants created and b) there is no other warehouse that supports all shipping zones.

dependencies do not guarantee that created warehouse will still exist in time for migrating stocks.

saleor/graphql/schema.graphql
maarcingebala: Yep we cannot remove fields from the schema like that, we should deprecate them and say in which Saleor version it will be removed ( in this case 2.11 - two versions from the current).

saleor/graphql/schema.graphql
```diff
@@ -3319,6 +3323,8 @@ type ProductVariant implements Node {
   images: [ProductImage]
   translation(languageCode: LanguageCodeEnum!): ProductVariantTranslation
   digitalContent: DigitalContent
+  stocks: Stock
```
maarcingebala: Idea - why not `stock(country:  CountryCode): Stock` instead of these two queries?

saleor/graphql/schema.graphql
```diff
@@ -2930,7 +2937,7 @@ type Product implements Node {
   thumbnail(size: Int): Image
   availability: ProductPricingInfo @deprecated(reason: "DEPRECATED: Will be removed in Saleor 2.10, Has been renamed to `pricing`.")
   pricing: ProductPricingInfo
-  isAvailable: Boolean
+  isAvailableInCountry: Boolean
```
maarcingebala: `isAvailable(country: CountryCode): Boolean` - for easier adoption country could be nullable if we have some way to return a default if none is provided.

saleor/graphql/schema.graphql
```diff
@@ -4401,6 +4482,91 @@ type VoucherUpdate {
...
 }
 
+type Warehouse implements Node {
+  id: ID!
+  name: String!
+  companyName: String!
+  shippingZones(before: String, after: String, first: Int, last: Int): ShippingZoneCountableConnection!
+  address: Address!
+  email: String!
+  stockSet(before: String, after: String, first: Int, last: Int): StockCountableConnection!
```
maarcingebala: Is this field here on purpose? It looks like it was automatically added by Graphene.

saleor/graphql/stock/mutations.py
```diff
@@ -0,0 +1,59 @@
...
+        model = models.Stock
+        permissions = ("stock.manage_stocks",)
+        description = "Update given stock."
+        error_type_class = StockError
+        error_type_field = "stock_error"
+
+
+class StockDelete(ModelDeleteMutation):
+    class Arguments:
+        id = graphene.ID(required=True, description="ID fo stock to delete.")
```
maarcingebala:
```suggestion
        id = graphene.ID(required=True, description="ID of stock to delete.")
```

saleor/graphql/stock/schema.py
```diff
@@ -0,0 +1,45 @@
...
+    stocks = FilterInputConnectionField(
+        Stock,
+        description="List of stocks.",
+        filter=StockFilterInput(),
+        query=graphene.String(),
+    )
+
+    @permission_required("stock.manage_stocks")
+    def resolve_stock(self, info, **kwargs):
+        stock_pk = kwargs.get("id")
```
maarcingebala: Is this Base64 ID or pk -primary key? If the former then I would name it `stock_id`. We use both across our API so it's good to have a clear distinction in naming.

saleor/graphql/stock/types.py
maarcingebala: The description is missing here as well.

saleor/graphql/stock/schema.py
```diff
@@ -0,0 +1,45 @@
...
+        stock = graphene.Node.get_node_from_global_id(info, stock_pk, Stock)
+        return stock
+
+    @permission_required("stock.manage_stocks")
+    def resolve_stocks(self, info, **data):
+        qs = (
+            models.Stock.objects.select_related("warehouse")
```
maarcingebala: Not a big issue but we can also use `gql_optimizer` for that to dynamically add these prefetches only when these fields are requested. See how in some types we use `gql_optimizer` to decorate resolvers.

saleor/checkout/models.py
```diff
@@ -148,6 +148,25 @@ def get_last_active_payment(self):
         payments = [payment for payment in self.payments.all() if payment.is_active]
         return max(payments, default=None, key=attrgetter("pk"))
 
+    def set_country(self, country_code: str, commit: bool = False, replace=False):
```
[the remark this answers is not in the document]
krzysztofwolski: Default `replace=False` seems strange.

saleor/checkout/models.py
```diff
@@ -148,6 +148,25 @@ def get_last_active_payment(self):
...
+        if not replace and self.country is not None:
+            return
+        self.store_private_meta("Shipping_info", "", {"country_code": country_code})
+        if commit:
+            self.save(update_fields=["private_meta"])
+
+    @property
+    def country(self):
+        address = self.shipping_address or self.billing_address
```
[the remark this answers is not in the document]
krzysztofwolski: Can we have `get_country` method? Properties are always confusing when used in models.

saleor/checkout/models.py
```diff
@@ -148,6 +148,25 @@ def get_last_active_payment(self):
         payments = [payment for payment in self.payments.all() if payment.is_active]
         return max(payments, default=None, key=attrgetter("pk"))
 
+    def set_country(self, country_code: str, commit: bool = False, replace=False):
+        if not replace and self.country is not None:
+            return
+        self.store_private_meta("Shipping_info", "", {"country_code": country_code})
```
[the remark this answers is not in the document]
krzysztofwolski: For sure there is need for storing client country:
- new user adds sth to the cart
- no address is set 
- all we have is geoip/country user selected in country selector (if storefront has one)

saleor/graphql/stock/types.py
```diff
@@ -0,0 +1,31 @@
...
+
+    class Meta:
+        description = "Represents stock."
+        model = models.Stock
+        interfaces = [graphene.relay.Node]
+
+    @staticmethod
+    @permission_required("stock.manage_stocks")
+    def resolve_available(root, *_args):
+        return min(root.quantity_available, settings.MAX_CHECKOUT_LINE_QUANTITY)
```
krzysztofwolski: Stock needs `get_user_presentable_quantity_available` util. No logic in API! I guess Google feeds should use the same value as well if presenting available quantity.

saleor/order/utils.py
```diff
@@ -15,12 +15,22 @@
...
 from ..order.models import Order, OrderLine
-from ..product.utils import allocate_stock, deallocate_stock, increase_stock
 from ..product.utils.digital_products import get_default_digital_content_settings
 from ..shipping.models import ShippingMethod
+from ..stock.availability import check_stock_quantity
+from ..stock.management import allocate_stock, deallocate_stock, increase_stock
 from . import events
 
 
+def get_order_country(order) -> str:
```
krzysztofwolski: Types, docstrings

saleor/graphql/checkout/mutations.py
```diff
@@ -259,6 +262,8 @@ def save_addresses(cls, instance: models.Checkout, cleaned_input: dict):
     def save(cls, info, instance: models.Checkout, cleaned_input):
         # Create the checkout object
         instance.save()
+        country = info.context.country
```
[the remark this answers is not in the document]
szewczykmira: IMO this is outside fo the scope - this PR is already huge and I don't think we should add any more feature here.

saleor/graphql/schema.graphql
szewczykmira: So how do you suggest to handle this? With warehouses and stocks this field will blow up and returns an error.

maarcingebala: We need to provide some workaround to return here a meaningful value. E.g. do we have the concept of "allocated quantity" with warehouses? I guess we do but now it's kept at the stock level? Are we able to calculate the sum of allocated quantities for a single variant in all stocks? That could be one solution.

saleor/graphql/stock/types.py
szewczykmira: About Google Feed, please discuss this with Saleor Team as this is outside of scope for this PR.

saleor/warehouse/models.py
szewczykmira: The name will be presented only in the dashboard. If there is any need for translating this - I do not know anything about this.

saleor/graphql/schema.graphql
szewczykmira: @maarcingebala I will think about this after testing current changes as everything might change after that :) There is allocated quantity in stocks, but I will need to check where this field is used as for storefront we can define quantity allocated by country, but for the dashboard, we would need the sum

saleor/checkout/models.py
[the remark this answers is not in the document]
szewczykmira: @maarcingebala I wonder - isn't there a chance to have stocks for digital products as well?
For eg shop selling access codes for games. They can have limited stock for every country/shipping zone. Actually I'm not sure if in that case, we should allocate stock by billing address or by country from geolocalization...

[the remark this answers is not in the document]
maarcingebala: Not sure how it works now, we should ask @korycins who implemented the backend part of digital products.

saleor/graphql/stock/types.py
krzysztofwolski: Having one function for user presentable available amount is in scope of this PR. Feeds is just an example.

saleor/graphql/schema.graphql
szewczykmira: yeah, it's been added automatically

saleor/graphql/warehouse/filters.py
```diff
@@ -0,0 +1,40 @@
...
+        qs = filter_by_query_param(qs, value, search_fields)
+    return qs
+
+
+class WarehouseFilter(django_filters.FilterSet):
+    search = django_filters.CharFilter(method=filter_search)
+
+    class Meta:
+        model = Warehouse
+        fields = []
```
korycins: `fields = ['search']` or AFAIK you could remove this line

saleor/order/utils.py
```diff
@@ -13,12 +13,23 @@
...
+from ..stock.management import allocate_stock, deallocate_stock, increase_stock
 from . import events
 
 
+def get_order_country(order: Order) -> str:
+    """Return country to which order will be shipped."""
+    address = order.billing_address
+    if order.is_shipping_required():
+        address = order.shipping_address
+    if address is None:
```
korycins: IMO we should try to fetch `order.billing_address` first, then we can try with `DEFAULT_COUNTRY`

saleor/product/utils/availability.py
```diff
@@ -144,6 +148,12 @@ def get_product_availability(
     )
 
     is_on_sale = product.is_visible and discount is not None
+    if country is not None:
```
korycins: Instead of `if else` condition, You could do something like this:
`country = country or DEFAULT_COUNTRY`
and you will have only a single call of `is_product_in_stock`

korycins: but this is only suggestion :)

saleor/stock/management.py
```diff
@@ -0,0 +1,42 @@
...
+
+if TYPE_CHECKING:
+    from ..product.models import ProductVariant
+
+
+def allocate_stock(
+    variant: "ProductVariant", country_code: str, quantity: int, commit: bool = True
+) -> Stock:
+    stock = Stock.objects.for_country(country_code).get(product_variant=variant)
+    stock.allocate_stock(quantity, commit=commit)
```
korycins: We should lock database when we are changing a quantity. `select_for_update` would be helpful

tests/api/test_page.py
```diff
@@ -360,6 +360,7 @@ def test_query_pages_with_sort(
     response = staff_api_client.post_graphql(QUERY_PAGE_WITH_SORT, variables)
     content = get_graphql_content(response)
     pages = content["data"]["pages"]["edges"]
+    print(result_order)
```
korycins: `debug print`

tests/api/test_product.py
```diff
@@ -910,7 +878,6 @@ def test_create_product(
         $name: String!,
         $basePrice: Decimal!,
         $sku: String,
-        $quantity: Int,
```
korycins: IMHO we shouldn't remove the tests for deprecated part of code. We will drop it at the same time as deprecated fields

tests/test_stock.py
```diff
@@ -0,0 +1,115 @@
...
+    assert COUNTRY_CODE in warehouse.countries
+    assert stock.warehouse == warehouse
+
+    stock_qs = Stock.objects.for_country(COUNTRY_CODE)
+    assert stock_qs.count() == 1
+    assert stock_qs.first() == stock
+
+
+def test_stock_for_country_does_not_exists(product, warehouse):
+    fake_country_code = "OO"
```
korycins: Can we check also a real country which is not assigned to the ​given warehouse?

saleor/checkout/migrations/0022_checkout_country.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [("checkout", "0022_auto_20191219_1137")]
+
+    operations = [
+        migrations.AddField(
+            model_name="checkout",
+            name="country",
+            field=django_countries.fields.CountryField(default="US", max_length=2),
```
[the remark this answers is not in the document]
szewczykmira: Damn, it's tricky one as in code there is `country = CountryField(default=settings.DEFAULT_COUNTRY)` I think I'll try to use some method there to get this value from settings.

saleor/graphql/product/mutations/products.py
```diff
@@ -887,13 +887,9 @@ def save(cls, info, instance, cleaned_input):
...
                 "track_inventory", site_settings.track_inventory_by_default
             )
-            quantity = cleaned_input.get("quantity", 0)
             sku = cleaned_input.get("sku")
             models.ProductVariant.objects.create(
-                product=instance,
-                track_inventory=track_inventory,
-                sku=sku,
-                quantity=quantity,
+                product=instance, track_inventory=track_inventory, sku=sku
```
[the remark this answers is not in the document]
szewczykmira: `ProductVariant` has no field quantity so we cannot

saleor/graphql/product/types/products.py
```diff
@@ -187,13 +193,19 @@ class Meta:
...
 class ProductVariant(CountableDjangoObjectType, MetadataObjectType):
+    # DEPRECATED
     quantity = graphene.Int(
         required=True,
-        description="Quantity of a product in the store's possession, "
+        description="DEPRECATED: Will be removed in favour of stock. "
+        "Quantity of a product in the store's possession, "
         "including the allocated stock that is waiting for shipment.",
     )
+    quantity_allocated = graphene.Int(required=False, description="DEPRECATED")
```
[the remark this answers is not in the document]
szewczykmira: This seems like good idea, but I'm not sure in which version we will drop those fields ;)

saleor/graphql/warehouse/filters.py
szewczykmira: So for a model, we must define either `fields` or `exclude`. But I'm not sure it's a good idea to set it to `search` as this is not one of model field. @maarcingebala  what do you think?

saleor/order/utils.py
szewczykmira: I'm not sure I understand that - we are checking `billing address` at the beginning.

tests/api/test_product.py
szewczykmira: Hmm, I have a question though - this is the mutation that creates a product with a variant, but right now variant doesn't have `quantity` field. It's easier to keep fields as deprecated in queries, but in mutation?

saleor/graphql/schema.graphql
maarcingebala: Yes in the `Warehouse`'s metaclass we should add `only_fields` where we'll explicitly define which fields should be converted to GraphQL type from the model.

CHANGELOG.md
```diff
@@ -46,6 +46,9 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add mypy checks - #5150 by @IKarbowiak
 - Move extracting user or service_account from context to utils - #5152 by @kswiatek92
 - Add deprecate description to order status/created arguments - #5076 by @kswiatek92
+- Fixed serialization error on weight fields when running `loaddata` and `dumpdb` - #5005 by @NyanKiyoshi
+- Add warehouse model - #4986 by @szewczykmira
+- Assign quantity by warehouse - #4986 by @szewczykmira
```
maarcingebala: I think one entry for this entire PR /feature is fine.

saleor/graphql/warehouse/filters.py
maarcingebala: If there are no other filters than `search` I think this should be empty. If I'm correct it's mandatory to have either `fields` or `exclude`.

tests/test_core.py
```diff
@@ -125,12 +125,12 @@ def test_create_fake_users(db):
 
 
 def test_create_address(db):
-    assert Address.objects.all().count() == 0
+    assert not Address.objects.exists()
```
krzysztofwolski:
```suggestion
    counter = Address.objects.all().count()
```

tests/test_core.py
```diff
@@ -125,12 +125,12 @@ def test_create_fake_users(db):
...
     random_data.create_address()
     assert Address.objects.all().count() == 1
```
krzysztofwolski:
```suggestion
    assert Address.objects.all().count() == counter + 1
```

This way test is not depending on empty database.

szewczykmira: I don't like this. That way we are missing information whether other conftest are adding additional data. Also checking if an object is saved with proper data will require additional check - if there is already something in the database or not.

krzysztofwolski: When maintaining forks we are sometimes creating preexisting objects in database (for example main company address). If we add such fixture, then all tests with 'frozen' counters will fail.

szewczykmira: I'm not sure I follow - you are adding objects in migrations, right? Are they required in migrations or could be replaced with fixture used with `Django admin`? Is there any discussion about that so I could read more about this issue?
I understand your point, but I also see other issues. For eg 
- what should happen when we want to check if proper data are saved to DB? Couse `Model.objects.first()` might fail. 
- what if we have a field with `unique=True` and we have a fixture with `autouse=True` that sets this field to the same value as we set it in our test? We've got an error and it will take additional time to find why this is happening.

saleor/graphql/product/types/products.py
```diff
@@ -332,13 +370,16 @@ def resolve_quantity_ordered(root: models.ProductVariant, *_args):
...
-        return root.quantity_allocated
+    def resolve_quantity_allocated(root: models.ProductVariant, info):
+        country = info.context.country
+        stock = stock_models.Stock.objects.get_variant_stock_for_country(country, root)
+        return stock.quantity_allocated
 
     @staticmethod
-    @permission_required(ProductPermissions.MANAGE_PRODUCTS)
+    @permission_required(
+        [OrderPermissions.MANAGE_ORDERS, ProductPermissions.MANAGE_PRODUCTS]
```
maarcingebala: I think we've changed that on master to only require `MANAGE_PRODUCTS` permission.

saleor/graphql/product/types/products.py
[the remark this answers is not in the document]
maarcingebala: Wouldn't it be useful to keep all of those `quantity_*` fields but only provide additional param `country: CountryCode` (or any other meaningful way to identify warehouse underneath)? I think it's still pretty convenient API to fetch the quantities through products/variants.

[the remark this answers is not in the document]
maarcingebala: OK nevermind my last comment, this is still pretty convenient to use if we can get all of that from the `stock` field.

saleor/graphql/product/types/products.py
szewczykmira: @maarcingebala Fixed

saleor/graphql/product/types/products.py
```diff
@@ -189,11 +196,15 @@ class Meta:
 class ProductVariant(CountableDjangoObjectType, MetadataObjectType):
     quantity = graphene.Int(
         required=True,
-        description="Quantity of a product in the store's possession, "
+        description="DEPRECATED: Will be removed in favour of stock. "
```
maarcingebala: All of these deprecated fields are regular Graphene fields that provide `deprecation_reason` for the deprecation message. We should revert the original descriptions and use `deprecation_reason` instead. The message can be the same in all these cases: "This field will be removed in Saleor 2.11. Use the `stock` field instead.`

saleor/graphql/warehouse/types.py
```diff
@@ -0,0 +1,80 @@
...
+
+class StockInput(graphene.InputObjectType):
+    product_variant = graphene.ID(
+        required=True, description="Product variant assiociated with stock."
+    )
+    warehouse = graphene.ID(
+        required=True, description="Warehouse in which stock is lockated."
+    )
+    quantity = graphene.Int(description="Quantity of items available for sell.")
+    quantity_allocated = graphene.Int(description="Quantity allocated for orders.")
```
maarcingebala: In what cases we want to manually set allocated quantity? You should be able to define the quantity in stock, but this one is calculated based on checkouts and orders.

saleor/graphql/warehouse/types.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class Stock(CountableDjangoObjectType):
+    available = graphene.Int(description="Quantity of a product available for sale.")
+
+    class Meta:
```
maarcingebala: We should add `only_fields` here as well to explicitly say what fields are available in this type.

saleor/warehouse/models.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def __str__(self):
+        return f"{self.product_variant} - {self.warehouse.name}"
+
+    @property
+    def quantity_available(self) -> int:
+        return max(self.quantity - self.quantity_allocated, 0)
+
+    @property
+    def is_available(self):
```
maarcingebala: What do you think about renaming it to `is_in_stock`? We used to have a function like this and it was checking if the quantity available is above zero so that would be similar to the previous implementation. Also, we tend to describe a product as available if it is in stock and it's published.

saleor/graphql/warehouse/types.py
szewczykmira: Good catch - you're right - we should not define it by ourselves.

saleor/warehouse/models.py
szewczykmira: We have two functions added - `is_variant_in_stock` and `is_product_in_stock` that should be used in other places (they use `stock.is_available`). I'm open to suggestion to rename this property but not exactly for `is_in_stock`. If we call it as `stock.is_in_stock` then this sounds funny ;)

saleor/graphql/warehouse/types.py
```diff
@@ -0,0 +1,80 @@
...
+        required=True, description="Product variant assiociated with stock."
+    )
+    warehouse = graphene.ID(
+        required=True, description="Warehouse in which stock is lockated."
+    )
+    quantity = graphene.Int(description="Quantity of items available for sell.")
+
+
+class Stock(CountableDjangoObjectType):
+    available = graphene.Int(description="Quantity of a product available for sale.")
```
maarcingebala: This type exposes three quantities:

```
  quantity: Int!
  quantityAllocated: Int!
  available: Int
```

It's really important that we provide meaningful names and descriptions here, otherwise, it will be really confusing to understand and use.

- We need to override `quantity` and `quantity_allocated` fields in this Graphene type so that we can provide descriptions. I would take descriptions from the fields that this PR deprecates in the `ProductVariant` type.
- the field `available` is nullable in it the schema - is it on purpose? I think we will always have an integer value here.
- if I'm correct the field `available` does what `ProductVariant.stockQuantity` used to do. Can we use the same name for this field then?

saleor/warehouse/models.py
maarcingebala: I don't have any better idea.. maybe it can stay as it is- it's not that big deal.

szewczykmira:
@maarcingebala all updated according to your suggestions

saleor/graphql/schema.graphql
```diff
@@ -3485,8 +3493,12 @@ type Query {
   webhooks(sortBy: WebhookSortingInput, filter: WebhookFilterInput, before: String, after: String, first: Int, last: Int): WebhookCountableConnection
   webhookEvents: [WebhookEvent]
   webhookSamplePayload(eventType: WebhookEventTypeEnum!): JSONString
+  warehouse(id: ID!): Warehouse
+  warehouses(filter: WarehouseFilterInput, before: String, after: String, first: Int, last: Int): WarehouseCountableConnection
   translations(kind: TranslatableKinds!, before: String, after: String, first: Int, last: Int): TranslatableItemConnection
   translation(id: ID!, kind: TranslatableKinds!): TranslatableItem
+  stock(id: ID!): Stock
+  stocks(filter: StockFilterInput, query: String, before: String, after: String, first: Int, last: Int): StockCountableConnection
```
maarcingebala: I see that `StockFilterInput` contains the `search` field already, so we don't need to introduce `query`, as this was the old approach.

saleor/graphql/warehouse/types.py
```diff
@@ -0,0 +1,96 @@
...
+        model = models.Stock
+        interfaces = [graphene.relay.Node]
+        only_fields = ["warehouse", "product_variant", "quantity", "quantity_allocated"]
+
+    @staticmethod
+    def resolve_stock_quantity(root, *_args):
+        return get_available_quantity_for_customer(root)
+
+    @staticmethod
+    def resolve_quantity(root, *_args):
```
maarcingebala: This one and `resolve_quantity_allocated` should be wrapped with `@permission_required(ProductPermissions.MANAGE_PRODUCTS)`. These fields had these permissions in the `ProductVariant` type, where we deprecated them. We have to maintain the same permissions as we had there. Only `stockQuantity` is meant for display in public storefronts.

PR #4987, opened 2019-11-19
Extend JWT payload

salwator:
When extending saleor with external services it is useful to reuse its JWT token for authentication everywhere, currently we can only extract email address for user which is not enough to do basic access control.

This PR adds possibility to inject additional user information into JWT token payload, starting with _is_staff_ and _is_superuser_ flags.

There might be need to provide (optional) full permission list in the future using same mechanism.

No tests added since no business logic really introduced and tests complexity would greatly outgrow production code.

### Pull Request Checklist


django-queries:
Here is the report for 2e4551f3cb69e7c95e956f2cd0c6ac682fbbbc0f (mirumee/saleor @ extend-jwt-payload)
Base comparison is f0cdaf50dfb046b06a8a1f173cf15bcf4f2c6d49.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          6	          6	              0
  create checkout                            	         50	         50	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	         13	         13	              2

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         15	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

salwator:
> Test?

No point, really. I wrote a few words about that in the PR description.
No business logic, no reason to change this code if you're not integrating something, trivial code.

maarcingebala:
I agree the code is trivial, but saying that there is no point of writing tests for any piece of code is at least controversial ;)

NyanKiyoshi:
Without test there is no proof the JWT payload is actually what we expect it to be.

salwator:
What a nice and unexpected place to have an interesting discussion, as a testing geek I can't miss that opportunity to talk! 😉 
Actually, JWT does not _need_ to do anything and we outsource it to a separate library, it is used outside of the saleor and you could/should easily test it in integration testing or just by seeing that your local dev env is working **at all**.
Obviously, it is a complex and hot topic, but I would like to kindly remind everyone that testing is **not** prooving correctness and there is nothing strange in consciously choosing not to test a piece of code, and I write this as a longterm TDD proponent!
Tests are long term investments that are subject to constant justification and balancing risk vs cost. When testing trivial _glue_ code we might generate more pain by writing, running and maintaining tests themselves then they bring to the table in terms of increased safety and coding. Nice take on it is [here](https://www.philosophicalhacker.com/post/maybe-dont-write-that-test/) and somewhere in Kent Becks book I totally forgot the title 😄

NyanKiyoshi:
What I mean is that when we upgrade the django JWT dependency from X major version to X other major version, we would have no way of telling "They dropped the setting without deprecating it, and the extension is no longer called".
What bothers me is that it is implicit. But I get it, it is fair; and we assume the third-party will make sure of not doing a major breaking change without actually deprecating first to let us know–which is fair.

PR #5000, opened 2019-11-22
Drop dashboard 1.0

### LINKED ISSUE #4463: Drop Dashboard 1.0
Soon we'll merge a PR which migrates HTML content (product descriptions/pages) to JSON (#4366) and adds support for rendering JSON in Storefront 1.0. It is probably the last issue that blocks us from making Dashboard 2.0 the default one.

There are no plans to archive Dashboard 1.0 in some other repository. If anyone would ever need it, it will remain in the history of Saleor main repository. Dashboard 2.0, on the other hand, will be moved to a separate repository, as this is technically a separate application using a different stack. It will be also easier to track issues separately for both parts.

If there is anything else that we should take care of before the migration - let's discuss it in this issue.

IKarbowiak:
Drop dashboard 1.0: views, templates and tests.

Closes #4463 

### Pull Request Checklist


django-queries:
Here is the report for 54b1337ae7a9377c8d5b2f08934a0fcf0136d90d (mirumee/saleor @ drop-dashboard-1.0)
Base comparison is aec8d5cc933b1614601115091a69b7eb5af9814d.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         14	         14	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	         50	         50	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         15	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/account/utils.py
```diff
@@ -2,6 +2,9 @@
 import random
 
 from django.conf import settings
+from django.contrib.admin.views.decorators import (
+    staff_member_required as _staff_member_required,
+)
```
NyanKiyoshi: What about naming it `django_staff_member_required` so we explicitly know it is not internal, but django's?

tests/api/test_account.py
```diff
@@ -1424,7 +1424,7 @@ def test_staff_create(
     )
 
 
-@patch("saleor.dashboard.emails._send_set_user_password_email_with_url.delay")
+@patch("saleor.core.emails._send_set_user_password_email_with_url.delay")
```
NyanKiyoshi: All the above: maybe it would be better to move those tasks to `saleor.account.emails`?

tests/test_elasticsearch.py
```diff
@@ -97,114 +95,3 @@ def test_new_search_doesnt_show_unpublished(db, products_with_mixed_publishing,
...
-    response = client.get(reverse("dashboard:search"), {"q": phrase})
-    assert phrase == response.context["query"]
-    found_prod = {p.pk for p in response.context["products"]}
-    found_users = {p.pk for p in response.context["users"]}
-    found_orders = {p.pk for p in response.context["orders"]}
-    return found_prod, found_users, found_orders
-
-
-@pytest.mark.integration
-@pytest.mark.vcr(record_mode="once", match_on=MATCH_SEARCH_REQUEST)
```
NyanKiyoshi: Don't forget to remove the cassettes in that case 👍

saleor/account/utils.py
IKarbowiak: Good idea. :)

saleor/account/utils.py
```diff
@@ -101,3 +104,20 @@ def get_random_avatar():
...
+    """
+    if staff.orders.exists():
+        staff.is_staff = False
+        staff.user_permissions.clear()
+        staff.save()
+    else:
+        staff.delete()
+
+
+def staff_member_required(f):
```
korycins: Maybe for better readability, we should introduce `account/decorators` file?

saleor/account/utils.py
```diff
@@ -101,3 +104,20 @@ def get_random_avatar():
     avatar_name = random.choice(os.listdir(AVATARS_PATH))
     avatar_path = os.path.join(AVATARS_PATH, avatar_name)
     return File(open(avatar_path, "rb"), name=avatar_name)
+
+
+def remove_staff_member(staff):
```
korycins: Shouldn't we separate staff actions from normal users? I mean, maybe we should have another file for the logic which comes from dashboard 1.0?

saleor/core/emails.py
```diff
@@ -14,3 +19,34 @@ def get_email_context():
         "site_name": site.name,
     }
     return send_email_kwargs, email_template_context
+
+
+def send_set_password_email_with_url(redirect_url, user, staff=False):
```
korycins: Shouldn't belong to the `account` module?

saleor/shipping/models.py
```diff
@@ -84,7 +84,7 @@ def __str__(self):
     def countries_display(self):
         countries = self.countries
         if self.default:
-            from ..dashboard.shipping.forms import get_available_countries
+            from .utils import get_available_countries
```
korycins: Are you sure that circular imports still exist? Maybe we could move this import at the beginning of the file?

saleor/core/emails.py
IKarbowiak: Yes, I will move that.

saleor/shipping/models.py
IKarbowiak: I checked, circular imports still exists after moving this import to beginning of the file. `saleor/shipping/utils` imports `ShippingZone` from `saleor/shipping/models.py`. We can drop this import by moving `currently_used_countries` and `get_available_countries` functions from `utils.py` to `models.py` in shipping. But I am not sure if it's a good solution.

tests/test_elasticsearch.py
IKarbowiak: Done 👍

saleor/shipping/models.py
NyanKiyoshi: The functions referencing the `ShippingZone` and being used by `ShippingZone` are query-set methods/utilities. They should definitively be inside the `QuerySet` manager of `ShippingZone`

tests/api/test_account.py
IKarbowiak: Done 👍

saleor/shipping/models.py
IKarbowiak: But `currently_used_countries` and `get_available_countries` operate on all existing `ShippingZone` objects, not on specific (e.g. filtered) queryset. So I'm not sure if putting this functions in `QuerySet` manager is a good solution.

saleor/account/utils.py
IKarbowiak: Probably it is, but it's used only in Django views. As far as I know, we will be removing all views soon. So I'm not sure if we care. 🤔 
What do you think @maarcingebala, @NyanKiyoshi?

maarcingebala: I wouldn't bother with that, soon it won't be needed at all.

saleor/account/utils.py
IKarbowiak: It seems that everything fit somewhere except for this staff functions (and one of them: `staff_member_required` will be removed soon). So we can create separate file for this stuff actions, but it will be for one function. Also user and staff user have a lot in common (only `is_staff` field differ them), actions applicable to user are also applicable to staff user, so I am not sure if this distinction is needed.

saleor/account/utils.py
korycins: So if this will be not needed soon then maybe it is a good moment to put such logic in one place? Dropping the whole file of old methods/functions will be faster (and I assume safer) than checking each function one by one. We had a similar solution for old mutations and queries. We just moved them to the `deprecated_` files. It was easier to remove deprecated mutations/queries when we had them in one place, instead of searching in different files
@maarcingebala @IKarbowiak ?

saleor/shipping/models.py
NyanKiyoshi: Yes, but it is returning a query-set. Having all query-set filtering is better in a single same place (manager). Just like we do for `visible_to_user` (there is `get_visible_to_user` too looks like 😕).

Personally, when working on a django project, I prefer finding all the filtering operations in a single same place. But we don't have any convention for that currently... We are doing both query-set and `utils.py`.

Maybe @maarcingebala or @korycins have an input about it?

krzysztofwolski: Do we even need this method anymore? It was used only in old dashboard
Screenshot 2019-11-25 at 15 34 12

krzysztofwolski: Do we need this method anymore? It seems to be used only in dashboard we want to drop.
Screenshot 2019-11-25 at 15 34 12

saleor/order/utils.py
```diff
@@ -321,3 +322,32 @@ def get_valid_shipping_methods_for_order(order: Order):
     return ShippingMethod.objects.applicable_shipping_methods_for_instance(
         order, price=order.get_subtotal().gross
     )
+
+
+def get_products_voucher_discount_for_order(voucher):
```
krzysztofwolski: Missing types

saleor/shipping/utils.py
```diff
@@ -0,0 +1,20 @@
+from ..account.i18n import COUNTRY_CHOICES
+from .models import ShippingZone
+
+
+def currently_used_countries(zone_pk=None):
```
krzysztofwolski: Missing types and docstrings

saleor/shipping/models.py
IKarbowiak: Oh, good point! Then `countries_display` could be removed, also `get_available_countries` and `currently_used_countries`will be not needed anymore. So I will remove those and there will be no problem anymore in this case. 😌

saleor/order/utils.py
IKarbowiak: Done 👍

saleor/shipping/utils.py
IKarbowiak: I removed that function.

saleor/account/utils.py
maarcingebala: I would first merge this PR, then drop Storefront 1.0 and we'll see what needs to be moved to a seaprate file.

CHANGELOG.md
```diff
@@ -18,8 +18,10 @@ All notable, unreleased changes to this project will be documented in this file.
 - Now force pytest to ignore the environment variable containing the django settings module - #4992 by @NyanKiyoshi
 - Extend JWT token payload with user information - #4987 by @salwator
 - Optimize the queries for product list in the dashboard - #4995 by @IKarbowiak
+- Drop dashboard 1.0 - #5000 by @IKarbowiak
 - Fixed serialization error on weight fields when running `loaddata` and `dumpdb` - #5005 by @NyanKiyoshi
 - Fixed JSON encoding error on Google Analytics reporting - #5004 by @NyanKiyoshi
+- Drop dashboard 1.0 - #5000 by @IKarbowiak
```
krzysztofwolski: Rebase issue

mimidotsuser:
the [docs](https://docs.saleor.io/docs/getting-started/creating-superuser/) should also be update to reflect the updates

PR #5021, opened 2019-11-28
Feature/wishlist

derenio:
This PR adds the GraphQL's API for customers' `wishlists`.

1. [ ] ~Privileged views and APIs are guarded by proper permission checks.~ views use only logged-in users data.
1. [ ] ~All visible strings are translated with proper context.~
1. [ ] ~All data-formatting is locale-aware (dates, numbers, and so on).~


django-queries:
Here is the report for 4a683adddd1e405ebf91efbc4ce8f53e971dd4f2 (derenio/saleor @ feature/wishlist)
Base comparison is 5f1437b91400f033ef23782fa919487c44f79093.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	         50	         50	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         15	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/graphql/wishlist/mutations.py
```diff
@@ -0,0 +1,59 @@
...
+            info, variant_id, only_type=ProductVariant, field="variant_id"
+        )
+        wishlist.add_variant(variant)
+        wishlist_items = wishlist.items.all()
+        return WishlistAddProductVariantMutation(wishlist_items=wishlist_items)
+
+
+class WishlistRemoveProductVariantMutation(BaseMutation):
+    wishlist_items = graphene.List(
+        WishlistItem, description="The list of WIshlistItems of the current user."
```
krzysztofwolski:
```suggestion
        WishlistItem, description="The list of WishlistItems of the current user."
```

krzysztofwolski:
@derenio Proposition: create unique together on wishlist item fields: (token, variant) This way we will avoid duplicates and handling logic for those. WDYT?

saleor/wishlist/models.py
```diff
@@ -0,0 +1,75 @@
...
+            return
+        else:
+            item.variants.remove(variant)
+            # If it was the last variant, delete the whole item
+            if item.variants.count() == 0:
+                item.delete()
+
+
+class WishlistItemQuerySet(models.QuerySet):
+    def move_items_between_wishlists(self, src_wishlist, dst_wishlist):
```
krzysztofwolski: Possible issues with handling `WishlistItem  unique_together `.

saleor/graphql/wishlist/resolvers.py
```diff
@@ -0,0 +1,20 @@
+from typing import List
+
+from ...wishlist.models import Wishlist, WishlistItem
+from ...wishlist.session_helpers import WishlistSessionHelper
+
+
+def get_wishlist_from_info(info) -> Wishlist:
```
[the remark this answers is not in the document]
derenio: @krzysztofwolski do you know from where to import type of the `info` object?

saleor/wishlist/models.py
```diff
@@ -0,0 +1,75 @@
...
+        self.user = user
+        self.save()
+
+    def get_all_variants(self):
+        return ProductVariant.objects.filter(
+            wishlist_items__wishlist_id=self.pk
+        ).distinct()
+
+    def add_variant(self, variant: ProductVariant):
+        item, is_created = self.items.get_or_create(product_id=variant.product_id)
```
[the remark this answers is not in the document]
derenio: Good catch, it will. Though `add_variant` is resistant to these cases as `get_or_create` ensures we don't create a duplicate `wishlist, product` instance and `item.variants.add` already handles duplicates (it doesn't make them).

saleor/graphql/wishlist/resolvers.py
```diff
@@ -0,0 +1,20 @@
...
+    user = info.context.user
+    if user.is_anonymous:
```
maarcingebala: Wishlists will be basically implemented client-side and only when you log in we will synchronize them with the backend, just as we do with carts/checkouts. One reason for that is that we want to support wishlist in the offline mode which client-side implementation allows.

saleor/graphql/wishlist/resolvers.py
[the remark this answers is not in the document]
krzysztofwolski: Seems its ResolveInfo objects: https://github.com/graphql-python/graphql-core/blob/master/graphql/execution/base.py

saleor/wishlist/error_codes.py
```diff
@@ -0,0 +1,9 @@
+from enum import Enum
+
+
+class WishlistErrorCode(Enum):
```
krzysztofwolski: According to our Notion docs: 
```suggestion
class WishlistErrorCode(str, Enum):
```

saleor/wishlist/models.py
```diff
@@ -0,0 +1,92 @@
...
+from django.db import models, transaction
+
+from ..account.models import User
+from ..product.models import Product, ProductVariant
+
+
+class WishlistQuerySet(models.QuerySet):
+    def get_or_create_wishlist_for_user(self, user):
+        try:
+            return user.wishlist
```
korycins: Why not - `get_or_create`? https://docs.djangoproject.com/en/3.0/ref/models/querysets/

tests/conftest.py
```diff
@@ -632,7 +634,47 @@ def product(product_type, category):
...
+    )
+
+    for i in (1, 2):
+        ProductVariant.objects.create(
+            product=product,
+            sku=f"prodVar{i}",
+            cost_price=Money("1.00", "USD"),
+            quantity=10,
+            quantity_allocated=1,
+        )
```
korycins: We should keep the number of DB queries as small as possible. Especially for fixtures.
https://docs.djangoproject.com/en/2.2/ref/models/querysets/#bulk-create

korycins:
Mr. @krzysztofwolski Can you review it again?

saleor/graphql/wishlist/schema.py
```diff
@@ -0,0 +1,24 @@
...
+    WishlistRemoveProductVariantMutation,
+)
+
+
+# User's wishlist queries are located in the "saleor.graphql.account" module:
+#
+#     me {
+#         wishlist
+#     }
+class WishlistQueries(graphene.ObjectType):
```
korycins: Do we need to store this class? I think that the comment contains enough info.

saleor/graphql/account/types.py
```diff
@@ -17,6 +17,8 @@
 from ..core.utils import get_node_optimized
 from ..decorators import one_of_permissions_required
 from ..utils import format_permissions_for_display
+from ..wishlist.resolvers import resolve_wishlist_items_from_user
```
korycins: I am not confident that mixin resolvers between modules is the best idea. Maybe it should be places in `account.resolvers`?

korycins:
@derenio let's solve conflicts and we should be ready to finish it today
@maarcingebala Do you have any suggestions?

saleor/graphql/account/types.py
derenio: The resolver contains logic specific to the wishlist. I think it's ok to keep it separate, even though it's exposed under the `me {...}` endpoint in the `account` module.

saleor/graphql/wishlist/schema.py
derenio: I kept it in the case we want to add some general wishlist specific queries (e.g. from top of my head, how many likes does `product XYZ` has). Although if we will need it we can add it, so removing :+1: .

PR #5035, opened 2019-12-04
Use calculate methods from calculation wrapper method

kswiatek92:
Vatlayer is missing calculate subtotal method
which raise an error
```
TypeError: unsupported operand type(s) for +: 'NotImplementedType' and 'TaxedMoney'
```
used `checkout_subtotal` method from calculation wrapper to fix this issue

### Pull Request 

### DIFF
--- a/saleor/extensions/plugins/vatlayer/plugin.py
+++ b/saleor/extensions/plugins/vatlayer/plugin.py
@@ -7,6 +7,8 @@ from django_countries.fields import Country
 from django_prices_vatlayer.utils import get_tax_rate_types
 from prices import Money, MoneyRange, TaxedMoney, TaxedMoneyRange
 
+
+from ....checkout import calculations
 from ....core.taxes import TaxType
 from ....graphql.core.utils.error_codes import ExtensionsErrorCode
 from ...base_plugin import BasePlugin
@@ -72,12 +74,9 @@ class VatlayerPlugin(BasePlugin):
         if self._skip_plugin(previous_value):
             return previous_value
 
-        zero_total = Money(0, currency=previous_value.currency)
-        taxed_zero = TaxedMoney(zero_total, zero_total)
-
         return (
-            self.calculate_checkout_subtotal(checkout, discounts, previous_value)
-            + self.calculate_checkout_shipping(checkout, discounts, taxed_zero)
+            calculations.checkout_subtotal(checkout, discounts)
+            + calculations.checkout_shipping_price(checkout, discounts)
             - checkout.discount
         )
 
--- a/saleor/graphql/checkout/types.py
+++ b/saleor/graphql/checkout/types.py
@@ -1,7 +1,7 @@
 import graphene
 import graphene_django_optimizer as gql_optimizer
 
-from ...checkout import models
+from ...checkout import calculations, models
 from ...checkout.utils import get_valid_shipping_methods_for_checkout
 from ...core.taxes import display_gross_prices, zero_taxed_money
 from ...extensions.manager import get_extensions_manager
@@ -146,22 +146,20 @@ class Checkout(MetadataObjectType, CountableDjangoObjectType):
     @staticmethod
     def resolve_total_price(root: models.Checkout, info):
         taxed_total = (
-            info.context.extensions.calculate_checkout_total(
-                checkout=root, discounts=info.context.discounts
-            )
+            calculations.checkout_total(checkout=root, discounts=info.context.discounts)
             - root.get_total_gift_cards_balance()
         )
         return max(taxed_total, zero_taxed_money())
 
     @staticmethod
     def resolve_subtotal_price(root: models.Checkout, info):
-        return info.context.extensions.calculate_checkout_subtotal(
+        return calculations.checkout_subtotal(
             checkout=root, discounts=info.context.discounts
         )
 
     @staticmethod
     def resolve_shipping_price(root: models.Checkout, info):
-        return info.context.extensions.calculate_checkout_shipping(
+        return calculations.checkout_shipping_price(
             checkout=root, discounts=info.context.discounts
         )
 
--- a/tests/extensions/plugins/test_vatlayer.py
+++ b/tests/extensions/plugins/test_vatlayer.py
@@ -8,6 +8,7 @@ from django.urls import reverse
 from django_countries.fields import Country
 from prices import Money, MoneyRange, TaxedMoney, TaxedMoneyRange
 
+from saleor.checkout import calculations
 from saleor.checkout.utils import add_variant_to_checkout
 from saleor.core.taxes import quantize_price
 from saleor.extensions.manager import get_extensions_manager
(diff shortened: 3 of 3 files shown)


django-queries:
Here is the report for 4fd4fcab6dd9a8484dc8e0b90f0a93352cd42ad4 (kswiatek92/saleor @ fix-error-with-missing-vatlayer-calculate-subtotal)
Base comparison is 81383c01022666b48a280982f27cf7e039272930.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	         50	         50	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         15	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/checkout/calculations.py
```diff
@@ -1,35 +1,44 @@
...
+    checkout: "Checkout",
+    discounts: "DiscountsListType" = None,
+    extensions: Optional["ExtensionsManager"] = None,
 ) -> "TaxedMoney":
     """Return checkout shipping price.
 
     It takes in account all extensions.
     """
-    return get_extensions_manager().calculate_checkout_shipping(checkout, discounts)
+    extensions = extensions if extensions else get_extensions_manager()
```
maarcingebala:
```suggestion
    extensions = extensions or get_extensions_manager()
```

saleor/checkout/calculations.py
```diff
@@ -1,35 +1,44 @@
...
     from .models import Checkout, CheckoutLine
     from ..discount.types import DiscountsListType
+    from ..extensions.manager import ExtensionsManager
 
 
 def checkout_shipping_price(
-    checkout: "Checkout", discounts: "DiscountsListType" = None
+    checkout: "Checkout",
+    discounts: "DiscountsListType" = None,
+    extensions: Optional["ExtensionsManager"] = None,
```
korycins: I am not sure about that. What was the reason to add it to all `checkout_*` methods? I don't see any benefits of using this, only additional `if`/`or`

saleor/checkout/calculations.py
kswiatek92: fixed

saleor/checkout/calculations.py
kswiatek92: correct me if I'm wrong but
resolvers store extension in `info` arg,
this way we just call "get_extension" once not every time resolver is called

korycins: @kswiatek92  See `saleor.core.middleware.extensions`. Manager is as appened as a `SimpleLazyObject` to `info.context.extensions`. So it doesn't give you any benefit to pass it as an optional argument. 
You will get the same number of `function calls` in both cases.

saleor/extensions/plugins/vatlayer/plugin.py
```diff
@@ -72,12 +74,9 @@ def calculate_checkout_total(
         if self._skip_plugin(previous_value):
             return previous_value
 
-        zero_total = Money(0, currency=previous_value.currency)
-        taxed_zero = TaxedMoney(zero_total, zero_total)
-
         return (
-            self.calculate_checkout_subtotal(checkout, discounts, previous_value)
-            + self.calculate_checkout_shipping(checkout, discounts, taxed_zero)
+            calculations.checkout_subtotal(checkout, discounts)
```
korycins: Also - I am not sure of this. @krzysztofwolski Can you help?

korycins: AFAIK `checkout .calculation` module is the `top` level of calculation.  So I am not sure that we should trigger it inside a plugin, which should respect hierachy `calculation` -> `manager` -> `plugin`

saleor/checkout/calculations.py
kswiatek92: in that case, your'e right I will revert this to previous version

kswiatek92: done

saleor/extensions/plugins/vatlayer/plugin.py
kswiatek92: I consulted this with @krzysztofwolski 
For what I understand we want to run `checkout_subtotal` for every plugin
and this should do it.
But we can discuss this if we want something different, or if we should use different approach

korycins: Maybe I am wrong but let's take as an example `checkout_shipping_price`.
All Saleor logic uses `calculation.checkout_shipping_price`, this calls manager and manager calls each plugin one by one.
The same flow is for the rest methods. Also, Avatax uses the same flow.

korycins: In your case we have: `calculation.checkout_subtotal` -> `manager` -> plugin one by one -> `Vatlayer` -> `calculation.checkout_subtotal`

korycins: Ok, my bad.

saleor/graphql/checkout/types.py
```diff
@@ -146,22 +146,20 @@ def resolve_email(root: models.Checkout, info):
...
-            )
+            calculations.checkout_total(checkout=root, discounts=info.context.discounts)
             - root.get_total_gift_cards_balance()
         )
         return max(taxed_total, zero_taxed_money())
 
     @staticmethod
     def resolve_subtotal_price(root: models.Checkout, info):
-        return info.context.extensions.calculate_checkout_subtotal(
+        return calculations.checkout_subtotal(
```
krzysztofwolski: Use ext. mgr. from context

saleor/graphql/checkout/types.py
```diff
@@ -146,22 +146,20 @@ def resolve_email(root: models.Checkout, info):
...
             checkout=root, discounts=info.context.discounts
         )
 
     @staticmethod
     def resolve_shipping_price(root: models.Checkout, info):
-        return info.context.extensions.calculate_checkout_shipping(
+        return calculations.checkout_shipping_price(
```
krzysztofwolski: Use ext. mgr. from context (pass it as `checkout_shipping_price` argument)

korycins: Do you mean optional argument `extension=None` for `checkout_shipping_price`?

kswiatek92:
> We discussed passing extensions - it can stay the way it is.
> 
> Before merging there needs to be created regression tests. Also, please check the second plugin for tax calculation is such issues does not happen there.

second plugin should be ok,
tests are on the way

PR #5126, opened 2020-01-02
Account confirmation by email when creating new account

### LINKED ISSUE #1527: Newly registered user e-mail confirmation
I see that according to previous issues and pull requests, that e-mail address confirmation for new users was a feature in the past but currently is not. I would like to add it to my site, I can work on it as a pull request for Saleor (mostly taken from the django-allauth implementation, it would be enabled/disabled in settings.py) but first I wanted to check if it's something that would likely be added to Saleor. Also wanted to ask if there is already an easy way to enable it that perhaps I'm missing?

...Users need to confirm their e-mail address before logging in

1.User registers
2.User is immediately logged in, no activation e-mail

### What I expected to happen

...User needs to confirm e-mail address via e-mail activation link

### What happened instead/how it failed

...User is immediately allowed to login

tomaszszymanski129:
I want to merge this change because it resolves #1527 
Cannot be merged until appropriate frontend view is ready (confirmation part that triggers the mutation)
Should also be merged with storefront mutation switch (which can be merged even sooner)
https://github.com/mirumee/saleor-storefront/pull/549

### Pull Request Checklist


django-queries:
Here is the report for 106fe83d89535597adb46f9eada591e548afb740 (mirumee/saleor @ 1527/account_confirmation)
Base comparison is f22ee176016250156cf1ed392b8e3f3b76e2a669.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	         50	         50	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         15	         15	              3
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         15	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/graphql/account/mutations/base.py
```diff
@@ -153,6 +153,44 @@ def perform_mutation(cls, _root, info, **data):
...
+                }
+            )
+
+        if not default_token_generator.check_token(user, data["token"]):
+            raise ValidationError(
+                {"token": ValidationError(INVALID_TOKEN, code=AccountErrorCode.INVALID)}
+            )
+
+        user.is_active = True
+        user.save()
```
NyanKiyoshi:
```suggestion
        user.save(update_fields=["is_active"])
```

tests/api/test_account.py
```diff
@@ -2115,6 +2157,41 @@ def test_account_reset_password(
...
+    }
+    user_api_client.post_graphql(CONFIRM_ACCOUNT_MUTATION, variables)
+
+    customer_user.refresh_from_db()
+    assert customer_user.is_active is True
+
+
+def test_account_confirmation_invalid_user(user_api_client, customer_user):
+    variables = {
+        "email": "non-existing@nope.com",
```
NyanKiyoshi: RFC
```suggestion
        "email": "non-existing@example.com",
```

saleor/settings.py
```diff
@@ -128,6 +128,8 @@ def get_bool_from_env(name, default_value):
 EMAIL_USE_TLS = email_config["EMAIL_USE_TLS"]
 EMAIL_USE_SSL = email_config["EMAIL_USE_SSL"]
 
+ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL = True
```
NyanKiyoshi: It should take the value from env and shouldn't be true by default, otherwise it could be a breaking or unwanted change

NyanKiyoshi: cc @maarcingebala that could be up for discussion whether it should be `True` or `False` out of the box. Or maybe later on.

tomaszszymanski129: @NyanKiyoshi Discussed with Marcin, we think it should be enabled by default as it's the default for most of the internet.

saleor/graphql/account/mutations/account.py
```diff
@@ -28,6 +29,12 @@
 class AccountRegisterInput(graphene.InputObjectType):
     email = graphene.String(description="The email address of the user.", required=True)
     password = graphene.String(description="Password.", required=True)
+    redirect_origin = graphene.String(
```
maarcingebala: In other mutations where this pattern is used this argument is called `redirectUrl` - for consistency, I would use it here as well.

saleor/graphql/account/mutations/account.py
```diff
@@ -36,18 +43,46 @@ class Arguments:
             description="Fields required to create a user.", required=True
         )
 
+    requires_confirmation = graphene.Boolean()
```
maarcingebala: Let's add some description here as well e.g. "Informs whether users need to confirm their email address."

saleor/settings.py
```diff
@@ -541,3 +548,14 @@ def get_host():
 JWT_TOKEN_SECRET = os.environ.get("JWT_TOKEN_SECRET", "saleor")
 if not DEBUG:
     JWT_VERIFY_EXPIRATION = True
+
+
+if (
+    not DEBUG
```
maarcingebala: Not if this is needed, this warning might be useful with `DEBUG=True` as well.

NyanKiyoshi: Maybe we shouldn't raise but only warn then?

tomaszszymanski129: Feature works out of the box for development since `localhost` and `127.0.0.1` are default for `ALLOWED_CLIENT_HOSTS` (but removing debug line would raise the exception for them). Assuming no production environment is running with DEBUG on, I think it's preferred to leave it as is, it's better to have an exception about this rather than warning in production env (and if we remove the debug it would have to be changed to warning).

saleor/graphql/account/mutations/account.py
```diff
@@ -36,18 +43,48 @@ class Arguments:
             description="Fields required to create a user.", required=True
         )
 
+    requires_confirmation = graphene.Boolean(
+        description="Informs whether users need to confirm their email address."
```
maarcingebala: Let's add `required=True` since this field will always return some boolean value and never a `null`.

PR #5149, opened 2020-01-08
Fast pagination without OFFSET, skip counting rows unless requested

### LINKED ISSUE #2377: Pagination: better backend implementation
Currently we're exposing a cursor-based API (which is great) but internally we still cast it to page numbers which result in an `OFFSET` + `LIMIT` database query.

Unfortunately using `OFFSET` means the database has to fetch `OFFSET` + `LIMIT` rows and then discard the first `OFFSET` results before sending the remaining ones to the client. This means the cost of each page increases linearly with the page number and there's risk of temporary disk tables being used for pages where the number of required rows exceeds the available results buffer.

It's slow when using pagination in your own shop but can become a denial-of-service attack vector for deployments under software-as-a-service model where a malicious agent could intentionally put stress on the database server.

I propose that we cease to use the connection fields provided by `graphene-django` and instead write a resolver that takes care of both filtering and pagination and embraces the cursor based approach (where the resulting query should only utilize `WHERE` and `LIMIT` without an `OFFSET` clause). This way database indexes can be used and the workload related to fetching any particular page is independent of the page number.

In addition we would be free to use data loaders which are currently incompatible with the `DjangoFilterConnectionField`. This is a major performance win.

I've considered extending the `DjangoConnectionField` but as the query is sliced inside the field's implementation we would likely need a separate class for each model. I think just writing a resolver for each case and maintaining a common structure for results is much easier.

### LINKED ISSUE #4652: Upgrade JWT middleware from 0.3.0 to 0.3.2
The JWT middleware is no longer a django middleware, thus, upgrading to this new "minor" version is breaking the middlewares. *More* details here https://github.com/flavors/django-graphql-jwt/pull/125.

For reference about why those changes have happened: https://docs.djangoproject.com/en/2.2/topics/http/middleware/

patrys:
Proof of concept implementation that limits the cost of paginated connections unless `totalCount` is explicitly requested.

⚠️ WARNING

* Not fully tested.
* Not sure if it's used by all connection classes.
* I likely don't have the time necessary to bring this into a mergeable state.

---
Fixes #4652
Fixes #2377

### Pull Request Checklist


patrys:
CC @NyanKiyoshi

django-queries:
Here is the report for f45a9f0200e077836638ee618227176cc805c563 (patrys/saleor @ skip-count-unless-necessary)
Base comparison is b342d0184e9dc4c043a310326dfed875acec045f.



<details><summary>**Found 1 differences!** (click me)</summary>
<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	         50	         50	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         14	         14	              2
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
+ retrieve variant list                      	         16	         15	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

requirements.txt
```diff
@@ -1,113 +1,496 @@
...
-tinycss2==1.0.2
-tqdm==4.40.2
-typing==3.7.4.1
-urllib3==1.25.7
-uwsgi==2.0.18; sys_platform != "win32"
-vine==1.3.0
-weasyprint==50
-webencodings==0.5.1
+amqp==2.5.2 \
+    --hash=sha256:6e649ca13a7df3faacdc8bbb280aa9a6602d22fd9d545336077e573a1f4ff3b8 \
```
maarcingebala: We don't want these hashes: https://docs.getsaleor.com/docs/next/advanced/managing-dependencies/#other-tools

patrys: Why? It's an important security feature.

maarcingebala: For some reason, it was breaking our deployment on Heroku. If I remember correctly the problem was that hashes were missing for some requirements and pip failed to install such requirements file.

patrys: Maybe it's no longer an issue as it works fine with `pip install`. By the way, maybe we should switch our Heroku support to use Docker? https://devcenter.heroku.com/articles/build-docker-images-heroku-yml

maarcingebala: @IKarbowiak Do you know exactly why it wasn't working? I think it was your PR that removed hashes recently.

IKarbowiak: Yes, it was my PR, but I worked with @fowczarek (because I don't have account on Heroku, and he has one) and he was checking it on Heroku. We saw that poeple writing that it helps. Also I saw that @NyanKiyoshi opened PR to remove hashes, so maybe he can also help here.

NyanKiyoshi: In my PR I added details on what went wrong:
> Noticed it was not possible to install requirements with some machines as some architecture have wheels available--poetry does not detect the wheel requirements at this time.

But I also saw that somebody said in saleor-docs to use Python 3.8 instead of 3.7. I am pretty sure I was using 3.7, so I would like to try that even though I doubt that–maybe there is no wheel available for 3.8, so it doesn't use them.

For reference, I was installing on the kernel image `4.14.152-127.182.amzn2.x86_64` using `pip install -r requirements.txt` and latest `pip` version that was available.

Once again, 99% sure I was using Python 3.7.4.

pyproject.toml
```diff
@@ -24,7 +24,7 @@ django = "~2.2"
 django-babel = "^0"
 django-countries = "^5.3"
 django-filter = "^2.2"
-django-graphql-jwt = "<=0.2.1" # https://github.com/mirumee/saleor/issues/4652
+django-graphql-jwt = "^0.3.0" # https://github.com/mirumee/saleor/issues/4652
```
NyanKiyoshi: Obviously, the comment is out of date from that change
```suggestion
django-graphql-jwt = "^0.3.0"
```

requirements.txt
maarcingebala: We've been using Python 3.8.1 on Heroku for some time already and it was failing with this version as well. We were getting this error:

```
Collecting setuptools>=39.2.0 (from cairocffi==1.1.0->-r /tmp/build_2886b32ba14cbeeee7a9d26444bc1fea/requirements.txt (line 34))
       In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
           setuptools>=39.2.0 from https://files.pythonhosted.org/packages/f9/d3/955738b20d3832dfa3cd3d9b07e29a8162edb480bf988332f5e6e48ca444/setuptools-44.0.0-py2.py3-none-any.whl#sha256=180081a244d0888b0065e18206950d603f6550721bd6f8c0a10221ed467dd78e (from cairocffi==1.1.0->-r /tmp/build_2886b32ba14cbeeee7a9d26444bc1fea/requirements.txt (line 34))
 !     Push rejected, failed to compile Python app.
```

maarcingebala: Also, `setuptools` is not specified in our requirements files. Others seem to have this issue as well: https://github.com/python-poetry/poetry/issues/1584

patrys: Wait, why are we running with `--require-hashes`?

maarcingebala: I can't see `--require-hashes` anywhere in our code, not sure where it comes from.

NyanKiyoshi: It comes by default when you have hashes in the file. There is no flag to tell it to act differently from the requirements of wheels. I believe the issue is that poetry doesn't/is not able to detect the requirements from wheels thus generate the requirement file with hashes for it as the wheels can differ from platform to platform.

patrys: Why do you think it's related to wheels? The only issue I've seen in this thread is related to missing hashed for the three blacklisted packages (setuptools, pip, distribute).

NyanKiyoshi: I recall having troubles with other dependencies than setuptools. But I'm not sure anymore, I would need to check if that was the only failure.

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,118 @@
...
+
+    cursor_after = graphene.Node.from_global_id(after) if after else None
+    cursor_before = graphene.Node.from_global_id(before) if before else None
+    cursor_model = qs.model
+    cursor = cursor_after or cursor_before
+    cursor_offset = 1 if cursor else 0
+
+    sort_by = args.get("sort_by", {})
+    sort_by_field = sort_by.get("field")
+    sort_by_direction = sort_by.get("direction", "")
```
patrys: I think these should be an enum and a boolean in the input type. I like Shopify's simple interface:

```graphql
{
  products(first: 50, sortKey: CREATED_AT, reverse: true) {
    ...
  }
}
```

patrys: Giving it a second thought I think the mutation should take care of mapping from `sortKey` to `sort_field` (available choices will depend on the collection being paginated) so this function could accept `sort_field: str, reverse: bool`.

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,118 @@
...
+        raise ValueError("Argument 'last' cannot be combined with 'first'.")
+    if isinstance(first, int) and args.get("before"):
+        raise ValueError("Argument 'first' cannot be combined with 'before'.")
+    if isinstance(last, int) and args.get("after"):
+        raise ValueError("Argument 'last' cannot be combined with 'after'.")
+    if isinstance(first, int) and first < 0 or first and not isinstance(first, int):
+        raise ValueError("Argument 'first' must be a non-negative integer.")
+    if isinstance(last, int) and last < 0 or last and not isinstance(last, int):
+        raise ValueError("Argument 'last' must be a non-negative integer.")
+
```
IKarbowiak: I think that `connection_args_validation` could be simplified. Firstly we should check if `last` and `first` are non-negative integer and then `isinstance()` check is not needed anymore. Like:
```suggestion
    for arg in ["first", "last"]:
        arg_value = args.get(arg)
        if isinstance(arg_value, int) and arg_value < 0 or arg_value and not isinstance(arg_value, int):
            raise ValueError(f"Argument '{arg}' must be a non-negative integer.")
    for arg1, arg2 in [("first", "last"), ("first", "before"), ("last", "after")]:
        if args.get(arg1) and args.get(arg2):
            raise ValueError(f"Argument '{arg1}' cannot be combined with '{arg2}'.")
```

saleor/graphql/core/connection.py
maarcingebala: On `master` we have `products(sortBy: {direction: ASC, field: NAME})`. Does this PR change it or you don't want to have it inside the `sortBy` input object @patrys ?

patrys: Disregard my comment then. It does look like passing any direction would result in the order being reversed though.

requirements.txt
maarcingebala: @kswiatek92 Please see this comment, maybe it will help: https://github.com/mirumee/saleor/pull/5109#issuecomment-575041918

saleor/graphql/utils.py
```diff
@@ -134,15 +134,15 @@ def sort_queryset(
 
     """
     if sort_by is None or not sort_by.field:
+        print("xaxaxa")
```
IKarbowiak: to delete

saleor/product/models.py
```diff
@@ -228,18 +229,25 @@ def sort_by_attribute(self, attribute_pk: Union[int, str], ascending: bool = Tru
...
+                output_field=CharField(),
+            )
         )
 
         # Sort by concatenated_values_order then
         # Sort each group of products (0, 1, 2, ...) per attribute values
         # Sort each group of products by name,
         # if they have the same values or not values
-        qs = qs.order_by("concatenated_values_order", "concatenated_values", "name")
+        # qs = qs.order_by("concatenated_values_order", "concatenated_values", "name")
```
IKarbowiak: remove comment

kswiatek92: done

saleor/graphql/utils.py
kswiatek92: done

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,168 @@
...
+    last = args.get("last")
+    connection_args_validation(args)
+
+    cursor_after = from_global_cursor(after) if after else None
+    cursor_before = from_global_cursor(before) if before else None
+    cursor = cursor_after or cursor_before
+    cursor_offset = 1 if cursor else 0
+
+    sort_by = args.get("sort_by", {})
+    sorting_direction = sort_by.get("direction", "")
```
patrys: Shouldn't this default to `OrderDirection.ASC`?

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,168 @@
...
+    if query_model:
+        model_fields = [field.name for field in query_model._meta.get_fields()]
+        default_ordering = query_model._meta.ordering
+        default_ordering = list(default_ordering) if default_ordering else []
+    else:
+        model_fields = default_ordering = []
+
+    if "sorting_value" in query_annotations:
+        order_expression = [f"{sorting_direction}sorting_value"] + default_ordering
+    elif sorting_field in query_annotations or sorting_field in model_fields:
```
patrys: Why do we need this case? When is `sorting_value` not possible to have?

kswiatek92: I don't want to force everyone to add `.annotate(sorting_value=F("pk"))` everywhere
if you don't need to sort you result don't add it
every model should have default ordering set in meta class
this way we only need `sorting_value` in custom sort

saleor/graphql/core/connection.py
kswiatek92: I need `"-"` and `""` in order_by that's why I did it this way

patrys: What if I pass `OrderDirection.ASC` as `direction`?

saleor/graphql/core/connection.py
patrys: I would agree if the sorting was on PK but this code allows any field to be used as the sorting field without annotating the queryset. I don't think this function should care about `field` at all. This would save us from having to construct keyword arguments with `f""` expressions which I find very hard to read.

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,168 @@
...
+    query_model = qs.query.model
+    if query_model:
+        model_fields = [field.name for field in query_model._meta.get_fields()]
+        default_ordering = query_model._meta.ordering
+        default_ordering = list(default_ordering) if default_ordering else []
+    else:
+        model_fields = default_ordering = []
+
+    if "sorting_value" in query_annotations:
+        order_expression = [f"{sorting_direction}sorting_value"] + default_ordering
```
patrys:
```python
if sorting_order == OrderDirection.ASC:
    order_expression = ["sorting_value", "pk"]
else:
    order_expression = ["-sorting_value", "-pk"]
```

patrys: Using `model._meta.ordering` makes this complex and we only need a single column to break ties.

saleor/graphql/core/connection.py
kswiatek92: If you pass `OrderDirection` it will be parsed as `""` or `"-"` so I end up having string

saleor/graphql/core/types/sort_input.py
```diff
@@ -10,10 +15,18 @@ class SortInputObjectType(graphene.InputObjectType):
...
+    class Meta:
+        abstract = True
+
     @classmethod
     def __init_subclass_with_meta__(
         cls, container=None, _meta=None, sort_enum=None, type_name=None, **options
     ):
+        if not _meta:
+            _meta = SortInputMeta(cls)
+        if sort_enum and type_name:
```
kswiatek92: refactor in progress

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,168 @@
...
+    filter_kwargs = Q()
+    if cursor and sorting_fields:
+        for field_id in range(len(sorting_fields)):
+            field_expression = {}
+            for val_id in range(len(cursor[:field_id])):
+                field_expression[sorting_fields[val_id]] = cursor[val_id]
+            field_expression[
+                f"{sorting_fields[field_id]}__{sorting_direction_kw}"
+            ] = cursor[field_id]
+            filter_kwargs |= Q(**field_expression)
```
patrys: Not sure this is correct, for sorting fields `["foo", "bar", "baz"]` and cursor values `[1, 2, 3]` we should filter by:

```python
Q(foo__gt=1) | Q(foo=1, bar__gt=2) | Q(foo=1, bar=2, baz__gt=3)
```

kswiatek92: that's what i'm doing
first have for loop for every index in sorting_fields list
then we take index we currently have and use it to slice sorting_fields (i.e. index 1 will slice sorting_fields so it will be [], index 2 will slice sorting_fields so it will be [field_1] etc.) and we run through this list and create `Q(foo=1),  Q(foo=1, bar=2)` etc
after that we append comparing condition with last field from loop
and we repeat this for every field in sorting_fields list

I know it's hard to read, that's why I will try to simplify this, or move this to separate method

patrys: I think some comment blocks would also help a lot 😄

maarcingebala: Agree, they would help a lot.

setup.cfg
```diff
@@ -38,7 +38,7 @@ exclude =
     __pycache__/,
     node_modules/,
     */migrations/
-ignore = H101,H238,H301,H306,W503
+ignore = E203,H101,H238,H301,H306,W503
```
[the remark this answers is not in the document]
NyanKiyoshi: It is because black and flake8 conflict with this check. Example:

```python
# black
value = fruits[start : length - 1]

# flake8
value = fruits[start:length - 1]
```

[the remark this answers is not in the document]
fowczarek: Maybe we should turn off an error by line? 
(I don't know in how many lines it appears.)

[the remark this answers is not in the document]
NyanKiyoshi: It's honestly rate. When I want to avoid the conflict I would just do:
```python
a = x + 2
b = y - 1
value = fruits[a:b]
```
Then it won't conflict.

saleor/graphql/extensions/sorters.py
```diff
@@ -1,12 +1,11 @@
 import graphene
-from django.db.models import QuerySet
 
 from ..core.types import SortInputObjectType
 
 
 class PluginSortField(graphene.Enum):
     NAME = "name"
```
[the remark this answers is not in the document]
kswiatek92: I thought about this, but then if someone pass str as enum value then code will raise meaningless error, so then I should add validation, which is more code.
Personally this way it's easier for me :D

saleor/graphql/page/resolvers.py
```diff
@@ -19,8 +18,7 @@ def resolve_page(info, global_page_id=None, slug=None):
     return page
 
 
-def resolve_pages(info, query, sort_by=None, **_kwargs):
+def resolve_pages(info, query, **_kwargs):
```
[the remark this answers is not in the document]
kswiatek92: query here it's not queryset
it's misleading, and I'm not sure to what name  should change it

btw
is filter_by_query_param deprecated?

saleor/graphql/extensions/sorters.py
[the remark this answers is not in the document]
fowczarek: In the current solution in one enum, I have string or list of strings. This solution isn't clear to me. 
@maarcingebala  What do you prefer?

saleor/graphql/page/resolvers.py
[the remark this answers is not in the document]
kswiatek92: removed

saleor/graphql/extensions/sorters.py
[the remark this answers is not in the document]
maarcingebala: @kswiatek92 I think it's the same amount of code as you have now. If a value is not a list we can raise `TypeError("Sorting enum value must be a list")`.

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,180 @@
...
+        matching_records = matching_records[:requested_count]
+
+    edges = [
+        edge_type(
+            node=record,
+            cursor=to_global_cursor(
+                [get_field_value(record, field) for field in sorting_fields]
+            ),
+        )
+        for index, record in enumerate(matching_records)
```
drozdowsky: Excuse me passing by...
looking at commit history, `index` is not needed anymore
also why do we need to calculate cursor for every edge? (is there any spec?)

patrys: The spec is here: https://facebook.github.io/relay/graphql/connections.htm

requirements.txt
kswiatek92: @marcin
it doesn't help
it brings different problems
when removing `sys_platform` from `python-magic-bin` it returns
```
No matching distribution found for python-magic-bin==0.4.14 (from -r /tmp/build_5ee20a5feba89390713ce1aa800698d1/requirements.txt (line 421))
```

saleor/graphql/middleware.py
```diff
@@ -60,13 +43,14 @@ def get_service_account(auth_token) -> Optional[ServiceAccount]:
...
-def service_account_middleware(get_response):
+def service_account_middleware(next, root, info, **kwargs):
 
     service_account_auth_header = "HTTP_AUTHORIZATION"
     prefix = "bearer"
+    request = info.context
 
-    def _service_account_middleware(request):
-        if request.path == API_PATH:
+    if request.path == API_PATH:
```
korycins: AFAIK we have only API

tests/api/test_account.py
```diff
@@ -2764,7 +2764,7 @@ def test_query_customers_with_filter_placed_orders__(
         ({"field": "EMAIL", "direction": "ASC"}, ["John", "Leslie", "Joe"]),
         ({"field": "EMAIL", "direction": "DESC"}, ["Joe", "Leslie", "John"]),
         ({"field": "ORDER_COUNT", "direction": "ASC"}, ["John", "Leslie", "Joe"]),
-        ({"field": "ORDER_COUNT", "direction": "DESC"}, ["Joe", "John", "Leslie"]),
+        ({"field": "ORDER_COUNT", "direction": "DESC"}, ["Joe", "Leslie", "John"]),
```
korycins: Why did you change order here?

tests/api/test_account.py
```diff
@@ -2965,7 +2965,7 @@ def test_query_staff_members_with_filter_search(
         ({"field": "EMAIL", "direction": "ASC"}, ["John", "Leslie", "", "Joe"]),
         ({"field": "EMAIL", "direction": "DESC"}, ["Joe", "", "Leslie", "John"]),
         ({"field": "ORDER_COUNT", "direction": "ASC"}, ["John", "Leslie", "", "Joe"]),
-        ({"field": "ORDER_COUNT", "direction": "DESC"}, ["Joe", "John", "Leslie", ""]),
+        ({"field": "ORDER_COUNT", "direction": "DESC"}, ["Joe", "", "Leslie", "John"]),
```
korycins: ☝️

tests/api/test_category.py
```diff
@@ -19,7 +19,7 @@ def test_category_query(user_api_client, product):
         category(id: "%(category_pk)s") {
             id
             name
-            ancestors(last: 20) {
+            ancestors(first: 20) {
```
korycins: Why did you change last to first?

saleor/graphql/account/sorters.py
```diff
@@ -46,12 +46,8 @@ def description(self):
         raise ValueError("Unsupported enum value: %s" % self.value)
 
     @staticmethod
-    def sort_by_order_count(
-        queryset: QuerySet, sort_by: SortInputObjectType
-    ) -> QuerySet:
-        return queryset.annotate(order_count=Count("orders__id")).order_by(
-            f"{sort_by.direction}order_count", "email"
-        )
+    def prepare_qs_for_sort_order_count(queryset: QuerySet) -> QuerySet:
```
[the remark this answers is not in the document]
kswiatek92: it doesn't sort here
I moved sorting to one place, custom sorters only prepare qs for sorting (i.e. annotate)

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,180 @@
...
 from graphene_django_optimizer.types import OptimizedDjangoObjectType
+from graphql_relay.connection.connectiontypes import Edge, PageInfo
+from graphql_relay.utils import base64, unbase64
+
+from ..core.enums import OrderDirection
+
+ConnectionArguments = typing.Dict[str, typing.Any]
+
+
+def to_global_cursor(values):
```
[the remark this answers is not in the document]
kswiatek92: graphene converter expect only 2 args (type, id)
we have from 1 to 3 args

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,180 @@
...
+            attr = getattr(attr, elem)
+        except AttributeError:
+            return None
+    return attr
+
+
+def prepare_filter(cursor, sorting_fields, sorting_direction):
+    """Create filter arguments based on sorting fields.
+
+    :param cursor: list of values that are passed from page_info, used for filtering.
```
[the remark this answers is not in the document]
kswiatek92: here cursor is a list, 
list of fields values used for sorting of first/last edge in result

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,180 @@
...
+        (OR: ('first_field__gt', 'first_value_form_cursor'),
+            (AND: ('second_field__gt', 'second_value_form_cursor'),
+                ('first_field', 'first_value_form_cursor')),
+            (AND: ('thrid_field__gt', 'third_value_form_cursor'),
+                ('second_field__gt', 'second_value_form_cursor'),
+                ('first_field', 'first_value_form_cursor'))
+        )
+    """
+    filter_kwargs = Q()
+    for field_id in range(len(sorting_fields)):
```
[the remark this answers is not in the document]
kswiatek92: forgot about that :) thanks

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,180 @@
...
+        field_expression = {}
+        for val_id in range(len(cursor[:field_id])):
+            field_expression[sorting_fields[val_id]] = cursor[val_id]
```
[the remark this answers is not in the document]
kswiatek92: sorting_fields is list of model fields,
cusros is list of values of these fields

saleor/graphql/menu/sorters.py
```diff
@@ -17,12 +17,8 @@ def description(self):
         raise ValueError("Unsupported enum value: %s" % self.value)
 
     @staticmethod
-    def sort_by_items_count(
-        queryset: QuerySet, sort_by: SortInputObjectType
-    ) -> QuerySet:
-        return queryset.annotate(items_count=Count("items__id")).order_by(
-            f"{sort_by.direction}items_count", "name"
-        )
+    def prepare_qs_for_sort_items_count(queryset: QuerySet) -> QuerySet:
```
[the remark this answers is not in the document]
kswiatek92: `qs_with_items_count` would be more meaningful in my opinion

tests/api/test_account.py
kswiatek92: in sorters file I change enum values to list of fields use for sorting 
this way we don't have to worry about conflict when sorting by fields that aren't unique
as result some sorting have slightly different output
like in this example we sort by order_count and email

tests/api/test_category.py
kswiatek92: I think it was part of @patrys  testing when he started this PR

kswiatek92: I can revert this change

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,180 @@
...
+                ('first_field', 'first_value_form_cursor')),
+            (AND: ('thrid_field__gt', 'third_value_form_cursor'),
+                ('second_field__gt', 'second_value_form_cursor'),
+                ('first_field', 'first_value_form_cursor'))
+        )
+    """
+    filter_kwargs = Q()
+    for field_id in range(len(sorting_fields)):
+        field_expression = {}
+        for val_id in range(len(cursor[:field_id])):
```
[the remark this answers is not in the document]
kswiatek92: I'm thinking what short comment would fit here

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,180 @@
...
+
+def get_field_value(instance, field_name):
+    return get_repr(get_field(instance, field_name))
+
+
+def connection_args_validation(args):
+    first = args.get("first")
+    last = args.get("last")
+
+    if isinstance(first, int) and first < 0 or first and not isinstance(first, int):
```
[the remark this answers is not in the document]
kswiatek92: looks good

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,180 @@
...
+        raise ValueError("Argument 'first' must be a non-negative integer.")
+    if isinstance(last, int) and last < 0 or last and not isinstance(last, int):
+        raise ValueError("Argument 'last' must be a non-negative integer.")
+    if first and last:
+        raise ValueError("Argument 'last' cannot be combined with 'first'.")
+    if first and args.get("before"):
+        raise ValueError("Argument 'first' cannot be combined with 'before'.")
```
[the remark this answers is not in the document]
kswiatek92: fixed

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,180 @@
...
+    edge_type: typing.Any = Edge,
+    page_info_type: typing.Any = PageInfo,
+) -> Connection:
+    """Create a connection object from a QuerySet."""
+    args = args or {}
+    before = args.get("before")
+    after = args.get("after")
+    first = args.get("first")
+    last = args.get("last")
+    connection_args_validation(args)
```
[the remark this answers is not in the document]
kswiatek92: it would :)

saleor/graphql/core/connection.py
```diff
@@ -1,7 +1,180 @@
...
+
+    cursor_after = from_global_cursor(after) if after else None
+    cursor_before = from_global_cursor(before) if before else None
+    cursor = cursor_after or cursor_before
+    cursor_offset = 1 if cursor else 0
+
+    sort_by = args.get("sort_by", {})
+    sorting_direction = sort_by.get("direction", "")
+    sorting_fields = sort_by.get("field")
```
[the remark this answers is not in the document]
kswiatek92: field can be passes as None
then we will have `sorting_fields = None`
so still have to have `elif not sorting_fields; sorting_fields = []`

tests/api/test_category.py
patrys: No, it's no longer legal to specify `last` before specifying `before`.

kswiatek92: actually I talked (i think with @maarcingebala if he can confirm) that you can specify `last` without `before` becuase we can actually get last records (it's pointless because it same results as first but with revert ordering)

patrys: There is already an option to reverse the results, I don't think it makes sense to support `last` without `before`.

saleor/graphql/middleware.py
maarcingebala: What do you mean @korycins? We have one or two more views e.g. to serve digital content files.

tests/api/test_category.py
kswiatek92: but when you won't provide sort_by arg then you cannot reverse results
this doesn't add complicity in code or anything
we support only last as default
I don't see a point for removing support for `last` now

saleor/graphql/account/sorters.py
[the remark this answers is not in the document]
kswiatek92: changed

requirements.txt
kswiatek92: removed hashes and created bug ticket for it

tests/api/test_category.py
patrys: That's interesting, why can't you reverse it then?

kswiatek92: when you send query like this
`
query {
customers(first 100){
 name
}
}
`

you are unable to reverse it's order
you have to provide sorter_by
so you can use ASC DESC

saleor/graphql/core/connection.py
```diff
@@ -1,11 +1,228 @@
-from typing import Union
+import typing
```
patrys: Why not import the ones we want? `foo: typing.List[typing.Any]` is not very readable.

saleor/graphql/core/connection.py
```diff
@@ -1,11 +1,228 @@
...
+    return attr
+
+
+def _prepare_filter_expression(
+    field_name: str,
+    index: int,
+    cursor: typing.List,
+    sorting_fields: typing.List,
+    sorting_direction: str,
+) -> typing.Tuple:
```
patrys: `List` and `Tuple` are generics that should be further qualified with type parameters

tests/api/test_category.py
patrys: Isn't `first: 100` the same as `first: 100, sortBy: {field: SOMETHING}`? And if not, _why not_?

saleor/graphql/core/connection.py
kswiatek92: `from graphene import Field, List, NonNull, ObjectType, String`
`from typing import List, Union`

patrys: In this case, I think it's more readable to import `graphene` as a namespace.

kswiatek92: done

saleor/graphql/core/connection.py
kswiatek92: done

saleor/product/models.py
```diff
@@ -162,14 +163,24 @@ def collection_sorted(self, user: "User"):
...
         return qs
 
-    def sort_by_attribute(self, attribute_pk: Union[int, str], ascending: bool = True):
+    def sort_by_attribute(
+        self, attribute_pk: Union[int, str], descending: bool = False
+    ):
         """Sort a query set by the values of the given product attribute.
 
         :param attribute_pk: The database ID (must be a number) of the attribute
                              to sort by.
```
IKarbowiak: This description doesn't fit to typing in method definition :information_desk_person:

kswiatek92: fixed

tests/api/test_category.py
kswiatek92: not exactly

only if `SOMETHING` is default sorting

in that case
`first: 100` the same as `first: 100, sortBy: {field: SOMETHING, direction: ASC}`
and
`last: 100` the same as `first: 100, sortBy: {field: SOMETHING, direction: DESC}`

tests/api/test_pagination.py
```diff
@@ -0,0 +1,1798 @@
...
+        status=OrderStatus.FULFILLED,
+        token="order_token_5",
+        total_gross_amount=Decimal("50"),
+        billing_address=address5,
+    )
+
+    Payment.objects.bulk_create(
+        [
+            Payment(
+                gateway="Dummy",
```
korycins: What about some pytest factories to cover these objects? The same thing for order and address. Here you have a lot of duplicated code.

tests/api/test_pagination.py
```diff
@@ -0,0 +1,1798 @@
...
+                min_spent_amount=Decimal("100"),
+                usage_limit=10,
+            ),
+        ]
+    )
+
+
+@pytest.fixture
+def menus_for_pagination():
+    Menu.objects.bulk_create(
```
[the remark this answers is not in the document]
korycins:
```
In [10]: tt = Checkout.objects.bulk_create([Checkout(token="1fa91751-fd0a-45ca-8633-4bac9f5cb2a2"), Checkout(token="1fa91751-fd0a-45ca-8633-4bac9f5cb2a1")])
Number of queries 1

In [11]: tt
Out[11]: [Checkout(quantity=0), Checkout(quantity=0)]
```

tests/api/test_pagination.py
```diff
@@ -0,0 +1,1798 @@
...
+                storefront_search_position=5,
+            ),
+        ]
+    )
+
+
+@pytest.fixture
+def categories_for_pagination():
+
+    product_type1 = ProductType.objects.create(
```
korycins: bulk

tests/api/test_category.py
patrys: >`last: 100` the same as `first: 100, sortBy: {field: SOMETHING, direction: DESC}`

I think they are actually not the same. I would expect `last` without a `before` to return the correct items (eg. the same as `first` with `direction: DESC`) but in the wrong order (the entire list reversed).

Consider nine products with names A, B, C, D, E, F, G, H, I.

* `first: 3`, `direction: ASC` -> `["A", "B", "C"]`
* `last: 3`, `direction: ASC` -> `["G", "H", "I"]`
* `first: 3`, `direction: DESC` -> `["I", "H", "G"]`
* `last: 3`, `direction: DESC` -> `["G", "H", "I"]`

So it's not a convenient way to reverse the sorting order. I don't see a reason to support this.

kswiatek92: ok, you're right, I will add constraint on `before` :P

saleor/graphql/core/connection.py
```diff
@@ -1,11 +1,227 @@
...
+    return base64("::".join(values))
+
+
+def from_global_cursor(cursor) -> List[str]:
+    values = unbase64(cursor)
+    values = values.split("::")
+    if isinstance(values, list):
+        values = [None if value == "None" else value for value in values]
+        return values
+    if values == "None":
```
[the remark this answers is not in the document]
kswiatek92: I think it will make debugging and working with this harder

if I have sorting fields `[name, slug, pk]`
I should have "unpacked" cursor as `[name, slug, pk]`

if I have name == None and didn't returned value I would end up with different cursor length than sorting_fields length

I will be impossible to create filter out of this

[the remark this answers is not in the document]
kswiatek92: and None in filter is value as any other

saleor/graphql/core/connection.py
```diff
@@ -1,11 +1,227 @@
...
+        values = [str(values)]
+    else:
+        values = [str(value) for value in values]
+    return base64("::".join(values))
+
+
+def from_global_cursor(cursor) -> List[str]:
+    values = unbase64(cursor)
+    values = values.split("::")
+    if isinstance(values, list):
```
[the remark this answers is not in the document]
kswiatek92: cursor can be only 'pk' in that case it will be just one value

saleor/graphql/core/connection.py
```diff
@@ -1,11 +1,227 @@
...
+        values = None
+    return [values]
+
+
+def get_field_value(instance: DjangoModel, field_name: str):
+    """Get field value for given field in filter format 'field__foreign_key_field'."""
+    field_path = field_name.split("__")
+    attr = instance
+    for elem in field_path:
+        attr = getattr(attr, elem)
```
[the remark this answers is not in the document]
kswiatek92: no because I don't know where this last elem belong

in simplest case with "__" I could just call
attr = getattr(instance, field_path[1])
attr = getattr(attr, field_path[0])

but I cannot guarantee that this will be the case all the time

saleor/graphql/core/connection.py
```diff
@@ -1,11 +1,227 @@
...
+        return qs.model.sort_by_attribute_fields()
+    elif not sorting_fields:
+        raise ValueError("Error while preparing cursor values.")
+    return sorting_fields
+
+
+def _get_sorting_direction(sort_by, last=None):
+    sorting_direction = sort_by.get("direction", "")
+    sorting_direction = "lt" if sorting_direction == OrderDirection.DESC else "gt"
+    if last:
```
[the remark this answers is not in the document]
kswiatek92: when using last arg I have to reversed order and reversed filter

when I have list [1,2,3,4,5,6,7,8,9]
anf have cursor 5
`first: 2, after: 5` will return [6,7]
`last:2, before: 5' should return [3,4]
to do this I have to first reverse order [9,8,7,6,5,4,3,2,1] and gt replace with lt [4,3,2,1]

without reversing order I would end up with [1,2]

saleor/graphql/core/connection.py
```diff
@@ -1,11 +1,227 @@
...
+        # reversed direction
+        sorting_direction = "gt" if sorting_direction == "lt" else "lt"
+    return sorting_direction
```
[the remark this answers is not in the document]
kswiatek92: thanks, changed

saleor/graphql/core/fields.py
```diff
@@ -41,6 +42,36 @@ def __init__(self, *args, **kwargs):
...
+                pageinfo_type=PageInfo,
+            )
+            connection.iterable = iterable
+        else:
+            iterable, sort_by = sort_queryset_for_connection(
+                iterable=iterable, args=args
+            )
+            args["sort_by"] = sort_by
+            connection = connection_from_queryset_slice(
+                iterable,
```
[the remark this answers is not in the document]
kswiatek92: thanks, changed

tests/api/test_pagination.py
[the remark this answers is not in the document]
kswiatek92: fixed, thanks

saleor/graphql/core/connection.py
```diff
@@ -1,11 +1,227 @@
...
+from graphql_relay.connection.connectiontypes import Edge, PageInfo
+from graphql_relay.utils import base64, unbase64
+
+from ..core.enums import OrderDirection
+
+ConnectionArguments = Dict[str, Any]
+
+
+def to_global_cursor(values):
+    if not isinstance(values, list):
```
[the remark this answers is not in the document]
kswiatek92: I think Iterable is ok, testing and fixing

saleor/graphql/core/connection.py
```diff
@@ -1,11 +1,227 @@
...
+        values = [str(values)]
+    else:
+        values = [str(value) for value in values]
+    return base64("::".join(values))
```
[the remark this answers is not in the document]
kswiatek92: looks good, fixed

saleor/graphql/core/fields.py
```diff
@@ -41,6 +42,36 @@ def __init__(self, *args, **kwargs):
...
+                args,
+                slice_start=0,
+                list_length=_len,
+                list_slice_length=_len,
+                connection_type=connection,
+                edge_type=connection.Edge,
+                pageinfo_type=PageInfo,
+            )
+            connection.iterable = iterable
+        else:
```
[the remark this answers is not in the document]
kswiatek92: this can receive only 2 values, queryset or list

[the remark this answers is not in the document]
kswiatek92: and I only did this because there are some places which return "queryset" as list

saleor/graphql/core/connection.py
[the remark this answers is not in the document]
patrys: `None` is a Python literal but also a perfectly valid string, what if I name my product "None"?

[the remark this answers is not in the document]
kswiatek92: @patrys damn it :(

I don't have solution for that situation 😭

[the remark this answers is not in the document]
kswiatek92: ok I'll try `\N` and will see what happens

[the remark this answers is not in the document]
kswiatek92: @patrys proposed `base64(json.dumps([none, 1, "123"]))` and it look like it's working, perfect

saleor/graphql/core/connection.py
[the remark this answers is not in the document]
patrys: Why not use a single-element list in this case?

saleor/graphql/product/sorters.py
```diff
@@ -45,10 +53,10 @@ def description(self):
                 "Sort attributes by their position in storefront"
             ),
             AttributeSortField.DASHBOARD_VARIANT_POSITION.name: (
-                "Sort variant attributes by their position in dashboard."
+                "Deprecated: Sort variant attributes by their position in dashboard."
```
maarcingebala: For any deprecation notice, we should say in which version something will be removed. Since we have now 2.9 and removed stuff 2 versions later, it should say: "DEPRECATED: Will be removed in Saleor 2.11." Same below.

saleor/graphql/utils.py
```diff
@@ -8,13 +8,17 @@
...
 
-from .core.enums import PermissionEnum, ReportingPeriod
+from .core.enums import OrderDirection, PermissionEnum, ReportingPeriod
 from .core.types import PermissionDisplay, SortInputObjectType
 
 ERROR_COULD_NO_RESOLVE_GLOBAL_ID = (
     "Could not resolve to a node with the global id list of '%s'."
 )
 registry = get_global_registry()
+REVERSED_DIRECTION = {
```
maarcingebala: This PR adds multiple utilities to this file which all are related to sorting. I think it may be a good time to reorganize it a bit and have:
```
utils/
  __init__.py
  sorting.py
```

And put all this sorting stuff to `sorting.py` while `__init__.py` would contain all the other functions taht are currently in `utils.py`. What do you think?

saleor/product/models.py
```diff
@@ -180,67 +191,71 @@ def sort_by_attribute(self, attribute_pk: Union[int, str], ascending: bool = Tru
...
-        # Descending sorting
-        if not ascending:
-            return qs.reverse()
-        return qs
+        ordering = "-" if descending else ""
+        return qs.order_by(
+            f"{ordering}concatenated_values_order",
+            f"{ordering}concatenated_values",
+            f"{ordering}name",
+        )
```
NyanKiyoshi: Did you check the cost of the queries? Those were very optimized, and a lot of work was put into making them as optimal as possible.

Could you test them and see how they perform underneath? Using `EXPLAIN`

saleor/graphql/utils.py
kswiatek92: sure, could point

saleor/graphql/product/sorters.py
kswiatek92: sure

saleor/product/models.py
kswiatek92: I didn't changing anything here, except that I moved `if` statement from resolver to the beginning of this method

and this I don't think changing sorting direction make this code less optimal

saleor/graphql/utils.py
kswiatek92: done

saleor/graphql/product/sorters.py
kswiatek92: done

saleor/graphql/utils/sorting.py
```diff
@@ -0,0 +1,114 @@
...
+            deafault_ordering.append(f"{direction}{field.expression.name}")
+    return deafault_ordering
+
+
+def sort_queryset_by_default(
+    queryset: QuerySet, reversed: bool
+) -> Tuple[QuerySet, dict]:
+    """Sort queryset by it's default ordering."""
+    queryset_model = queryset.model
+    deafault_ordering = ["pk"]
```
NyanKiyoshi: Stomped on a typo

kswiatek92: done

saleor/order/migrations/0081_merge_20200309_0952.py
```diff
@@ -0,0 +1,13 @@
+# Generated by Django 2.2.10 on 2020-03-09 14:52
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
```
korycins: Instead of merging migration, you could just remove your migration and re-generate it one more time.

tests/api/test_pagination.py
```diff
@@ -0,0 +1,1654 @@
...
+                discount_value=Decimal("2"),
+                min_spent_amount=Decimal("100"),
+                usage_limit=10,
+            ),
+        ]
+    )
+
+
+@pytest.fixture
+def address():
```
korycins: global conf already has fixture with that name.

tests/api/test_pagination.py
```diff
@@ -0,0 +1,1654 @@
...
+@pytest.fixture
+def categories_for_pagination():
+    product_types = ProductType.objects.bulk_create(
+        [
+            ProductType(name="Default Type", slug="default-type",),
+            ProductType(name="Default Type", slug="second-type",),
+        ]
+    )
+
+    category1 = Category.objects.create(name="The Category", slug="the-category",)
```
korycins: Bulk?

tests/api/test_pagination.py
```diff
@@ -0,0 +1,1654 @@
...
+def test_user_with_order_pagination(
+    order_by,
+    cursor_fields,
+    user_order,
+    staff_api_client,
+    permission_manage_users,
+    users_for_pagination,
+):
+    query = """
+        query (
```
korycins: query from this test and from `test_user_pagination` looks very similar, maybe you are able to extract it to single variable?

korycins:
@kswiatek92  `test_pagination.py` will be very time & resource consuming. We should consider some step before merge:
 - add bulk everywhere
 - consider using fixture(scope='module')  for all `bulk` fixtures. 
- consider changing `create_cursor_from_*` to use somehow mock for it.

tests/api/test_pagination.py
kswiatek92: category below use this one as parent
using bulk I would have to update cetegories below

korycins: https://github.com/mirumee/saleor/pull/5149/files/82318b7f6a5eb6f34ffacb088336a525adaeda87#diff-f177a26d6e2119236d7f092903f38899R408-R409
This can be done.

tests/api/test_pagination.py
```diff
@@ -0,0 +1,1654 @@
...
+    attribute = Attribute.objects.get(slug=slug)
+    return create_global_cursor(attribute, fields_list)
+
+
+def create_cursor_from_category_slug(slug, fields_list):
+    category = Category.objects.filter(slug=slug)
+    category = CategorySortField.qs_with_product_count(category)
+    category = CategorySortField.qs_with_subcategory_count(category).first()
+
+    return create_global_cursor(category, fields_list)
```
patrys: Why do we need those? Why not use the API to fetch a page of 2 items and obtain the cursor to the next page and so on? We don't need to document how cursors are created as only the GraphQL server will ever create them and we should be able to change their format as often as we see fit.

tests/api/test_pagination.py
```diff
@@ -0,0 +1,1654 @@
...
+        query,
+        order_by,
+        permission_manage_users,
+        start_cursor_page1,
+        end_cursor_page1,
+        start_cursor_page2,
+        end_cursor_page2,
+        start_cursor_before,
+        resolver_name="customers",
+    )
```
patrys: It's impossible to tell what this test is checking as all the assertions are hidden inside a private helper.

tests/api/test_pagination.py
```diff
@@ -0,0 +1,1654 @@
...
+
+
+@pytest.mark.parametrize(
```
patrys: Why is this marked as parametrized?

tests/api/test_pagination.py
```diff
@@ -0,0 +1,1654 @@
...
+                "zordon@example.com",
+                "leslie@example.com",
+                "ellen@example.com",
+                "ddevito@example.com",
+                "allen@example.com",
+            ],
+        ),
+    ],
+)
+def test_user_pagination(
```
patrys: I can't tell what this test is checking as all of the data is prepared outside of test's body, all assertions are made outside of test's body and expected output refers to a field different than the one used for sorting so even verifying that the assumptions are correct is impossible. It seems to be five separate tests disguised as one.

It would be easier to read if it was five separate tests with fields used for sorting explicitly being created (or updated) inside the test body finished with an assertion that a particular field is properly sorted (eg. `assert [user["foo"] for user in results] == ["a", "b", "c", "d", "e"]`).

saleor/graphql/core/connection.py
```diff
@@ -216,6 +216,8 @@ def connection_from_queryset_slice(
     )
     qs = qs.filter(filter_kwargs)
     qs = qs[:end_margin]
+    print(end_margin)
+    print(qs)
```
fowczarek: We should remove this.

saleor/graphql/core/connection.py
```diff
@@ -1,11 +1,226 @@
...
+
+    return filter_kwargs
+
+
+def _validate_connection_args(args):
+    first = args.get("first")
+    last = args.get("last")
+
+    if first and not (isinstance(first, int) and first > 0):
+        raise ValueError('Argument "first" must be a non-negative integer.')
```
maarcingebala, line 100: I suggest that we change these `ValueError`s to `GraphQLError`. This would be consistent with other errors returned by Graphene when passing an invalid query.

PR #5152, opened 2020-01-09
move extracting requester to utils

kswiatek92:
service account can be None
but user will always be set (user or anonymous)
previous `info.context.user or info.context.service_account` always returned user

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/resolvers.py
+++ b/saleor/graphql/product/resolvers.py
@@ -11,6 +11,7 @@ from ..utils import (
     filter_by_period,
     filter_by_query_param,
     get_database_id,
+    get_user_or_service_account_from_context,
     get_nodes,
     sort_queryset,
 )
@@ -141,9 +142,7 @@ def resolve_products(
     **_kwargs,
 ):
 
-    user = info.context.user
-    if user.is_anonymous and info.context.service_account:
-        user = info.context.service_account
+    user = get_user_or_service_account_from_context(info.context)
     qs = models.Product.objects.visible_to_user(user)
     qs = sort_products(qs, sort_by)
 
--- a/saleor/graphql/utils.py
+++ b/saleor/graphql/utils.py
@@ -183,3 +183,9 @@ def create_jwt_payload(user, context=None):
     payload["is_staff"] = user.is_staff
     payload["is_superuser"] = user.is_superuser
     return payload
+
+
+def get_user_or_service_account_from_context(context):
+    # order is important
+    # service_account can be None but user if None then is passed as anonymous
+    return context.service_account or context.user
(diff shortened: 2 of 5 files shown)


saleor/graphql/product/resolvers.py
```diff
@@ -141,9 +142,7 @@ def resolve_products(
     **_kwargs,
 ):
 
-    user = info.context.user
```
fowczarek: Probably we should use `get_requester_from_context` in almost all places where `user = info.context.user` appear in code. All query and mutation should work for user and service account with proper permissions.

django-queries:
Here is the report for 07c977be21bc735e045b817581e0c903bdc312be (kswiatek92/saleor @ move-getting-requester-to-utils)
Base comparison is a0c379ef7a81073c509d11907b2bef341eb12da2.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	         50	         50	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         14	         14	              2
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         16	         16	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/graphql/product/resolvers.py
maarcingebala: Sometimes we want to get only user and sometimes user or service account. This is the second case. For getting user from the context we could also have a helper.

saleor/graphql/utils.py
```diff
@@ -183,3 +183,9 @@ def create_jwt_payload(user, context=None):
     payload["is_staff"] = user.is_staff
     payload["is_superuser"] = user.is_superuser
     return payload
+
+
+def get_requester_from_context(context):
```
maarcingebala: Maybe `get_user_or_service_account_from_context`? A bit long but I assume we could also have `get_user_from_context` and just `get_service_account_from_context`.

PR #5161, opened 2020-01-13
Add sample payload data anonymization

derenio:
This `PR` adds anonymization to the sample payloads of the GraphQL's `webhookSamplePayload` endpoint.


django-queries:
Here is the report for 0a09f406b7bf7f09234c338e42141559561f4491 (mirumee/saleor @ feature/anonymize_sample_payloads)
Base comparison is 3f5c4c870b4243bb8dcbee6855d687305e1d43de.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	         50	         50	             24

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         14	         14	              2
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         16	         16	              6

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/core/utils/anonymization.py
```diff
@@ -0,0 +1,71 @@
...
+from ...account.models import Address, User
+
+if TYPE_CHECKING:
+    from ...order.models import Order
+    from ...checkout.models import Checkout
+
+fake = Faker()
+
+
+def generate_fake_address() -> "Address":
```
maarcingebala: In `saleor/core/utils/random_data.py` we have a bunch of functions that do the same functionality:
- `create_address` which also handles generating valid states and postal codes for the US addresses
- `create_fake_user` 

Since we have a module for generating random data, I think we should keep all related functions there and try to reuse ones that do the same thing.

saleor/core/utils/anonymization.py
```diff
@@ -0,0 +1,71 @@
...
+        note=fake.paragraph(),
+        date_joined=fake.date_time(),
+        default_shipping_address=generate_fake_address(),
+        default_billing_address=generate_fake_address(),
+        meta=fake.pystruct(count=1),
+        private_meta=fake.pystruct(count=1),
+    )
+
+
+def anonymize_order(order: "Order"):
```
maarcingebala: For anonymization itself, we also have a module already - `saleor/extensions/plugins/anonymize/__init__.py` - it's a plugin which we use to anonymize checkouts in the demo store. I'm not sure if we can reuse those functions in those two places, but we could try to have less code duplication.

saleor/core/utils/anonymization.py
derenio: I've reused the `create_address` and `create_fake_user` from the `random_data`. Though I had to add the change to make the `saving` optional as we don't want to save new or update the existing instances.

I left the functions in the `anonymization` module. Its purpose is different from the `random_data`. The `anonymization` doesn't want to save/update anything in the database. Although, because of the complex structure of relations it wants to be based on an already existing instances.

saleor/core/utils/anonymization.py
derenio: The `anonymize` plugin also differs in that it saves the changes and it _obfuscate_ rather than generates _fake_ data.

I'm open for suggestions. Perhaps some name changing could clarify things. For me these modules perform similar but not identical functionality and I don't see how to merge them.

Issue #5162, opened 2020-01-13
Carry foreward arbitrary fields from checkout to order


dankolbman:
### What I'm trying to achieve
I'd like to include additional fields during checkout for the customer to fill out, such as agreeing to some terms, which will carry forward to the  order so it may be validated or acted on by an administrator.

Currently, we can store metadata fields on the checkout to keep track of these additional values, but once the checkout is completed, the fields are lost.

### Describe a proposed solution
There are three different solutions I can think of right now.

#### Copy checkout metadata to order
Perhaps the easiest solution is to copy the metadata, public and private, from the checkout to the new order when an order is completed. This will allow our current approach of storing additional values on checkout metadata while the user fills out the form.

#### Allow plugins to modify order data
The `manager.preprocess_order_creation()` function could be allowed to modify the `order_data`during the `prepare_order_data()` call which would let a plugin modify an order before it's created using information from the checkout.

#### Manage everything on the frontend
The possible noop solution here could be to update the order information post-creation on the front end. I'm not sure this is possible due to permissions, though.
It still wouldn't be ideal for our use as we would want to still add some backend logic to prevent creating orders that violate some conditions.

maarcingebala:
I think in your particular case, the fastest solution would be to build a plugin for that, so solution 2). But I think we should implement solution 1) anyway - if there is any additional metadata stored in checkout instance, it's likely that you may want to access it later once it becomes an order.

dankolbman:
The plugin seems that it should be the easiest way to get this functionality, but if I'm reading it correctly, the `order_created()` hook gets called after the order has been created and the checkout has been deleted, making it difficult to do without managing some temporary storage for the plugin.
The `preprocess_order_creation()` could be allowed to modify and return the `order_data` before the new order is constructed, but this may be unsafe?

I PR'd what I'm using to copy the metadata, if you're ok with having what may be an inconsistency with how metadata is handled. Otherwise, it may be worth looking more into accommodating the plugin approach which may have more benefits, too.

maarcingebala:
Fixed in #5165

PR #5165, opened 2020-01-14
Copy metadata to order upon checkout completion

dankolbman:
Metadata that has been added to a checkout may be needed post-completion. Currently, all metadata is lost on the checkout once it is completed as the checkout object is deleted during order creation.

This PR copies both the meta and private_meta from the checkout object to the new order when it is created.

Addresses #5162 

### Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Move extracting user or service_account from context to utils - #5152 by @kswiatek92
 - Add deprecate description to order status/created arguments - #5076 by @kswiatek92
 - Fix getting title field in page mutations #5160 by @maarcingebala
+- Copy public and private metadata from the checkout to the order upon creation -  #5165 by @dankolbman
 
 ## 2.9.0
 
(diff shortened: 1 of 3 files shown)


dankolbman:
Questions to address before merging:

1) Does it make sense to have an 'exception' to how the metadata behaves in this context? Is there another instance where metadata is populated or modified automatically?
2) Should a similar flow be implemented for draft orders? I'm not sure how drafts go to orders, although I see there may not be any metadata mutations in the schema, so maybe these should be added first?

maarcingebala:
@dankolbman First of all thanks for the PR, looks good! 

1. I don't recall any other case in our codebase where metadata would be populated automatically. I don't think this is a problem since we copy a lot of stuff from checkout to orders anyway. Maybe we could put checkout's metadata under a particular key in order's metadata, e.g. `CHECKOUT_META`. Just not to mix checkout's meta fields with your order meta keys.

2. I agree we should first add metadata mutations for draft orders (I thought we have them), but I would say it's beyond the scope of this PR.

dankolbman:
@maarcingebala thanks for the comments. It sounds like these changes should be ok for now.

I'll leave the decision on 1) to your discretion as I'm not very clear the philosophy of how the metadata is being organized regarding clients and namespaces. Maybe an opportunity for more documentation?

I opened #5171 to track adding the new draft order mutations for metadata.

PR #5179, opened 2020-01-20
Handle additional confirmation in checkout complete flow

### LINKED ISSUE #4773: Add additional confirmation state to checkout
Some payment methods might result in additional step required to finish payment after _process_payment_ is called. We do have process payment coupled with order creation in checkout flow and it is impossible to do that step using existing API.

_CheckoutComplete_ mutation could return either checkout or order, depending on payment status. Checkout can be extended with additional state field.
When going from intermediate state to finished, mutation could confirm payment instead of processing new one.

salwator:
This PR enables our checkout process to handle additional confirmation step needed for multi-factor authentication.

checkoutComplete mutation now returns additional flag indicating that action is required by payment gateway, next call will confirm payment instead of capturing again.

Fixes #4773

### Pull Request Checklist


django-queries:
Here is the report for 091e65c8af6b99366e04b7557570477fe4f32e31 (mirumee/saleor @ add-confirmation-to-checkout-complete)
Base comparison is 7f67c75348628d7296c964d99208cc9923ea3a70.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	          5	          5	              1

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         18	         18	              4
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         23	         23	              9

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/graphql/checkout/mutations.py
```diff
@@ -628,6 +628,13 @@ def perform_mutation(cls, _root, info, checkout_id, shipping_method_id):
 
 class CheckoutComplete(BaseMutation):
     order = graphene.Field(Order, description="Placed order.")
+    confirmation_needed = graphene.Boolean(
```
korycins: I assume that after we added this flag we should also modify the description of the `class Meta` as there is a possibility that we won't create new order.

saleor/graphql/checkout/mutations.py
```diff
@@ -733,19 +744,23 @@ def perform_mutation(cls, _root, info, checkout_id, store_source, **data):
...
+            checkout.delete()
 
-        # remove checkout after order is successfully paid
-        checkout.delete()
+            # return the success response with the newly created order data
+            return CheckoutComplete(order=order)
 
-        # return the success response with the newly created order data
-        return CheckoutComplete(order=order)
+        return CheckoutComplete(order=None, confirmation_needed=True)
```
korycins: Don't we need a separate mutation for payment that needs confirmation?
Now we have `CheckoutComplete` but we don't complete the checkout as we need to do additional steps.

korycins: What should be the next step in the checkout flow when we get `CheckoutComplete(order=None, confirmation_needed=True)`?

saleor/payment/gateway.py
```diff
@@ -177,7 +179,7 @@ def void(payment: Payment) -> Transaction:
 @require_active_payment
 def confirm(payment: Payment) -> Transaction:
     plugin_manager = get_extensions_manager()
-    token = _get_past_transaction_token(payment, TransactionKind.AUTH)
+    token = _get_past_transaction_token(payment, TransactionKind.CAPTURE)
```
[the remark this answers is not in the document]
salwator: We do capture when processing payment during checkoutComplete. Doesn't really make sense to start with supporting auth.

saleor/graphql/checkout/mutations.py
salwator: You confirm payment client side (depends on gateway) and call again.

saleor/payment/gateway.py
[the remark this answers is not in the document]
salwator: We do capture when processing payment during checkoutComplete. Doesn't really make sense to start with supporting auth.

saleor/graphql/checkout/mutations.py
korycins: So to finish checkout I need to execute mutation `CheckoutComplete` twice?

salwator: yes, in case there is additional confirmation required, you need to call mutation again

salwator:
@korycins 
After brief attempt to integrate this branch with stripe client, I discovered additional issues to be addressed before full two-factor payment is possible.
Since those are loosly connected issues and this checkout change has far more implications, I suggest merging this and solving remaining issues separately @maarcingebala .

saleor/graphql/checkout/mutations.py
```diff
@@ -655,6 +655,13 @@ def perform_mutation(cls, _root, info, checkout_id, shipping_method_id):
 
 class CheckoutComplete(BaseMutation):
     order = graphene.Field(Order, description="Placed order.")
+    confirmation_needed = graphene.Boolean(
+        required=False,
```
maarcingebala: This is `required=False` on purpose? I mean are there cases where we want to return `null` here and not `true` or `false`?

maarcingebala:
We officially support Braintree and Stripe. Is this PR tested with these gateways @salwator?

saleor/graphql/checkout/mutations.py
salwator: Well, good point.

maarcingebala:
@salwator OK so if we know what needs to be done to fully support two-factor payments let's gather these requirements in one GH issue.

saleor/payment/migrations/0016_payment_to_confirm.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.9 on 2020-01-20 17:30
```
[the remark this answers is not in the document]
korycins: @salwator  what about this?

PR #5196, opened 2020-01-27
Add slug to Warehouse, Product, ProductType and update slug in models which already using it

### LINKED ISSUE #5131: Slugify of floats and integers can lead to unique failures
Let's say I have products having for diameters:
- 40.5
- 405

If I was to provide the values like so (40.5 as float and 405 as int), it would generate the slugs `405` and `405`; triggering an error saying the values are duplicated. The solution could be to make sure to always enter a floating number–but it could fail as well depending on the values.

### What I expected to happen
Having a special type or handling for floating values.

### LINKED ISSUE #5132: Cannot create multiple values when they sometime can contain a special character
Let's say I want to sell processors having for possible socket attribute values:
- AM3
- AM3+
- AM4
- FM1
- FM1+

The following slugs would be created:
- AM3
- AM3
- AM4
- FM1
- FM1

Leading to being unable to create those values contain a `+` sign which has a totally different meaning.

IKarbowiak:
Add slug field to Warehouse, Product and ProductType models. Update slug field in Category, Collection models (make it unique) and also in Attribute, AttributeValu models - change max_length. 

Closes #5132 
Closes #5131 

### Pull Request Checklist


django-queries:
Here is the report for c6d138566f25b473f5432299ba5a2d9c15f8f0ed (mirumee/saleor @ add-slug-to-warehouse-and-producttype)
Base comparison is 9dfd1bce43b534883e5c62b6cdab019b01d89480.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	          5	          5	              1

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         18	         18	              4
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         23	         23	              9

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/product/models.py
```diff
@@ -98,6 +98,7 @@ def __repr__(self) -> str:
 
 class ProductType(ModelWithMetadata):
     name = models.CharField(max_length=128)
+    slug = models.SlugField(max_length=50, unique=True)
```
maarcingebala: `slug` fields should have max_length the same as fields they can be generated from.

saleor/warehouse/models.py
```diff
@@ -22,6 +22,7 @@ def for_country(self, country: str):
 class Warehouse(models.Model):
     id = models.UUIDField(default=uuid.uuid4, primary_key=True)
     name = models.CharField(max_length=255)
+    slug = models.SlugField(max_length=50, unique=True)
```
maarcingebala: Same here about `max_length`.

saleor/product/models.py
fowczarek: @maarcingebala  We should have at least a few more allowed characters in the slug field, for ower unique slug generator.

IKarbowiak: I think it should be 4 or 5 more characters. One for `-` sign and 3 or 4 for digits.

saleor/core/utils/__init__.py
```diff
@@ -111,3 +118,37 @@ def create_thumbnails(pk, model, size_set, image_attr=None):
         logger.info("Created %d thumbnails", num_created)
     if failed_to_create:
         logger.error("Failed to generate thumbnails", extra={"paths": failed_to_create})
+
+
+def generate_unique_slug(
+    instance: Type[Model], slugable_value: str, slug_field_name: str = "slug",
+) -> str:
+    """Create unique slug for model instance.
```
fowczarek: Maby some hints with an example of how we generate this unique slug. Good to know what is appended to slug without reading the whole function.

saleor/graphql/page/mutations.py
```diff
@@ -38,10 +38,18 @@ class Meta:
...
-            cleaned_input["slug"] = slugify(title)
+        try:
+            cleaned_input = validate_slug_and_generate_if_needed(
+                instance, "title", cleaned_input
+            )
+        except ValidationError:
+            raise ValidationError(
+                {
+                    "slug": ValidationError(
+                        "Slug value cannot be blank.", code="required"
```
fowczarek: We should use PageErrorCode.

saleor/graphql/product/mutations/products.py
```diff
@@ -204,8 +216,19 @@ class Meta:
...
-            cleaned_input["slug"] = slugify(cleaned_input["name"])
+        try:
+            cleaned_input = validate_slug_and_generate_if_needed(
+                instance, "name", cleaned_input
+            )
+        except ValidationError:
+            raise ValidationError(
+                {
+                    "slug": ValidationError(
+                        "Slug value cannot be blank.",
```
fowczarek: Maby we should use the message from catched exception, in all places when we use `validate_slug_and_generate_if_needed`?

saleor/graphql/warehouse/types.py
```diff
@@ -20,7 +20,7 @@ class WarehouseAddressInput(graphene.InputObjectType):
 
 
 class WarehouseInput(graphene.InputObjectType):
-    name = graphene.String(description="Warehouse name.", required=True)
+    slug = graphene.String(description="Waregouse slug.")
```
maarcingebala:
```suggestion
    slug = graphene.String(description="Warehouse slug.")
```

saleor/core/utils/__init__.py
IKarbowiak: You thinking about something like this in docstring:
```
    Example:
        For Product model where product variable is a Product instance:
        >>> generate_unique_slug(product, product.name, "slug")
```

maarcingebala: I think @fowczarek meant that we could roughly explain the logic behind generating unique slugs. Something like:

```
"""
Create unique slug for model instance.

The function uses `django.utils.text.slugify` to generate a slug from the `slugable_value` model field. If the slug already exists it adds a numeric suffix and increments it until a unique value is found.
"""
```

saleor/graphql/core/utils/__init__.py
```diff
@@ -62,3 +68,32 @@ def get_node_optimized(qs, lookup, info):
...
+
+def validate_slug_and_generate_if_needed(
+    instance: Type["Model"],
+    slugable_field: str,
+    cleaned_input: dict,
+    slug_field_name: str = "slug",
+) -> dict:
+    """Validate slug from input and generate in create mutation if is not given."""
+    # update mutation - just check if value is not empty
+    if getattr(instance, slug_field_name):
```
maarcingebala: If `slug_field_name` is not present in the `instance` this would raise an error. Don't we want to prevent from that by using something like `getattr(instance, slug_field_name, None)`?

IKarbowiak: Maybe the best solution will be checking if `instance` has `pk` value instead of checking `slug_field_name`, because `pk` is always present in model instance.

saleor/graphql/core/utils/__init__.py
```diff
@@ -62,3 +68,34 @@ def get_node_optimized(qs, lookup, info):
...
+
+def validate_slug_and_generate_if_needed(
+    instance: Type["Model"],
+    slugable_field: str,
+    cleaned_input: dict,
+    slug_field_name: str = "slug",
+) -> dict:
+    """Validate slug from input and generate in create mutation if is not given."""
+    # update mutation - just check if slug value is not empty
+    # we use _state.adding to checking if instance is saved because if pk has default
```
fowczarek: _state.adding is True only when it's new not save instance.

PR #5206, opened 2020-01-28
Update payment information before processing payment

Boquete:
I want to merge this change because in authentication / processing payment, order information is essential 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/mutations.py
+++ b/saleor/graphql/checkout/mutations.py
@@ -735,7 +735,18 @@ class CheckoutComplete(BaseMutation):
 
         if shipping_address is not None:
             shipping_address = AddressData(**shipping_address.as_data())
+            
+        # create the order into the database
+        order = create_order(
+            checkout=checkout,
+            order_data=order_data,
+            user=user,
+            redirect_url=redirect_url,
+        )
 
+        payment.order = order
+        payment.save(update_fields=["order"])
+        
         try:
             txn = gateway.process_payment(
                 payment=payment, token=payment.token, store_source=store_source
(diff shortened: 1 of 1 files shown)


patrys:
@Boquete Can you describe the use case where this becomes useful? We explicitly try to avoid placing orders without a valid payment method.

Boquete:
> @Boquete Can you describe the use case where this becomes useful? We explicitly try to avoid placing orders without a valid payment method.

I've seen recently issue reported that someone needed it in some payment integration gateway.

In our case, which is PayU, we need order details to authorize payment, in which redirect url is generated for payu site.

But you are right, placing order without valid payment method is not a great idea too. Not sure which one is worse, but placing order without valid payment method looks less scary, then not allowing other gateways to be implemented. Maybe there is some compromise (?)

Boquete:
> > @Boquete Can you describe the use case where this becomes useful? We explicitly try to avoid placing orders without a valid payment method.
> 
> I've seen recently issue reported that someone needed it in some payment integration gateway.
> 
> In our case, which is PayU, we need order details to authorize payment, in which redirect url is generated for payu site.
> 
> But you are right, placing order without valid payment method is not a great idea too. Not sure which one is worse, but placing order without valid payment method looks less scary, then not allowing other gateways to be implemented. Maybe there is some compromise (?)

Related issue : https://github.com/mirumee/saleor/issues/5193

patrys:
In the case where you need a redirect URL ahead of the time, you can use the Payment ID or the Checkout token in the return URL and then redirect to the appropriate order which should already exist by the time the redirect happens.

maarcingebala:
Thanks for opening the PR, but the issue should be solved in https://github.com/mirumee/saleor/issues/5193. Also, we don't want to change the flow where payment is required to create orders. As @patrys wrote, we don't want to have unpaid orders in the system.

PR #5209, opened 2020-01-29
Fix the way account confirm URL is being created

tomaszszymanski129:
I want to merge this change because it changes the way confirmation URL is being created.
(In addition to #1527)
should be merged along with https://github.com/mirumee/saleor-storefront/pull/586

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/mutations/account.py
+++ b/saleor/graphql/account/mutations/account.py
@@ -33,7 +33,7 @@ class AccountRegisterInput(graphene.InputObjectType):
         description=(
             "Base of frontend URL that will be needed to create confirmation URL."
         ),
-        required=settings.ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL,
+        required=False,
     )
 
 
@@ -73,6 +73,11 @@ class AccountRegister(ModelMutation):
         except ValidationError as error:
             raise ValidationError(
                 {"redirect_url": error}, code=AccountErrorCode.INVALID
+            )
+        except KeyError:
+            raise ValidationError(
+                {"redirect_url": "This field is required."},
+                code=AccountErrorCode.INVALID,
             )
         return super().clean_input(info, instance, data, input_cls=None)
 
--- a/saleor/graphql/account/mutations/account.py
+++ b/saleor/graphql/account/mutations/account.py
@@ -33,7 +33,7 @@ class AccountRegisterInput(graphene.InputObjectType):
         description=(
             "Base of frontend URL that will be needed to create confirmation URL."
         ),
-        required=True,
+        required=settings.ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL,
     )
 
 
@@ -65,6 +65,9 @@ class AccountRegister(ModelMutation):
 
     @classmethod
     def clean_input(cls, info, instance, data, input_cls=None):
+        if not settings.ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL:
+            return super().clean_input(info, instance, data, input_cls=None)
+
         try:
             validate_storefront_url(data["redirect_url"])
         except ValidationError as error:
--- a/saleor/graphql/account/mutations/account.py
+++ b/saleor/graphql/account/mutations/account.py
@@ -67,6 +67,11 @@ class AccountRegister(ModelMutation):
     def clean_input(cls, info, instance, data, input_cls=None):
         if not settings.ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL:
             return super().clean_input(info, instance, data, input_cls=None)
+        elif not data.get("redirect_url"):
+            raise ValidationError(
+                {"redirect_url": "This field is required."},
+                code=AccountErrorCode.INVALID,
+            )
 
         try:
             validate_storefront_url(data["redirect_url"])
@@ -74,11 +79,7 @@ class AccountRegister(ModelMutation):
             raise ValidationError(
                 {"redirect_url": error}, code=AccountErrorCode.INVALID
             )
-        except KeyError:
-            raise ValidationError(
-                {"redirect_url": "This field is required."},
-                code=AccountErrorCode.INVALID,
-            )
+
         return super().clean_input(info, instance, data, input_cls=None)
 
     @classmethod
--- a/tests/api/test_account.py
+++ b/tests/api/test_account.py
@@ -640,7 +640,21 @@ def test_customer_register_disabled_email_confirmation(
     send_account_confirmation_email_mock, user_api_client
 ):
     variables = {"email": "customer@example.com", "password": "Password"}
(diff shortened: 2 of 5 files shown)


django-queries:
Here is the report for 19825ec3d9d5d53063e1486c881a19b25a09a102 (mirumee/saleor @ fix/email_confirm_url_creation)
Base comparison is be18c5e049e67c89ff8a52f196f5a0b35521db26.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	          5	          5	              1

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         18	         18	              4
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         23	         23	              9

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

fowczarek:
@tomaszszymanski129  Can you change current solution to `prepare_url` in `_send_account_delete_confirmation_email_with_url` function? 
https://github.com/mirumee/saleor/pull/5209/files#diff-742fc523ca372ea4d4662056c57c6eefL112

fowczarek:
@maarcingebala After this, we should fix storefront ASAP.

fowczarek:
@tomaszszymanski129 We should fix one more issue related to account register mutation. 
Inside `AccountRegisterInput` we have required `redirect_url` field. This field shouldn't be required by schema because it depends on the value of `ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL`. When we don't need account confirmation this field shouldn't be required, but when we required confirmation this filed should be required by validation.

saleor/graphql/account/mutations/account.py
```diff
@@ -33,7 +33,7 @@ class AccountRegisterInput(graphene.InputObjectType):
         description=(
             "Base of frontend URL that will be needed to create confirmation URL."
         ),
-        required=True,
+        required=settings.ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL,
```
maarcingebala: I think schema.graphql should be always static, e.g. changing env variable `ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL` shouldn't require re-generating it. Maybe we could just change it to `required=False` and have a validation rule in `clean_input` of that mutation to raise an error if this setting is enabled and the value was not provided.

maarcingebala: Now I see that we have changes in `clean_input` as well - maybe it's enough; if `validate_storefront_url` returns a meaningful error when url is not provided then it's fine.

tests/api/test_account.py
```diff
@@ -577,29 +577,32 @@ def test_user_with_cancelled_fulfillments(
...
-                field
-                message
-            }
-            user {
-                id
-            }
-        }
-    }
-"""
+def _build_account_register_query(requires_redirect=True):
```
fowczarek: I'm not sure if it's a good solution. We should build a query without any conditions. Instead of condition, you can create mutation with optional `$redirectUrl: String`, and everything should work in both cases with `redirectUrl` inside query variables and without it.

saleor/graphql/account/mutations/account.py
```diff
@@ -65,12 +65,20 @@ def mutate(cls, root, info, **data):
...
+        if not settings.ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL:
+            return super().clean_input(info, instance, data, input_cls=None)
+
         try:
             validate_storefront_url(data["redirect_url"])
         except ValidationError as error:
             raise ValidationError(
                 {"redirect_url": error}, code=AccountErrorCode.INVALID
             )
+        except KeyError:
```
fowczarek: I don't see the test for lack of "redirect_url" when it's required.

fowczarek: I'm not sure but in all mutation, we check if `key` in `input_data` instead of catching `KeyError` and I prefer a solution with the condition. It's more clear to me.  

@maarcingebala  What do you think about it?

tomaszszymanski129: `try - except` in general have better performance than `if key in` so I'm just used to that syntax for quite a while, however, since `if key in` is the way we do things here maybe it's better to stay consistent.
@maarcingebala thoughts?

tomaszszymanski129: @fowczarek will add the missing test, good idea.

maarcingebala: I would say `if-else` is used more often in our codebase in cases like this. But both solutions are perfectly fine and readable in my opinion and I wouldn't bother with changing that.

tests/api/test_account.py
```diff
@@ -640,7 +640,21 @@ def test_customer_register_disabled_email_confirmation(
...
+    assert send_account_confirmation_email_mock.delay.call_count == 0
+
+
+@override_settings(ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL=True)
+@patch("saleor.account.emails._send_account_confirmation_email")
+def test_customer_register_no_redirect_url(
+    send_account_confirmation_email_mock, user_api_client
+):
+    variables = {"email": "customer@example.com", "password": "Password"}
+    response = user_api_client.post_graphql(ACCOUNT_REGISTER_MUTATION, variables)
```
fowczarek: We should use `api_client` fixture instead of `user_api_client`. 
`user_api_client` - request as logged-in user. 
`api_client` -  request as anonymous user. 
`staff_api_client` - request as staff user. 
`service_account_api_client` - request as service account.

tests/api/test_account.py
```diff
@@ -640,7 +640,21 @@ def test_customer_register_disabled_email_confirmation(
     send_account_confirmation_email_mock, user_api_client
 ):
     variables = {"email": "customer@example.com", "password": "Password"}
-    user_api_client.post_graphql(ACCOUNT_REGISTER_MUTATION, variables)
+    response = user_api_client.post_graphql(ACCOUNT_REGISTER_MUTATION, variables)
```
fowczarek: Same as below.

PR #5254, opened 2020-02-14
Enabled opentracing segmentation of spans

### LINKED ISSUE #5249: OpenTracing segments are orphans instead of children of the root span
Saleor is using `start_span` instead of `start_active_span` which is resulting to having a trace for each span instead of a segment, making the tracing close to useless.

On top, we should add a test(s) using MockTracer to ensure it does not break in the future.

NyanKiyoshi:
This adds segmentation to the OpenTracing spans, so each child span has its parent correctly associated to it.

On top, I added a change that was in my logs: `PLAYGROUND_ENABLED` setting key, which was discussed some time ago that it would be nice to be able to use the playground in production mode.

Closes #5249.

### Pull Request Checklist

1. [ ] ~~Changes are mentioned in the changelog.~~ feature not released yet

### DIFF
--- a/saleor/graphql/middleware.py
+++ b/saleor/graphql/middleware.py
@@ -1,7 +1,7 @@
 from typing import Optional
 
 import opentracing as ot
-import opentracing.tags
+import opentracing.tags as ot_tags
 from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 from django.utils.functional import SimpleLazyObject
@@ -51,7 +51,7 @@ class OpentracingGrapheneMiddleware:
             operation_name=info.operation.operation
         ) as scope:
             span = scope.span
-            span.set_tag(ot.tags.COMPONENT, "graphql")
+            span.set_tag(ot_tags.COMPONENT, "graphql")
             span.set_tag("graphql.parent_type", info.parent_type.name)
             span.set_tag("graphql.field_name", info.field_name)
             return next(root, info, **kwargs)
--- a/saleor/graphql/middleware.py
+++ b/saleor/graphql/middleware.py
@@ -47,9 +47,10 @@ def should_trace(info):
 class OpentracingGrapheneMiddleware:
     @staticmethod
     def resolve(next, root, info, **kwargs):
-        with ot.global_tracer().start_active_span(
-            operation_name=info.operation.operation
-        ) as scope:
+        operation = (
+            f"{info.operation.operation}.{info.parent_type.name}.{info.field_name}"
+        )
+        with ot.global_tracer().start_active_span(operation_name=operation) as scope:
             span = scope.span
             span.set_tag(ot_tags.COMPONENT, "graphql")
             span.set_tag("graphql.parent_type", info.parent_type.name)
--- a/saleor/graphql/middleware.py
+++ b/saleor/graphql/middleware.py
@@ -1,6 +1,7 @@
 from typing import Optional
 
-import opentracing
+import opentracing as ot
+import opentracing.tags
 from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 from django.utils.functional import SimpleLazyObject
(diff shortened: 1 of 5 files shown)


saleor/graphql/views.py
```diff
@@ -29,18 +31,11 @@
...
-        try:
-            result = execute(sql, params, many, context)
-        except Exception as e:
-            span.set_tag("error", True)
-            span.set_tag("error.object", e)
-            raise
-        else:
-            span.set_tag("error", False)
-            return result
+    with ot.global_tracer().start_active_span(operation_name="query") as scope:
```
NyanKiyoshi: By the way, this change removes `error` but it is normal. OpenTracing is already handling this behind the scene.

django-queries:
Here is the report for efbebcc10aa24535496fd8a1630e273ffde82825 (NyanKiyoshi/saleor @ ot/add-segmentation)
Base comparison is 00284d0c65eb063c339fc69366bbbc21ac3955d3.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	          5	          5	              1

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         18	         18	              4
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         23	         23	              9

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/graphql/views.py
```diff
@@ -80,7 +75,7 @@ def __init__(
     def dispatch(self, request, *args, **kwargs):
         # Handle options method the GraphQlView restricts it.
         if request.method == "GET":
-            if settings.DEBUG:
+            if settings.PLAYGROUND_ENABLED:
```
maarcingebala: We agreed with @patrys and @cmiacz that Playground can be served always by default, so I wouldn't introduce another variable for that. We can also accept this change but it will be removed soon anyway.

NyanKiyoshi: Then instead of `PLAYGROUND_ENABLED = getbool(..., DEBUG)` we could set to `True` by default? I prefer the idea of letting the user decide whether or not to enable it.

maarcingebala: Yes, if we're adding it then it should be set to `True`. There is no much difference whether you enable or disable it, anyone can access your API from their standalone playground or any other client.

tests/api/test_view.py
```diff
@@ -52,16 +52,14 @@ def test_batch_queries(category, product, api_client):
     assert data["category"]["name"] == category.name
 
 
-def test_graphql_view_get_in_non_debug_mode(client):
+def test_graphql_view_get_enabled_or_disabled(client, settings):
```
tomaszszymanski129: I usually prefer to keep each case as separate test so in case of failure I don't have to analyze the logic to see whats wrong. But since it's so trivial can be left as is I guess.

saleor/graphql/middleware.py
```diff
@@ -1,6 +1,7 @@
 from typing import Optional
 
-import opentracing
+import opentracing as ot
+import opentracing.tags
```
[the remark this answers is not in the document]
NyanKiyoshi: The import is used through `ot`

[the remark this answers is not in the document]
NyanKiyoshi: Ah I see the bot is saying it is not used... but it is. Then I will do `... import tags as ot_tags` less good looking but at least the robot will be happy

Issue #5174, opened 2020-01-17
CSV import/export


maarcingebala:
Implement CSV import/export for products.

## Export

Proposed API:
```
type Mutation {
  exportProducts(scope: ExportScope, filter: ProductFilterInput, ids: [ID!]): Job!
}

enum ExportScope {
  ALL
  IDS
  FILTER
}

type Job {
  id: ID!
  status: String!
  createdAt: DateTime!
  url: String
}
```

### Description
Mutation triggers a Celery task that takes given queryset of products and exports them to a file. As a result of the mutation, we return a `Job` object that informs about the status of the operation. Once the task is finished, `Job` will contain a link to download the CSV file (or a dashboard view with the report and download link). We will send this link in an email to the user who requested the operation.

Job objects could be used in the dashboard to render a list of scheduled jobs, query their status or access past exports, but we'll need to agree if this in the scope of this PR.

Parms:
My idea is to use `scope`, `filter` and `ids` to determine which products to export. `filter` would be of the same shape as filter used in the `products` query, so we can filter out some products first on the list and then export them. `ids` is to allow exporting only selected set of product `ids`. There is also the ability to export all products in the database. The `scope` param would determine which of these three export options to use.

Permissions: `MANAGE_PRODUCTS` (or specific permission only for import/export - needs to be discussed)

## Import

```
type Mutation {
  importProducts(file: Upload!, updateExisting: Boolean!, updateStock: Boolean!): Job!
}
```

### Description

Mutation passes the uploaded file to a task that reads and parses the file. For each row, we create necessary model instances, create connections between them and save them in the database. Once the task is finished we send an email to the staff user who requested the operation with a report about the number of successful/failed rows and/or errors that occurred.

Params:
- `updateExisting` - if true and there is a product/variant ID in CSV that already exists in the database, the import will override these products with data from CSV. If false, lines that refer to existing products will be skipped.
- `updateStock` - whether to update stock quantities with values from the CSV file. This is to guard against overriding the inventory accidentally with quantities that may be outdated in the file.

Permissions: same as `exportProducts` mutation.

## Notes

- Rows in the CSV file will represent either products or variants. We will assume that the first line (after the header line) will represent a product and the following lines will represent its variants until the importer recognizes the next product.
- Product images will be exported as URLs; when importing we also expect URLs where we'll download the images from.
[...]

PR #5255, opened 2020-02-14
Add products csv export

IKarbowiak:
Add products csv exports.

Impplements export part of #5174.

 #### :warning: Questions to consider:
- ~~How many products user can export at once?~~
~~1) without limits - we should probably consider saving to csv in parts.~~ - **Done**
~~2) with some limits - how many items?~~
- Maybe we should add some download limits for user per day or week.
- How long csv files should be kept?
- ~~Should we create user event for export csv?~~ **Done**

Right now, memory usage for 50 000, 100 000 and 200 000 is the same. About 200 Mb.

**Time-consuming and number of queries:**
For CSV file:
- 200 000 products, 30 seconds, 100 db queries
- 100 000 products, 18 seconds, 60 db queries
- 50 000 products, 10 seconds, 40 db queries
- 20 000 products, 8 seconds, 28 db queries

XLSX is slow. Append to file is really time-consuming, so I think, there is nothing we can do with that.
- 50 000 products, 10 minutes, 47 db queries

Current flow diagram:
image

image

### Pull Request Checklist


django-queries:
Here is the report for 03905e046c6f30270b3a5b13f39a31e676da7139 (mirumee/saleor @ add-products-csv-export)
Base comparison is 73f40eb405709b2748bbe7714a5b173226d70533.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	          5	          5	              1

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         18	         18	              4
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         23	         23	              9

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/csv/emails.py
```diff
@@ -0,0 +1,24 @@
...
+from ..core.utils import build_absolute_uri
+
+if TYPE_CHECKING:
+    from .models import Job
+
+
+EXPORT_TEMPLATES = {"export_products": "csv/export_products_csv"}
+
+
+def send_email_with_link_to_download_csv(job: "Job", template_name: str):
```
NyanKiyoshi: Should be a celery task instead of being synchronous

saleor/csv/models.py
```diff
@@ -0,0 +1,14 @@
...
+
+from ..account.models import User
+from . import JobStatus
+
+
+class Job(models.Model):
+    status = models.CharField(
+        max_length=50, choices=JobStatus.CHOICES, default=JobStatus.PENDING
+    )
+    user = models.ForeignKey(User, related_name="jobs", on_delete=models.CASCADE)
```
NyanKiyoshi: `user` is not descriptive about what it is. `created_by` instead (I assume)?

saleor/csv/models.py
```diff
@@ -0,0 +1,14 @@
...
+    content_file = models.FileField(upload_to="csv_files", null=True)
+    created_at = models.DateTimeField(auto_now_add=True)
+    ended_at = models.DateTimeField(null=True)
```
NyanKiyoshi: It sounds weird to me. `completed_at` maybe?

saleor/csv/utils/export.py
```diff
@@ -0,0 +1,345 @@
...
+            "sku": "sku",
+            "price_override_amount": "price override",
+            "cost_price_amount": "cost price",
+            "variant_currency": "variant_currency",
+        },
+        "common": {"images": "images"},
+    }
+
+
+MEDIA_URL = settings.MEDIA_URL
```
NyanKiyoshi: Are you sure you need it here? I don't like it too much as it forces the lazy object to be resolved during import, which could potentially lead to unwanted side-effects

saleor/graphql/csv/mutations.py
```diff
@@ -0,0 +1,84 @@
...
+    ids = graphene.List(
+        graphene.NonNull(graphene.ID),
+        description="List of products IDS to export.",
+        required=False,
+    )
+
+
+class ExportProducts(BaseMutation):
+    job = graphene.Field(
+        Job, description="Tne newly created job which is responsible for export data."
```
NyanKiyoshi:
```suggestion
        Job, description="The newly created job which is responsible to export data."
```

saleor/graphql/csv/mutations.py
```diff
@@ -0,0 +1,84 @@
...
+                raise ValidationError(
+                    {
+                        "filter": ValidationError(
+                            "You must provide filter input.",
+                            code=CsvErrorCode.REQUIRED.value,
+                        )
+                    }
+                )
+            return {"filter": filter}
+        return {"all": ""}
```
NyanKiyoshi: Maybe we could split to reduce the complexity? `.clean_ids(...)`, `.clean_filter(...)`?

saleor/graphql/csv/types.py
```diff
@@ -0,0 +1,39 @@
...
+    status = JobStatusEnum(description="Job status.")
+
+    class Meta:
+        description = "Represents job data"
+        interfaces = [relay.Node]
+        model = models.Job
+        only_fields = ["id", "created_at", "ended_at", "status", "url", "user"]
+
+    @staticmethod
+    def resolve_url(root: models.Job, _info):
```
NyanKiyoshi: `_blah` means unused `blah`
```suggestion
    def resolve_url(root: models.Job, info):
```

saleor/graphql/csv/types.py
```diff
@@ -0,0 +1,39 @@
...
+        content_file = root.content_file
+        if not content_file:
+            return None
+        return _info.context.build_absolute_uri(content_file.url)
```
NyanKiyoshi:
```suggestion
        return info.context.build_absolute_uri(content_file.url)
```

templates/templated_email/csv/export_products_csv.email
```diff
@@ -0,0 +1,19 @@
...
+
+{% block subject %}
+  {% trans "Export products data" context "Export products e-mail subject" %}
+{% endblock %}
+
+{% block plain %}
+{% include 'templated_email/shared/_header.email' %}
+
+{% blocktrans context "Export products e-mail text" %}
+Your csv file with products data is ready to download: {{ csv_link }}
```
NyanKiyoshi:
```suggestion
Your CSV file with products data is ready to download: {{ csv_link }}
```

templates/templated_email/source/export_products_csv.mjml
```diff
@@ -0,0 +1,24 @@
...
+    </mj-raw>
+    <mj-include path="./shared/header.mjml" />
+    <mj-section>
+      <mj-column>
+        <mj-text font-size="16px">
+          {% trans "Hi!" context "Standard e-mail greeting" %}
+        </mj-text>
+        <mj-text>
+        {% blocktrans context "Export products e-mail text" %}
+            Your csv file with products data is ready to download: <a href="{{ csv_link }}">{{ csv_link }}</a>
```
NyanKiyoshi:
```suggestion
            Your CSV file with products data is ready to download: <a href="{{ csv_link }}">{{ csv_link }}</a>
```

tests/utils.py
```diff
@@ -61,3 +63,8 @@ def generate_attribute_map(obj: Union[Product, ProductVariant]) -> Dict[int, Set
         assignment.attribute.pk: {value.pk for value in assignment.values.all()}
         for assignment in qs
     }
+
+
+def clear_temporary_dir(upload_dir):
+    temp_dir = f"{tempfile.gettempdir()}/{upload_dir}"
+    shutil.rmtree(temp_dir)
```
NyanKiyoshi: There is a fixture for such case. Here is an example of mine on how to use it: https://github.com/NyanKiyoshi/pytest-django-queries/blob/47a1325c811de4f3d5cf0c032b36cd5994e1929e/tests/test_plugin.py#L44-L64

saleor/csv/utils/export.py
```diff
@@ -0,0 +1,345 @@
+import os
```
NyanKiyoshi: Whoever: make sure to document this, this will save a lot of time when doing the reverse: importing from Saleor CSV.

saleor/csv/utils/export.py
IKarbowiak: No, I don't need it here. I'll remove this global variable.

saleor/csv/models.py
IKarbowiak: Yeah, `completed_at` sounds better.

tests/utils.py
IKarbowiak: It seems that I still need te override `MEDIA_ROOT`. The problem here is that `save_csv_file_in_job` saves CSV file to job and without proper set of `MEDIA_ROOT` test failed.  
Current solution creates new file every time test is running, so because of that cleaning tempdir is necessary. But I can use `tmpdir`as a fixture, override `settings.MEDIA_ROOT` in test and in this case every time new `temp_directory` will be creating, but only 3 last directory are kept. However it will still creats something on your disk and not delete it.

We can also mock save to Django Storage but in this case we can't check if csv file is saved correctly.

NyanKiyoshi: If I understand correctly, your issue/worry is that pytest would not delete the temporary files? I looked into (was too lazy to try locally 😉 ), and found this:

> Temporary directories are by default created as sub-directories of the system temporary directory. The base name will be pytest-NUM where NUM will be incremented with each test run. Moreover,**entries older than 3 temporary directories will be removed.** - https://docs.pytest.org/en/latest/tmpdir.html#base-temporary-directory

Does it help?

CHANGELOG.md
```diff
@@ -67,8 +67,9 @@ All notable, unreleased changes to this project will be documented in this file.
 - Return `null` when querying `me` as an anonymous user - #5231 by @maarcingebala
 - Added `PLAYGROUND_ENABLED` environment variable/setting to allow to enable the GraphQL playground when `DEBUG` is disabled - #5254 by @NyanKiyoshi
 - Fix access to order query when request from service account - #5258 by @fowczarek
-- Customer shouldn't be able to see draft orders by token - #5259 by @fowczarek
+- Customer shouldn't be able to see draft orders by token  - #5259 by @fowczarek
```
NyanKiyoshi:
```suggestion
- Customer shouldn't be able to see draft orders by token - #5259 by @fowczarek
```

saleor/graphql/csv/mutations.py
```diff
@@ -0,0 +1,92 @@
...
+        if not ids:
+            raise ValidationError(
+                {
+                    "ids": ValidationError(
+                        "You must provide at least one product id.",
+                        code=CsvErrorCode.REQUIRED.value,
+                    )
+                }
+            )
+        _, pks = _resolve_nodes(ids, graphene_type=Product)
```
NyanKiyoshi: Maybe we should check if there is a reason why this is private? Dangerous to run alone? Or fine to publicly? If that's fine to run publicly, maybe we should make it public–for now, it is private meaning this wasn't designed to be called outside than the file itself.

tests/utils.py
IKarbowiak: Yes, that's true. But it's not so easy case.
I want to test if the proper data in proper format are saved in csv file. So I provide data and file name to function, which creates temporary file and saved it into the job instace. Then I read the file from job instance and checked it's content. 
So I can't use file as a fixture because it's generated in the function. With overriding `MEDIA_ROOT` I ensure that this file will be saved in specific directory (what's more it doesn't work without that ). So it would be best if the directory with this new files will be removed.  

Maybe not so bad is using `tmpdir` as `MEDIA_ROOT` which kept only 3 last directories (https://docs.pytest.org/en/latest/tmpdir.html#the-default-base-temporary-directory) if we don't want to use current solution.

saleor/graphql/csv/mutations.py
IKarbowiak: I don't see the reason why it's private. As far as I can see you are the person who refactored `get_nodes` and created this private method. So it seems you are the right person to give the answer 😄

NyanKiyoshi: Yes I know I am the author but I don't remember the exact reasons 👀 (git blame strikes)

By the looks of it, I didn't want to give direct access to the raw information, and prevent human-errors while handling the results and raising errors. To prevent:
- Disclosing internal IDs (notice `get_nodes` throws the base64 IDs and not the primary keys, as we don't know what will those contain in the future)
- Mixed error raising/formatting/...
- Prevent mixed logic
- Prevent accidental usage of unwanted complex "utilities"/logics

What is the reason behind you not wanting it to run `get_nodes` as is? It seems fine for your case? Am I missing something?

tests/utils.py
NyanKiyoshi: I don't understand why you don't want to set `MEDIA_ROOT` to the `str()` of the path given by pytest's fixture?

saleor/csv/utils/export.py
maarcingebala: In the future, we want to also support the export of orders and customers. We should think about how to organize the logic in this file, so we have some abstraction for the common machinery that will be used in all cases. Maybe it's too early for this but we should keep it in mind.

saleor/graphql/csv/enums.py
```diff
@@ -0,0 +1,13 @@
+import graphene
+
+from ...csv import JobStatus
+from ...graphql.core.enums import to_enum
+
+
+class ExportScope(graphene.Enum):
```
maarcingebala: We should provide descriptions for these enums so that it's clear in the API what's the difference between these three export modes.

saleor/graphql/csv/types.py
```diff
@@ -0,0 +1,46 @@
...
+from graphene_federation import key
+from graphql_jwt.exceptions import PermissionDenied
+
+from ...core.permissions import AccountPermissions
+from ...csv import models
+from ..core.connection import CountableDjangoObjectType
+from .enums import JobStatusEnum
+
+
+@key(fields="id")
```
maarcingebala: Not sure if this is needed for this type. @salwator I know it's related to Federation, do you think we need to support this for the export job type?

saleor/graphql/csv/types.py
```diff
@@ -0,0 +1,46 @@
...
+class Job(CountableDjangoObjectType):
+    url = graphene.String(description="The URL of field to download.")
+    status = JobStatusEnum(description="Job status.")
+
+    class Meta:
+        description = "Represents job data"
```
maarcingebala: A little bit more explicit:

```suggestion
        description = "Represents a data export job."
```

saleor/graphql/csv/mutations.py
IKarbowiak: I don't want to use `get_nodes` because it returns queryset and I don't need it there (I just need `pks` of products). It will generate additional useless database query.

saleor/graphql/csv/mutations.py
```diff
@@ -0,0 +1,92 @@
...
+        if not filter:
+            raise ValidationError(
+                {
+                    "filter": ValidationError(
+                        "You must provide filter input.",
+                        code=CsvErrorCode.REQUIRED.value,
+                    )
+                }
+            )
+        return {"filter": filter}
```
NyanKiyoshi: Can you add type hinting for the return values? When I read `perform_mutation` and see `get_products_scope`, I'm not instantly sure what it could be returning

saleor/graphql/csv/mutations.py
NyanKiyoshi: Okay, I see then. Then, refactor `_resolve_nodes` to `resolve_global_ids_to_primary_keys`

saleor/graphql/csv/types.py
IKarbowiak: I removed that.

saleor/csv/utils/export.py
```diff
@@ -0,0 +1,95 @@
...
+    headers: List[str],
+    csv_headers_mapping: Dict[str, str],
+    delimiter: str,
+    job: Job,
+    file_name: str,
+):
+    table = etl.fromdicts(export_data, header=headers, missing=" ")
+    table = etl.rename(table, csv_headers_mapping)
+
+    temporary_file = NamedTemporaryFile()
```
korycins: `with NamedTemporaryFile as temporary_file:` ? 
then you don't need to close file manually

tests/conftest.py
```diff
@@ -1910,3 +1913,52 @@ def stock(variant, warehouse):
...
+    new_created_dates = [
+        created_date,
+        created_date,
+        created_date + datetime.timedelta(hours=2),
+        created_date - datetime.timedelta(days=2),
+        created_date - datetime.timedelta(days=5),
+    ]
+    for counter, job in enumerate(job_list):
+        job.created_at = new_created_dates[counter]
+        job.save()
```
korycins: https://docs.djangoproject.com/en/3.0/ref/models/querysets/#bulk-update

saleor/csv/emails.py
```diff
@@ -0,0 +1,26 @@
...
+if TYPE_CHECKING:
+    from .models import Job
+
+
+EXPORT_TEMPLATES = {"export_products": "csv/export_products_csv"}
+
+
+@app.task
+def send_email_with_link_to_download_csv(job: "Job", template_name: str):
+    recipient_email = job.created_by.email
```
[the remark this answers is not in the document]
maarcingebala: I think it's fine for the MVP.

saleor/graphql/csv/mutations.py
```diff
@@ -0,0 +1,141 @@
...
+        error_type_field = "csv_errors"
+
+    @classmethod
+    def perform_mutation(cls, root, info, **data):
+        user = info.context.user
+        input = data["input"]
+        scope = cls.get_products_scope(input)
+        export_info = cls.get_export_info(input["export_info"])
+        export_file = csv_models.ExportFile.objects.create(created_by=user)
+        export_started_event(export_file=export_file, user=user)
```
korycins, line 80: Are you aware that if someone will try to run this mutation with App token, this will raise an unhandled exception?

tests/test_emails.py
```diff
@@ -359,3 +363,48 @@ def test_send_email_changed_notification(
...
     email_connection.get_email_message(to=recipients, **expected_call_kwargs)
+
+
+@mock.patch("saleor.csv.emails.send_templated_mail")
+def test_send_email_with_link_to_download_csv(
+    mocked_templated_email, site_settings, export_file, tmpdir
+):
+    from django.conf import settings
+
+    settings.MEDIA_ROOT = tmpdir
```
korycins, line 374: there is a fixture media_root

tests/test_emails.py
```diff
@@ -359,3 +363,48 @@ def test_send_email_changed_notification(
     # Render the email to ensure there is no error
     email_connection = get_connection()
     email_connection.get_email_message(to=recipients, **expected_call_kwargs)
+
+
+@mock.patch("saleor.csv.emails.send_templated_mail")
+def test_send_email_with_link_to_download_csv(
+    mocked_templated_email, site_settings, export_file, tmpdir
+):
+    from django.conf import settings
```
korycins, line 372: pytest-django gives you a fixture `settings`. Why do you import it directly?

poetry.lock
```diff
@@ -148,10 +148,10 @@ description = "The AWS SDK for Python"
 name = "boto3"
 optional = false
 python-versions = "*"
-version = "1.13.19"
+version = "1.13.15"
```
fowczarek, line 151: Downgrade here. Please check all versions of the packages.

fowczarek:
Maby we should have 1 CSV migration instead of 4?

poetry.lock
IKarbowiak: Done

IKarbowiak:
> Maby we should have 1 CSV migration instead of 4?

Yes, thanks @fowczarek. Done.

saleor/csv/tests/test_products_data.py
```diff
@@ -0,0 +1,1096 @@
...
+    warehouse_headers = [f"{w.slug} (warehouse quantity)" for w in warehouses]
+    excepted_headers = (
+        expected_fields + product_headers + variant_headers + warehouse_headers
+    )
+
+    expected_file_headers += product_headers + variant_headers + warehouse_headers
+
+    assert expected_file_headers == file_headers
+    assert set(export_fields) == set(expected_fields)
+    assert data_headers == excepted_headers
```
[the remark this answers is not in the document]
korycins: Also, you could create an `export` directory as we are going to implement `import` also. IMHO - It would be good to separate them now. What do you think?

saleor/csv/utils/export.py
```diff
@@ -0,0 +1,156 @@
...
+        data_headers,
+        delimiter,
+        export_file,
+        file_type,
+    )
+
+
+def get_filename(model_name: str, file_type: str) -> str:
+    return "{}_data_{}.{}".format(
+        model_name, timezone.now().strftime("%d_%m_%Y"), file_type
```
korycins, line 50: Are you sure that `"%d_%m_%Y"` will be enough? What about the case when I want to export two lists of products on the same day? This method will return the same path to file and will overwrite the previous one. Am I right? Is it expected behavior?

IKarbowiak: No, files won't be overridden. Django always creates unique names like `product_data_01_06_2020_Y9pH4kO.csv`

saleor/csv/tests/test_export.py
```diff
@@ -0,0 +1,403 @@
...
+
+@patch("saleor.csv.utils.export.BATCH_SIZE", 1)
+@patch("saleor.csv.utils.export.send_email_with_link_to_download_file")
+def test_export_products_in_batches_for_csv(
+    save_csv_file_in_export_file_mock, product_list, export_file, tmpdir, media_root
+):
+    # given
+    qs = Product.objects.all()
+    export_info = {
+        "fields": [ProductFieldEnum.NAME.value, ProductFieldEnum.VARIANT_SKU.value],
```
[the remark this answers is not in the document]
IKarbowiak: This test checks if export in batches works correctly, tests with all the fields you can find in `saleor/saleor/csv/tests/test_products_data.py`, test: `test_get_products_data`

saleor/csv/tests/test_export.py
```diff
@@ -0,0 +1,403 @@
...
+@patch("saleor.csv.utils.export.BATCH_SIZE", 1)
+@patch("saleor.csv.utils.export.send_email_with_link_to_download_file")
+def test_export_products_in_batches_for_xlsx(
+    save_csv_file_in_export_file_mock, product_list, export_file, tmpdir, media_root
+):
+    # given
+    qs = Product.objects.all()
+    export_info = {
+        "fields": [ProductFieldEnum.NAME.value, ProductFieldEnum.VARIANT_SKU.value],
+        "warehouses": [],
```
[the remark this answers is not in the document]
IKarbowiak: Again, this testing export in batches for XLSX. But true, maybe I should add a test where attributes and warehouses are provided to check if it doesn't fail.

saleor/csv/tests/test_products_data.py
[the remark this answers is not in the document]
IKarbowiak: I can create an export directory, but I think it doesn't matter if I do it now or later.

What about file size, I was wondering about splitting it, but this is one file, so there should be a separate file for every function? I am not sure if it's a good idea.

saleor/csv/tests/test_products_data.py
```diff
@@ -0,0 +1,1096 @@
...
+                    f"{slug} (warehouse quantity)",
+                ]
+                data[warehouse_headers[0]] = stock.quantity
+
+            assigned_attribute = variant.attributes.first()
+            if assigned_attribute:
+                header = f"{assigned_attribute.attribute.slug} (variant attribute)"
+                data[header] = assigned_attribute.values.first().slug
+
+            expected_data.append(data)
```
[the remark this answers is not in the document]
IKarbowiak: It's different depends on the given fields. To make it reusable there will be lots of parameters, I think.

saleor/csv/tests/test_products_data.py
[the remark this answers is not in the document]
korycins: - When we will introduce tests for import I assume that we will try to separate them from the export. In that case, you will need to move all files. But it was only an idea so the decision is your ;)

- Based on experience from tests for mutation/queries and splitting the files per:  mutation/query I would say that smaller scope of the test file is better than the huge one.
cc @fowczarek  What do you think?

saleor/csv/tests/test_export.py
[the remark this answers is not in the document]
korycins: > Again, this testing export in batches for XLSX. 
The question was generic, not connected directly with this file :P

saleor/csv/tests/test_products_data.py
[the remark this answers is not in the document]
IKarbowiak: I agree that splitting files to smaller scope increases test readability.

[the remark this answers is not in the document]
fowczarek: In my opinion, working on small files is a lot easier than one big file.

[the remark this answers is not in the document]
IKarbowiak: I create PR for that:
#5751

saleor/csv/tests/test_export.py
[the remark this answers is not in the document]
IKarbowiak: So the answer is yes, in `saleor/csv/tests/export/products_data/test_get_products_data.py`.

saleor/csv/tests/test_products_data.py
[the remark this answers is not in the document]
IKarbowiak: I added some utils functions to simplify a little preparing expected data.

saleor/graphql/schema.graphql
```diff
@@ -1457,6 +1457,18 @@ type CreditCard {
   expYear: Int!
 }
 
+type CsvError {
```
dominik-zeglen, line 1460: Why not just `ExportError` or `ProductExportError`?
We got `ExportEvent`, `ExportFile` and so on, and suddenly `CsvSomething` :)

IKarbowiak: True, `ExportError` will be better.

IKarbowiak: I'll change that.

saleor/graphql/schema.graphql
```diff
@@ -1751,6 +1763,96 @@ type Error {
...
+  errors: [Error!]! @deprecated(reason: "Use typed errors with error codes. This field will be removed after 2020-07-31.")
+  exportFile: ExportFile
+  csvErrors: [CsvError!]!
+}
+
+input ExportProductsInput {
+  scope: ExportScope!
+  filter: ProductFilterInput
+  ids: [ID!]
+  exportInfo: ExportInfoInput!
```
dominik-zeglen, line 1841: None of the fields in `ExportInfoInput` is required, so I's not necessary to required it at all 🤷‍♂️

saleor/graphql/schema.graphql
IKarbowiak: I remembered why I left here `CsvError`, because `csv` corresponds to our python module, where the export is, so it will be generic for export and import. Unless we want to have separate ones for export and import. @maarcingebala

maarcingebala: I think in the future we may need a separate type for import so `ImportError` and `ImportErrorCode`, as there may be a different set of codes there. Here I would use `ExportError` with `ExportErrorCode`. The type of export file (csv or xlsx) shouldn't matter for the error type.

locale/en/LC_MESSAGES/django.po
```diff
@@ -501,6 +525,26 @@ msgid ""
...
 msgstr ""
 
+#: templates/templated_email/source/export_failed.mjml:16
+msgctxt "Export products e-mail text"
+msgid ""
+"\n"
+"            Sorry, we couldn't finish exporting products, some unexpected "
+"errors occurred.\n"
+"            Please try again.\n"
+"        "
```
maarcingebala, line 535: We shouldn't have newline characters and empty spaces here, because it's harder to provide translations then. All `blocktrans` tags should also have the `trimmed` parameter. See what [the docs](https://docs.djangoproject.com/en/3.0/topics/i18n/translation/) say about it and how we use blocktrans in other email templates.

PR #5256, opened 2020-02-14
Move demo-specific changes to master

### LINKED ISSUE #4971: Adding a command to clear the database
One would want to keep their database structure but be able to clear the database from all the data they have (testings, populatedb(s), etc.).

We would like to have `./manage cleardb [--delete-staff]`

Default behaviors:
- Deletes everything except: `sites`, `site_settings` and staff users
- `--delete-staff` flag to still delete staff users, excluding the superuser
- We **do not** want to reset the tables sequences

### LINKED ISSUE #5170: Move demo specific changes to master
Deployment of the demo store requires us to rebase the `demo` branch every time we want to push a new version which is problematic. There are a few customizations on `demo` branch, but we could actually move them to `master` and enable the "demo" mode with a flag in settings (possibly configured as environment variable).

Todo:
- move demo-specific code to `master`
- add an environment variable to control whether the "demo" mode is turned on
- enable GraphQL Playground in `DEBUG=False` mode.
- make sure that we have tests that check that in the `demo` mode we actually anonymize checkout data or block admin mutations

maarcingebala:
Fixes #5170
Fixes #4971

Moves changes from `demo` branch to `master` so that we no longer have to maintain this branch. Changes:
- added `settings_demo.py` that enables some of the demo-specific configurations
- `populatedb` generates Braintree configuration when proper environment variables are set
- rendering example GraphQL query in Playground in demo mode (but simpler - it doesn't require a database call now as it used to)
- added `cleardb` command
- fixed `populatedb` command that was randomly failing when trying to create an order for a staff user without any address assigned


django-queries:
Here is the report for c66097acd11862a71a363005681666170b372a3d (maarcingebala/saleor @ move-demo-changes)
Base comparison is 959149677d5f7efd6665e6f377b1be6a3285bca4.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	          5	          5	              1

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         18	         18	              4
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         23	         23	              9

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

templates/graphql/playground.html
```diff
@@ -46,7 +46,11 @@
       GraphQLPlayground.init(document.getElementById('root'), {
         settings: {
           'schema.polling.enable': false
-        }
+        },
+        tabs: [{
+          endpoint: '{{ api_url }}',
+          query: `{{ query|safe }}`
+        }]
```
NyanKiyoshi: Could we try to fix the issue where this would override the tabs from local storage? The feature of saving tabs is really nice and useful

maarcingebala: Good point. According to Playground's [readme](https://github.com/prisma-labs/graphql-playground) that's not possible:

> Note: When using this feature, tabs will be resetted each time the page is reloaded

I'll update the template to load this query only in demo mode then.

saleor/core/management/commands/cleardb.py
```diff
@@ -0,0 +1,76 @@
...
+from saleor.account.models import User
+from saleor.checkout.models import Checkout
+from saleor.discount.models import Sale, Voucher
+from saleor.order.models import Order
+from saleor.payment.models import Payment, Transaction
+from saleor.product.models import Product
+from saleor.shipping.models import ShippingMethod
+
+
+class Command(BaseCommand):
```
fowczarek: How about delete:
1. ServiceAccounts
1. GiftCards
1. Categories
1. Collections
1. ProductTypes
1. Attributes
1. ShippingZones
1. Warehouses
1. Webhooks

saleor/core/utils/random_data.py
```diff
@@ -505,14 +508,14 @@ def create_fulfillments(order):
 
 
 def create_fake_order(discounts, max_order_lines=5):
-    user = random.choice(
-        [None, User.objects.filter(is_superuser=False).order_by("?").first()]
-    )
-    if user:
-        address = user.default_shipping_address
+    customers = User.objects.filter(is_superuser=False, is_staff=False).order_by("?")
```
fowczarek: Maybe we should have a few orders for staff users(Not for superuser) in random data?

saleor/core/management/commands/cleardb.py
maarcingebala: The idea behind the original command that I moved from `demo` branch was to remove data that people create in our demo instance - checkouts, orders, payments, changes in product stock. Then you can easily run the command periodically and repopulate database with fresh data.

saleor/core/utils/random_data.py
maarcingebala: Maybe yes, but not sure if this brings much value to the demo shop?

fowczarek: This function is used in `populatedb` on local instances of Saleor. We should have orders for staff users for the purpose of testing Saleor. It's allowed to find edge cases in the development process.

saleor/core/management/commands/cleardb.py
fowczarek: In that case, we shouldn't close #4971

saleor/core/management/commands/populatedb.py
```diff
@@ -120,3 +122,10 @@ def handle(self, *args, **options):
 
         for msg in create_permission_groups():
             self.stdout.write(msg)
+
+        if is_demo_mode():
```
korycins: Do we really want to add a `demo things` inside the repo? I know that deploying new changes can be problematic but I believe that we can figure out some different approaches for that.

korycins: If you really want to keep that in main repository I would suggest creating a separate file for all `demo staff` in utils directory. Just let's keep them togheter

saleor/core/management/commands/cleardb.py
maarcingebala: After speaking with @NyanKiyoshi I'll fix this command to remove other models as well, except for service accounts, staff users, site settings and plugins configuration.

saleor/core/utils/random_data.py
maarcingebala: Good point, I'll change it then.

fowczarek:
@maarcingebala  Do you know why `LGTM analysis: JavaScript` failing?

maarcingebala:
@fowczarek No idea about failing JS.

saleor/core/management/commands/cleardb.py
```diff
@@ -0,0 +1,108 @@
...
+        Warehouse.objects.all().delete()
+
+        Page.objects.all().delete()
+        self.stdout.write("Removed pages")
+
+        Webhook.objects.all().delete()
+        self.stdout.write("Removed webhooks")
+
+        # Delete all users except for staff members.
+        staff = User.objects.filter(Q(is_staff=True) | Q(is_superuser=True))
```
korycins: The query `User.objects.filter(Q(is_staff=True) | Q(is_superuser=True))` will return the same results as `User.objects.filter(is_staff=True)`
```
In [20]: User.objects.filter(Q(is_staff=True) | Q(is_superuser=True))
Out[20]: <QuerySet [<User: michele.west@example.com>, <User: stacey.moore@example.com>, <User: alicia.clark@example.com>, <User: admin@example.com>, <User: james.smith@example.com>, <User: michelle.ross@example.com>]>
```

```
In [18]: User.objects.filter(Q(is_staff=True))
Out[18]: <QuerySet [<User: michele.west@example.com>, <User: stacey.moore@example.com>, <User: alicia.clark@example.com>, <User: admin@example.com>, <User: james.smith@example.com>, <User: michelle.ross@example.com>]>
```

saleor/demo/management/commands/populatedb.py
```diff
@@ -0,0 +1,43 @@
...
+
+class Command(PopulateDBCommand):
+    def handle(self, *args, **options):
+        super().handle(*args, **options)
+        is_configured = configure_braintree()
+        if is_configured:
+            self.stdout.write("Configured Braintree")
+        else:
+            self.stdout.write(
+                "Failed to configure Braintree. Check if proper environment variables "
```
korycins: This is only a suggestion but for someone who doesn't have any knowledge about populateDB or braintree, this error can be hard to understand. What do you think about adding some str to this msg like: `Check if Braintree variables: BRAINTREE_API_KEY, BRAINTREE_API_KEY, BRAINTREE_API_KEY are set`
Also, I wouldn't use the word `Failed` as this can trigger a lot of issues from the community. I would use something like - `Braintree env variables(BRAINTREE_API_KEY, BRAINTREE_API_KEY, BRAINTREE_API_KEY) not found, using DUMMY gateway as a defualt`

saleor/core/management/commands/cleardb.py
maarcingebala: In the case of our example data, yes, but does `is_superuser=True` imply `is_staff=True`? These are two separate fields in the database.

saleor/demo/management/commands/populatedb.py
maarcingebala: The community is not supposed to use this demo configuration. But I can names these vars explicitly if you like.

korycins: Ach, my bad. I thought that you changed  the standard `populatedb` command. Please, ignore this comment

saleor/core/management/commands/cleardb.py
korycins: Yes, you are right. I thought that there are some correlations between these fields.

saleor/demo/management/commands/populatedb.py
maarcingebala: Thank you, sir.

NyanKiyoshi:
From the documentation, it means it detected a JS or TS file but the logs says there are none. Sounds random and from what the docs are saying it wouldn't happen again and will not try later on to build JS/TS if there are no such files detected (you cannot disable languages).

saleor/core/management/commands/cleardb.py
```diff
@@ -0,0 +1,107 @@
+"""Clear the database preserving shop's configuration.
+
+This command clears the database from data such as orders, products or customer
+accounts. It doesn't remove shop's configuration, such as: staff accounts, service
+accounts, plugin configurations, site settings or navigation menus.
+"""
+
+from __future__ import unicode_literals
```
NyanKiyoshi: Seems like left-overs from Python 2 code

maarcingebala: Fixed

Issue #5274, opened 2020-02-19
Events don't take into consideration ServiceAccount


korycins:
All events from `order/events`  require log-in User but it doesn't give the possibility to pass ServiceAccount 
DoD:
- ForeignKey to ServiceAccount should be added to OrdedEvents
- Tests from test_order_events should also check events for SA
- All events method should take into account ServiceAccount also. 
- All methods/functions which call any event should have the possibility to pass ServiceAccount also
- Some events from `AccountEvents` will also require SA (example: `staff_user_assigned_name_to_a_customer_event`).

maarcingebala:
Customer and order events now supports boh user and app references.

PR #5275, opened 2020-02-19
Add validation if user is not anonymous before saving an order events

korycins:
There is an issue with the `OrderEvents`. They don't take into consideration that an event can be created by ServiceAccount, they expect `User` object and they don't check if the user is a real object. 
This PR adds only validation to check if the provided user is not anonymous.

### DIFF
--- a/saleor/order/events.py
+++ b/saleor/order/events.py
@@ -28,6 +28,10 @@ def _get_payment_data(amount: Optional[Decimal], payment: Payment) -> Dict:
     }
 
 
+def _user_is_valid(user: UserType) -> bool:
+    return user and not user.is_anonymous
+
+
 def email_sent_event(
     *,
     order: Order,
@@ -58,6 +62,8 @@ def email_resent_event(
 
 
 def draft_order_created_event(*, order: Order, user: UserType) -> OrderEvent:
+    if not _user_is_valid(user):
+        user = None
     return OrderEvent.objects.create(
         order=order, type=OrderEvents.DRAFT_CREATED, user=user
     )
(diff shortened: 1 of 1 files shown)


korycins:
@IKarbowiak @fowczarek @maarcingebala 
It is only a quick fix for failing events. The proper solution is described here - #5274, I didn't add any tests as it should be added in #5274 when proper handling a ServiceAccounts will be added. 
For now, we are not able to use API with SA for Orders.

django-queries:
Here is the report for c71842b47f861f7518f474a0ea7904a16f741a81 (mirumee/saleor @ fix/add_user_validation_for_order_event)
Base comparison is ee60c497c0dd41a590642e0f0f20e2ccb1e053d0.



<p>

```diff
# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	          5	          5	              1

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  retrieve main menu                         	          5	          5	              0
  retrieve product list                      	          4	          4	              0
  retrieve secondary menu                    	          5	          5	              0
  retrieve shop                              	          2	          2	              0

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         17	         17	              3
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         23	         23	              9

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

PR #5304, opened 2020-02-27
Initial work on extension manager

patrys:
This reworks how extensions are loaded and reduces the number of database queries to one per request.

### Pull Request Checklist

### DIFF
--- a/saleor/checkout/base_calculations.py
+++ b/saleor/checkout/base_calculations.py
@@ -4,16 +4,16 @@ It's recommended to use functions from calculations.py module to take in account
 manager.
 """
 
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, Iterable, List, Optional
 
 from prices import TaxedMoney
 
 from ..core.taxes import quantize_price, zero_taxed_money
+from ..discount import DiscountInfo
 
 if TYPE_CHECKING:
     # flake8: noqa
     from .models import Checkout, CheckoutLine
-    from ..discount.types import DiscountsListType
 
 
 def base_checkout_shipping_price(checkout: "Checkout") -> TaxedMoney:
@@ -44,9 +44,9 @@ def base_checkout_total(
 
 
 def base_checkout_line_total(
-    line: "CheckoutLine", discounts: "DiscountsListType" = None
+    line: "CheckoutLine", discounts: Optional[Iterable[DiscountInfo]] = None
 ) -> TaxedMoney:
     """Return the total price of this line."""
-    amount = line.quantity * line.variant.get_price(discounts)
+    amount = line.quantity * line.variant.get_price(discounts or [])
     price = quantize_price(amount, amount.currency)
     return TaxedMoney(net=price, gross=price)
--- a/saleor/graphql/extensions/mutations.py
+++ b/saleor/graphql/extensions/mutations.py
@@ -44,8 +44,10 @@ class PluginUpdate(BaseMutation):
     @classmethod
     def perform_mutation(cls, root, info, **data):
         plugin_id = data.get("id")
-        input = data.get("input")
-        instance = cls.get_node_or_error(info, plugin_id, only_type=Plugin)
+        data = data.get("input")
         manager = get_extensions_manager()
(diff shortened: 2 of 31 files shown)


saleor/graphql/extensions/mutations.py
```diff
@@ -44,8 +46,12 @@ class Meta:
...
         plugin_id = data.get("id")
-        input = data.get("input")
-        instance = cls.get_node_or_error(info, plugin_id, only_type=Plugin)
+        data = data.get("input")
         manager = get_extensions_manager()
-        instance = manager.save_plugin_configuration(instance.name, input)
+        plugin = manager.get_plugin(plugin_id)
+        if not plugin:
+            raise ValidationError(
+                {"id": ValidationError("Plugin doesn't exist", code="not_found")}
```
maarcingebala, line 54: Instead of hardcoded `not_found` we shuold use the code from enum, probably from `ExtensionsError` in this case.

saleor/graphql/extensions/resolvers.py
```diff
@@ -1,20 +1,56 @@
...
+    if sort_field == PluginSortField.IS_ACTIVE:
+
+        plugins = sorted(
+            manager.plugins,
+            key=lambda p: (not p.active if sort_reverse else p.active, p.PLUGIN_NAME),
+        )
+    else:
+        plugins = sorted(manager.plugins, key=lambda p: p.PLUGIN_NAME)
+        if sort_reverse:
+            plugins = reversed(plugins)
```
maarcingebala, line 38: IMO this could go to a function to make the resolver more readable. Here we could have then e.g. `sort_plugins(manager.plugins, sort_by)`.

saleor/graphql/extensions/types.py
```diff
@@ -42,22 +43,26 @@ class Meta:
...
 
     class Meta:
         description = "Plugin."
         model = models.PluginConfiguration
         interfaces = [graphene.relay.Node]
-        only_fields = ["name", "description", "active", "configuration"]
+        only_fields = ["id", "name", "description", "active", "configuration"]
+
+    def resolve_id(self: models.PluginConfiguration, _info):
+        return self.name
```
maarcingebala, line 56: Not sure if this is in the scope of this PR, but don't we want to have slugs for this purpose?

maarcingebala:
@patrys @bogdal Do we want to unify naming "plugins" vs. "extensions" in this PR as well or is it a separate task?

patrys:
I would do it in a separate PR to avoid scope creep. This one should be mergeable soon.

saleor/graphql/extensions/types.py
patrys: We should not depend on names being constant. I suggest introducing a Java-like reverse-domain identifier: `mirumee.avatax` or `saleor.tax.vatlayer`

saleor/checkout/base_calculations.py
```diff
@@ -4,16 +4,16 @@
 manager.
 """
 
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, Iterable, List, Optional
 
 from prices import TaxedMoney
 
 from ..core.taxes import quantize_price, zero_taxed_money
+from ..discount import DiscountInfo
```
korycins, line 12: I see that you moved most of the imports of  `DiscountInfo`  to above `TYPE_CHECKING`.
Do we want to do that? IMO the safest way is to reduce import dependencies to minimal.

maarcingebala:
Does this PR affect [this section in the docs](https://docs.saleor.io/docs/next/advanced/extensions/#base-plugin)? If so, it needs to be updated accordingly [in this PR](https://github.com/mirumee/saleor-docs/pull/112).

saleor/checkout/base_calculations.py
patrys: `TYPE_CHECKING` is an ugly workaround for dependency loops, not the best practice for type imports 😉

saleor/graphql/extensions/types.py
bogdal: As agreed, we will return to this in another PR.

PR #5310, opened 2020-03-02
Create invoice model

tomaszszymanski129:
I want to merge this change because it adds a model for invoice.

### Pull Request Checklist


django-queries:
Here is the report for 9b6962834536ddb157d3fe9625d39eb8383b46c5 (mirumee/saleor @ invoice-model)
Base comparison is ef7d090af614d2117972fba1b00d17ecf0fbd671.



<p>

```diff
# api.benchmark category
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  category view                              	         23	         23	              0

# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	          5	          5	              1

# api.benchmark collection
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  collection view                            	         21	         21	              0

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  featured products list                     	         20	         20	              0
  retrieve main menu                         	          8	          8	              0
  retrieve product list                      	          6	          6	              0
  retrieve secondary menu                    	          8	          8	              0
  retrieve shop                              	          2	          2	              0
  user checkout details                      	         53	         53	             28

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         17	         17	              3
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         51	         51	              3
  retrieve variant list                      	         23	         23	              9

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         21	         21	              0
```

</p>
</details>

saleor/order/models.py
```diff
@@ -500,3 +500,10 @@ class Meta:
 
     def __repr__(self):
         return f"{self.__class__.__name__}(type={self.type!r}, user={self.user!r})"
+
+
+class Invoice(models.Model):
+    order = models.ForeignKey(Order, on_delete=models.CASCADE)
```
tomaszszymanski129: should this be cascade?

maarcingebala: Good question. @korycins @patrys should we keep invoices in DB when you delete related orders? It could be useful to store them as historical data. It's not usual (and not possible through UI/API) to remove orders but still, maybe some integration could be able to do that at some point?

patrys: We could make this nullable but I don't think we want to go beyond that for now (eg. no new UI to show abandoned invoices). I think it would be better to anonymize and flag orders instead of actually deleting them as we're also deleting their event logs.

saleor/order/models.py
```diff
@@ -500,3 +500,10 @@ class Meta:
 
     def __repr__(self):
         return f"{self.__class__.__name__}(type={self.type!r}, user={self.user!r})"
+
+
+class Invoice(models.Model):
+    order = models.ForeignKey(Order, null=True, on_delete=models.CASCADE)
```
maarcingebala: So if we agree on `null=True` here let's make it `on_delete=models.SET_NULL`.

saleor/order/models.py
```diff
@@ -500,3 +500,10 @@ class Meta:
 
     def __repr__(self):
         return f"{self.__class__.__name__}(type={self.type!r}, user={self.user!r})"
+
+
+class Invoice(models.Model):
```
NyanKiyoshi: I don't know what are the plans for Invoicing, but please be aware that depending on countries, the voice identifier (ID) can have different regulations. In France we recommend using a prefix, e.g. `FR000001`, `SHOP000001`.

Only important rule for France is that they have to be continuous (growing) numbers, and must have no gaps, e.g. `FR202003041` is illegal.

Anyway apart the examples, to come back to the point, maybe we would want to allow meta-data and extensions for invoices so people can adapt if they have special or odd needs?

maarcingebala: Yes, we will have metadata for invoices but that's not in the scope of this task. To ensure correct IDs - it will be implemented at the plugin level, so you can have different plugins to handle different formats.

saleor/order/models.py
```diff
@@ -500,3 +500,10 @@ class Meta:
...
+    order = models.ForeignKey(Order, null=True, on_delete=models.SET_NULL)
+    number = models.CharField(max_length=64)
+    created = models.DateTimeField(auto_now_add=True)
+    url = models.URLField(max_length=256)
```
korycins: Are you sure that 256 is enough in any cases?

saleor/order/models.py
```diff
@@ -500,3 +500,10 @@ class Meta:
 
     def __repr__(self):
         return f"{self.__class__.__name__}(type={self.type!r}, user={self.user!r})"
+
+
+class Invoice(models.Model):
+    order = models.ForeignKey(Order, null=True, on_delete=models.SET_NULL)
+    number = models.CharField(max_length=64)
```
korycins: Are you sure that 64 is enough in any cases?

saleor/order/models.py
korycins: https://stackoverflow.com/a/417184

saleor/order/models.py
tomaszszymanski129: in Poland in my entire life i didn't see a number larger than 20. But since this is a multi country project we can bump it up to 128. That should be more than safe since i googled a bit and couldn't find informations about extra long invoice numbers, in countries i've seen it's kept short.

saleor/order/models.py
tomaszszymanski129: such a long url's are a common effect of bad design but since we don't have impact on what will be integrated with this, we can go with 2000 limit for safety. Though I cannot imagine URL such long unless you pass everything in query params which is a horrible idea ;0

saleor/order/models.py
NyanKiyoshi: I mean, I believe `999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999` (64) orders is enough.

saleor/order/models.py
maarcingebala: I would stick with `255` which we use as a max value for other URL fields (e.g. webhook URLs). We control our integrations and if one of them produces longer URLs than this could be changed.

patrys: We should support at least 2048 characters, we have no control over what external vendors want to put in their URLs. Look at what Facebook does to their image and sharing URLs for an example of what's out there.

PR #5353, opened 2020-03-12
Add schema for invoices

dominik-zeglen:
Schema proposal.

### DIFF
--- a/saleor/graphql/schema.graphql
+++ b/saleor/graphql/schema.graphql
@@ -2473,7 +2473,7 @@ type Mutation {
   permissionGroupUnassignUsers(id: ID!, users: [ID!]!): PermissionGroupUnassignUsers
   invoiceCreate(orderID: ID!, input: InvoiceInput): InvoiceCreate
   invoiceDelete(orderID: ID!, invoiceID: ID!): InvoiceDelete
-  invoiceUpdate(invoiceID: ID!, input: InvoiceInput): InvoiceUpdate
+  invoiceUpdate(id: ID!, input: InvoiceInput): InvoiceUpdate
   invoiceSend(id: ID!, email: String): InvoiceSend
   requestInvoiceCreate(orderID: ID!, input: RequestInvoiceCreateInput): RequestInvoiceCreate
   requestInvoiceDelete(orderID: ID!, invoiceID: ID!): RequestInvoiceDelete
--- a/saleor/graphql/schema.graphql
+++ b/saleor/graphql/schema.graphql
@@ -1831,9 +1831,73 @@ type Image {
   alt: String
 }
 
+type Invoice implements Node & ObjectWithMetadata {
+  id: ID!
+  created: DateTime!
+  number: String!
+  order: Order!
+  url: String
+  privateMetadata: [MetadataItem]!
+  metadata: [MetadataItem]!
+  privateMeta: [MetaStore]! @deprecated(reason: "DEPRECATED: Will be removed in Saleor 2.11. use the `privetaMetadata` field instead. ")
+  meta: [MetaStore]! @deprecated(reason: "DEPRECATED: Will be removed in Saleor 2.11. use the `metadata` field instead. ")
+}
+
+type InvoiceCreate {
+  errors: [Error!]!
+  invoice: Invoice
+  invoiceErrors: [InvoiceError!]!
+  order: Order
+}
+
+type InvoiceDelete {
+  errors: [Error!]!
+  invoiceErrors: [InvoiceError!]!
+  order: Order
+}
+
+type InvoiceError {
+  code: InvoiceErrorCode!
+  field: String
+  message: String
+}
+
+enum InvoiceErrorCode {
+  PLUGIN_NOT_ENABLED
+}
+
+type InvoiceUpdate {
+  errors: [Error!]!
+  invoice: Invoice
+  invoiceErrors: [InvoiceError!]!
+  order: Order
+}
+
+input InvoiceInput {
+  created: DateTime
+  number: String
+  url: String
+}
+
+type InvoiceSend {
+  errors: [Error!]!
+  invoiceErrors: [InvoiceError!]!
+}
+
 input IntRangeInput {
   gte: Int
   lte: Int
+}
+
+interface Job {
+  status: JobStatus
+}
+
+enum JobStatus {
+  PENDING_CREATION
+  OK
+  PENDING_DELETION
+  DELETED
 }
 
 scalar JSONString
@@ -2407,6 +2471,12 @@ type Mutation {
   permissionGroupDelete(id: ID!): PermissionGroupDelete
   permissionGroupAssignUsers(id: ID!, users: [ID!]!): PermissionGroupAssignUsers
   permissionGroupUnassignUsers(id: ID!, users: [ID!]!): PermissionGroupUnassignUsers
+  invoiceCreate(orderID: ID!, input: InvoiceInput): InvoiceCreate
+  invoiceDelete(orderID: ID!, invoiceID: ID!): InvoiceDelete
+  invoiceUpdate(invoiceID: ID!, input: InvoiceInput): InvoiceUpdate
+  invoiceSend(id: ID!, email: String): InvoiceSend
+  requestInvoiceCreate(orderID: ID!, input: RequestInvoiceCreateInput): RequestInvoiceCreate
+  requestInvoiceDelete(orderID: ID!): RequestInvoiceDelete
 }
 
 input NameTranslationInput {
@@ -2478,6 +2548,7 @@ type Order implements Node & ObjectWithMetadata {
   totalBalance: Money!
   userEmail: String
   isShippingRequired: Boolean!
+  invoices: [Invoice!]
 }
 
 enum OrderAction {
@@ -2594,6 +2665,7 @@ type OrderEvent implements Node {
   oversoldItems: [String]
   lines: [OrderEventOrderLineObject]
   fulfilledItems: [FulfillmentLine]
+  invoice: Invoice
 }
 
 type OrderEventCountableConnection {
@@ -2644,6 +2716,9 @@ enum OrderEventsEnum {
   TRACKING_UPDATED
   NOTE_ADDED
   OTHER
+  INVOICE_CREATED
+  INVOICE_DELETED
+  INVOICE_SENT
 }
 
 input OrderFilterInput {
@@ -3793,6 +3868,23 @@ type RequestEmailChange {
   errors: [Error!]!
   user: User
   accountErrors: [AccountError!]!
+}
+
+type RequestInvoiceCreate {
+  errors: [Error!]!
+  invoice: Invoice
+  invoiceErrors: [InvoiceError!]!
+  order: Order
+}
+
+input RequestInvoiceCreateInput {
+  number: String
+}
+
+type RequestInvoiceDelete {
+  errors: [Error!]!
+  invoiceErrors: [InvoiceError!]!
+  order: Order
 }
 
 type RequestPasswordReset {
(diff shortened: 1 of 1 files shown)


django-queries:
Here is the report for 9bdb7a894e7106cf82b49fcabedceeea84c2804f (mirumee/saleor @ schema/invoices)
Base comparison is a0b14fb206003f1a7e65d911882cbd74f97c9b5e.



<p>

```diff
# api.benchmark category
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  category view                              	         23	         23	              0

# api.benchmark checkout
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  add billing address to checkout            	         34	         34	             20
  add shipping to checkout                   	          7	          7	              0
  checkout payment charge                    	         10	         10	              0
  complete checkout                          	          8	          8	              0
  create checkout                            	          5	          5	              1

# api.benchmark collection
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  collection view                            	         21	         21	              0

# api.benchmark homepage
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  featured products list                     	         20	         20	              0
  retrieve main menu                         	          8	          8	              0
  retrieve product list                      	          6	          6	              0
  retrieve secondary menu                    	          8	          8	              0
  retrieve shop                              	          2	          2	              0
  user checkout details                      	         53	         53	             28

# api.benchmark product
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product details                            	         17	         17	              3
  retrieve product attributes                	          9	          9	              0

# api.benchmark variant
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variant bulk create                	         48	         48	              3
  retrieve variant list                      	         23	         23	              9

# api.benchmark variant stocks
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  product variants stocks create             	         19	         19	              4
  product variants stocks delete             	         16	         16	              4
  product variants stocks update             	         23	         23	              4

# api product sorting attributes
  test name                                  	left count 	right count	duplicate count
  -------------------------------------------	-----------	-----------	---------------
  sort product not having attribute data     	         20	         20	              0
```

</p>
</details>

saleor/graphql/schema.graphql
```diff
@@ -1831,11 +1831,75 @@ type Image {
...
 
+type Invoice implements Node & ObjectWithMetadata {
+  id: ID!
+  created: DateTime!
+  number: String!
+  order: Order!
+  url: String
+  privateMetadata: [MetadataItem]!
+  metadata: [MetadataItem]!
+  privateMeta: [MetaStore]! @deprecated(reason: "DEPRECATED: Will be removed in Saleor 2.11. use the `privetaMetadata` field instead. ")
```
maarcingebala: There is no point in adding deprecated fields; if possible we shouldn't do that, but I'm not sure if it will be possible in Graphene. We have `ObjectWithMetadata` that adds them. @tomaszszymanski129 When implementing this I'd check if we can somehow manually exclude these fields.

saleor/graphql/schema.graphql
```diff
@@ -1831,11 +1831,75 @@ type Image {
...
+  errors: [Error!]!
+  invoiceErrors: [InvoiceError!]!
+}
+
 input IntRangeInput {
   gte: Int
   lte: Int
 }
 
+interface Job {
```
maarcingebala: Where is this used? I guess it should be used by "request" mutations?

We should have the common `Job` interface that will be reused and specific job types for different actions. I 
 would suggest these fields for the interface:
- `id: ID!` - as we're going to store them in DB
- `status: JobStatus!`
- `createdAt: DateTime!`
- `updatedAt: DateTime!`

saleor/graphql/schema.graphql
```diff
@@ -1831,11 +1831,75 @@ type Image {
...
+  status: JobStatus
+}
+
+enum JobStatus {
+  PENDING_CREATION
+  OK
+  PENDING_DELETION
+  DELETED
```
IKarbowiak: For me, `SUCCESS ` sounds better than `OK`.  Also, we need status for failed, e.g. `FAILED`.

saleor/graphql/schema.graphql
```diff
@@ -1831,11 +1831,75 @@ type Image {
...
+  errors: [Error!]!
+  invoice: Invoice
+  invoiceErrors: [InvoiceError!]!
+  order: Order
+}
+
+input InvoiceInput {
+  created: DateTime
+  number: String
+  url: String
```
IKarbowiak: If invoices are generated by the backend, what is the reason for passing `url` in `input`?

korycins: We don't force you to generate invoice by the backend. Basic plugin will do that, but external services need to have the possibility to pass URL to invoice.

saleor/graphql/schema.graphql
```diff
@@ -2407,6 +2471,12 @@ type Mutation {
   permissionGroupDelete(id: ID!): PermissionGroupDelete
   permissionGroupAssignUsers(id: ID!, users: [ID!]!): PermissionGroupAssignUsers
   permissionGroupUnassignUsers(id: ID!, users: [ID!]!): PermissionGroupUnassignUsers
+  invoiceCreate(orderID: ID!, input: InvoiceInput): InvoiceCreate
+  invoiceDelete(orderID: ID!, invoiceID: ID!): InvoiceDelete
+  invoiceUpdate(invoiceID: ID!, input: InvoiceInput): InvoiceUpdate
+  invoiceSend(id: ID!, email: String): InvoiceSend
+  requestInvoiceCreate(orderID: ID!, input: RequestInvoiceCreateInput): RequestInvoiceCreate
+  requestInvoiceDelete(orderID: ID!): RequestInvoiceDelete
```
korycins: Assuming, the order has 5 invoices. Which one do you want to remove here? :D

saleor/graphql/schema.graphql
```diff
@@ -1831,11 +1831,75 @@ type Image {
...
+}
+
+type InvoiceError {
+  code: InvoiceErrorCode!
+  field: String
+  message: String
+}
+
+enum InvoiceErrorCode {
+  PLUGIN_NOT_ENABLED
```
korycins: We could consider some generic enum for that. @maarcingebala what do you think?

maarcingebala: Probably yes, but I'm not sure if we already know what errors could happen here?

saleor/graphql/schema.graphql
dominik-zeglen: Typo!

saleor/graphql/schema.graphql
dominik-zeglen: If we want to implement `ObjectWithMetadata` we should add these fields, even if it's pointless 🤷‍♂️

saleor/graphql/schema.graphql
```diff
@@ -2407,6 +2471,12 @@ type Mutation {
   permissionGroupDelete(id: ID!): PermissionGroupDelete
   permissionGroupAssignUsers(id: ID!, users: [ID!]!): PermissionGroupAssignUsers
   permissionGroupUnassignUsers(id: ID!, users: [ID!]!): PermissionGroupUnassignUsers
+  invoiceCreate(orderID: ID!, input: InvoiceInput): InvoiceCreate
+  invoiceDelete(orderID: ID!, invoiceID: ID!): InvoiceDelete
```
maarcingebala: Why we need `orderID` here? Just the invoice ID should be enough:
```suggestion
  invoiceDelete(id: ID!): InvoiceDelete
```

saleor/graphql/schema.graphql
```diff
@@ -2407,6 +2471,12 @@ type Mutation {
...
+  invoiceUpdate(invoiceID: ID!, input: InvoiceInput): InvoiceUpdate
```
maarcingebala:
```suggestion
  invoiceUpdate(id: ID!, input: InvoiceInput): InvoiceUpdate
```

saleor/graphql/schema.graphql
maarcingebala: The same question for `requestInvoiceDelete`.

dominik-zeglen:
I didn't notice I pushed a little bit outdated schema. 
So the `Job` interface is implemented by `Invoice` since there is no point creating a separate object type if all fields can be accessed from the `Invoice` type.

saleor/graphql/schema.graphql
dominik-zeglen: @patrys explicitly asked for this to add additional level of security I guess.

patrys: I'd consider calling them something like `orderCreateInvoice` as it's really an operation on the order.

saleor/graphql/schema.graphql
```diff
@@ -2407,6 +2471,12 @@ type Mutation {
   permissionGroupDelete(id: ID!): PermissionGroupDelete
   permissionGroupAssignUsers(id: ID!, users: [ID!]!): PermissionGroupAssignUsers
   permissionGroupUnassignUsers(id: ID!, users: [ID!]!): PermissionGroupUnassignUsers
+  invoiceCreate(orderID: ID!, input: InvoiceInput): InvoiceCreate
+  invoiceDelete(orderID: ID!, invoiceID: ID!): InvoiceDelete
+  invoiceUpdate(id: ID!, input: InvoiceInput): InvoiceUpdate
+  invoiceSend(id: ID!, email: String): InvoiceSend
+  requestInvoiceCreate(orderID: ID!, input: RequestInvoiceCreateInput): RequestInvoiceCreate
```
patrys: Same here, `orderRequestInvoiceCreation`?

saleor/graphql/schema.graphql
```diff
@@ -3795,6 +3870,23 @@ type RequestEmailChange {
...
 
+type RequestInvoiceCreate {
+  errors: [Error!]!
+  invoice: Invoice
+  invoiceErrors: [InvoiceError!]!
+  order: Order
+}
+
+input RequestInvoiceCreateInput {
+  number: String
```
tomaszszymanski129: shouldn't that have order also? yeah the number is optional but we also need info which order you are requesting this for...

saleor/graphql/schema.graphql
tomaszszymanski129: i will have a look, maybe we can exclude it. I agree there is no point in making use of deprecated field.

saleor/graphql/schema.graphql
dominik-zeglen: It's in a mutation as a separate argument

saleor/graphql/schema.graphql
```diff
@@ -1831,11 +1831,55 @@ type Image {
   alt: String
 }
 
+type Invoice implements Job & Node & ObjectWithMetadata {
```
maarcingebala: I'm not sure the invoice should implement the `Job` interface? Excuse me If I'm wrong, but I thought in cases like generating files we want to first return a job that you can keep querying to check the status of the operation. Once the job is done it will return the invoice. Was it supposed to work like that @korycins @patrys?

patrys: The idea was to make long-running tasks implement the interface so you can monitor the object itself instead of an abstract job class.

maarcingebala: Sounds good. We will need to make sure that we have it consistent with other jobs, e.g. CSV exports.

maarcingebala:
Closing this, I think we have an agreement on the schema.

PR #5362, opened 2020-03-18
Allow specifying users in groups mutations

IKarbowiak:
Update create and update permission group mutations: add users to input and required validations to mutations.
Remove mutations for assign and unassign users. 

### Pull Request Checklist


saleor/graphql/account/mutations/permission_group.py
```diff
@@ -38,60 +45,153 @@ class Meta:
...
+    @classmethod
+    def perform_mutation(cls, _root, info, **data):
+        group = cls.get_instance(info, **data)
+        data = data.get("input")
+        cleaned_input = cls.clean_input(info, group, data)
+        group = cls.construct_instance(group, cleaned_input)
+        cls.clean_instance(info, group)
+        cls.save(info, group, cleaned_input)
+        cls._save_m2m(info, group, cleaned_input)
+        group.user_set.add(*cleaned_input["users_pks"])
```
NyanKiyoshi: transaction, if not already

saleor/graphql/account/mutations/permission_group.py
```diff
@@ -38,60 +45,153 @@ class Meta:
...
+    def get_user_pks(cls, cleaned_input: dict, field: str):
+        if field not in cleaned_input:
+            return []
+
+        user_ids: List[str] = cleaned_input[field]
+
+        user_pks = [
+            from_global_id_strict_type(user_id, only_type=User, field="id")
+            for user_id in user_ids
         ]
```
NyanKiyoshi: There is `resolve_global_ids_to_primary_keys` for that

saleor/graphql/account/mutations/permission_group.py
```diff
@@ -38,60 +45,153 @@ class Meta:
...
+        field: str,
+        user_pks: List[str],
+        error_field: Optional[str] = None,
+    ):
+        non_staff_users = list(
+            account_models.User.objects.filter(pk__in=user_pks)
+            .filter(is_staff=False)
+            .values_list("pk", flat=True)
+        )
+        if non_staff_users:
```
NyanKiyoshi: `exists()`

saleor/graphql/account/mutations/permission_group.py
```diff
@@ -38,60 +45,153 @@ class Meta:
...
+            ids = [graphene.Node.to_global_id("User", pk) for pk in non_staff_users]
```
NyanKiyoshi: Really needed?

saleor/graphql/account/mutations/permission_group.py
```diff
@@ -38,60 +45,153 @@ class Meta:
...
+    def update_errors(
+        cls,
+        errors: Dict[Optional[str], List[ValidationError]],
+        msg: str,
+        field: Optional[str],
+        code: str,
+        values: list,
+        error_field: Optional[str] = None,
+    ):
+        error_field = error_field or field
```
NyanKiyoshi: Why do we have two parameters?

saleor/graphql/account/mutations/permission_group.py
```diff
@@ -38,60 +45,153 @@ class Meta:
...
-    def handle_typed_errors(cls, errors: list, **extra):
-        typed_errors = [
-            cls._meta.error_type_class(
-                field=e.field,
-                message=e.message,
-                code=code,
-                index=params.get("index") if params else None,
-            )
-            for e, code, params in errors
+    def get_user_pks(cls, cleaned_input: dict, field: str):
```
NyanKiyoshi: Could we add a type hint for the return value?

saleor/graphql/account/mutations/permission_group.py
```diff
@@ -107,118 +207,96 @@ class Meta:
...
+                    "for adding and removing items."
+                )
+                cls.update_errors(
+                    errors,
+                    error_msg,
+                    None,
+                    PermissionGroupErrorCode.CANNOT_ADD_AND_REMOVE.value,
+                    values,
+                    field,
+                )
```
NyanKiyoshi: Could really use some comments

saleor/graphql/core/mutations.py
```diff
@@ -302,10 +302,18 @@ def handle_typed_errors(cls, errors: list, **extra):
...
+            for e, code, params in errors:
+                error_instance = cls._meta.error_type_class(
+                    field=e.field, message=e.message, code=code
+                )
+                if params:
+                    error_fields_in_params = set(params.keys()) & error_class_fields
+                    for error_field in error_fields_in_params:
+                        setattr(error_instance, error_field, params[error_field])
+                typed_errors.append(error_instance)
+
```
NyanKiyoshi: Could use some comments

saleor/graphql/core/types/common.py
```diff
@@ -30,6 +31,21 @@ class CountryDisplay(graphene.ObjectType):
     vat = graphene.Field(VAT, description="Country tax.")
 
 
+class LanguageDisplay(graphene.ObjectType):
+    code = LanguageCodeEnum(description="Language code.", required=True)
+    language = graphene.String(description="Language.", required=True)
```
NyanKiyoshi: Description are not useful, the user already knows that

tests/api/benchmark/test_permission_group.py
```diff
@@ -114,7 +120,12 @@ def test_permission_group_update(
         "id": graphene.Node.to_global_id("Group", group.id),
         "input": {
             "name": "New permission group",
-            "permissions": [AccountPermissions.MANAGE_SERVICE_ACCOUNTS.name],
+            "addPermissions": [AccountPermissions.MANAGE_SERVICE_ACCOUNTS.name],
+            "removePermissions": [AccountPermissions.MANAGE_USERS.name],
+            "addUsers": [graphene.Node.to_global_id("User", staff_user.pk)],
+            "removeUsers": [
+                graphene.Node.to_global_id("User", group.user_set.first().pk)
+            ],
```
NyanKiyoshi: Could use some comments or query string to explain what we are doing and testing

tests/api/test_account_permission_group.py
```diff
@@ -40,6 +44,11 @@ def test_permission_group_create_mutation(
     permission_manage_users,
     permission_manage_service_accounts,
 ):
+    staff_user2 = User.objects.get(pk=staff_user.pk)
+    staff_user2.id = None
+    staff_user2.email = "test@example.com"
+    staff_user2.save()
```
NyanKiyoshi: A comment saying it creates a new one, could be nice

tests/api/test_account_permission_group.py
```diff
@@ -127,21 +151,116 @@ def test_permission_group_create_mutation_group_exists(
...
+    customer_user,
+    permission_manage_staff,
+    staff_api_client,
+    permission_manage_users,
+    permission_manage_service_accounts,
+):
+    second_customer = User.objects.get(pk=customer_user.pk)
+    second_customer.id = None
+    second_customer.email = "second_customer@test.com"
+    second_customer.save()
```
NyanKiyoshi: same

tests/api/test_account_permission_group.py
```diff
@@ -127,21 +151,116 @@ def test_permission_group_create_mutation_group_exists(
...
+    data = content["data"]["permissionGroupCreate"]
+    errors = data["permissionGroupErrors"]
+
+    assert errors
+    assert len(errors) == 1
+    assert errors[0]["field"] == "users"
+    assert errors[0]["permissions"] is None
+    assert set(errors[0]["users"]) == set(user_ids[1:])
+    assert errors[0]["code"] == AccountErrorCode.ASSIGN_NON_STAFF_MEMBER.name
+    assert data["group"] is None
```
NyanKiyoshi: Docstring or comments explaining what we are testing and expecting would be nice

tests/api/test_account_permission_group.py
```diff
@@ -217,6 +342,8 @@ def test_permission_group_update_mutation(
     data = content["data"]["permissionGroupUpdate"]
     permission_group_data = data["group"]
 
+    users.remove(group_user.email)
+    users.add(staff_user.email)
```
NyanKiyoshi: too

tests/api/test_account_permission_group.py
```diff
@@ -336,223 +503,322 @@ def test_permission_group_update_mutation_no_input_data(
...
-    assert_no_permission(response)
+    assert len(errors) == 2
+    assert {error["code"] for error in errors} == {
+        PermissionGroupErrorCode.CANNOT_ADD_AND_REMOVE.name
+    }
+    assert {error["field"] for error in errors} == {None}
+    assert set(permissions) in [
+        set(error["permissions"]) if error["permissions"] else None for error in errors
+    ]
+    assert [staff_user2_id] in [error["users"] for error in errors]
```
NyanKiyoshi: same

tests/api/test_account_permission_group.py
```diff
@@ -336,223 +503,322 @@ def test_permission_group_update_mutation_no_input_data(
...
+    permission_group_manage_users,
+    staff_user,
+    customer_user,
+    permission_manage_staff,
+    staff_api_client,
+    permission_manage_users,
+    permission_manage_service_accounts,
+    permission_manage_orders,
+):
+    staff_user.user_permissions.add(
```
NyanKiyoshi: too

tests/api/test_account_permission_group.py
```diff
@@ -336,223 +503,322 @@ def test_permission_group_update_mutation_no_input_data(
...
+    assert permissions[:1] in permission_errors
+    assert permissions[1:] in permission_errors
+    user_errors = [error["users"] for error in errors]
+    assert user_ids[1:] in user_errors
+    assert user_ids[:1] in user_errors
+    assert {error["code"] for error in errors} == {
+        PermissionGroupErrorCode.ASSIGN_NON_STAFF_MEMBER.name,
+        PermissionGroupErrorCode.NO_PERMISSION.name,
+        PermissionGroupErrorCode.CANNOT_ADD_AND_REMOVE.name,
+    }
```
NyanKiyoshi: very confusing. Add descriptive assertion messages

tests/api/test_account_permission_group.py
```diff
@@ -575,137 +838,37 @@ def test_permission_group_assign_users_mutation_customer_user(
...
-    )
-    permissions_codes = {
-        permission["code"].lower()
-        for permission in permission_group_data["permissions"]
-    }
-    assert (
-        set(group.permissions.all().values_list("codename", flat=True))
-        == permissions_codes
-    )
+    with pytest.raises(AssertionError):
```
NyanKiyoshi: Check the message is the one excepted and not something else

saleor/graphql/account/mutations/permission_group.py
IKarbowiak: I need users' graphene `ids` to add them to error.

saleor/graphql/account/mutations/permission_group.py
IKarbowiak: Because in `PermissionGroupError` are fields `permissions` and `users` and mutation have two list for both of them. One to add and one to remove `users` or `permissions`, so in this cases `error_field` will be different than `field`.

saleor/graphql/account/mutations/permission_group.py
IKarbowiak: I don't see the reason to use `exists()` here.

saleor/graphql/core/types/common.py
IKarbowiak: I only moved it from another place. But I can remove these descriptions if it's needed.

tests/api/benchmark/test_permission_group.py
IKarbowiak: I am just testing `PermissionGroupUpdate` mutation, I think that's clear.

tests/api/test_account_permission_group.py
IKarbowiak: I think it's clear and comment here is useless.

tests/api/test_account_permission_group.py
IKarbowiak: I think it's clear and comment here is useless.

saleor/graphql/account/mutations/permission_group.py
NyanKiyoshi: But is it really needed to tell which ones?

saleor/graphql/core/types/common.py
NyanKiyoshi: Okay, but I mean we could probably have something more descriptive.
- code: ISO 639 representation of the language name
- language: full (translated?) name of the language

tests/api/benchmark/test_permission_group.py
NyanKiyoshi: I mean, explaining what's going on, that you are ensuring it correctly adds the 'manage service accounts' permission and drops the 'manage users' permissions and removes add the default staff user into the group and then remove the <???> user from the group.

A docstring explaining what is going on that wall of code

tests/api/test_account_permission_group.py
NyanKiyoshi: Explicit is better than having to guess the magic happening.
okay... the code getting a user... copying it... changing it's ID (easily missable), change its email and then... save the copied user. Does it override? Oh wait no. Seems like django is doing an implicit create in background because the ID is none.

I prefer the user reading the code of someone else to avoid to having to go through all this though process and be able to go on with his life by looking at one short comment saying "Create a second user object for ensuring blablabla". Comments are there for that: making it quick to understand, not waste time reading and analyzing each line and noticing every details and magics hiding underneath of django.

Comments save time for maintaining and we trust them, we don't try to analyze each lines on by one.

saleor/graphql/account/mutations/permission_group.py
NyanKiyoshi: Maybe we could come up with a better name? Or at least have a docstring describing the arguments so we know what is the difference between `error_field` and `field`. Reading the code seems like both are the same things–which is not the case.
A docstring would fix the issue and prevent questions, doubts and waste of time.

saleor/graphql/account/mutations/permission_group.py
NyanKiyoshi: Depends whether it is really useful to return the ids inside the error. Not doing exist is more expensive, but if there is an error, then it is more expensive to do exist if we want to render/return the IDs.

tests/api/test_account_permission_group.py
maarcingebala: Quick `cmd+shift+f` in my VSCode reveals almost 50 usages of this pattern in our tests, I think it's pretty familiar to any Django dev.

saleor/graphql/account/mutations/permission_group.py
IKarbowiak: During one of our daily, we decided that we want to.

saleor/graphql/account/mutations/permission_group.py
IKarbowiak: As above, we decided to return IDs on one of our daily.

tests/api/benchmark/test_permission_group.py
IKarbowiak: But it's a benchmark test. There is simple logic in these tests, just uses all available input field. As far as I know, the aim of the benchmark tests is just checking if the number of queries doesn't change. I really don't see any reason to add a comment here.

tests/api/test_account_permission_group.py
IKarbowiak: I'm not sure if we want to check error messages. @maarcingebala

saleor/graphql/account/mutations/permission_group.py
```diff
@@ -107,118 +214,102 @@ class Meta:
...
-            for user_id in user_ids
-        ]
+    @classmethod
+    def clean_input(
+        cls, info, instance, data,
+    ):
+        cleaned_input = super().clean_input(info, instance, data)
+        errors = defaultdict(list)
+        for field in ["users", "permissions"]:
+            cls.check_for_duplicates(errors, cleaned_input, field)
```
maarcingebala: I think it would be more readable if we had it more explicitly:
```
cls.check_for_duplicates(cleaned_input, "add_permissions", "remove_permissions")
cls.check_for_duplicates(cleaned_input, "add_users", "remove_users")
```
It could be moved before the initial `super().clean_input()` call because we can find out pretty quickly that you have duplicates and raise the validation error early.

saleor/graphql/account/mutations/permission_group.py
```diff
@@ -107,118 +214,102 @@ class Meta:
...
+        remove_field = f"remove_{field}"
+        if add_field in cleaned_input and remove_field in cleaned_input:
+            # get items which are in both list
+            common_items = set(cleaned_input[add_field]) & set(
+                cleaned_input[remove_field]
             )
+            if common_items:
+                if field == "permission":
+                    # prepare permissions enum list for error
+                    values = [PermissionEnum.get(perm) for perm in common_items]
```
maarcingebala: Do we have to do this transformation? Aren't `common_items` already list of permission names that were provided in the input and could be returned directly to the error params?

saleor/account/error_codes.py
```diff
@@ -15,10 +15,17 @@ class AccountErrorCode(Enum):
...
     PASSWORD_TOO_SHORT = "password_too_short"
     PASSWORD_TOO_SIMILAR = "password_too_similar"
     REQUIRED = "required"
     UNIQUE = "unique"
+
+
+class PermissionGroupErrorCode(Enum):
+    ASSIGN_NON_STAFF_MEMBER = "assign_non_staff_member"
+    CANNOT_ADD_AND_REMOVE = "cannot_add_and_remove"
+    NO_PERMISSION = "no_permission"
```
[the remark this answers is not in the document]
IKarbowiak: Yes, I will change that.

tests/api/test_account_utils.py
```diff
@@ -0,0 +1,69 @@
...
+    permission_group_manage_users,
+    permission_manage_users,
+    permission_manage_orders,
+):
+    staff_user.user_permissions.add(permission_manage_orders)
+    result = can_user_manage_group(staff_user, permission_group_manage_users)
+    assert not result
+
+
+def test_can_manage_group_true(
```
fowczarek: I think we can remove `true` from the test name.

saleor/graphql/account/utils.py
```diff
@@ -101,9 +104,16 @@ def get_allowed_fields_camel_case(allowed_fields: set) -> set:
 
 
 def get_permissions_user_has_not(user: "User", permissions: List[str]):
```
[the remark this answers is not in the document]
IKarbowiak: Yes, that's better.

saleor/graphql/account/mutations/permission_group.py
IKarbowiak: Yeah, I missed it.

tests/api/test_account_permission_group.py
NyanKiyoshi: That would be better, otherwise you don't know for sure, it can be something else. It is often bad to check a very common exception without checking it is actually the proper and expected reason/trigger

tests/api/test_account_permission_group.py
fowczarek: We need at least 2 staff users in these tests. We should introduce fixture like `staffs_users` and inside it use bulk_create.

fowczarek:
@maarcingebala  @patrys 
Should we add test for:
1. Remove Me from the last group
2. Remove the last user in the group. 

Should we allow to create an empty group?

tests/api/test_account_permission_group.py
IKarbowiak: I removed that function because it checks only graphQL schema.

maarcingebala:
@fowczarek If we're adding these two validations now then yes, we should add tests for them I think. I think it's fine to create an empty group. You may want to first create some groups and assign users later.

saleor/graphql/account/utils.py
```diff
@@ -101,9 +104,16 @@ def get_allowed_fields_camel_case(allowed_fields: set) -> set:
...
-            indexes.append(index)
-    return indexes
+            missing_permissions.append(perm)
+    return missing_permissions
+
+
+def can_user_manage_group(user: "User", group: "Group"):
+    permissions = group.permissions.annotate(
+        lookup_field=Concat("content_type__app_label", Value("."), "codename")
+    ).values_list("lookup_field", flat=True)
```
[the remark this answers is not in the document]
IKarbowiak: I created a function from this because it'll be used in further permission group implementation.

saleor/graphql/account/mutations/permission_group.py
```diff
@@ -107,118 +155,106 @@ class Meta:
...
+    ):
+        if "add_users" in cleaned_input:
+            group.user_set.add(*cleaned_input["add_users"])
+        if "remove_users" in cleaned_input:
+            group.user_set.remove(*cleaned_input["remove_users"])
+
+        if "add_permissions" in cleaned_input:
+            group.permissions.add(*cleaned_input["add_permissions"])
+        if "remove_perissions" in cleaned_input:
+            remove_permissions = get_permissions(cleaned_input["remove_perissions"])
```
maarcingebala: `get_permissions` is needed here? I'm asking because it's not used for adding.

saleor/graphql/account/mutations/permission_group.py
```diff
@@ -107,118 +155,106 @@ class Meta:
...
+            group.permissions.remove(*remove_permissions)
 
+    @classmethod
+    def clean_input(
+        cls, info, instance, data,
+    ):
+        if not can_user_manage_group(info.context.user, instance):
+            error_msg = "You can't manage group with permissions out of your scope."
+            code = PermissionGroupErrorCode.OUT_OF_SCOPE_PERMISSION.value
+            raise ValidationError({None: ValidationError(error_msg, code)})
```
maarcingebala: Since the error is not assigned to a field, couldn't we just do this?
```suggestion
            raise ValidationError(error_msg, code)
```

saleor/graphql/account/mutations/permission_group.py
IKarbowiak: It's also used for adding but in `clean_permissions`. But maybe it probably shouldn't be there, because it's not clear what going on in `update_group_permissions_and_users`.

IKarbowiak: I left `get_permission` in `clean_permissions` but moved the one from `update_group_permissions_and_users`  to `clean_input` in `PermissionGrpupUpdate`. I hope it's more clear right now.

PR #5368, opened 2020-03-19
Add groups to user type

IKarbowiak:
Add permission groups to user type. 

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/types.py
+++ b/saleor/graphql/account/types.py
@@ -326,12 +326,12 @@ class User(CountableDjangoObjectType):
         if root.is_superuser:
             permissions = get_permissions()
         else:
-            permissions = root.user_permissions.prefetch_related(
-                "content_type"
-            ).order_by("codename")
             groups = root.groups.all()
-            permissions = permissions | auth_models.Permission.objects.filter(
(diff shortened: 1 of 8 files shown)


saleor/graphql/account/types.py
```diff
@@ -318,11 +326,19 @@ def resolve_permissions(root: models.User, _info, **_kwargs):
...
-            permissions = root.user_permissions.prefetch_related(
-                "content_type"
-            ).order_by("codename")
+            groups = root.groups.all()
+            user_permissions = root.user_permissions.all()
+            group_permissions = auth_models.Permission.objects.filter(group__in=groups)
+            permissions = user_permissions | group_permissions
+            permissions = permissions.prefetch_related("content_type").order_by(
+                "codename"
+            )
```
maarcingebala: I think this could be a utility function e.g. `get_user_permissions(user: models.User)` with a docstring saying that it returns a sum of group and user's individual permissions. But then we should rename `saleor.core.permissions.get_permissions` to `get_all_permissions` with a docstring "returns a list of all available permissions`, so it's easier to understand the difference between them. I think it would improve the readability of the code.

PR #5369, opened 2020-03-19
Add command createaccount

korycins:
Django command for creating service account instance.
```
./manage.py createserviceaccount service_name --permission account.manage_users --target_url http://127.0.0.1:8080/webhooks/register
```
`target_url` is optional argument. If provided, the command will send service_account data to given URL

### DIFF
--- a/saleor/account/management/commands/createserviceaccount.py
+++ b/saleor/account/management/commands/createserviceaccount.py
@@ -71,8 +71,6 @@ class Command(BaseCommand):
         token_obj = service_account.tokens.create()
         data = {
             "auth_token": token_obj.auth_token,
-            "name": name,
-            "permissions": permissions,
         }
         if target_url:
             self.send_service_account_data(target_url, data)
--- a/saleor/account/management/commands/createserviceaccount.py
+++ b/saleor/account/management/commands/createserviceaccount.py
@@ -1,0 +1,79 @@
+import json
+from typing import Any, Dict, List, Optional
+
+import requests
+from django.contrib.auth.models import Permission
+from django.contrib.sites.models import Site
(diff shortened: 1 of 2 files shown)


saleor/account/management/commands/createserviceaccount.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+        parser.add_argument("--is_active", default=True, dest="is_active")
+        parser.add_argument(
+            "--target_url",
+            dest="target_url",
+            help="Url which will receive newly created data of service account object.",
+        )
+
+    def validate_permissions(self, required_permissions: List[str]):
+        permissions = list(map(lambda x: x[1], get_permissions_enum_list()))
```
maarcingebala:
```suggestion
        permissions = [x[1] for x in get_permissions_enum_list()]
```

korycins: <3

saleor/account/management/commands/createserviceaccount.py
```diff
@@ -0,0 +1,78 @@
...
+        permissions_qs = self.clean_permissions(permissions)
+        service_account.permissions.add(*permissions_qs)
+        token_obj = service_account.tokens.create()
+        data = {
+            "auth_token": token_obj.auth_token,
+        }
+        if target_url:
+            self.send_service_account_data(target_url, data)
+
+        return json.dumps(data) if not target_url else ""
```
NyanKiyoshi: Could we maybe describe the behavior in the docstring? Why does it not return the JSON when there is not `target_url`? (because it sent it) that would be nice so we don't have to look at the whole code to understand what is going on

PR #5384, opened 2020-03-24
Ability to mutate product variant name.

zzart:
productvariant.name is present in both schemas :
product_productvariant
order_orderline
I'm not sure why it has been ommited in graphql spec but it's pretty useful to have ability to set this field.  
In case it is a deliberate omission ignore this PR.

### Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/mutations/products.py
+++ b/saleor/graphql/product/mutations/products.py
@@ -1096,6 +1096,7 @@ class ProductVariantInput(graphene.InputObjectType):
     cost_price = Decimal(description="Cost price of the variant.")
     price_override = Decimal(description="Special price of the particular variant.")
     sku = graphene.String(description="Stock keeping unit.")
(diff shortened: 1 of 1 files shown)


maarcingebala:
We omitted this field on purpose, as the idea behind variant names was that they will hold the string representation of attributes assigned to a variant. So if my variant has two attributes "size" and "color" with some values assigned, the variant name would be e.g. "Red - XL". This saves database queries when rendering variants as we don't have to fetch all attribute values.

I think this may be a bit confusing and should store the representation of the attributes in some other field, while "name" would be just the optional name of a variant? What do you think @patrys?

zzart:
We have very custom use case when we have just a few products and 100's of variants. So the `name` definitely helps to distinguish the variants. For the case @maarcingebala is describing postgresql's  JSONB or Array fields would be a more suited alternative. I also agree it is a bit confusing to be using `name` field as a string representation of variants attributes.

PR #5394, opened 2020-03-26
Push version tag to Docker Hub

machadovilaca:
I want to merge this change because currently only `latest` and `COMMIT_SHA1` are being pushed to Docker Hub and for anyone who wants to run saleor on his own using Docker, without a need to build own image, it's harder to keep track of released versions without the version tag.

In this PR I enable circleci on tag releases, and push, on every deploy, the `SALEOR_VERSION` tag to Docker Hub.

# Impact

* [ ] New migrations
* [ ] New API fields or mutations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,9 +33,10 @@ jobs:
           name: Push Docker image
           command: |
             docker login -u $DOCKERCLOUD_USER -p $DOCKERCLOUD_PASS
-            docker tag mirumee/saleor:latest mirumee/saleor:$CIRCLE_SHA1
+            docker tag mirumee/saleor:latest mirumee/saleor:$CIRCLE_SHA1 mirumee/saleor:$SALEOR_VERSION
             docker push mirumee/saleor:$CIRCLE_SHA1
             docker push mirumee/saleor:latest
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -33,9 +33,10 @@ jobs:
           name: Push Docker image
           command: |
             docker login -u $DOCKERCLOUD_USER -p $DOCKERCLOUD_PASS
-            docker tag mirumee/saleor:latest mirumee/saleor:$CIRCLE_SHA1
+            docker tag mirumee/saleor:latest mirumee/saleor:$CIRCLE_SHA1 mirumee/saleor:$SALEOR_VERSION
             docker push mirumee/saleor:$CIRCLE_SHA1
             docker push mirumee/saleor:latest
+            docker push mirumee/saleor:$SALEOR_VERSION
```
NyanKiyoshi, line 39: Sounds like it would push for every build on master instead of for every release?

machadovilaca: On every push to master `SALEOR_VERSION` already had the value of `git rev-parse --short HEAD` but was not used except on image build. So also pushing that tag to Docker Hub. Anyway can remove that if you do think it's better.

PR #5395, opened 2020-03-27
Update user to group assign restriction

IKarbowiak:
Prevent from assign and unassign users to group. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/graphql/account/utils.py
```diff
@@ -122,6 +124,30 @@ def get_out_of_scope_permissions(user: "User", permissions: List[str]):
...
+    users_data = users.annotate(
+        group_perms=ArrayAgg("groups__permissions"),
+        user_perms=ArrayAgg("user_permissions__pk"),
+    ).values("group_perms", "user_perms")
+
+    for index, user_data in enumerate(users_data):
+        user_permissions = set(user_data["group_perms"]).union(
+            set(user_data["user_perms"])
+        )
+        user_permissions.discard(None)
```
maarcingebala: Shouldn't we somehow use `get_user_permissions` here? One problem that I see is that `get_user_permission` always performs database query so this might inefficient.

IKarbowiak: Because of that, I decided to not use it for every user which we need to check. When it will be more users there will be plenty of database queries.

PR #5400, opened 2020-03-27
Invoice send email mutation

tomaszszymanski129:
Send Invoice email mutation, unit tests.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested


saleor/graphql/order/mutations/orders.py
```diff
@@ -671,3 +672,42 @@ def perform_mutation(cls, _root, info, **data):
...
+        model = models.Invoice
+        permissions = (OrderPermissions.MANAGE_ORDERS,)
+        error_type_class = InvoiceError
+        error_type_field = "invoice_errors"
+
+    @classmethod
+    def clean_instance(cls, info, instance):
+        validation_errors = {}
+        if instance.status != InvoiceStatus.READY:
+            validation_errors["invoice"] = ValidationError(
```
maarcingebala: Here you're using `"invoice"` as a field for the validation error, but we use these fields to indicate which input field causes the error. In the case of this mutation, the validation error isn't really related to any input field (there is only `id`), so I would leave it empty and just return a generic error.  But then we would have to also change the generic `InvoiceErrorCode.INVALID` to something more specific like e.g `NOT_READY`.

saleor/graphql/order/mutations/orders.py
```diff
@@ -671,3 +672,42 @@ def perform_mutation(cls, _root, info, **data):
...
+                "Invoice needs to be ready in order to send it.",
+                code=InvoiceErrorCode.INVALID,
+            )
+        for field in ("url", "number"):
+            if not getattr(instance, field):
+                validation_errors[field] = ValidationError(
```
maarcingebala: Similar as above, here we're matching the error with a particular field, but none of these fields were passed in the input. I would say this error means that the invoice object is in a state that doesn't allow us to perform this operation, so it's a general error. We also have `URL_OR_NUMBER_NOT_SET` error code from what I see (which I thought we dropped 🧐) - maybe it would make sense to use this code here.

tests/api/test_order.py
```diff
@@ -3162,3 +3149,39 @@ def test_create_invoice_empty_params(
     assert not Invoice.objects.filter(
         order_id=order.pk, status=InvoiceStatus.READY
     ).exists()
+
+
+@mock.patch("saleor.order.emails.send_invoice.delay")
+def test_send_invoice(email_mock, staff_api_client, permission_manage_orders, orders):
```
fowczarek: Why are you using `orders` fixture instead of `order`?

tests/api/test_order.py
```diff
@@ -3162,3 +3149,39 @@ def test_create_invoice_empty_params(
...
+        SEND_INVOICE_MUTATION, variables, permissions=[permission_manage_orders]
+    )
+    content = get_graphql_content(response)
+    assert not content["data"]["sendInvoiceEmail"]["invoiceErrors"]
+    email_mock.assert_called_with(invoice.pk)
+
+
+@mock.patch("saleor.order.emails.send_invoice.delay")
+def test_send_pending_invoice(
+    email_mock, staff_api_client, permission_manage_orders, orders
```
fowczarek: Why are you using `orders` fixture instead of `order`?

tests/api/test_order.py
tomaszszymanski129: Because i didn't see it. Will change that here and everywhere else.

PR #5408, opened 2020-03-30
Add editable groups on user type

IKarbowiak:
Editable groups returns groups which the user can manage. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/graphql/account/types.py
```diff
@@ -310,6 +310,13 @@ class User(CountableDjangoObjectType):
         ),
         model_field="groups",
     )
+    editable_groups = gql_optimizer.field(
```
maarcingebala: Optimizer here is not needed because we don't use `user.groups` in the resolver anyway.

PR #5411, opened 2020-03-30
Add shipping rate price validation

### LINKED ISSUE #5361: It should not be possible to have negative price on shipping rate
1. Go to the dashboard's shipping section
2. Add new rate or edit existing one
3. Set -1 in price field

### What I expected to happen
To get an error like "shipping rate price cannot be lower than 0".

kswiatek92:
closes #5361 

# Pull Request Checklist

* [x] The changes are tested


tests/api/test_shipping_method.py
```diff
@@ -305,6 +305,32 @@ def test_create_shipping_method(
...
+        "type": ShippingMethodTypeEnum.PRICE.name,
+    }
+    response = staff_api_client.post_graphql(
+        query, variables, permissions=[permission_manage_shipping]
+    )
+    content = get_graphql_content(response)
+    data = content["data"]["shippingPriceCreate"]
+    assert data["errors"][0] == {
+        "field": "price",
+        "message": "Shipping rate price cannot be lower than 0.",
```
maarcingebala: In the `PRICE_BASED_SHIPPING_QUERY` we should include:
```
shippingErrors {
  field
  code
}
```
And in this line instead of comparing the `message`, we should check if the `field` and `code` are as we expected.

kswiatek92: sure

tests/api/test_shipping_method.py
```diff
@@ -305,6 +309,30 @@ def test_create_shipping_method(
...
+        "maximumOrderPrice": 20,
+        "type": ShippingMethodTypeEnum.PRICE.name,
+    }
+    response = staff_api_client.post_graphql(
+        query, variables, permissions=[permission_manage_shipping]
+    )
+    content = get_graphql_content(response)
+    data = content["data"]["shippingPriceCreate"]
+    assert data["shippingErrors"][0]["field"] == "price"
+    assert data["shippingErrors"][0]["code"] == "INVALID"
```
maarcingebala: We should import the enum in this file and use it here instead of hardcoding the value.

PR #5423, opened 2020-04-01
Add attribute validation to attributeAssign

### LINKED ISSUE #5140: Assigning a non-existing attribute to a product type crashes
**Query:**
```gql
mutation AssignAttributesToPT(
  $productTypeID: ID!
  $operations: [AttributeAssignInput]!
) {
  attributeAssign(productTypeId: $productTypeID, operations: $operations) {
    errors {
      field
      message
    }
  }
}
```

**Variables:**
```python
{
  "variables": {
    "productTypeID": "UHJvZHVjdFR5cGU6MTE2",   # <- valid and existing product type ID
    "operations": [
      {
        "id": "QXR0cmlidXRlOjU1NQ==",  # <- valid attribute ID but non-existing
        "type": "PRODUCT"
      }
    ]
  }
}
```

### What I expected to happen
To have a check in the mutation to prevent it from happening.

kswiatek92:
closes #5140 
# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/core/types/common.py
+++ b/saleor/graphql/core/types/common.py
@@ -73,6 +73,14 @@ class ProductError(Error):
     code = ProductErrorCode(description="The error code.", required=True)
 
 
+class ProductAttributeError(ProductError):
+    attributes = graphene.List(
+        graphene.NonNull(graphene.ID),
+        description="List of attributes IDs which causes the error.",
+        required=False,
+    )
+
+
 class BulkProductError(ProductError):
     index = graphene.Int(
         description="Index of an input list item that caused the error."
--- a/saleor/graphql/product/mutations/attributes.py
+++ b/saleor/graphql/product/mutations/attributes.py
@@ -405,6 +405,9 @@ class AttributeAssign(BaseMutation):
         )
         if len(attrs_pk) != len(attributes):
             invalid_attrs = set(attrs_pk) - set(attributes)
+            invalid_attrs = [
+                graphene.Node.to_global_id("Attribute", pk) for pk in invalid_attrs
+            ]
             raise ValidationError(
                 {
                     "operations": ValidationError(
--- a/saleor/graphql/product/mutations/attributes.py
+++ b/saleor/graphql/product/mutations/attributes.py
@@ -317,18 +317,6 @@ class AttributeAssign(BaseMutation):
             pk = from_global_id_strict_type(
                 operation.id, only_type=Attribute, field="operations"
             )
-
-            try:
-                models.Attribute.objects.get(pk=pk)
-            except models.Attribute.DoesNotExist:
-                raise ValidationError(
-                    {
-                        "operations": ValidationError(
-                            f"Attribute doesn't exist.",
-                            code=ProductErrorCode.NOT_FOUND.value,
-                        )
-                    }
-                )
             if operation.type == AttributeTypeEnum.PRODUCT:
                 product_attrs_pks.append(pk)
             else:
@@ -397,6 +385,17 @@ class AttributeAssign(BaseMutation):
     @classmethod
     def clean_operations(cls, product_type, product_attrs_pks, variant_attrs_pks):
         """Ensure the attributes are not already assigned to the product type."""
+        attrs_pk = product_attrs_pks + variant_attrs_pks
+        attributes_count = models.Attribute.objects.filter(id__in=attrs_pk).count()
+        if len(attrs_pk) != attributes_count:
+            raise ValidationError(
+                {
+                    "operations": ValidationError(
+                        f"Attribute doesn't exist.",
+                        code=ProductErrorCode.NOT_FOUND.value,
+                    )
+                }
+            )
         cls.check_product_operations_are_assignable(product_attrs_pks)
         cls.check_operations_not_assigned_already(
             product_type, product_attrs_pks, variant_attrs_pks
--- a/saleor/graphql/product/mutations/attributes.py
+++ b/saleor/graphql/product/mutations/attributes.py
@@ -317,6 +317,18 @@ class AttributeAssign(BaseMutation):
             pk = from_global_id_strict_type(
                 operation.id, only_type=Attribute, field="operations"
             )
+
+            try:
+                models.Attribute.objects.get(pk=pk)
+            except models.Attribute.DoesNotExist:
+                raise ValidationError(
+                    {
+                        "operations": ValidationError(
+                            f"Attribute doesn't exist.",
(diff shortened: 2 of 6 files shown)


saleor/graphql/product/mutations/attributes.py
```diff
@@ -317,6 +317,18 @@ def get_operations(cls, info, operations: List[AttributeAssignInput]):
             pk = from_global_id_strict_type(
                 operation.id, only_type=Attribute, field="operations"
             )
+
+            try:
+                models.Attribute.objects.get(pk=pk)
```
maarcingebala, line 322: It's not an optimal solution - it performs a database query for each operation. You could instead do one query always if you used `Attribute.objects.filter(id__in=all_attribute_ids)` or something like that and compare the number of instances you got from DB with the number of IDs passed in the input. 

There is also `clean_operations` function which does other validations as well, so this one would fit there.

We are able to pass additional context in the error response to tell the user which IDs are wrong, @IKarbowiak did that in one of her PRs; Iga can you paste some quick example how can we pass attribute IDs to the error context?

kswiatek92: sure I'll fix fix that

IKarbowiak: 1) Extend `ErrorClass` for the additional field with attributes ids
2) To add ids to `ValidationError` you need to provide it to `params` attribute in form of dict, where `key` is name of field from error class, and `value` is list of ids or something else. Like that:
`ValidationError(message=msg, code=code, params={"field_name": [IDs]})`
3) You must also update handle_errors. We have a solution for that in BaseMutation but it's still on my branch, unfortunately 😞. So for now you can just override `handle_typed_errors` as is done here https://github.com/mirumee/saleor/pull/5399/files#diff-5cc938d1fb7222180f0de3a3eb4fb4e1R102 .

kswiatek92: @IKarbowiak something like this?

saleor/graphql/product/mutations/attributes.py
```diff
@@ -382,9 +382,38 @@ def check_product_operations_are_assignable(cls, product_attrs_pks):
...
+            "pk", flat=True
+        )
+        if len(attrs_pk) != len(attributes):
+            invalid_attrs = set(attrs_pk) - set(attributes)
+            raise ValidationError(
+                {
+                    "operations": ValidationError(
+                        f"Attribute doesn't exist.",
+                        code=ProductErrorCode.NOT_FOUND.value,
+                        params={"attributes": list(invalid_attrs)},
```
IKarbowiak, line 413: Did you check if it gives you the correct graphql IDs? You probably need to convert invalid_attrs pks to graphql ids.

saleor/graphql/product/mutations/attributes.py
```diff
@@ -300,7 +300,7 @@ class Arguments:
 
     class Meta:
         description = "Assign attributes to a given product type."
-        error_type_class = ProductError
+        error_type_class = ProductAttributeError
         error_type_field = "product_errors"
```
IKarbowiak, line 304: If `error_type_class` is changed we probably should also change `error_type_field`.

tests/api/test_attributes.py
```diff
@@ -1252,6 +1253,27 @@ def test_assign_attributes_to_product_type(
...
+    ]
+    variables = {"productTypeId": product_type_global_id, "operations": operations}
+
+    response = staff_api_client.post_graphql(
+        query, variables, permissions=[permission_manage_products]
+    )
+    content = get_graphql_content(response)
+    content = content["data"]["attributeAssign"]
+    assert content["productErrors"][0]["code"] == ProductErrorCode.NOT_FOUND.name
+    assert content["productErrors"][0]["field"] == "operations"
```
IKarbowiak, line 1274: Check also `attributes` field.

saleor/graphql/core/types/common.py
```diff
@@ -73,6 +73,14 @@ class ProductError(Error):
     code = ProductErrorCode(description="The error code.", required=True)
 
 
+class ProductAttributeError(ProductError):
+    attributes = graphene.List(
+        graphene.NonNull(graphene.ID),
+        description="List of attributes IDs which causes the error.",
+        required=False,
+    )
```
IKarbowiak, line 81: I am not sure if we should create a new class or just add the `attributes` field to `ProductError`. I think it might be useful also in other mutations and checks.

saleor/graphql/product/mutations/attributes.py
kswiatek92: forgot about this, thanks

saleor/graphql/core/types/common.py
kswiatek92: Some other errors inherit from ProductError
by adding attributes to ProductError I will add attributes to all of them

saleor/graphql/product/mutations/attributes.py
kswiatek92: I thought about this, and I'm not sure if in this case we should change it

saleor/graphql/product/mutations/attributes.py
```diff
@@ -382,9 +382,41 @@ def check_product_operations_are_assignable(cls, product_attrs_pks):
...
+        if len(attrs_pk) != len(attributes):
+            invalid_attrs = set(attrs_pk) - set(attributes)
+            invalid_attrs = [
+                graphene.Node.to_global_id("Attribute", pk) for pk in invalid_attrs
+            ]
+            raise ValidationError(
+                {
+                    "operations": ValidationError(
+                        f"Attribute doesn't exist.",
+                        code=ProductErrorCode.NOT_FOUND.value,
```
maarcingebala, line 415: You don't need `.value`, it's done automatically in `get_error_code_from_error`.

saleor/graphql/core/types/common.py
maarcingebala: I think this is fine.

kswiatek92: @maarcingebala you mean it's fine like this, or it is fine if all errors which inherit from ProductError will have `attributes`?

maarcingebala: Sorry, I meant the current solution is fine so having a separate ProductAttributeError that inherits from ProductError. There is no need to introduce the `attributes` field to all product errors as it doesn't make sense in some cases.

maarcingebala:
@IKarbowiak Can you please re-review this one?

PR #5424, opened 2020-04-02
Add delete group restriction

IKarbowiak:
Add group delete restriction. Ensure that after deleting the group for each permission, there will be at least one staff member who can manage it (has both “manage staff” and this permission).

Please notice that this is just for removing the group. Implementation for removing users from groups will be in separate PR.

⚠️ 
I am open to any suggestions for changing some of the names of the functions. It was hard to find something as short as possible and as much descriptive as possible. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


fowczarek:
@IKarbowiak  Have we any benchmark for `group_delete` with multiple groups and users. For checking the number of database queries in this logic?

IKarbowiak:
Probably not @fowczarek . I'll check it and add if there isn't.

saleor/graphql/account/utils.py
```diff
@@ -173,3 +173,133 @@ def get_groups_which_user_can_manage(user: "User") -> List[Optional[Group]]:
...
+    who can manage it (has both “manage staff” and this permission).
+    """
+    group_pk = group.pk
+    groups_data = get_group_to_permissions_and_users_mapping()
+    not_manageable_permissions = set(groups_data[group_pk]["permissions"])
+
+    # get users from groups with manage staff and look for not_manageable_permissions
+    # if any of not_manageable_permissions is found it is removed from set
+    manage_staff_users = get_users_and_look_for_permissions_in_groups_with_manage_staff(
+        group_pk, groups_data, not_manageable_permissions
```
maarcingebala, line 191: `group_pk` is passed to this function only to skip this particular PK when iterating over `groups_data`. What if we did something like:

```
not_manageable_permissions = set(groups_data.pop(group_pk)["permissions"])
```

Then we would have one less parameter in `get_users_and_look_for_permissions_in_groups_with_manage_staff` and in `look_for_permission_in_users_with_manage_staff` plus we wouldn't have to do the `if pk == group_pk:` skipping in both functions.

saleor/graphql/account/utils.py
```diff
@@ -173,3 +173,133 @@ def get_groups_which_user_can_manage(user: "User") -> List[Optional[Group]]:
...
+                filter=Q(permissions__isnull=False),
+            ),
+            users=ArrayAgg("user", filter=Q(user__is_active=True)),
+        )
+        .values("pk", "perm_codenames", "users")
+    )
+
+    for data in groups_data:
+        mapping[data["pk"]] = {
+            "permissions": data["perm_codenames"],
```
maarcingebala, line 243: In few other places where you use this entire data structure, you need to make `set` from either `permissions` or `users`. Can we do the `set` here for both lists so we don't have to do it later when we use this structure?

saleor/graphql/account/utils.py
IKarbowiak: Yes, I changed it in #5438

Issue #4933, opened 2019-11-04
productVariantUpdate mutation fails


maarcingebala:
### What I'm trying to achieve
I'm trying to update quantity of a variant (but it will probably fail for other fields -it's a general problem with updating variants).

### Steps to reproduce the problem
It can be tested in the dashboard:
1. Go to variant details page.
2. Change quantity in save changes.
3. "This variant already exists" error is shown:

![image](https://user-images.githubusercontent.com/5421321/68127984-3761d580-ff17-11e9-84b6-5520df526c69.png)

PR #5427, opened 2020-04-03
Fix: Variant update using graphql.

sebasgoldberg:

#4933 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested


sebasgoldberg:
@NyanKiyoshi 

Yes you are missing the following:

The `validate_duplicated_attribute_values` method was called always.

With this modification it is called only when the variant is created (only inside the `else` of the `if instance.product_id is not None`.

This fix the problem reported in the issue #4933.

PR #5430, opened 2020-04-05
Lock django-graphql-jwt to 0.3.0

adithyabsk:
Addresses: https://github.com/mirumee/saleor/issues/4652#issuecomment-609491605

* Remove semver caret from `django-graphql-jwt`
* Run `poetry lock`

### Why this is needed
When users of the project add dependencies using `poetry add [PACKAGE_NAME]`, `poetry` re-locks the project updating the `poetry.lock` file. You then encounter strange errors because the `django-graphql-jwt` package upgrades to 0.3.1 because of the semver caret resolution.

This version of `django-graphql-jwt` is incompatible with `saleor` mainline.

I assume this change will need to be reverted when https://github.com/mirumee/saleor/pull/5149 goes through or if this is a recurring issue, semver carets should not be used on this package.


adithyabsk:
Another thing to note is that this might make the `requirements.txt` out of sync with `poetry.lock` should I update that as well? Also is there any reason you all are using `requirements.txt` if you have the poetry.lock file?

NyanKiyoshi:
Yes, they might be outdated due to `lock`.
```
poetry export -f requirements.txt -o requirements.txt --without-hashes
poetry export -f requirements.txt -o requirements_dev.txt --dev --without-hashes
```

We are using them as it allows to be independent from poetry, especially for production usage–but there is currently an issue with hashes making pip unable to successfully check for hashes in the requirement files.

adithyabsk:
@NyanKiyoshi that's fair enough. You can still install a poetry project without poetry though, you just `pip install .` and that should work as far as build processes are concerned as long as you have a version of `pip > 19.0.1`. I'll run your above commands and amend my PR

maarcingebala:
@adithyabsk Thanks for opening the PR! Would you mind updating the poetry.lock file as there are some conflicts now?

adithyabsk:
@maarcingebala done

maarcingebala:
Thanks @adithyabsk!

PR #5440, opened 2020-04-06
Add missing data migration

d-wysocki:
Data migration for https://github.com/mirumee/saleor/pull/5426

# Impact

* [X] New migrations

# Pull Request Checklist

* [X] The changes are tested


saleor/order/migrations/0081_auto_20200406_0456.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 3.0.4 on 2020-04-06 09:56
+
+from django.db import migrations
+
+
+def match_orders_with_users(apps, *_args, **_kwargs):
+    Order = apps.get_model("order", "Order")
+    User = apps.get_model("account", "User")
+
+    orders_with_no_user = Order.objects.filter(user_email__isnull=False, user=None)
```
maarcingebala, line 10: We forgot about draft orders. Draft orders are those that can be created by admins in the dashboard. Before we assign any users to drafts, they would be filtered here as well. To get only confirmed orders (non-drafts) we need to use `Order.objects.confirmed()` queryset.

We need to change this behavior in a function that matches orders in registration as well. To be sure I would also add a test that make sure that drafts are not matched with any users.

tomaszszymanski129, line 10: `orders_without_user`?

PR #5443, opened 2020-04-06
Check if sale contains product, category, collection before run celery

### LINKED ISSUE #5351: Creating a new sale raises error in Celery task
1. Run the following mutation as an admin user (with `MANAGE_DISCOUNTS` permission):
```
mutation {
  saleCreate(input: {name: "Test"}) {
    errors {
      field
      message
    }
    sale {
      id
      name
    }
  }
}
```

The response from API is successful, but in the Django server console I'm getting the following error:

```
ERROR celery.app.trace Task saleor.product.tasks.update_products_minimal_variant_prices_of_discount_task[4ec46245-d1f1-47ae-ab23-0c0ab73a9981] raised unexpected: ValueError('Provide at least one of the ID lists:\n\tproduct_ids,\n\tcategory_ids,\n\tcollection_ids.') [PID:31316:Thread-175]
  File "/Users/marcin/.pyenv/versions/saleor3.8.1/lib/python3.8/site-packages/celery/app/trace.py", line 385, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/Users/marcin/mirumee/saleor-platform/saleor/saleor/product/tasks.py", line 64, in update_products_minimal_variant_prices_of_discount_task
    update_products_minimal_variant_prices_of_discount(discount)
  File "/Users/marcin/mirumee/saleor-platform/saleor/saleor/product/utils/variant_prices.py", line 76, in update_products_minimal_variant_prices_of_discount
    update_products_minimal_variant_prices_of_catalogues(
  File "/Users/marcin/mirumee/saleor-platform/saleor/saleor/product/utils/variant_prices.py", line 62, in update_products_minimal_variant_prices_of_catalogues
    raise ValueError(
ValueError: Provide at least one of the ID lists:
	product_ids,
	category_ids,
	collection_ids.
```

I suppose that the Celery task that recalculates minimal variant prices is run even there are no products to update. Probably an additional check needs to be added to not run the task in this case.

kswiatek92:
closes #5351

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/discount/mutations.py
+++ b/saleor/graphql/discount/mutations.py
@@ -271,9 +271,9 @@ class SaleUpdateMinimalVariantPriceMixin:
         # Update the "minimal_variant_prices" of the associated, discounted
         # products (including collections and categories).
         if (
-            instance.products.all()
-            or instance.categories.all()
-            or instance.collections.all()
+            instance.products.all().exists()
+            or instance.categories.all().exists()
+            or instance.collections.all().exists()
         ):
             update_products_minimal_variant_prices_of_discount_task.delay(instance.pk)
         return super().success_response(instance)
--- a/saleor/graphql/discount/mutations.py
+++ b/saleor/graphql/discount/mutations.py
@@ -270,12 +270,7 @@ class SaleUpdateMinimalVariantPriceMixin:
     def success_response(cls, instance):
         # Update the "minimal_variant_prices" of the associated, discounted
         # products (including collections and categories).
-        if (
-            instance.products.all().exists()
-            or instance.categories.all().exists()
-            or instance.collections.all().exists()
-        ):
-            update_products_minimal_variant_prices_of_discount_task.delay(instance.pk)
+        update_products_minimal_variant_prices_of_discount_task.delay(instance.pk)
         return super().success_response(instance)
 
 
--- a/saleor/graphql/discount/mutations.py
+++ b/saleor/graphql/discount/mutations.py
@@ -270,10 +270,11 @@ class SaleUpdateMinimalVariantPriceMixin:
     def success_response(cls, instance):
         # Update the "minimal_variant_prices" of the associated, discounted
         # products (including collections and categories).
-        products_exists = instance.products.all().exists()
-        categories_exists = instance.categories.all().exists()
-        collections_exists = instance.collections.all().exists()
-        if products_exists or categories_exists or collections_exists:
+        if (
+            instance.products.all()
+            or instance.categories.all()
+            or instance.collections.all()
+        ):
             update_products_minimal_variant_prices_of_discount_task.delay(instance.pk)
         return super().success_response(instance)
 
--- a/saleor/graphql/discount/mutations.py
+++ b/saleor/graphql/discount/mutations.py
@@ -270,7 +270,11 @@ class SaleUpdateMinimalVariantPriceMixin:
     def success_response(cls, instance):
         # Update the "minimal_variant_prices" of the associated, discounted
         # products (including collections and categories).
-        update_products_minimal_variant_prices_of_discount_task.delay(instance.pk)
+        products_exists = instance.products.all().exists()
+        categories_exists = instance.categories.all().exists()
+        collections_exists = instance.collections.all().exists()
+        if products_exists or categories_exists or collections_exists:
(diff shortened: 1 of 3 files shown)


saleor/graphql/discount/mutations.py
```diff
@@ -270,7 +270,11 @@ class SaleUpdateMinimalVariantPriceMixin:
     def success_response(cls, instance):
         # Update the "minimal_variant_prices" of the associated, discounted
         # products (including collections and categories).
-        update_products_minimal_variant_prices_of_discount_task.delay(instance.pk)
+        products_exists = instance.products.all().exists()
+        categories_exists = instance.categories.all().exists()
+        collections_exists = instance.collections.all().exists()
```
NyanKiyoshi, line 275: I wonder, `instance.products.all() or instance.categories.all() or ...` would resolve to `exists()` to make it lazy?

tests/api/test_product_minimal_variant_price.py
```diff
@@ -600,6 +603,8 @@ def test_sale_delete_updates_products_minimal_variant_prices(
         }
     }
     """
+    sale.products.add(product)
+    sale.save()
```
NyanKiyoshi, line 607: Are you actually editing the sale? Seems like you only act on a M2M

kswiatek92: sorry my bad

saleor/graphql/discount/mutations.py
```diff
@@ -270,7 +270,12 @@ class SaleUpdateMinimalVariantPriceMixin:
     def success_response(cls, instance):
         # Update the "minimal_variant_prices" of the associated, discounted
         # products (including collections and categories).
-        update_products_minimal_variant_prices_of_discount_task.delay(instance.pk)
+        if (
+            instance.products.all()
```
maarcingebala, line 274: Why not `instance.products.exsists()`?  I think it should be much faster. Also, why not to move this entire check to `update_products_minimal_variant_prices_of_discount_task`? Then we would be sure that it is checked in every case that you actually use this task.

NyanKiyoshi: Depends whether `all()` evaluates to `exists()`, that was my initial question. I don't know if @kswiatek92 checked or not what behavior Django adopts in this case

kswiatek92: I didn't move this to task, because in task I only have pk, which later is used to get instance, then it checks if have any products, categories, collections, if not raise error

kswiatek92: @NyanKiyoshi sorry I thought it was statement not question, I checked it and `.all()` doesn't evaluates to `.exists()`

saleor/graphql/discount/mutations.py
```diff
@@ -270,7 +270,12 @@ class SaleUpdateMinimalVariantPriceMixin:
     def success_response(cls, instance):
         # Update the "minimal_variant_prices" of the associated, discounted
         # products (including collections and categories).
-        update_products_minimal_variant_prices_of_discount_task.delay(instance.pk)
+        if (
+            instance.products.all().exists()
```
maarcingebala, line 274: `instance.products.exists()`, you don't need `all()`.

kswiatek92: you right, my bad

maarcingebala:
@kswiatek92 I looked where this error described in the issue comes from: https://github.com/mirumee/saleor/blob/master/saleor/product/utils/variant_prices.py#L62

Maybe it would be a better idea to just get rid of this `ValueError` and instead have:
```
if q_list:
    q_or = reduce(operator.or_, q_list)
    products = Product.objects.filter(q_or).distinct()
    update_products_minimal_variant_prices(products)
```

And when `q_list` is empty just do nothing? That would solve the problem. I don't think the error is useful.

NyanKiyoshi:
I agree with @maarcingebala because the issue would still be there in case of a race condition, we don't know when the task will run it could be in the next 1/2 second, or in 2 minutes, ... or more.

kswiatek92:
@maarcingebala sure, as I said on daily, I wasn't sure why this value error was in the first place, but you both are right, it doesn't make sense to have this error there

PR #5445, opened 2020-04-06
Refactor quantity allocated implementation

fowczarek:
1. [x] Populatedb should create the real allocation.
1. [x] Move quantity allocation to the relation between stock and order line.
1. [x] Make data migration to a new solution for allocations
1. [x] When the User buys something we should allocate Stocks by iteration. (Always same order operation should use select_for_update)
1. [x] Extend the FulfillmentLine with a relationship to the Stock model to track from which warehouse a line was fulfilled.

The general flow of allocations:
mermaid-diagram-20200420152014

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] The changes are tested


fowczarek:
@NyanKiyoshi  There will be more changes. I do it on the end of my task.

saleor/core/utils/random_data.py
```diff
@@ -481,13 +475,19 @@ def create_order_lines(order, discounts, how_many=10):
...
     manager = get_plugins_manager()
+    country = order.shipping_method.shipping_zone.countries[0]
     for line in lines:
         unit_price = manager.calculate_order_line_unit(line)
         line.unit_price = unit_price
         line.tax_rate = unit_price.tax / unit_price.net
+        warehouse = (
+            Warehouse.objects.filter(shipping_zones__countries__contains=country)
+            .order_by("?")
+            .first()
```
IKarbowiak, line 488: I think we should move this Warehouse query outside the loop and add some counter to getting another warehouse for every line.

saleor/order/utils.py
```diff
@@ -278,9 +263,16 @@ def restock_order_lines(order):
...
+                deallocate_stock(line, line.quantity_unfulfilled)
             if line.quantity_fulfilled > 0:
-                increase_stock(line.variant, country, line.quantity_fulfilled)
+                allocation = line.allocations.first()
+                if allocation:
+                    warehouse = allocation.stock.warehouse
+                else:
+                    warehouse = Warehouse.objects.filter(
+                        shipping_zones__countries__contains=country
+                    ).first()
```
IKarbowiak, line 274: The same here. It can be moved outside the for loop.

saleor/order/utils.py
```diff
@@ -292,9 +284,14 @@ def restock_fulfillment_lines(fulfillment):
...
-            increase_stock(
-                line.order_line.variant, country, line.quantity, allocate=True
-            )
+            allocation = line.order_line.allocations.first()
+            if allocation:
+                warehouse = allocation.stock.warehouse
+            else:
+                warehouse = Warehouse.objects.filter(
+                    shipping_zones__countries__contains=country
+                ).first()
```
IKarbowiak, line 293: Here the same 👮‍♀️

saleor/order/utils.py
```diff
@@ -290,11 +283,18 @@ def restock_order_lines(order):
...
+        shipping_zones__countries__contains=country
+    ).first()
+
     for line in fulfillment:
         if line.order_line.variant and line.order_line.variant.track_inventory:
-            increase_stock(
-                line.order_line.variant, country, line.quantity, allocate=True
-            )
+            allocation = line.order_line.allocations.first()
+            if allocation:
```
tomaszszymanski129, line 293: `warehouse = allocation.stock.warehouse if allocation else default_warehouse`

tests/test_stock_management.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+def test_decrease_stock_many_allocations(order_line_with_allocation_in_many_stocks,):
+    order_line = order_line_with_allocation_in_many_stocks
+
+    decrease_stock(order_line, 3)
+
+    allocations = order_line.allocations.all()
+    if not allocations[0].quantity_allocated == 0:
+        import ipdb
```
tomaszszymanski129, line 211: :smile: quite a few of those here and there

fowczarek:
@NyanKiyoshi Yea I know about it. This increase of DB queries appears because queries before this pr are not optimized. I missed that at the start of work on this PR. We had a meeting about it the best solution, and we decided to fix it when implementing data loaders for stock management. 

The second action that we should take is that we should create allocations for all order lines in one transaction. This action needs to be done in separate PR to not block other warehouse related tasks.

NyanKiyoshi:
Gotcha.

saleor/order/actions.py
```diff
@@ -208,9 +207,8 @@ def clean_mark_order_as_paid(order: "Order"):
 
 def fulfill_order_line(order_line, quantity):
     """Fulfill order line with given quantity."""
-    country = get_order_country(order_line.order)
     if order_line.variant and order_line.variant.track_inventory:
-        decrease_stock(order_line.variant, country, quantity)
+        decrease_stock(order_line, quantity)
```
maarcingebala, line 211: What if an exception is raised?

saleor/order/utils.py
```diff
@@ -274,13 +259,20 @@ def delete_order_line(line):
 def restock_order_lines(order):
     """Return ordered products to corresponding stocks."""
     country = get_order_country(order)
+    default_warehouse = Warehouse.objects.filter(
+        shipping_zones__countries__contains=country
+    ).first()
```
maarcingebala, line 264: Using the first warehouse from the list is our assumption for now - maybe it would be better to put that to some function? `get_default_warehouse`? It will be easier to find later when we decide to move it to a plugin or something.

fowczarek: I think it not worth changing it because in the next PR we add warehouse to mutation input, and if we still need it we add this function to code.

saleor/order/utils.py
```diff
@@ -274,13 +259,20 @@ def delete_order_line(line):
...
 
     for line in order:
         if line.variant and line.variant.track_inventory:
             if line.quantity_unfulfilled > 0:
-                deallocate_stock(line.variant, country, line.quantity_unfulfilled)
+                deallocate_stock(line, line.quantity_unfulfilled)
```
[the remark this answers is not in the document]
fowczarek: If this line rise error it's mean that this order hasn't enough allocated items. 
Not sure we system should do in this situation. Currently, we don't have any views allowing to modify allocations.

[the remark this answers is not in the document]
fowczarek: Maby we should change the behavior of this function to deallocate all stocks if you pass more quantity that is currently allocated?

saleor/order/actions.py
fowczarek: If this line rise error it's mean that this order hasn't enough allocated items. For a better solution here we should refactor the whole FulfillmentCreate, we should have to fulfill all lines in one transaction instead of doing it line by line.

saleor/order/utils.py
maarcingebala: 👍

PR #5449, opened 2020-04-07
Add restriction to deactivating staff 

IKarbowiak:
Add restrictions for deactivating staff. Ensure that after deactivating user for each permission, there will be at least one staff member who can manage it (has both “manage staff” and this permission).

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/graphql/account/utils.py
```diff
@@ -326,3 +326,35 @@ def look_for_permission_in_users_with_manage_staff(
...
+
+    groups_data = get_group_to_permissions_and_users_mapping()
+
+    # get users from groups with manage staff
+    manage_staff_users = get_users_and_look_for_permissions_in_groups_with_manage_staff(
+        groups_data, set()
+    )
+
+    # remove deactivating or removing user from manage staff users
+    if user.pk in manage_staff_users:
```
maarcingebala, line 349: According to check in line 338 here we should be sure that the user has `manage_staff` permission, shouldn't we? Or maybe I don't understand what `get_users_and_look_for_permissions_in_groups_with_manage_staff` does.

IKarbowiak: Yes, but if the `manage_staff_users` is empty, just `remove` will fail. A small refactor is needed here (`not_manageable_permissions` should be before, and then should be check if `manage_staff_users` exists), but this function changed in #5456, so is useless to make it now.  :no_mouth:

PR #5456, opened 2020-04-08
Add restriction to bulk staff delete mutation

IKarbowiak:
Add restriction to bulk staff delete mutation.  Ensure that after removing users for each permission, there will be at least one staff member who can manage it (has both “manage staff” and this permission).

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/account/bulk_mutations.py
+++ b/saleor/graphql/account/bulk_mutations.py
@@ -61,10 +61,11 @@ class StaffBulkDelete(StaffDeleteMixin, UserBulkDelete):
     @classmethod
     def clean_instances(cls, info, users):
         errors = defaultdict(list)
+        requestor = info.context.user
         cls.check_if_users_can_be_deleted(info, users, "ids", errors)
-        cls.check_if_requestor_can_manage_users(info, users, "ids", errors)
+        cls.check_if_requestor_can_manage_users(requestor, users, "ids", errors)
         cls.check_if_removing_left_not_manageable_permissions(
-            info, users, "ids", errors
+            requestor, users, "ids", errors
         )
         return ValidationError(errors) if errors else {}
 
--- a/saleor/graphql/account/bulk_mutations.py
+++ b/saleor/graphql/account/bulk_mutations.py
@@ -63,7 +63,9 @@ class StaffBulkDelete(StaffDeleteMixin, UserBulkDelete):
         errors = defaultdict(list)
         cls.check_if_users_can_be_deleted(info, users, "ids", errors)
         cls.check_if_requestor_can_manage_users(info, users, "ids", errors)
-        cls.check_if_removing_left_not_manageable_permissions(users, "ids", errors)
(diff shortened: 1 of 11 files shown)


saleor/graphql/account/bulk_mutations.py
```diff
@@ -39,8 +42,31 @@ class Meta:
...
+
+    @classmethod
+    def clean_instances(cls, info, users):
+        errors = defaultdict(list)
+        cls.check_if_users_can_be_deleted(info, users, "ids", errors)
+        cls.check_if_requestor_can_manage_users(info, users, "ids", errors)
+        cls.check_if_removing_left_not_manageable_permissions(
+            info, users, "ids", errors
+        )
+        return ValidationError(errors) if errors else {}
```
tomaszszymanski129, line 69: Function can return either instance of ValidationError or a dictionary so that doesn't seem like a good pattern...

IKarbowiak: I can do it here, or move to `perform_mutation`. But still `perform_mutation`will return `ValidationError` or empty dict. But this is how our `BaseBulkMutations` works. So refactor should be done in `BaseBulkMutations` first.

saleor/graphql/account/utils.py
```diff
@@ -73,46 +75,74 @@ class Meta:
...
+            msg = "Cannot delete a non-staff users."
+            code = AccountErrorCode.DELETE_NON_STAFF_USER
+            params = {"users": user_pks}
+            errors[field].append(ValidationError(msg, code=code, params=params))
 
     @classmethod
-    def check_if_requestor_can_manage_user(cls, info, instance):
-        if get_out_of_scope_users(info.context.user, [instance]):
-            msg = "You can't manage this user."
+    def check_if_requestor_can_manage_users(cls, info, instances, field, errors):
```
maarcingebala, line 108: If possible I would avoid passing `info` to utility functions. In this case, only `user` is needed from `info`. To make it less confusing with `users` we could consistently call it `current_user`.

maarcingebala: Also maybe `current_user` would sound better than `requestor`.

maarcingebala: And the third thing, we're missing a docstring here.

saleor/graphql/account/utils.py
```diff
@@ -73,46 +75,74 @@ class Meta:
...
+        cls.check_if_removing_left_not_manageable_permissions(
+            info, [instance], "id", errors
+        )
+        if errors:
+            raise ValidationError(errors)
 
-        cls.check_if_requestor_can_manage_user(info, instance)
-        cls.check_if_removing_left_not_manageable_permissions(instance)
+    @classmethod
+    def check_if_users_can_be_deleted(cls, info, instances, field, errors):
```
maarcingebala, line 88: We should add a docstring explaining that it checks that we're not deleting non-staff users here.

PR #5457, opened 2020-04-08
Fix a race condition between Django and Celery

### LINKED ISSUE #5455: Having trouble with emails during order checkout
…
Email sent with the order summary

1. Register a new acc.
2. Pick desired items in the store.
3. Checkout with valid parameters.
4. Hit place order button.

### What I expected to happen
…
Celery worker sends email with order, instead it raises exception.

### Screenshots

image

**System information**
Operating system: Windows 10

patrys:
This delays the task until the transaction is committed thus making sure the object actually exists by the time the worker tries to fetch it. We should consider wrapping most if not all celery tasks in `transaction.on_commit`, it's safe to execute outside a transaction.

Fixes #5455 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -758,8 +758,12 @@ def create_order(
     order_created(order=order, user=user)
 
     # Send the order confirmation email
-    send_order_confirmation.delay(order.pk, redirect_url, user.pk)
-    send_staff_order_confirmation.delay(order.pk, redirect_url)
+    transaction.on_commit(
+        lambda: send_order_confirmation.delay(order.pk, redirect_url, user.pk)
+    )
+    transaction.on_commit(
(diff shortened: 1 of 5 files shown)


tests/api/utils.py
```diff
@@ -16,6 +18,7 @@ def get_graphql_content(response, *, ignore_errors: bool = False):
     permissions.
     """
     content = _get_graphql_content_from_response(response)
+    flush_post_commit_hooks()
```
NyanKiyoshi, line 21: After running super of `tests.api.conftest.ApiClient.post_graphql` would probably be a better place?

patrys: Good idea, updated.

PR #5460, opened 2020-04-09
Raise GraphQLError if filter has invalid values

### LINKED ISSUE #5416: Filter by ids in query.
When doing any query with filter by ids and provide invalid id query return all objects instead of all matching form ids in the input. 

Working query on [demo](https://pwa.demo.saleor.rocks/graphql/)
```
{
  categories(first:10, filter:{ids: ["Q2F0ZWdvcnk6MTU=", "Q2F0ZWdvcnk6MTI="]}){
    edges{
      node{
        id
      }
    }
  }
}
```
For not working query chage ids to `["Q2F0ZWdvcnk6MTU=", "Q2F0ZWdvcnkxMg=="]`
Secound query retrun all categories.

gabmartinez:
I want to merge this change because it fixes the issue that was returning all objects when the ids are not valid wherever we use the GlobalIDMultipleChoiceFilter class.

fixed #5416

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/graphql/core/fields.py
```diff
@@ -186,9 +187,14 @@ def connection_resolver(
...
-            iterable = filterset_class(
+            instance = filterset_class(
                 data=dict(filter_input), queryset=iterable, request=info.context
-            ).qs
+            )
+            # Make sure filter input has valid values
+            if not instance.is_valid():
+                raise ValidationError(instance.errors, code="invalid_list")
+            else:
+                iterable = instance.qs
```
koradon, line 197: This `else` is not needed

tests/api/test_attributes.py
```diff
@@ -98,6 +98,32 @@ def test_attributes_query(user_api_client, product):
...
+            }
+        }
+    }
+    """
+
+
+@pytest.mark.parametrize(
+    "attribute_filter", [{"ids": ["ygRqjpmXYqaTD9r=", "PBa4ZLBhnXHSz6v="]}],
+)
+def test_attribute_not_exit_ids(attribute_filter, user_api_client, category):
```
koradon, line 118: I would change def's to `test_attributes_query_ids_not_exists`.

saleor/graphql/core/fields.py
gabmartinez: good catch. thanks :)

tests/api/test_attributes.py
gabmartinez: done. thank you 👍

saleor/graphql/core/fields.py
```diff
@@ -186,9 +187,13 @@ def connection_resolver(
...
 
         if filter_input and filterset_class:
-            iterable = filterset_class(
+            instance = filterset_class(
                 data=dict(filter_input), queryset=iterable, request=info.context
-            ).qs
+            )
+            # Make sure filter input has valid values
+            if not instance.is_valid():
+                raise ValidationError(instance.errors, code="invalid_list")
```
NyanKiyoshi, line 195: Please check whether that error is handled or not by GraphQL.

In order to achieve this, trigger the error and look at the server logs, if you see something saying the error was not handled, "query failed unexpectedly" it means you have to raise a `GraphQLError` instead:
```
ERROR saleor.graphql.errors.unhandled A query failed unexpectedly [PID:18114:Thread-3]
```

maarcingebala: I was going to write, that we had discussion about what type of errors should we return in queries and it seems that in this case `GraphQLError` is the best one. Validation errors are rather for mutations where errors can be part of the response.

gabmartinez: gotcha. Working on it. Thank you :)

gabmartinez: Take a look again, please.

Thanks 👍

gabmartinez: Btw, using `instance.errors` rather than a custom message (You must provide valid fields.), it returns a message like this `<ul class=\"errorlist\"><li>ids<ul class=\"errorlist\"><li>Invalid ID specified.</li></ul></li></ul>`. Let me know which one do we prefer to use?

saleor/graphql/core/fields.py
```diff
@@ -186,9 +187,13 @@ def connection_resolver(
...
 
         if filter_input and filterset_class:
-            iterable = filterset_class(
+            instance = filterset_class(
                 data=dict(filter_input), queryset=iterable, request=info.context
-            ).qs
+            )
+            # Make sure filter input has valid values
+            if not instance.is_valid():
+                raise GraphQLError("You must provide valid fields.")
```
maarcingebala, line 195: I think this is fine, but was wondering if we can use `instance.errors` to show what was wrong with the values. The thing is it's an error dict and there may be multiple errors.

saleor/graphql/core/fields.py
maarcingebala: Yup, but we definitely shouldn't return HTML tags there. Maybe some something like `str(instance.errors)`. Also, error dict has methods like `as_data` etc. If we can easily figure out how to get these errors as a message, I would try to do that. Otherwise, we'll merge it as it is, because this is also good enough.

gabmartinez: cool, checking again. :)

saleor/graphql/core/fields.py
gabmartinez: Please check again. Using like this `GraphQLError(instance.errors.as_data())`,  we got a message error like this `{'ids': [ValidationError(['Invalid ID specified.'])]}`. :)

gabmartinez:
> Please check the errors in tests, you are expecting an error but not checking if that's actually the error you are expecting and wanting to happen
Cool, working on it. :)

saleor/graphql/core/fields.py
maarcingebala: I think it's fine now. In the future, I would think about improving it to return messages formatted as "field: message", but it's good enough already.

gabmartinez: I'm going to check if we already have a method to format those messages and if not what we can do. ;)

gabmartinez: Using `get_json_data` we got a result like this `{'ids': [{'message': 'Invalid ID specified.', 'code': ''}]}`. Please let me know if it better. ;)

tests/api/test_attributes.py
```diff
@@ -98,6 +98,35 @@ def test_attributes_query(user_api_client, product):
...
+    "attribute_filter", [{"ids": ["ygRqjpmXYqaTD9r=", "PBa4ZLBhnXHSz6v="]}],
+)
+def test_attributes_query_ids_not_exists(attribute_filter, user_api_client, category):
+    query = NOT_EXISTS_IDS_ATTRIBUTES_QUERY
+    variables = {"filter": attribute_filter}
+    response = user_api_client.post_graphql(query, variables)
+    content = get_graphql_content(response, ignore_errors=True)
+    message_error = "{'ids': [{'message': 'Invalid ID specified.', 'code': ''}]}"
+
+    assert content["errors"]
```
NyanKiyoshi, line 125: I would suggest doing the following:
```python
assert len(content["errors"]) == 1
```

This allow to ensure there are errors, and actually the amount that we are expecting

tests/api/test_attributes.py
```diff
@@ -98,6 +98,35 @@ def test_attributes_query(user_api_client, product):
...
+    assert content["errors"][0]["message"] == message_error
```
NyanKiyoshi, line 126: Are you sure about this? The message shouldn't be a string but a dict

gabmartinez: it's the data type that came from the response. if you want we can cast that string to a dictionary before comparing them in those tests.
Screen Shot 2020-04-09 at 12 05 12 PM

tests/api/test_attributes.py
gabmartinez: working on it. thanks :)

gabmartinez: Done. Please take a look again ;)

tests/api/test_attributes.py
NyanKiyoshi: So the issue is that GraphQL we _cannot_ tell GraphQL or the client that they can retrieve a `errors` field, like done in mutations. So we only can use the current structure and typings, that are:
```yaml
errors:
  - message: str
    locations:
      - line: int
        column: int
```

So we should return a string as message. And not a python representation of a dictionary (that's totally wrong, we should definitely not do that; and it is an invalid parameter type for GraphQL, it is expecting a string, not a dict).

It is the only possibility we have.

maarcingebala: It's valid - it is a dict but wrapped in a string. If @gabmartinez is willing to keep improving that, we can do that but let's agree on what format we want to have in the `message` field. I'm also fine with the current solution as it's better what we had.

gabmartinez: @maarcingebala @NyanKiyoshi I think we can create another task to work on that because maybe we are going to need more time to review what we can do.

NyanKiyoshi: In that case, if we do want a dict, we have to pass a valid JSON object, not a python representation of a Python dictionary, that's invalid JSON. On top, we need to pass it a string not as a dict to GraphQL, it is expecting a string.

Overall, I don't like the idea of passing a JSON object inside the error message, as it is a simple error telling the developer that they passed something invalid. We shouldn't add an expensive JSON serialization just to tell the developer they passed an invalid ID type.

gabmartinez: sorry, if I did wrong, but the `message_error` variable in those tests is not an argument that we are passing to GraphQL. We use it to compare the result content. Please let me know if that was you were worrying.

NyanKiyoshi: When you raise `GraphQLError(instance.errors.get_json_data())`, it is expecting a string, not a dict.

gabmartinez: Gotcha, it is correct. So, your suggestion is to put all of them in one message string. Quick question, if more than one field is not valid, which format should we use to return the message, separate them by a comma?

Like this? `field_1: message_1, field_2: message_2`

NyanKiyoshi: I don't understand the question. The signature is:
```
GraphQLError(message: str, nodes, ...)
```

gabmartinez: ok, thinking in the future if we have other fields to filter the result and more than one field is not valid in the input, how should we return the message error in one string. 

For example, the  `instance.errors` has a value like this `"{'ids': [ValidationError(['Invalid ID specified.'])]}, {'names': [ValidationError(['Invalid data type specified.'])]}"`, the response error message should be `ids: Invalid ID specified., names: Invalid data type specified.` in one string?

If it doesn't matter update it.

NyanKiyoshi: For now, dump the dict to JSON and pass it into `message`

gabmartinez: gotcha, one it. :)

tests/api/test_attributes.py
```diff
@@ -120,10 +121,11 @@ def test_attributes_query_ids_not_exists(attribute_filter, user_api_client, cate
     variables = {"filter": attribute_filter}
     response = user_api_client.post_graphql(query, variables)
     content = get_graphql_content(response, ignore_errors=True)
-    message_error = "{'ids': [{'message': 'Invalid ID specified.', 'code': ''}]}"
 
     assert len(content["errors"]) == 1
-    assert content["errors"][0]["message"] == message_error
+
+    error = json.loads(content["errors"][0]["message"])
+    assert error["ids"][0]["message"] == "Invalid ID specified."
```
NyanKiyoshi, line 128: Just compare directly the raw JSON string, no need to parse the JSON

gabmartinez: gotcha, reverting.

gabmartinez: Please check again :)

CHANGELOG.md
```diff
@@ -93,6 +93,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Check if image exists before validating - #5425 by @kswiatek92
 - Fix warehouses query not working without id - #5441 by @koradon
 - Add accountErrors to CreateToken as a required field - #5437 by @koradon
+- Raise GraphQLError if filter has not valid IDs - #5460 @gabmartinez
```
NyanKiyoshi, line 96:
```suggestion
- Raise GraphQLError if filter has not valid IDs - #5460 by @gabmartinez
```

gabmartinez: On it and thank you :)

gabmartinez: done 👍

tests/api/test_attributes.py
```diff
@@ -98,6 +98,35 @@ def test_attributes_query(user_api_client, product):
...
+            }
+        }
+    }
+    """
+
+
+@pytest.mark.parametrize(
+    "attribute_filter", [{"ids": ["ygRqjpmXYqaTD9r=", "PBa4ZLBhnXHSz6v="]}],
+)
+def test_attributes_query_ids_not_exists(attribute_filter, user_api_client, category):
```
fowczarek, line 118: Better to use `attribute_filter = {"ids": ["ygRqjpmXYqaTD9r=", "PBa4ZLBhnXHSz6v="]}`  insted of using `parametrize` with one parameter.

tests/api/test_category.py
```diff
@@ -681,6 +681,35 @@ def test_category_level(user_api_client, category):
...
+            }
+        }
+    }
+    """
+
+
+@pytest.mark.parametrize(
+    "category_filter", [{"ids": ["W3KATGDn3fq3ZH4=", "zH9pYmz7yWD3Hy8="]}],
+)
+def test_categories_query_ids_not_exists(category_filter, user_api_client, category):
```
fowczarek, line 701: Same issue with one parameter in parametrize.

tests/api/test_category.py
```diff
@@ -681,6 +681,35 @@ def test_category_level(user_api_client, category):
...
+        categories(first: 5, filter: $filter) {
+            edges {
+                node {
+                    id
+                    name
+                }
+            }
+        }
+    }
+    """
```
fowczarek, line 695:
```suggestion
"""
```

tests/api/test_attributes.py
```diff
@@ -98,6 +98,35 @@ def test_attributes_query(user_api_client, product):
...
+        attributes(first: 5, filter: $filter) {
+            edges {
+                node {
+                    id
+                    name
+                }
+            }
+        }
+    }
+    """
```
fowczarek, line 112:
```suggestion
"""
```

tests/api/test_collection.py
```diff
@@ -598,6 +598,35 @@ def test_remove_products_from_collection(
...
+            }
+        }
+    }
+    """
+
+
+@pytest.mark.parametrize(
+    "collection_filter", [{"ids": ["ncXc5tP7kmV6pxE=", "yMyDVE5S2LWWTqK="]}],
+)
+def test_collections_query_ids_not_exists(collection_filter, user_api_client, category):
```
fowczarek, line 618: Same issue with one parameter in parametrize.

tests/api/test_collection.py
```diff
@@ -598,6 +598,35 @@ def test_remove_products_from_collection(
...
+        collections(first: 5, filter: $filter) {
+            edges {
+                node {
+                    id
+                    name
+                }
+            }
+        }
+    }
+    """
```
fowczarek, line 612:
```suggestion
"""
```

tests/api/test_product.py
```diff
@@ -3444,6 +3444,37 @@ def test_product_type_query_with_sort(
...
+        productTypes(first: 5, filter: $filter) {
+            edges {
+                node {
+                    id
+                    name
+                }
+            }
+        }
+    }
+    """
```
fowczarek, line 3458:
```suggestion
"""
```

tests/api/test_product.py
```diff
@@ -3444,6 +3444,37 @@ def test_product_type_query_with_sort(
...
+
+
+@pytest.mark.parametrize(
+    "product_type_filter", [{"ids": ["fTEJRuFHU6fd2RU=", "2XwnQNNhwCdEjhP="]}],
+)
+def test_product_types_query_ids_not_exists(
```
fowczarek, line 3464: Same issue with one parameter in parametrize.

tests/api/test_attributes.py
gabmartinez: Done, please take a look again :)

maarcingebala:
Great job @gabmartinez, thanks a lot for your effort and contribution! 🚀

maarcingebala:
Thanks for your feedback @NyanKiyoshi @koradon @fowczarek.

gabmartinez:
> Thanks for your feedback @NyanKiyoshi @koradon @fowczarek.

Yes, thank you it was really helpful :)

Issue #5415, opened 2020-03-31
Errors in schema should be required.


fowczarek:
In PR #5293 we introduce a rule that all error fields should be non-nullable. 
We should fix error fields in mutations: 
- [ ] `SetPassword`
- [ ] `CreateToken`

PR #5465, opened 2020-04-09
Fix missing account_errors in CreateToken mutation

koradon:
This PR adds missing account_errors to CreateToken mutation when JSONWebTokenError is raised.
Adds new error code to AccountErrorCodes.
Adds tests.

#5415 
# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/account/error_codes.py
```diff
@@ -14,6 +14,7 @@ class AccountErrorCode(Enum):
     GRAPHQL_ERROR = "graphql_error"
     INVALID = "invalid"
     INVALID_PASSWORD = "invalid_password"
+    INVALID_CREDENTIALS = "invalid_credentials"
```
maarcingebala, line 17: We could actually drop `INVALID_PASSWORD` at all and replace it with `INVALID_CREDENTIALS` which is more universal and if needed, it can be paired with either email or password input field.

koradon: Do you want to do this in this PR?

maarcingebala: Up to you, whatever is easier/more convenient.

tests/api/test_core.py
```diff
@@ -288,6 +289,68 @@ def _run_test():
...
+
+
+def test_create_token(api_client, customer_user):
+    variables = {"email": customer_user.email, "password": customer_user._password}
+    response = api_client.post_graphql(MUTATION_CREATE_TOKEN, variables)
+    content = get_graphql_content(response)
+    user_email = content["data"]["tokenCreate"]["user"]["email"]
+    assert customer_user.email == user_email
+    assert content["data"]["tokenCreate"]["token"]
+    assert content["data"]["tokenCreate"]["errors"] == []
```
fowczarek, line 320: Checking only custom errors is enough we can remove this line.

tests/api/test_core.py
```diff
@@ -288,6 +289,68 @@ def _run_test():
...
+    content = get_graphql_content(response)
+    assert (
+        content["data"]["tokenCreate"]["errors"][0]["message"]
+        == "Please, enter valid credentials"
+    )
+    assert (
+        content["data"]["tokenCreate"]["accountErrors"][0]["code"]
+        == AccountErrorCode.INVALID_CREDENTIALS.value.upper()
+    )
+    assert content["data"]["tokenCreate"]["accountErrors"][0]["field"] == "email"
```
fowczarek, line 336: We shouldn't check messages form errors
```suggestion
    error = content["data"]["tokenCreate"]["errors"][0]
    assert error["code"] == AccountErrorCode.INVALID_CREDENTIALS.value.upper()
    assert error["field"] == "email"
```

tests/api/test_core.py
```diff
@@ -288,6 +289,68 @@ def _run_test():
...
+
+
+def test_create_token_invalid_email(api_client, customer_user):
+    variables = {"email": "wrongemail", "password": "wrongpassword"}
+    response = api_client.post_graphql(MUTATION_CREATE_TOKEN, variables)
+    content = get_graphql_content(response)
+    assert (
```
fowczarek, line 343: Same here :D

saleor/account/error_codes.py
koradon: Let's do this in next PR

tests/api/test_core.py
koradon: This is a check fo error code. We can check if it is not empty. Not sure what is the preferred option here.

fowczarek: We prefer checking `code` and `field` these fields should provide enough information to display error on storefronts/dashboard. Messages are information for developers. We should have possibility to change the message without changing tests.

In operation with return params in errors, we should check params too. For example `ProductVariantBulkCreate` contains `index` in error and this field needs to be checked in tests.

koradon: That's what I am checking here. AccountErrorCode and field.

koradon: Ow damn. Yes I left the message...

PR #5471, opened 2020-04-10
Drop permissions input field from staff mutations

IKarbowiak:
Drop permissions input field from staff mutations. Groups will be used instead.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/tests/api/benchmark/test_account.py
+++ b/tests/api/benchmark/test_account.py
@@ -6,7 +6,6 @@ from django.contrib.auth.models import Group
 from django.core.files import File
 
 from saleor.account.models import User
-from saleor.graphql.core.enums import PermissionEnum
 from tests.api.utils import get_graphql_content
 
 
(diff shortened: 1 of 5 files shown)


tests/api/benchmark/test_account.py
```diff
@@ -118,6 +117,75 @@ def test_query_staff_user(
...
+    }
+
+    staff_count = User.objects.filter(is_staff=True).count()
+    response = staff_api_client.post_graphql(
+        query, variables, permissions=[permission_manage_staff]
+    )
+    content = get_graphql_content(response)
+    data = content["data"]["staffCreate"]
+
+    assert User.objects.filter(is_staff=True).count() + staff_count + 1
```
tomaszszymanski129, line 184: :warning: This will always be true as count returns 0 when not found and we always add hardcoded 1

tests/api/benchmark/test_account.py
```diff
@@ -118,6 +117,75 @@ def test_query_staff_user(
...
+    assert data
```
tomaszszymanski129, line 185: we ask of so much data in the query and don't assert anything in here.

tests/api/test_account.py
```diff
@@ -1521,102 +1526,80 @@ def test_staff_create(
...
         data["user"]["avatar"]["url"],
     )
+
+    expected_perms = {
+        permission_manage_products.codename,
+        permission_manage_users.codename,
+    }
     permissions = data["user"]["userPermissions"]
-    assert permissions[0]["code"] == "MANAGE_PRODUCTS"
+    assert len(permissions) == 2
```
tomaszszymanski129, line 1535: no need to check the length if in line below we're comparing the whole object

tests/api/test_account.py
```diff
@@ -1521,102 +1526,80 @@ def test_staff_create(
...
+    assert {perm["code"].lower() for perm in permissions} == expected_perms
 
     # deprecated, to remove in #5389
     permissions = data["user"]["permissions"]
-    assert permissions[0]["code"] == "MANAGE_PRODUCTS"
+    assert len(permissions) == 2
```
tomaszszymanski129, line 1540: same here and in other places, can be safely removed.

tests/api/benchmark/test_account.py
IKarbowiak: It's benchmark test. Look at others, most of them don't check anything.

PR #5472, opened 2020-04-11
Fix: The products that are out of stock are not filtered as expected …

sebasgoldberg:

The products that are in stock, and out of stock, are not filtered as expected in the dashboard.

This modification:

- Validates the quantity in stock considering the quantity already allocated (if a product has quantity 5 and has a quantity allocated of 5 too, then the product should not be considered in stock).

- In case of an OUT_OF_STOCK query, if a variant is out of stock, then the product is considered out of stock.

Please note that with this modification, if a product P has two variants A and B, and the first one is in stock, and the second one is out of stock, then:

- An OUT_OF_STOCK query, returns P, because it as a variant out of stock.
- An IN_STOCK query, returns P too, because it as a variant in stock.

So, a product that has variants (more than 1), could be OUT_OF_STOCK and IN_STOCK at the same time.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested

### DIFF
--- a/saleor/graphql/product/filters.py
+++ b/saleor/graphql/product/filters.py
@@ -99,11 +99,10 @@ def sort_qs(qs, sort_by):
 
 
 def filter_products_by_stock_availability(qs, stock_availability):
-    qs = qs.annotate(total_quantity=Sum("variants__quantity"))
     if stock_availability == StockAvailability.IN_STOCK:
-        qs = qs.filter(total_quantity__gt=0)
+        qs = qs.filter(variants__quantity__isnull=False).extra(where=["product_productvariant.quantity > product_productvariant.quantity_allocated"]).distinct()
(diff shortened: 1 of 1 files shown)


maarcingebala:
Hey, thanks for opening the PR! We're just working on a change where stock allocation will be stored in a separate model: https://github.com/mirumee/saleor/pull/5445. This is related to our multi-warehouses feature. I'll wait with reviewing your PR since it's a lot of code that does the allocations is going to change.

@fowczarek We need to make sure that filtering that this PR fixes, works correctly in your PR.

maarcingebala:
I'm closing this PR as it refers to the legacy Saleor which we don't maintain anymore. Please check the [legacy-views](https://github.com/mirumee/legacy-views/) repo where Saleor 2.9.0 is archived.

PR #5485, opened 2020-04-15
Remove service account access to staff related operations

IKarbowiak:
Ensure that the service account doesn't have access to any operation/query protected by `MANAGE_STAFF`.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/core/mutations.py
+++ b/saleor/graphql/core/mutations.py
@@ -274,6 +274,7 @@ class BaseMutation(graphene.Mutation):
             return True
         service_account = getattr(context, "service_account", None)
         if service_account:
+            # for now MANAGE_STAFF permission for service account is not supported
             if AccountPermissions.MANAGE_STAFF in permissions:
                 return False
             return service_account.has_perms(permissions)
--- a/saleor/graphql/core/mutations.py
+++ b/saleor/graphql/core/mutations.py
@@ -18,6 +18,7 @@ from graphql_jwt.exceptions import JSONWebTokenError, PermissionDenied
 
 from ...account import models
 from ...account.error_codes import AccountErrorCode
+from ...core.permissions import AccountPermissions
 from ..account.types import User
 from ..utils import get_nodes
 from .types import Error, Upload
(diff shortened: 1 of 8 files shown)


saleor/graphql/core/mutations.py
```diff
@@ -269,8 +270,10 @@ def check_permissions(cls, context, permissions=None):
         if context.user.has_perms(permissions):
             return True
         service_account = getattr(context, "service_account", None)
-        if service_account and service_account.has_perms(permissions):
-            return True
+        if service_account:
+            if AccountPermissions.MANAGE_STAFF in permissions:
+                return False
+            return service_account.has_perms(permissions)
```
NyanKiyoshi, line 276: As a user, I give `Manage Staff` permission to my service account, then I try to manage staff, it doesn't work for some reasons. I open a ticket about this bug.

I believe it is a bad way of doing.

On top, a user, I would like to be able to manage the staff to perform automatic actions, such as adding new employees. Maybe putting a site settings or something that only shop owners can set?

IKarbowiak: Hm, during some daily or PR meetings we decided to don't allow service account to handle operations/queries which are protected by `MANAGE_STAFF`, so I am not sure it should be handle in settings. But maybe further discussion is needed. @maarcingebala @patrys

maarcingebala: @NyanKiyoshi It will be added in the future, for now we want to disable managing staff for service accounts.

NyanKiyoshi: Shouldn't we remove `Manage Staff` from the services?

saleor/graphql/core/mutations.py
```diff
@@ -272,8 +273,10 @@ def check_permissions(cls, context, permissions=None):
         if context.user.has_perms(permissions):
             return True
         service_account = getattr(context, "service_account", None)
-        if service_account and service_account.has_perms(permissions):
-            return True
+        if service_account:
+            if AccountPermissions.MANAGE_STAFF in permissions:
```
maarcingebala, line 277: I would add some comment here explaining that for now MANAGE_STAFF is not supported for service accounts.

NyanKiyoshi: Documenting it as well? Before we get bug tickets about it

maarcingebala: Yup we will take care of that in saleor-docs.

PR #5487, opened 2020-04-15
change extension permission to plugin permission

d-wysocki:
I want to merge this change because... after changing permission name from `extensions.manage_plugins` to `plugins.manage_plugins`, there was missing data migration for current users in database which already had this permission with old name.

# Impact

* [X] New migrations

# Pull Request Checklist

* [X] The changes are tested


saleor/account/migrations/0040_auto_20200415_0443.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 3.0.5 on 2020-04-15 09:43
+
+from django.db import migrations
+
+
+def change_extension_permission_to_plugin_permission(apps, schema_editor):
+    permission = apps.get_model("auth", "Permission")
+    users = apps.get_model("account", "User")
+    plugin_permission = permission.objects.get(
```
maarcingebala, line 9: I think this `get()` and the one below may be the reason for failing CI now. Instead maybe you should use `filter()[0]` with some null check; then based on that run the right logic.

NyanKiyoshi: `filter` then `first`

maarcingebala: Yup that what I actually meant ;)

PR #5494, opened 2020-04-17
Add home page

IKarbowiak:
Add view for a splash page and create a template with links to the playground, dashboard, and storefront. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/core/views.py
+++ b/saleor/core/views.py
@@ -1,0 +1,11 @@
+import os
+
+from django.template.response import TemplateResponse
+
+
+def home(request):
+    storefront = os.environ.get("STOREFRONT_URL", "")
+    dashboard = os.environ.get("DASHBOARD_URL", "")
+    return TemplateResponse(
+        request, "home.html", {"storefront": storefront, "dashboard": dashboard}
+    )
--- a/templates/home.html
+++ b/templates/home.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="UTF-8">
+  <title>It works!</title>
+</head>
+
+<body>
+  <a href={{dashboard}} target="_blank">Dashboard</a>
+  <a href={{storefront}} target="_blank">Storefront</a>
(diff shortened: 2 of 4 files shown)


fowczarek:
Are "STOREFRONT_URL" and "DASHBOARD_URL" set in saleor-platform?

IKarbowiak:
@fowczarek, there is PR for it right now: https://github.com/mirumee/saleor-platform/pull/24

templates/home.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="UTF-8">
+  <title>It works!</title>
+</head>
+
+<body>
+  <a href={{dashboard}} target="_blank">Dashboard</a>
```
fowczarek, line 10: Maby some if with an error if the link is empty?

maarcingebala: Yup these links should be probably be wrapped with some `if` and rendered only when the variable is set.

IKarbowiak: Right now, when dashboard and storefront urls are not set, it just stays at the home page.  So we should wrap this in if statement and show this elements only when variables are set?

maarcingebala: I think so. If the env is not set the link won't work anyway. Then we could even display some guide/message to tell the user how to start the service (but we don't have that on designs yet).

saleor/core/views.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+from django.template.response import TemplateResponse
+
+
+def home(request):
+    storefront = os.environ.get("STOREFRONT_URL", "")
```
tomaszszymanski129, line 7: If we don't have those set (for ex. someone cloned saleor and running it outside of docker) that will result in broken link in template. Maybe worth adding localhost:3000 as second arg?

IKarbowiak: The link will not be broken, links to the dashboard and storefront won't appear in this situation. If we put there default links and somebody doesn't have running instances of the dash and store it will cause broken links.

PR #5501, opened 2020-04-20
Fix: Change Payment Gateway

sebasgoldberg:

In case there is more than one option for payment gateway, once one of them is selected, it is not possible to select the other option. This change fix this problem.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/payment/utils.py
+++ b/saleor/payment/utils.py
@@ -108,7 +108,7 @@ def create_payment(
     if checkout is not None:
         data["checkout"] = checkout
 
-    payment, _ = Payment.objects.get_or_create(defaults=defaults, **data)
+    payment, _ = Payment.objects.get_or_create(gateway=gateway, defaults=defaults, **data)
     return payment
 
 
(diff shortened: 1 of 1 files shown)


saleor/payment/utils.py
```diff
@@ -108,7 +108,7 @@ def create_payment(
     if checkout is not None:
         data["checkout"] = checkout
 
-    payment, _ = Payment.objects.get_or_create(defaults=defaults, **data)
+    payment, _ = Payment.objects.get_or_create(gateway=gateway, defaults=defaults, **data)
```
korycins, line 111: `data` is a correct object to store items for `get_or_create`. Could you move the `gateway` there (to keep the consistency)?
Lines 99-104

korycins:
@sebasgoldberg  thanks for the fix! 
@maarcingebala - Do we want to merge any changes to `2.9` or we should merge it to `master`?

saleor/payment/utils.py
sebasgoldberg: @korycins, I have done the requested change.

PR #5509, opened 2020-04-20
Remove unused permission from db

d-wysocki:
I want to merge this change because...after renaming permission from `extension permission` to `plugin permission`, I want to delete permission with old name.
[#5487](https://github.com/mirumee/saleor/pull/5487)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested

### DIFF
--- a/saleor/account/migrations/0040_auto_20200415_0443.py
+++ b/saleor/account/migrations/0040_auto_20200415_0443.py
@@ -25,6 +25,9 @@ def change_extension_permission_to_plugin_permission(apps, schema_editor):
         user.user_permissions.remove(extension_permission)
         user.user_permissions.add(plugin_permission)
 
+    if extension_permission:
+        extension_permission.delete()
+
 
 class Migration(migrations.Migration):
 


saleor/account/migrations/0040_auto_20200415_0443.py
```diff
@@ -25,6 +25,9 @@ def change_extension_permission_to_plugin_permission(apps, schema_editor):
         user.user_permissions.remove(extension_permission)
         user.user_permissions.add(plugin_permission)
 
+    if extension_permission:
```
tomaszszymanski129, line 28: could simply be `extension_permission.delete()` without the `if` as it doesn't raise an exception in case of empty QS.

d-wysocki: I'm using `.first()` method on QS, and it's return None for empty QS and this will raise an exception

tomaszszymanski129: Aye, didn't see that, nvm.

PR #5529, opened 2020-04-23
Pairing product and ProductTranslation name fields

daltonsena:

I don't know if the difference between the length of these two fields is deliberated by the business rule. If this is the case this PR must be discarded, if not, I believe it is relevant that both fields have the same length

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


maarcingebala:
The descriptions in different languages may be of totally different length but it definitely doesn't make sense to limit the translation description to 128.

PR #5530, opened 2020-04-23
Fix/weight scalar

### LINKED ISSUE #5496: ProductCreate weight mutation
p1
p2

productCreate mutation

I followed the "amount unit" as said in the comment, but returns a "'StringValue(value='10.00 kg')' value must be a float."
Sorry just a beginner

koradon:
Right now WeightScalar is only working when the value is passed by variable to mutation.
This PR fixes that and adds support for passing value directly to mutation input (mostly used in playground).
Now accepted values are:
```
weight: 10.0
weight: 10
weight: "10"
weight: "10.0"
weight: {
    value: 10 // and all combinations from above
    unit: "kg"
}
```
For both variable and input

fixes #5496
# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/core/scalars.py
+++ b/saleor/graphql/core/scalars.py
@@ -2,6 +2,7 @@ import decimal
 
 import graphene
 from graphql.language import ast
+from graphql.error import GraphQLError
 from measurement.measures import Weight
 
 from ...core.weight import convert_weight, get_default_weight_unit
@@ -46,7 +47,7 @@ class WeightScalar(graphene.Scalar):
             default_unit = get_default_weight_unit()
             weight = Weight(**{default_unit: value})
         if not weight:
-            raise ValueError(f"Unsupported value: {value}")
+            raise GraphQLError(f"Unsupported value: {value}")
         return weight
 
     @staticmethod
@@ -66,7 +67,7 @@ class WeightScalar(graphene.Scalar):
         else:
             weight = WeightScalar.parse_literal_decimal(node)
         if not weight:
-            raise ValueError(f"Unsupported value: {node.value}")
+            raise GraphQLError(f"Unsupported value: {node.value}")
         return weight
 
     @staticmethod
@@ -88,7 +89,7 @@ class WeightScalar(graphene.Scalar):
                 try:
                     value = decimal.Decimal(field.value.value)
                 except decimal.DecimalException:
-                    return None
+                    raise GraphQLError(f"Unsupported value: {field.value.value}")
             if field.name.value == "unit":
                 unit = field.value.value
         return Weight(**{unit: value})
--- a/saleor/graphql/core/scalars.py
+++ b/saleor/graphql/core/scalars.py
@@ -1,6 +1,7 @@
 import decimal
 
 import graphene
+from graphql.language import ast
 from measurement.measures import Weight
 
 from ...core.weight import convert_weight, get_default_weight_unit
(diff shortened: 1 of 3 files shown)


tests/api/test_product.py
```diff
@@ -3961,3 +3961,184 @@ def test_update_or_create_variant_stocks_empty_stocks_data(variant, warehouses):
     stock = variant.stocks.first()
     assert stock.warehouse == warehouses[0]
     assert stock.quantity == 5
+
+
+# Because we use Scalars for Weight this test query tests only a scenario when weight
+# value is passed by a variable
+MUTATION_CREATE_PRODUCT_WITH_WEIGHT_GQL_VARIABLE = """
```
fowczarek, line 3968: Can we remove useless fields from the response in both mutation?

tests/api/test_product.py
```diff
@@ -3961,3 +3961,184 @@ def test_update_or_create_variant_stocks_empty_stocks_data(variant, warehouses):
...
+# value is passed by directly in input
+MUTATION_CREATE_PRODUCT_WITH_WEIGHT_GQL_INPUT = """
+mutation createProduct(
+        $productType: ID!,
+        $category: ID!
+        $name: String!,
+        $sku: String,
+        $basePrice: Decimal!
+        $trackInventory: Boolean)
+    {{
```
fowczarek, line 4022: Duble `{`

tests/api/test_product.py
```diff
@@ -3961,3 +3961,184 @@ def test_update_or_create_variant_stocks_empty_stocks_data(variant, warehouses):
...
+    category_id = graphene.Node.to_global_id("Category", category.pk)
+    product_type_id = graphene.Node.to_global_id(
+        "ProductType", product_type_without_variant.pk
+    )
+    variables = {
+        "category": category_id,
+        "productType": product_type_id,
+        "name": "Test",
+        "sku": "23434",
+        "trackInventory": True,
```
fowczarek, line 4088: Can we remove warehouse fixture, if we turn off `trackInventory`?

tests/api/test_product.py
koradon: This is needed if You want to use '{ }' in string with '.format()'.
We need to use '.format()' because we want to check if passing data directly in input weight field works.

fowczarek: Use `%` instead of format like in `test_query_countries_with_translation`. In this solution, we don't have doubled `{`.

koradon: `%` smells like 90ties. Is that a real problem? It is used only in this test and You can see for the first sight that it is different than the rest of defined queries so probably you shouldn't use it in any other test. And maybe we should move it inside the TC and use `f-strings` instead?

maarcingebala: It's no longer Python 2.7, let's use f-strings where we can.

tests/api/test_product.py
koradon: Done

tests/api/test_product.py
```diff
@@ -3961,3 +3961,161 @@ def test_update_or_create_variant_stocks_empty_stocks_data(variant, warehouses):
...
+    expected_weight_unit,
+    staff_api_client,
+    category,
+    permission_manage_products,
+    product_type_without_variant,
+):
+    # Because we use Scalars for Weight this test query tests only a scenario when
+    # weight value is passed by directly in input
+    MUTATION_CREATE_PRODUCT_WITH_WEIGHT_GQL_INPUT = f"""
+mutation createProduct(
```
fowczarek, line 4070: This line can start with the gap.

saleor/graphql/core/scalars.py
```diff
@@ -34,14 +36,19 @@ def parse_value(value):
...
+        weight = None
+        if isinstance(value, dict):
+            weight = Weight(**{value["unit"]: value["value"]})
+        else:
+            try:
+                value = decimal.Decimal(value)
+            except decimal.DecimalException:
+                return None
+            default_unit = get_default_weight_unit()
+            weight = Weight(**{default_unit: value})
```
IKarbowiak, line 48: Can't we use here `parse_literal_decimal` method?

koradon: Yes, we can

koradon: Done

maarcingebala:
@koradon CI failed

PR #5564, opened 2020-04-29
Negative weight validation

fowczarek:
I want to merge this change because adding negative weight validation and fix 0 value in weight scalar. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/product/mutations/products.py
+++ b/saleor/graphql/product/mutations/products.py
@@ -802,6 +802,18 @@ class ProductCreate(ModelMutation):
     @classmethod
     def clean_input(cls, info, instance, data):
         cleaned_input = super().clean_input(info, instance, data)
+
+        weight = cleaned_input.get("weight")
+        if weight and weight.value < 0:
+            raise ValidationError(
+                {
+                    "weight": ValidationError(
+                        "Product can't have negative weight.",
+                        code=ProductErrorCode.INVALID,
+                    )
+                }
+            )
+
         # Attributes are provided as list of `AttributeValueInput` objects.
         # We need to transform them into the format they're stored in the
         # `Product` model, which is HStore field that maps attribute's PK to
@@ -1158,6 +1170,17 @@ class ProductVariantCreate(ModelMutation):
     ):
         cleaned_input = super().clean_input(info, instance, data)
 
+        weight = cleaned_input.get("weight")
+        if weight and weight.value < 0:
+            raise ValidationError(
+                {
+                    "weight": ValidationError(
+                        "Product variant can't have negative weight.",
+                        code=ProductErrorCode.INVALID.value,
+                    )
+                }
+            )
+
         if "cost_price" in cleaned_input:
             cost_price = cleaned_input.pop("cost_price")
             if cost_price and cost_price < 0:
@@ -1419,6 +1442,18 @@ class ProductTypeCreate(ModelMutation):
     @classmethod
     def clean_input(cls, info, instance, data):
         cleaned_input = super().clean_input(info, instance, data)
+
+        weight = cleaned_input.get("weight")
+        if weight and weight.value < 0:
+            raise ValidationError(
+                {
+                    "weight": ValidationError(
+                        "Product type can't have negative weight.",
+                        code=ProductErrorCode.INVALID,
+                    )
+                }
+            )
+
         try:
             cleaned_input = validate_slug_and_generate_if_needed(
                 instance, "name", cleaned_input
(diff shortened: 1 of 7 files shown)


saleor/graphql/product/mutations/products.py
```diff
@@ -802,6 +802,18 @@ def clean_attributes(
     @classmethod
     def clean_input(cls, info, instance, data):
         cleaned_input = super().clean_input(info, instance, data)
+
+        weight = cleaned_input.get("weight")
+        if weight and weight.value < 0:
```
tomaszszymanski129, line 807: btw. can something have 0? if not that would have to be `<=`

saleor/graphql/product/mutations/products.py
```diff
@@ -1158,6 +1170,17 @@ def clean_input(
...
+            raise ValidationError(
+                {
+                    "weight": ValidationError(
+                        "Product variant can't have negative weight.",
+                        code=ProductErrorCode.INVALID.value,
+                    )
+                }
+            )
+
```
tomaszszymanski129, line 1183: can be extracted to something like `validate_weight` and used here and there as there are already 3 places with the same code

saleor/graphql/product/mutations/products.py
fowczarek: We talk about it yesterday, and in some cases, you want to set products (e.g. Post Cards) Weight to 0.

saleor/graphql/product/mutations/products.py
```diff
@@ -802,6 +802,18 @@ def clean_attributes(
...
     def clean_input(cls, info, instance, data):
         cleaned_input = super().clean_input(info, instance, data)
+
+        weight = cleaned_input.get("weight")
+        if weight and weight.value < 0:
+            raise ValidationError(
+                {
+                    "weight": ValidationError(
+                        "Product can't have negative weight.",
+                        code=ProductErrorCode.INVALID,
```
d-wysocki, line 812: should be `ProductErrorCode.INVALID.value`?

saleor/graphql/product/mutations/products.py
```diff
@@ -1419,6 +1442,18 @@ class Meta:
...
     def clean_input(cls, info, instance, data):
         cleaned_input = super().clean_input(info, instance, data)
+
+        weight = cleaned_input.get("weight")
+        if weight and weight.value < 0:
+            raise ValidationError(
+                {
+                    "weight": ValidationError(
+                        "Product type can't have negative weight.",
+                        code=ProductErrorCode.INVALID,
```
d-wysocki, line 1452: should be `ProductErrorCode.INVALID.value`?

saleor/graphql/shipping/mutations.py
```diff
@@ -207,6 +207,27 @@ def clean_input(cls, info, instance, data, input_cls=None):
             else:
                 min_weight = cleaned_input.get("minimum_order_weight")
                 max_weight = cleaned_input.get("maximum_order_weight")
+
+                if min_weight and min_weight.value < 0:
+                    raise ValidationError(
+                        {
+                            "minimum_order_weight": ValidationError(
+                                "Shipping can't have negative weight.",
+                                code=ShippingErrorCode.INVALID,
```
d-wysocki, line 216: should be `ShippingErrorCode.INVALID.value`?

saleor/graphql/shipping/mutations.py
```diff
@@ -207,6 +207,27 @@ def clean_input(cls, info, instance, data, input_cls=None):
...
+                            )
+                        }
+                    )
+
+                if max_weight and max_weight.value < 0:
+                    raise ValidationError(
+                        {
+                            "maximum_order_weight": ValidationError(
+                                "Shipping can't have negative weight.",
+                                code=ShippingErrorCode.INVALID,
```
d-wysocki, line 226: should be `ShippingErrorCode.INVALID.value`?

saleor/graphql/product/mutations/products.py
fowczarek: But errors are different. Second, think is that we have more duplicated logic in `clean_input` in ProductVariantCreate, ProductTypeCreate, and ProcutCreate, We should consider refactor it globally instead of having some logic duplicated and some logic hidden in functions.

saleor/graphql/product/mutations/products.py
tomaszszymanski129: Well, post cards have some not-null weight but I guess there might be someone that sells virtual goods among physical products so that makes sense anyway.

saleor/graphql/product/mutations/products.py
fowczarek: I try to use ErrorCode without `.value` in all places where `mypy` don't rise error.

fowczarek: We shouldn't use `.value` if `mypy` don't raise an error.

PR #5566, opened 2020-04-29
Invoice PDF multi page support

tomaszszymanski129:
I want to merge this change because it adds multi page support for invoice PDFs
Also contains some visual adjustments
Tested all cases to see whether products are bring sliced properly and pages don't glitch.

Sample PDF: 
~~[9281ff95-9fdf-429f-9423-23983022316f.pdf](https://github.com/mirumee/saleor/files/4586483/9281ff95-9fdf-429f-9423-23983022316f.pdf)~~
~~[0f1d501a-96fd-404e-8ca6-4742996bb174.pdf](https://github.com/mirumee/saleor/files/4591904/0f1d501a-96fd-404e-8ca6-4742996bb174.pdf)~~
~~[f0fe58b5-0457-47c1-b2e1-285ca0146fc1.pdf](https://github.com/mirumee/saleor/files/4592515/f0fe58b5-0457-47c1-b2e1-285ca0146fc1.pdf)~~
~~[41d45631-4839-4ef2-ae85-88bc5aeebdf1.pdf](https://github.com/mirumee/saleor/files/4608947/41d45631-4839-4ef2-ae85-88bc5aeebdf1.pdf)~~
[4cd604dd-a96d-41a4-9f88-0cb472e144ce.pdf](https://github.com/mirumee/saleor/files/4627863/4cd604dd-a96d-41a4-9f88-0cb472e144ce.pdf)

Image from iOS

TODO:
* [X] switch logo to what we have on design
* [X] Fix date not showing up and show it in correct format
* [X] Add some test coverage
* [X] Increase template paging logic readability a bit

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested


IKarbowiak:
Why `Date` is `None` in the sample PDF?

NyanKiyoshi:
Weird, it's not nullable. Good question.

tomaszszymanski129:
PDF might contain some gibberish since that worked on mocked data however I will take a look

fowczarek:
In PDF example of Figma we have:
1. different saleor logo 
2. The invoice information has a background
3. separator between parts
4. separator between order line 

@pwgryglak I thinking about margins in PDF file. From my perspective, they are pretty small. I don't know much about printing invoices. Can you say anything about this topic?

saleor/order/utils.py
```diff
@@ -342,10 +342,41 @@ def match_orders_with_new_user(user: User) -> None:
...
+        return 1
+
+
 def generate_invoice_pdf_for_order(invoice):
     logo_path = static_finders.find("images/logo-light.svg")
+
+    all_products = invoice.order.lines.all()
+    _product_limit = 3 if len(all_products) < 4 else 4
+    products_first_page = all_products[:_product_limit]
+    rest_of_products = _chunk_products(all_products[_product_limit:], 13)
```
maarcingebala, line 369: There is a lot of magic numbers here; I'm wondering if we can come up with some meaningful names for them (at least for some of them) and make them constants. E.g. this `13` could become something as `MAX_PRODUCTS_PER_PAGE` and the number `4` above `MAX_PRODUCTS_FIRST_PAGE`, what do you think?

tomaszszymanski129: Good idea, I know this logic is quite enigmatic, will try to make it more explicit.

maarcingebala:
Isn't the font size too large on the invoice? It feels so for the products list. Also, I'm wondering if it's not a problem that we're trimming the product (variant) name:
image

saleor/order/utils.py
```diff
@@ -342,10 +342,41 @@ def match_orders_with_new_user(user: User) -> None:
...
 
+def _chunk_products(products, product_limit):
+    chunks = []
+    for i in range(0, len(products), product_limit):
+        limit = i + product_limit
+        chunks.append(products[i:limit])
+    return chunks
+
+
+def _get_page_count(products_first_page, rest_of_products):
```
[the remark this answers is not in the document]
fowczarek: Did you check the test coverage of these functions?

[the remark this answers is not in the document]
tomaszszymanski129: There is none coverage for these since the func is not connected anywhere at this point. Will add some units.

tomaszszymanski129:
@pwgryglak please advise what to do when product name is too long to fit in single row of product in products table.

tomaszszymanski129:
Managed to use auto-page numbering using CSS print rules, code is much less spaghetti now! :tada:

saleor/order/utils.py
```diff
@@ -343,10 +347,67 @@ def match_orders_with_new_user(user: User) -> None:
...
+
+    return MAX_PRODUCTS_WITHOUT_TABLE
+
+
 def generate_invoice_pdf_for_order(invoice):
-    logo_path = static_finders.find("images/logo-light.svg")
+    logo_path = static_finders.find("images/logo.svg")
+    MAX_PRODUCTS_PER_PAGE = 13
+
+    if not getattr(invoice, "order"):
```
fowczarek, line 376: This validation should be part of mutation instead of part of the utils. 
When using som validation in utils we shouldn't use "GraphQL errors".

pwgryglak:
@fowczarek - margins are less problematic here as PDFs will be treated as

pwgryglak:
@fowczarek There shouldn't be any problem with printing this design. See attachment
@tomaszszymanski129 - PDF prints with background for some reason. See attachment.
@maarcingebala - We can use smaller fonts for the PDF. I've just printed the final pdf Tomasz provided and we can go down. See designs in Figma.

To do list for Tomasz:
- [ ] Change font size according to design
- [ ] Regulate spacings for smaller fonts according to design
- [ ] Figure out the pdf background issue

97161669_2610220952584698_7879868037162074112_n

saleor/order/utils.py
maarcingebala: Agree, `ValidationError`s should ideally live in the GraphQL layer.

tomaszszymanski129: Yeah, furthermore I think I might have covered that already, will remove it from here.

tomaszszymanski129:
@pwgryglak 
The background I've set to what we have on design, didn't think much about printing - yep that would leave a margin on most of printers because of machine margin limitations, will delete the background color property so we won't have that problem.

About the font sizes - what should we set them to? I mean, is the design A4? It doesn't seem like A4 since for example the "INVOICE INFORMATION" in design is 12px and on A4 that looks small, that's the reason behind enlarging it a bit.

pwgryglak:
@tomaszszymanski129 
Design was created specifically for A4 paper. In case of changing font sizes it would be great if you could provide the information that you changed something to me as I'm responsible for the design. Changes that you made without giving me any information about it - I can't make the adjustments to things I don't know about. There is a probability that if they weren't made @maarcingebala feedback would be obsolete. Now I have changed the design based on feedback to an implementation, that was not done as designed.
I would say that the background issue is not fault of the design as it was always set as #FFFFFF.

tomaszszymanski129:
All right will shortly prepare a PDF where we have exact copy of design properties

saleor/order/utils.py
```diff
@@ -343,10 +343,50 @@ def match_orders_with_new_user(user: User) -> None:
...
+
+    if len(products) < MAX_PRODUCTS_WITHOUT_TABLE:
+        return MAX_PRODUCTS_WITH_TABLE
+
+    return MAX_PRODUCTS_WITHOUT_TABLE
+
+
 def generate_invoice_pdf_for_order(invoice):
-    logo_path = static_finders.find("images/logo-light.svg")
+    logo_path = static_finders.find("images/logo.svg")
```
korycins, line 369: Shouldn't be a part of our basic Invoice plugin? Will Saleor use these functions somewhere else? If not I assume that we should move them to InvoicePlugin logic

saleor/order/utils.py
```diff
@@ -343,10 +343,50 @@ def match_orders_with_new_user(user: User) -> None:
...
+    """
+    chunks = []
+    for i in range(0, len(products), product_limit):
+        limit = i + product_limit
+        chunks.append(products[i:limit])
+    return chunks
+
+
+def get_product_limit_first_page(products):
+    MAX_PRODUCTS_WITH_TABLE = 3
```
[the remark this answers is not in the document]
tomaszszymanski129: Since that is a constant, I decided to leave it this way but I'm open for suggestions

saleor/order/utils.py
tomaszszymanski129: Can be moved, perhaps will do so, but in a separate PR when I will have the logic behind core plugin (generating invoice by Saleor)

korycins: Okay, so maybe a gh issue to remember about it? No need to keep pdf logic in saleor core as we will use it only inside a plugin.

tomaszszymanski129:
Re-worked entire PDF, discussed with @pwgryglak, it looks fine now.
Few notes:
- weasyprint scales the PDF a bit, for example font 14 is equivalent to 11PX height in resulting PDF if you measure it by any photo manipulation program. Dindn't manage to find the root cause but reworked the sizes so it looks right.
- All fonts have been adjusted so they are nearly pixel-pixel like on design
- Used the Inter font as we use it on our site, downloaded from official google site, license is Open Font License.
- Font is included in resulting PDF so there is no need to set fallback font
- One thing left to do - with very long product names i decided to leave it trimmed for now - computing the height of a row, based on how many lines each product have would be a nightmare to fit it in multi-pages without cutting them improperly. It will be a nightmare but not impossible nightmare - let's leave it for post-mvp.

PR #5573, opened 2020-04-30
External App support

korycins:
- query `AppInstallations`
- `AppInstall` mutation
- job for handling app installation
- `appRetryInstall` mutation
- `appDeleteFailedInstallation` mutation
- extend App model by new fields

---
#### App installation via API
mermaid-diagram-20200512135931

---
#### Installation job flow
mermaid-diagram-20200512140046


saleor/graphql/app/mutations.py
```diff
@@ -220,3 +206,133 @@ def clean_instance(cls, info, instance):
...
+            for perm in perms
+        ]
+        requestor = get_user_or_app_from_context(info.context)
+        ensure_can_manage_permissions(requestor, perms)
+
+
+class InstallAppInput(graphene.InputObjectType):
+    name = graphene.String(description="Name of the app to install.")
+    manifest_url = graphene.String(description="Url to app's manifest in JSON format.")
+    activate_after_installation = graphene.Boolean(default_value=True, required=False)
```
fowczarek, line 290: Missing description :policeman:

saleor/graphql/app/schema.py
```diff
@@ -17,6 +26,9 @@ class Meta:
 
 
 class AppQueries(graphene.ObjectType):
+    ongoing_apps_installations = PrefetchingConnectionField(
```
fowczarek, line 29: Is pagination required here?

saleor/app/tasks.py
```diff
@@ -0,0 +1,27 @@
...
+from .models import AppJob
+
+
+@celeryconf.app.task
+def install_app_task(job_id, activate=False):
+    app_job = AppJob.objects.get(id=job_id)
+    try:
+        install_app(app_job, activate=activate)
+        app_job.delete()
+        return
```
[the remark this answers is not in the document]
korycins: Not sure If I understand it. So you propose to remove the `return` statement and add bool flag `success=True`?
This approach will require additional `if` condition.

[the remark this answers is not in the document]
patrys: Early returns are more readable than putting most of the code in `else`.

saleor/graphql/app/schema.py
NyanKiyoshi: Agreed

saleor/graphql/app/schema.py
```diff
@@ -26,7 +26,7 @@ class Meta:
 
 
 class AppQueries(graphene.ObjectType):
-    ongoing_apps_installations = PrefetchingConnectionField(
+    ongoing_apps_installations = graphene.List(
         OngoingAppInstallation, description="List of all ongoing apps installations"
```
fowczarek, line 30: @dominik-zeglen  Should it be `NonNull` list?

dominik-zeglen: I don't see any reason why there should be used null or list of nulls 🤷‍♂️

saleor/graphql/app/utils.py
```diff
@@ -0,0 +1,23 @@
+from django.core.exceptions import ValidationError
+
+from ..account.utils import get_out_of_scope_permissions
+from ..core.enums import AppErrorCode
+from ..utils import requestor_is_superuser
+
+
+def ensure_can_manage_permissions(requestor, permission_items):
```
IKarbowiak, line 8: We have the same functionality in the few more places. Maybe we should do this more generic, returning just ValidationError(error_msg, code=code, params=params) and handling codename in mutations (or we can pass codename as an argument).

saleor/app/models.py
```diff
@@ -61,3 +74,15 @@ class AppToken(models.Model):
     app = models.ForeignKey(App, on_delete=models.CASCADE, related_name="tokens")
     name = models.CharField(blank=True, default="", max_length=128)
     auth_token = models.CharField(default=generate_token, unique=True, max_length=30)
+
+
+class AppJob(Job):
+    name = models.CharField(max_length=60)
```
IKarbowiak, line 80: It's an app name or app job name?

saleor/graphql/app/utils.py
korycins: Yes, I was thinking about this, but it needs to be done as a separate PR as this branch will be merged into the feature branch. Will create a gh issue for that

saleor/app/models.py
korycins: changed to `app_name`

saleor/graphql/app/mutations.py
```diff
@@ -220,3 +206,139 @@ def clean_instance(cls, info, instance):
...
+        if not requestor_is_superuser(requestor) and not requestor.has_perms(
+            permissions
+        ):
+            msg = "You don't have enough permission to perform this action."
+            code = AppErrorCode.OUT_OF_SCOPE_APP.value
+            raise ValidationError({"id": ValidationError(msg, code=code)})
+
+        if instance.status != JobStatus.FAILED:
+            msg = "Cannot drop installation with different status than failed."
+            code = AppErrorCode.FORBIDDEN.value
```
maarcingebala, line 237: Maybe a more explicit error code like `INVALID_STATUS`? `FORBIDDEN` don't say too much.

saleor/app/types.py
```diff
@@ -0,0 +1,5 @@
+class AppType:
+    CUSTOM = "custom"
+    EXTERNAL = "external"
```
maarcingebala, line 3: Wondering if this will be clear enough but I don't have any better name :/

saleor/graphql/app/mutations.py
```diff
@@ -220,3 +206,139 @@ def clean_instance(cls, info, instance):
             msg = "You can't delete this app."
             code = AppErrorCode.OUT_OF_SCOPE_APP.value
             raise ValidationError({"id": ValidationError(msg, code=code)})
+
+
+class AppDropFailedInstallation(ModelDeleteMutation):
```
maarcingebala, line 211: To be consistent with the rest of API I would say `AppDeleteFailedInstallation` and replace all "drops" in descriptions below.

saleor/graphql/schema.graphql
```diff
@@ -256,6 +272,42 @@ input AppInput {
...
+}
+
+enum AppJobStatus {
+  PENDING
+  SUCCESS
+  FAILED
+  DELETED
+}
+
+type AppOngoingInstallation {
```
maarcingebala, line 295: I would suggest removing the word "ongoing" as this is actually determined by the job status. Just `AppInstallation` here and in queries and mutations would be short and clear.

saleor/graphql/schema.graphql
```diff
@@ -256,6 +272,42 @@ input AppInput {
...
+}
+
+input AppInstallInput {
+  appName: String
+  manifestUrl: String
+  activateAfterInstallation: Boolean = true
+  permissions: [PermissionEnum]
+}
+
+enum AppJobStatus {
```
maarcingebala, line 288: @IKarbowiak is working on CSV import/export and there we have generic enum for job statuses, couldn't we use it here? I'm trying to avoid a situation when we merged these two branches and end up with the same enum doubled in the schema with different names. 

https://github.com/mirumee/saleor/blob/add-products-csv-export/saleor/graphql/schema.graphql#L2035

IKarbowiak: Good point. `JobStatusEnum` is on master. https://github.com/mirumee/saleor/blob/master/saleor/graphql/core/enums.py#L79

saleor/app/installation_utils.py
```diff
@@ -0,0 +1,62 @@
...
+        is_active=activate,
+        identificator=manifest_data.get("identificator"),
+        about_app=manifest_data.get("about_app"),
+        data_privacy=manifest_data.get("data_privacy"),
+        data_privacy_url=manifest_data.get("data_privacy_url"),
+        homepage_url=manifest_data.get("homepage_url"),
+        support_url=manifest_data.get("support_url"),
+        configuration_url=manifest_data.get("configuration_url"),
+        app_url=manifest_data.get("app_url"),
+        version=manifest_data.get("version"),
```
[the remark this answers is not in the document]
patrys: `manifest_data` is an untrusted source, I'd rather not blindly pass this to our code.

saleor/graphql/schema.graphql
korycins: Now should be fine

saleor/graphql/schema.graphql
korycins: Done, Sir! 👮‍♂️

korycins:
@mirumee/saleor-core  Ready to review!

saleor/app/app_manifest_sample.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "saleor.app.sample",
+  "version": "1.0.0",
+  "name": "My Wonderful App",
+  "about": "My Wonderful App is a wonderful App for Saleor.",
```
maarcingebala, line 5: This file will be used as official template for app manifests, are you sure we want to have it?

saleor/app/error_codes.py
```diff
@@ -2,8 +2,10 @@
 
 
 class AppErrorCode(Enum):
+    FORBIDDEN = "forbidden"
```
maarcingebala, line 5: I think is it not used anymore.

saleor/app/types.py
fowczarek: If we don't have better names we can at least add a comment here. For example: 
Custom - App integrated through webhooks. 
External - App integrated through the python interface.

maarcingebala: Agree with Owczar.

patrys: I don't think these fit: `CUSTOM` is really "created by the staff" and `EXTERNAL` is "installed using a manifest".

saleor/app/app_manifest_sample.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "saleor.app.sample",
```
patrys, line 2: Please don't use `saleor` in examples, we don't want others to use that prefix.

saleor/app/types.py
maarcingebala: @patrys Any ideas for enum names? It's important to sort out naming issues early, as this later escalates to the entire codebase (plugin vs. extension vs. app).

patrys: Maybe `LOCAL` and `THIRDPARTY`?

saleor/app/app_manifest_sample.json
korycins: We have the same sample in docs. What is wrong with this?

PR #5585, opened 2020-05-04
Add error when user pass empty object as address. 

### LINKED ISSUE #5572: No errors returned when empty address set
I am trying to display errors in address form when user skips filling any input within it and I rely on errors returned by backend, but backend does not return any error.

1. Many possibilities:
- Create checkout with empty object set as shipping address (probably also billing address), or
- Update billing address with empty object set as billing address (probably also shipping address)
2. Response from backend contains checkout set to `undefined` and errors set to empty array

### What I expected to happen
I expect to get in response errors array with errors associated with required, but not properly filled fields or not filled at all.

### Screenshots

Zrzut ekranu 2020-04-30 o 14 27 08
Zrzut ekranu 2020-04-30 o 14 25 41
Zrzut ekranu 2020-04-30 o 14 32 32
Zrzut ekranu 2020-04-30 o 14 32 41

**System information**
Operating system: MacOS

fowczarek:
I want to merge this change because resolve #5572 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/account/i18n.py
+++ b/saleor/graphql/account/i18n.py
@@ -43,6 +43,10 @@ class I18nMixin:
 
     @classmethod
     def validate_address(cls, address_data: dict, instance=None, info=None):
+        if address_data.get("country") is None:
+            raise ValidationError(
+                {"country": ValidationError("This field is required.", code="required")}
+            )
         address_form = cls.validate_address_form(address_data)
         if not instance:
             instance = Address()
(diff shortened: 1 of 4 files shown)


saleor/graphql/account/i18n.py
```diff
@@ -43,6 +43,10 @@ def validate_address_form(cls, address_data: dict, instance=None):
 
     @classmethod
     def validate_address(cls, address_data: dict, instance=None, info=None):
+        if address_data.get("country") is None:
+            raise ValidationError(
+                {"country": ValidationError("This field is required.", code="required")}
```
tomaszszymanski129, line 48: Isn't there a pattern to use enums as error codes?

fowczarek: Yes, but we don't have `AddressErrorCode`. In some universal validation, we using strings instead of enums. E.g. `saleor/graphql/core/utils::validate_image_file`

maarcingebala: We should have a base enum for all error codes that we could use in cases like this.

PR #5602, opened 2020-05-06
Refactor read only middleware

maarcingebala:
This PR refactors `ReadOnlyMiddleware`:
- reimplemented as Graphene middleware instead of Django's which makes it a bit simpler
- removed outdated HTML template and checking Django paths (no longer needed since we've dropped static views)
- added `ROOT_EMAIL` feature (which was used on `demo` branch but not moved to `master`)
- removed deprecated meta mutations from `ALLOWED_MUTATIONS` (I think it's not necessary to allow users to test them in demo mode).

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/middleware.py
```diff
@@ -58,6 +59,51 @@ def app_middleware(next, root, info, **kwargs):
     return next(root, info, **kwargs)
 
 
+class ReadOnlyMiddleware:
+    ALLOWED_MUTATIONS = [
```
NyanKiyoshi, line 63: I noticed we don't allow users to register new accounts. Is it meant?

saleor/demo/settings.py
```diff
@@ -81,3 +81,6 @@ def before_send(event: dict, _hint: dict):
     sentry_sdk.init(
         DEMO_SENTRY_DSN, integrations=[DjangoIntegration()], before_send=before_send,
     )
+
+
+ROOT_EMAIL = os.environ.get("ROOT_EMAIL")
```
NyanKiyoshi, line 86: We already have it on line 43

saleor/demo/settings.py
```diff
@@ -81,3 +81,6 @@ def before_send(event: dict, _hint: dict):
     sentry_sdk.init(
         DEMO_SENTRY_DSN, integrations=[DjangoIntegration()], before_send=before_send,
```
NyanKiyoshi, line 82: We need to enable Celery too

saleor/graphql/middleware.py
maarcingebala: Yes it is, we don't want them to create accounts on demo.

saleor/demo/settings.py
maarcingebala: Fixed

saleor/demo/settings.py
maarcingebala: Good catch, fixed.

PR #5613, opened 2020-05-08
Add export events

IKarbowiak:
Add export events to csv export actions.

Flow diagram of current export:
image

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/csv/emails.py
```diff
@@ -14,13 +15,14 @@
 
 
 @app.task
-def send_email_with_link_to_download_csv(job: "ExportFile", template_name: str):
-    recipient_email = job.created_by.email
+def send_email_with_link_to_download_csv(export_file: "ExportFile", template_name: str):
+    recipient_email = export_file.created_by.email
```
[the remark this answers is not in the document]
IKarbowiak: It's a good question. In current implementation no, but maybe they should can.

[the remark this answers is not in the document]
IKarbowiak: But there will be no way to send them email with the url to download file.

saleor/csv/__init__.py
```diff
@@ -0,0 +1,19 @@
+class ExportEvents:
+    """The different csv events types."""
+
+    DATA_EXPORT_PENDING = "data_export_pending"
```
maarcingebala, line 4: Maybe just `EXPORT_PENDING`? A bit less typing and it still makes sense :)

saleor/csv/models.py
```diff
@@ -1,9 +1,32 @@
...
+
+
+class ExportEvent(models.Model):
+    """Model used to store events that happened during the export file lifecycle."""
+
+    date = models.DateTimeField(default=timezone.now, editable=False)
+    type = models.CharField(
+        max_length=255,
+        choices=[
+            (type_name.upper(), type_name) for type_name, _ in ExportEvents.CHOICES
```
maarcingebala, line 23: Do we have to unwrap it like that? In few other models where we have `choices` we just do `choices=SomeClass.CHOICES`.

saleor/graphql/schema.graphql
```diff
@@ -1756,13 +1756,30 @@ type Error {
   message: String
 }
 
+type ExportEvent implements Node {
+  id: ID!
+  date: DateTime
+  type: ExportEventsEnum
+  user: User
+  message: String
```
maarcingebala, line 1764: I think all of these fields are required and should be marked so in the schema.

saleor/csv/models.py
IKarbowiak: It depends. In ExportEvents, as the second argument in CHOICES, there is a description of Enum field. So without wrap, the choices will be like:
```choices=[('export_pending', 'Data export was started.'), ('export_success', 'Data export was completed successfully.')]```
instead of
```choices=[("EXPORT_PENDING", "export_pending"), ("EXPORT_SUCCESS", "export_success")]``` .

I based on `OrderEvent`.

Comment (maarcingebala) [reply saleor/csv/models.py:23]: Yes, I saw the `OrderEvent` uses the same pattern but I'm wondering why it's done like that.  From the two examples you gave above I thought we want the first one. When using `choices` the second item in the choice tuple is supposed to be a human-readable description.

File: saleor/csv/events.py [hunk 5]
```diff
@@ -0,0 +1,41 @@
...
+
+from ..account.models import User
+from . import ExportEvents
+from .models import ExportEvent
+
+if TYPE_CHECKING:
+    from .models import ExportFile
+
+
+def export_event(*, export_file: "ExportFile", user: User):
```
maarcingebala, line 11: One last change - I would rename this function to `export_started_event` - it will be more consistent with other functions in this module. Now it sounds like this function was used to export some events which is not exactly true :)

saleor/csv/models.py
IKarbowiak: :information_desk_person:
I will change it in ExportEvent.

Issue #4955, opened 2019-11-08
Create django view for plugins


korycins:
Some plugins require two ways of communication. Most of the services use rest to communicate so we can't use our API. We need to have a generic view like `/integrations/<slug:plugin>/` which will call method for the given plugin(via ExtensionManager). The plugin will handle the request based on its custom logic.
Related to #4939

PR #5619, opened 2020-05-10
Add url handler for plugins

korycins:
Some plugins (especially payment gateways) require two ways of communication. This PR adds possibility to extend Saleor.urls by custom URLs defined by plugins.
This is an implementation for #4955

It adds:
- new flag to `BasePlugin` - `URL_ID` which is required to generate URL path for plugin
- new method to `BasePlugin` - `register_urls` which should return a list of Django URLs.

The structure of URLs will be as below:
 - `/plugins/<plugin.URL-ID>/register_view1` - where `register_view1` is view defined by a given plugin. 

Example of possible urls based on Braintree:
 - `plugins/braintree/webhooks/handle_payment`

### DIFF
--- a/tests/plugins/test_urls.py
+++ b/tests/plugins/test_urls.py
@@ -1,0 +1,30 @@
+from saleor.plugins.urls import register_plugins_urls
+
+
+def test_register_plugins_urls(settings):
+    settings.PLUGINS = ["tests.plugins.sample_plugins.PluginSample"]
+
+    urls = register_plugins_urls()
+
+    assert len(urls) == 1
+    plugin_resolver = urls[0]
+    assert len(plugin_resolver.url_patterns) == 2
+
+    # direct call on index list because we know position of the patterns.
+    test_request = plugin_resolver.url_patterns[0]
+    test_request_detail = plugin_resolver.url_patterns[1]
+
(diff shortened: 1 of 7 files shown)


NyanKiyoshi:
> This is an implementation for #4955

Seems like this PR closes it?

tests/plugins/test_urls.py
```diff
@@ -0,0 +1,30 @@
...
+        "tests.plugins.sample_plugins.PluginSample."
+        "register_urls.<locals>.handle_request"
+    )
+    assert test_request_detail.lookup_str == (
+        "tests.plugins.sample_plugins.PluginSample."
+        "register_urls.<locals>.handle_detail_request"
+    )
+
+
+def test_test_register_plugins_urls_plugins_without_urls(settings):
```
IKarbowiak, line 27: Double `test` in test name. 👮‍♀️

maarcingebala: 👮‍♀️

korycins:
I am closing it as we should use a different approach for this feature.
We should handle the request as a simple plugin method(instead of django view). No need to use django urls route. The manager should identify plugin by the name from URL and call method on the given plugin.

PR #5638, opened 2020-05-13
Activate and Deactivate App mutations

korycins:
- AppActivate mutation
- AppDeactivate mutation
- tests


tests/api/app/mutations/test_app_activate.py
```diff
@@ -0,0 +1,167 @@
...
+    variables = {
+        "id": id,
+    }
+    response = staff_api_client.post_graphql(
+        query, variables=variables, permissions=(permission_manage_orders,)
+    )
+    assert_no_permission(response)
+
+    app.refresh_from_db()
+    assert not app.is_active
```
IKarbowiak, line 98: This test is not needed, `post_graphql` is doing this check.

tests/api/app/mutations/test_app_deactivate.py
```diff
@@ -0,0 +1,167 @@
...
+    variables = {
+        "id": id,
+    }
+    response = staff_api_client.post_graphql(
+        query, variables=variables, permissions=(permission_manage_orders,)
+    )
+    assert_no_permission(response)
+
+    app.refresh_from_db()
+    assert app.is_active
```
IKarbowiak, line 98: Same here, `post_graphql` is doing this check.

saleor/graphql/app/mutations.py
```diff
@@ -208,6 +211,75 @@ def clean_instance(cls, info, instance):
...
+            msg = "App is already deactivated."
+            code = AppErrorCode.INVALID.value
+            raise ValidationError({"id": ValidationError(msg, code=code)})
+
+    @classmethod
+    def perform_mutation(cls, root, info, **data):
+        app = cls.get_instance(info, **data)
+        cls.clean_instance(info, app)
+        app.is_active = False
+        cls.save(info, app, cleaned_input=None)
```
fowczarek, line 279: Maby? 
```suggestion
        app.save(update_fields=["is_active"])
```

tests/api/app/mutations/test_app_activate.py
fowczarek: We have tasks to remove this "feature" from post_graphql.

saleor/graphql/app/mutations.py
```diff
@@ -208,6 +211,75 @@ def clean_instance(cls, info, instance):
...
+            permissions
+        ):
+            msg = "You don't have enough permission to perform this action."
+            code = AppErrorCode.OUT_OF_SCOPE_APP.value
+            raise ValidationError({"id": ValidationError(msg, code=code)})
+
+        if app.is_active:
+            msg = "App is already active."
+            code = AppErrorCode.INVALID.value
+            raise ValidationError({"id": ValidationError(msg, code=code)})
```
maarcingebala, line 239: I'm wondering, is it useful to raise this validation error? What if the mutation just did nothing and returned the unmodified object?

saleor/graphql/app/mutations.py
```diff
@@ -208,6 +211,75 @@ def clean_instance(cls, info, instance):
...
+    @classmethod
+    def clean_instance(cls, info, app: App):
+        requestor = get_user_or_app_from_context(info.context)
+        permissions = app.permissions.all()
+        if not requestor_is_superuser(requestor) and not requestor.has_perms(
+            permissions
+        ):
+            msg = "You don't have enough permission to perform this action."
+            code = AppErrorCode.OUT_OF_SCOPE_APP.value
+            raise ValidationError({"id": ValidationError(msg, code=code)})
```
maarcingebala, line 234: I'm a bit lost on why we prefer `ValidationError` and not `PermissionDenied` here.

saleor/graphql/app/mutations.py
korycins: Not sure of this. ModelMutation has its own flow, if we are able then we should use it but you are right - having `update_fields` in `cls.save()` would be a good idea.

saleor/graphql/app/mutations.py
korycins: @maarcingebala I just aligned to the current flow here.
PermissionDenied is dedicated to raising when the user/app doesn't have the required permission for a given section. Here, we use ValidationError as this error is not connected only with the user. The user has access to Apps, but he can't manage Apps with bigger permissions than he has for his own.  
If you don't like it then what do you propose?

PR #5646, opened 2020-05-14
Add our implementation of UUID Scalar

### LINKED ISSUE #5008: Resolving a checkout by providing an invalid UUID should be handled
When running a query like the following:

```
{
  checkout(token: "Best-UUID") {
    created
  }
}
```

A `ValueError` will occur. We should whether return `None` or raise a validation error.

Note: we need to update the other queries and mutations that are using the `UUID` scalar.

koradon:
This PR gives us control over UUID scalar and changes thrown errors to `GraphQLError` when provided a wrong value for the token.

Fixes #5008 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/checkout/types.py
+++ b/saleor/graphql/checkout/types.py
@@ -110,7 +110,7 @@ class Checkout(CountableDjangoObjectType):
         TaxedMoney,
         description="The price of the checkout before shipping, with taxes included.",
     )
-    token = graphene.Field(UUID, description=("UUID of checkout."))
+    token = graphene.Field(UUID, description=("UUID of checkout."), required=True)
     total_price = graphene.Field(
         TaxedMoney,
         description=(
--- a/saleor/graphql/checkout/types.py
+++ b/saleor/graphql/checkout/types.py
@@ -8,6 +8,7 @@ from ...core.permissions import AccountPermissions, CheckoutPermissions
 from ...core.taxes import display_gross_prices, zero_taxed_money
 from ...plugins.manager import get_plugins_manager
 from ..core.connection import CountableDjangoObjectType
+from ..core.scalars import UUID
 from ..core.types.money import TaxedMoney
 from ..decorators import permission_required
 from ..discount.dataloaders import DiscountsByDateTimeLoader
@@ -109,6 +110,7 @@ class Checkout(CountableDjangoObjectType):
         TaxedMoney,
         description="The price of the checkout before shipping, with taxes included.",
     )
+    token = graphene.Field(UUID, description=("UUID of checkout."))
     total_price = graphene.Field(
         TaxedMoney,
         description=(
@@ -129,7 +131,6 @@ class Checkout(CountableDjangoObjectType):
             "quantity",
             "shipping_address",
             "shipping_method",
-            "token",
             "translated_discount_name",
             "user",
             "voucher_code",
--- a/saleor/graphql/checkout/types.py
+++ b/saleor/graphql/checkout/types.py
@@ -110,7 +110,7 @@ class Checkout(CountableDjangoObjectType):
         TaxedMoney,
         description="The price of the checkout before shipping, with taxes included.",
     )
-    token = graphene.Field(UUID, description=("UUID of checkout."), required=True)
+    token = graphene.Field(UUID, description=("The checkout's token."), required=True)
     total_price = graphene.Field(
         TaxedMoney,
         description=(
--- a/saleor/graphql/core/scalars.py
+++ b/saleor/graphql/core/scalars.py
@@ -88,3 +88,23 @@ class WeightScalar(graphene.Scalar):
             if field.name.value == "unit":
                 unit = field.value.value
         return Weight(**{unit: value})
+
+
+class UUID(graphene.UUID):
+    @staticmethod
+    def serialize(uuid):
+        return super(UUID, UUID).serialize(uuid)
+
+    @staticmethod
+    def parse_literal(node):
+        try:
+            return super(UUID, UUID).parse_literal(node)
+        except ValueError:
+            raise GraphQLError(f"Unsuported value: {node.value}")
+
+    @staticmethod
+    def parse_value(value):
+        try:
+            return super(UUID, UUID).parse_value(value)
+        except ValueError:
+            raise GraphQLError(f"Unsuported value: {value}")
--- a/saleor/graphql/core/scalars.py
+++ b/saleor/graphql/core/scalars.py
@@ -99,12 +99,12 @@ class UUID(graphene.UUID):
     def parse_literal(node):
         try:
             return super(UUID, UUID).parse_literal(node)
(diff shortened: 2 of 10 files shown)


saleor/graphql/schema.graphql
```diff
@@ -780,6 +779,7 @@ type Checkout implements Node & ObjectWithMetadata {
   lines: [CheckoutLine]
   shippingPrice: TaxedMoney
   subtotalPrice: TaxedMoney
+  token: UUID
```
fowczarek, line 782: Why token is not required now?

saleor/graphql/schema.graphql
```diff
@@ -1609,12 +1609,12 @@ input DigitalContentUploadInput {
 }
 
 type DigitalContentUrl implements Node {
-  token: UUID!
   content: DigitalContent!
   created: DateTime!
   downloadNum: Int!
   id: ID!
   url: String
+  token: UUID
```
fowczarek, line 1617: Why token is not required now?

tests/api/test_scalars.py
```diff
@@ -0,0 +1,44 @@
+from tests.api.test_checkout import QUERY_CHECKOUT
```
fowczarek, line 1: I thin we can duplicate the query here. In the near feature, we want to move tests next to code and it can be not good practice if someone changes mutation in checkout test and gets errors in scalar tests.

saleor/graphql/schema.graphql
koradon: Good catch

tests/api/test_scalars.py
koradon: Done

saleor/graphql/product/types/digital_contents.py
```diff
@@ -4,17 +4,21 @@
...
 from ...core.connection import CountableDjangoObjectType
+from ...core.scalars import UUID
 from ...decorators import permission_required
 from ...meta.deprecated.resolvers import resolve_meta, resolve_private_meta
 from ...meta.types import ObjectWithMetadata
 
 
 class DigitalContentUrl(CountableDjangoObjectType):
     url = graphene.String(description="URL for digital content.")
+    token = graphene.Field(
```
fowczarek, line 15: @maarcingebala If we use toke as a custom field, should we add resolver for this field?

koradon: This only overwrites the default UUID field to our UUID which is default UUID with our errors, so probably not.

maarcingebala: If it works without a custom resolver then we don't need it.

saleor/graphql/checkout/types.py
```diff
@@ -109,6 +110,7 @@ class Checkout(CountableDjangoObjectType):
         TaxedMoney,
         description="The price of the checkout before shipping, with taxes included.",
     )
+    token = graphene.Field(UUID, description=("UUID of checkout."), required=True)
```
maarcingebala, line 113: I would change it to have consistency with other description of that field above:
```suggestion
    token = graphene.Field(UUID, description=("The checkout's token."), required=True)
```

saleor/graphql/core/scalars.py
```diff
@@ -88,3 +88,23 @@ def parse_literal_object(node):
             if field.name.value == "unit":
                 unit = field.value.value
         return Weight(**{unit: value})
+
+
+class UUID(graphene.UUID):
+    @staticmethod
+    def serialize(uuid):
+        return super(UUID, UUID).serialize(uuid)
```
korycins, line 96: `super().serialize(uuid)` won't work?

saleor/graphql/core/scalars.py
```diff
@@ -88,3 +88,23 @@ def parse_literal_object(node):
             if field.name.value == "unit":
                 unit = field.value.value
         return Weight(**{unit: value})
+
+
+class UUID(graphene.UUID):
+    @staticmethod
+    def serialize(uuid):
```
korycins, line 95: All of these methods can raise `ValueError` and `TypeError`. Why do you catch only `ValueError`?

korycins:
All graphene's fields are catching the exceptions. Like here:
```
class Int(Scalar):
    """
    The `Int` scalar type represents non-fractional signed whole numeric
    values. Int can represent values between -(2^53 - 1) and 2^53 - 1 since
    represented in JSON as double-precision floating point numbers specified
    by [IEEE 754](http://en.wikipedia.org/wiki/IEEE_floating_point).
    """

    @staticmethod
    def coerce_int(value):
        try:
            num = int(value)
        except ValueError:
            try:
                num = int(float(value))
            except ValueError:
                return None
        if MIN_INT <= num <= MAX_INT:
            return num
```
For me, this PR is like workaround, we should introduce pull request to graphene repository with these changes as it seems it is not expected behaviour.
cc: @koradon  @maarcingebala

saleor/graphql/core/scalars.py
maarcingebala: Yes, in Python 3 we should use `super` without arguments as @korycins suggests. It should be changed here and in methods below as well.

maarcingebala:
@korycins Graphene is a bit inconsistent about it and we could open an issue and discuss it, but unfortunately, there seems to be a lot of open issues without any answers there :/
I think this solution is fine - it doesn't cost us much and makes our API more consistent.

korycins:
@maarcingebala Okay, I just thought that we could create an issue and create PR with `try except`  logic. The same class as we did here. Of course, this is not solution `for now` and we will need these changes.

saleor/graphql/core/scalars.py
koradon: @korycins @maarcingebala 
Well this was new for me too. According to this thread [https://bugs.python.org/issue31118](https://bugs.python.org/issue31118) super without arguments does not work with static methods by design. There are two possible solutions.
1. Pass class as I did above for both super arguments
2. Use `graphene.UUID.method(...)`
For me both are fine.

saleor/graphql/core/scalars.py
koradon: The issue was to throw ValidationError instead of ValueError. Since in other Scalars we throw GraphQLError I did that here too. Other changes were not requested and I can add them.

saleor/graphql/checkout/types.py
koradon: Done

saleor/graphql/core/scalars.py
maarcingebala: If the "Python 3 approach" doesn't work here I'm fine with the current solution.

saleor/graphql/core/scalars.py
maarcingebala: Do we know when a TypeError is raised? We know about ValueError and this is now handled.

PR #5648, opened 2020-05-14
Add missing quantize method on checkout in Avatax plugin

d-wysocki:
I want to merge this change because...closes https://github.com/mirumee/saleor/issues/5519

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested

### DIFF
--- a/saleor/checkout/calculations.py
+++ b/saleor/checkout/calculations.py
@@ -1,5 +1,6 @@
 from typing import TYPE_CHECKING, Iterable, Optional
 
+from ..core.taxes import quantize_price
 from ..discount import DiscountInfo
 from ..plugins.manager import get_plugins_manager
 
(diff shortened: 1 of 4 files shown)


korycins:
Wouldn't it be better to handle quantization on the level where Saleor runs plugins methods connected with a price? - like `saleor/checkout/calculations.py`. 
Thanks to it we will be sure that all plugin returns a quantized price.

maarcingebala:
Agree with @korycins, I would check if we can handle `quanatize` at the more general level.

d-wysocki:
> Wouldn't it be better to handle quantization on the level where Saleor runs plugins methods connected with a price? - like `saleor/checkout/calculations.py`.
> Thanks to it we will be sure that all plugin returns a quantized price.

Agree, I'm working on that.

PR #5652, opened 2020-05-15
Fix for extending an configuration of plugins

korycins:
- Fix for losing reference for configuration object in `get_plugin_configuration` method
- Vatlayer plugin - set active to false if used with VATLAYER_ACCESS_KEY variable
- Fix for saving data in plugin config
- Add tests

### DIFF
--- a/saleor/plugins/base_plugin.py
+++ b/saleor/plugins/base_plugin.py
@@ -352,7 +352,7 @@ class BasePlugin:
                         new_value = new_value.lower() == "true"
                     config_item.update([("value", new_value)])
 
-        # Get new keys that are not exists in current_config and extend it.
+        # Get new keys that don't exist in current_config and extend it.
         current_config_keys = set(c_field["name"] for c_field in current_config)
         configuration_to_update_dict = {
             c_field["name"]: c_field["value"] for c_field in configuration_to_update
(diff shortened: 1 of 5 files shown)


saleor/plugins/base_plugin.py
```diff
@@ -352,6 +352,22 @@ def _update_config_items(
                         new_value = new_value.lower() == "true"
                     config_item.update([("value", new_value)])
 
+        # Get new keys that are not exists in current_config and extend it.
```
maarcingebala, line 355: 🚔 Grammar police 🚔 
```suggestion
        # Get new keys that don't exist in current_config and extend it.
```

saleor/plugins/base_plugin.py
```diff
@@ -352,6 +352,22 @@ def _update_config_items(
...
+        current_config_keys = set(c_field["name"] for c_field in current_config)
+        configuration_to_update_dict = {
+            c_field["name"]: c_field["value"] for c_field in configuration_to_update
+        }
+        missing_keys = set(configuration_to_update_dict.keys()) - current_config_keys
+        for missing_key in missing_keys:
+            if not config_structure.get(missing_key):
```
maarcingebala, line 362: Not a big deal but it could be:
```
if config_structure.get(missing_key):
    current_config.append(
...
```
One less line and no `continue`.

saleor/plugins/base_plugin.py
korycins: 😱

PR #5656, opened 2020-05-16
Complete invoicing plugin

tomaszszymanski129:
I want to merge this change because it adds a full invocing plugin for invoce auto generation.

TODO:
* [X] Write and enable default invocing plugin
* [X] Modify RequestInvoice to return the status so frontend knows whether invoice needs to be created async (pending) or is ready. This will have to be discussed further.
* [X] Refactor the util for PDF creation, move it to newly created plugin
* [x] Bit more refactor to entire thing
* [x] Unit tests
* [X] Create legit URL instead of mocked one in case of serving the invoice when we crerated it locally

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested


saleor/graphql/schema.graphql
```diff
@@ -1959,6 +1959,7 @@ type Invoice implements Node {
   id: ID!
   number: String
   url: String
+  status: String!
```
tomaszszymanski129, line 1962: That's gonna get refactored to Job soon.

saleor/plugins/invoicing/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+    chunks = []
+    for i in range(0, len(products), product_limit):
+        limit = i + product_limit
+        chunks.append(products[i:limit])
+    return chunks
+
+
+def get_product_limit_first_page(products):
+    MAX_PRODUCTS_WITH_TABLE = 3
+    MAX_PRODUCTS_WITHOUT_TABLE = 4
```
fowczarek, line 53: Maby we should move all const outside functions?

saleor/plugins/invoicing/__init__.py
```diff
@@ -0,0 +1,93 @@
+import os
```
korycins, line 1: Maybe separate file would be a better idea than __init__

tests/conftest.py
```diff
@@ -180,6 +181,16 @@ def setup_vatlayer(settings):
     return settings
 
 
+@pytest.fixture
+def setup_invoicing(settings):
+    settings.PLUGINS = ["saleor.plugins.invoicing.plugin.InvoicingPlugin"]
+    data = {
+        "active": True,
```
korycins, line 188: Do you need to activate it? I though it is active by default

saleor/plugins/invoicing/__init__.py
tomaszszymanski129: Not my favorite approach, in typed language we would just put `const` before and leave it inside func as it's used only here and I like it more than putting it anywhere outside - though PEP8 suggest that's the way to do it so lets go with it.

tomaszszymanski129:
Closing in favor of https://github.com/mirumee/saleor/pull/5698, will resolve all issues there and let involved people know.

PR #5657, opened 2020-05-16
Invoice events

tomaszszymanski129:
I want to merge this change because it adds events to invoices.

# Impact

* [X] New migrations

# Pull Request Checklist

* [X] The changes are tested


saleor/graphql/order/mutations/orders.py
```diff
@@ -590,6 +595,12 @@ def perform_mutation(cls, _root, info, **data):
         invoice.order = instance
         invoice.status = InvoiceStatus.READY
         invoice.save()
+        events.invoice_created_event(
+            user=info.context.user,
+            invoice=invoice,
+            number=data["input"].get("number"),
+            url=data["input"].get("url"),
+        )
         return CreateInvoice(invoice=invoice)
```
fowczarek, line 604: The current solution allows executing mutations on draft orders. Should we allow this?

saleor/graphql/order/mutations/orders.py
```diff
@@ -626,6 +638,13 @@ class Meta:
         error_type_class = InvoiceError
         error_type_field = "invoice_errors"
 
+    @classmethod
+    def perform_mutation(cls, _root, info, **data):
+        invoice_pk = copy.copy(cls.get_instance(info, **data).pk)
```
fowczarek, line 643: Why we use `copy` here?

saleor/order/events.py
```diff
@@ -95,6 +95,53 @@ def draft_order_removed_products_event(
     )
 
 
+def invoice_requested_event(
```
korycins, line 98: This will fail if mutation will be triggered by App Token.

korycins: All your events methods will fail

maarcingebala: We had the same issue with CSV import/export. What approach did we agree on there? I think we decided to add it later and move the app support to another epic?

tomaszszymanski129: Just added a simple fail-safe for that which we use in other functions in that file

saleor/graphql/order/mutations/orders.py
tomaszszymanski129: Leftover, removing

saleor/graphql/order/mutations/orders.py
tomaszszymanski129: It's a good point, I thing generating an invoice should only be allowed for fullfilled order.

fowczarek: @maarcingebala For witch orders, we want to create invoices?

maarcingebala: For all orders except drafts? If an order was placed successfully, the customer should receive the invoice. If we can make any changes to orders we can also generate new invoices.

saleor/order/events.py
IKarbowiak: Yes, I created an epic for that.

tomaszszymanski129:
@fowczarek Will move the invoice mutations to separate file for sure, at first it didn't seem like there will be much but it grown quickly. I'm also planning some more refactor, I will handle it in separate PR, before merging feature branch to master.

@fowczarek, @maarcingebala please re-review, everything is handled except moving invoice related mutations to separate file which I will cover on separate refactor PR that will also include some other stuff.

saleor/graphql/order/mutations/orders.py
```diff
@@ -580,16 +583,35 @@ def clean_input(cls, info, instance, data):
...
         return data["input"]
 
+    @classmethod
+    def clean_instance(cls, info, instance):
+        if instance.status == OrderStatus.DRAFT:
+            raise ValidationError(
+                {
+                    "orderId": ValidationError(
+                        "Provided order cannot be draft.",
+                        code=InvoiceErrorCode.NOT_READY,
```
maarcingebala, line 593: Do we `NOT_READY` in some other cases as well? I'm wondering if it is clear here. Maybe something more explicit like `INVALID_STATUS` would be more appropriate?

tomaszszymanski129: We use it also for invoice when trying to send it and due to some async stuff it is not marked as ready yet. But maybe it's good to use new status here, furthermore now i'm using InvoiceErrorCode on field orderId... So it would be for the best imo to create `INVALID_STATUS` on `OrderErrorCode` and use it here.

maarcingebala: It should be added to InvoiceErrorCode so its easier for fronted what error codes can happen in invoice-related mutations.

tomaszszymanski129: Done

saleor/graphql/order/mutations/orders.py
fowczarek: We should add the same validation to `RequestInvoice`

saleor/graphql/order/mutations/orders.py
```diff
@@ -580,16 +583,35 @@ def clean_input(cls, info, instance, data):
...
         cleaned_input = cls.clean_input(info, instance, data)
         invoice = cls.construct_instance(cls.get_instance(info, **data), cleaned_input)
         invoice.order = instance
         invoice.status = InvoiceStatus.READY
         invoice.save()
+        events.invoice_created_event(
+            user=info.context.user,
+            invoice=invoice,
+            number=data["input"].get("number"),
+            url=data["input"].get("url"),
```
IKarbowiak, line 613: You can use `cleaned_input` here.

PR #5659, opened 2020-05-17
Remove task from sending fulfillment update

Boquete:
I want to merge this change because this more hot fix then proper solution.

Right now, when buying *DIGITAL* product, which is automatically fulfilled, when this task is in queue, in 99% of runs it returns error:
Fulfillment.DoesNotExist
Fulfillment matching query does not exist.

Looks like some strange bug with async actions, but removing it from queue, just running as normal function fixes problem.

As I said, that's not really ready to merge solution, more like reporting issue with fast proposal if someone is using automatic fulfillment.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/order/emails.py
+++ b/saleor/order/emails.py
@@ -101,7 +101,6 @@ def send_staff_order_confirmation(order_pk, redirect_url):
         send_templated_mail(**staff_email_data)
 
 
-@app.task
 def send_fulfillment_confirmation(order_pk, fulfillment_pk):
     email_data = collect_data_for_fullfillment_email(
         order_pk, CONFIRM_FULFILLMENT_TEMPLATE, fulfillment_pk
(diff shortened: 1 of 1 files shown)


saleor/order/emails.py
```diff
@@ -110,7 +109,7 @@ def send_fulfillment_confirmation(order_pk, fulfillment_pk):
 
 
 def send_fulfillment_confirmation_to_customer(order, fulfillment, user):
-    send_fulfillment_confirmation.delay(order.pk, fulfillment.pk)
+    send_fulfillment_confirmation(order.pk, fulfillment.pk)
```
NyanKiyoshi, line 112: Emails should be a task. This is not solving the root of the issue.

Boquete: > "As I said, that's not really ready to merge solution, more like reporting issue with fast proposal if someone is using automatic fulfillment."

Boquete: Created issue for that instead, and added mention for this PR. Closing this PR as it's not needed

PR #5667, opened 2020-05-18
Update version and license

maarcingebala:
Updated package.json and LICENSE.md with up-to-date info.


patrys:
@tomaszszymanski129 We are planning to move email templates outside of the core, then we can think about dropping the remaining Node.js scripts.

PR #5682, opened 2020-05-21
Prevent creating checkout with unpublished product

d-wysocki:
I want to merge this change because...closes https://github.com/mirumee/saleor/issues/5676

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested


saleor/graphql/checkout/mutations.py
```diff
@@ -283,7 +283,11 @@ def save(cls, info, instance: models.Checkout, cleaned_input):
                     raise ValidationError(
                         f"Insufficient product stock: {exc.item}", code=exc.code
                     )
-
+                except ProductNotPublished as exc:
+                    raise ValidationError(
+                        "Can't create checkout with unpublished product.",
```
korycins, line 288: AFAIK Checkout can be already created. This should be something like `Can't add ...`

tests/api/test_checkout.py
```diff
@@ -754,6 +754,28 @@ def test_checkout_lines_add(
...
+    variables = {
+        "checkoutId": checkout_id,
+        "lines": [{"variantId": variant_id, "quantity": 1}],
+    }
+
+    response = user_api_client.post_graphql(MUTATION_CHECKOUT_LINES_ADD, variables)
+
+    content = get_graphql_content(response)
+    error = content["data"]["checkoutLinesAdd"]["errors"][0]
+    assert error["message"] == "Can't create checkout with unpublished product."
```
korycins, line 776: Can you add assert for `error['code']` here?

fowczarek:
How about `checkoutLinesUpdate` mutation with unpublished products?

@maarcingebala  Should we allow to add unpublished products to draft order? 
If yes, should we allow to finalize the draft order with unpublished products?

tests/api/benchmark/test_checkout_mutations.py
```diff
@@ -228,6 +228,50 @@ def test_create_checkout(
     assert checkout_counts + 1 == Checkout.objects.count()
 
 
+@pytest.mark.django_db
+@pytest.mark.count_queries(autouse=False)
+def test_create_checkout_with_unpublished_product(
```
fowczarek, line 233: Why this test is in `benchmark` folder?

d-wysocki: In this folder there are also other tests related to checkout like `test_create_checkout`, `test_update_checkout_lines.

maarcingebala:
@fowczarek I think it shouldn't be possible to add unpublished products to draft orders as well. At least for now it's safer to assume this approach, and if we have a case that we need to add them to draft orders, then we'll make sure that it's properly implemented.

saleor/graphql/checkout/mutations.py
```diff
@@ -354,6 +358,11 @@ def perform_mutation(cls, _root, info, checkout_id, lines, replace=False):
                     raise ValidationError(
                         f"Insufficient product stock: {exc.item}", code=exc.code
                     )
+                except ProductNotPublished as exc:
+                    raise ValidationError(
+                        "Can't create checkout with unpublished product.",
```
fowczarek, line 363: This isn't `checkout_create` mutation. :police_officer:

PR #5684, opened 2020-05-22
Add missing OrderEvents in checkout flow

### LINKED ISSUE #5535: Missing payment_captured and order_fully_paid events
When placing orders using a gateway that captures the amount (without using the dashboard), the events payment_captured and order_fully_paid are not registered. These events are registered only when marking orders as paid in the dashboard.

1. Make an Order.
2. Pay using Dummy (fully-charged) or a credit card Gateway.
3. Look into table order_orderevent.
4. Events payment_captured and order_fully_paid are missing.

### What I expected to happen
I expect these events registered when capturing a payment.

Thanks and awesome project!

koradon:
it adds missing OrderEvents in checkout flow.

fixes #5535 

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/order/actions.py
```diff
@@ -33,14 +33,30 @@ def order_created(order: "Order", user: "User", from_draft: bool = False):
     events.order_created_event(order=order, user=user, from_draft=from_draft)
     manager = get_plugins_manager()
     manager.order_created(order)
+    handle_post_checkout_events(order=order, user=user)
```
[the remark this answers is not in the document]
koradon: Well they shouldn't have any payments and transactions so those events should not be created. I will check it and add test for it.

saleor/order/actions.py
```diff
@@ -33,14 +33,30 @@ def order_created(order: "Order", user: "User", from_draft: bool = False):
...
+        if order.is_captured():
+            order_captured(
+                order=order, user=user, amount=payment.total, payment=payment
+            )
+        elif order.is_pre_authorized():
+            order_authorized(
+                order=order, user=user, amount=payment.total, payment=payment
+            )
+        if order.is_fully_paid():
+            handle_fully_paid_order(order=order, user=user)
```
maarcingebala, line 51: This function is also called in `gateway_postprocess`. Wouldn't calls to this function be duplicated in some flows? Ideally, we should remove it from `gateway_postprocess`.

saleor/order/actions.py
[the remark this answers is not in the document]
maarcingebala: I'm not sure introducing a new function `handle_post_checkout_events` helps - now we have three functions that are called only once and they all do similar things: `order_created`, `handle_post_checkout_events`, `handle_fully_paid_order`. I'd rather move this logic to `order_created`. `handle_fully_paid_order` also mixes different things but maybe refactoring this is out of the scope of this PR.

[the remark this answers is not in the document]
koradon: Done

saleor/graphql/order/mutations/orders.py
```diff
@@ -391,7 +392,8 @@ def perform_mutation(cls, _root, info, amount, **data):
         try_payment_action(
             order, info.context.user, payment, gateway.capture, payment, amount
         )
-
+        if order.is_fully_paid():
+            handle_fully_paid_order(order)
```
maarcingebala, line 396: This logic should go after `order_captured` call I guess? We first capture funds, then we're able to say that the order was fully paid. It's done like that in `order_created`.

koradon: Yes, I wanted to ask about this because right now it works that way (there are even tests that are checking this exact order) and for me, it should be as you described.

koradon: Fixed

saleor/graphql/order/tests/test_order.py
```diff
@@ -1576,16 +1576,12 @@ def test_order_capture(
     assert data["isPaid"]
     assert data["totalCaptured"]["amount"] == float(amount)
 
-    event_order_paid = order.events.first()
-    assert event_order_paid.type == order_events.OrderEvents.ORDER_FULLY_PAID
-    assert event_order_paid.user is None
+    from pprint import pprint
+
+    pprint(order.events.all())
```
maarcingebala, line 1581: Print :)

koradon: Done

PR #5698, opened 2020-05-25
Full invoicing plugin, invoicing module and job refactor

tomaszszymanski129:
Refactor of invoices
* [X] Move Invoices to it's own module
* [X] Refactor Invoices to use Job interface
* [x] Change plugin according to PR changes in https://github.com/mirumee/saleor/pull/5656

Ready for review, there are two major questions that I pointed out and moved from previous discussion.

Additional notes for reviewers:
- You can notice that we are dropping `Invoice` model from `Order` app. The reason it was there in first place was that it (model only) was merged before feature branch. It should not contain any data at all since we did not connect it anywhere on `master` so it should be safe with dropping that and re-creating by `0001_initial.py` in invoices module.

Sorry for long PR, jumping between few branches was a nightmare with refactor in progress, so that PR solves SP-629, SP-1512 and SP-1514.

# Impact

* [X] New migrations
* [X] Deprecated API fields or mutations
* [X] Removed API types, fields, or mutations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/core/types/common.py
```diff
@@ -301,3 +302,21 @@ def resolve_type(cls, instance, _info):
             # <DjangoModel>: <GrapheneType>
         }
         return MODEL_TO_TYPE_MAP.get(type(instance))
+
+
+class InvoiceJobInterface(Job):
+    status = InvoiceStatusEnum(description="Job status.", required=True)
```
tomaszszymanski129, line 308: We have two options here
1) leave it modified as is (new interface)
2) Remove `status` from here and just use common Job interface `status` but that doesn't have entity like `pending_delete` so we would have to add some filed like "pending_target" on the job. Or an option "pending_delete" to job. Moving to Job's `status` benefits in that we have same `status` for new interface. What you guys think is better?

saleor/urls.py
```diff
@@ -17,6 +19,11 @@
         digital_product,
         name="digital-product",
     ),
+    path(
+        "download-invoice/<uuid:invoice_hash>/",
```
tomaszszymanski129, line 23: Backstory behind this is that we have a plugin that generates an invoice file on our side (saves to local storage to be compliant with people using S3 for ex.) Plugin saves the invoice then it needs to generate an URL and save it on Invoice model (this is a flow when our built-in plugin handles the creation). It can generate such URL due to having this view so we can serve it later. Then an email mutation can be trigerred and the mail is being sent with URL already saved on model.
Another options we have instead of having that view:
- combine it with existing view `digital_content` so that is our only file-serving endpoint
- re-phrase it to more generic view that can serve content for all plugins for example if there will be need.
- drop the endpoint at all and serve somehow else?

Please advise, especially @patrys since I didn't explain it properly to you, hope this clarifies what needs to be accomplished.

saleor/graphql/core/types/common.py
patrys: I would prefer to have a single common interface. We can extend the base `Job` if needed (as long as the statuses are generic).

saleor/urls.py
patrys: Django media files have a `url` property that returns a public URL of the file. We should not act as a proxy here.

saleor/invoice/error_codes.py
```diff
@@ -0,0 +1,9 @@
+from enum import Enum
+
+
+class InvoiceErrorCode(Enum):
```
fowczarek, line 4: We should add this enum to `SALEOR_ERROR_CODE_ENUMS` in `saleor.graphql.core.utils`.

saleor/core/models.py
```diff
@@ -119,6 +120,7 @@ class Job(models.Model):
     status = models.CharField(
         max_length=50, choices=JobStatus.CHOICES, default=JobStatus.PENDING
     )
+    pending_target = models.CharField(max_length=50, choices=PendingTarget.CHOICES)
```
fowczarek, line 123: If we change something in any of the core functionality, we should do it in separate PR merged to master instead of the feature branch. 

If anybody wants to use this field, they'll have to wait for your feature branch to be merged.

tomaszszymanski129: My god, that meant to be added to InvoiceJob model as it is specific. Will change that!

saleor/invoice/models.py
```diff
@@ -0,0 +1,71 @@
...
+from ..order.models import Order
+from . import InvoiceEvents
+
+
+class InvoiceQueryset(models.QuerySet):
+    def ready(self):
+        return self.filter(job__status=JobStatus.SUCCESS)
+
+
+class Invoice(ModelWithMetadata):
```
fowczarek, line 18: We should add tests for metadata in the invoice.

saleor/graphql/invoice/types.py
```diff
@@ -0,0 +1,23 @@
...
+
+from ...invoice import models
+from ..core.connection import CountableDjangoObjectType
+from ..core.types.common import InvoiceJobInterface
+
+
+class Invoice(CountableDjangoObjectType):
+    class Meta:
+        description = "Represents an Invoice."
+        interfaces = [graphene.relay.Node]
```
fowczarek, line 11: Model inherit form `ModelWithMetadata` and support metadata. 
We should add the proper metadata interface to API or remove the metadata form model.

saleor/invoice/error_codes.py
tomaszszymanski129: It was added already.

tomaszszymanski129:
> Do we plan to have more types that implement the `InvoideJobInterface`? Maybe we can move the implementation straight into `InvoiceJob`?

Dropped implementation of `InvoiceJobInterface` entirely, simplified to use just `Job` as that suits well after some refactor.

saleor/plugins/invoicing/utils.py
```diff
@@ -74,7 +71,6 @@ def generate_invoice_pdf(invoice):
         all_products[product_limit_first_page:], MAX_PRODUCTS_PER_PAGE
     )
     cretion_date = datetime.now(tz=pytz.utc)
```
abodacs, line 73: I think there misspell  here `cretion_date`   should be `creation_date`

saleor/invoice/models.py
```diff
@@ -0,0 +1,82 @@
...
+        if url is not None:
+            self.url = url
+        self.save()
+
+
+class InvoiceJob(Job):
+    invoice = models.OneToOneField(
+        Invoice, on_delete=models.CASCADE, related_name="job"
+    )
+    pending_target = models.CharField(max_length=50, choices=PendingTarget.CHOICES)
```
IKarbowiak, line 52: Could you explain what is the reason for using the additional status field? Why Job's `status` is not enough?

saleor/plugins/invoicing/plugin.py
```diff
@@ -0,0 +1,35 @@
...
+        self,
+        order: "Order",
+        invoice_job: "InvoiceJob",
+        number: Optional[str],
+        previous_value: Any,
+    ) -> Any:
+        file_content, creation_date = generate_invoice_pdf(invoice_job.invoice)
+        invoice_job.invoice.created = creation_date
+        invoice_job.invoice.update_invoice(number=generate_invoice_number())
+        invoice_job.invoice.save()
```
fowczarek, line 29: Save should be used with `update_fields`. 

(`update_invoice` this function use `save`. That's the reason why we should avoid it)

saleor/plugins/invoicing/plugin.py
```diff
@@ -0,0 +1,35 @@
...
+        invoice_job.invoice.invoice_file.save(
+            f"{uuid4()}.pdf", ContentFile(file_content)
+        )
+        invoice_job.status = JobStatus.SUCCESS
+        invoice_job.save()
```
fowczarek, line 34: Save should be used with `update_fields`.

saleor/graphql/invoice/mutations.py
```diff
@@ -0,0 +1,278 @@
...
+
+    @classmethod
+    def perform_mutation(cls, _root, info, **data):
+        instance = cls.get_instance(info, **data)
+        cleaned_input = cls.clean_input(info, instance, data)
+        instance.invoice.update_invoice(
+            number=cleaned_input.get("number"), url=cleaned_input.get("url")
+        )
+        instance.status = JobStatus.SUCCESS
+        instance.save()
```
fowczarek, line 243: Save should be used with `update_fields`.

saleor/graphql/invoice/mutations.py
```diff
@@ -0,0 +1,278 @@
...
+            required=False, description="Invoice number (optional)."
+        )
+
+    @classmethod
+    def clean_instance(cls, info, instance):
+        if instance.status == OrderStatus.DRAFT:
+            raise ValidationError(
+                {
+                    "orderId": ValidationError(
+                        "Provided order status cannot be draft.",
```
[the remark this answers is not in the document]
maarcingebala: I guess correct is "Cannot request an invoice for draft order."

saleor/graphql/invoice/mutations.py
```diff
@@ -0,0 +1,278 @@
...
+        description = "Deletes an invoice."
+        model = models.InvoiceJob
+        permissions = (OrderPermissions.MANAGE_ORDERS,)
+        error_type_class = InvoiceError
+        error_type_field = "invoice_errors"
+
+    @classmethod
+    def perform_mutation(cls, _root, info, **data):
+        invoice_job = cls.get_instance(info, **data)
+        response = super().perform_mutation(_root, info, **data)
```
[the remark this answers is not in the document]
tomaszszymanski129: Refactored so now it uses just invoice model.

saleor/invoice/models.py
tomaszszymanski129: Added a docstring. When an Invoice is pending it can transform to DELTED or SUCCESS (we don't have "PENDING_DELETE" status on Job), this field holds that information for extra logic in plugin.

patrys: Do we need a pending delete status (is it going to be used for anything)? If so, we can add it to the Job interface.

saleor/plugins/invoicing/plugin.py
```diff
@@ -0,0 +1,41 @@
...
+        self,
+        order: "Order",
+        invoice: "Invoice",
+        number: Optional[str],
+        previous_value: Any,
+    ) -> Any:
+        file_content, creation_date = generate_invoice_pdf(invoice)
+        invoice.created = creation_date
+        invoice.update_invoice(number=generate_invoice_number())
+        invoice.invoice_file.save(f"{uuid4()}.pdf", ContentFile(file_content))
```
maarcingebala, line 29: I think it could be useful to have the order number in the file name. E.g. format `order-{order_number}-invoice-{number}` - that would be more human-readable. If we cannot guarantee the number is unique, we could then append some short UUID, like 6 characters or so.

saleor/graphql/invoice/mutations.py
```diff
@@ -0,0 +1,273 @@
...
+        permissions = (OrderPermissions.MANAGE_ORDERS,)
+        error_type_class = InvoiceError
+        error_type_field = "invoice_errors"
+
+    class Arguments:
+        order_id = graphene.ID(
+            required=True, description="ID of the order related to invoice."
+        )
+        number = graphene.String(
+            required=False, description="Invoice number (optional)."
```
maarcingebala, line 30: It would be useful to explain it a bit more e.g. "Invoice number (optional). Provide it if you want to set a specific number on the invoice. Otherwise, a unique number will be generated." - something like that?

saleor/graphql/invoice/mutations.py
```diff
@@ -0,0 +1,273 @@
...
+    class Arguments:
+        order_id = graphene.ID(
+            required=True, description="ID of the order related to invoice."
+        )
+        input = CreateInvoiceInput(
+            required=True, description="Fields required when creating an invoice."
+        )
+
+    class Meta:
+        description = "Creates an invoice."
```
maarcingebala, line 93: As an API user, I see here two similar mutations - `requestInvoice` and `createInvoice`? Which one and when should I use? These descriptions don't tell that. Can we explain it shortly in both mutations?

maarcingebala: The same applies to `deleteInvoice` and `requestDeleteInvoice` mutations.

saleor/invoice/models.py
IKarbowiak: DELETING invoice is such a long action that we need special status for that?

saleor/graphql/invoice/types.py
```diff
@@ -0,0 +1,22 @@
+import graphene
+
+from ...invoice import models
+from ..core.connection import CountableDjangoObjectType
+from ..core.types.common import Job
+
+
+class Invoice(CountableDjangoObjectType):
+    url = graphene.String()
```
maarcingebala, line 9: Missing description 👮

saleor/invoice/models.py
tomaszszymanski129: @IKarbowiak It can be done async through external service.

saleor/graphql/order/types.py
```diff
@@ -506,7 +499,7 @@ def resolve_available_shipping_methods(root: models.Order, _info):
 
     @staticmethod
     def resolve_invoices(root: models.Order, info):
-        return root.invoices.ready()
+        return root.invoices.all()
```
maarcingebala, line 502: Do we want to return all invoices for any order? I think we should use similar logic as in `resolve_user` function in this type, so you can only fetch invoices if you're an admin user with proper permission or you're a customer and this is your order. Should we return the invoices for anonymous users as well? They'll get them by email and I don't think we'll render them in the storefront. What do you think @patrys?

saleor/plugins/invoicing/utils.py
```diff
@@ -0,0 +1,85 @@
...
+    rest_of_products = chunk_products(
+        all_products[product_limit_first_page:], MAX_PRODUCTS_PER_PAGE
+    )
+    creation_date = datetime.now(tz=pytz.utc)
+    rendered_template = get_template("invoices/invoice.html").render(
+        {
+            "invoice": invoice,
+            "creation_date": creation_date.strftime("%d %b %Y"),
+            "order": invoice.order,
+            "logo_path": f"file://{logo_path}",
```
maarcingebala, line 79: We dropped the Saleor logo from emails, so I think we should also drop it from invoices. On the other hand, it could be useful if you're running Saleor yourself as you have to only replace the logo file to have yours.

saleor/graphql/invoice/mutations.py
tomaszszymanski129: Will do so, furthermore - after everything is merged to master I will spend a while and update the docs. We have a synchronous flow, async flow, plugin to handle invoices. A guide how to make use of all of this or customize it to one needs would be super handy.

saleor/plugins/invoicing/utils.py
tomaszszymanski129: Let's make it configurable, it will default to Saleor's logo and can be overriden by settings, what do you think @maarcingebala ?

saleor/plugins/invoicing/plugin.py
tomaszszymanski129: It might be better to stick with UUID:
- no os.path calls to check whether file exists
- some setups might use S3, would have to check if it exists on S3 in case we use S3 :fire: 
So i think it's best to do:
`order-{order_number}-invoice-{number}-{uuid4()}`
so we have both uniqueness and information in the file, then it's easier to figure what that PDF file in your downloads folder is.

maarcingebala: Sounds good!

patrys: I would also stick to UUID because we don't want to risk users enumerating existing invoices by guessing their filenames. As for checking whether a file exists, we may still need to do that, UUID4 is not safe from collisions (even if they are unlikely, see the birthday paradox).

saleor/plugins/invoicing/plugin.py
```diff
@@ -0,0 +1,44 @@
...
+from ...invoice.models import Invoice
+from ...order.models import Order
+from ..base_plugin import BasePlugin
+from .utils import generate_invoice_number, generate_invoice_pdf
+
+
+class InvoicingPlugin(BasePlugin):
+    PLUGIN_ID = "mirumee.invoicing"
+    PLUGIN_NAME = "Invoicing"
+    DEFAULT_ACTIVE = True
```
korycins, line 16: Just to be sure. Did you test the functionality with the scenario when the user has turned off `InvoicingPlugin` and he doesn't have any other Apps/Plugins for invoicing?

saleor/plugins/invoicing/utils.py
korycins: Fyi We dropped it from all emails, so if you want to add it here you need to add them in all emails.

saleor/settings.py
```diff
@@ -556,3 +557,5 @@ def get_host():
 if REDIS_URL:
     CACHE_URL = os.environ.setdefault("CACHE_URL", REDIS_URL)
 CACHES = {"default": django_cache_url.config()}
+
+INVOICE_LOGO_PATH = os.environ.get("INVOICE_LOGO_PATH", "images/logo.svg")
```
korycins, line 561: As @maarcingebala  said, we don't want to use saleor's logo by default.
And this will fail. How do you handle the scenario when a user doesn't want to add own logo? I see that `INVOICE_LOGO_PATH` requires path.

saleor/plugins/invoicing/plugin.py
tomaszszymanski129: If this is what the user wants to do, it is an intentional configuration. No invoice will be generated in that case but if this is how system is configured it should work this way I assume :man_shrugging:

saleor/graphql/invoice/mutations.py
```diff
@@ -0,0 +1,274 @@
...
+        error_type_field = "invoice_errors"
+
+    @classmethod
+    def clean_instance(cls, info, instance):
+        error_code = None
+
+        if instance.status != JobStatus.SUCCESS:
+            error_code = InvoiceErrorCode.NOT_READY
+        elif not instance.external_url or not instance.number:
+            error_code = InvoiceErrorCode.URL_OR_NUMBER_NOT_SET
```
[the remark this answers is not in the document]
tomaszszymanski129: User shall not get this error because this is triggered by dashboard which implementation will show only ready invoices with send possibility. Also I doubt if anyone while building their own frontend would like to send not-ready invoices, therefore if someone sees this error there is something wrong with his system and that requires debugging the root cause anyway.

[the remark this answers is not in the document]
korycins: Don't you think that for debugging, two different errors would be better than one for two causes?

saleor/plugins/invoicing/plugin.py
korycins: So my question was. Did you test this scenario? It is a new functionality we need to assume that someone will turn it off.

saleor/graphql/invoice/mutations.py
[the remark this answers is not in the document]
tomaszszymanski129: It's exactly same because both fields are in the same place so you see at first glance what is not set. Anyway, all right, explicit is better than implicit.

saleor/plugins/invoicing/plugin.py
tomaszszymanski129: Tested, doesn't break anything other that invoice is not being generated.

saleor/graphql/invoice/types.py
```diff
@@ -0,0 +1,22 @@
...
+from ..core.types.common import Job
+from ..meta.types import ObjectWithMetadata
+
+
+class Invoice(CountableDjangoObjectType):
+    url = graphene.String(description="URL to download an invoice.")
+
+    class Meta:
+        description = "Represents an Invoice."
+        interfaces = [ObjectWithMetadata, Job]
```
fowczarek, line 14: Can you check if Invoice resolver returns ID as graphene ID instead of PK?
If yes we should add: 
```suggestion
        interfaces = [ObjectWithMetadata, Job, relay.Node]
```

saleor/graphql/meta/resolvers.py
```diff
@@ -19,6 +19,8 @@ def resolve_object_with_metadata_type(instance: ModelWithMetadata):
     from ..checkout import types as checkout_types
     from ..order import types as order_types
     from ..product import types as product_types
+    from ..invoice import types as invoice_types
+    from ...invoice import models as invoice_models
```
fowczarek, line 23: We can import model outside function.

saleor/graphql/order/types.py
```diff
@@ -506,7 +499,10 @@ def resolve_available_shipping_methods(root: models.Order, _info):
 
     @staticmethod
     def resolve_invoices(root: models.Order, info):
-        return root.invoices.ready()
+        user = info.context.user
```
fowczarek, line 502: How about the service account?

saleor/plugins/invoicing/utils.py
```diff
@@ -0,0 +1,82 @@
...
+    month_and_year = datetime.now().strftime("%m/%Y")
+    return f"{number}/{month_and_year}"
+
+
+def parse_invoice_number(invoice):
+    number = re.match(r"^(\d+)\/", invoice.number).group(1)
+    return int(number)
+
+
+def generate_invoice_number():
```
fowczarek, line 27: Should we add tests for this function?

saleor/graphql/meta/resolvers.py
tomaszszymanski129: Cannot do, it is done that way (like other impoets here) so we don't fall into circular import.

saleor/graphql/order/types.py
tomaszszymanski129: we already test that

saleor/plugins/invoicing/utils.py
tomaszszymanski129: Added.

tomaszszymanski129:
@maarcingebala @fowczarek please check last commit where I add missing tests, otherwise it's all done.

saleor/graphql/order/types.py
```diff
@@ -499,8 +499,9 @@ def resolve_available_shipping_methods(root: models.Order, _info):
 
     @staticmethod
     def resolve_invoices(root: models.Order, info):
-        user = info.context.user
-        if user == root.user or user.has_perm(OrderPermissions.MANAGE_ORDERS):
+        if info.operation.name.value == "OrderByToken" or info.context.user.has_perm(
+            OrderPermissions.MANAGE_ORDERS
+        ):
```
fowczarek, line 504: Maby we can use a similar approach like in `resolve_checkout`?

Have we tests for this resolver as an anonymous customer, logged-in customer, staff user, and app? 

```suggestion
        # resolve invoice for anonymous customer
        if root is None or order.user is None:
            return PermissionDenied()

        # resolve invoice for logged-in customer
        if order.user == info.context.user:
            return root.invoices.all()
            
        # resolve checkout for staff user or app
        requester = get_user_or_app_from_context(info.context)
        if requester.has_perm(OrderPermissions.MANAGE_ORDERS):
```

fowczarek: Have we tests for this resolver as an anonymous customer, logged-in customer, staff user, and app?

maarcingebala: We shouldn't hardcode names of queries here. I think it was correct last time I check this. Something like that should work (we have the same pattern is for `resolve_user` resolver in this and few other types):
```
user = info.context.user
if user == root.user or user.has_perm(AccountPermissions.MANAGE_USERS):
    return root.invoices.all()
raise PermissionDenied()
```
This way we only return invoices if the requesting user is also the owner of the order or a staff member with MANAGE_USERS.

maarcingebala: We have some tests for that.

fowczarek: @maarcingebala If an app tries to get an invoice resolver will raise PermissionDenied.

maarcingebala: I'm fine with adding support for apps here, but there are anyway more places in the entire API that don't support that; for example `resolve_user` function in this type.

PR #5700, opened 2020-05-26
Export products in batches

IKarbowiak:
Export products in batches to avoid OOM errors.
Refactor for reducing exporting time.

Right now, memory usage for 50 000, 100 000 and 200 000 is the same. About 200 Mb.

#### Time-consuming and number of queries:
For CSV file:
- 200 000 products, **30 seconds**, 100 db queries
- 100 000 products, **18 seconds**, 60 db queries
- 50 000 products,  **10 seconds**, 40 db queries
- 20 000 products, **8 seconds**, 28 db queries

XLSX is slow. Append to file is really time-consuming, so I think, there is nothing we can do with that.
- 50 000 products, **10 minutes**, 47 db queries

Flow diagram:
image

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/csv/utils/export.py
+++ b/saleor/csv/utils/export.py
@@ -105,10 +105,7 @@ def queryset_in_batches(queryset):
     """
     start_pk = 0
 
-    while True:
-        if not queryset.filter(pk__gt=start_pk).exists():
-            break
-
+    while queryset.filter(pk__gt=start_pk).exists():
         qs = queryset.filter(pk__gt=start_pk)
 
         pks = qs.values_list("pk", flat=True)
--- a/saleor/csv/utils/export.py
+++ b/saleor/csv/utils/export.py
@@ -74,7 +74,6 @@ def export_products(
         file_name,
         file_type,
     )
-    send_email_with_link_to_download_csv(export_file, "export_products_success")
 
 
 def get_filename(model_name: str, file_type: str) -> str:
@@ -167,7 +166,7 @@ def export_products_in_batches(
         else:
             append_to_file(export_data, headers, export_file, file_type, delimiter)
 
-    send_email_with_link_to_download_csv(export_file, "export_products")
+    send_email_with_link_to_download_csv(export_file, "export_products_success")
 
 
 def create_csv_file_and_save_in_export_file(
--- a/saleor/csv/utils/export.py
+++ b/saleor/csv/utils/export.py
@@ -99,6 +99,30 @@ def get_product_queryset(scope: Dict[str, Union[str, dict]]) -> "QuerySet":
     return queryset
 
 
+def queryset_in_batches(queryset):
+    """Slice a queryset into batches.
+
+    Input queryset should be sorted be pk.
+    """
+    start_pk = 0
+
+    while True:
+        if not queryset.filter(pk__gt=start_pk).exists():
+            break
+
+        qs = queryset.filter(pk__gt=start_pk)
+
+        pks = qs.values_list("pk", flat=True)
+        try:
+            end_pk = pks[BATCH_SIZE - 1]
+        except IndexError:
+            end_pk = pks.last()
+
+        yield qs.filter(pk__lte=end_pk)
+
+        start_pk = end_pk
+
+
 def export_products_in_batches(
     queryset: "QuerySet",
     export_info: Dict[str, list],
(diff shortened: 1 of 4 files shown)


saleor/csv/utils/export.py
```diff
@@ -84,13 +92,68 @@ def get_product_queryset(scope: Dict[str, Union[str, dict]]) -> "QuerySet":
...
+                "category",
+            )
+        )
+
+        export_data = get_products_data(
+            product_batch, export_fields, warehouses, attributes,
+        )
+
+        if create_file:
+            create_csv_file_and_save_in_export_file(
```
korycins, line 139: Question: What should we do in a case when `products_count` == 0? What will happen with `send_email_with_link_to_download_csv` when file will be empty?
- Do we want to create an empty file only with headers? 
- we don't want to create a file in that case?

Suggestion: If you would have a function for `create_csv_file_and_headers`, then you could call it before loop and this `if...else` condition would not be needed.

IKarbowiak: IMO we should create the empty file only with headers. I'll test it for such a case.

maarcingebala: We can also check what other systems do in such a case. I expect that when having 0 products the export operation will be blocked at least in UI.

IKarbowiak: But I think we still should handle this situation because right now it fails if there are no products.

maarcingebala: Yes, API should handle that as well. I guess creating an empty file is acceptable but I would check what other systems do. As I said - it should be blocked in UI so it's not common case that someone runs the export though API only.

saleor/csv/utils/export.py
```diff
@@ -84,13 +92,68 @@ def get_product_queryset(scope: Dict[str, Union[str, dict]]) -> "QuerySet":
...
+    file_type: str,
+):
+    products_count = queryset.count()
+    warehouses = export_info.get("warehouses")
+    attributes = export_info.get("attributes")
+
+    create_file = True
+    counter = 0
+    while counter < products_count:
+        qs = queryset[counter : counter + BATCH_SIZE]
```
maarcingebala, line 118: Just for the record - `count` and slicing the queryset (database `OFFSET`) will be expensive and slow for large querysets. We agreed with @IKarbowiak to try to refactor this to use cursor-based approach as we do for pagination.

saleor/csv/utils/export.py
```diff
@@ -10,17 +10,24 @@
...
 from ..models import ExportFile
-from .products_data import get_products_data
+from .products_data import get_export_fields_and_headers_info, get_products_data
 
 if TYPE_CHECKING:
     # flake8: noqa
     from django.db.models import QuerySet
 
 
+BATCH_SIZE = 10000
```
NyanKiyoshi, line 20: I wonder if that's the right place? I'm not really a fan of tunable values being inside python (code) files. Seems like with meaningful name it would be better of inside `settings.py`? Unless there is good a reason.

IKarbowiak: But it concerns only export

maarcingebala: But should this value be configurable? It's rather a constant where we should find the optimal value in tests on different datasets.

saleor/csv/utils/export.py
```diff
@@ -85,24 +93,74 @@ def get_product_queryset(scope: Dict[str, Union[str, dict]]) -> "QuerySet":
...
+def queryset_in_batches(queryset):
+    """Slice a queryset into batches.
+
+    Input queryset should be sorted be pk.
+    """
+    start_pk = 0
+
+    while True:
+        if not queryset.filter(pk__gt=start_pk).exists():
+            break
```
fowczarek, line 110: Maby?
```suggestion
    while queryset.filter(pk__gt=start_pk).exists():
```

saleor/csv/utils/export.py
```diff
@@ -85,24 +93,71 @@ def get_product_queryset(scope: Dict[str, Union[str, dict]]) -> "QuerySet":
...
-def create_csv_file_and_save_in_export_file(
-    export_data: List[Dict[str, Union[str, bool]]],
+def queryset_in_batches(queryset):
+    """Slice a queryset into batches.
+
+    Input queryset should be sorted be pk.
+    """
+    start_pk = 0
+
+    while queryset.filter(pk__gt=start_pk).exists():
```
patrys, line 108: This executes two additional queries per batch (one for `.exists()`, one for `pks[...]`). An optimized version could be something like this:

```python
while True:
    qs = queryset.filter(pk__gt=start_pk)[:BATCH_SIZE]
    pks = list(qs.values_list("pk", flat=True))
    if not pks:
        break
    yield pks
```

IKarbowiak: I'm testing it right now, and yes, it produces fewer queries. For 200k products, it's 100 db queries instead of 122, but time is about 35 seconds instead of 30 seconds. It's a small difference. But I'll be checking if the offset takes longer than filter or just my system is overwhelmed right now.

IKarbowiak: I changed the solution to your suggestion.

patrys: I hope my version doesn't use OFFSET. As far as I know both use LIMIT: the original to calculate `pks[BATCH_SIZE-1]`. my version to slice `...[:BATCH_SIZE]`.

IKarbowiak: Yes, according to Django docs is LIMIT. My mistake 😉

PR #5706, opened 2020-05-27
Remove base price field

d-wysocki:

# Impact

* [X] New migrations
* [X] Removed API types, fields, or mutations

# Pull Request Checklist

* [X] The changes are tested


saleor/graphql/product/filters.py
```diff
@@ -298,9 +298,6 @@ class ProductFilter(django_filters.FilterSet):
     collections = GlobalIDMultipleChoiceFilter(method=filter_collections)
     categories = GlobalIDMultipleChoiceFilter(method=filter_categories)
     has_category = django_filters.BooleanFilter(method=filter_has_category)
-    price = ObjectTypeFilter(
```
maarcingebala, line 301: Although we remove this field from the product model, I think it would be still useful to be able to filter products by price. We could rewrite the filter logic to use variant prices instead.

saleor/graphql/product/mutations/products.py
```diff
@@ -556,6 +555,7 @@ class ProductCreateInput(ProductInput):
         ),
         required=False,
     )
+    base_price = Decimal(description="Default price for product variant.",)
```
maarcingebala, line 558: What about the `productUpdate` mutation? It uses `ProductInput` and for now we want to use it to update prices of simple products (with the one default variant).

saleor/graphql/product/types/products.py
```diff
@@ -192,13 +192,8 @@ class ProductVariant(CountableDjangoObjectType):
...
         ),
     )
-    price_override = graphene.Field(
-        Money,
-        description=(
-            "Override the base price of a product if necessary. A value of `null` "
-            "indicates that the default product price is used."
-        ),
-    )
+    price = graphene.Field(Money, description=("Base price of a product variant."),)
```
maarcingebala, line 195: I would extend this description with sth like: "This field is restricted for admins. Use the `pricing` field to get the public price for customers." It will be less confusing for API users when they see that the regular price field is protected with permissions.

saleor/product/models.py
```diff
@@ -273,15 +244,11 @@ class Product(SeoModel, ModelWithMetadata, PublishableModel):
         default=settings.DEFAULT_CURRENCY,
     )
 
-    price_amount = models.DecimalField(
-        max_digits=settings.DEFAULT_MAX_DIGITS,
-        decimal_places=settings.DEFAULT_DECIMAL_PLACES,
-    )
-    price = MoneyField(amount_field="price_amount", currency_field="currency")
```
maarcingebala, line 280: I guess `currency` field could be also dropped. It was used for the compound `price` field, but not sure if it is used anywhere else - we would have to check this.

saleor/graphql/product/sorters.py
```diff
@@ -142,7 +141,6 @@ def description(self):
         # pylint: disable=no-member
         descriptions = {
             ProductOrderField.NAME.name: "name",
-            ProductOrderField.PRICE.name: "price",
```
maarcingebala, line 145: I think some client apps can already use this field. Maybe it would be safer to keep it but underneath use the minimal variant price field for sorting. At the same time, we could deprecate the "minimal price" filter and say in the description that from now the "price" filter should be used. Later we'll remove it and we'll have only one way of sorting by price.

d-wysocki:
@maarcingebala good points, working on changes

tests/api/test_product.py
```diff
@@ -1763,44 +1718,44 @@ def test_update_product_slug_and_name(
...
-        UPDATE_PRODUCT_PRICE_MUTATION,
-        variables,
-        permissions=[permission_manage_products],
-    )
-    content = get_graphql_content(response)
-    data = content["data"]["productUpdate"]
-    errors = data["productErrors"]
-    assert errors[0]["field"] == "basePrice"
-    assert errors[0]["code"] == ProductErrorCode.INVALID.name
+# UPDATE_PRODUCT_PRICE_MUTATION = """
```
fowczarek, line 1721: If you want temporary turn off tests you should use skip instead of comment out. 
`@pytest.mark.skip(reason="<Reason>")`

tests/api/test_product.py
```diff
@@ -1763,44 +1718,44 @@ def test_update_product_slug_and_name(
...
+#                 field
+#                 message
+#                 code
+#             }
+#         }
+#     }
+# """
+#
+#
+# def test_update_product_invalid_price(
```
fowczarek, line 1743: If you want temporary turn off tests you should use skip instead of comment out. 
`@pytest.mark.skip(reason="<Reason>")`

tests/api/test_product.py
```diff
@@ -3930,28 +3872,28 @@ def test_bulk_unpublish_products(
...
-
-    variables = {"productID": product_id}
-    permissions = [permission_manage_products]
-
-    response = staff_api_client.post_graphql(query, variables, permissions)
-    content = get_graphql_content(response)
-
-    assert "basePrice" in content["data"]["product"]
-    assert content["data"]["product"]["basePrice"]["amount"] == product.price.amount
+# def test_product_base_price_permission(
```
fowczarek, line 3875: If you want temporary turn off tests you should use skip instead of comment out. 
`@pytest.mark.skip(reason="<Reason>")`

tests/test_product_minimal_variant_price.py
```diff
@@ -13,161 +13,91 @@
...
-    assert product1.price == product1.minimal_variant_price == Money("10", "USD")
-
-    assert product2.minimal_variant_price
-    assert product2.price != product2.minimal_variant_price
-    assert product2.minimal_variant_price == Money("20", "USD")
+        assert product.minimal_variant_price == price
 
 
 def test_product_variant_objects_create_updates_minimal_variant_price(product):
+    product.refresh_from_db()
```
fowczarek, line 87: Why we do this?

tests/test_webhook_payloads.py
```diff
@@ -127,6 +127,7 @@ def test_generate_sample_customer_payload(customer_user):
 
 
 def test_generate_sample_product_payload(variant):
+    variant.product.refresh_from_db()
```
fowczarek, line 130: Why we do this?

saleor/product/models.py
d-wysocki: `currency` field is also used in `minimal_variant_price` field.

saleor/graphql/product/mutations/products.py
d-wysocki: I just restored old logic.

maarcingebala: So to not break creating products in the dashboard, we should have the `base_price` field in both mutations - for creating and updating products. Now we have it only in mutation to create. If I'm correct OFC, it's Friday afternoon ;)

saleor/product/models.py
```diff
@@ -337,20 +295,14 @@ def get_first_image(self):
 
     def get_price_range(
```
korycins, line 296: Is this method still used?

tests/api/test_product.py
```diff
@@ -469,16 +466,15 @@ def test_products_query_with_filter(
     product,
     permission_manage_products,
 ):
-    assert product.price == Money("10.00", "USD")
-    assert product.minimal_variant_price == Money("10.00", "USD")
     assert product.is_published is True
     assert "Juice1" not in product.name
+    product.minimal_variant_price = Money("10.00", "USD")
```
korycins, line 471: Is it needed?

tests/api/test_product_minimal_variant_price.py
```diff
@@ -3,117 +3,12 @@
...
 from freezegun import freeze_time
 from graphql_relay import from_global_id, to_global_id
-from prices import Money
 
 from saleor.graphql.discount.enums import DiscountValueTypeEnum
 from saleor.product.error_codes import ProductErrorCode
 from tests.api.utils import get_graphql_content
 
 
-def test_product_create_sets_minimal_variant_price(
```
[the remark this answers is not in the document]
d-wysocki: We don't need this test because it is based on a removed field (`price`) to update `minimal_variant_price`, since this field is removed, this test case is invalid.

tests/api/test_product_minimal_variant_price.py
```diff
@@ -3,117 +3,12 @@
...
-
-    content = get_graphql_content(response)
-    data = content["data"]["productCreate"]
-    assert data["errors"] == []
-    assert str(data["product"]["minimalVariantPrice"]["amount"]) == product_price
-
-
-@patch(
-    "saleor.graphql.product.mutations.products"
-    ".update_product_minimal_variant_price_task"
```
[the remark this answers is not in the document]
d-wysocki: This test also based on removed `price` field.

saleor/product/migrations/0119_auto_20200527_0402.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 3.0.6 on 2020-05-27 09:02
```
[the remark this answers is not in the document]
korycins: IMHO You can do this in one migration file. No need to separate it as it is related to the same changes.
CC @maarcingebala what do you think?

saleor/product/utils/availability.py
```diff
@@ -154,13 +149,8 @@ def get_product_price_range(
...
             return MoneyRange(min(prices), max(prices))
-        price = calculate_discounted_price(
-            product=product,
-            price=product.price,
-            collections=collections,
-            discounts=discounts,
-        )
-        return MoneyRange(start=price, stop=price)
+
+        return MoneyRange(zero_money(), zero_money())
```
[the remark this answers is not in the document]
maarcingebala: Maybe it would be better to return `None` than zero price, which is a valid value. But then we'd have to change `ProductPricingInfo` GraphQL type as now the price range there is not nullable. Also, we would have to make sure that products without variants are not returned for customers in the storefront.

tests/api/test_product_minimal_variant_price.py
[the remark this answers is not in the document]
korycins: But `ProductUpdate` still uses `update_product_minimal_variant_price_task`. 
Correct me if I am wrong. I thought that this test checks if we call `update_product_minimal_variant_price_task` when we call `ProductUpdate` mutation. So I am not sure why this scenario is not valid when we still trigger this task in `ProductUpdate` mutation

saleor/product/utils/availability.py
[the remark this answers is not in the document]
patrys: We can also make the entire pricing info nullable if that makes more sense. If we return a zero price, someone will render it and people will expect a free product.

saleor/product/migrations/0119_auto_20200527_0402.py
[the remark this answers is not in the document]
maarcingebala: Looks like it should be doable in one migration. The fewer migration files, the better, so we can try.

tests/api/test_product_minimal_variant_price.py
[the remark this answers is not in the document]
korycins: So what values I will get for field `minimalVariantProduct` when I call query:
```
{
  products(first:35, filter:{price: { lte: 20}}){
    totalCount
    edges{
      node{
        id
        minimalVariantPrice{
          amount
        }
      }
    }
  }
}
```
Assuming that I have mixed Product With and Without Variants.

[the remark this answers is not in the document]
patrys: All products have variants (or at least all products that you can purchase have them).

saleor/product/utils/availability.py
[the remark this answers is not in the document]
maarcingebala: Yes, I think we should make the `Product.pricing` field nullable and return `None` when there are no variants. Also, we should change the `Product.objects.visible_to_user()` QS to not return these products for users without `MANAGE_PRODUCTS` permission. Staff users having it should see these products (we want to keep rendering them in the dashboard).

saleor/product/migrations/0119_auto_20200527_0402.py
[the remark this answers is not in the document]
d-wysocki: Okey

tests/api/test_product.py
```diff
@@ -791,22 +787,22 @@ def test_filter_products_by_wrong_attributes(user_api_client, product):
...
     # Create the second product with higher price and date
     product.pk = None
     product.slug = "second-product"
-    product.price = Money("20.00", "USD")
-    product.minimal_variant_price = Money("20.00", "USD")
+    product.minimal_variant_amount = 20
     product.updated_at = datetime.utcnow()
     product.save()
+    ProductVariant.objects.create(
+        product=product, sku="1234", cost_price=Money("1.00", "USD"), price_amount=20
```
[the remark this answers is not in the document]
d-wysocki: 👍

saleor/product/utils/availability.py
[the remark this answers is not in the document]
d-wysocki: 👍

tests/api/test_product_minimal_variant_price.py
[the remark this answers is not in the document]
d-wysocki: 👍

saleor/product/utils/availability.py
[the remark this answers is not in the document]
patrys: Since we render other unavailable-for-purchase products like out-of-stock ones, wouldn't it be expected that we do return them to the storefront if they are published? You wouldn't be able to add them to the cart and we would not show a price but it should be possible to at least see their photos and descriptions.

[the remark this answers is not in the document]
maarcingebala: I think we shouldn't do that since a product without a variant is not a valid product in our data model. This would also require changes in the storefront so I think it's easier to not return them now and if we decide to render them, implement that properly in the storefront.

saleor/graphql/product/tests/test_product_pagination.py
```diff
@@ -463,7 +453,6 @@ def test_products_pagination_with_sorting_by_attribute(
     "filter_by, products_order",
     [
         ({"isPublished": False}, ["Product2", "ProductProduct1"]),
-        ({"price": {"gte": 8, "lte": 12}}, ["Product1", "ProductProduct2"]),
```
fowczarek, line 466: We still have this filter option.

saleor/graphql/product/tests/test_product_pagination.py
```diff
@@ -395,10 +389,6 @@ def products_for_pagination(product_type, color_attribute, category, warehouse):
             {"field": "NAME", "direction": "DESC"},
             ["ProductProduct2", "ProductProduct1", "Product3"],
         ),
-        (
-            {"field": "PRICE", "direction": "ASC"},
```
fowczarek, line 399: We still have this sorting option.

saleor/graphql/product/tests/test_product.py
```diff
@@ -458,10 +456,10 @@ def test_products_query_with_filter_collection(
 @pytest.mark.parametrize(
     "products_filter",
     [
-        {"price": {"gte": 5.0, "lte": 9.0}},
         {"minimalPrice": {"gte": 1.0, "lte": 2.0}},
         {"isPublished": False},
         {"search": "Juice1"},
+        {"price": {"gte": 1.0, "lte": 2.0}},
```
fowczarek, line 462: If you restore some code please do it without changing line order. It's easier to do a review without looking at the changes that "change nothing".

PR #5707, opened 2020-05-27
Introduce a dataloader for available quantity

patrys:
It's not very readable but we want to return the maximum available in any single shipping zone which means the loader needs to operate on several tables at once.

TODO:
- ~~Figure out how to handle variant.track_inventory, it seems wasteful to calculate things for products that don't care about stock, maybe we should rethink requiring a stock record to exist?~~

New flow for products without inventory tracking:

- If staff user create an available product without inventory tracking, the product is available to buy in any shipping zone and max quantity is `settings.MAX_CHECKOUT_LINE_QUANTITY`. 

- Stocks with any quantity for a product without inventory tracking are needed when any staff user want to create fulfillment with this product. (Fulfillment don't decrease  quantity for a product without inventory tracking)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/warehouse/dataloaders.py
```diff
@@ -0,0 +1,88 @@
...
+    def batch_load_country(
+        self, country_code: CountryCode, variant_ids: Iterable[int]
+    ) -> Iterable[Tuple[int, int]]:
+        filters = Q(product_variant_id__in=variant_ids)
+        if country_code:
+            filters &= Q(warehouse__shipping_zones__countries__contains=country_code)
+        results = (
+            Stock.objects.filter(filters)
+            .annotate(
+                available_quantity=Sum("quantity", distinct=True)
```
maarcingebala, line 55: Why `Sum` aggregation on `quantity`, which is a single value in each stock row? Wouldn't `F("quantity")` be enough? In `StockQuerySet` queryset we the following function:

```
def annotate_available_quantity(self):
        return self.annotate(
            available_quantity=F("quantity")
            - Coalesce(Sum("allocations__quantity_allocated"), 0)
        )
```

Ideally, we should reuse this function in all places that need to calculate the available quantity on stock records (there are few more functions that do that in the `warehouse/availability.py` module.

PR #5728, opened 2020-06-02
Invoices webhooks

tomaszszymanski129:
I want to merge this change because it add webhook events for invoicing.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/invoice/mutations.py
```diff
@@ -274,4 +274,5 @@ def perform_mutation(cls, _root, info, **data):
         cls.clean_instance(info, instance)
         send_invoice.delay(instance.pk)
         events.invoice_sent_event(user=info.context.user, invoice=instance)
+        info.context.plugins.invoice_sent(instance, instance.order.user.email)
```
korycins, line 277: Don't you think that this call and `events.invoice_sent_event` should be called inside `send_invoice.delay` ?
We don't know the time when celery worker will process this task. This could cause strange cases like:
 - we have events that invoice was send but we didn't send it because our celery worker is under the huge load
 - we have events that invoice was send but sending an email for some reason failed.

saleor/plugins/webhook/plugin.py
```diff
@@ -45,6 +47,32 @@ def order_updated(self, order: "Order", previous_value: Any) -> Any:
...
+    def invoice_request(
+        self,
+        order: "Order",
+        invoice: "Invoice",
+        number: Optional[str],
+        previous_value: Any,
+    ) -> Any:
+        if not self.active:
+            return previous_value
+        invoice_data = generate_invoice_payload(invoice)
```
korycins, line 59: Do you want to send `invoice` object here? AFAIK this method will be called when staff user wants to generate an Invoice, which means that this object will be empty. Shouldn't we send here an order data?

saleor/webhook/payloads.py
```diff
@@ -105,6 +106,12 @@ def generate_order_payload(order: "Order"):
     return order_data
 
 
+def generate_invoice_payload(invoice: "Invoice"):
+    serializer = PayloadSerializer()
+    invoice_fields = ("id", "number", "external_url", "created")
```
korycins, line 111: IMHO You should consider an `order` field here

saleor/graphql/invoice/mutations.py
tomaszszymanski129: Very good point! Will defienietely move that.

saleor/plugins/webhook/plugin.py
tomaszszymanski129: invoice might get in handy since that contains job id. And yes, will add order inside.

saleor/graphql/invoice/tests/test_invoice.py
```diff
@@ -470,8 +470,7 @@ def test_create_invoice_empty_params(staff_api_client, permission_manage_orders,
     ).exists()
 
 
-@patch("saleor.invoice.emails.send_invoice.delay")
```
korycins, line 473: Don't you need a mock here?  Do we want to call `send_templated_mail`?

saleor/invoice/emails.py
```diff
@@ -23,7 +25,11 @@ def collect_invoice_data_for_email(invoice_pk, template):
...
-def send_invoice(invoice_pk):
+def send_invoice(invoice_pk, user_pk):
     """Send an invoice to user of related order with URL to download it."""
-    email_data = collect_invoice_data_for_email(invoice_pk, "order/send_invoice")
+    invoice = Invoice.objects.get(pk=invoice_pk)
+    email_data = collect_invoice_data_for_email(invoice, "order/send_invoice")
     send_templated_mail(**email_data)
+    events.invoice_sent_event(user=User.objects.get(pk=user_pk), invoice=invoice)
+    manager = get_plugins_manager()
+    manager.invoice_sent(invoice, invoice.order.user.email)
```
fowczarek, line 35: Not all orders have assigned users.

saleor/invoice/emails.py
```diff
@@ -23,7 +25,11 @@ def collect_invoice_data_for_email(invoice_pk, template):
 
 
 @app.task
-def send_invoice(invoice_pk):
+def send_invoice(invoice_pk, user_pk):
```
fowczarek, line 28: Maybe we should use more specific name for `user_pk` like `staff_user_pk` or `requester_user_pk`. 
Another solution is add docstring with args descriptions. 

Currently, I don't know it a customer pk or staff user pk.

saleor/invoice/emails.py
```diff
@@ -1,14 +1,16 @@
...
+from ..plugins.manager import get_plugins_manager
 from .models import Invoice
 
 
-def collect_invoice_data_for_email(invoice_pk, template):
+def collect_invoice_data_for_email(invoice, template):
     """Collect the required data for sending emails."""
-    invoice = Invoice.objects.get(pk=invoice_pk)
-    recipient_email = invoice.order.user.email
+    recipient_email = invoice.order.user_email or invoice.order.user.email
```
fowczarek, line 13: If I create a draft order with billing address and without email and customer? 
I'm not sure what should we do.

fowczarek: Egzample:
image

tomaszszymanski129: For now we only allow even requesting an Invoice for an Order that is not draft but indeed if that transforms to order without `order.user_email` or `order.user.email` then we effectively have no way to know the email to send it to?
So I would add that check to SendInvoice mutation unless there is some other way we can get the email address?
Please advise, @maarcingebala maybe?

tomaszszymanski129: @fowczarek discussed with Marcin, there is a requirement of adding the email in checkout so that's saved on `order.user_email`, otherwise we'll have the user so we can still get the email, so we don't have a case where both can be empty - current implementation should be okay.

saleor/order/models.py
```diff
@@ -317,6 +317,9 @@ def total_captured(self):
     def total_balance(self):
         return self.total_captured - self.total.gross
 
+    def get_email(self):
```
maarcingebala, line 320: We already have `get_customer_email`.

PR #5736, opened 2020-06-05
Explicit country assignment in default shipping zones

maarcingebala:
This PR changes the way in which default shipping zones are created in API. From now on, when creating a default shipping zone, we explicitly assign all countries that are not yet in any other shipping zone. The "default" field in the ShippingZone model is left for compatibility with UI (rendering the "default" checkbox in the dashboard).

In a separate PR, I'll also rewrite how countries are stored, instead of the inefficient comma-separated format we'll have regular M2M relation, but that requires changes in more places and I wanted to do that separately.

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/shipping/migrations/0018_default_zones_countries.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 3.0.6 on 2020-06-05 14:35
+
+from django.db import migrations
+
+from ..utils import get_countries_without_shipping_zone
```
IKarbowiak, line 5: Shouldn't we copy here this function instead of importing it?

maarcingebala: I wasn't sure but can change that.

PR #5757, opened 2020-06-10
Add support for exporting products by app

IKarbowiak:
Allow exporting products by app.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/csv/emails.py
```diff
@@ -21,7 +21,10 @@
 def send_email_with_link_to_download_file(
     export_file: "ExportFile", template_name: str
 ):
-    recipient_email = export_file.created_by.email
+    user = export_file.user
+    if not user:
+        return
```
maarcingebala, line 26: In what cases there will be no user? When an app is requesting the export? How would we provide the export results then?

saleor/graphql/csv/filters.py
```diff
@@ -5,19 +5,29 @@
...
 
 
-def filter_created_by(qs, _, value):
+def filter_user(qs, _, value):
     user_fields = [
-        "created_by__pk",
-        "created_by__first_name",
-        "created_by__last_name",
-        "created_by__email",
+        "user__pk",
```
maarcingebala, line 10: Filtering by PK was here before but I'm wondering if this is useful - PKs are internal and are not known to the user so I'm not sure we should filter by them.

saleor/csv/emails.py
IKarbowiak: I talked with @korycins about that. The app can query `ExportFile` with the use of ExportFile ID returned by mutation (current solution). Another possible solution, which can be implemented, is adding an optional `email` field to `exportProducts` mutation.

saleor/graphql/csv/mutations.py
```diff
@@ -73,14 +73,19 @@ class Meta:
...
         export_info = cls.get_export_info(input["export_info"])
         file_type = input["file_type"]
 
-        export_file = csv_models.ExportFile.objects.create(created_by=user)
-        export_started_event(export_file=export_file, user=user)
+        app = info.context.app
+        if app:
+            kwargs = {"app": app}
+        else:
+            kwargs = {"user": info.context.user}
```
tomaszszymanski129, line 85: maybe `kwargs = {"app": app} if app else {"user": info.context.user}` ?

saleor/csv/emails.py
maarcingebala: OK, but I think these functions are expected to send the email, and these checks should be moved outside, to where these tasks are actually called. Same for the other email sending function where we added this check.

IKarbowiak: I did it in this way first (as you proposed), but then I moved it here because it will fail when user in export_file will be None. But maybe I should add the `user` parameter to force providing it.

maarcingebala: Adding explicit `user` param sounds good to me, I think we do that in other email functions, or we explicitly pass email as string even. It's not very consistent, unfortunately.

saleor/graphql/csv/sorters.py
```diff
@@ -7,7 +7,8 @@ class ExportFileSortField(graphene.Enum):
     STATUS = ["status"]
     CREATED_AT = ["created_at"]
     UPDATED_AT = ["updated_at"]
-    CREATED_BY = ["created_by__id"]
+    USER = ["user__id"]
```
fowczarek, line 10: Maby sort by `user__email`, `app_name`?

saleor/graphql/csv/types.py
```diff
@@ -63,12 +74,21 @@ def resolve_url(root: models.ExportFile, info):
...
 
     @staticmethod
-    def resolve_created_by(root: models.ExportFile, info):
+    def resolve_user(root: models.ExportFile, info):
         requestor = get_user_or_app_from_context(info.context)
-        if requestor == root.created_by or requestor.has_perm(
-            AccountPermissions.MANAGE_USERS
+        if (
+            requestor == root.user
+            or requestor.has_perm(AccountPermissions.MANAGE_USERS)
```
fowczarek, line 81: I think it should be only `MANAGE_STAFF` permission. 
Export could be created only by staff users.

saleor/graphql/csv/sorters.py
IKarbowiak: `app__name` sounds good, but I am not totally convinced about sorting by user email. Maybe first user `last_name` and then `first_name`

maarcingebala: I was thinking if we need sorting by user/app at all, we don't have any designs for that yet so it would be fine for me to no add it all.

IKarbowiak: That's true. So I'll remove that.

saleor/csv/events.py
```diff
@@ -1,49 +1,64 @@
...
+        export_file=export_file, user=user, app=app, type=ExportEvents.EXPORT_SUCCESS
     )
 
 
 def export_failed_event(
-    *, export_file: "ExportFile", user: User, message: str, error_type: str
+    *,
+    export_file: "ExportFile",
+    user: UserType = None,
+    app: "App" = None,
```
korycins, line 34: `Optional["App"]`? ;>

saleor/graphql/csv/filters.py
```diff
@@ -5,19 +5,28 @@
...
+        "user__first_name",
+        "user__last_name",
+        "user__email",
+    ]
+    qs = filter_by_query_param(qs, value, user_fields)
+    return qs
+
+
+def filter_app(qs, _, value):
+    user_fields = [
```
korycins, line 19: app_fields

saleor/graphql/csv/filters.py
```diff
@@ -5,19 +5,28 @@
...
+        "app__pk",
```
korycins, line 20: not sure if we need to have filter over `pk`

PR #5766, opened 2020-06-12
Strip warehouse name in mutations

### LINKED ISSUE #5673: Warehouse name is not stripped in validation
User can create a warehouse without a name by passing empty spaces as the name in API.

1. Run this mutation:
```
mutation {
  createWarehouse(
    input: {
      name: " "
      slug: "test-12345"
      address: {
        country: PL
        city: "Test"
        streetAddress1: "Test"
        postalCode: "12-345"
      }
    }
  ) {
    warehouseErrors {
      field
      code
    }
    warehouse {
      id
      name
      slug
    }
  }
}
```
Note that `name` is ` `.

2. Mutation creates a warehouse with an empty name.

### What I expected to happen
I should get the error:
```
{
    "field": "name",
    "code": "REQUIRED"
},
```

koradon:

fixes #5673 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/core/utils/__init__.py
+++ b/saleor/graphql/core/utils/__init__.py
@@ -108,3 +108,15 @@ def get_duplicates_ids(first_list, second_list):
 def get_duplicated_values(values):
     """Return set of duplicated values."""
     return {value for value in values if values.count(value) > 1}
+
+
+def validate_and_strip_field_str_value(cleaned_input, field_name: str):
+    """Strip and validate field value."""
+    if field_name in cleaned_input:
+        field_value = cleaned_input.get(field_name)
+        field_value = field_value.strip() if field_value else ""
+        if field_value:
+            cleaned_input[field_name] = field_value
+        else:
+            raise ValidationError(f"{field_name.capitalize()} is required.")
+    return cleaned_input
--- a/saleor/graphql/core/utils/__init__.py
+++ b/saleor/graphql/core/utils/__init__.py
@@ -110,13 +110,12 @@ def get_duplicated_values(values):
     return {value for value in values if values.count(value) > 1}
 
 
-def validate_and_strip_field_str_value(cleaned_input, field_name: str):
+def validate_required_string_field(cleaned_input, field_name: str):
     """Strip and validate field value."""
-    if field_name in cleaned_input:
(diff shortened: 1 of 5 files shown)


saleor/graphql/warehouse/tests/test_warehouse.py
```diff
@@ -374,6 +374,42 @@ def test_mutation_create_warehouse(
     assert created_warehouse["slug"] == warehouse.slug
 
 
+def test_mutation_create_warehouse_does_not_create_when_name_is_empty_string(
+    staff_api_client, permission_manage_products, shipping_zone
+):
+    Warehouse.objects.all().delete()
```
fowczarek, line 380: Why are you removing all the warehouses?

koradon: Honestly, I copied the test above and changed just what I needed. Missed that line.

saleor/graphql/core/utils/__init__.py
```diff
@@ -108,3 +108,15 @@ def get_duplicates_ids(first_list, second_list):
 def get_duplicated_values(values):
     """Return set of duplicated values."""
     return {value for value in values if values.count(value) > 1}
+
+
+def validate_and_strip_field_str_value(cleaned_input, field_name: str):
```
maarcingebala, line 113: A detail, but maybe `validate_required_string_field` would sound more universal?

koradon: Yes, I was thinking about it but it's not exactly the case here. Field `name` is required to create mutation and optional for the update.
`not_empty` or other combinations are also not the best IMO. I tried to connect validation and the exact action here. But I am not happy either with the result.
(Naming is hard sometimes)

maarcingebala: The difference between updating and creating is handled by the schema where the field is required or not. Another option is to move `if field_name in cleaned_input` outside and in the function assume that `cleaned_input["value"]` is present? 

We have the same logic in `OrderAddNote` and we should reuse this function there as well.

koradon: Ok I will reuse it there.

PR #5769, opened 2020-06-14
[Suggestion / POC] Cart allocation for logged in users ala Zalando Lounge plugin

Boquete:
(not ready for full merge) 
The below solutions allow us to allocate products for logged in users.
In big sales actions, users get anoyed that before they get to end of checkout all of the products are already sold.

Current solution need two more thing: 
1) crone task started each minute to delete checkout objects where user is not None, and checkout expired is True, and delete such checkout
2) on frontend - after user add item to "cart" it should create new Checkout obj for user, instead of adding cart items to local storage, use mutation CheckoutLineAdd/Update etc.
3) on frontend - if checkout is deleted during some stage of putting in information (shipping etc.) the error will be that Node couldn't be resolved
3) on frontend - some timer to show how much time left user have untill checkout expires.

Whole idea is based on Zalando Lounge. There's also checkbox in plugin configuration, defines if user add something to checkout, should it add extra time to checkout validity or expiration date should only be created when checkout is created (or during first iteration) 

Once again - it's just POC - but it works. Allocate stock function should be prettier for sure, but made this PR to show our idea for it.

# Impact

* [X] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/warehouse/management.py
```diff
@@ -54,20 +82,29 @@ def allocate_stock(
             (quantity - quantity_allocated), quantity_available_in_stock
         )
         if quantity_to_allocate > 0:
-            allocations.append(
-                Allocation(
-                    order_line=order_line,
-                    stock=stock,
-                    quantity_allocated=quantity_to_allocate,
+            if type(line) == OrderLine:
```
Boquete, line 85: those two if's are expecially ugly

saleor/warehouse/management.py
```diff
@@ -24,13 +26,39 @@ def allocate_stock(
...
-        .get_variant_stocks_for_country(country_code, order_line.variant)
-        .order_by("pk")
+            .get_variant_stocks_for_country(country_code, line.variant)
+            .order_by("pk")
     )
 
+    if type(line) == CheckoutLine:
+        checkout_allocations = Allocation.objects.filter(stock__in=stocks, checkout_line=line)
+        if checkout_allocations is not None:
+            # checkout_allocations._raw_delete(checkout_allocations.db)
```
Boquete, line 41: raw_delete could be faster, as there is not dependencies which has to be deleted with Allocation obj

saleor/warehouse/management.py
```diff
@@ -24,13 +26,39 @@ def allocate_stock(
...
+        if not checkout.expired():
+            checkout_line = checkout.lines.get(variant=line.variant)
+            if checkout_line:
+                checkout_allocations = Allocation.objects.filter(stock__in=stocks, checkout_line=checkout_line)
+                quantity_allocated_for_checkout = 0
+                for allocation in checkout_allocations:
+                    quantity_allocated_for_checkout += allocation.quantity_allocated
+
+                if quantity != quantity_allocated_for_checkout:
+                    checkout_allocations.delete()
```
Boquete, line 54: deleting old allocations might be not best option, but it looks like safest option - updateing allocations on multiple stocks sounds like something which will most likely fail at some point

saleor/warehouse/models.py
```diff
@@ -128,6 +137,13 @@ class Allocation(models.Model):
     )
     quantity_allocated = models.PositiveIntegerField(default=0)
 
+    def clean(self):
+        if self.order_line is None and self.checkout_line is None:
+            raise ValidationError("Order line or checkout line is required")
+
     class Meta:
-        unique_together = [["order_line", "stock"]]
```
Boquete, line 132: moving to UniqueConstraint as it's faster then unique_together, and also might be decaprated in the future

https://docs.djangoproject.com/en/3.0/ref/models/options/#django.db.models.Options.unique_together

patrys:
May I suggest a different approach? I was thinking (disclaimer: this is not an official statement of the team) about a potential API to create temporary allocations for particular users, entirely independent of the Checkout API.

Consider a mutation like `createStockReservation(productVariant: ..., quantity: ..., duration: ...)`.
* The core creates a reservation object with an expiration time.
* The core treats the reservation as allocated stock for all users but the owner of the reservation (so if you reserve the last item in stock you can still place an order that contains it).
* Once the owner places an order, the quantities ordered are subtracted from their reservations.

There would have to be a setting limiting the maximum duration allowed and a setting limiting the number of allocated stock a single person can hold to prevent abuse. I think if we approach this the right way, such an API could be used in other creative ways. For example, a privileged version of the mutation could allow Apps to create such reservations for other users while not being bound by either the maximum duration or the quantity limit.

Boquete:
> * Once the owner places an order, the quantities ordered are subtracted from their reservations.

So your solution is more like adding new models for reservation and also keep all mutation separate. This sounds good although: 
right now quantity available is sql query which counts / SUM Allocation objects. If we want to make reservation seperate from checkout, should we maybe.. change allocation model and instead of order_line required we should add User field over there which can be used intead of order line? And keep expiration date in some Reservation model which has: fk do allocation, expiration date?

patrys:
CC @fowczarek

fowczarek:
Adding the "User field" can be a good solution but we should remember that `Checkouts` can be created by an anonymous users(For this kind of users we have an only email address.).

Boquete:
> Adding the "User field" can be a good solution but we should remember that `Checkouts` can be created by an anonymous users(For this kind of users we have an only email address.).

I think it’s bad idea for implementing allocation system to anonymous users. Blocking all inventory with one bot would take less then few seconds.
That’s why whole solution is based on logged in users only

PR #5794, opened 2020-06-17
Remove creating default variants for product

d-wysocki:
I want to merge this change because...we want to create default variants for product with createProductVariant mutation

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested


d-wysocki:
This part is only for delete creating defaults variant in `productCreate` and `productUpdate` mutation and its ready for partial code review.

fowczarek:
We have a lot of new missed line in the coverage report. Probably you should remove some class function from mutation(e.g. `clean_sku`) or add a new tests for this lines.

d-wysocki:
Code is ready for CR, working on add more unit tests.

saleor/product/utils/product_variants.py
```diff
@@ -0,0 +1,26 @@
...
+
+from ..error_codes import ProductErrorCode
+
+
+def product_variant_exist(product):
+    if product.variants.first() is None:
+        raise ValidationError(
+            {
+                "product": ValidationError(
+                    f"Cannot add product `{product.id}` without variant.",
```
IKarbowiak, line 11: I wondering if error message shouldn't be more generic, like: `Cannot manage product '{product.id}' without variant.`

saleor/graphql/discount/mutations.py
```diff
@@ -67,6 +69,7 @@ def remove_catalogues_from_node(cls, node, input):
         products = input.get("products", [])
         if products:
             products = cls.get_nodes_or_error(products, "products", Product)
+            products_variant_exist(products)
```
IKarbowiak, line 72: Do we need this check here? If we removing connections between products and the node which already exists?

saleor/graphql/product/mutations/products.py
```diff
@@ -934,23 +868,6 @@ def get_instance(cls, info, **data):
...
-
-            variant = models.ProductVariant.objects.create(
-                product=instance,
-                track_inventory=track_inventory,
-                sku=sku,
-                price_amount=variant_price,
-            )
-            stocks = cleaned_input.get("stocks")
-            if stocks:
-                cls.create_variant_stocks(variant, stocks)
```
fowczarek, line 953: Should we remove this function?

saleor/product/utils/product_variants.py
```diff
@@ -0,0 +1,26 @@
+from django.core.exceptions import ValidationError
+
+from ..error_codes import ProductErrorCode
+
+
+def product_variant_exist(product):
+    if product.variants.first() is None:
+        raise ValidationError(
```
fowczarek, line 8: We should raise graphql error in "core utils".  If you want to raise errors in utils we should create custom exceptions like `InsufficientStock` and handle it in mutation.

saleor/product/utils/product_variants.py
```diff
@@ -0,0 +1,26 @@
...
+def products_variant_exist(products):
+    errors = []
+    for product in products:
+        try:
+            product_variant_exist(product)
+        except ValidationError as exc:
+            errors.append(exc.message_dict)
+
+    if errors:
+        raise ValidationError(errors, code=ProductErrorCode.PRODUCT_WITHOUT_VARIANT)
```
fowczarek, line 26: Same here.

saleor/product/utils/product_variants.py
```diff
@@ -0,0 +1,26 @@
...
+                )
+            }
+        )
+
+
+def products_variant_exist(products):
+    errors = []
+    for product in products:
+        try:
+            product_variant_exist(product)
```
fowczarek, line 21: For 100 products we execute 100 database queries :cry:

saleor/tests/fixtures.py
```diff
@@ -1061,6 +1061,9 @@ def product_with_images(product_type, category, media_root):
         category=category,
         is_published=True,
     )
+    ProductVariant.objects.create(
```
fowczarek, line 1064: Why? 

`product_type` have `size_attribute` as `variant_attribute`. 
If you want to create a Variant for this product you should assign value for this attribute.

saleor/product/utils/product_variants.py
```diff
@@ -0,0 +1,26 @@
+from django.core.exceptions import ValidationError
+
+from ..error_codes import ProductErrorCode
+
+
+def product_variant_exist(product):
```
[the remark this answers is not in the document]
maarcingebala: Also, the name could be more descriptive, something like `check_if_variant_exists`? Not perfect since the function not only checks that but also raises an error. Maybe this function should just do the check and return a bool, while error should be raised in mutation modules.

saleor/product/utils/product_variants.py
```diff
@@ -0,0 +1,26 @@
...
+    if product.variants.first() is None:
```
maarcingebala, line 7: `exists()` is the fastest way to check the existence of objects in the database. But anyway, this function is also used in `products_variant_exists` and both `first()` and `exists()` will result in duplicated database queries.

saleor/graphql/product/mutations/products.py
```diff
@@ -1656,6 +1540,7 @@ def perform_mutation(cls, _root, info, product_id, images_ids):
         product = cls.get_node_or_error(
             info, product_id, field="product_id", only_type=Product
         )
+        product_variant_exist(product)
```
maarcingebala, line 1543: Reordering product images seems independent from variants, why do we disallow it here?

saleor/graphql/product/mutations/products.py
maarcingebala: If it's no longer needed then yes.

saleor/graphql/schema.graphql
```diff
@@ -3446,6 +3442,7 @@ enum ProductErrorCode {
   REQUIRED
   UNIQUE
   VARIANT_NO_DIGITAL_CONTENT
+  PRODUCT_WITHOUT_VARIANT
```
maarcingebala, line 3445: This code could be more descriptive, e.g. `CANNOT_MANAGE_PRODUCT_WITHOUT_VARIANT`.

maarcingebala: We need to create a ticket for the frontend team to handle this error in the dashboard.

maarcingebala:
@d-wysocki We said that products that don't have any variants shouldn't be returned in the public products query, do we have any logic for that?

saleor/graphql/product/mutations/products.py
d-wysocki: This function is related with variants, if we want to separate creating products and creating variants, we also should remove this function.

saleor/product/utils/product_variants.py
d-wysocki: good catch :)

saleor/graphql/product/mutations/products.py
d-wysocki: good catch, thanks. I'll remove this.

d-wysocki:
@maarcingebala I did change in my last PR about returning products with variants in public query.
https://github.com/mirumee/saleor/pull/5706/files#diff-79417314910e0179330376e3da486e0cR130

saleor/product/utils/product_variants.py
```diff
@@ -0,0 +1,21 @@
+from ...graphql.core.utils import from_global_id_strict_type
```
fowczarek, line 1: I'm not sure is `product_variants.py` is good name for this function?

saleor/graphql/discount/mutations.py
```diff
@@ -49,6 +51,17 @@ def recalculate_minimal_prices(cls, products, categories, collections):
     def add_catalogues_to_node(cls, node, input):
         products = input.get("products", [])
         if products:
+            products_without_variants = get_products_without_variants(products)
+            if products_without_variants:
+                raise ValidationError(
+                    {
+                        "products": ValidationError(
+                            f"Cannot manage products without variants. "
+                            f"Products ids: {products_without_variants}.",
```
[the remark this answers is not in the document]
maarcingebala: We could do that actually, but maybe we should discuss it with the frontend team. I'm fine with adding it in separate PR as well. This message is only information for the developer, it won't help to render anything useful in the frontend.

saleor/product/utils/product_variants.py
maarcingebala: Maybe we don't need a new module for these two functions? We can move them to `product/utils/__init__.py`.

saleor/product/utils/product_variants.py
```diff
@@ -0,0 +1,21 @@
...
+from ...product.models import Product
+
+
+def get_products_without_variants(products_list):
+    """Return list of product's ids without variants."""
+    products_ids = get_products_from_global_ids(products_list)
```
maarcingebala, line 7: If this function is a global utility function (not living in the GraphQL-layer) I think it should already receive resolved product objects and not GraphQL IDs that you have to resolve here.

saleor/graphql/product/mutations/products.py
```diff
@@ -359,12 +360,23 @@ def perform_mutation(cls, _root, info, collection_id, products):
...
+        if products_without_variants:
+            raise ValidationError(
+                {
+                    "products": ValidationError(
+                        f"Cannot manage products without variants. "
+                        f"Products ids: {products_without_variants}.",
+                        code=ProductErrorCode.CANNOT_MANAGE_PRODUCT_WITHOUT_VARIANT,
+                    )
+                }
+            )
```
maarcingebala, line 373: Maybe something like this:

```
products = cls.get_nodes_or_error(products, "products", Product)
clean_products(products)
# rest of mutation logic
```

And above we'd add a new function in this mutation:
```
def clean_products(...):
    products_without_variants = get_products_without_variants(products)
    if products_without_variants:
        raise ValidationError(...)
```

That would be consistent with the rest of the codebase.

maarcingebala:
Looks good now. We'll wait with merging for the frontend part since these changes would break the dashboard.

PR #5802, opened 2020-06-19
Feature/add jwt for apps

korycins:
- Add new field to App type - `accessToken`
- rebuild AuthBackend to use `effective_permissions`
- add `permission` field to app token
- add tests for decorator `permission_required`
- add tests for authentication backend

Expected usage of accessToken:
mermaid-diagram-20200619122753

### DIFF
--- a/saleor/account/models.py
+++ b/saleor/account/models.py
@@ -4,11 +4,12 @@ from django.conf import settings
 from django.contrib.auth.models import (
     AbstractBaseUser,
     BaseUserManager,
+    Permission,
     PermissionsMixin,
 )
 from django.contrib.postgres.fields import JSONField
 from django.db import models
-from django.db.models import Q, Value
+from django.db.models import Q, QuerySet, Value
 from django.forms.models import model_to_dict
 from django.utils import timezone
 from django.utils.crypto import get_random_string
@@ -161,6 +162,32 @@ class User(PermissionsMixin, ModelWithMetadata, AbstractBaseUser):
             (AccountPermissions.MANAGE_USERS.codename, "Manage customers."),
             (AccountPermissions.MANAGE_STAFF.codename, "Manage staff."),
         )
+
+    def __init__(self, *args, **kwargs):
+        super(User, self).__init__(*args, **kwargs)
+        self._effective_permissions = None
+
+    @property
+    def effective_permissions(self) -> "QuerySet[Permission]":
+        if self._effective_permissions is None:
+            if self.is_superuser:
+                self._effective_permissions = Permission.objects.prefetch_related(
+                    "content_type"
+                ).all()
+            else:
+                permissions = self.user_permissions.prefetch_related(
+                    "content_type"
+                ).all()
+                self._effective_permissions = permissions | Permission.objects.filter(
+                    group__user=self
+                )
+        return self._effective_permissions
+
+    @effective_permissions.setter
+    def effective_permissions(self, value: "QuerySet[Permission]"):
+        self._effective_permissions = value
+        # Drop cache for authentication backend
+        self._effective_permissions_cache = None
 
     def get_full_name(self):
         if self.first_name or self.last_name:
--- a/saleor/account/models.py
+++ b/saleor/account/models.py
@@ -1,6 +1,7 @@
 from typing import Union
 
 from django.conf import settings
+from django.contrib.auth.models import _user_has_perm  # type: ignore
 from django.contrib.auth.models import (
     AbstractBaseUser,
     BaseUserManager,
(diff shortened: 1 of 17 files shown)


saleor/core/jwt.py
```diff
@@ -103,6 +108,49 @@ def get_user_from_payload(payload: Dict[str, Any]) -> Optional[User]:
...
 def get_user_from_access_token(token: str) -> Optional[User]:
     payload = jwt_decode(token)
-    if payload["type"] != JWT_ACCESS_TYPE:
+    if payload["type"] not in [JWT_ACCESS_TYPE, JWT_THIRDPARTY_ACCESS_TYPE]:
         return None
-    return get_user_from_payload(payload)
+    permission_limits = payload.get(PERMISSION_LIMITS_FIELD, None)
+    user = get_user_from_payload(payload)
+    if permission_limits is not None:
+        user.permission_limits = permission_limits  # type: ignore
```
NyanKiyoshi, line 116: I don't like too much having a magic field, sometimes it can be there, sometimes not.

Could we have a default value in `User` and comment it?

saleor/core/jwt.py
```diff
@@ -103,6 +108,49 @@ def get_user_from_payload(payload: Dict[str, Any]) -> Optional[User]:
...
+    The token which can be used by App has additional field defining the permissions
+    assigned to it. The permissions set is the intersection of user permissions and
+    app permissions.
+    """
+
+    permissions_dict = {
+        enum.codename: enum.name
+        for permission_enum in PERMISSIONS_ENUMS
+        for enum in permission_enum
+    }
```
fowczarek, line 133: Maby we should create the function for this?
We have similar functions `get_permissions_codename` and `get_permissions_enum_list` in `saleor\core\permissions.py`

saleor/core/jwt.py
patrys: Or call the field `effective_permissions` and populate with whatever sits in the database when not set by the token. This way we can only check one place (and we can probably speed up most of the permission-related checks by removing the database queries required to insert the permissions).

saleor/core/jwt.py
```diff
@@ -103,6 +108,49 @@ def get_user_from_payload(payload: Dict[str, Any]) -> Optional[User]:
...
+    app_permissions = app.permissions.all()
+    app_permission_enums = {permissions_dict[perm.codename] for perm in app_permissions}
+
+    if user.is_superuser:
+        user_permissions = get_permissions()
```
fowczarek, line 138: When I inspecting the coverage report I found that we don't have any test for this line. Can you add a test for this function as a superuser.

saleor/core/jwt.py
korycins: How would you like to name it?

fowczarek: Maby `get_permissions_dict` ?

saleor/core/jwt.py
korycins: I've changed the workflow. I dropped the changes in `permission_required` and extend our AuthBackend to use `effective_permissions`.

saleor/core/jwt.py
```diff
@@ -110,10 +110,10 @@ def get_user_from_access_token(token: str) -> Optional[User]:
     payload = jwt_decode(token)
     if payload["type"] not in [JWT_ACCESS_TYPE, JWT_THIRDPARTY_ACCESS_TYPE]:
         return None
-    permission_limits = payload.get(PERMISSION_LIMITS_FIELD, None)
+    permissions = payload.get(PERMISSIONS_FIELD, None)
     user = get_user_from_payload(payload)
-    if permission_limits is not None:
-        user.permission_limits = permission_limits  # type: ignore
+    if user and permissions is not None:
+        user.effective_permissions = get_permissions_from_names(permissions)
```
patrys, line 116: Shouldn't we cross-check those with the database permissions? Can a token grant me permission I did not previously hold?

NyanKiyoshi: Unless the secret is known or was leaked (not impossible!), the HMAC should ensure nothing was altered/injected. What we can't ensure, is that permissions are actually up to date.

korycins: @patrys Yes, I am adding it now :)

saleor/graphql/account/mutations/jwt.py
```diff
@@ -71,10 +71,8 @@ class Meta:
 
     @classmethod
     def get_user(cls, info, data):
-        user = authenticate(
-            request=info.context, username=data["email"], password=data["password"],
-        )
-        if not user:
+        user = models.User.objects.filter(email=data["email"]).first()
+        if not user or not user.is_active or not user.check_password(data["password"]):
```
IKarbowiak, line 75: Maybe we should create utils function for that?

saleor/graphql/account/mutations/jwt.py
```diff
@@ -69,11 +71,20 @@ class Meta:
...
+    @classmethod
+    def _retrieve_user_from_credentials(cls, email, password) -> Optional[models.User]:
+        user = models.User.objects.filter(email=email).first()
+        if not user:
+            return None
+        if not user.is_active:
+            return None
+        if not user.check_password(password):
+            return None
+        return user
```
maarcingebala, line 83: A bit shorter and should do the job:

```suggestion
        user = models.User.objects.filter(email=email, is_active=True).first()
        if user and user.check_password(password):
            return user
        return None
```

What do you think?

saleor/account/models.py
```diff
@@ -162,6 +164,29 @@ class Meta:
...
+    def effective_permissions(self) -> "QuerySet[Permission]":
+        if self._effective_permissions is None:
+            self._effective_permissions = get_permissions()
+            if not self.is_superuser:
+                self._effective_permissions = self._effective_permissions.filter(
+                    user=self
+                )
+                self._effective_permissions |= Permission.objects.filter(
+                    group__user=self
+                )
```
patrys, line 181: Shouldn't that be `.filter(Q(user=self) | Q(group__user=self))`? It would be even better if it used a subquery with `exists` to eliminate the problem of duplicate permissions.

PR #5816, opened 2020-06-22
Better way of handling token expiration and old type of jwt tokens

korycins:
- Change way of handling tokens when JWT_EXPIRE is set to False
- Raise exceptions when token payload doesn't have all required fields

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 All notable, unreleased changes to this project will be documented in this file. For the released changes, please visit the [Releases](https://github.com/mirumee/saleor/releases) page.
 
+
 ## [Unreleased]
 
 - Add our implementation of UUID scalar - #5646 by @koradon
@@ -9,9 +10,11 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fix specific product voucher in draft orders - #5727 by @fowczarek
 - Explicit country assignment in default shipping zones - #5736 by @maarcingebala
 - Drop `json_content` field from the `Menu` model - #5761 by @maarcingebala
-- Refactor JWT support - These changes could require a handling JWT token in the storefront. Storefront needs to handle a case when the backend returns the exception about the invalid token. - #5734, #5816 by @korycins
 - Strip warehouse name in mutations - #5766 by @koradon
 - Add missing OrderEvents during checkout flow - #5684 by @koradon
+    ### Breaking Changes
+   - Refactor JWT support - These changes could require a handling JWT token in the storefront. Storefront needs to handle a case when the backend returns the exception about the invalid token. - #5734, #5816 by @korycins
+
 
 ## 2.10.1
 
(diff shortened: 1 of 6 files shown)


saleor/core/jwt.py
```diff
@@ -96,13 +95,24 @@ def get_token_from_request(request: WSGIRequest) -> Optional[str]:
...
-        return user
-    return None
+    user_jwt_token = payload.get("token")
+    if not user_jwt_token or not user:
+        raise InvalidTokenError(
+            "Invalid token. Create new one by using tokenCreate mutation."
+        )
+    if user.jwt_token_key != user_jwt_token:
+        raise ExpiredUserSignatureError(
+            "User requested to expire this token. Create new one by using tokenCreate "
```
maarcingebala, line 105: Isn't it the same as "token expired" error from the API-user point of view? Not sure what is the use case of this exception but I don't know the whole context probably.

maarcingebala: Unless we don't have any exception for token expiration. But still, I'm wondering if we have to say "User requested to expire this token" in the message.

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fix specific product voucher in draft orders - #5727 by @fowczarek
 - Explicit country assignment in default shipping zones - #5736 by @maarcingebala
 - Drop `json_content` field from the `Menu` model - #5761 by @maarcingebala
-- Refactor JWT support - #5734 by @korycins
+- Refactor JWT support - These changes could require a handling JWT token in the storefront. Storefront needs to handle a case when the backend returns the exception about the invalid token. - #5734, #5816 by @korycins
```
maarcingebala, line 12: We we mention this in the changelog? There are usually some changes in the backend that require changes in frontend as well.

saleor/core/jwt.py
patrys: I'm with @maarcingebala here but I'm not sure we should be treating this the same as token expired or as an invalid token. Clients will likely attempt to refresh expired tokens which will fail in this case so it's a wasted API call.

CHANGELOG.md
korycins: I thought that this will be a good idea to inform the community that some changes from the changelog will require extra actions.  A similar case was with the Vatlayer, where we deactivate the plugin but we didn't mention it explicitly. 
If you don't like this, then I will revert it.

maarcingebala: It's a good idea, I'm just thinking what is the best place to mention this. Here it may be hard to find, but maybe we should extract these changes required for the frontend when releasing a new version and add some section "Frontend breaking changes" or something.

patrys: Or just a large "breaking changes" section at the very top of the changelog? Once Cloud becomes public we will have to withhold any breaking changes until the next major release.

PR #5819, opened 2020-06-25
Performance: creating checkout

maarcingebala:
This PR improves the performance of creating checkouts:
- Optimize checking stock quantities before creating new checkout
- Use bulk inserts to create new checkout lines instead of creating them one by one.
- Optimize fields in `Checkout` type returned as a result of the `checkoutCreate` operation
- Change plugin signatures to receive data from arguments instead of making database queries

Todo:
- Optimize `is_shipping_required` functions
- Optimize `get_total_weight` functions
- Optimize completing checkout
- Fix all "fixme's" and "todo's" in code

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/plugins/base_plugin.py
```diff
@@ -120,6 +122,10 @@ def calculate_order_shipping(
     def calculate_checkout_line_total(
         self,
         checkout_line: "CheckoutLine",
+        variant: "ProductVariant",
```
korycins, line 125: The changes in this file will break compatibility with the custom plugins - as the interface is changed. 
Is it required?  How do we want to handle this case (break compatibility of the custom plugins)?

maarcingebala: Bad news - I need to change signatures of most plugins related to checkout as now they do A LOT of queries, while they should receive already prefetched data. I don't know yet how we want to handle compatibility.

ohld (approved):

saleor/graphql/checkout/mutations.py
```diff
@@ -199,23 +196,30 @@ class Meta:
...
         cls, lines, country
     ) -> Tuple[List[product_models.ProductVariant], List[int]]:
-        variant_ids = [line.get("variant_id") for line in lines]
+        variant_ids = [line["variant_id"] for line in lines]
         variants = cls.get_nodes_or_error(
             variant_ids,
             "variant_id",
             ProductVariant,
-            qs=product_models.ProductVariant.objects.prefetch_related(
+            qs=product_models.ProductVariant.objects.select_related(
```
korycins, line 207: https://github.com/mirumee/saleor/pull/6138#discussion_r490095738

saleor/graphql/checkout/mutations.py
```diff
@@ -376,7 +367,7 @@ def perform_mutation(cls, _root, info, checkout_id, lines, replace=False):
         if variants and quantities:
             for variant, quantity in zip(variants, quantities):
                 try:
-                    add_variant_to_checkout(
+                    checkout = add_variant_to_checkout(
```
korycins, line 370: IMO, no need to return and assign checkout here. Checkout is passed to function as a reference, so all changes will be visible after the function will be finished.  Second thing, it is strange that the method responsible for adding variant to checkout returns checkout

korycins (commented):
It looks really great but we have to test it hard with different cases. Adyen, Vatlayer, Avalara etc. This PR changes a lot of things related to these plugins so we need to double-check all of them

patrys:
Completing checkout got a lot slower. Can we improve that before merging?

saleor/graphql/checkout/types.py
```diff
@@ -197,90 +235,115 @@ def wrap_shipping_method_with_channel_context(data):
...
+        lines = CheckoutLinesInfoByCheckoutTokenLoader(info.context).load(root.token)
         discounts = DiscountsByDateTimeLoader(info.context).load(
             info.context.request_time
         )
-
-        return Promise.all([lines, discounts]).then(calculate_total_price)
+        return Promise.all([address, lines, discounts]).then(calculate_total_price)
 
     @staticmethod
     # TODO: We should optimize it in/after PR#5819
```
IKarbowiak, line 262: Same here. Can we drop the comment?

d-wysocki (approved):

saleor/order/utils.py
```diff
@@ -308,11 +311,39 @@ def sum_order_totals(qs, currency_code):
...
+    if not order.is_shipping_required():
+        return None
+    if not order.shipping_address:
+        return None
     return ShippingMethod.objects.applicable_shipping_methods_for_instance(
         order, channel_id=order.channel_id, price=order.get_subtotal().gross
     )
 
 
+def get_discounted_lines(lines, voucher):
```
IKarbowiak, line 323: Function with the same name is in `saleor/checkout/utils.py`. The difference I see is that the one in `checkout` module uses `CheckoutLineInfo` dataclass for lines. Do we need both of them?

saleor/plugins/manager.py
```diff
@@ -172,22 +204,32 @@ def calculate_order_shipping(self, order: "Order") -> TaxedMoney:
...
+        checkout: "Checkout",
         checkout_line: "CheckoutLine",
-        discounts: Iterable[DiscountInfo],
+        variant: "ProductVariant",
+        product: "Product",
+        collections: Iterable["Collection"],
+        address: Optional["Address"],
         channel: "Channel",
+        discounts: Iterable[DiscountInfo],
     ):
```
IKarbowiak, line 215: Can we also use `CheckoutLineInfo` here?

IKarbowiak (approved):

d-wysocki:
@IKarbowiak I will save your comments and fix them in other PR. For now we create feature branch related to checkout optimizations.

PR #5820, opened 2020-06-25
Add supported currencies to payment gateways

IKarbowiak:
- Add supported currencies to plugins configurations
- Add validation in checkoutPaymentCreate
- Match supported currencies in Checkout and Shop type

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/plugins/manager.py
```diff
@@ -320,8 +320,15 @@ def list_payment_plugin_names(self, active_only: bool = False) -> List[tuple]:
             if payment_method in type(plugin).__dict__
         ]
 
-    def list_payment_gateways(self, active_only: bool = True) -> List[dict]:
+    def list_payment_gateways(
+        self, currency: Optional[str] = None, active_only: bool = True
+    ) -> List[dict]:
         payment_plugins = self.list_payment_plugin_names(active_only=active_only)
+        # if currency is given return only gateways which support given currency
+        if currency:
```
korycins, line 328: Don't you think, that it would be better to append currencies to the response of the `list_payment_gateways`?
The `list_payment_plugin_names`  is used only in `list_payment_gateways`. Same as `__get_payment_config`. You can modify them to operate directly on the `plugin` not on the `name` - with `name` you need to fetch plugin every time when you want to get plugin data
Something like this:
```
    def get_payment_plugins(self, active_only: bool = False) -> Dict[str, "BasePlugin"]:
        payment_method = "process_payment"
        plugins = self.plugins
        if active_only:
            plugins = self.get_active_plugins()
        return {
            plugin.PLUGIN_ID: plugin
            for plugin in plugins
            if payment_method in type(plugin).__dict__
        }

    def list_payment_gateways(self, currency: None, active_only: bool = True) -> List[dict]:
        payment_plugins = self.get_payment_plugins(active_only=active_only)
        gateways = [
            {
                "id": plugin_id,
                "name": plugin.PLUGIN_NAME,
                "config": self.__get_payment_config(plugin),
                "currencies": self.__get_payment_currencies(plugin)
            }
            for plugin_id, plugin in payment_plugins
        ]
        if currency:
            return [gtw for gtw in gateways if currency in gtw["currencies"]]
        return gateways

    def __get_payment_currencies(self, gateway: "BasePlugin"):
        method_name = "get_payment_config"
        default_value: list = []
        config = self.__run_method_on_single_plugin(gateway, method_name, default_value)
        config_dict = {item["field"]:item['value'] for item in config}
        return config_dict.get('supported_currencies')


    def __get_payment_config(self, gateway: "BasePlugin") -> List[dict]:
        method_name = "get_payment_config"
        default_value: list = []
        return self.__run_method_on_single_plugin(gateway, method_name, default_value)
```

saleor/graphql/shop/types.py
```diff
@@ -66,6 +68,11 @@ class Meta:
 class Shop(graphene.ObjectType):
     available_payment_gateways = graphene.List(
         graphene.NonNull(PaymentGateway),
+        currency=graphene.Argument(
```
korycins, line 71: Not sure of this.
- How storefront will fetch what currencies are available for the shop?
- IMO `PaymentGateway` type should also have a field `currencies` or `supported_currencies`. - It would be easier for storefronts with multiple currencies support.

maarcingebala: 1) There is `shop { currencies }` query. Not sure it will be used in multichannel as well but I'm sure we'll return some channel-specific info so there should be some way for storefront to know the currencies it operates in.
2) Agree that it could be useful, we can discuss with the frontend team if they needed it or look on designs.

saleor/graphql/payment/mutations.py
```diff
@@ -105,6 +106,21 @@ def clean_payment_amount(cls, info, checkout_total, amount):
                 }
             )
 
+    @classmethod
+    def validate_gateway(cls, gateway_id, currency):
+        available_gateways = get_plugins_manager().list_payment_gateways(
+            currency=currency
+        )
+        if not any([gateway["id"] == gateway_id for gateway in available_gateways]):
```
maarcingebala, line 114: Not sure if it's needed anywhere else but this could be also a utility function in the payment module, e.g. `is_supported_currency(gateway, currency)`.

saleor/payment/gateways/utils.py
```diff
@@ -0,0 +1,22 @@
+import warnings
+from typing import TYPE_CHECKING, List
+
+from django.conf import settings
+
+if TYPE_CHECKING:
+    from ..interface import GatewayConfig
+
+
+def get_supported_currencies(config: "GatewayConfig", gateway_name: str) -> List[str]:
```
maarcingebala, line 10: Docstring would be useful here since this is a utility function that can be reused in more places.

saleor/graphql/shop/types.py
korycins: Remember that someone can use one storefront for multiple currencies. So single storefront- multiple channels. I don't think that extending PaymentGateway would cost us a lot of work, it would be an easier way to fetch all gateways (by storefront) and combine them with currencies.

saleor/payment/gateways/razorpay/tests/test_razorpay.py
```diff
@@ -27,6 +27,7 @@ def gateway_config():
     return GatewayConfig(
         gateway_name="razorpay",
         auto_capture=False,
+        supported_currencies="USD",
```
NyanKiyoshi, line 30: Do we check whether the provider supports to passed currency? `INR` is the only supported currency for RazorPay

IKarbowiak: We decided to make supported currencies only configurable on the dashboard right now. There are plans to fetch supported currencies from gateway.

saleor/plugins/manager.py
patrys: I have a feeling that `if "process_payment" in type(plugin).__dict__` should really be an [abstract base class](https://docs.python.org/3/library/abc.html#abc.ABCMeta.__subclasshook__) so you could just use `isinstance(plugin, PaymentPlugin)`.

saleor/payment/gateways/razorpay/tests/test_razorpay.py
patrys: Is there an easy way to use INR in this test case?

saleor/graphql/shop/types.py
patrys: Shouldn't available payment gateways live under Checkout rather than Shop?

IKarbowiak: We have it in both places.

saleor/payment/gateways/razorpay/tests/test_razorpay.py
IKarbowiak: I will replace here `supported_currencies` with `INR`.

saleor/graphql/shop/types.py
maarcingebala: We used to have it only in the `Checkout` type but the storefront team asked to add it here as well. @orzechdev Do you remember why we need `availablePaymentGateways` in the `Shop` type?

saleor/plugins/manager.py
IKarbowiak: We decided that the plugins need a bigger refactor, and we will leave it as it is for now.

PR #5836, opened 2020-06-30
Fix SimpleLazyObject in payload generator. Improve calls for quantity-changed event

### LINKED ISSUE #5830: Checkout process doesn't go ahead from "Address"
…
Purchase a product via checkout.

1. Go to store-front
2. Login with your credentials
3. Select category from home page
4. Select product.
5. Select product variant and quantity.
6. Click on add to basket
7. Open basket and checkout
8. Link is: "localhost:3000/checkout/address"
9. Enter new address or select existing address and click continue to shipping.
10. Nothing at all happens. In developer tools/network tab the api hits with status 200 with the following error: 
"errors": [{"message": "'User' object is not iterable", "locations": [{"line": 170, "column": 3}], "path": ["checkoutCreate"], 

### What I expected to happen
Move to shipping 

### Screenshots

image

**System information**
Operating system: N/A
Browser: Chrome (83.0)

korycins:
closes #5830

### DIFF
--- a/saleor/graphql/checkout/mutations.py
+++ b/saleor/graphql/checkout/mutations.py
@@ -287,6 +287,7 @@ class CheckoutCreate(ModelMutation, I18nMixin):
                         "Can't create checkout with unpublished product.",
                         code=exc.code,
                     )
+            info.context.plugins.checkout_quantity_changed(instance)
         # Save provided addresses and associate them to the checkout
         cls.save_addresses(instance, cleaned_input)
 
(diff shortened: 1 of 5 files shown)


saleor/graphql/checkout/mutations.py
```diff
@@ -287,6 +287,7 @@ def save(cls, info, instance: models.Checkout, cleaned_input):
                         "Can't create checkout with unpublished product.",
                         code=exc.code,
                     )
+            info.context.plugins.checkout_quantity_changed(instance)
```
maarcingebala, line 290: Let's discuss if this is needed here - we create a new checkout instance, so should we call a "quantity changed" hook? If anything, I would expect "checkout created" hook, if we had it.

maarcingebala: It will be removed in separate PR.

PR #5837, opened 2020-06-30
Drop CHECKOUT_QUANTITY_CHANGED, Add CHECKOUT_CREATED, CHECKOUT_UPDATED

korycins:
- Deprecate - WebhookEventType.CHECKOUT_QUANTITY_CHANGED
- WebhookEventType.CHECKOUT_QUANTITY_CHANGED will be removed in Saleor 3.0
- Add CHECKOUT_CREATED
- Add CHECKOUT_UPDATED


saleor/core/permissions.py
```diff
@@ -35,6 +35,7 @@ class MenuPermissions(BasePermissionEnum):
     MANAGE_MENUS = "menu.manage_menus"
 
 
+# Depreacted. It will be removed in Saleor 3.0
```
patrys, line 38: Typo: "depreacted" -> "deprecated"

saleor/graphql/webhook/mutations.py
```diff
@@ -13,7 +13,11 @@ class WebhookCreateInput(graphene.InputObjectType):
     name = graphene.String(description="The name of the webhook.", required=False)
     target_url = graphene.String(description="The url to receive the payload.")
     events = graphene.List(
-        WebhookEventTypeEnum, description="The events that webhook wants to subscribe."
+        WebhookEventTypeEnum,
+        description=(
+            "The events that webhook wants to subscribe. The CHECKOUT_QUANTITY_CHANGED"
+            " is depreacted. It will be removed in Saleor 3.0"
```
patrys, line 19: Can we recommend which hook to use instead?

korycins: @patrys IMHO -  handling events for `checkout quantity changed` should be done on the storefront(with JS).

patrys: What about abandoned cart handling?

korycins: The current workflow for creating Checkout in Saleor requires to move at least to `ShippingMethod` section. This is the moment when Storfreont trigger CreateCheckout mutation. Saleor creates a checkout object and assigns it to user (if authenticated). If we want to handle all abandoned carts we need to change workflow to intercept all carts for authenticated users before they moved to the 2nd step of checkout process. I am not sure what we should do for non-authenticated users - do we want to handle data from the email input?
Even if we decide to handle the cart by Saleor then we can introduce CheckoutCreate webhook event.
@patrys What do you think?

patrys: A company may want to create a storefront that creates a checkout object as soon as I know the user's email address. It doesn't really matter if our default storefront works differently. We may want to give them a webhook that is called when a checkout object is updated, we may also tell them that the right way to approach this is to periodically poll for data from Saleor (potential downside is the number of checkout objects could be large).

korycins: Okay, so what do you propose? Should we create two new webhooks instead of this one? `checkout_created` and `checkout_updated` ?

patrys: I don't have a solution ready, I'm thinking about the implications of this getting deprecated. See my answer above, telling users to poll for abandoned carts is also an option.

korycins: I would say that the better way would be adding a new webhooks like `checkout_created` and `checkout_updated` and deprecating this one. I can do this in this PR (will change the name of pull request). Do you agree? @patrys @maarcingebala

patrys: Sounds like a plan 😄

korycins: 👨‍🏭

PR #5860, opened 2020-07-09
Fix invoice PDF number

tomaszszymanski129:
- when we can have no user on email object, we should use the `get_customer_email` method, otherwise it will fail
- it fixes invoice number not showing in resulting PDF file
- it fixes invoice number incrementation logic

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/invoice/mutations.py
+++ b/saleor/graphql/invoice/mutations.py
@@ -19,7 +19,7 @@ class InvoiceRequest(ModelMutation):
     class Meta:
         description = "Request an invoice for the order using plugin."
         model = models.Invoice
-        # permissions = (OrderPermissions.MANAGE_ORDERS,)
+        permissions = (OrderPermissions.MANAGE_ORDERS,)
         error_type_class = InvoiceError
         error_type_field = "invoice_errors"
 
--- a/saleor/graphql/invoice/mutations.py
+++ b/saleor/graphql/invoice/mutations.py
@@ -19,7 +19,7 @@ class InvoiceRequest(ModelMutation):
     class Meta:
         description = "Request an invoice for the order using plugin."
         model = models.Invoice
-        permissions = (OrderPermissions.MANAGE_ORDERS,)
+        # permissions = (OrderPermissions.MANAGE_ORDERS,)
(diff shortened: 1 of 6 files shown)


saleor/graphql/invoice/mutations.py
```diff
@@ -19,7 +19,7 @@ class InvoiceRequest(ModelMutation):
     class Meta:
         description = "Request an invoice for the order using plugin."
         model = models.Invoice
-        permissions = (OrderPermissions.MANAGE_ORDERS,)
+        # permissions = (OrderPermissions.MANAGE_ORDERS,)
```
maarcingebala: What about this line?

saleor/plugins/invoicing/utils.py
```diff
@@ -14,24 +14,31 @@
...
-    return int(number)
+
+def parse_invoice_dates(invoice):
+    match = re.match(r"^(\d+)\/(\d+)\/(\d+)", invoice.number)
+    return int(match.group(1)), int(match.group(2)), int(match.group(3))
 
 
 def generate_invoice_number():
-    last_invoice = Invoice.objects.last()
+    last_invoice = Invoice.objects.filter(number__isnull=False).last()
```
maarcingebala, line 35: Why and when we may have invoices without a number?

tomaszszymanski129: Point of the invoice plugin is to provide it with empty invoice object for manipulation. That object is already created but doesn't have a number yet. This is why we want to fetch the last invoice that did had the number. When the plugin finish its job (this `generate_invoice_number` is actually called in plugin), the number will be there.

saleor/graphql/invoice/mutations.py
tomaszszymanski129: reverted it very quickly after push but you were fast enough to spot it :police_officer:

saleor/plugins/invoicing/utils.py
maarcingebala: 👍

PR #5818, opened 2020-06-24
Add product availability in channel

fowczarek:
- [x] Create a default channel in mutation if exists any Product or Order or Sale or Voucher or Checkout.
- [x] Create a new model for `ProductChannelListing`
- [x] Migrate publication data from `Product` to `ProductChannelListing`
- [x] Drop publication field from `Product`
- [x] Refactor `Product` querysets
- [x] Rafactor all availability usages in code not related to graphql 
- [x] Add protected queries for getting all `ProductChannelListing` for product
- [x] Refactor existing Product queries
- [x] Add warrining if query don't have `ChannelId` in request
- [x] Add mutation to manage `ProductChannelListing` 
- [x] Update fixtures in tests
- [ ] Update populate db - Separete PR 

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #5883, opened 2020-07-14
Add channel slug to checkoutCreate

d-wysocki:
I want to merge this change because...I want to add channelSlug to input in checkoutCreate mutation.

# Impact

* [X] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/channel/utils.py
```diff
@@ -0,0 +1,9 @@
+from .models import Channel
+
+
+def raise_channel_validation():
```
fowczarek, line 4: When we merge #5818 you should use `get_default_channel_slug_if_available` for getting default channel.

saleor/order/models.py
```diff
@@ -99,7 +100,13 @@ class Order(ModelWithMetadata):
     shipping_method_name = models.CharField(
         max_length=255, null=True, default=None, blank=True, editable=False
     )
-
+    channel = models.ForeignKey(
+        Channel,
+        related_name="orders",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
```
fowczarek, line 108: We shouldn't allow orders without channels. When someone remove a channel all related orders should be moved to new target channel .

saleor/tests/fixtures.py
```diff
@@ -227,8 +227,8 @@ def site_settings(db, settings) -> SiteSettings:
 
 
 @pytest.fixture
-def checkout(db):
-    checkout = Checkout.objects.create(currency="USD")
+def checkout(db, channel):
+    checkout = Checkout.objects.create(currency="USD", channel=channel)
```
fowczarek, line 231:
```suggestion
    checkout = Checkout.objects.create(currency=channel.currency_code, channel=channel)
```

saleor/order/models.py
fowczarek: Removing channel is a separate task.

saleor/graphql/checkout/mutations.py
```diff
@@ -142,6 +144,7 @@ class CheckoutLineInput(graphene.InputObjectType):
 
 
 class CheckoutCreateInput(graphene.InputObjectType):
+    channel_slug = graphene.String(description="Channel slug for current checkout.")
```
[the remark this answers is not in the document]
d-wysocki: It's need for deprecation.

saleor/checkout/migrations/0027_checkout_channel.py
```diff
@@ -0,0 +1,52 @@
...
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def add_channel_slug(apps, schema_editor):
+    Channel = apps.get_model("channel", "Channel")
+    Checkout = apps.get_model("checkout", "Checkout")
+
+    if Checkout.objects.exists():
```
maarcingebala, line 11: Why do we need this? The iterator below won't run without if there are no checkouts.

saleor/graphql/checkout/mutations.py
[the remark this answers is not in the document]
maarcingebala: In Owczar's PR we changed "channelSlug" in the schema to "channel". Slug is an implementation detail here and doesn't need to be exposed in the schema. I would change this field name to "channel". Also, a slight improvement of the description - "Slug of a channel in which to create a checkout"

saleor/graphql/checkout/mutations.py
```diff
@@ -315,8 +335,8 @@ def perform_mutation(cls, _root, info, **data):
             checkout = models.Checkout(user=user)
         else:
             checkout = models.Checkout()
-
         cleaned_input = cls.clean_input(info, checkout, data.get("input"))
+        checkout.channel = cleaned_input["channel"]
```
maarcingebala, line 339: I think this should happen in `clean_input`. We either use the value passed in the input or use the default from `info.context`.

saleor/graphql/checkout/mutations.py
```diff
@@ -351,6 +371,7 @@ def perform_mutation(cls, _root, info, checkout_id, lines, replace=False):
         )
 
         variant_ids = [line.get("variant_id") for line in lines]
+        info.context.channel_slug = checkout.channel.slug
```
maarcingebala, line 374: Why do we need this? This value comes from middleware and IMO it should be always set by the middleware, otherwise we'll end up with magic which is really difficult to debug.

saleor/graphql/checkout/schema.py
```diff
@@ -33,6 +33,9 @@ class CheckoutQueries(graphene.ObjectType):
         Checkout,
         description="Look up a checkout by token.",
         token=graphene.Argument(UUID, description="The checkout's token."),
+        channel_slug=graphene.Argument(
```
maarcingebala, line 36: "channel"

saleor/graphql/checkout/types.py
```diff
@@ -116,6 +116,9 @@ class Checkout(CountableDjangoObjectType):
         description="The price of the checkout before shipping, with taxes included.",
     )
     token = graphene.Field(UUID, description=("The checkout's token."), required=True)
+    channel_slug = graphene.String(
+        description=("The checkout's channel slug."), required=True
+    )
```
maarcingebala, line 121: There is `channel` added to `only_fields` below. Why do we need `channel_slug` here?

saleor/order/migrations/0086_order_channel.py
```diff
@@ -0,0 +1,53 @@
...
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def add_channel_slug(apps, schema_editor):
+    Channel = apps.get_model("channel", "Channel")
+    Order = apps.get_model("order", "Order")
+
+    if Order.objects.exists():
```
maarcingebala, line 11: We don't need this.

saleor/graphql/middleware.py
```diff
@@ -37,8 +37,14 @@ def user():
 class ChannelMiddleware:
     def resolve(self, next, root, info, **kwargs):
         request = info.context
-
-        channel_slug = kwargs.get("channel")
+        request_input = kwargs.get("input", {})
```
maarcingebala, line 40: This is magic. Some mutations will have `input` param, some others will not. Second thing - I'm afraid it will be really hard to guess what value is actually stored in the request if you're using it somewhere in the code. We need to discuss why we need these changes and if can be done in some more explicit and easy to follow way.

saleor/checkout/error_codes.py
```diff
@@ -21,3 +21,4 @@ class CheckoutErrorCode(Enum):
     UNIQUE = "unique"
     VOUCHER_NOT_APPLICABLE = "voucher_not_applicable"
     ZERO_QUANTITY = "zero_quantity"
+    MISSING_CHANNEL_SLUG = "missing_channel_slug"
```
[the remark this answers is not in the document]
d-wysocki: No, that error code left from other solution, I will remove this.

saleor/graphql/middleware.py
fowczarek: Maby if these changes are required only for CreateCheckout mutation we shouldn't put in inside mutation?

fowczarek:
You should add tests for all checkout line related mutation like `checkoutLinesAdd` and `checkoutLinesUpdate`. 
You should check adding product witch is not available in channel related to checkout but is available in a different channel.

saleor/graphql/checkout/schema.py
```diff
@@ -29,10 +33,11 @@
 
 
 class CheckoutQueries(graphene.ObjectType):
-    checkout = graphene.Field(
+    checkout = FieldWithChannel(
         Checkout,
-        description="Look up a checkout by token.",
+        description="Look up a checkout by token and slug of channel.",
         token=graphene.Argument(UUID, description="The checkout's token."),
+        slug=graphene.Argument(graphene.String, description="Slug of the checkout"),
```
maarcingebala, line 40: Wrong description.

saleor/graphql/middleware.py
```diff
@@ -37,8 +37,14 @@ def user():
...
+        channel = kwargs.get("channel")
+        if isinstance(request_input, dict) and request_input:
+            channel_slug = request_input.get("channel")
+        elif channel is not None:
+            channel_slug = channel
+        else:
+            channel_slug = None
```
maarcingebala, line 47: Do we need this in this PR? I'm removing the middleware at all on my branch so this will be removed anyway.

saleor/graphql/checkout/schema.py
maarcingebala: We now have `FieldWithChannel` which adds `channel` arg and the `slug` arg which seems like not needed?

PR #5884, opened 2020-07-14
Add endpoint handler for plugins

korycins:
- Add webhook handler which can be used by plugins.

Example:
 Request to  `/plugins/vatlayer/some/path` will call `VatlayerPlugin.webhook` method

`VatlayerPlugin` will receive the `request` WSGIRequest object and internal plugin path `/some/path`.
This is super useful when we need to have the possibility to communicate with external service. 

it closes https://github.com/mirumee/saleor/issues/4955

### DIFF
--- a/saleor/plugins/tests/test_manager.py
+++ b/saleor/plugins/tests/test_manager.py
@@ -1,6 +1,8 @@
+import json
 from decimal import Decimal
 
 import pytest
+from django.http import HttpResponseNotFound, JsonResponse
 from django_countries.fields import Country
 from prices import Money, TaxedMoney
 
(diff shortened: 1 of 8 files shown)


saleor/plugins/tests/test_manager.py
```diff
@@ -340,3 +342,44 @@ def test_manager_serve_list_all_payment_gateways_specified_currency_two_gateways
...
+    assert response.status_code == 404
+
+
+def test_manager_inncorrect_plugin(rf):
+    plugins = [
+        "saleor.plugins.tests.sample_plugins.PluginSample",
+        "saleor.plugins.tests.sample_plugins.PluginInactive",
+    ]
+    manager = PluginsManager(plugins=plugins)
+    plugin_path = "/webhook/paid"
```
IKarbowiak, line 381: Do you have test for `plugin_path="/webhook/failed"`?

korycins: No need as I added failed only to show how to handle different paths in the plugin.
Here, I just want to confirm that we call all the required actions for the manager. I don't want to test the methods of  `SamplePlugin` ;)

saleor/urls.py
```diff
@@ -17,6 +18,11 @@
         digital_product,
         name="digital-product",
     ),
+    url(
```
auvipy, line 21: why not path() or re_path here?

korycins: to keep consistency with the above patterns ☝️ . I know that `url()` will be deprecated in the future but I think that migrating to `path` or `re_path` should be done in one and separate PR

PR #5885, opened 2020-07-14
variants sorting by attributes and attributes values specified in product type

Dom4n:
We implemented cool feature, which allows sorting variants based on attribute and attribute values order provided in product type.

The solution is really lightweight because I have added sort field to ProductVariant model - simple IntegerField.
Sorting is calculated asynchronously when someone changes attribute, attribute value or creates new ProductVariant.
I thought about using SortableModel but it does not fit in this use case. 

The solution is working flawlessly in our shop on production.

Visual representation:

sorted attribute values:
obraz

storefront:
obraz

and now I moved 80 behind 86:
obraz

It got updated right away in storefront:
obraz

# Impact

* [X] New migrations

# Pull Request Checklist

* [X] The changes are tested


korycins:
@Dom4n  Thanks for the PR, it seems like quite nice work but tests are failing. Can you take a look at that? Also, it would be good if you could attach some new tests for your logic.

Dom4n:
Yes, I will refactor it during weekend probably and add a few tests.

PR #5887, opened 2020-07-15
Use generic scalara instead of string in PaymentGateway config

korycins:
This PR changes the type of `GatewayConfigLine` to `GenericScalar`.
Some payment gateway can require more advanced structure and with the current type, we are not able to pass them to a storefront.


fowczarek:
@korycins Maby we should add test for some advanced structure in the `value` field?

korycins:
> @korycins Maby we should add test for some advanced structure in the `value` field?

@fowczarek Yeap, you are right. I've added test.

korycins:
Closing it as we discussed that we will just dump the structure and send it as a str

PR #5889, opened 2020-07-15
Show payment gateway in checkout based on checkout object

korycins:
Some Payment plugins require the data from the checkout object. 
Sometimes payment plugin needs to use something from meta of checkout. This PR changes a way of handling the available payment methods

### DIFF
--- a/saleor/graphql/checkout/types.py
+++ b/saleor/graphql/checkout/types.py
@@ -244,7 +244,7 @@ class Checkout(CountableDjangoObjectType):
 
     @staticmethod
     def resolve_available_payment_gateways(root: models.Checkout, _info):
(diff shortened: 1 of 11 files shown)


saleor/graphql/checkout/types.py
```diff
@@ -21,7 +22,7 @@
 
 class GatewayConfigLine(graphene.ObjectType):
     field = graphene.String(required=True, description="Gateway config key.")
-    value = graphene.String(description="Gateway config value for key.")
+    value = GenericScalar(description="Gateway config value for key.")
```
maarcingebala, line 25: I would only consult with the frontend team if they can handle such fields easily in their code, when you don't know what exactly will be returned from a field.

PR #5893, opened 2020-07-17
Return weights in default weight unit

### LINKED ISSUE #5874: Product type shipping weight gets saved in kilograms when set to pounds
Shipping weight unit is set to POUNDS and i am trying to configure weight on product types in pounds. When saved and refreshed it gets converted to kilograms.

1. Configuration -> Shipping Methods -> Set shipping weight unit to LB
2. Go to product type that is shippable and add weight in pounds
3. Refresh page

### What I expected to happen
I expect the weight to be shown and used in pounds when set to pounds.

### Screenshots

Step 1
shipping-weight

Step 2
product-type-weight-unsaved

Step 3
product-type-weight-saved

### LINKED ISSUE #5875: Shipping rate is converted to KG when set to different shipping weight units
I want to set the shipping weight unit to pounds and be able to configure the weight based rates in pounds.

1. Configuration -> Shipping Methods -> Set shipping weight unit to LB
2.  Create a shipping zone
3. Create a weight based rate using pounds as unit type

### What I expected to happen
I expected the weight range to show up in pounds.

### Screenshots
shipping-weight

edit-weight

weight-based-rate

IKarbowiak:
Weights should be returned in shop default unit weights.
Fixes #5874
Fixes #5875 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

image
image
image

### DIFF
--- a/saleor/core/weight.py
+++ b/saleor/core/weight.py
@@ -53,4 +53,6 @@ def covert_weight_to_default_weight_unit(weight):
     default_unit = get_default_weight_unit()
     if weight and weight.unit != default_unit:
         weight = convert_weight(weight, default_unit)
+    elif weight:
+        weight.value = round(weight.value, 3)
     return weight
--- a/saleor/core/weight.py
+++ b/saleor/core/weight.py
@@ -39,9 +39,18 @@ def convert_weight(weight, unit):
     # Weight amount from the Weight instance can be retrived in serveral units
     # via its properties. eg. Weight(lb=10).kg
     converted_weight = getattr(weight, unit)
-    return Weight(**{unit: converted_weight})
+    weight = Weight(**{unit: converted_weight})
+    weight.value = round(weight.value, 3)
+    return weight
 
 
 def get_default_weight_unit():
     site = Site.objects.get_current()
     return site.settings.default_weight_unit
+
+
+def covert_weight_to_default_weight_unit(weight):
+    default_unit = get_default_weight_unit()
+    if weight and weight.unit != default_unit:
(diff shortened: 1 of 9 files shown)


saleor/core/weight.py
```diff
@@ -35,13 +35,26 @@ def zero_weight():
     return Weight(kg=0)
 
 
-def convert_weight(weight, unit):
+def convert_weight(weight: Weight, unit: str) -> Weight:
```
fowczarek, line 38: Maby better to use `WeightUnits` instead of `str` as `unit` type?

saleor/core/weight.py
```diff
@@ -35,13 +35,26 @@ def zero_weight():
...
+    weight.value = round(weight.value, 3)
+    return weight
 
 
 def get_default_weight_unit():
     site = Site.objects.get_current()
     return site.settings.default_weight_unit
+
+
+def covert_weight_to_default_weight_unit(weight: Weight) -> Weight:
```
maarcingebala, line 52: Typo: covert :)

saleor/core/weight.py
maarcingebala: Would be nice if we could use enum here.

maarcingebala: Also let's add a docstring here, even very simple that it just coverts weight to given unit.

IKarbowiak: There is a short description in a comment. I can covert it to the docstring

IKarbowiak: Must be `str` because `getattr` require `str`.

PR #5901, opened 2020-07-21
Add product variant in channel

fowczarek:
I want to merge this change because I move variant price data to variant-channel relation.
TODO: 
- [x] drop price field from ProductVariant mutation
- [x] drop price field from ProductVariant type
- [x] add `channel_listing` field to `ProductVaraitnt` similar field is in `Product` type.
- [x] resolver for `channel_listing` should be implemented with dataloader similar to `ProductChannelListingByProductIdLoader`
- [x] add `updateProductVariantChannelListing` mutation for creating pricing info per channels. (Similar solution to `ProductChannelListingUpdate` user should be able to set price only for available products and when product is removed from channel all prices for this channel should be dropped.) 
- [ ] ~~consider what to do with fields like `margin` because this field depends on `cost_price` and `price`. `Cost_price` have unchanged currency. Probably we should temporarily remove these fields.~~ will be done separate task
- [x] make sure that any of ProductVariant model function don't use removed field
- [x] Any test shouldn't raise uncached warning from channel middleware
- [x] Fix mypy errors

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


IKarbowiak (approved):

saleor/product/tasks.py
```diff
@@ -43,8 +42,11 @@ def update_variants_names(product_type_pk: int, saved_attributes_ids: List[int])
 
 @app.task
 def update_product_minimal_variant_price_task(product_pk: int):
-    product = Product.objects.get(pk=product_pk)
-    update_product_minimal_variant_price(product)
+    # TODO: We should refactor this in separete PR
+    # https://app.clickup.com/t/6a5txz
+    return
+    # product = Product.objects.get(pk=product_pk)
```
d-wysocki, line 48: This comment is necessary?

saleor/product/tasks.py
```diff
@@ -4,8 +4,7 @@
 from ..discount.models import Sale
 from .models import Attribute, Product, ProductType, ProductVariant
 from .utils.attributes import generate_name_for_variant
-from .utils.variant_prices import (
-    update_product_minimal_variant_price,
+from .utils.variant_prices import (  # update_product_minimal_variant_price,
```
d-wysocki, line 7: This comment is necessary?

saleor/checkout/calculations.py
```diff
@@ -73,13 +74,16 @@ def checkout_total(
 
 
 def checkout_line_total(
-    *, line: "CheckoutLine", discounts: Optional[Iterable[DiscountInfo]] = None
+    *,
+    line: "CheckoutLine",
+    discounts: Optional[Iterable[DiscountInfo]] = None,
+    channel: "Channel"
```
d-wysocki, line 80: What do you think do pass here `channel_id` instead of channel object?

saleor/plugins/base_plugin.py
```diff
@@ -132,10 +133,12 @@ def calculate_order_shipping(
         """
         return NotImplemented
 
+    # TODO: Add information about this change to `breaking changes in changelog`
     def calculate_checkout_line_total(
         self,
         checkout_line: "CheckoutLine",
         discounts: List["DiscountInfo"],
+        channel: "Channel",
```
d-wysocki, line 141: What do you thing to pass here `channel_id` instead of channel object?

saleor/plugins/manager.py
```diff
@@ -154,13 +156,20 @@ def calculate_order_shipping(self, order: "Order") -> TaxedMoney:
         )
 
     def calculate_checkout_line_total(
-        self, checkout_line: "CheckoutLine", discounts: Iterable[DiscountInfo]
+        self,
+        checkout_line: "CheckoutLine",
+        discounts: Iterable[DiscountInfo],
+        channel: "Channel",
```
d-wysocki, line 162: What do you thing to pass here `channel_id` instead of channel object?

saleor/plugins/base_plugin.py
fowczarek: In this case, we can't do it because we pass the channel to the plugin. Some plugin might to require channel slug and others can require channel id.

saleor/checkout/calculations.py
fowczarek: In this case, we can't do it because we pass the channel to the plugin. Some plugin might to require channel slug and others can require channel id.

saleor/product/tasks.py
fowczarek: This code should be refactored in next PR's I prefer left it commented instead of removing it. It is easier to look at how this code looks before my changes if the old code is commented instead of looking for it in repository/history.

saleor/product/tasks.py
fowczarek: This code should be refactored in next PR's I prefer left it commented instead of removing it. It is easier to look at how this code looks before my changes if the old code is commented instead of looking for it in repository/history.

d-wysocki (approved):

PR #5933, opened 2020-07-30
Add channel id to draftOrder

d-wysocki:
I want to merge this change because...I want to add channel id to draftORder mutations

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/core/types/common.py
+++ b/saleor/graphql/core/types/common.py
@@ -142,10 +142,7 @@ class OrderError(Error):
     )
     variants = graphene.List(
         graphene.NonNull(graphene.ID),
-        description=(
-            "List of product variants product variant that are not published "
-            "in the channel associated with this order."
-        ),
+        description="List of product variants that are associated with the error",
         required=False,
     )
 
--- a/saleor/graphql/core/types/common.py
+++ b/saleor/graphql/core/types/common.py
@@ -140,6 +140,14 @@ class OrderError(Error):
     order_line = graphene.ID(
         description="Order line ID which causes the error.", required=False,
     )
+    variants = graphene.List(
+        graphene.NonNull(graphene.ID),
+        description=(
+            "List of product variants product variant that are not published "
+            "in the channel associated with this order."
+        ),
+        required=False,
+    )
 
 
 class InvoiceError(Error):
--- a/saleor/graphql/order/mutations/draft_orders.py
+++ b/saleor/graphql/order/mutations/draft_orders.py
@@ -57,6 +57,9 @@ class DraftOrderInput(InputObjectType):
     customer_note = graphene.String(
         description="A note from a customer. Visible by customers in the order summary."
     )
+    channel = graphene.ID(
+        description="ID of the channel associated with the order.", name="channel"
+    )
 
 
 class DraftOrderCreateInput(DraftOrderInput):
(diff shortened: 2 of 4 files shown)


saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -81,16 +87,59 @@ class Meta:
...
+                            "Can't add unpublished products.",
+                            code=OrderErrorCode.PRODUCT_NOT_PUBLISHED,
+                        )
+                    }
+                )
+
+    @classmethod
+    def clean_channel_id(cls, instance, channel):
+        if channel:
+            if instance.channel is not None:
```
IKarbowiak, line 118:
```suggestion
if channel and instance.channel is not None:
```

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -81,16 +87,59 @@ class Meta:
...
+            unpublished_products = ProductChannelListing.objects.filter(
+                Q(product__variants__id__in=variant_ids),
+                Q(channel__id=channel),
+                Q(is_published=False),
+            )
+            if unpublished_products:
+                raise ValidationError(
+                    {
+                        "lines": ValidationError(
+                            "Can't add unpublished products.",
```
maarcingebala, line 109: I would put it more explicitly. We need to make sure that our error messages precisely inform the user what happened: "Can't add products that are not published in the channel associated with this order."

maarcingebala: It would be really helpful if we returned IDs of variant IDs from the input that cannot be added. We can extend the `OrderError` type with `variants: [ID]` field and pass the IDs using `params` argument that is available in our validation errors.

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -81,16 +87,59 @@ class Meta:
         error_type_class = OrderError
         error_type_field = "order_errors"
 
+    @classmethod
+    def validate_product_is_published_in_channel(
+        cls, info, instance, variants, channel_id
+    ):
+        channel = getattr(instance.channel, "id", None)
```
maarcingebala, line 94: I think we could just do `channel = instance.channel_id`.

saleor/graphql/core/types/common.py
```diff
@@ -140,6 +140,14 @@ class OrderError(Error):
     order_line = graphene.ID(
         description="Order line ID which causes the error.", required=False,
     )
+    variants = graphene.List(
+        graphene.NonNull(graphene.ID),
+        description=(
+            "List of product variants product variant that are not published "
```
maarcingebala, line 146: This field can be used in all order mutations so I would say "List of product variants that are associated with the error".

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -81,16 +87,60 @@ class Meta:
...
+                Q(is_published=False),
+            ).values_list("product__variants__id", flat=True)
+            if unpublished_variants:
+                raise ValidationError(
+                    {
+                        "lines": ValidationError(
+                            "Can't add product variant that are not published in "
+                            "the channel associated with this order.",
+                            code=OrderErrorCode.PRODUCT_NOT_PUBLISHED,
+                            params={"variants": unpublished_variants},
```
maarcingebala, line 112: This should be a list of GraphQL IDs (base64 encoded). Otherwise (when returning database ID) client won't know how to handle them (in Apollo cache we identify objects by GraphQL ID).

PR #5937, opened 2020-07-31
Add adyen support for 3D secure payments

IKarbowiak:
Handle additional actions for 3D secure payments for adyen.
Implement step 4 and 5: https://docs.adyen.com/checkout/drop-in-web?tab=codeBlockoDBM2_3#step-4-additional-front-end
It must be checked with storefront.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/payment/gateways/adyen/plugin.py
```diff
@@ -242,16 +247,37 @@ def process_payment(
...
+        if action:
+            _type, payment_id = from_global_id(payment_information.payment_id)
+            try:
+                payment = Payment.objects.get(pk=payment_id)
+                details = result.message.get("details")
+                payment.extra_data = json.dumps(
+                    {
+                        "payment_data": action["paymentData"],
+                        "parameters": [detail["key"] for detail in details]
+                        if details
```
korycins, line 262: This `if else` will not be needed if you assign `list` in `get` call:
`details = result.message.get("details", [])`

saleor/payment/gateways/adyen/webhooks.py
```diff
@@ -468,8 +470,14 @@ def validate_auth_user(headers: HttpHeaders, gateway_config: "GatewayConfig") ->
...
 
-def handle_webhook(request: WSGIRequest, gateway_config: "GatewayConfig"):
-    json_data = json.loads(request.body)
+def handle_webhook(
+    request: WSGIRequest, gateway_config: "GatewayConfig", payment_details: Callable
+):
+    try:
+        json_data = json.loads(request.body)
+    except ValueError:
+        return handle_additional_actions(request, payment_details)
```
korycins, line 479: Not sure of this. 
Inside AdyenGatewayPlugin.webhook you can create simple `if` for separating webhook notification and additional actions. 
You have a request object so you can use request.path to define if this request should be treated as webhook notification or as a payment additional action. 
```
def webhook(request):
    if reguest.path == WEBHOOK_PATH:
        return handle_webhook()
    elif request.path == ADDITIONAL_PATH:
       return handle_additional_path()
```

saleor/payment/gateways/adyen/plugin.py
```diff
@@ -242,16 +247,37 @@ def process_payment(
             kind = TransactionKind.CAPTURE
         else:
             kind = TransactionKind.AUTH
+
+        action = result.message.get("action")
```
korycins, line 251: I would suggest extracting `action` logic to separate function. Process_payment is growing and adding helper function for `action` would improve readability.

saleor/payment/gateways/adyen/webhooks.py
```diff
@@ -486,3 +488,31 @@ def handle_webhook(request: WSGIRequest, gateway_config: "GatewayConfig"):
...
+    try:
+        payment = Payment.objects.get(pk=payment_pk)
+    except ObjectDoesNotExist:
+        raise Exception("Cannot perform payment. Payment does not exists.")
+
+    data = json.loads(payment.extra_data)
+    return_url = payment.return_url
+
+    if not return_url:
+        raise Exception("Cannot perform payment. Lack of data about returnUrl.")
```
korycins, line 505: Not sure if low lvl `Exception` is a good way.  The payment module has its own error lvl `PaymentError`.
Also. Is it expected that we don't catch these raises? Won't it close the python proces?

IKarbowiak: I am returning `HttpResponseNotFound` right now.

IKarbowiak:
This test `test_get_payment_gateway_for_checkout` failing because of changes in `adyen_integration` branch will be fixed in different PR.

saleor/payment/gateways/adyen/plugin.py
```diff
@@ -229,9 +235,16 @@ def get_payment_gateway_for_checkout(
     def process_payment(
         self, payment_information: "PaymentData", previous_value
     ) -> "GatewayResponse":
+        params = urlencode({"payment": payment_information.payment_id})
```
korycins, line 238: It is not safe to build URL with the only payment_id. It is base64 which is easy to decode. We should add here checkout_id also

PR #5938, opened 2020-07-31
Products export unsupported operation fix

tomaszszymanski129:
I want to merge this change because it fixes broken CSV export when using S3 as storage.

Tested locally both with local storage and S3.

Removed broken tests for now, they have to be rerwitten and so will be done in subsequent PR.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


tomaszszymanski129:
@IKarbowiak The file is being deleted after the function ends as it closes automatically and is being deleted (as default behavior of `NamedTemporaryFile`) but you're right, it's better to explictily close it. As for tests i plan to re-write them shortly so that doesn't really matter if we remove them here.

IKarbowiak:
In my opinion is, if somebody else will be working on fixing the tests, it will be easier to refactor existing one, that writing new ones from the beginning or looking for old ones in history.

IKarbowiak:
Looks good 👍, but did you test it on the big sample like 50 000 and 100 000 products?

tomaszszymanski129:
@IKarbowiak tested with local storage and S3 with over 100k products in database. Resulting file is about 5mb, takes some time but works flawlessly (and of course uploads without any problems to S3).

PR #5949, opened 2020-08-03
Add deleteChannel mutation

d-wysocki:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested

### DIFF
--- a/saleor/graphql/channel/mutations.py
+++ b/saleor/graphql/channel/mutations.py
@@ -1,5 +1,6 @@
 import graphene
 from django.core.exceptions import ValidationError
+from django.db import transaction
 
 from ...channel import models
 from ...checkout.models import Checkout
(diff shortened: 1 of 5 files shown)


saleor/graphql/channel/mutations.py
```diff
@@ -51,3 +54,63 @@ class Meta:
...
+
+class ChannelDelete(ModelDeleteMutation):
+    class Arguments:
+        id = graphene.ID(required=True, description="ID of a channel to delete.")
+        input = ChannelDeleteInput(
+            required=True, description="Fields required to delete a channel."
+        )
+
+    class Meta:
+        description = "Delete a channel."
```
maarcingebala, line 74: This description is part of the documentation so we can use it to describe the behavior a bit:

"Delete a channel. Orders associated with the deleted channel will be moved to the target channel. Checkouts, product availability, and pricing will be removed."

saleor/graphql/channel/mutations.py
```diff
@@ -51,3 +54,63 @@ class Meta:
...
+                        code=ChannelErrorCode.CHANNEL_TARGET_ID_MUST_BE_DIFFERENT,
+                    )
+                }
+            )
+
+    @classmethod
+    def perform_delete(cls, origin_channel, target_channel):
+        cls.validate_input(origin_channel, target_channel)
+        cls.migrate_orders_to_target_channel(origin_channel, target_channel)
+        cls.delete_checkouts(origin_channel)
```
maarcingebala, line 97: To make it safer and resilient to race conditions, we could wrap these with `transaction.atomic()` and use `select_for_update` on Order and Checkout querysets inside these two functions. Then we should avoid the risk that new order is created in a channel that will be soon deleted.

PR #5952, opened 2020-08-04
Drop channel middleware

maarcingebala:
Drop channel middleware and replace it with explicitly passed context objects that store the channel that was passed in the top-level resolver and the resolved object.

Some product mutations may still not work but I think we should merge it to not use channel middleware anymore in other PRs and in the meantime I'll be fixing other things that do not work.

Other changes:
- Removed implicit channel arg from `PrefetchingConnectionField` - I think we should use it explicitly and make sure that it's properly resolved, instead of automatically adding it to all connections. Also, it's not needed in all of them now. The same applies to `FieldWithChannel` field, for now, I removed its usages and will keep adding channel arg where it's needed.
- added some docstrings to utility functions
- renamed custom exceptions

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/channel/__init__.py
+++ b/saleor/graphql/channel/__init__.py
@@ -1,0 +1,16 @@
+from dataclasses import dataclass
+from typing import Any, Optional
+
+from django.db.models import QuerySet
+
+
+@dataclass
+class ChannelContext:
+    node: Any
+    channel_slug: Optional[str] = None
+
+
+@dataclass
+class ChannelQsContext:
+    qs: QuerySet
+    channel_slug: str
--- a/saleor/graphql/channel/__init__.py
+++ b/saleor/graphql/channel/__init__.py
@@ -7,10 +7,10 @@ from django.db.models import QuerySet
 @dataclass
 class ChannelContext:
     node: Any
-    channel_slug: Optional[str] = None
+    channel_slug: Optional[str]
 
 
 @dataclass
 class ChannelQsContext:
     qs: QuerySet
-    channel_slug: Optional[str] = None
+    channel_slug: Optional[str]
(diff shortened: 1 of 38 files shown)


saleor/graphql/channel/__init__.py
```diff
@@ -0,0 +1,16 @@
+from dataclasses import dataclass
+from typing import Any, Optional
+
+from django.db.models import QuerySet
+
+
+@dataclass
+class ChannelContext:
+    node: Any
```
patrys, line 9: This works, but consider:

```python
from typing import Generic, TypeVar

N = TypeVar("N")

@dataclass
class ChannelContext(Generic[N]):
    node: N
    channel_slug: Optional[str] = None
```

then you can use strong typing:

```python
def resolve_foo(root: ChannelContext[Product], ...):
    root.node # <- will be typed automatically

def resolve_bar(...) -> ChannelContext[ProductVariant]:
    return ...
```

A similar construct could be used for ChannelQsContext (assuming we have `django-stubs` installed for mypy).

maarcingebala: 👍

saleor/graphql/channel/types.py
```diff
@@ -1,7 +1,49 @@
...
+from graphene_django import DjangoObjectType
 
 from ...channel import models
 from ..core.connection import CountableDjangoObjectType
+from ..meta.types import ObjectWithMetadata
+from ..translations.resolvers import resolve_translation
+from . import ChannelContext
+
+
+class ChannelContextType(DjangoObjectType):
```
patrys, line 12: This could also be a generic:

```python
N = TypeVar("N")

class ChannelContextType(DjangoObjectType, Generic[N]):
    ...

    @staticmethod
    def resolver_with_context(
        attname, default_value, root: ChannelContext[N], info, **args
    ):
        ...

    # etc.
```

PR #5956, opened 2020-08-05
Add auto capture flags for Adyen and Saleor

korycins:
Add two auto_capture flags to handle the cases where:
- Immediately auto-capture is enabled on the Adyen side
- auto-capture on the Adyen side is disabled and Saleor uses own auto-capture logic

### DIFF
--- a/saleor/payment/gateways/adyen/plugin.py
+++ b/saleor/payment/gateways/adyen/plugin.py
@@ -15,9 +15,9 @@ from ...models import Transaction
 from ..utils import get_supported_currencies
 from .utils import (
     api_call,
+    call_capture,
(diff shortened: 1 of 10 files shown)


saleor/payment/gateways/adyen/plugin.py
```diff
@@ -294,6 +314,7 @@ def refund_payment(
     ) -> "GatewayResponse":
 
         _type, payment_id = from_global_id(payment_information.payment_id)
+        # we take Auth kind because it contains the transaction it that we need
```
IKarbowiak, line 317: Do you mean:
```suggestion
        # we take Auth kind because it contains the transaction id that we need
```

PR #5985, opened 2020-08-12
Adyen add void, rebuild confirm, lock payment changes 

korycins:
- add cancel(void) to adyen
- rebuild confirm action to properly handle the async data
- add lock to confirm that we don't process the same transaction twice. (Thread race - adyen webhook vs checkoutComplete)


saleor/payment/gateway.py
```diff
@@ -54,9 +57,20 @@ def wrapped(payment: Payment, *args, **kwargs):
     return wrapped
 
 
-@payment_postprocess
+def transaction_block(fn: Callable) -> Callable:
```
fowczarek, line 60: The name is not clear from my perspective but I can't find better name. Maby we should add docstring for this decorator?

saleor/payment/gateway.py
```diff
@@ -128,6 +154,10 @@ def capture(
     )
     if response and response.payment_method_info:
         update_payment_method_details(payment, response)
+    if response and response.transaction_already_processed:
+        transaction = get_already_processed_transaction(payment, response)
+        if transaction:
+            return transaction
```
fowczarek, line 160: This code is duplicated in many places. Maby we should extract it to some utils?

saleor/payment/gateway.py
maarcingebala: Maybe `with_locked_payment`?

maarcingebala: Plus a docstring.

saleor/payment/gateways/adyen/plugin.py
```diff
@@ -442,3 +479,37 @@ def capture_payment(
...
+            Transaction.objects.filter(
+                payment__id=payment_information.payment_id,
+                kind=TransactionKind.AUTH,
+                is_success=True,
+            )
+            .exclude(token__isnull=True, token__exact="")
+            .last()
+        )
+        if not transaction:
+            raise PaymentError("Cannot find a payment reference to void.")
```
maarcingebala, line 497: Just wondering - is this Adyen-specific or it should be part of `can_void`?

korycins: It is specific for Adyen as any operation requires the reference received from the Auth operation.

PR #5991, opened 2020-08-12
Use product and variant names for klarna product description

korycins:
Instead of description on the klarna side like:
image
We use product and variant names:
image


saleor/payment/gateways/adyen/tests/test_utils.py
```diff
@@ -52,7 +52,7 @@ def test_append_klarna_data(dummy_payment_data, payment_dummy, checkout_with_ite
         "shopperEmail": dummy_payment_data.customer_email,
         "lineItems": [
             {
-                "description": line.variant.product.description,
+                "description": f"{line.variant.product.name} - {line.variant.name}",
```
patrys, line 55: Maybe a comma would be better than a hyphen?

PR #6017, opened 2020-08-17
Send a confirmation email when the order is canceled or refunded

IKarbowiak:
Send a confirmation email when the order is canceled or refunded.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/order/actions.py
```diff
@@ -83,6 +88,8 @@ def cancel_order(order: "Order", user: Optional["User"]):
     manager.order_cancelled(order)
     manager.order_updated(order)
 
+    send_order_canceled_confirmation(order, user)
```
maarcingebala, line 91: Not a big issue but sometimes we say "cancelled" and sometimes "canceled" - both are correct but we have them mixed in entire codebase. Do you have any opinion on which one should we use @patrys?

saleor/order/emails.py
```diff
@@ -88,7 +96,7 @@ def send_order_confirmation(order_pk, redirect_url, user_pk=None):
         order=email_data["context"]["order"],
         user=None,
         user_pk=user_pk,
-        email_type=events.OrderEventsEmails.ORDER,
+        email_type=events.OrderEventsEmails.ORDER_CONFIRMATION,
```
maarcingebala, line 99: This is change is probably correct but we need to confirm with @dominik-zeglen if this doesn't break the events UI in the dashboard. I think we use these type fields for rendering there but I'm not sure.

maarcingebala: If we used this `OrderEventsEmails.ORDER`, we will have to open a task for dashboard to change it to `ORDER_CONFIRMATION`.

saleor/order/actions.py
patrys: "Cancelled" is British, "canceled" is American.

maarcingebala: Yeah I know, but do we have any preference on which to use?

patrys: Looking at the code it seems American English is used almost everywhere (including the code itself).

saleor/order/emails.py
dominik-zeglen: Well, we do :)

maarcingebala: Then please open a task to change it in the dashboard or if this is a simple replacement, we can fix it right away.

IKarbowiak: I created a task.

Issue #6001, opened 2020-08-14
Insufficient Stock Error for Digital  Product Type


neerajgupta2407:
### What I'm trying to achieve

I want to checkout for digital type of product irrespective of country of shipping address and without worrying about 
inventory.

Currently i want to offer online E-Learning course which user can buy across world. 
I don't want to setup any warehouse or its inventory as it doesn't make sense for Digital Type Product.

Now when i am checking out on Storefront, its throwing error saying "Not enough Quantity" ( as Saleor core checkout if failing due to ` InsufficientStock  Exception in check_stock_quantity(variant, country, quantity)` 

I have also verified these boolean fields. . `product_type.is_digital = True, product_type.is_shipping_required = False`

### Steps to reproduce the problem
1. Create Digital Product Type. 
2. Create Product with Digital Product type  and don't assign quantity on any warehouse.
3. Checkout the product on Storefront

### What I expected to happen
I should be able to checkout without any InsufficientStock exception for Digital Type of products.

### Screenshots

https://prnt.sc/tzj3vw

**System information**
Operating system:
Linux

maarcingebala:
`ProductVariant` model has a field `track_inventory`. Can you see what value is set for that field in your product variant? When it's set to `false` than the quantity checks should be bypassed in checkout and it would work as you expect. This value defaults to `true` at the model level, but maybe it should be changed in case of digital products.

neerajgupta2407:
Hi @maarcingebala , Yes, track_inventory is set to false.
Problem which i was able to figure out is `InsufficientStock` exception is thrown before the `variant.track_inventory` check.

```
Current code : function: saleor/warehouse/availability.py: check_stock_quantity

stocks = Stock.objects.get_variant_stocks_for_country(country_code, variant)
    if  not stocks:
        raise InsufficientStock(variant).   ### Exception is thrown here..

    if variant.track_inventory and quantity > _get_available_quantity(stocks):
        raise InsufficientStock(variant)
```
This should fix the problem:

```
    if variant.track_inventory:
        stocks = Stock.objects.get_variant_stocks_for_country(country_code, variant)
        if not stocks:
            raise InsufficientStock(variant)

        if quantity > _get_available_quantity(stocks):
            raise InsufficientStock(variant)
```

Kindly suggest if this change should fix or it requires more changes?

maarcingebala:
[...]

PR #6018, opened 2020-08-17
Fix for variant false with inventory tracking.  Issue #6001

neerajgupta2407:
Fix for variant without inventory tracking should be able to create order without throwing InsufficientStock exception.
Currently Exception is thrown when we add a variant with tracking_ventory set as False

Issue: #6001 
# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "saleor"
-version = "2.11.0-a.0"
+version = "2.11.0-rc.1"
 description = "A modular, high performance e-commerce storefront built with GraphQL, Django, and ReactJS."
 authors = ["Mirumee Software <hello@mirumee.com>"]
 license = "BSD-3-Clause"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "saleor"
-version = "2.11.0-rc.1"
+version = "2.11.0-rc.2"
 description = "A modular, high performance e-commerce storefront built with GraphQL, Django, and ReactJS."
 authors = ["Mirumee Software <hello@mirumee.com>"]
 license = "BSD-3-Clause"
--- a/saleor/warehouse/availability.py
+++ b/saleor/warehouse/availability.py
@@ -35,12 +35,13 @@ def check_stock_quantity(variant: "ProductVariant", country_code: str, quantity:
     If so - returns None. If there is less stock then required raise InsufficientStock
     exception.
     """
-    stocks = Stock.objects.get_variant_stocks_for_country(country_code, variant)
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "saleor"
-version = "2.11.0-a.0"
+version = "2.11.0-rc.2"
```
maarcingebala, line 3: These version changes (here and above) are unrelated to this PR so we shouldn't change it.

saleor/warehouse/availability.py
```diff
@@ -35,12 +35,13 @@ def check_stock_quantity(variant: "ProductVariant", country_code: str, quantity:
     If so - returns None. If there is less stock then required raise InsufficientStock
     exception.
     """
-    stocks = Stock.objects.get_variant_stocks_for_country(country_code, variant)
-    if not stocks:
-        raise InsufficientStock(variant)
+    if True: #variant.track_inventory:
```
maarcingebala, line 38: Did you mean `if variant.track_inventory`? Because `if True` does nothing and will always pass.

saleor/warehouse/availability.py
```diff
@@ -35,12 +35,13 @@ def check_stock_quantity(variant: "ProductVariant", country_code: str, quantity:
...
+        stocks = Stock.objects.get_variant_stocks_for_country(country_code, variant)
+        if not stocks:
+            raise InsufficientStock(variant)
 
-    if variant.track_inventory and quantity > _get_available_quantity(stocks):
-        raise InsufficientStock(variant)
+        if variant.track_inventory and quantity > _get_available_quantity(stocks):
```
maarcingebala, line 43: If we change the first condition as I suggested then we won't need to check `if variant.track_inventory` again. The second part `if quantity > _get_available_quantity(stocks)` will be enough.

pyproject.toml
neerajgupta2407: I checked out from 2.11 branch and create merge request for Master?
 
Is it right or i was supposed to do something else?

maarcingebala: You should check out from `master`. `2.11` is a release branch that we maintain and we're currently using it to prepare the `2.11` release.

maarcingebala:
Our CI requires a few checks to pass in order to maintain code quality, please take a look at this two failing checks: https://github.com/mirumee/saleor/pull/6018/checks?check_run_id=994042418 
The code needs to be formatted with [Black](https://github.com/psf/black) and apply to Flake8 rules.

pyproject.toml
neerajgupta2407: OK.I will create a new PR.

PR #6020, opened 2020-08-18
Update MenuItemMoveInput sort_order description

IKarbowiak:
Update `MenuItemMoveInput` `sort_order` description. Current description is misleading.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/menu/types.py
+++ b/saleor/graphql/menu/types.py
@@ -96,5 +96,5 @@ class MenuItemMoveInput(graphene.InputObjectType):
         description="ID of the parent menu. If empty, menu will be top level menu."
     )
     sort_order = graphene.Int(
-        description="Sorting position of the menu item (from 0 to x)."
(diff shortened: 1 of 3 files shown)


saleor/graphql/menu/types.py
```diff
@@ -96,5 +96,5 @@ class MenuItemMoveInput(graphene.InputObjectType):
         description="ID of the parent menu. If empty, menu will be top level menu."
     )
     sort_order = graphene.Int(
-        description="Sorting position of the menu item (from 0 to x)."
+        description="The new relative sorting position of the item (from -inf to +inf)."
```
maarcingebala, line 99: Maybe we should also add an example here like e.g:  `1` moves the item one position forward, `-1` moves the item one position backward, `0` leaves the item unchanged. This could go to all `sort_order` args if we have more of them that work that way.

PR #6021, opened 2020-08-18
Fix: checking variant.track_inventory while checking for stock_quantity

neerajgupta2407:

Fix for variant without inventory tracking should be able to create order without throwing InsufficientStock exception.
Currently Exception is thrown when we add a variant with tracking_ventory set as False

Issue: #6001 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/tests/test_checkout.py
+++ b/saleor/checkout/tests/test_checkout.py
@@ -9,6 +9,19 @@ from django_countries.fields import Country
 from freezegun import freeze_time
 from prices import Money, TaxedMoney
 
+from ...account import CustomerEvents
+from ...account.models import Address, CustomerEvent, User
+from ...account.utils import store_user_address
+from ...core.exceptions import InsufficientStock
+from ...core.taxes import zero_money, zero_taxed_money
+from ...discount import DiscountValueType, VoucherType
+from ...discount.models import NotApplicable, Voucher
+from ...order import OrderEvents, OrderEventsEmails
+from ...order.models import OrderEvent
+from ...payment.models import Payment
+from ...plugins.manager import get_plugins_manager
+from ...shipping.models import ShippingZone
+from ...tests.utils import flush_post_commit_hooks
 from .. import AddressType, calculations
 from ..models import Checkout
 from ..utils import (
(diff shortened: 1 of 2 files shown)


saleor/checkout/tests/test_checkout.py
```diff
@@ -40,6 +27,19 @@
...
+from ...core.exceptions import InsufficientStock
+from ...core.taxes import zero_money, zero_taxed_money
+from ...discount import DiscountValueType, VoucherType
+from ...discount.models import NotApplicable, Voucher
+from ...order import OrderEvents, OrderEventsEmails
+from ...order.models import OrderEvent
+from ...payment.models import Payment
+from ...plugins.manager import get_plugins_manager
+from ...shipping.models import ShippingZone
+from ...tests.utils import flush_post_commit_hooks
```
maarcingebala, line 42: I think this change is not needed; also `isort` fails in CI now supposedly because of it.

neerajgupta2407: I See. 
I applied the Black and it changes this import.
Isort is reverting to earlier version.

Is there any steps which i can follow while pushing PR?

saleor/checkout/tests/test_checkout.py
```diff
@@ -1491,3 +1491,26 @@ def test_cancel_active_payments(checkout_with_payments):
...
+
+def test_create_order_with_variant_tracking_false(
+    checkout, customer_user, variant_without_inventory_tracking
+):
+    variant = variant_without_inventory_tracking
+    add_variant_to_checkout(checkout, variant, 10, check_quantity=False)
+    checkout.user = customer_user
+    checkout.billing_address = customer_user.default_billing_address
+    checkout.shipping_address = customer_user.default_billing_address
+    checkout.save()
```
fowczarek, line 1504:
```suggestion
    checkout.user = customer_user
    checkout.billing_address = customer_user.default_billing_address
    checkout.shipping_address = customer_user.default_billing_address
    checkout.save()
    add_variant_to_checkout(checkout, variant, 10, check_quantity=False)
```
We should first set checkout addresses then assign variant to checkout because the possibility to assign variant to checkout is related to address.

neerajgupta2407: @fowczarek Thanks for the suggestion.
But during `add_variant_to_checkout` , no address check happens.  So i don't see anything specific check in your version of code. Let me know if i am missing on some scenarios.

Also the same flow is used in `test_create_order_insufficient_stock`.  So i am assuming the flow is correct.

fowczarek: `add_variant_to_checkout` call `check_variant_in_stock` and this use `checkout.get_country()` 
Addreses are used by  `checkout.get_country()`.

In `test_create_order_insufficient_stock` we use different order to check if `prepare_order_data` raises an error when one of a product becomes unavailable after is added to the checkout.

neerajgupta2407: @fowczarek Got your point and this should be the case of a general product where inventory_tracking is set to True.

But i has one doubt. Here we are checking for variant for which inventory_tracking is set to false.

Now when we are not not tracking inventory, skipping the stock availability check in `check_stock_quantity`, then does this suggestion really affect the flow ?

maarcingebala:
Thanks for your contribution @neerajgupta2407!

PR #6048, opened 2020-08-24
Lock checkoutComplete, create order from adyen notification

korycins:
- refactor checkoutComplete mutation
- extract whole checkoutComplete logic to separate file
- use checkout_complete logic to create order from Adyen notification
- apply lock for handling the adyen notifciation and mutation checkoutComplete


fowczarek (approved):
For now it great.

patrys:
Are the schema changes intentional?

korycins:
> Are the schema changes intentional?
@patrys , I didn't add any changes for schema. Not sure why it raises the error. I assume that I need to update my branch with master

saleor/checkout/utils.py
```diff
@@ -695,9 +696,14 @@ def abort_order_data(order_data: dict):
...
 
+def get_order(checkout_token: UUID) -> Optional[Order]:
+    """Get order based on checkout token."""
+    return Order.objects.confirmed().filter(checkout_token=checkout_token).first()
+
+
 @transaction.atomic
 def create_order(
-    *, checkout: Checkout, order_data: dict, user: User, redirect_url: str
+    *, checkout: Checkout, order_data: dict, user: User  # type: ignore
```
IKarbowiak, line 706: Why there is `# type: ignore` in function definition?

IKarbowiak (approved):

saleor/checkout/utils.py
fowczarek: :policewoman:

korycins: 👮‍♀️ 🚔

fowczarek (approved):
:rocket:

d-wysocki (approved):

saleor/checkout/complete_checkout.py
```diff
@@ -0,0 +1,190 @@
...
+        gateway.payment_refund_or_void(payment)
+        raise ValidationError(
+            "Payment does not cover all checkout value.",
+            code=CheckoutErrorCode.CHECKOUT_NOT_FULLY_PAID.value,
+        )
+
+
+def get_order_data(checkout: models.Checkout, discounts: List[DiscountInfo]) -> dict:
+    try:
+        with transaction.atomic():
```
tomaszszymanski129, line 85: is this atomic needed here?

saleor/payment/utils.py
```diff
@@ -194,9 +194,9 @@ def get_already_processed_transaction_or_create_new_transaction(
     error_msg=None,
 ) -> Transaction:
     if gateway_response and gateway_response.transaction_already_processed:
-        transaction = get_already_processed_transaction(payment, gateway_response)
-        if transaction:
-            return transaction
+        txn = get_already_processed_transaction(payment, gateway_response)
+        if txn:
+            return txn
     return create_transaction(
```
tomaszszymanski129, line 200:
```suggestion
    return txn or create_transaction(
```

tomaszszymanski129 (approved):

patrys:
13 new database queries, that's a lot.

korycins:
> 13 new database queries, that's a lot.

@patrys, I will doube check them

saleor/checkout/complete_checkout.py
korycins: So what do you propose? It prepares the data for order like locking the vouchers etc

patrys: Keep in mind that if we do locking, anything we `SELECT FOR UPDATE` will be unlocked as soon as the transaction terminates.

saleor/checkout/complete_checkout.py
```diff
@@ -0,0 +1,178 @@
...
+from ..payment.models import Payment, Transaction
+from ..payment.utils import store_customer_id
+from . import models
+from .checkout_cleaner import clean_checkout_payment, clean_checkout_shipping
+from .error_codes import CheckoutErrorCode
+from .utils import abort_order_data, create_order, get_order, prepare_order_data
+
+
+@transaction.atomic
+def prepare_checkout(
```
maarcingebala, line 25: I'm not sure what's the responsibility of this function, the name "prepare_checkout" doesn't explain much, there is also no docstring here.

saleor/checkout/complete_checkout.py
maarcingebala: I agree with @tomaszszymanski129, I would expect this function to only return some data. Also, `prepare_order_data` seems to be used only in "checkout complete", so why isn't it in this module?

saleor/checkout/complete_checkout.py
```diff
@@ -0,0 +1,178 @@
...
+                payment=payment,
+                token=payment.token,
+                store_source=store_source,
+                additional_data=payment_data,
+            )
+        payment.refresh_from_db()
+        if not txn.is_success:
+            raise PaymentError(txn.error)
+    except PaymentError as e:
+        abort_order_data(order_data)
```
maarcingebala, line 110: `abort_order_data` is also only used for completing checkouts, can we move it to this module? Also, this name is really awkward, we could fix it on this occasion.

saleor/checkout/complete_checkout.py
```diff
@@ -0,0 +1,178 @@
...
+    discounts,
+    user,
+    tracking_code=None,
+    redirect_url=None,
+) -> Tuple[Optional[Order], bool, dict]:
+    action_required = False
+    action_data = {}  # type: ignore
+    order = get_order(checkout.token)
+    if order:
+        return order, action_required, action_data
```
maarcingebala, line 128: What do you think about having a dataclass to represent the results of this function e.g. `CheckoutCompleteResult` @korycins  @patrys? It would be more elegant to process in places where the function is called, instead of expecting three tuple elements in this particular order? Just a suggestion, if you think it's overengineering, we can leave it as it is.

saleor/checkout/complete_checkout.py
korycins: @maarcingebala  @tomaszszymanski129  So what name do you propose?

tomaszszymanski129: Hard to think of anything appropriate, maybe something like assemble_order? @maarcingebala any other idea?

saleor/checkout/complete_checkout.py
korycins: I've added the docstring. See if it is clear enough now

saleor/checkout/complete_checkout.py
korycins: Changed

saleor/checkout/complete_checkout.py
korycins: I would say it is overengineering but I can change it if you think that this is a good place for data class.

maarcingebala: OK so let's leave it as it is now.

korycins:
@patrys About the number of queries. After I've updated the benchmark tests, master has 145 queries for checkout complete. Here we have 119
```
test name                             left count      right count     duplicate count
complete checkout                                 145             119                  25
```

patrys:
@korycins Was this branch rebased after the test change that increased the number of queries on master? Trying to make sure we're not comparing apples to oranges here.

korycins:
@patrys Yes. I did that to get the current number of queries. - commit 3b9cafa

saleor/checkout/checkout_cleaner.py
```diff
@@ -1,27 +1,28 @@
...
+    error_code: Union[Type[CheckoutErrorCode], Type[PaymentErrorCode]],
 ):
     if checkout.is_shipping_required():
         if not checkout.shipping_method:
             raise ValidationError(
                 {
                     "shipping_method": ValidationError(
                         "Shipping method is not set",
-                        code=error_code.SHIPPING_METHOD_NOT_SET,
+                        code=error_code.SHIPPING_METHOD_NOT_SET.value,
```
patrys, line 25: Why are those changes needed?

saleor/core/transactions.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@contextmanager
+def transaction_with_commit_on_errors():
+    """Perform transaction and raise an error in any occurred."""
+    error = None
+    with transaction.atomic():
+        try:
+            yield
+        except (DatabaseError, IntegrityError):
```
patrys, line 13: `IntegrityError` is a subclass of `DatabaseError`.

korycins: Yes, I know but I thought that passing it here explicitly will be a better way

maarcingebala (approved):

saleor/checkout/checkout_cleaner.py
korycins: I am getting: 
```
Unexpected type(s): (str, Union[CheckoutErrorCode, PaymentErrorCode]) Possible types: (Any, Optional[str]) (Any, Optional[str]) 
```
And then it raises the exception on mypy:
```
saleor/checkout/checkout_cleaner.py:43: error: Argument "code" to "ValidationError" has incompatible type "Union[CheckoutErrorCode, PaymentErrorCode]"; expected "Optional[str]"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

saleor/core/transactions.py
patrys: To me, it's confusing right now and caused me to double-check if there were maybe different classes with the same names.

korycins: Okay, I will remove `IntegrityError`

saleor/core/transactions.py
```diff
@@ -0,0 +1,18 @@
...
+
+@contextmanager
+def transaction_with_commit_on_errors():
+    """Perform transaction and raise an error in any occurred."""
+    error = None
+    with transaction.atomic():
+        try:
+            yield
+        except DatabaseError:
+            raise
```
IKarbowiak, line 14: Shouldn't we raise something here?

korycins: We re-raise the DatabaseError here

IKarbowiak (approved):

fowczarek (approved):

PR #6069, opened 2020-08-28
Update translations

korycins:
Update translations


saleor/settings.py
```diff
@@ -95,9 +95,11 @@ def get_bool_from_env(name, default_value):
     ("is", "Icelandic"),
     ("it", "Italian"),
     ("ja", "Japanese"),
+    ("km", "Khmer"),
```
patrys, line 98: Are those languages supported by Django? If not, we cannot add them here as Django will crash when you try to activate them.

patrys: Nevermind, both are supported!

patrys (approved):

saleor/settings.py
korycins: Yes, ' Ive checked the,:
Khmer - https://github.com/django/django/blob/master/django/conf/locale/__init__.py#L323
https://github.com/django/django/blob/master/django/conf/locale/__init__.py#L389

PR #6087, opened 2020-09-01
Use checkout object with many lines in benchmark tests

korycins:
Most of the benchmark tests for counting the queries for checkout use simple relation with only one line. As a result, we don't see the issues with missing prefetches or increasing DB counts. This PR changes the checkout object used in benchmark test.
An increased number of queries in `django-queries` is expected.

### DIFF
--- a/saleor/graphql/checkout/tests/benchmark/conftest.py
+++ b/saleor/graphql/checkout/tests/benchmark/conftest.py
@@ -14,17 +14,26 @@ def customer_checkout(customer_user, checkout_with_voucher_percentage_and_shippi
 
 
 @pytest.fixture()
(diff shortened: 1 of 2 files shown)


maarcingebala (approved):

maarcingebala:
"Found 8 differences" 😢 . But we will fix it in 3.0 with more data loaders.

korycins:
> "Found 8 differences" 😢 . But we will fix it in 3.0 with more data loaders.

Yeap, and thanks to this we will be sure that we have correct info about queries

maarcingebala:
True dat.

fowczarek (approved):

patrys (approved):

PR #6090, opened 2020-09-01
Refactor minimal variant price

d-wysocki:
I want to merge this change because...I want to refactor minimal_variant_price to discounted_price and move it to ProductChannelListing.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/product/utils/variant_prices.py
```diff
@@ -5,52 +5,63 @@
...
+    )
+    product = product_channel_listing.product
+    collections = (product.collections.all(),)
+    for variant_channel_listing in variants_channel_listing:
+        variant_price = calculate_discounted_price(
+            product=product,
+            price=variant_channel_listing.price,
+            collections=collections,
+            discounts=discounts,
+        )
```
IKarbowiak, line 29: Are you sure this works correctly? I am curious why you creating a tuple from collections. As I see `calculate_discounted_price` require `Iterable` and then collections are provided as an argument to `get_product_discounts`, where collections are used in this way:
`product_collections = set(pc.id for pc in collections)`. 

Maybe it should be like:
`collections = list(product.collections.all())` ?

d-wysocki: Good catch, thanks. I will fix it.

saleor/graphql/product/sorters.py
```diff
@@ -144,9 +144,9 @@ def description(self):
             ProductOrderField.NAME.name: "name",
             ProductOrderField.PRICE.name: "price",
             ProductOrderField.TYPE.name: "type",
-            ProductOrderField.MINIMAL_PRICE.name: (
-                "a minimal price of a product's variant"
-            ),
+            # ProductOrderField.MINIMAL_PRICE.name: (
+            #     "a minimal price of a product's variant"
+            # ),
```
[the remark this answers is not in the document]
maarcingebala: I guess this field should be renamed to `DISCOUNTED_PRICE`. If we change it we need to open a task for storefront because they're using `MINIMAL_PRICE` now if I'm correct.

saleor/product/filters.py
```diff
@@ -49,9 +49,9 @@ class ProductFilter(SortedFilterSet):
     sort_by = OrderingFilter(
         label="Sort by", fields=SORT_BY_FIELDS.keys(), field_labels=SORT_BY_FIELDS,
     )
-    minimal_variant_price = RangeFilter(
-        label="Price", field_name="minimal_variant_price_amount",
-    )
+    # minimal_variant_price = RangeFilter(
+    #     label="Price", field_name="minimal_variant_price_amount",
+    # )
```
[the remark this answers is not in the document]
maarcingebala: Why is it commented?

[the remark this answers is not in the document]
fowczarek: I don't know why Damian comment out filters source code. We should allow passing new arg with a channel to have proper working filters. It should be a separate [task.](https://app.clickup.com/t/2549495/SALEOR-655) In my code, I skip only tests and left this source code. In the current state of multichannel not all filters and sorters are working.

saleor/product/models.py
```diff
@@ -506,6 +466,42 @@ def __str__(self):
...
+        product_ids = list(product_ids)
+
+        from .tasks import update_products_discounted_prices_of_catalogues_task
+
+        update_products_discounted_prices_of_catalogues_task.delay(
+            product_ids=product_ids
+        )
+        return variants
+
+
```
maarcingebala, line 504: I'm thinking about dropping this code. IMO we should explicitly call tasks to update discounted prices in places where we know we need to do this (which we already do in mutations for example). I'm afraid that calling tasks automatically in ORM methods might bite us and e.g. trigger these tasks too often.

patrys: I wholeheartedly agree with @maarcingebala.

maarcingebala: Let's remove this and double-check that we update discount prices when creating or updating variants (variant channel listings).

saleor/graphql/product/sorters.py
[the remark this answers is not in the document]
d-wysocki: We will handle it in separate [task](https://app.clickup.com/t/2549495/SALEOR-655). I left the comment in the code.

saleor/product/utils/variant_prices.py
```diff
@@ -5,52 +5,63 @@
...
-        product.minimal_variant_price_amount = minimal_variant_price.amount
-        if save:
-            product.save(update_fields=["minimal_variant_price_amount", "updated_at"])
-    return product
+    product_channel_listings = product.channel_listing.all()
+    changed_products_channels_to_update = []
+    for product_channel_listing in product_channel_listings:
+        product_discounted_price = _get_product_discounted_price(
+            product_channel_listing, discounts
+        )
```
maarcingebala, line 45: `_get_product_discounted_price` makes a database query with `ProductVariantChannelListing.objects.filter()` so we will have duplicated queries in this loop. Maybe we can improve the performance here?

saleor/product/filters.py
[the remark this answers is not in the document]
d-wysocki: Yes, we need to this in separate task. I also left comment in the code, to do not forget to change it later

saleor/product/models.py
patrys: This may be a good opportunity to generate some charts with mermaid 😉

saleor/product/utils/variant_prices.py
```diff
@@ -1,56 +1,71 @@
...
-    # Start with the product's price as the minimal one
-    minimal_variant_price = None
-    for variant in product.variants.all():
-        variant_price = variant.get_price(discounts=discounts)
-        if minimal_variant_price is None:
-            minimal_variant_price = variant_price
-        else:
-            minimal_variant_price = min(minimal_variant_price, variant_price)
-    return minimal_variant_price
+def _get_varaint_prices_in_channels_dict(product):
```
maarcingebala, line 13: Typo in "varaint" 🚓

maarcingebala (approved):

IKarbowiak (approved):

PR #6095, opened 2020-09-03
Wrap order_created in transaction.on_commit

korycins:
- fix for calling the order_created before we saved the order.


patrys:
Task linked: [CU-7ux05t Action order_created called inside transaction and without wrapper transaction.on_commit](https://app.clickup.com/t/7ux05t)

maarcingebala (approved):

PR #6100, opened 2020-09-04
PRODUCT_UPDATED webhook

tomaszszymanski129:
I want to merge this change because it adds PRODUCT_UPDATED webhook.

TODO:
* [X] Add a new event to WebhookEvents
* [X] new hook to PluginManager - product_updated
* [X] new hook to BasePlugin - product_updated
* [X] new hook to WebhookPlugin - product_updated
* [X] payload should be generated by `saleor.webhook.payloads.generate_product_payload`

Where is this triggered?
* [X] `ProductUpdate` mutation
* [X] `ProductVariantCreate` mutation
* [X] `ProductVariantUpdate` mutation
* [x] `ProductSetAvailabilityForPurchase` mutation
* [x] `DeleteMetadata` mutation (selective)
* [x] `UpdatePrivateMetadata` mutation (selective)
* [x] `DeletePrivateMetadata` mutation (selective)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


IKarbowiak (approved):

saleor/graphql/product/mutations/products.py
```diff
@@ -1055,6 +1055,7 @@ def save(cls, info, instance, cleaned_input):
         attributes = cleaned_input.get("attributes")
         if attributes:
             AttributeAssignmentMixin.save(instance, attributes)
+        info.context.plugins.product_updated(instance)
```
korycins, line 1058: Is `order_updated` called when order is created? If yes, I assume we should have the same behaviour.
What about an update in the product's variants? Do we want to also call the hook in that case? cc @maarcingebala

korycins: What about the situation when the product has been deleted? What about the mutation for Product META?

tomaszszymanski129: > Is `order_updated` called when order is created? If yes, I assume we should have the same behaviour.

It's not called - there is a dedicated webhook product created for that, so I figured we shouldn't call update there.

> What about an update in the product's variants? Do we want to also call the hook in that case? cc @maarcingebala 

Added that according to the ticket, it makes sense.

About deleting the product - I think there should be separate webhook for that - product deleted.

About the META - probably we should trigger the update - it's an update to a product after all.

So the plan would be...
- create a ticket for product deleted webhook
- trigger product updated when META changes

CC @maarcingebala

fowczarek (commented):
Should we trigger this webhook on `ProductSetAvailabilityForPurchase` mutation?

korycins:
> Should we trigger this webhook on `ProductSetAvailabilityForPurchase` mutation?

Yes, everything related to changes in product/its variatns

saleor/graphql/product/mutations/products.py
```diff
@@ -1320,6 +1332,7 @@ def save(cls, info, instance, cleaned_input):
             AttributeAssignmentMixin.save(instance, attributes)
             instance.name = generate_name_for_variant(instance)
             instance.save(update_fields=["name"])
+        info.context.plugins.product_updated(instance.product)
```
korycins, line 1335: It seems that there are a lot of places where we could prefetch the related objects. Maybe we could move it `PluginManager.product_updated` ? What do you think?

maarcingebala (approved):

saleor/graphql/product/mutations/products.py
```diff
@@ -1890,4 +1903,5 @@ def perform_mutation(cls, _root, info, **data):
             product.available_for_purchase = start_date
 
         product.save(update_fields=["available_for_purchase", "updated_at"])
+        info.context.plugins.product_updated(product)
```
fowczarek, line 1906: We should prefetch `variants` in this mutation.

tomaszszymanski129: yep, im adding that as we speak

fowczarek:
@korycins @maarcingebala 
Maby `info.context.plugins.product_updated(product)` should get only product ID and fetch all needed data in an optimal way, instead of prefetching variants in all mutations witch trigger this webhook?

fowczarek (approved):

korycins (approved):

tomaszszymanski129:
Discussed with @fowczarek and @korycins, prefetching does no performance improvement since we call variants.all() later, there is a problem with the serializer that collects data for order (quantity is no longer on variant), that will be handled in subsequent PR.

saleor/graphql/meta/mutations.py
```diff
@@ -119,7 +119,21 @@ def perform_mutation(cls, root, info, **data):
         return cls.success_response(instance)
 
 
-class DeleteMetadata(BaseMetadataMutation):
+class ExtraMethodMetadataMixin:
```
fowczarek, line 122: Should we add this mixin to `UpdateMetadata` ?

fowczarek (approved):

PR #6117, opened 2020-09-10
Fix crashing system when avalara is improperly configured

IKarbowiak:
Fixing issue with `None` reaching quantize price - it was caused by improperly configured Avalara, so I add the check to prevent such a situation.

- Raise an error when avalara API return error in response
- Add authentication validation when activating avalara plugin
- Change the default value of `PluginConfiguration` to False

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/plugins/avatax/plugin.py
```diff
@@ -421,6 +422,24 @@ def fetch_taxes_data(self, previous_value):
...
+    @classmethod
+    def validate_authentication(cls, plugin_configuration: "PluginConfiguration"):
+        conf = {
+            data["name"]: data["value"] for data in plugin_configuration.configuration
+        }
+        url = urljoin(get_api_url(conf["Use sandbox"]), "utilities/ping")
+        response = api_get_request(
+            url,
+            username_or_account=conf["Username or account"],
+            password_or_license=conf["Password or license"],
```
maarcingebala, line 434: Do we use strings with spaces as keys in the configuration?

saleor/plugins/avatax/__init__.py
```diff
@@ -449,7 +455,9 @@ def get_cached_tax_codes_or_fetch(
     tax_codes = cache.get(TAX_CODES_CACHE_KEY, {})
     if not tax_codes:
         tax_codes_url = urljoin(get_api_url(config.use_sandbox), "definitions/taxcodes")
-        response = api_get_request(tax_codes_url, config)
+        response = api_get_request(
+            tax_codes_url, config.username_or_account, config.password_or_license
+        )
         if response and "error" not in response:
```
korycins, line 461: You raise the TaxError when response has error field so checking `"error" not in response` is not needed, right?

saleor/core/management/commands/populatedb.py
```diff
@@ -85,6 +86,11 @@ def sequence_reset(self):
             cursor.execute(commands.getvalue())
 
     def handle(self, *args, **options):
+        settings.PLUGINS = [
```
maarcingebala, line 89: Wondering if this is not a workaround. Maybe we should try setting some fake address in site settings that is required by Avalara.

saleor/plugins/avatax/plugin.py
```diff
@@ -432,12 +451,18 @@ def validate_plugin_configuration(cls, plugin_configuration: "PluginConfiguratio
...
+                    "To enable a plugin, you need to provide values for the "
+                    "following fields: "
+                )
+                raise ValidationError(
+                    error_msg + ", ".join(missing_fields),
+                    code=PluginErrorCode.PLUGIN_MISCONFIGURED.value,
+                )
+
+            try:
+                cls.validate_authentication(plugin_configuration)
```
korycins, line 466: Try catch is not needed here

saleor/core/management/commands/populatedb.py
korycins: I am not sure that we want to have each new order created by populated DB directly send to Avalara.

maarcingebala: Probably not, are we sending them to Avalara now?

IKarbowiak: Yes

korycins: Avalara as a plugin doesn't know if the order comes from populateDB or from normal user.

saleor/plugins/avatax/plugin.py
korycins: Unfortunately for Avalara, yes. 
This should be changed but imo not in this PR.

maarcingebala: Yes I agree it's a separate task but I thought we fixed it already, but now I see this was done properly only for Adyen.

korycins (approved):

maarcingebala (approved):

saleor/core/management/commands/populatedb.py
```diff
@@ -85,6 +86,13 @@ def sequence_reset(self):
             cursor.execute(commands.getvalue())
 
     def handle(self, *args, **options):
+        # set only our custom plugin to not call external API when preparing
+        # example database
+        settings.PLUGINS = [
+            "saleor.payment.gateways.dummy.plugin.DummyGatewayPlugin",
+            "saleor.payment.gateways.dummy_credit_card.plugin."
+            "DummyCreditCardGatewayPlugin",
```
patrys, line 94: @mirumee/saleor Please change this to be on one line, these two look like separate entries and it will lead to a bug when someone edits the list and either adds a comma or inserts a line between those two.

IKarbowiak: But then the line will be too long.

PR #6118, opened 2020-09-10
Allow REDIS_URL as CELERY_BROKER_URL

ohld:
I want to merge this change because in Heroku we can use only one REDIS_URL variable. And I don't want to create 2 Redis instances. I want to use one Redis for Celery broker and fo Cache. 

Is it a good fix?

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -470,7 +470,10 @@ AUTHENTICATION_BACKENDS = [
 # CELERY SETTINGS
 CELERY_TIMEZONE = TIME_ZONE
 CELERY_BROKER_URL = (
-    os.environ.get("CELERY_BROKER_URL", os.environ.get("CLOUDAMQP_URL")) or ""
+    os.environ.get(
+        "CELERY_BROKER_URL", 
+        os.environ.get("CLOUDAMQP_URL", os.environ.get("REDIS_URL"))
+    ) or ""
 )
 CELERY_TASK_ALWAYS_EAGER = not CELERY_BROKER_URL
 CELERY_ACCEPT_CONTENT = ["json"]


maarcingebala (approved):
I think this is a good idea, thanks for opening the PR!

maarcingebala:
@ohld There are two CI checks that are failing now - Black and flake8. Could you take a look at that? We use these linters to make sure that the code is formatted consistently. We also have precommit configuration so you can install it as well locally and these checks would be run when creating a commit. Let me know if you need any help.

korycins:
@ohld  I might be wrong as I don't know Heroku interface very well but If you want to have exactly the same instance, can't you provide the same value for  `CELERY_BROKER_URL` and `REDIS_URL`?
I don't think that we want to set `celery_broker` every time when we set `redis_url`.  When broker_url is empty we set `CELERY_TASK_ALWAYS_EAGER` to True.

korycins:
@marekbrzoska, @ohld I might be wrong as I don’t know Heroku interface very well but If you want to have exactly the same instance, can’t you provide the same value for  CELERY_BROKER_URL and REDIS_URL?
I don’t think that we want to set celery_broker every time when we set redis_url. When broker_url is empty we set CELERY_TASK_ALWAYS_EAGER to True.

maarcingebala:
That's a good point, @ohld have you tried just setting the same Redis URL in these two different envs separately?

fowczarek (approved):
LGFM. CI still failing and it should be fixed before the merge.

patrys:
I'm not a fan of this behavior as enabling cache introduces a surprising secondary effect of changing how background tasks are executed (or not executed if you never start a Celery worker). I'm afraid it will break environments that were previously working (small sites not wanting to spend resources on a separate background worker) and it will make local testing harder (as it ties two separate features to one env switch).

maarcingebala:
Thanks for your feedback and thank you @ohld for your effort, let me know if you have any questions!

PR #6119, opened 2020-09-11
Remove corresponding draft order lines when variant is removing

IKarbowiak:
When the variant is removing, the corresponding draft order lines too.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/product/mutations/products.py
+++ b/saleor/graphql/product/mutations/products.py
@@ -12,6 +12,7 @@ from graphql_relay import from_global_id
 
 from ....core.exceptions import PermissionDenied
 from ....core.permissions import ProductPermissions
+from ....order import OrderStatus, models as order_models
 from ....product import models
 from ....product.error_codes import ProductErrorCode
 from ....product.tasks import (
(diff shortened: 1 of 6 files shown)


fowczarek (approved):

saleor/graphql/product/mutations/products.py
```diff
@@ -1372,6 +1373,25 @@ def success_response(cls, instance):
...
+    @classmethod
+    def perform_mutation(cls, _root, info, **data):
+        node_id = data.get("id")
+        instance = cls.get_node_or_error(info, node_id, only_type=ProductVariant)
+
+        # get draft order lines for variant
+        line_pks = list(
+            order_models.OrderLine.objects.filter(
+                variant=instance, order__status=OrderStatus.DRAFT
+            ).values_list("pk", flat=True)
```
maarcingebala, line 1385: Minor thing but we could use the queryset that we have for drafts - `Order.objects.drafts()`.

maarcingebala (commented):
I think we need this logic in a few more places:
- `productVariantBulkDelete` mutation
- deleting a product, its variant will be deleted as well
- on deleting product types, as it also removes products

If it's problematic for some reasons, we don't have to cover all cases in this one PR, but we will need to have a task for that.

saleor/graphql/product/mutations/products.py
IKarbowiak: But we need here `OrderLine` instances not `Order`.

maarcingebala: Yep, sorry :)

maarcingebala (approved):

fowczarek (approved):

PR #6138, opened 2020-09-16
Allow product variants to be sortable

tomaszszymanski129:
I want to merge this change because it allows to change order of product variants.
TODO:
* [X] Variant model change (inherit from `SortableModel`)
* [X] Mutation for changing variants order
* [X] Make sure variants are returned in correct order
* [X] Unit tests
* [x] Update changelog

# Impact

* [X] New migrations

# Pull Request Checklist

* [X] The changes are tested


saleor/graphql/product/mutations/products.py
```diff
@@ -1784,6 +1784,67 @@ def perform_mutation(cls, _root, info, product_id, images_ids):
...
+    @classmethod
+    def perform_mutation(cls, _root, info, product_id, variants):
+        product = cls.get_node_or_error(
+            info, product_id, field="product_id", only_type=Product
+        )
+
+        if len(variants) != product.variants.count():
+            raise ValidationError(
+                {
+                    "order": ValidationError(
```
korycins, line 1817: Shouldn't `variants` here? The dashboard needs to know the field name to assign the error

saleor/graphql/product/mutations/products.py
```diff
@@ -1784,6 +1784,67 @@ def perform_mutation(cls, _root, info, product_id, images_ids):
...
+                    )
+                }
+            )
+
+        ordered_variants = []
+        for variant_id in variants:
+            variant = cls.get_node_or_error(
+                info, variant_id, field="order", only_type=ProductVariant
+            )
+            if variant and variant.product != product:
```
korycins, line 1829: IMO This will trigger additional DB queries, I don't think that we fetched product:
```
In [34]: from saleor.product.models import ProductVariant

In [35]: v = ProductVariant.objects.all()

In [36]: for x in v:
    ...:     w = x.product.name
    ...:
Number of queries 130

In [37]: v = ProductVariant.objects.select_related('product').all()

In [38]: for x in v:
    ...:     w = x.product.name
    ...:
Number of queries 1
```

saleor/graphql/product/mutations/products.py
```diff
@@ -1784,6 +1784,67 @@ def perform_mutation(cls, _root, info, product_id, images_ids):
...
+                raise ValidationError(
+                    {
+                        "order": ValidationError(
```
korycins, line 1832: `variants`?

saleor/graphql/product/mutations/products.py
```diff
@@ -1784,6 +1784,67 @@ def perform_mutation(cls, _root, info, product_id, images_ids):
...
+                            "Variant %(variant_id) does not belong to this product.",
+                            code=ProductErrorCode.NOT_PRODUCTS_VARIANT,
+                            params={"variant_id": variant.id},
+                        )
+                    }
+                )
+            ordered_variants.append(variant)
+
+        for _order, variant in enumerate(ordered_variants):
```
korycins, line 1841: Let's do this in bulk operation

korycins: https://docs.djangoproject.com/en/3.1/ref/models/querysets/#bulk-update

saleor/product/models.py
```diff
@@ -453,6 +453,9 @@ def get_first_image(self) -> "ProductImage":
         images = list(self.images.all())
         return images[0] if images else self.product.get_first_image()
 
+    def get_ordering_queryset(self):
```
fowczarek, line 456: Why do we add this function which is never called?

saleor/graphql/product/mutations/products.py
tomaszszymanski129: Yeah, copied that from other ordering mutations but this can be optimised further, get rid of `for` loop and use `get_nodeS_or_error`, will do that in a moment

saleor/graphql/product/mutations/products.py
fowczarek: 👮

saleor/graphql/product/mutations/products.py
```diff
@@ -1784,6 +1784,67 @@ def perform_mutation(cls, _root, info, product_id, images_ids):
...
+                        code=ProductErrorCode.INVALID,
+                    )
+                }
+            )
+
+        ordered_variants = []
+        for variant_id in variants:
+            variant = cls.get_node_or_error(
+                info, variant_id, field="order", only_type=ProductVariant
+            )
```
fowczarek, line 1828: Maybe we should use bulk load?

saleor/graphql/product/mutations/products.py
```diff
@@ -1784,6 +1784,67 @@ def perform_mutation(cls, _root, info, product_id, images_ids):
...
+                            code=ProductErrorCode.NOT_PRODUCTS_VARIANT,
+                            params={"variant_id": variant.id},
+                        )
+                    }
+                )
+            ordered_variants.append(variant)
+
+        for _order, variant in enumerate(ordered_variants):
+            variant.sort_order = _order
+            variant.save(update_fields=["sort_order"])
```
fowczarek, line 1843: @maarcingebala  Should we update `updated_at` on Product mode?

saleor/graphql/product/tests/test_product.py
```diff
@@ -1402,6 +1402,88 @@ def test_create_product(
...
+    }
+
+    response = staff_api_client.post_graphql(
+        REORDER_PRODUCT_VARIANTS_MUTATION,
+        variables,
+        permissions=[permission_manage_products],
+    )
+    content = get_graphql_content(response)
+    data = content["data"]["productVariantReorder"]
+    assert data["productErrors"][0]["code"] == ProductErrorCode.INVALID.name
```
fowczarek, line 1458: We should also check `field` on this error.

saleor/graphql/product/tests/test_product.py
```diff
@@ -1402,6 +1402,88 @@ def test_create_product(
...
+    }
+
+    response = staff_api_client.post_graphql(
+        REORDER_PRODUCT_VARIANTS_MUTATION,
+        variables,
+        permissions=[permission_manage_products],
+    )
+    content = get_graphql_content(response)
+    data = content["data"]["productVariantReorder"]
+    assert (
```
fowczarek, line 1482: We should also check `field` on this error.

fowczarek (commented):
Should we add some events or trigger some webhooks?

saleor/product/models.py
tomaszszymanski129: See `SortableModel` abstract which the class uses now

saleor/graphql/product/mutations/products.py
```diff
@@ -1784,6 +1784,67 @@ def perform_mutation(cls, _root, info, product_id, images_ids):
...
+class ProductVariantReorder(BaseMutation):
+    product = graphene.Field(Product)
+    variants = graphene.List(ProductVariant)
+
+    class Arguments:
+        product_id = graphene.ID(
+            required=True,
+            description="Id of product that variants order will be altered.",
+        )
+        variants = graphene.List(
```
maarcingebala, line 1796: We have two APIs for reordering from what I see now. The first approach with passing moves:
- `attributeReorderValues(attributeId: ID!, moves: [ReorderInput]!): AttributeReorderValues`
- `collectionReorderProducts(collectionId: ID!, moves: [MoveProductInput]!): CollectionReorderProducts`
- `productTypeReorderAttributes(moves: [ReorderInput]!, productTypeId: ID!, type: AttributeTypeEnum!): ProductTypeReorderAttributes`

And the second with passing a list of IDs in the new order:
- `productImageReorder(imagesIds: [ID]!, productId: ID!): ProductImageReorder`

Your new mutation uses the second approach, but I think we should everywhere use the first one with passing the list of particular reordering moves, is that right @patrys? The image reordering mutation should be in fact refactored as well (not in this PR ofc).

saleor/graphql/product/mutations/products.py
patrys: Please don't use `select_related` for sparse many-to-one relations (eg. when many variants can map to a single product) as it will result in the database sending us multiple copies of the same related row. `prefetch_related` is often a better idea when the relation is not one-to-one.

saleor/graphql/product/mutations/products.py
patrys: Yes, we want all mutations to be safe when used in a concurrent manner (two clients making changes that should not conflict with each other).

maarcingebala (approved):

fowczarek (approved):

PR #6140, opened 2020-09-18
Allow product variant to be set as defult

tomaszszymanski129:
I want to merge this change because it allows product variant to be set as default.
* [X] First variant created marks as default
* [X] `Product` graphql type returns `defaultVariant`
* [X] Mutation to change the default variant

[SALEOR-1047](https://app.clickup.com/t/2549495/SALEOR-1047)

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] The changes are tested


patrys:
Task linked: [SALEOR-1047 As a User I want to select the default variant of a product](https://app.clickup.com/t/2549495/SALEOR-1047)

saleor/graphql/product/types/products.py
```diff
@@ -516,6 +525,10 @@ def resolve_category(root: models.Product, info):
 
         return CategoryByIdLoader(info.context).load(category_id)
 
+    @staticmethod
+    def resolve_default_variant(root: models.Product, info):
+        return root.get_default_variant()
```
fowczarek, line 530: We should implement a data loader here.

saleor/product/models.py
```diff
@@ -313,6 +313,9 @@ def is_available_for_purchase(self):
             and datetime.date.today() >= self.available_for_purchase
         )
 
+    def get_default_variant(self):
```
fowczarek, line 316: We should avoid this kind of utils. Each usage of this function generate 1 database query.

saleor/product/models.py
```diff
@@ -462,6 +466,11 @@ def get_first_image(self) -> "ProductImage":
         images = list(self.images.all())
         return images[0] if images else self.product.get_first_image()
 
+    def set_as_default(self):
```
fowczarek, line 469: We should avoid this kind of utils. Each usage of this function generates 2 database queries. 
It's hard to optimize.

saleor/product/migrations/0125_productvariant_default.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.1.1 on 2020-09-17 15:46
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
fowczarek, line 6: Should we migrete old data?

fowczarek (commented):
What should the system do when the staff user removes the default variant?

fowczarek (commented):
We should add a test for passing `VariantId` not assigned to the product from `ProductId`

saleor/product/migrations/0125_productvariant_default.py
tomaszszymanski129: Was thinking about it but it wasn't in requirements and we don't have any logic for a product without default variant. For now (in this PR) the state is - when you create a new variant and there is no default, it sets itself as default.

saleor/product/models.py
maarcingebala: I would also avoid any model methods and instead created an optimal way of returning default variants in the resolver in the `Product` type.

maarcingebala: Another option is to have this as a utility function somewhere in `product/utils` but it should be prepared to work with dataloader later so accept list of variants as an argument and instead of doing filter it could just iterate over them and return the first variant where `default=True`.

saleor/product/models.py
maarcingebala: This utility is used in mutation so I'm not sure if we can further optimize it. But I would move it to `product/utils` to keep the model thinner.

saleor/product/migrations/0125_productvariant_default.py
tomaszszymanski129: @patrys please advise if we need the data migration.

saleor/graphql/product/types/products.py
```diff
@@ -474,6 +482,7 @@ class Product(CountableDjangoObjectType):
     variants = graphene.List(
         ProductVariant, description="List of variants for the product."
     )
+    default_variant = graphene.Field(ProductVariant)
```
maarcingebala, line 485: We should add a descritption here :)

maarcingebala (approved):

saleor/product/migrations/0125_productvariant_default.py
patrys: We need to set the first variant as default, the API has to return something and it's better to do this once instead of implementing costly runtime fallbacks

fowczarek: Is `default_variant` required?

fowczarek:
What do you think about store `default_value` in `Product` model? It's easier to get it, find it, and validate that we have default_variant eatch product. We can simply set it as required on model. 
@tomaszszymanski129  @maarcingebala

maarcingebala:
It could be implemented this way as well. Definitely it is easier to fetch such a variant from the `Product` model as you would already have `Product.default_variant_id` field in the instance (then we can also reuse the existing dataloader for loading variants). Not sure if there was a requirement to implement it as a boolean flag in the `ProductVariant` model?

tomaszszymanski129:
@maarcingebala there was no such requirement. It was just a way I thought of, but indeed, proposed solution is better, I'll refactoir that quickly - also logic with setting default will be simpler - no requirement of checking if any other variant has default on `True`.

patrys:
+1 from me on having this on Product level. The database schema should make it impossible to set more than one (or fewer than one) variant as default.

saleor/graphql/product/mutations/products.py
```diff
@@ -1785,6 +1788,45 @@ def perform_mutation(cls, _root, info, product_id, images_ids):
...
+            info, product_id, field="product_id", only_type=Product
+        )
+        variant = cls.get_node_or_error(
+            info,
+            variant_id,
+            field="variant_id",
+            only_type=ProductVariant,
+            qs=product.variants.all(),
+        )
+        product.default_variant = variant
```
maarcingebala, line 1824: I would add validation that checks if the variant belongs to the product, as now you can set any variant from the system.

maarcingebala (commented):
Now that we changed the relation we can easily reuse existing dataloader to load default variant. To do that we need to override the resolver in the Product type:

```
@staticmethod
def resolve_default_variant(root: models.ProductVariant, info):
    return ProductVariantByIdLoader(info.context).load(root.default_variant_id)
```

saleor/graphql/product/mutations/products.py
tomaszszymanski129: You cannot. See that the variant has `qs=product.variants.all()`, there is also a test for that `test_product_variant_set_default_not_products_variant`.

maarcingebala: Ahh missed that, nevermind then :)

saleor/product/migrations/0126_product_default_variant.py
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 3.1.1 on 2020-09-22 11:49
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def set_default_variants(apps, schema_editor):
```
tomaszszymanski129, line 7: Pay special attention to this while reviewing, this is done by the most bulletproof way which we discussed yesterday with the team.

maarcingebala: Looks good IMO

maarcingebala (approved):

maarcingebala:
Need re-review from @korycins  and @fowczarek

saleor/graphql/product/mutations/products.py
```diff
@@ -1785,6 +1788,45 @@ def perform_mutation(cls, _root, info, product_id, images_ids):
...
+        product.save(update_fields=["default_variant"])
```
fowczarek, line 1825:
```suggestion
        product.save(update_fields=["default_variant", "updated_at"])
```

saleor/graphql/product/mutations/products.py
```diff
@@ -1326,6 +1326,9 @@ def get_instance(cls, info, **data):
     @transaction.atomic()
     def save(cls, info, instance, cleaned_input):
         instance.save()
+        if not instance.product.default_variant:
+            instance.product.default_variant = instance
+            instance.product.save(update_fields=["default_variant", "updated_at"])
```
fowczarek, line 1331: Should we add `info.context.plugins.product_updated(product)`?

saleor/graphql/product/tests/test_product.py
```diff
@@ -1414,6 +1431,91 @@ def test_create_product(
...
+    )
+    product_with_two_variants.refresh_from_db()
+    assert not product_with_two_variants.default_variant
+    content = get_graphql_content(response)
+    data = content["data"]["productVariantSetDefault"]
+    assert data["productErrors"][0]["code"] == ProductErrorCode.INVALID.name
+    assert data["productErrors"][0]["field"] == "variantId"
+
+
+def test_product_variant_set_default_not_products_variant(
```
fowczarek, line 1489: I think that we should raise a different error code in this case. 
Error code `NOT_FOUND` on field `variantId` could be confusing because this variant exists in the system.

saleor/graphql/product/types/products.py
```diff
@@ -273,7 +274,14 @@ class Meta:
         description = (
             "Represents a version of a product such as different size or color."
         )
-        only_fields = ["id", "name", "product", "sku", "track_inventory", "weight"]
+        only_fields = [
```
fowczarek, line 277: In my opinion, this change is not required.

fowczarek (approved):

korycins (approved):

saleor/graphql/product/mutations/products.py
```diff
@@ -1785,6 +1788,54 @@ def perform_mutation(cls, _root, info, product_id, images_ids):
...
+            info, product_id, field="product_id", only_type=Product
+        )
+        variant = cls.get_node_or_error(
+            info,
+            variant_id,
+            field="variant_id",
+            only_type=ProductVariant,
+            qs=models.ProductVariant.objects.select_related("product"),
+        )
+        if variant.product != product:
```
fowczarek, line 1824: Maby?
```suggestion
            only_type=ProductVariant,
        )
        if variant.product_id != product.id:
```

d-wysocki (approved):

saleor/graphql/product/mutations/products.py
tomaszszymanski129: the qs saves one query as we compare resulting related object to one we already have, measured with `query_count`.

PR #6183, opened 2020-10-01
Use OAuth scopes to request and assign Saleor permission to token

korycins:
We request Saleor's scopes from OAuth provider. If given user has scope like `saleor:manage_orders`, we will assign the `MANAGE_ORDERS` permission to a token which will be used by Saleor

- add optional audience config
- use state field to pass redirect url
- map scope fields to Saleor perms


patrys:
Task linked: [SALEOR-1105 Use scope permissions](https://app.clickup.com/t/2549495/SALEOR-1105)

saleor/core/tests/test_auth_backend.py
```diff
@@ -128,29 +128,6 @@ def test_user_with_limited_permissions(
     assert set(user_permissions) == set(limited_permissions)
 
 
-def test_user_doesnt_have_permissions_from_token(staff_user, app, rf):
```
korycins, line 131: I've dropped this test as we create jwt tokens on our side, so we can trust them.  This test was checking an additional checker for confirming that `create_access_token_for_app` doesn't assign more permission than user & app has.

saleor/plugins/openid_connect/plugin.py
```diff
@@ -114,6 +116,14 @@ class OpenIDConnectPlugin(BasePlugin):
             ),
             "label": "OAuth logout URL",
         },
+        "audience": {
+            "type": ConfigurationTypeField.STRING,
+            "help_text": (
+                "The Oauth resource identifier. If provided, Saleor will define "
```
maarcingebala, line 122: I think the name is OAuth :) 
```suggestion
                "The OAuth resource identifier. If provided, Saleor will define "
```

saleor/plugins/openid_connect/plugin.py
```diff
@@ -163,9 +175,14 @@ def external_authentication(
             return previous_value
         storefront_redirect_url = data.get("redirectUrl")
         validate_storefront_redirect_url(storefront_redirect_url)
+        kwargs = {
+            "redirect_uri": build_absolute_uri(f"/plugins/{self.PLUGIN_ID}/callback"),
```
maarcingebala, line 179: In the good old Django views we used to use `reverse` function to get URLs by name, can we use it here as well?

fowczarek (approved):

saleor/plugins/manager.py
```diff
@@ -532,6 +532,13 @@ def authenticate_user(self, request: WSGIRequest) -> Optional["User"]:
         default_value = None
         return self.__run_method_on_plugins("authenticate_user", default_value, request)
 
+    def external_logout(self, data: dict, request: WSGIRequest) -> dict:
+        """Authenticate user which should be assigned to the request."""
+        default_value = {}  # type: ignore
```
IKarbowiak, line 537: Why `type: ignore` here?

IKarbowiak (approved):

saleor/plugins/manager.py
korycins: Mypy requires to provide the exact type of the dict, which plugin will return.  I can try to add some typing here.

PR #6195, opened 2020-10-02
Add external verify mutation

korycins:
The PR adds new mutation which will allow validating the external token


IKarbowiak (approved):

saleor/graphql/account/mutations/authentication.py
```diff
@@ -356,3 +356,32 @@ def perform_mutation(cls, root, info, **data):
...
+
+
+class ExternalVerify(BaseMutation):
+    user = graphene.Field(User, description="User assigned to data.")
+    is_valid = graphene.Boolean(
+        required=True,
+        default_value=False,
+        description="Determine if authentication data is valid or not.",
+    )
+    verify_data = graphene.JSONString(description="External data.")
```
fowczarek, line 368: Maybe `extra_data` is a more proper name for this field?

fowczarek (approved):

maarcingebala (approved):

saleor/graphql/account/mutations/authentication.py
korycins: I followed the same pattern for ExternalAuthentication, ExternalRefresh and ExternalVerify. If we need to change it here we should change this for all external mutations

PR #6209, opened 2020-10-05
Make TIME_ZONE and LANGUAGE_CODE configurable via env vars

maersu:

In order to make the settings more configurable, I would like to merge the following changes:

* TIME_ZONE and LANGUAGE_CODE are configurable via env vars
* Faker uses the specified language code so the dummy data are more accurate to the developers location

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/core/utils/random_data.py
+++ b/saleor/core/utils/random_data.py
@@ -69,7 +69,7 @@ from ...shipping.models import ShippingMethod, ShippingMethodType, ShippingZone
 from ...warehouse.management import increase_stock
 from ...warehouse.models import Stock, Warehouse
 
-fake = Factory.create()
+fake = Factory.create(settings.LANGUAGE_CODE)
 PRODUCTS_LIST_DIR = "products-list/"
 
 IMAGES_MAPPING = {
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -68,8 +68,8 @@ DATABASES = {
 }
 
 
-TIME_ZONE = "UTC"
-LANGUAGE_CODE = "en"
+TIME_ZONE = os.environ.get("TIME_ZONE", "UTC")
+LANGUAGE_CODE = os.environ.get("LANGUAGE_CODE", "en")
 LANGUAGES = [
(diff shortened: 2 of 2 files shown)


patrys:
Hi, `TIME_ZONE` is not meant to be changed. The API is supposed to only operate in UTC. The clients are expected to convert to the correct time zone for display/input. I know this currently affects time ranges that rely on dates alone, we're going to fix that by converting the relevant date inputs to date/time inputs.

maersu:
@patrys thanks! 

Just for my curiosity: Why can the API just operate with UTC?

patrys:
Given a UTC timestamp, you can accurately represent that point in time in any given time zone. The opposite is not true because of daylight saving times (an entire hour is duplicated whenever the clocks are shifted back).

maarcingebala:
What about LANGUAGE_CODE, do we want to set it with env variable @patrys?

patrys:
Not sure what it's used for, is it just a fallback language? If so, I think the Dashboard is a better place for it.

maarcingebala:
It's one of Django's standard configuration options: https://docs.djangoproject.com/en/3.1/ref/settings/#language-code.

maarcingebala:
We use it only for email translations as there are no other templates in Saleor now.

patrys:
Aren't emails translated based on the user's language? I know we record the language each order was placed in, not sure about signup and password reset but the user's preferred language should also be used there.

incorvia:
> Hi, the `TIME_ZONE` is not meant to be changed. The API is supposed to only operate in UTC. The clients are expected to convert to the correct time zone for display/input. I know this currently affects time ranges that rely on dates alone, we're going to fix that by converting the relevant date inputs to date/time inputs.

When Saleor's team plans to 'fix that by converting the relevant date inputs.. ' is the intention to then provide a configuration within the Dashboard to present everything in a configurable time zone?

patrys:
Yes, the dashboard will let you pick a timezone when entering a date/time (but the value will be stored as UTC in the backend).

Issue #6032, opened 2020-08-20
Missing validation for start and end dates on vouchers, sales, gift cards


cr0nex:
### What I'm trying to achieve
…

### Steps to reproduce the problem
1. Create a voucher or a sale through the dashboard, or directly through the API
2. Set the end date BEFORE the start date

### What I expected to happen
Should have prevented me from creating a voucher with incorrect dates on 3 levels:
1) Client side React of the dashboard
2) GraphQL API
3) Django models

### Screenshots

![image](https://user-images.githubusercontent.com/39736209/90797260-b4c4ce00-e310-11ea-8d53-d8aabc5ce00c.png)

**System information**
Saleor 2.10 (saleor-platform) run through Docker

Operating system:
Windows 10

### Question
This could've been partially addressed by having database-side validation. 
Is there a reason why saleor does not use Postgres built-in Range types? I know graphene support for those is patchy, but other than that?
Also, I believe DateTimeRangeField implements this type of validation out of the box.

akash1997:
Hi, can I work on this issue?

maarcingebala:
Feel free to pick it up @akash1997!

PR #6216, opened 2020-10-06
add check for end_date after start_date

akash1997:
The check is added as an assertion before the model object gets saved for the giftcard and discount models

I want to merge this change because it validates that the end tme is after the start date during creation.

#6032 

# Impact

# Pull Request Checklist

* [ ] The changes are tested


korycins:
Hi @akash1997
Thanks for this pull request!
IMHO the validation that you implementaed in `save` method should be moved to the mutations classes. You can use `clean_instance` method or call validation directly in `perform_mutation`

PR #6220, opened 2020-10-07
Add new field to shop type - availableExternalAuthentications

korycins:
Add a new field to shop type to give a possibility to discover External Authentication plugins.

### DIFF
--- a/saleor/graphql/shop/types.py
+++ b/saleor/graphql/shop/types.py
@@ -76,6 +76,11 @@ class Shop(graphene.ObjectType):
         description="List of available payment gateways.",
         required=True,
     )
+    available_external_authentications = graphene.List(
+        graphene.NonNull(graphene.String),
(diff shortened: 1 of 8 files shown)


IKarbowiak (approved):

saleor/graphql/shop/types.py
```diff
@@ -176,6 +181,10 @@ class Meta:
     def resolve_available_payment_gateways(_, _info, currency: Optional[str] = None):
         return get_plugins_manager().list_payment_gateways(currency=currency)
 
+    @staticmethod
+    def resolve_available_external_authentications(_, info):
```
fowczarek, line 185: Should this field be public?

korycins: Yeap, the same as we have public payment gateways :)

fowczarek (approved):

PR #6240, opened 2020-10-12
Remove requirements.txt and requirements_dev.txt

tomaszszymanski129:
I want to merge this change because it removes requirements.txt and requirements_dev.txt in favor of poetry, so that we use one source of requirements.

Linked task: [SALEOR-746](https://app.clickup.com/t/2549495/SALEOR-746)

What is changed here?
- `Dockerfile` to use `poetry` instead of `pip`
- `pyproject.toml` now uses python 3.9 (was 3.8 and docker image was updated at some point so this should also get bumped)
- `tox.ini` to support poetry
- `pytest` workflow for GH actions modified to use poetry
- `requirements.txt` and `requirements_dev.txt` are dropped entirely

Should be merged along with [platform update](https://github.com/mirumee/saleor-platform/pull/83) that adds an env var to install or not the dev dependencies in image. Explained more in the platform PR.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-746 Drop requirements.txt files](https://app.clickup.com/t/2549495/SALEOR-746)

Dockerfile
```diff
@@ -8,9 +8,12 @@ RUN apt-get -y update \
...
 
 # Install Python dependencies
-COPY requirements_dev.txt /app/
+COPY poetry.lock /app/
+COPY pyproject.toml /app/
 WORKDIR /app
-RUN pip install -r requirements_dev.txt
+RUN pip install poetry
+RUN poetry config virtualenvs.create false
+RUN poetry install --no-ansi
```
patrys, line 16: Do we need to force `--no-ansi` here? It seems to detect whether the controlling terminal has ANSI capabilities. I think we should add `--no-root` and possibly `--no-dev`.

tomaszszymanski129:
Tested locally, all builds and works. Will wait for travis to finish as it might need some adjustments, will mark as ready for review after that.

tomaszszymanski129:
Everything works except mypy which after update to 3.9 produces 69 type errors.

maarcingebala:
Closing this PR, since we're not working on it now, but we want to get back to this in the future.

PR #6244, opened 2020-10-12
Use is_visible to determine if object should be published

korycins:
Add resolvers for is_publish to calculate result based on db values - `is_published` and `publication_date`


saleor/graphql/page/types.py
```diff
@@ -8,6 +8,9 @@
 
 class Page(CountableDjangoObjectType):
     translation = TranslationField(PageTranslation, type_name="page")
+    is_published = Boolean(
+        required=True, description="Whether the collection is published."
```
IKarbowiak, line 12:
```suggestion
        required=True, description="Whether the page is published."
```

korycins: Thanks! 👮‍♀️

IKarbowiak (approved):

patrys:
Closed?

PR #6283, opened 2020-10-19
Drop currency from shipping_method

d-wysocki:
I want to merge this change because...I want to drop currency from shipping_method

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/order/types.py
```diff
@@ -528,23 +528,26 @@ def resolve_available_shipping_methods(root: models.Order, _info):
...
+            if shipping_channel_listing:
+                taxed_price = manager.apply_taxes_to_shipping(
+                    shipping_channel_listing.price,
+                    root.shipping_address,  # type: ignore
+                )
+                if display_gross:
+                    shipping_method.price = taxed_price.gross
+                else:
+                    shipping_method.price = taxed_price.net
+                available_shipping_methods.append(shipping_method)
```
IKarbowiak, line 550: Why do you introduce this `available_shipping_methods` list if you don't use it anywhere?

d-wysocki: Good catch, thanks!

saleor/graphql/order/types.py
```diff
@@ -528,27 +528,32 @@ def resolve_available_shipping_methods(root: models.Order, _info):
...
         channel_slug = root.channel.slug
+        if available_shipping_methods is None:
+            return []
```
IKarbowiak, line 553: Right now, there is no chance that `available_shipping_methods` will be `None` because you declare it with an empty list. Do you have a test for the situation when `available_shipping_methods` will be empty?

d-wysocki: I can remove this check because `get_valid_shipping_methods_for_order` already filter shipping method related to specific channel. Also I will add test for order without available shipping method.

IKarbowiak (approved):

maarcingebala (approved):

fowczarek (approved):

PR #6285, opened 2020-10-19
Make shipping / billing address and method required on order complete

tomaszszymanski129:
- it resolves issue with avatax plugin when there is no billing or shipping address on order and its converted from draft to normal
- it makes shipping and billing address of order required when converting from draft
- it makes shipping method of order required when converting from draft

Related ticket: [SALEOR-1264](https://app.clickup.com/t/2549495/SALEOR-1264)

Corresponding [dashboard PR](https://github.com/mirumee/saleor-dashboard/pull/784) will be created to remove the warning modal and handle validations error properly.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-1264 Order without user details - TaxError ](https://app.clickup.com/t/2549495/SALEOR-1264)

IKarbowiak (approved):

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -255,6 +257,18 @@ def update_user_fields(cls, order):
             except User.DoesNotExist:
                 order.user = None
 
+    @staticmethod
+    def avalara_active():
+        manager = get_plugins_manager()
```
maarcingebala, line 262: The plugin manager is available in the request context: `info.context.plugins`, we should use it to save a database query which `get_plugins_manager` makes.

fowczarek (approved):

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -255,6 +257,18 @@ def update_user_fields(cls, order):
             except User.DoesNotExist:
                 order.user = None
 
+    @staticmethod
+    def avalara_active():
```
korycins, line 261: - We shouldn't add solutions specific to a single plugin to core code. 
- Saleor should not know what avalara is.
- The saleor's responsibility is to call the expected tax method on PluginManager.
- It is plugin responsibility to handle each specific case. We don't want to have inside the Saleor code the `if else` cases related to calls on specific plugins.

korycins (changes requested):

saleor/graphql/order/mutations/draft_orders.py
tomaszszymanski129: Will move that to plugin, considered placing it there myself, just one question - what exactly should we do when case from the ticket happens?
1) raise TaxError in plugin, catch it in mutation and re-raise it
2) don't calculate taxes and just return `order_line.unit_price` in plugin

_I mean `_validate_order` would get patched to return `False` in case of no address and this fragment of plugin code remains:_
```
if _validate_order(order_line.order):
    return self._calculate_order_line_unit(order_line)
return order_line.unit_price
```

tomaszszymanski129:
Discussed with @korycins and @pwgryglak - on backend we shouldn't allow finalizing draft order if there is no shipping or billing address and if shipping method isn't set. Will modify this PR accordingly.

korycins (approved):
What about the changes for the dashboard?

tomaszszymanski129:
@korycins on it, mentioned in description that it will be handled shortly after this is merged.

fowczarek:
@tomaszszymanski129  Are staff user able to create and finalize digital draft order? (Without shipping address and shipping method)?

tomaszszymanski129:
@fowczarek they are not. Only if they fill all of these.

tomaszszymanski129:
@fowczarek Added what you suggested, now it looks we've covered all the cases.
To sum it up:
- if shipping is required, shipping address and method will be validated
- if shipping is not required, adress won't be validated
- if shipping is not required, method won't be validated (but will be if provided)
- billing address is always validated, that fixes avatax issue

fowczarek (approved):

maarcingebala (approved):

saleor/graphql/order/utils.py
```diff
@@ -98,8 +128,12 @@ def validate_draft_order(order, country):
 
     Returns a list of errors if any were found.
     """
+    validate_billing_address(order)
     if order.is_shipping_required():
-        validate_shipping_method(order)
+        validate_shipping_address(order)
```
korycins, line 133: Not sure of this `if..else` 
If the order doesn't require shipping, then why do we validate shipping_method?

korycins: @tomaszszymanski129  IMHO, the parameter `required` is not needed. this `if else` is also not needed.
```
if order.is_shipping_required():
    validate_shipping_address(order)
    validate_shipping_method(order)
```
If order requires the shipping then we will validate it.

korycins (changes requested):

korycins (approved):

PR #6297, opened 2020-10-20
Move csv emails to plugin interface

korycins:
Migrate csv emails to AdminEmailPlugin

- Saleor defines the type of the notify events like `CSV_PRODUCT_EXPORT_SUCCESS`
- Saleor prepares the payload and pass it to the PluginManager

Note: Please do not take into consideration the logic from `saleor/plugins/admin_plugins/*`, most of the code from there will be rebuild.


IKarbowiak (approved):

saleor/plugins/admin_email/tests/test_csv_emails.py
```diff
@@ -0,0 +1,87 @@
+#
+# @mock.patch("saleor.csv.emails.send_templated_mail")
```
fowczarek, line 2: Maybe instead of comment-out this test we can use `@pytest.mark.skip`?

fowczarek (approved):

saleor/plugins/admin_email/tests/test_csv_emails.py
korycins: Not sure if this will solve the problem, the mock paths have been changed.

fowczarek: If you add `skip` on those tests, Pytest didn't validate mocked paths.

saleor/plugins/admin_email/emails.py
```diff
@@ -0,0 +1,25 @@
+from .tasks import (
+    send_email_with_link_to_download_file,
+    send_export_failed_info,
+    send_set_staff_password_email_task,
+)
+
+
+def handle_account_set_staff_password(payload):
```
maarcingebala, line 8: I would suggest a different naming here, because if you read this function name, "handle" doesn't say what the function does. Couldn't we just have `send_set_staff_password_email`? The word "account" can be omitted, I think it's obvious.

saleor/csv/notifications.py
```diff
@@ -0,0 +1,40 @@
...
+        "user_email": user_email,
+        "app_id": app_id,
+        "status": export_file.status,
+        "message": export_file.message,
+        "created_at": export_file.created_at,
+        "updated_at": export_file.updated_at,
+    }
+
+
+def send_email_with_link_to_download_file(export_file: "ExportFile"):
```
maarcingebala, line 27: A little naming inconsistency - here we use "email" in the name, while the module is called "notification". If I understand we can implement all sort of notification and emails are just one type of them, is that correct? Then I would call these function in the same manner like `send_export_download_link_notification` or `notify_export_download_link`.

maarcingebala: Also we should add some docstrings.

maarcingebala (approved):

PR #6319, opened 2020-10-25
Added support for queries & mutations from external sources

LeOndaz:
Okay so I was creating a plugin and I wanted to add product reviews to the schema to allow the user to search for reviews by product and I wanted to merge this in the current /graphql/ instead of creating a new custom endpoint for my resolvers

### DIFF
--- a/saleor/graphql/api.py
+++ b/saleor/graphql/api.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from graphene_federation import build_schema
 
 from .account.schema import AccountMutations, AccountQueries
@@ -21,8 +22,7 @@ from .translations.schema import TranslationQueries
 from .warehouse.schema import StockQueries, WarehouseMutations, WarehouseQueries
 from .webhook.schema import WebhookMutations, WebhookQueries
 
-
-class Query(
+QUERY_CLASSES = (
     AccountQueries,
     AppQueries,
     CheckoutQueries,
@@ -42,11 +42,9 @@ class Query(
     TranslationQueries,
     WarehouseQueries,
     WebhookQueries,
-):
-    pass
+)
 
-
-class Mutation(
+MUTATION_CLASSES = (
     AccountMutations,
     AppMutations,
     CheckoutMutations,
@@ -65,8 +63,12 @@ class Mutation(
     ShopMutations,
     WarehouseMutations,
     WebhookMutations,
-):
-    pass
+)
 
+QUERY_CLASSES += getattr(settings, "GRAPHQL_QUERY_CLASSES", tuple())
+MUTATION_CLASSES += getattr(settings, "GRAPHQL_MUTATION_CLASSES", tuple())
 
-schema = build_schema(Query, mutation=Mutation)
+QueryClass = type("Query", QUERY_CLASSES, {})
+MutationClass = type("Mutation", MUTATION_CLASSES, {})
+
+schema = build_schema(QueryClass, mutation=MutationClass)


LeOndaz:
I could actually make the `QUERY_CLASSES` and `MUTATION_CLASSES` inside settings.py and allow the user to remove queries/mutations from the API depending on his needs. You may want to say your opinion about this, instead of putting the `QUERY_CLASSES` and `MUTATION_CLASSES` in api.py, we could move them to settings.py

maarcingebala:
Extending the main GraphQL schema by plugins is definitely something that we want to support in Saleor. What you suggest would work, but in my opinion, developers shouldn't change settings to extend the schema per plugin. In settings, you should only define what plugins are installed/enabled. Then we should add a new hook in the plugin manager that would load the plugin's schema and add it to the main Saleor's schema. How it could work:

- we could assume that plugin's schema should always live in the `schema.py` file in the plugin's main directory. This file would define queries and mutations specific to that plugin.
- in module `saleor/graphql/api.py`, after the main schema is initialized, we should iterate over enabled plugins and discover their `schema.py` files and load operations that are defined there. Probably there should be some utility function that would handle merging the schemas.

That's just a rough idea, but I think this is how we would like to implement it eventually. What you suggest will work for you @LeOndaz and the solution is fine, as it doesn't add many changes to Saleor, so it will be easier to upgrade to newer versions in the future if you decide to use it.

@korycins Take a look at this when you have time as you have more experience with plugins than I do ;)

LeOndaz:
Well, both do work, I think you're may be right about "developers shouldn't change settings....." and I agree with you,  we agreed that the idea is to be implemented anyway (ignoring "how" for now), I'll be waiting for us to reach an agreement and for @korycins to reply then I'll implement what we agree on as I don't like to alter with saleor in my project, I usually run git pull tags/<tag> to update it, I don't want the files to conflict with it (probably everyone who uses saleor shouldn't play with it's core files), but plugins aren't really that useful without this feature

korycins:
Hey @LeOndaz ,
Just letting you know about this as I am not sure if you're already familiar - https://docs.saleor.io/docs/next/developer/extending/plugins#webhook-handler


> * we could assume that plugin's schema should always live in the `schema.py` file in the plugin's main directory. This file would define queries and mutations specific to that plugin. 
> * in module `saleor/graphql/api.py`, after the main schema is initialized, we should iterate over enabled plugins and discover their `schema.py` files and load operations that are defined there. Probably there should be some utility function that would handle merging the schemas.

IMHO the best way would be extending the `BasePlugin` and add methods responsible for serving additional plugin's queries and mutations. These methods should be called by the PluginManager to extend the Saleor API. 
The easiest way would be, extending Saleor's scheme by all plugins API, no matter if the plugin is active or not. 
If we would like to include only the API for active plugins this will require some scheme rebuilds in real-time as the plugin can be activated/deactivated from the dashboard. 

@patrys  What do you think?

patrys:
The biggest challenge I see is the need to rebuild the schema each time the list of active plugins changes as those can be enabled and disabled. There's no point in rebuilding the schema for each individual request but also no obvious way to invalidate an already built one in a multi-server environment.

LeOndaz:
I think using this method of webhook method is somewhat hacky, I agree with auto detecting schema.py of active plugins + installed apps, Sounds like the right way to do it for me, better ways can come in the future and this specific way can be deprecated by then as it's not something to worry about in the future, but extending is the schema is really needed. Building the schema when a plugin active status changes in real-time is challenging and should be somewhat slow enough to annoy some users, this is why I think it should be a compile time operation, only when the server runs, it builds it, if someone disabled a plugin, basically he should restart the server, what do you think ?

patrys:
I don't think the server should ever need to be restarted. Saleor was built with ease of containerization and horizontal scalability in mind. Some deployments will run dozens of Docker images, some may run hundreds. I'm also not convinced that all use cases need the schema to change. I could see a plugin callback to synthesize a key-value structure similar to the metadata that would already be part of Saleor's schema. If you absolutely need to modify the schema, there's always [Apollo Federation](https://www.apollographql.com/docs/federation/implementing-services/) which is already supported by our API.

maarcingebala:
@patrys Do you something like this?
```graphql
type Mutation {
    pluginCallback(pluginId: ID!, data: JSON): JSON
}
```
I think this should do the job for a lot of plugins.

patrys:
@maarcingebala I was thinking about something like this for queries:

```graphql
query Foo {
  product(id: "...") {
    name
    generatedFields {
      key
      value
    }
  }
}
```

For mutations, we could indeed allow something like that:

```graphql
mutation Foo {
  result: callPlugin(plugin: "...", data: "...")
}
```

LeOndaz:
@patrys I think federation is a good solution but it's an overkill just for a plugin that adds a basic thing like reviews, but anyway I used it, have a look at #6439

LeOndaz:
not stale

LeOndaz:
@patrys Months had passed, I'd like to ask if this solution is still relevant, in case I wanted to implement it

patrys:
As I've said above, I'm not too fond of the idea of altering the schema of a running service. If needed, this can always be achieved using GraphQL federation. For other cases, I'd prefer to analyze each problem individually rather than try to find a blanket solution to cover all possible use cases.

LeOndaz:
Okay then, I'll put some attention on #6439

LeOndaz:
Btw I was talking about this

> @maarcingebala I was thinking about something like this for queries:
> 
> ```graphql
> query Foo {
>   product(id: "...") {
>     name
>     generatedFields {
>       key
>       value
>     }
>   }
> }
> ```
> 
> For mutations, we could indeed allow something like that:
> 
> ```graphql
> mutation Foo {
>   result: callPlugin(plugin: "...", data: "...")
> }
> ```

PR #6320, opened 2020-10-26
Add possibility to add external reference to order

korycins:
- Add possibility to provide external payment ID during the conversion draft order to order


saleor/order/actions.py
```diff
@@ -216,6 +217,17 @@ def mark_order_as_paid(order: "Order", request_user: "User"):
...
+        payment=payment,
+        action_required=False,
+        kind=TransactionKind.EXTERNAL,
+        token=external_reference or "",
+        is_success=True,
+        amount=order.total.gross.amount,
+        currency=order.total.gross.currency,
+        searchable_key=external_reference or "",
+        gateway_response={},
+    )
```
IKarbowiak, line 230: Maybe we should create this only when `external_reference` is provided?

maarcingebala (approved):

fowczarek (approved):

saleor/order/actions.py
korycins: @IKarbowiak I was thinking about this, but this will introduce the inconsistency in the database. The order with manual capture but without the reference will not have the transaction, but the order with manual capture and reference will have this.
IMO If we want to record the transaction process we should do that for each case. 
cc @maarcingebala @fowczarek

patrys:
Task linked: [SALEOR-1385 Reference for external payments](https://app.clickup.com/t/2549495/SALEOR-1385)

krzysztofwolski:
Ref https://github.com/mirumee/saleor-dashboard/pull/837

PR #6338, opened 2020-10-27
Allow associating attribute values with pages

IKarbowiak:
* Add `Page.attributes` field to return page attributes and their values
* Allow setting attribute values for a page instance - update `pageCreate` and `pageUpdate` mutations

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/product/mutations/products.py
+++ b/saleor/graphql/product/mutations/products.py
@@ -13,6 +13,8 @@ from graphql_relay import from_global_id
 from ....core.exceptions import PermissionDenied
 from ....core.permissions import ProductPermissions, ProductTypePermissions
 from ....order import OrderStatus, models as order_models
+from ....page import models as page_models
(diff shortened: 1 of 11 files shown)


fowczarek (approved):

saleor/graphql/product/mutations/products.py
```diff
@@ -743,16 +747,55 @@ def _validate_input(
         else:
             return cls._check_input_for_product(cleaned_input, attribute_qs)
 
+    @classmethod
+    def _validate_page_attributes_input(
```
maarcingebala, line 751: This function is the same as `_check_input_for_product`. Maybe we could just move the logic from this function to `validate_attributes_input_for_product_and_page`? It looks like some validation is done here and some in the utils functions. It would be better to have validation in one place. You can take a look if we can fix it somehow.

maarcingebala (approved):

PR #6381, opened 2020-11-02
New order status - unconfirmed

tomaszszymanski129:
I want to merge this change because it adds new order status, which is `unconfirmed`.
- Added order status `unconfirmed`.
- Site setting `automatically_confirm_all_new_orders` that defaults to `True`.
- Order that is created from checkout complete has status `unconfirmed` if `automatically_confirm_all_new_orders`
 is `False`.
- Unit test for the case obove (when `automatically_confirm_all_new_orders` is `False`, order should be `unconfirmed`)

_This is a PR to feature branch._

Related ticket: [SALEOR-1388](https://app.clickup.com/t/2549495/SALEOR-1388)

# Impact

* [X] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-1388 Add new order status - unconfirmed](https://app.clickup.com/t/2549495/SALEOR-1388)

IKarbowiak (approved):

saleor/checkout/complete_checkout.py
```diff
@@ -248,7 +250,15 @@ def _create_order(*, checkout: Checkout, order_data: dict, user: User) -> Order:
     total_price_left = order_data.pop("total_price_left")
     order_lines = order_data.pop("lines")
 
-    order = Order.objects.create(**order_data, checkout_token=checkout.token)
+    site_settings = Site.objects.get_current().settings
```
maarcingebala, line 253: Small optimization - current site is available in `request.site`, it can be used in GraphQL mutation with `info.context.site`. We could pass it as param to save duplicated database query.

tomaszszymanski129: @maarcingebala the location of that code line is to say the least.. complicated :)
in order to get to the request we would have to pass it in chain like:
`entry point` -> `complete_checkout` -> `_create_order`
and there are multiple entry points (`adyen webhooks` which would have to reach even higher for it, `checkout complete` mutation)
we sould save one query but add a lot of extra mess on the way

maarcingebala: That's exactly what we do to optimize e.g. checkout API in Saleor 3.0, where a lot of low-level functions fetch the plugin manager, while it is available in `info.context` in resolvers.

patrys: Just a note, we should pass the site, not the entire request object.

fowczarek (approved):

saleor/checkout/complete_checkout.py
tomaszszymanski129: Added TODO note for now as checkouts are being refactored and this will have to be changed anyway. After further investigation that would require a lot of work (in adyen plugin) to make it site-aware just to pass it to complete_checkout.

maarcingebala (approved):

korycins (approved):

PR #6401, opened 2020-11-04
Update populatedb with Featured Products

d-wysocki:
After removing homepage collection we want to show featured products, in this case storefront will be using collection with `featured-products` slug. So I want to update populatedb with collection with `featured-products` slug.

### DIFF
--- a/saleor/static/populatedb_data.json
+++ b/saleor/static/populatedb_data.json
@@ -10309,5 +10309,15 @@
       "collection": 3,
       "channel": 1
     }
+  },
+  {
+    "model": "product.collectionchannellisting",
+    "pk": 6,
+    "fields": {
+      "publication_date": "2020-01-01",
+      "is_published": true,
+      "collection": 3,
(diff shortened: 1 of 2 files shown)


IKarbowiak (approved):

maarcingebala (approved):

saleor/static/populatedb_data.json
```diff
@@ -10236,5 +10299,15 @@
       "collection": 2,
       "channel": 2
     }
+  },
+  {
+    "model": "product.collectionchannellisting",
```
fowczarek, line 10304: Maybe this collection should be available on both channels?

fowczarek (approved):

saleor/static/populatedb_data.json
d-wysocki: Good point, thanks :)

fowczarek (approved):

PR #6432, opened 2020-11-06
Order confirmed webhook

tomaszszymanski129:
I want to merge this change because it adds a new webhook - order_confirmed.

Related ticket: [SALEOR-1406](https://app.clickup.com/t/2549495/SALEOR-1406)

_This is a PR to feature branch_

Frontend part merged in: https://github.com/mirumee/saleor-dashboard/pull/820

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-1406 Add new plugin/webhook hooks](https://app.clickup.com/t/2549495/SALEOR-1406)

IKarbowiak (approved):

fowczarek (approved):

maarcingebala (approved):

saleor/graphql/order/mutations/orders.py
```diff
@@ -519,4 +519,5 @@ def perform_mutation(cls, root, info, **data):
         order.status = OrderStatus.UNFULFILLED
         order.save(update_fields=["status"])
         events.order_confirmed_event(order=order, user=info.context.user)
+        info.context.plugins.order_confirmed(order)
```
korycins, line 522: What about the case when we automatically mark order as confirmed?

maarcingebala: When an order is created we trigger a different webhook, right? You would then get the order data and see that the status is already confirmed.

tomaszszymanski129: On order creation (from checkout) we now trigger ORDER_CREATED. This mutation here is only called when order is confirmed by it. So when setting is disabled (automatically_confirm_all_new_orders), that triggers:
ORDER_CREATED when user finish his checkout
ORDER_CONFIRMED when staff user confirms the order.
When it's enabled only ORDER_CREATED will get triggered - should it stay that way?

maarcingebala: I think it's fine right now, webhook "order confirmed" seems specific to this particular manual action, se I don't see a reason to trigger it on order creation. Information that an order is "confirmed" will be passed anyway as a status field in that "order created" webhook.

patrys: I prefer to trigger all the appropriate events. Imagine writing an app that looks for confirmed orders. You'd have to listen to both hooks just in case someone has automatic confirmation enabled.

maarcingebala: 👍

saleor/graphql/checkout/mutations.py
```diff
@@ -835,6 +835,8 @@ def perform_mutation(cls, _root, info, checkout_id, store_source, **data):
                 tracking_code=tracking_code,
                 redirect_url=data.get("redirect_url"),
             )
+        if info.context.site.settings.automatically_confirm_all_new_orders:
```
korycins, line 838: The saleor.order module has file `actions`. The file defines all mandatory order events like `order_created`, `order_cancelled`, `order_fully_paid` etc. 
We should create there the `order_confirmed` action. I assume that this function will trigger the `plugins.order_confirmed(order)` event, but what about the `OrderEvents`? Don't we want anything about it in order history?
If you add `order_confirmed` to `actions`, you could move this `if` to action `order_created` and trigger `order_confirmed` when `automatically_confirm_all_new_orders` is enabled.
For `OrderConfirm` mutation you could only trigger the action -`order_confirmed`

saleor/graphql/checkout/mutations.py
```diff
@@ -835,6 +836,8 @@ def perform_mutation(cls, _root, info, checkout_id, store_source, **data):
                 tracking_code=tracking_code,
                 redirect_url=data.get("redirect_url"),
             )
+        if info.context.site.settings.automatically_confirm_all_new_orders and order:
```
korycins, line 839: there is an action `order_created` which is called when the order is created. IMO this part should be there

tomaszszymanski129:
@korycins please re-review, after adding the action to order_created i've seen there are some edge-cases i've missed - like when order is created from draft. That is now fully covered and tested.

PR #6433, opened 2020-11-06
Added GitHub Workflow publishing container images

NyanKiyoshi:
Tags: https://github.com/mirumee/saleor/blob/924e16595c4b992857a8a9dd4a2880143a27e8eb/.github/workflows/publish-containers.rst

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


.github/workflows/publish-containers.yml
```diff
@@ -0,0 +1,131 @@
+name: Publish
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
```
patrys, line 6: I don't think we're using the `v` prefix any more? cc @maarcingebala

maarcingebala: We're not. Can we change it here or should we start using the `v` prefix for future releases instead? I don't have much preference, but the format without the prefix seems a bit cleaner to me.

NyanKiyoshi: I will change it no problem 👍

patrys (approved):

maarcingebala (approved):

PR #6444, opened 2020-11-09
Plugin interface - add fields for configuring smtp server

korycins:
- Add email configuration fields for `UserEmailPlugin`
- Add email configuration fields for `AdminEmailPlugin`
- Add tests for plugin.py file
- Add validation for EmailBackend
- Add logic to send email

Note: Do not take into consideration the missing tests for the notify_events.py and tasks.py files. Both of them will be covered  when I will add the logic for customizing the email messages


fowczarek (approved):

IKarbowiak (approved):

saleor/plugins/user_email/plugin.py
```diff
@@ -22,63 +29,61 @@
...
-    #         "help_text": "",
-    #         "label": "Account confirmation template",
-    #     },
-    #     "account-password-reset-template": {
-    #         "type": ConfigurationTypeField.SECRET,
-    #         "help_text": "",
-    #         "label": "Account password reset template",
-    #     },
-    # }
+    DEFAULT_CONFIGURATION = [] + DEFAULT_EMAIL_CONFIGURATION  # type: ignore
```
tomaszszymanski129, line 57: `DEFAULT_EMAIL_CONFIGURATION` is a list already, why do we need `[] +`?

saleor/plugins/user_email/plugin.py
```diff
@@ -22,63 +29,61 @@
...
+    CONFIG_STRUCTURE = {}
+    CONFIG_STRUCTURE.update(DEFAULT_EMAIL_CONFIG_STRUCTURE)
```
tomaszszymanski129, line 59:
```suggestion
    CONFIG_STRUCTURE = dict(DEFAULT_EMAIL_CONFIG_STRUCTURE)
```

saleor/plugins/user_email/plugin.py
```diff
@@ -22,63 +29,61 @@
...
+        self.config = EmailConfig(
+            host=configuration["host"],
+            port=configuration["port"],
+            username=configuration["username"],
+            password=configuration["password"],
+            sender_name=configuration["sender_name"],
+            sender_address=configuration["sender_address"],
+            use_tls=configuration["use_tls"],
+            use_ssl=configuration["use_ssl"],
+        )
```
tomaszszymanski129, line 73: Keys of configuration dict are exact match of kwarg names passed to `EmailConfig`, are there any other keys in configuration? if not, we can pass it as
`self.config = EmailConfig(**configuration)`?

tomaszszymanski129 (approved):

saleor/plugins/user_email/plugin.py
korycins: We want to create a new list here, not copy the reference of the DEFAULT one.  In next PRs this list will not be empty :D

saleor/plugins/user_email/plugin.py
korycins: @tomaszszymanski129  I will skip this suggestion - In the next PR I will add the templates input which will cause, that I will have to add .update() again.

PR #6447, opened 2020-11-10
Squash migrations in product module

d-wysocki:
I want do squash all migration in product module related to multichannel including

- Product
- ProductVariant
- Collections


IKarbowiak (approved):
Looks good 👍 , but I'm not sure if we need these comments with names of previous migrations. The migrations won't exist anymore, so IMO it will be useless. And one more thing, sometimes you use migration name like: `product_variant_chanel_listings` and sometimes like `productchannellisting`. It would be nice to change it to one format. 👮‍♀️

d-wysocki:
I will remove comments after code review and I change the name of migration file.

fowczarek (approved):

PR #6465, opened 2020-11-12
Authorize net integration

bufke:
This adds an authorize.net plugin to Saleor as a payment gateway. It assumes usage of authorize.net's [accept.js](https://developer.authorize.net/api/reference/features/acceptjs.html)'s frontend integration to get a payment token. Payment capture is based on [this code sample](https://github.com/AuthorizeNet/sample-code-python/blob/master/AcceptSuite/create-an-accept-payment-transaction.py).

This introduces one new python dependency, `authorizenet`.

Currently this is a draft that shows only the implementation of the authorize.net payment capture and doesn't yet contain tests. The intention of submitting this request as a draft is to get review before completing all features.

# Interesting things

- Authorize.net is misleading about what is private any public. In [this](https://account.authorize.net/helpCP/Account/Settings/Security_Settings/General_Settings/API_Login_ID_and_Transaction_Key.htm) support page, "The API Login ID and Transaction Key protect your payment gateway account from unauthorized use and should not be shared with anyone." but that appears false. The API Login ID is actually public and must be given to the browser for usage with their acceptjs iframe as described [here](https://developer.authorize.net/api/reference/features/acceptjs.html). It is my determination that the API Login ID is public and only the Transaction Key is private. As such, I've implemented only the transaction key as `ConfigurationTypeField.SECRET` and we pass the API Login ID in to the browser in `get_payment_config`
- The payment_information.token is referred to as opaque data in authorize.net terms. However it's clear that this token is actually a base64 string of JSON which contains a token and other fields. I see no need to consider this implementation detail outside an interesting fact. The base64 string treated blindly as a token that works with the authorizenet python package.
- Authorize.net is an XML based API. However Saleor's GatewayResponse.raw_response assumes a Python dictionary type and it must be JSON serializable. It is not technically possible to perfectly transform XML into JSON, at least without making some assumptions. As such, I convert it to a string and submit that as the raw_response which is JSON serializable. It is not a dict which means I'm lying about it's type. I'm happy to change this if desired but I would think raw_response handling XML and other formats is beyond the scope of the PR.
- Authorize.net supports multiple currencies but not at the same time in the same account. You must use one account per currency. That could impact how Saleor supports authorize.net when multi currency support lands. You might need one set of credentials per currency. Since it's not available in saleor yet anyway - I decided not to think too hard about it and supported it the same way every other plugin does with a supported currencies field. I think we can deal with this later if needed.
- The upstream authorizenet python packages raises [log spam](https://github.com/AuthorizeNet/sdk-python/issues/109). It's based on SOAP which makes me sad generally. They offer a JSON API but it doesn't work with their python package. It's also just their SOAP API run through a converter to JSON. As such, using their python package is the least bad approach in my opinion.
- Refund behaviour is not ideal. You cannot refund a transaction while it's in a "settling" status for 24 hours after capture. This results in an error when attempting to refund from the Saleor dashboard.
- list_payment_sources is not returning brand in grahpql. May be a Saleor issue?
- Until payment_info.customer_id is supported by Saleor, I cannot finish reusing credit cards. The feature as a whole is incomplete. A new customer can get the card to show up in the Authorize.net interface as a "profile" and can even list their payment sources. But they can't use the card to pay. Nor can they add additional cards.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/payment/gateways/authorize_net/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+            ):
+                error = response.transactionResponse.errors.error[0].errorText
+            else:
+                error = response.messages.message[0]["text"].text
+    else:
+        error = "Null Response"
+
+    if not transaction_id:
+        transaction_id = payment_information.token
+    return GatewayResponse(
```
korycins, line 95: Not sure if authorize.net gives such a possibility, but the `GatewayResponse` has `searchable_key` field which can be used to search assigned order from the API side. 
Example based on the Adyen implementation:
- adyen returns the payment reference
- saleor saves it as a searchable_key
- staff user takes payment reference from Adyen admin panel
- staff user copies the adyen reference to Dashboard/orders search bar
- saleor return orders related to the  provided key.

bufke: Auhth.net returns a "transId" which is a unique integer only on success. I can set that to both "transaction_id" and "searchable_key" which seems to match how the Adyen plugin does it.

On failure, there's nothing unique I can use.

saleor/payment/gateways/authorize_net/plugin.py
```diff
@@ -0,0 +1,124 @@
...
+    @require_active_plugin
+    def get_payment_config(self, previous_value):
+        config = self._get_gateway_config()
+        return [
+            {
+                "field": "api_login_id",
+                "value": config.connection_params["api_login_id"],
+            },
+            {"field": "client_key", "value": config.connection_params["client_key"]},
+        ]
```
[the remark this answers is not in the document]
bufke: I think it does if I use their API directly. Described [here](https://developer.authorize.net/api/reference/index.html#gettingstarted-section-section-header). It appears the Python SDK cannot support this. I can just implement this with requests...though it will be a odd function.

bufke:
@korycins is there a way to get this nice error message to show up?

image

I also tried this as seen in the Vatlayer plugin but it didn't work either.

```
raise ValidationError(
                {
                    "api_login_id",
                    ValidationError(
                        message, code=PluginErrorCode.PLUGIN_MISCONFIGURED.value,
                    ),
                }
            )
```

korycins:
@bufke It seems that the dashboard doesn't handle the errors related to the plugin fields.

saleor/payment/gateways/authorize_net/__init__.py
```diff
@@ -0,0 +1,425 @@
...
+    get_customer_profile = apicontractsv1.getCustomerProfileRequest()
+    get_customer_profile.merchantAuthentication = merchant_auth
+    get_customer_profile.customerProfileId = customer_id
+    controller = getCustomerProfileController(get_customer_profile)
+    controller.execute()
+
+    response = controller.getresponse()
+    results = []
+    raw_response = etree.tostring(response).decode()
+    print(raw_response)
```
korycins, line 253: print

saleor/payment/gateways/authorize_net/__init__.py
```diff
@@ -0,0 +1,425 @@
...
+    if response is not None:
+        raw_response = etree.tostring(response).decode()
+        if hasattr(response, "transactionResponse"):
+            transaction_response = response.transactionResponse[0]
+            if hasattr(transaction_response, "transId"):
+                transaction_id = transaction_response.transId.pyval
+        if response.messages.resultCode == "Ok":
+            if hasattr(response.transactionResponse, "messages"):
+                success = True
+            else:
```
korycins, line 351: elif instead of `else` and `if`

saleor/payment/gateways/authorize_net/plugin.py
```diff
@@ -0,0 +1,188 @@
...
+        {"name": "use_sandbox", "value": True},
+        {"name": "store_customers_card", "value": False},
+        {"name": "automatic_payment_capture", "value": True},
+        {"name": "supported_currencies", "value": ""},
+    ]
+
+    CONFIG_STRUCTURE = {
+        "api_login_id": {
+            "type": ConfigurationTypeField.STRING,
+            "help_text": ("Provide public Authorize.Net Login ID."),
```
korycins, line 52: It is a detail but you don't need a brackets here.

saleor/payment/gateways/authorize_net/plugin.py
```diff
@@ -0,0 +1,188 @@
...
+        """Validate if provided configuration is correct."""
+        configuration = {
+            item["name"]: item["value"] for item in plugin_configuration.configuration
+        }
+
+        api_login_id = configuration.get("api_login_id", None)
+        transaction_key = configuration.get("transaction_key", None)
+
+        # Only check when both are set, otherwise the dashboard is hard to use
+        if api_login_id and transaction_key:
```
korycins, line 119: I would check this only if the plugin is active.  Sometimes you want to modify something when the plugin is disabled. Now you're trying to check the values even if the staff user doesn't activate the plugin.

saleor/payment/gateways/authorize_net/plugin.py
```diff
@@ -0,0 +1,188 @@
...
+            success, message = authenticate_test(
+                api_login_id, transaction_key, configuration.get("use_sandbox")
+            )
+            if not success:
+                raise ValidationError(
```
korycins, line 124: The structure like this would be better:
```
                raise ValidationError(
                    {
                        "apple-pay-cert": ValidationError(
                            "The provided apple certificate is invalid.",
                            code=PluginErrorCode.INVALID.value,
                        )
                    }
                )
```
Where the `apple-pay-cert` is the defined field.
This is the standard structure of error which we use. I know there are some parts of code which the old error structure but the new part of code should use the proposed structure.

saleor/payment/gateways/authorize_net/plugin.py
```diff
@@ -0,0 +1,188 @@
...
+        return capture(payment_information, self._get_gateway_config())
+
+    @require_active_plugin
+    def refund_payment(
+        self, payment_information: "PaymentData", previous_value
+    ) -> "GatewayResponse":
+        try:
+            payment = Payment.objects.get(pk=payment_information.payment_id)
+        except Payment.DoesNotExist:
+            raise PaymentError("Cannot find Payment.")
```
korycins, line 147: I would add here the ID of the payment which cannot be found.
`f"Cannot find Payment {payment_information.payment_id}.`

saleor/payment/gateways/authorize_net/plugin.py
patrys: @korycins as long as you spell Apple with a capital "A." 😄

bufke: @korycins What about this case where it's at the form level and not a field level (because authentication failure is the result of 2 fields). In Django forms they refer to them as "non_field_errors". I could just pick on field and have it there.

bufke:
Should I ignore this? The imports added in the PR have been run through isort

```
seed isort known_third_party.............................................Failed
- hook id: seed-isort-config
- exit code: 1
- files were modified by this hook

./.isort.cfg updated.
```

saleor/payment/gateways/authorize_net/plugin.py
korycins: If the failure is related to more than one field you can still do something like this:
```
raise ValidationError(
                    {
                        "api-public-key": ValidationError(
                            "Provided keys are incorrect..",
                            code=PluginErrorCode.INVALID.value,
                        ),
                       "api-private-key": ValidationError(
                            "Provided keys are incorrect..",
                            code=PluginErrorCode.INVALID.value,
                        )
                    }
                )
```
The message is a little bit generic, but as a response, you return fields responsible for failure. 


You can still return the generic error if you don't know what fields are incorrect or missing.

korycins:
> Should I ignore this? The imports added in the PR have been run through isort
> 
> ```
> seed isort known_third_party.............................................Failed
> - hook id: seed-isort-config
> - exit code: 1
> - files were modified by this hook
> 
> ./.isort.cfg updated.
> ```

It seems that isort found a missing package in own config file (I assume it is authorize.net) and it wants to update the file. Do you use `pre-commit`? This should fail before executing the `git commit` and should update the file. 

`pre-commit install`

bufke:
I want to check if I'm using list_client_sources right.

I can set this CustomerSource object`CustomerSource(id=1841309241, gateway='authorize.net', credit_card_info=PaymentMethodInfo(last_4='1111', exp_year=None, exp_month=None, brand='Visa', name='John Doe', type=None))`

But when I view it in graphql, I only get some of the data

```
{
  me {
    storedPaymentSources {
      gateway
      creditCardInfo {
        brand
        lastDigits
      }
    }
  }
}
```

yields

```
{
  "data": {
    "me": {
      "storedPaymentSources": [
        {
          "gateway": "authorize.net",
          "creditCardInfo": {
            "brand": "",
            "lastDigits": "1111"
          }
        }
      ]
    }
  }
}
```

Brand seems supported but doesn't show up. I'd expect it to say Visa

korycins:
> I want to check if I'm using list_client_sources right.
> 
> I can set this CustomerSource object`CustomerSource(id=1841309241, gateway='authorize.net', credit_card_info=PaymentMethodInfo(last_4='1111', exp_year=None, exp_month=None, brand='Visa', name='John Doe', type=None))`
> 
> But when I view it in graphql, I only get some of the data
> 
> ```
> {
>   me {
>     storedPaymentSources {
>       gateway
>       creditCardInfo {
>         brand
>         lastDigits
>       }
>     }
>   }
> }
> ```
> 
> yields
> 
> ```
> {
>   "data": {
>     "me": {
>       "storedPaymentSources": [
>         {
>           "gateway": "authorize.net",
>           "creditCardInfo": {
>             "brand": "",
>             "lastDigits": "1111"
>           }
>         }
>       ]
>     }
>   }
> }
> ```
> 
> Brand seems supported but doesn't show up. I'd expect it to say Visa

I see that you pushed commit which can be related to this question. Is it something that you already solved?

bufke:
The brand issue is not resolved. It still shows up empty in graphql.

saleor/payment/gateways/authorize_net/tests/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+            "transaction_key": "secret",
+            "client_key": "public",
+            "use_sandbox": True,
+        },
+    )
+
+
+@pytest.fixture()
+def authorize_net_payment(payment_dummy):
+    return payment_dummy
```
korycins, line 27: I would recommend changing at least the gateway name to authorize net (now it points to dummy) which can cause a lot of trouble in the future development

korycins (approved):

bufke:
Rebased. Ready to merge IMO. We discussed on slack the possibility of a second Pull Request with some improvements including fixture names, unsettled transaction status, and cross referencing between the Saleor Dashboard and Authorize.net web admin.

saleor/payment/gateways/authorize_net/plugin.py
```diff
@@ -0,0 +1,208 @@
...
+        },
+        "use_sandbox": {
+            "type": ConfigurationTypeField.BOOLEAN,
+            "help_text": "Determines if Saleor should use Authorize.Net sandbox environment.",
+            "label": "Use sandbox",
+        },
+        "store_customers_card": {
+            "type": ConfigurationTypeField.BOOLEAN,
+            "help_text": "Determines if Saleor should store cards on payments "
+            "in Stripe customer.",
```
korycins, line 80: `Stripe` -> `Authorize.net`

maarcingebala (approved):

maarcingebala:
Black is failing on CI. We bumped it some time ago to version `20.8b1` and we should format these changes with this version. The easiest way to make sure that all linters are passing is to run them locally through `pre-commit` ([link](https://docs.saleor.io/docs/developer/community/contributing#coding-style)).

fowczarek (approved):

Issue #6439, opened 2020-11-08
Apollo federation causes a data graph is missing a valid config error


LeOndaz:
Well, I'm trying to use apollo federation with saleor to extend the schema.

steps to reproduce this: create a server like this, I've a plugin that listens for graphql, (ignore, useless for the purpose of this)
```
import {ApolloServer} from 'apollo-server'
import {ApolloGateway} from '@apollo/gateway'

const SALEOR_SERVICE: string = 'http://localhost:8000/graphql/';
const REVIEWS_SERVICE: string = 'http://localhost:8000/plugins/reviews.plugin/';


const gateway = new ApolloGateway({
    serviceList: [
        { name: 'Saleor', url: SALEOR_SERVICE },
        { name: 'Reviews Plugin', url: REVIEWS_SERVICE },
    ],
});

const server = new ApolloServer({
    gateway,
    subscriptions: false
});

server.listen(8001).then(({ url }) => {
  console.log(`🚀 Server ready at ${url}`);
});
```
start your server 
```
npm start
```
The error is 
```
� Server ready at http://localhost:8001/
Encountered error when loading Saleor at http://localhost:8000/graphql/: Syntax Error: Unexpected Name "ObjectWithMetadata"
This data graph is missing a valid configuration. Type Query must define one or more fields.
```
Even though the saleor server /graphql/ itself is running perfectly, this happens when it's used as a federation service. I think the schema is not 100% federation valid, should be easily solvable.

maarcingebala:
Thanks for reporting @LeOndaz, would you be able to send us a PR to fix it since you have Saleor up and running with federation? If not, we'll fix it but I'm afraid not too soon, the team is pretty busy with multichannel now. I'll add it to our backlog anyway.

LeOndaz:
I didn't get it working and I used another workaround, but I've debugged the problem for hours and I've got everything we need to start solving it.
By querying the 
```
{
  _service {
      sdl
   }
{
```
You can find that all ampersands (&) are replaced with commas (old version of graphql)
1) A wide guess, graphene federation repo is unmaintained for 1.5 years and probably it's the one causing this. this is what made me read the source code of the library to see if it's causing any errors, maybe the `get_sdl` function returns a wrong sdl, actually the library is working like a charm, it does what it's supposed to do, the schema is received in graphene federation with commas and this implies that something external made this.
[...]

PR #6467, opened 2020-11-12
Fixed graphene fideration #6439

LeOndaz:


LeOndaz:
Well, I forgot to format with black

LeOndaz:
Think of this as a temporary solution till we use bump graphene to v3 (requires lots of changes to the core)

LeOndaz:
I think this fixed the problem from saleor side, but when I try to extend a model in my plugin say by doing
```
@extend(fields='id')
class Product(CountableDjangoObjectType): # or just ObjectType
    id = external(graphene.ID())

    class Meta:
        model = ProductModel
```
I get "AssertionError: Found different types with the same name in the schema: Product, Product."
I think the graphql core v2 itself doesn't support federation, or what am I missing

LeOndaz:
Let's close this for now, I'll wait and try to bump graphql core to v3

PR #6470, opened 2020-11-13
Add generic FileUpload mutation

IKarbowiak:
Add generic mutation for file upload, save file in default storage.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/core/error_codes.py
```diff
@@ -22,3 +22,9 @@ class TranslationErrorCode(Enum):
     GRAPHQL_ERROR = "graphql_error"
     NOT_FOUND = "not_found"
     REQUIRED = "required"
+
+
+class UploadErrorCode(Enum):
+    GRAPHQL_ERROR = "graphql_error"
+    NOT_FOUND = "not_found"
```
maarcingebala, line 29: Is there a case when the `NOT_FOUND` error code would be returned for this mutation? Same for `REQUIRED`, since we don't have any validation that would use it. Fields are required at the schema level.

maarcingebala (approved):

saleor/core/error_codes.py
IKarbowiak: No, I dropped those error codes.

saleor/graphql/core/mutations.py
```diff
@@ -618,3 +621,31 @@ class Meta:
...
+            "https://github.com/jaydenseric/graphql-multipart-request-spec"
+        )
+        error_type_class = UploadError
+        error_type_field = "upload_errors"
+
+    @classmethod
+    @staff_member_or_app_required
+    def perform_mutation(cls, _root, info, **data):
+        file_data = info.context.FILES.get(data["file"])
+        path = default_storage.save(file_data._name, file_data.file)
```
tomaszszymanski129, line 647: [django docs about storage save](https://docs.djangoproject.com/en/3.1/ref/files/storage/#django.core.files.storage.Storage.save) states that it returns name of saved file not a path, is it a name or a path?

tomaszszymanski129 (approved):
Looks good, it would be good to smoke-test it with S3 as storage if there is time.

saleor/graphql/core/mutations.py
IKarbowiak: Good point, I was suggesting by [this](https://docs.djangoproject.com/en/3.1/topics/files/#storage-objects), where the returned value is assigned to `path` variable...

IKarbowiak: But, the value returned by saving can be used to open the file from storage, but cannot be open by URL returned by `url` [method](https://docs.djangoproject.com/en/3.1/ref/files/storage/#django.core.files.storage.Storage.url). I'll check it with S3 as storage.

IKarbowiak: I checked with S3 and it works correctly.

PR #6477, opened 2020-11-16
Added support for querying attributes by slug

LeOndaz:
demonstration
```
query attrBySlug() {
  attribute(slug: "brand"){
    id
    name
    values{
      name
    }
  }
}
```

### DIFF
--- a/saleor/graphql/attribute/resolvers.py
+++ b/saleor/graphql/attribute/resolvers.py
@@ -6,3 +6,7 @@ def resolve_attributes(info, qs=None, **_kwargs):
     requestor = get_user_or_app_from_context(info.context)
     qs = qs or models.Attribute.objects.get_visible_to_user(requestor)
     return qs.distinct()
+
+
+def resolve_attribute_by_slug(slug):
+    return models.Attribute.objects.filter(slug=slug).first()
--- a/saleor/graphql/attribute/schema.py
+++ b/saleor/graphql/attribute/schema.py
@@ -13,7 +13,7 @@ from .mutations import (
     AttributeValueDelete,
     AttributeValueUpdate,
 )
-from .resolvers import resolve_attributes
+from .resolvers import resolve_attribute_by_slug, resolve_attributes
 from .sorters import AttributeSortingInput
 from .types import Attribute
 
@@ -27,17 +27,18 @@ class AttributeQueries(graphene.ObjectType):
     )
     attribute = graphene.Field(
         Attribute,
-        id=graphene.Argument(
-            graphene.ID, description="ID of the attribute.", required=True
-        ),
+        id=graphene.Argument(graphene.ID, description="ID of the attribute."),
+        slug=graphene.Argument(graphene.String, description="Slug of the attribute."),
         description="Look up an attribute by ID.",
     )
 
     def resolve_attributes(self, info, **kwargs):
         return resolve_attributes(info, **kwargs)
 
-    def resolve_attribute(self, info, id):
-        return graphene.Node.get_node_from_global_id(info, id, Attribute)
+    def resolve_attribute(self, info, id=None, slug=None):
+        if id:
+            return graphene.Node.get_node_from_global_id(info, id, Attribute)
+        return resolve_attribute_by_slug(slug=slug)
 
 
 class AttributeMutations(graphene.ObjectType):
--- a/saleor/graphql/attribute/tests/queries/test_attribute_query.py
+++ b/saleor/graphql/attribute/tests/queries/test_attribute_query.py
@@ -34,6 +34,30 @@ def test_get_single_attribute_by_id_as_customer(
     assert content["data"]["attribute"]["slug"] == color_attribute_without_values.slug
 
 
+def test_get_single_attribute_by_slug_as_customer(
+    user_api_client, color_attribute_without_values
+):
+    attribute_gql_slug = color_attribute_without_values.slug
+    query = """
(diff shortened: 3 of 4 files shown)


maarcingebala:
Thanks for opening the PR! The code looks good but we need two more things:
- GraphQL schema needs to be updated and committed, you can update it locally with `npm run build-schema`.
- Add a test for fetching an attribute by slug.

saleor/graphql/product/schema.py
```diff
@@ -231,8 +231,12 @@ class ProductQueries(graphene.ObjectType):
     def resolve_attributes(self, info, **kwargs):
         return resolve_attributes(info, **kwargs)
 
-    def resolve_attribute(self, info, id):
-        return graphene.Node.get_node_from_global_id(info, id, Attribute)
+    def resolve_attribute(self, info, id=None, slug=None):
+        validate_one_of_args_is_in_query("id", id, "slug", slug)
+        if id:
+            return graphene.Node.get_node_from_global_id(info, id, Product)
+        if slug:
```
korycins, line 238: IMO the second if is no needed. You validated the args in 235, which means that you have id or slug.
```
if id:
    return graphene.Node.get_node_from_global_id(info, id, Product)
return resolve_attribute_by_slug(slug=slug)
```

maarcingebala: @korycins You're right, but this solution is consistent with our codebase, we have the same structure if few more resolvers. So I think this is fine for this PR.

LeOndaz: > IMO the second if is no needed. You validated the args in 235, which means that you have id or slug.
> 
> ```
> if id:
>     return graphene.Node.get_node_from_global_id(info, id, Product)
> return resolve_attribute_by_slug(slug=slug)
> ```

I borrowed this from the other resolvers for consistency

LeOndaz: I'll make it like this
```
if id:
    return graphene.Node.get_node_from_global_id(info, id, Product)
return resolve_attribute_by_slug(slug=slug)
```
but for all the codebase, soon

korycins: 👍 


> @korycins You're right, but this solution is consistent with our codebase, we have the same structure if few more resolvers. So I think this is fine for this PR.

Sounds good 👍

LeOndaz: @korycins, @maarcingebala Can anyone review the current commit?

korycins (approved):

LeOndaz:
May I know the what will happen next?

korycins:
@LeOndaz We are waiting for @maarcingebala 's review

LeOndaz:
> @LeOndaz We are waiting for @maarcingebala 's review

Okay, thanks

maarcingebala:
@LeOndaz Looks good! The PR needs to be rebased with master as there are some conflicts, probably caused by multichannel or page types that we merged this week.

maarcingebala (approved):

LeOndaz:
I commited by mistake, don't waste time revising this, I'll also add another feature after this as there's something essential missing,

LeOndaz:
I reversed the mistaken commit, I've a question, why the attributes queries where removed from ProductQueries? as this is the main conflict

maarcingebala:
Yes, I should have mentioned that in my previous comment - attributes queries were moved to the `graphql/attribute/schema.py` file. The attributes logic is growing, as now they can be used not only with products but also with pages. To make the logic more universal and easier to maintain, we moved attributes to their own graphql module. You should move your changes from `saleor/graphql/product` to `saleor/graphql/attribute`.

LeOndaz:
Did I just run black and isort on the whole project instead of just the files I edited.. hmmm, anyway, this should be working, Review and if there're no issues and you're okay with it accept asap as I want to make another PR

maarcingebala:
These linters are still failing: `isort`, `black` and `flake8`. Make sure to use exact versions of linters that are defined in `pyproject.toml` and run them against the entire codebase.

LeOndaz:
I was busy the last week so I couldn't do this, anyway, everything is up and running and every test should be passed soon,. I'll be waiting for the merge and I'll be creating another PR soon.

maarcingebala (approved):

korycins (approved):

LeOndaz:
@korycins @maarcingebala I have 2 questions if I may, I have some intentions to PR a big update, adding reviews to Saleor system (I've already implemented most of this), your're okay with me adding it? (Asking because II know it's on your future plans). Secondly, what is the product rating included in Saleor by default? if it's just a normal rating, what's the purpose of it being a normal number (not from 1-5?)

patrys:
@LeOndaz When such large chunks of functionality are at stake, we will need to work on the design first (both UI and API). I suggest approaching the team before working on a particular implementation, so there's time to discuss the approach (e.g., what ranged to use and whether we need to define the ranges at all). A thread in GitHub discussions may be a good place to gather feedback.

For this particular PR, we'll have to wait for a bit. Our legal department recommended that we implement a CLA process, but we're still waiting for the actual CLA text for contributors to agree to.

LeOndaz:
@patrys I understand, it was easier for me to PR this than creating a plugin to handle it because graphene federation doesn't work in saleor (because of graphene v2 and not saleor), anyway thanks for replying. I'd like to know what's the product rating and why it is implemented like this (a number from 0 to whatever)

PR #6490, opened 2020-11-17
Update attribute queries with variant selection filter

IKarbowiak:
* Update `productType.variantAttributes` query with `variantSelection` argument
* Update `variant.attributes` query with `variantSelection` argument

As default `variantSelection` is set to False and all attributes are returned, when `variantSelection` is set to True, only attributes that can be used in a variant selector are returned (attribute must be product attribute and `input_type` must be `DROPDOWN`). 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/product/types/products.py
+++ b/saleor/graphql/product/types/products.py
@@ -7,7 +7,7 @@ from graphene import relay
 from graphene_federation import key
 from graphql.error import GraphQLError
 
-from ....attribute import AttributeInputType, AttributeType, models as attribute_models
+from ....attribute import models as attribute_models
 from ....core.permissions import OrderPermissions, ProductPermissions
 from ....core.weight import convert_weight_to_default_weight_unit
 from ....product import models
@@ -21,6 +21,7 @@ from ....product.utils.availability import (
     get_variant_availability,
 )
 from ....product.utils.costs import get_margin_for_variant, get_product_costs_data
+from ....product.utils.variants import get_variant_selection_attributes
 from ....warehouse.availability import (
     get_available_quantity,
     get_quantity_allocated,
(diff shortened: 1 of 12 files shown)


saleor/graphql/product/types/products.py
```diff
@@ -699,8 +724,24 @@ def resolve_product_attributes(root: models.ProductType, info):
...
+    ):
+        def apply_variant_selection_filter(attributes):
+            if not variant_selection:
+                return attributes
+            return [
+                attribute
+                for attribute in attributes
+                if attribute.input_type == AttributeInputType.DROPDOWN
+                and attribute.type == AttributeType.PRODUCT_TYPE
+            ]
```
maarcingebala, line 738: I think it would be useful to have a single function that would handle this logic. It could accept a list/qs of `Attribute` instances.

maarcingebala: We need to reuse this logic in two resolvers, also, in the future we may extend it so it's better to have it in a function.

IKarbowiak: The logic in those 2 resolvers is different, as the data structure is different. I can create a method from part which is used in `PageType`.

IKarbowiak: I introduced a function for this logic.

maarcingebala: Looks good now. Maybe just add a docstring to that function and it will be perfect :)

maarcingebala (approved):

d-wysocki (approved):

korycins (approved):

fowczarek (approved):

PR #6495, opened 2020-11-17
Handlebars for templates and custom email template

korycins:
- use handlebars instead of django template
- convert templates to use handlebars
- extend admin email plugin to provide input for custom subjects and templates 🍾 🍾 🍾 
- tests


saleor/plugins/email_common.py
```diff
@@ -105,12 +115,54 @@ class EmailConfig:
 }
 
 
-def send_email(config: EmailConfig, recipient_list, template_name, context):
+def block_trans(this, options, *args, **kwargs):
+    # TODO add translation here
```
fowczarek, line 119: Should we fix it before merge to the feature branch or it should be separate PR?

fowczarek (approved):

saleor/plugins/email_common.py
korycins: Not in this PR. I need to figure out the best way for this.

IKarbowiak (approved):

patrys (changes requested):
Please make LGTM happy. It shows 13 function calls using an unsupported argument.

korycins:
@patrys  Yes, I am aware, it is related with the changes for user_email_plugin. I was plingin to fix it in second PR where I will use the correct function calls.

PR #6506, opened 2020-11-19
Add exclude products mutation, and exclude products field

korycins:
- add mutation to include products, categories, or collections to excluded product list for shipping method
- add mutation to remove products from excluded product list for shipping method
- add new type excluded_products to shipping method


saleor/graphql/schema.graphql
```diff
@@ -2557,6 +2557,8 @@ type Mutation {
   shippingPriceBulkDelete(ids: [ID]!): ShippingPriceBulkDelete
   shippingPriceUpdate(id: ID!, input: ShippingPriceInput!): ShippingPriceUpdate
   shippingPriceTranslate(id: ID!, input: NameTranslationInput!, languageCode: LanguageCodeEnum!): ShippingPriceTranslate
+  shippingPriceExcludeProducts(id: ID!, input: ShippingPriceExcludeProductsInput!): ShippingPriceExcludeProducts
+  shippingPriceRemoveProductFromExclude(id: ID!, input: ShippingPriceRemoveProductFromExcludeInput!): ShippingPriceRemoveProductFromExclude
```
maarcingebala, line 2561: Maybe instead of the `input` object, we could have just the `products` list here? This input doesn't have any other fields.

maarcingebala (commented):
We could add a benchmark test for the `shippingPriceExcludeProducts` mutation since there is a lot of queries and it would be nice to have some record for that.

IKarbowiak (approved):

saleor/shipping/models.py
```diff
@@ -96,8 +96,14 @@ def applicable_shipping_methods_by_channel(self, shipping_methods, channel_id):
...
 
+    def exclude_shipping_methods_for_excluded_products(
+        self, qs, product_ids: List[int]
+    ):
+        """Exclude the ShippingMethods which have excluded given products."""
+        return qs.exclude(excluded_products__id__in=product_ids)
+
     def applicable_shipping_methods(
-        self, price: Money, channel_id, weight, country_code
+        self, price: Money, channel_id, weight, country_code, excluded_product_ids=None
```
maarcingebala, line 106: Let's change `excluded_product_ids` to `product_ids` and add a comment in code to explain what it does.

maarcingebala (approved):

PR #6529, opened 2020-11-25
Exclude zip codes for available shipping resolvers

tomaszszymanski129:
- it adds a new model - ShippingMethodZipCode
- it modifies resolvers for available shipping on order and checkout to exclude zip codes
- it adds logic of excluding shipping methods qs based on method that checks validity of customer zip code (the mentioned method doesn't contain algorythm for now, it will be done in separate ticket)
- CheckoutShippingMethodUpdate and OrderUpdateShipping will need to have that check but it will be done in separate ticket.

Related ticket: [SALEOR-1626](https://app.clickup.com/t/2549495/SALEOR-1626)

_This is PR to feature branch._

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-1626 Extend shipping method model to contain zip-codes](https://app.clickup.com/t/2549495/SALEOR-1626)

saleor/order/utils.py
```diff
@@ -331,6 +336,20 @@ def get_prices_of_discounted_specific_product(
     return line_prices
 
 
+def check_zip_code_in_range(code, start, end):
+    # TODO: do actual logic on checking if the zip code is in range
```
tomaszszymanski129, line 340: Will be done in separate ticket

IKarbowiak (approved):

saleor/shipping/models.py
```diff
@@ -173,6 +180,17 @@ def __repr__(self):
         )
 
 
+class ShippingMethodZipCode(models.Model):
+    shipping_method = models.ForeignKey(
+        ShippingMethod, on_delete=models.PROTECT, related_name="zip_codes"
+    )
+    start = models.CharField(max_length=32)
+    end = models.CharField(max_length=32)
```
korycins, line 188: The value which you get for the `end` can be blank/null. Not sure if the field will be able to handle it.

saleor/graphql/checkout/types.py
```diff
@@ -273,6 +277,12 @@ def calculate_available_shipping_methods(data):
                     shipping_method.price = taxed_price.gross
                 else:
                     shipping_method.price = taxed_price.net
+                if not check_shipping_method_for_zip_code(
```
korycins, line 280: IMO, this should be placed in the `applicable_shipping_methods_for_instance`. Then, It will work for order and checkout.  See the PR for excluding products from shipping_method. The case is pretty similar - [shipping/models.py](https://github.com/mirumee/saleor/pull/6506/files#diff-9afc58bf8ebfa6a87d537779f3c333d8310297ad86c01f90adb0c0c9173802c0R143)

korycins (commented):
@tomaszszymanski129  Please, double-check the number of queries which we get after your changes

saleor/shipping/models.py
```diff
@@ -173,6 +180,17 @@ def __repr__(self):
         )
 
 
+class ShippingMethodZipCode(models.Model):
+    shipping_method = models.ForeignKey(
+        ShippingMethod, on_delete=models.PROTECT, related_name="zip_codes"
```
maarcingebala, line 185: Why `PROTECT` here? I think we can remove zip codes when removing a shipping method?

saleor/checkout/utils.py
```diff
@@ -429,13 +429,15 @@ def get_valid_shipping_methods_for_checkout(
...
     country_code: Optional[str] = None,
+    prefetch_zip_codes: bool = False,
 ):
     manager = get_plugins_manager()
     return ShippingMethod.objects.applicable_shipping_methods_for_instance(
         checkout,
         channel_id=checkout.channel_id,
         price=manager.calculate_checkout_subtotal(checkout, lines, discounts).gross,
         country_code=country_code,
+        prefetch_zip_codes=prefetch_zip_codes,
```
maarcingebala, line 440: Not convinced about this pattern but don't have yet any idea how to fix it.

tomaszszymanski129: applicable_shipping_methods_for_instance should prefetch it always, as we can make check of the zip codes inside it, would be even better if that method take it into consideration

saleor/shipping/models.py
```diff
@@ -123,17 +124,28 @@ def applicable_shipping_methods_for_instance(
...
             price=price,
             channel_id=channel_id,
             weight=instance.get_total_weight(),
             country_code=country_code or instance.shipping_address.country.code,
         )
+        excluded_methods_by_zip_code = []
+        for method in applicable_methods:
+            if check_shipping_method_for_zip_code(
+                instance.shipping_address.postal_code, method
+            ):
```
maarcingebala, line 144: Did you check the performance of this piece of code? I see a few issues here:
- `instance.shipping_address` called in a loop - can result in duplicated database queries
- `check_shipping_method_for_zip_code` calls `method.zip_code_rules.all()` which again is called within a loop so again, potentially duplicated queries.

All data should be prefetched earlier (possibly in GraphQL resolvers) and passed down the function calls so we don't have any database calls in utility functions. I just want to point this out, since there are many more optimization issues on `master` now and we're working on a PR that will fix this. CC @d-wysocki.

tomaszszymanski129: @maarcingebala `instance.shipping_address` is called in a loop but we loop over `applicable_methods` instead of instances so fortunately that doesn't call extra queries. As for `method.zip_code_rules.all()` that should be better now, I've added prefetching (for now at `applicable_shipping_methods_for_instance` level) and ran the tests - queries count dropped.

maarcingebala: Ok thanks, performance is super important so we need to make sure that we pay attention to it in all our changes.

maarcingebala (approved):

PR #6551, opened 2020-11-30
Add availableShippingMethods to the Shop type

IKarbowiak:
Allow fetching `availableShippingMethods` from `Shop` type, for given `address`.

Linked task: [SALEOR-1630](https://app.clickup.com/t/2549495/SALEOR-1630)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/shop/resolvers.py
+++ b/saleor/graphql/shop/resolvers.py
@@ -1,0 +1,35 @@
+from ...core.taxes import display_gross_prices
+from ...plugins.manager import get_plugins_manager
+from ...shipping.models import ShippingMethod
+from ..channel import ChannelContext
+
+
+def resolve_available_shipping_methods(channel_slug: str, address):
+    available = ShippingMethod.objects.filter(
+        channel_listings__channel__slug=channel_slug
+    )
+    if address and address.country:
+        available = available.filter(
+            shipping_zone__countries__contains=address.country,
+        )
+
+    if available is None:
+        return []
+    manager = get_plugins_manager()
+    display_gross = display_gross_prices()
+    for shipping_method in available:
+        shipping_channel_listing = shipping_method.channel_listings.get(
+            channel__slug=channel_slug
+        )
+        taxed_price = manager.apply_taxes_to_shipping(
+            shipping_channel_listing.price, address
+        )
+        if display_gross:
+            shipping_method.price = taxed_price.gross
+        else:
+            shipping_method.price = taxed_price.net
+
+    return [
+        ChannelContext(node=shipping, channel_slug=channel_slug)
+        for shipping in available
+    ]
--- a/saleor/graphql/shop/resolvers.py
+++ b/saleor/graphql/shop/resolvers.py
@@ -1,3 +1,4 @@
+from ...account.models import Address
 from ...shipping.models import ShippingMethod, ShippingMethodChannelListing
 from ..channel import ChannelContext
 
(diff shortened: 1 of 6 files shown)


patrys:
Task linked: [SALEOR-1630 Add minimal order total for free shipping for API](https://app.clickup.com/t/2549495/SALEOR-1630)

saleor/graphql/shop/resolvers.py
```diff
@@ -0,0 +1,35 @@
+from ...core.taxes import display_gross_prices
+from ...plugins.manager import get_plugins_manager
+from ...shipping.models import ShippingMethod
+from ..channel import ChannelContext
+
+
+def resolve_available_shipping_methods(channel_slug: str, address):
+    available = ShippingMethod.objects.filter(
+        channel_listings__channel__slug=channel_slug
+    ).price_based()
```
IKarbowiak, line 10: I'm not sure if we should return only `price_based` shipping methods or all? I added `price_based` as I know it's for showing minimal order total for free shipping by the storefront. But maybe it will be misleading. @maarcingebala

saleor/graphql/shop/resolvers.py
```diff
@@ -0,0 +1,35 @@
...
+        channel_listings__channel__slug=channel_slug
+    )
+    if address and address.country:
+        available = available.filter(
+            shipping_zone__countries__contains=address.country,
+        )
+
+    if available is None:
+        return []
+    manager = get_plugins_manager()
```
maarcingebala, line 18: This makes a database query, while the manager is already prefetched and available as `info.context.plugins`. Since this function is called as a resolver we can pass the `info` object as param (or just the manager if it makes more sense).

saleor/graphql/shop/resolvers.py
```diff
@@ -0,0 +1,35 @@
...
+    display_gross = display_gross_prices()
```
maarcingebala, line 19: This information comes from `SiteSettings` object and I think it should be already prefetched as well in the info object: `info.context.site.settings.display_gross_prices`. This function doesn't seem too useful, I think we will drop it later during performance fixes.

saleor/graphql/shop/resolvers.py
```diff
@@ -0,0 +1,35 @@
...
+    for shipping_method in available:
+        shipping_channel_listing = shipping_method.channel_listings.get(
```
maarcingebala, line 21: N+1 issue. It would be better to create some dict with this data so that we don't query channel listings in a loop.

IKarbowiak: I refactored the function.

saleor/graphql/shop/resolvers.py
```diff
@@ -0,0 +1,44 @@
...
+    if available is None:
+        return []
+    manager = info.context.plugins
+    display_gross = info.context.site.settings.display_gross_prices
+    shipping_mapping = get_shipping_method_to_shipping_price_mapping(
+        available, channel_slug
+    )
+    for shipping_method in available:
+        shipping_price = shipping_mapping[shipping_method.pk]
+        taxed_price = manager.apply_taxes_to_shipping(shipping_price, address)
```
korycins, line 23: Have you checked this with avalara? I am afraid that current implementation assumes that it will get the full address data

saleor/graphql/shop/resolvers.py
korycins: `manager.apply_taxes_to_shipping(shipping_price, address)` will produce a lot of queries (until we finish the refactor started by @maarcingebala) and hits to external providers like Avalara.

saleor/graphql/shop/resolvers.py
IKarbowiak: I'm working on a fix for `Vatlayer` (`Avalara` doesn't have `apply_taxes_to_shipping` method).

maarcingebala (approved):

korycins (approved):

PR #6567, opened 2020-12-02
calculate_available_shipping_methods optimalization

d-wysocki:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -571,7 +571,7 @@ def is_valid_shipping_method(
     if not checkout.shipping_method:
         return False
     if not checkout.shipping_address:
-        return None
+        return False
 
     valid_methods = get_valid_shipping_methods_for_checkout(checkout, lines, discounts)
     if valid_methods is None or checkout.shipping_method not in valid_methods:
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -38,6 +38,7 @@ from .models import Checkout, CheckoutLine
 if TYPE_CHECKING:
     # flake8: noqa
     from ..account.models import Address
+    from prices import TaxedMoney
 
 
 def get_user_checkout(
@@ -539,7 +540,7 @@ def get_valid_shipping_methods_for_checkout(
     lines: Iterable["CheckoutLineInfo"],
     discounts: Iterable[DiscountInfo],
     country_code: Optional[str] = None,
-    subtotal=None,
+    subtotal: Optional["TaxedMoney"] = None,
 ):
     if not is_shipping_required(lines):
         return None
@@ -567,6 +568,8 @@ def is_valid_shipping_method(
     """Check if shipping method is valid and remove (if not)."""
     if not checkout.shipping_method:
         return False
+    if not checkout.shipping_address:
+        return None
 
     valid_methods = get_valid_shipping_methods_for_checkout(checkout, lines, discounts)
     if valid_methods is None or checkout.shipping_method not in valid_methods:
@@ -684,7 +687,10 @@ def cancel_active_payments(checkout: Checkout):
 def fetch_checkout_lines(checkout: Checkout) -> Iterable[CheckoutLineInfo]:
     """Fetch checkout lines as CheckoutLineInfo objects."""
     lines = CheckoutLine.objects.filter(checkout=checkout).prefetch_related(
-        "variant__product__collections", "variant__channel_listings"
+        "variant__product__collections",
+        "variant__channel_listings",
+        "variant__channel_listings__channel",
+        "variant__product__product_type",
     )
     lines_info = []
 
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -539,21 +539,23 @@ def get_valid_shipping_methods_for_checkout(
     lines: Iterable["CheckoutLineInfo"],
     discounts: Iterable[DiscountInfo],
     country_code: Optional[str] = None,
+    subtotal=None,
 ):
     if not is_shipping_required(lines):
         return None
-    if not checkout.shipping_address:
-        return None
-
-    manager = get_plugins_manager()
-    subtotal = manager.calculate_checkout_subtotal(
-        checkout, lines, checkout.shipping_address, discounts
-    )
+    # TODO: subtotal should comes from arg instead of calculate it in this function
+    # use info.context.plugins from resolver
+    if subtotal is None:
+        manager = get_plugins_manager()
+        subtotal = manager.calculate_checkout_subtotal(
+            checkout, lines, checkout.shipping_address, discounts
+        )
     return ShippingMethod.objects.applicable_shipping_methods_for_instance(
         checkout,
         channel_id=checkout.channel_id,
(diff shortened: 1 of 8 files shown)


saleor/graphql/checkout/types.py
```diff
@@ -233,6 +236,22 @@ class Meta:
         interfaces = [graphene.relay.Node, ObjectWithMetadata]
         filter_fields = ["token"]
 
+    @staticmethod
+    def resolve_shipping_address(root: models.Checkout, info):
+        return (
+            AddressByIdLoader(info.context).load(root.shipping_address_id)
+            if root.shipping_address_id
+            else None
+        )
```
IKarbowiak, line 245: Isn't this way more readable?
```suggestion
        if not root.shipping_address_id:
            return
        return AddressByIdLoader(info.context).load(root.shipping_address_id)
```

saleor/graphql/checkout/types.py
```diff
@@ -233,6 +236,22 @@ class Meta:
...
+
+    @staticmethod
+    def resolve_billing_address(root: models.Checkout, info):
+        return (
+            AddressByIdLoader(info.context).load(root.billing_address_id)
+            if root.billing_address_id
+            else None
+        )
```
IKarbowiak, line 253: Same here.

IKarbowiak (approved):

saleor/checkout/utils.py
```diff
@@ -539,21 +539,23 @@ def get_valid_shipping_methods_for_checkout(
     lines: Iterable["CheckoutLineInfo"],
     discounts: Iterable[DiscountInfo],
     country_code: Optional[str] = None,
+    subtotal=None,
 ):
     if not is_shipping_required(lines):
         return None
-    if not checkout.shipping_address:
```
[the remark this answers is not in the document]
korycins: below this line, manager calls the method on plugin which can expect that shipping_adress exists.

saleor/graphql/checkout/types.py
d-wysocki: I removed it because this doesn't affect on performance.

saleor/graphql/checkout/types.py
d-wysocki: I removed it because this doesn't affect on performance.

saleor/checkout/utils.py
```diff
@@ -682,7 +689,10 @@ def cancel_active_payments(checkout: Checkout):
 def fetch_checkout_lines(checkout: Checkout) -> Iterable[CheckoutLineInfo]:
     """Fetch checkout lines as CheckoutLineInfo objects."""
     lines = CheckoutLine.objects.filter(checkout=checkout).prefetch_related(
-        "variant__product__collections", "variant__channel_listings"
+        "variant__product__collections",
+        "variant__channel_listings",
```
korycins, line 693: Single `"variant__channel_listings__channel"` will prefetch also `variant__channel_listings`

korycins: image

maarcingebala (approved):

IKarbowiak (approved):

saleor/checkout/utils.py
```diff
@@ -565,6 +570,8 @@ def is_valid_shipping_method(
     """Check if shipping method is valid and remove (if not)."""
     if not checkout.shipping_method:
         return False
+    if not checkout.shipping_address:
+        return None
```
korycins, line 574: Why `None`?

saleor/graphql/checkout/types.py
```diff
@@ -344,26 +359,67 @@ def resolve_lines(root: models.Checkout, info):
...
+            )
             if available is None:
                 return []
-            display_gross = display_gross_prices()
-            for shipping_method in available:
-                # ignore mypy checking because it is checked in
-                # get_valid_shipping_methods_for_checkout
-                shipping_channel_listing = shipping_method.channel_listings.get(
-                    channel=root.channel
+            else:
```
korycins, line 380: else is not needed here. you're leaving the method line above.

saleor/graphql/checkout/dataloaders.py
```diff
@@ -26,54 +26,65 @@ class CheckoutLinesInfoByCheckoutTokenLoader(DataLoader):
...
-                        ]
-                    )
-                return lines_info
-
-            variants = ProductVariantByIdLoader(self.context).load_many(variants_pks)
-            products = ProductByVariantIdLoader(self.context).load_many(variants_pks)
-            collections = CollectionsByVariantIdLoader(self.context).load_many(
-                variants_pks
+            channel = ChannelByCheckoutLineIDLoader(self.context).load(
+                checkout_lines[0][0].id
```
fowczarek, line 38: What if `checkout_lines` is an empty list?

saleor/graphql/checkout/types.py
```diff
@@ -344,26 +359,66 @@ def resolve_lines(root: models.Checkout, info):
...
+            address, lines, discounts, channel = data
+            channel_slug = channel.slug
+            display_gross = info.context.site.settings.display_gross_prices
+            manager = info.context.plugins
+            subtotal = manager.calculate_checkout_subtotal(
+                root, lines, address, discounts
+            )
+            if not address:
+                return []
+            available = get_valid_shipping_methods_for_checkout(
```
fowczarek, line 371: This method can generate additional queries. Should we optimize it in this PR or separate PR?

saleor/plugins/manager.py
```diff
@@ -151,7 +151,7 @@ def calculate_checkout_subtotal(
                 line_info.product,
                 line_info.collections,
                 address,
-                checkout.channel,
+                line_info.channel_listing.channel,
```
fowczarek, line 154: This line can generate additional DB queries. Maby we should store `channel` in `line_info`

maarcingebala: I would change that in a separate PR - we need to add `channel` to `CheckoutLineInfo` dataclass and then update all other places to use it.

saleor/graphql/checkout/types.py
maarcingebala: I would say a separate PR for that. Those shipping methods are used for both checkouts and orders and if you change one of them, you need to change many more places.

saleor/graphql/checkout/dataloaders.py
maarcingebala: `variants_pks` will be empty then and the function will return in line 36.

saleor/graphql/checkout/types.py
d-wysocki: You have right but this function in big and is used in many places. We should fix it in seperate PR.

saleor/plugins/manager.py
d-wysocki: Channel is prefetched and it should not generate additional db queries.

maarcingebala: I think we should change it later anyway - instead of using Django's prefetches, it should be better to load it with data loaders - I think then we will load it only when it is necessary.

PR #6599, opened 2020-12-08
Return Products

korycins:
Sorry for the big PR 😢 

- extract from refund logic the generic logic which can be used to return.
- add return logic with possibility to make refund and replace
- add tests


saleor/order/actions.py
```diff
@@ -607,32 +594,318 @@ def create_refund_fulfillment(
...
+        fulfillment_lines_to_replace=fulfillment_lines,
+    )
+    replaced_lines = [(line.quantity, line) for line in new_order.lines.all()]
+    fulfillment_replaced_event(
+        order=order, user=requester, replaced_lines=replaced_lines,
+    )
+    order_replace_created(
+        original_order=order, replace_order=new_order, user=requester,
+    )
+    # TODO call order_returned
```
korycins, line 812: This will be handled in separate PR

saleor/order/actions.py
```diff
@@ -40,13 +50,29 @@
...
+    line: OrderLine
+    quantity: int
+    replace: bool = False
+
+
+@dataclass
+class FulfillmentLineData:
+    line: FulfillmentLine
+    quantity: int
+    replace: bool = False
```
maarcingebala, line 73: On the "checkout performance" branch we keep similar dataclasses in `order/__init__.py`, maybe we should move these as well? What do you think @fowczarek ?

saleor/order/actions.py
```diff
@@ -607,32 +594,318 @@ def create_refund_fulfillment(
...
+                    order_line,
+                )
+
+        fulfillment_returned_event(
+            order=order, user=requester, returned_lines=list(returned_lines.values()),
+        )
+    return return_fulfillment
+
+
+def process_replace(
```
maarcingebala, line 787: Docstrings would be helpful since these are "actions".

saleor/graphql/schema.graphql
```diff
@@ -3031,6 +3046,7 @@ enum OrderEventsEnum {
   CANCELED
   ORDER_MARKED_AS_PAID
   ORDER_FULLY_PAID
+  ORDER_REPLACE_CREATED
```
maarcingebala, line 3049: A noun is "replacement" so `ORDER_REPLACEMENT_CREATED`.

saleor/graphql/order/types.py
```diff
@@ -79,6 +79,9 @@ class OrderEvent(CountableDjangoObjectType):
     shipping_costs_included = graphene.Boolean(
         description="Define if shipping costs were included to the refund."
     )
+    related_order = graphene.Field(
```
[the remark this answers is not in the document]
korycins: Maybe the order is also a good idea. I was thinking that I should emphasize that it is not the order which events belong to

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -520,11 +559,156 @@ def perform_mutation(cls, _root, info, **data):
...
+        Fulfillment, description="A replace fulfillment."
+    )
+    order = graphene.Field(Order, description="Order which fulfillment was returned.")
+    replace_order = graphene.Field(
+        Order,
+        description="A draft order which was created for products with replace flag.",
+    )
+
+    class Meta:
+        description = "Refund products."
```
IKarbowiak, line 640:
```suggestion
        description = "Return products."
```

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -520,11 +559,156 @@ def perform_mutation(cls, _root, info, **data):
...
+        permissions = (OrderPermissions.MANAGE_ORDERS,)
+        error_type_class = OrderError
+        error_type_field = "order_errors"
+
+    class Arguments:
+        order = graphene.ID(
+            description="ID of the order to be refunded.", required=True
```
IKarbowiak, line 647:
```suggestion
            description="ID of the order to be returned.", required=True
```

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -520,11 +559,156 @@ def perform_mutation(cls, _root, info, **data):
...
+        )
+        input = OrderReturnProductsInput(
+            required=True,
+            description="Fields required to create an refund fulfillment.",
```
IKarbowiak, line 651: 👮‍♀️ 
```suggestion
            description="Fields required to return products.",
```

saleor/order/actions.py
```diff
@@ -428,126 +454,100 @@ def _get_fulfillment_line_if_exists(
...
-    return refund_amount
 
 
-def _refund_fulfillment_for_fulfillment_lines(
-    fulfillment_lines_to_refund: List[FulfillmentLine],
-    fulfillment_lines_quantities_to_refund: List[int],
-    already_refunded_lines: List[FulfillmentLine],
-    refunded_fulfillment: Fulfillment,
-    all_refunded_lines: Dict[int, Tuple[int, OrderLine]],
+def _move_fulfillment_lines_to_target_fulfillment(
```
IKarbowiak, line 514: The main part of `_move_order_lines_to_target_fulfillment` and `_move_fulfillment_lines_to_target_fulfillment` looks very similar, maybe some part could be extracted to separate function?

saleor/order/actions.py
fowczarek: It's will be great to have all data classes in the same place for each module.

saleor/order/actions.py
```diff
@@ -607,32 +594,318 @@ def create_refund_fulfillment(
...
+    fulfillment_lines_to_replace: List[FulfillmentLineData],
+) -> "Order":
+    """Create draft order with lines to replace."""
+    # fetch original order to not lose reference for original_order
+    replace_order = Order.objects.get(pk=original_order.pk)
+    replace_order.pk = None
+    replace_order.token = None
+    replace_order.status = OrderStatus.DRAFT
+    replace_order.created = now()
+    replace_order.save()
```
fowczarek, line 642: I'm not sure that we should copy the order in this way. 
Currently, we copy all payments, gift cards, shipping method, order weight, and total price.

fowczarek: What about the situation the new order weight is too low to send it with the original shipping method?

saleor/order/actions.py
```diff
@@ -607,32 +594,318 @@ def create_refund_fulfillment(
...
+            lines_in_target_fulfillment=already_refunded_lines,
+            target_fulfillment=refunded_fulfillment,
+        )
+
+        Fulfillment.objects.filter(order=order, lines=None).delete()
+
+    return refunded_fulfillment
+
+
+def create_replace_order(
```
fowczarek, line 629: Should this action be a transaction?

saleor/order/actions.py
```diff
@@ -607,32 +594,318 @@ def create_refund_fulfillment(
...
+    if not return_fulfillment_lines:
+        return refund_amount
+
+    order_lines_with_fulfillment = OrderLine.objects.in_bulk(
+        [line_data.line.order_line_id for line_data in return_fulfillment_lines]
+    )
+    for line_data in return_fulfillment_lines:
+        # skip lines which were already refunded
+        if line_data.line.fulfillment.status == FulfillmentStatus.REFUNDED:
+            continue
```
fowczarek, line 911: We should add a test for this case.

saleor/order/actions.py
korycins: So in that case, it shouldn't be a file like `dataclasses.py` or something? I worry that `__init__` can be like `utils.py`

saleor/order/actions.py
korycins: - payment - Payment object stores the relation for order. The new order will not have a relation to the payment from original_order
- giftcards - same thing - ManyToMany relations are stored in separate obejct
- shipping-method, order weight it should as now
- voucher- this relations should be cleaned
-  good catch with total_price and weights. I will re-calculate them

fowczarek (approved):

maarcingebala (approved):

PR #6607, opened 2020-12-11
Do not create new attribute values when instance is updated

IKarbowiak:
When the instance has assigned attribute value and in update value is not changed, don't create a new `AttributeValue` instance.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/attribute/utils.py
+++ b/saleor/graphql/attribute/utils.py
@@ -122,7 +122,10 @@ class AttributeAssignmentMixin:
 
     @classmethod
     def _pre_save_file_value(
-        cls, attribute: attribute_models.Attribute, attr_value: AttrValuesInput
+        cls,
+        instance,
+        attribute: attribute_models.Attribute,
+        attr_value: AttrValuesInput,
     ):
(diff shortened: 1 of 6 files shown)


saleor/graphql/attribute/utils.py
```diff
@@ -133,16 +136,30 @@ def _pre_save_file_value(
...
         return (value,)
 
+    @classmethod
+    def _get_assigned_attribute_value_if_exists(
+        cls, instance: T_INSTANCE, attribute: attribute_models.Attribute, file_url
+    ):
+        assignment = instance.attributes.filter(
+            assignment__attribute=attribute, values__file_url=file_url
+        ).first()
+        return None if assignment is None else assignment.values.get(file_url=file_url)
```
maarcingebala, line 161: `values.get()` could raise an error if the value is not present. I think this solution would be safer and should give the same result:
```
return assignment.values.filter(file_url=file_url).first()
```

maarcingebala (approved):

fowczarek (approved):

PR #6608, opened 2020-12-12
Drop blocktrans function and tags from default email templates

korycins:
Drop dummy `blocktrans` translation method, as we plan to add language_code to the email payloads.

### DIFF
--- a/saleor/plugins/email_common.py
+++ b/saleor/plugins/email_common.py
@@ -125,11 +125,6 @@ DEFAULT_EMAIL_CONFIG_STRUCTURE = {
 }
 
 
-def block_trans(this, options, *args, **kwargs):
-    # TODO add translation here
-    return options["fn"](this)
-
-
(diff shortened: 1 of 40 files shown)


maarcingebala (approved):

maarcingebala:
Should we also generate new translation files? A lot of translations will be removed.

patrys (approved):

patrys:
Yes, let's drop the unused translations from translation files. What remains there to be translated in Core?

korycins:
@patrys  I would prefer to drop the translation when the email plugins will be merged. 

> What remains there to be translated in Core?

I see that most of the translation will be removed. The only one part which uses the translation is the invoice plugin. IMO we could rebuild it in similar way as we did with the emails(using handlebars). Then we could drop a logic related to the django templates from Saleor (tempalte tags etc).

PR #6629, opened 2020-12-17
Feature: reserve stock allocated to authenticated user's cart

rafalp:
This PR introduces new "reservation" app with single model `Reservation` that holds reservation of quantity of product variant for given user and shipping zone together with reservation's expiration date. New mutation `reservationCreate` is introduced that allows authenticated users to create or update reservations, and checkout mutations were updated to account for reserved stock when checking variant's availability. Additionally `reservationsRemove` mutation is provided to allow users to explicitly remove reservations, and there's management command `remove_expired_stock_reservations` to remove expired reservations from DB, keeping table size more reasonable.

# Todo

* [x] Solve the deadlock issue
* [x] Rebase against master
* [x] Add `reservationCreate` mutation for when item is reserved
* [x] Add `reservationsRemove` mutation for when one or more items are removed from cart
* [x] Add reservations check to order creation mutation
* [x]  Add reservations check to `CheckoutLinesUpdate` and `CheckoutLinesAdd` mutations
* [x] Remove reservation after checkout
* [x] Remove expired reservations
* [x] Fix `reservationCreate` failing for product
* [x] Fix `reservationCreate` failing for variant ID not having node
* [x] Fix linters

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] The changes are tested


rafalp:
I've rebased the PR against master and updated the implementation to tie reservations to shipping zones and warehouse availability checks.

So `reserveStock()` requires authenticated user to provide country code in addition to quantity and product variant ID in order for reservation to work. It also runs availability check against existing stock and other users reservations.

`saleor.warehouse.availability` now takes the optional `user` argument and also includes reservations in the available quantity checks. Perhaps argument should be named `context_user` or `exclude_user` instead?

@maarcingebala I would appreciate feedback if this approach is correct before I'll continue this any further. Seems like there's still quite a few places that would need update to accommodate for new `Reservation` logic.

Thanks!

saleor/graphql/reservation/tests/test_reservation_creation.py
```diff
@@ -0,0 +1,223 @@
...
+            "countryCode": "US",
+            "quantity": 0,
+            "variantId": variant_id,
+        }
+    }
+    assert not Reservation.objects.exists()
+    response = user_api_client.post_graphql(MUTATION_RESERVATION_CREATE, variables)
+    content = get_graphql_content(response)
+    data = content["data"]["reservationCreate"]
+    assert data["reservationErrors"][0]["message"] == (
```
fowczarek, line 137: In tests, we should compare `code` and `field` instead of the message.  It's should be fixed in all tests in this PR.

fowczarek (approved):

saleor/reservation/models.py
```diff
@@ -0,0 +1,64 @@
...
+        return self.filter(expires__lte=timezone.now())
+
+
+class Reservation(models.Model):
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        related_name="reservations",
+        on_delete=models.CASCADE,
+    )
+    shipping_zone = models.ForeignKey(
```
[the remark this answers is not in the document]
rafalp: I see now that I will need a way to correlate reservations with warehouses and not shipping zones. How can I find which warehouse should I reserve stock for when using user's country? Should I find shipping zones for given country/address and then check warehouses to see which one has enough required stock to reserve?

saleor/graphql/reservation/mutations.py
```diff
@@ -0,0 +1,299 @@
...
+        if quantity < 1:
+            raise ValidationError(
+                {
+                    "quantity": ValidationError(
+                        "The quantity should be higher than zero.",
+                        code=ReservationErrorCode.ZERO_QUANTITY.value,
+                    )
+                }
+            )
+        if quantity > RESERVATION_SIZE_LIMIT:
```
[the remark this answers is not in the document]
rafalp: New reservation will replace old one, eg. you've reserved 5 units, now you create reservation for 8, final reservation will befor 8 units and not 13.

saleor/graphql/reservation/mutations.py
```diff
@@ -0,0 +1,299 @@
...
+    variants_ids = graphene.List(
+        graphene.ID,
+        description="A list of IDs of products variants reservations to remove.",
+        required=True,
+    )
+
+
+class ReservationsRemove(BaseMutation):
+    removed_reservations = graphene.List(
+        RemovedReservation, description="List of removed reservations."
```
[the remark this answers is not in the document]
maarcingebala: This is a bulk mutation and our other bulk mutations return only the number of items affected. Maybe we could use `BaseBulkMutation` or `ModelBulkDeleteMutation` here but I'm not sure how helpful each of them would be.

saleor/reservation/models.py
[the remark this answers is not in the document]
maarcingebala: I think we should extract the logic from `allocate_stocks` and share it between allocations and reservations. Also, a side note, when reservations become more warehouse-related, I think we could keep the model and logic in the `warehouse` app.

saleor/warehouse/availability.py
```diff
@@ -22,7 +24,12 @@ def _get_available_quantity(stocks: StockQuerySet) -> int:
     return max(total_quantity - quantity_allocated, 0)
 
 
-def check_stock_quantity(variant: "ProductVariant", country_code: str, quantity: int):
+def check_stock_quantity(
+    variant: "ProductVariant",
+    country_code: str,
+    quantity: int,
+    user: Optional["User"] = None,
```
maarcingebala, line 31: Again a loose idea but maybe we could decouple this module from users and just operate on stocks and reservations - instead of passing the user, we could pass a list of reservations to take into account when checking stock quantity. We wouldn't have to pass the user a few levels down. But the current implementation is also reasonable so I won't argue about that :)

saleor/graphql/reservation/mutations.py
```diff
@@ -0,0 +1,299 @@
...
+
+class ReservationCreateInput(graphene.InputObjectType):
+    country_code = CountryCodeEnum(description="Country code.", required=True)
+    quantity = graphene.Int(
+        required=True, description="The number of items to reserve."
+    )
+    variant_id = graphene.ID(required=True, description="ID of the product variant.")
+
+
+class ReservationCreate(BaseMutation):
```
korycins, line 47: Why the reservation flow is a logic separated from the checkout flow? By adding this we will have two different path of going through the checkout, right?  Why? Shouldn't we include it directly in the checkout flow and allow staff users to control if this feature should be enabled?  This is an additional checkout path to maintain.

In a few weeks, we are going to work on the possibility to extend shipping method by external services, which will need to reserve the stock on the checkout level - as we need to determine the address of "shipFrom" (warehouse address).

korycins (commented):
Also, not sure why this is a separate module.  I would say that the reservation logic should be placed in the same where the management of allocations and stocks.  IMO reservation is the first stage of preparing stock data for the order.
cc @maarcingebala  @fowczarek  what do you think?

saleor/graphql/reservation/mutations.py
korycins: cc @patrys

patrys: This will be changed. I've talked to Rafał about how the approach changed after talking to some customers and partners.

korycins: By saying `will be changed` do you mean the current feature branch or sometime in the future?

patrys: I mean "in this pull request" 😄

rafalp:
I'll be closing this PR and opening new one taking new approach as agreed with @patrys:

- Reservations will be created/updated together with checkouts.
- Reservations will live in the `warehouse.reservations` package and will share approach with `Allocation`, except they will be related to checkouts instead of orders.
- GraphQL API changes will be limited to new field on checkout returning the minimum possible time before shortest reservation expires.

Thank ya'all for feedback and suggestions!

Szomy:
Hi @rafalp!

I'm very interested where this feature heading. 
Do you have any news about this PR? It would be a "life changing" feature for my webshop. 😄

It would be super helpful if this reservation system could send some webhook events based on the reservation 🤔 I'm currently trying to solve some "external stock reservation" system.

Thank you for your hard work!

PR #6554, opened 2020-11-30
Full shiping method exclusion by zip codes backend

tomaszszymanski129:
Entire feature composed from already merged and reviewed branches.

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #6664, opened 2021-01-04
Feature: add inclusion type to ZIP code rules

Kwaidan00:
I want to merge this change because it adds an ability to define shipping method areas with inclusive postal code rules. It extends feature implemented in #6554.

With this PR backend can store and use the rules like "this shipping method is available only for the following postal code ranges".

This PR also fixes typo in the Ireland ISO-3166 code usage and typo in fixture name.

The GraphQL API was inspired by Saleor Dashboard usage of this feature; however, I'm open to change it if needed.

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/shipping/mutations/shippings.py
```diff
@@ -171,6 +171,9 @@ def success_response(cls, instance):
 class ShippingZipCodeRulesCreateInputRange(graphene.InputObjectType):
     start = graphene.String(required=True, description="Start range of the zip code.")
     end = graphene.String(required=False, description="End range of the zip code.")
+    inclusion_type = ZipCodeRuleInclusionTypeEnum(
+        required=False, description="Inclusion type for given zip code rules."
+    )
```
Kwaidan00, line 176: I've added it as non required field to not break existing Saleor Dashboard code. I can change it.

maarcingebala (approved):

maarcingebala:
The code looks good, I would just double check that this API (`shippingMethodZipCodeRulesUpdateInclusionType`) will work fine with the frontend part. It looks like it should do the job, but I haven't seen the designs of the dashboard with both inclusion and exclusion rules.

maarcingebala:
@Kwaidan00 We upgraded `Black` and `isort` versions today and your changes may require reformatting before merging.

fowczarek (approved):

IKarbowiak (approved):

saleor/graphql/shipping/mutations/shippings.py
```diff
@@ -260,6 +269,44 @@ def perform_mutation(cls, _root, info, **data):
...
+        error_type_field = "shipping_errors"
+
+    @classmethod
+    def perform_mutation(cls, root, info, **data):
+        shipping_method = cls.get_node_or_error(
+            info, data["shipping_method_id"], only_type=ShippingMethod
+        )
+        inclusion_type = data["inclusion_type"]
+
+        shipping_method.zip_code_rules.all().update(inclusion_type=inclusion_type)
```
tomaszszymanski129, line 302: Not sure if this is okay. When admin changes radio button in UI, mutation triggers. From now on all of zip codes I meant to be excluded are included. As an admin I have to delete all zips from the list as it is unlikely I would like to now include all that was excluded and vice-versa. Perhaps im overthinking it and it's not a real problem after all, what do you think?

saleor/shipping/zip_codes.py
```diff
@@ -69,23 +71,63 @@ def check_any_zip_code(code, start, end):
...
-            return True
+    zip_code_rules = method.zip_code_rules.all()
+    if not zip_code_rules:
+        return {}
+    results = {}
+    for zip_code_rule in zip_code_rules:
+        results[zip_code_rule] = check_zip_code_in_range(
+            country, postal_code, zip_code_rule.start, zip_code_rule.end
+        )
+    return results
```
tomaszszymanski129, line 96:
```
return {
    rule: check_zip_code_in_range(country, postal_code, rule.start, rule.end)
    for rule in zip_code_rules
}
```

saleor/graphql/shipping/mutations/shippings.py
maarcingebala: We need to discuss the design with @pwgryglak but I think this mutation may not be needed at all. I think on the design we should have two separate sections: "excluded zip codes" and "included zip codes". Then you would only have to specify the type of a rule when adding it.

maarcingebala: I'll mark this PR as blocked before we have an agreement.

saleor/shipping/zip_codes.py
Kwaidan00: Great thanks, I didn't notice it :)

saleor/graphql/shipping/mutations/shippings.py
Kwaidan00: Please let me know when you will decide how to handle it. As I mentioned, I've created this mutation based on current Dashboard design. 
(For me you're proposal sounds better than it @maarcingebala.)

maarcingebala:
@Kwaidan00 We looked at the designs and current API with @tomaszszymanski129 and decided that we should change the API. The problem is that currently, adding or removing a postal code rule has an instant effect i.e. UI sends a mutation that takes effect immediately. Instead, we should save the entire state in API when users click the "Save" button, to avoid intermediate states that may be incorrect. For this reason, we also don't want to introduce a mutation that this PR proposes.

Here is what we agreed about the desired API.

1. Extend the `ShippingPriceInput` with the following fields: 

```graphql
input ShippingPriceInput {
    addPostalCodeRules: [ShippingZipCodeRulesCreateInputRange!]!
    deletePostalCodeRules: [ID!]!
    inclusionType: ZipCodeRuleInclusionTypeEnum!
}
```

- `addPostalCodeRules` - list of rules to add 
- `deletePostalCodeRules` - list of rules to delete
- `inclusionType` - determines the types of currently assigned rules

This API will allow us to save all changes at once on "save" in the UI.

2. Since we're changing the API, I would use that occasion to rename occurrences of "zip code" to "postal code", since "zip code" is US-specific while "postal code" is more general.

3. Drop the following mutations: 
- shippingMethodZipCodeRulesCreate
- shippingMethodZipCodeRulesDelete

fowczarek (approved):

tomaszszymanski129 (approved):

IKarbowiak (approved):

saleor/graphql/shipping/mutations/shippings.py
```diff
@@ -40,6 +47,20 @@ class ShippingPriceInput(graphene.InputObjectType):
...
+    )
+    delete_postal_code_rules = graphene.List(
+        graphene.NonNull(graphene.ID),
+        required=True,
+        description="Postal code rules to delete.",
+    )
+    inclusion_type = PostalCodeRuleInclusionTypeEnum(
+        required=True,
+        description="Inclusion type for currently assigned postal code rules.",
+    )
```
maarcingebala, line 63: I initially suggested these fields to be required in one of previous my comments, but now I think they should be optional since the postal code rules are also optional. We should be able to create a shipping zone without passing any of those fields.

saleor/graphql/shipping/mutations/shippings.py
```diff
@@ -18,8 +18,15 @@
...
-from ..types import ShippingMethod, ShippingMethodZipCodeRule, ShippingZone
+from ..enums import PostalCodeRuleInclusionTypeEnum, ShippingMethodTypeEnum
+from ..types import ShippingMethod, ShippingMethodPostalCodeRule, ShippingZone
+
+
+class ShippingPostalCodeRulesCreateInputRange(graphene.InputObjectType):
+    start = graphene.String(
+        required=True, description="Start range of the postal code."
+    )
+    end = graphene.String(required=False, description="End range of the postal code.")
```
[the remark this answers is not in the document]
Kwaidan00: Yes, it does. The function for comparing postal code with rule handles it: https://github.com/mirumee/saleor/blob/9b668be5ad96a438e8c439c3ea1435300535c6ce/saleor/shipping/zip_codes.py#L37-L38. Moreover, the old mutations for creating postal code rule allows such ranges also: https://github.com/mirumee/saleor/blob/9b668be5ad96a438e8c439c3ea1435300535c6ce/saleor/graphql/shipping/mutations/shippings.py#L176

saleor/graphql/shipping/enums.py
```diff
@@ -1,4 +1,7 @@
 from ...graphql.core.enums import to_enum
-from ...shipping import ShippingMethodType
+from ...shipping import PostalCodeRuleInclusionType, ShippingMethodType
 
 ShippingMethodTypeEnum = to_enum(ShippingMethodType, type_name="ShippingMethodTypeEnum")
+PostalCodeRuleInclusionTypeEnum = to_enum(
```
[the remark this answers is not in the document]
d-wysocki: We would like to remove suffix `Enum`, so this could be called just a `PostalCodeRuleInclusionType`

PR #6678, opened 2021-01-04
Add order shipping tax rate

IKarbowiak:
- Add plugins methods for fetching shipping tax rates.
- Add `shipping_tax_rate` field on `Order`.
- Set order `shipping_tax_rate` when it's possible

Linked task: [SALEOR-1703](https://app.clickup.com/t/2549495/SALEOR-1703)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-1703 Missing shipping_rate for order](https://app.clickup.com/t/2549495/SALEOR-1703)

saleor/plugins/vatlayer/plugin.py
```diff
@@ -206,6 +206,65 @@ def calculate_order_line_unit(
...
+        checkout: "Checkout",
+        lines: Iterable["CheckoutLineInfo"],
+        address: Optional["Address"],
+        discounts: Iterable["DiscountInfo"],
+        previous_value: Decimal,
+    ):
+        return self._get_shipping_tax_rate(address, previous_value)
+
+    def get_order_shipping_tax_rate(self, order: "Order", previous_value: Decimal):
+        address = order.shipping_address or order.billing_address
```
maarcingebala, line 252: I think only `order.shipping_address` should be involved in this calculation. The billing address might be something totally different, wouldn't that result in a wrong calculation? E.g. for orders without shipping that don't have a shipping address?

IKarbowiak: I agree with you, that it seems like only `order.shipping_address` should be taken into consideration. I did it in this way because the `calculate_order_shipping` is doing like that and both must work the same. We should ask @korycins why this is done like this.

fowczarek (approved):

saleor/plugins/vatlayer/plugin.py
maarcingebala: OK so if it's consistent with the previous implementation let's leave it for now and ping Maciek when he's back from the holidays.

maarcingebala (approved):

saleor/plugins/vatlayer/plugin.py
```diff
@@ -206,6 +206,65 @@ def calculate_order_line_unit(
...
+    ):
+        if self._skip_plugin(previous_value):
+            return previous_value
+        country = address.country if address else None
+        taxes = self._get_taxes_for_country(country)
+        if not taxes:
+            return previous_value
+        tax = taxes.get(DEFAULT_TAX_RATE_NAME)
+        # tax value is given in precentage so it need be be converted into decimal value
+        return Decimal(tax["value"] / 100)
```
korycins, line 266: @IKarbowiak  Not sure of this. What about the case when `tax["value"]` will be 0?

patrys: It'll work for 0 (as 0/100 is 0) but it should really be `Decimal(tax["value"]) / 100`, otherwise we're creating a `float` and then converting it to a `Decimal`.

IKarbowiak: Good point, I'll change that.

PR #6683, opened 2021-01-05
Added support for querying objects by metadata fields

LeOndaz:
I'm leaving this PR here as in case you wanted to discuss the approach with me, I'll be adding tests and generalizing the metadata for all models that need it soon and I'll also rebuild the schema as the tests below failed because of it, just lemme hear your thoughts

### DIFF
--- a/saleor/graphql/core/filters.py
+++ b/saleor/graphql/core/filters.py
@@ -70,6 +70,13 @@ def filter_status(qs, _, value):
     return qs.filter(status=value)
 
 
+def filter_metadata(qs, _, value):
+    for metadata_item in value:
+        qs = qs.filter(metadata__contains={metadata_item.key: metadata_item.value})
+
+    return qs
+
+
 class BaseJobFilter(django_filters.FilterSet):
     created_at = ObjectTypeFilter(
         input_class=DateTimeRangeInput, method=filter_created_at
@@ -78,3 +85,16 @@ class BaseJobFilter(django_filters.FilterSet):
         input_class=DateTimeRangeInput, method=filter_updated_at
     )
     status = EnumFilter(input_class=JobStatusEnum, method=filter_status)
+
+
+class MetadataFilter(ListObjectTypeFilter):
+    def __init__(self, *args, **kwargs):
+        input_class = "saleor.graphql.meta.mutations.MetadataInput"
+        super().__init__(input_class, method=filter_metadata, *args, **kwargs)
+
+
+class MetadataFilterBase(django_filters.FilterSet):
+    metadata = MetadataFilter()
+
+    class Meta:
+        abstract = True
--- a/saleor/graphql/product/filters.py
+++ b/saleor/graphql/product/filters.py
@@ -16,7 +16,12 @@ from ...attribute.models import (
 from ...product.models import Category, Collection, Product, ProductType, ProductVariant
 from ...warehouse.models import Stock
 from ..channel.filters import get_channel_slug_from_filter_data
-from ..core.filters import EnumFilter, ListObjectTypeFilter, ObjectTypeFilter
+from ..core.filters import (
+    EnumFilter,
+    ListObjectTypeFilter,
+    MetadataFilterBase,
+    ObjectTypeFilter,
+)
 from ..core.types import ChannelFilterInputObjectType, FilterInputObjectType
 from ..core.types.common import IntRangeInput, PriceRangeInput
 from ..utils import get_nodes, resolve_global_ids_to_primary_keys
@@ -313,7 +318,7 @@ class ProductStockFilterInput(graphene.InputObjectType):
     quantity = graphene.Field(IntRangeInput, required=False)
 
 
-class ProductFilter(django_filters.FilterSet):
+class ProductFilter(MetadataFilterBase):
     is_published = django_filters.BooleanFilter(method="filter_is_published")
     collections = GlobalIDMultipleChoiceFilter(method=filter_collections)
     categories = GlobalIDMultipleChoiceFilter(method=filter_categories)
@@ -364,7 +369,7 @@ class ProductFilter(django_filters.FilterSet):
         return _filter_is_published(queryset, name, value, channel_slug)
 
 
-class ProductVariantFilter(django_filters.FilterSet):
+class ProductVariantFilter(MetadataFilterBase):
     search = django_filters.CharFilter(
         method=filter_fields_containing_value("name", "product__name", "sku")
     )
@@ -375,7 +380,7 @@ class ProductVariantFilter(django_filters.FilterSet):
         fields = ["search", "sku"]
 
 
-class CollectionFilter(django_filters.FilterSet):
+class CollectionFilter(MetadataFilterBase):
     published = EnumFilter(
         input_class=CollectionPublished, method="filter_is_published"
     )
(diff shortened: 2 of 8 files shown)


LeOndaz:
I'd appreciate suggestions about this test should be implemented

saleor/graphql/core/filters.py
```diff
@@ -70,6 +76,14 @@ def filter_status(qs, _, value):
     return qs.filter(status=value)
 
 
+def filter_metadata(qs, _, value):
+    for metadata_item in value:
+        lookup = "__".join(["metadata", metadata_item.key])
```
maarcingebala, line 81: This is correct, but I would suggest using an f-string for better readability:
```suggestion
        lookup = f'metadata__{metadata_item.key}'
```

maarcingebala (commented):
Tests for metadata API can be found in `saleor/graphql/meta/tests`. We could add them to `test_meta_queries.py`, but this file is pretty large and should be split into smaller files, so these tests could also be added to a new file e.g. `test_meta_filtering.py`. In each test we could take an object from a queryset, add a metadata key and value; then run a GraphQL query that would filter out this object.

saleor/graphql/core/filters.py
LeOndaz: @maarcingebala I've used fstrings at the beginning but I said to myself I don't know what's the least python version you're supporting, so I chose the one that's more supported, I'll change it for sure!

LeOndaz: And thanks for the hints, I'll do them, can you have a look at #6684

LeOndaz:
The test is added, it should pass after the checks are finished. Please don't forget to have a look at #6684 as this is important for my next PRs.

saleor/graphql/product/filters.py
```diff
@@ -313,6 +318,7 @@ class ProductFilter(django_filters.FilterSet):
     stocks = ObjectTypeFilter(input_class=ProductStockFilterInput, method=filter_stocks)
     search = django_filters.CharFilter(method=filter_search)
     ids = GlobalIDMultipleChoiceFilter(field_name="id")
+    metadata = MetadataFilter()
```
korycins, line 321: I would suggest creating the MetadataFilterBase class, inherited from django_filters.FilterSet. It should be placed somewhere in the `core` directory.  Then all filtres which should have metada field can use this class to automatically include it. Something similar we have for db models where we have `ModelWithMetadata` class to include meta fields. 
It should be similar to this example:
```
class Test(django_filters.FilterSet):
    metadata = MetadataFilter()
    class Meta:
        abstract=True

class ProductFilter(Test):
```

LeOndaz: At first I did that, but I didn't do it this way, probably because I'm still convinced that Metadata API is somewhat weird, to get my point can you have a look at the mentioned issue?

LeOndaz:
Any ideas on the adding more tests ?

LeOndaz:
I have an idea in mind that I'd like to share, I'm thinking of extending this search to a better version
```
metadata: {
   items: [],
   searchAlgorithm: CONCAT_RESULTS, FILTER_STACK, WHATEVER,
   keyCase: MATCH_CASE, ANY_CASE,
   value: ... 
}
```
and if value is provided, we assume it's the value, if not, we assume it has no value and it's an empty metadata item.
Just an overview of what I'm thinking, what do you think? ignoring the weird names, will rethink them if we agree on them

LeOndaz:
Also before putting effort in this, I'd like you to have a look at the mentioned issue kindly to see what I'm thinking

korycins:
The one thing that I am worried about is the performance of DB, when Saleor will have a lot of products. 
@patrys  what do you think?

patrys:
I agree. We don't want to filter on non-indexed data. We've previously kept metadata in a JSONB structure, but is there anything stopping us from using an indexed HSTORE?

LeOndaz:
@patrys Tbh I've never used HSTOREs, what do you think is the possible future of this PR ? if it may be accepted then we may in the future change the implementation itself while providing the same interface (or whatever), I'm stuck in my project because there's no way to add featured categories, collections, ...etc this is actually why I PRed this

patrys:
I understand how this implementation is already helpful in your case, but I'm hesitant to merge anything that we already know won't scale for at least some of our users. This is a great feature to have, but I believe we should tackle it in a performance-friendly way (which could mean migrating from JSONB to an indexed HSTORE column).

LeOndaz:
@patrys totally it's totally fine, if I were you I'd just do the same so I understand, I may help with this soon after I finish my current project s & PRs. Have a nice day!

patrys:
@LeOndaz see #6979 as it's adding indexes that make `metadata__contains=...` lookups fast with the existing JSONB fields.

saleor/graphql/core/filters.py
```diff
@@ -70,6 +70,14 @@ def filter_status(qs, _, value):
     return qs.filter(status=value)
 
 
+def filter_metadata(qs, _, value):
+    for metadata_item in value:
+        lookup = f"metadata__{metadata_item.key}"
+        qs = qs.filter(**{lookup: metadata_item.value})
```
patrys, line 76:
```suggestion
        qs = qs.filter(metadata__contains={metadata_item.key: metadata_item.value})
```

LeOndaz:
Great! here we go, I'll do it maybe later today

patrys (commented):
I would love to make sure we get the expected SQL execution plan (e.g., EXPLAIN not including any seq scans) for queries generated by this filter (especially when combined with another filter).

LeOndaz:
We may have a nice test for this, unfortunately, I'm having my exams these days, after I finish I may do it, you can merge this (if you see it's okay) and the test will be in a separate PR

korycins (approved):

maarcingebala (approved):

maarcingebala:
@LeOndaz Would you have some capacity to fix the linters? I should only require running `Black` and submitting reformatted code. After that, I think this PR should be ready to merge.

LeOndaz:
@maarcingebala sure, it's done

korycins:
@LeOndaz  It seems that you pushed the incorrect branch here

LeOndaz:
Sorry about that, I didn't work on this for weeks and I forgot where goes where, anyway I think I've done it right now

LeOndaz:
I thought I've wasted the files actually, I did reflog the repo to see if it's still there and I got it from there, you'll need to have a look

LeOndaz:
I think some other commits I've done before was pushed with this, I'll mention you when everything is resolved

LeOndaz:
@maarcingebala I've done it, everything should be working yet I've absolutely zero idea about what's failing, I've read it and saw some code like `assert 4 == 5` which is weird, anyway, I've updated saleor version to the master one too, so this is coming directly from master branch

fowczarek (approved):

IKarbowiak (approved):

d-wysocki (approved):

tomaszszymanski129 (approved):

maarcingebala:
@LeOndaz Thank you for contributing to Saleor 🎉

LeOndaz:
I hope for even more than contributions someday, have a great day!

PR #6718, opened 2021-01-14
Add search product by description

d-wysocki:
I want to merge this change because...I want to add SearchVectorField in product 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/product/mutations/products.py
```diff
@@ -525,10 +527,38 @@ def clean_attributes(
         )
         return attributes
 
+    @staticmethod
+    def parse_description_json_to_string(description):
```
maarcingebala, line 531: This is a core utility that could be used in many modules of Saleor, it should be moved out from a mutation. I suggest creating a file `saleor/core/utils/editorjs.py` and moving this function there. We could also move there stuff from `editorjs_sanitizer.py`. Then we will have one file with utilities for editor.js.

saleor/graphql/product/mutations/products.py
```diff
@@ -525,10 +527,38 @@ def clean_attributes(
...
+                    continue
+                string += clean_text_data(text)
+        return string
+
     @classmethod
     def clean_input(cls, info, instance, data):
         cleaned_input = super().clean_input(info, instance, data)
 
+        description = cleaned_input.get("description")
+        if description:
```
maarcingebala, line 557: What if the description is empty? One may want to clear the content of the description field and it should result in clearing the `description_search` field. Would that happen with the current implementation?

saleor/search/backends/postgresql_storefront.py
```diff
@@ -16,7 +16,7 @@ def search(phrase):
 
     """
     name_sim = TrigramSimilarity("name", phrase)
-    ft_in_description = Q(description__search=phrase)
+    ft_in_description = Q(search_vector=phrase)
```
maarcingebala, line 19: I would use the occasion and get rid of entire "saleor/search" module. We don't need it - it was used long time ago when we had ElasticSearch integration. If we wanted to add it now, we would implement it as a plugin. But for now it is not needed, so I would move this "search" function to where it is used: `graphql/product/filters.py`.

saleor/graphql/product/tests/test_product.py
```diff
@@ -1684,6 +1685,20 @@ def test_products_query_with_filter(
...
+    parser = ProductCreate.parse_description_json_to_string
+
+    assert parser({}) == ""
+
+    assert parser({"blocks": {"data": "some data"}}) == ""
+
+    data = {"blocks": [{"type": "list", "data": {"items": "some text"}}]}
+    assert parser(data) == "some text"
+
+    data = {"blocks": [{"type": "unstyled", "data": {"text": "some text"}}]}
```
maarcingebala, line 1698: Maybe we could also add a test for parsing multiple blocks?

saleor/graphql/product/mutations/products.py
d-wysocki: I updated this functionality to clear `description_search` when there is no `description` in payload

maarcingebala (approved):

saleor/graphql/product/mutations/products.py
```diff
@@ -529,6 +531,12 @@ def clean_attributes(
...
+            cleaned_input["description_search"] = clean_editor_js(description, True)
```
IKarbowiak, line 536: Maybe:
```suggestion
        cleaned_input["description_search"] = clean_editor_js(description, True) if description else ""
```

saleor/core/utils/editorjs.py
```diff
@@ -7,15 +7,16 @@
 HYPERLINK_TAG_WITH_URL_PATTERN = r"(.*?<a\s+href=\\?\")(\w+://\S+[^\\])(\\?\">)"
 
 
-def clean_editor_js(definitions: dict):
+def clean_editor_js(definitions: dict, to_string=False):
     """Sanitize a given EditorJS JSON definitions.
 
     Look for not allowed URLs, replaced them with `invalid` value, and clean valid ones.
     """
```
IKarbowiak, line 14: Could you add a short description for what the flag`to_string` is used to? I also think that we should change `to_string` to keyword-only parameter to force passing this by key.  For me call like `clean_editor_js(some_data, True)` is not clear and I need to go to the function to check what the `True` value is.

```suggestion
def clean_editor_js(definitions: dict, *, to_string=False):
```

d-wysocki: Good point, I will do that

IKarbowiak (approved):

saleor/graphql/product/tests/test_product.py
```diff
@@ -2242,6 +2242,73 @@ def test_create_product(
     assert color_value_slug in values
 
 
+def test_create_search_product_by_description(
```
tomaszszymanski129, line 2245: We have separate tests for product creation so IMO this test should be strictly about the search. If someone breaks product creation and run the tests, this one would also fail (because it's going to create mutation) and we'd have to wonder why the search broke 🤔 I would suggest to use the fixture and modify it to our needs / create new one if needed.

tomaszszymanski129 (changes requested):

saleor/graphql/product/tests/test_product.py
d-wysocki: To work properly `SearchVector` needs to be updated after create/update product also we have logic for update `SearchVector` in update/create product mutation. This test check if after create product via mutation, search function working good. For strictly search functionality we have test in `saleor/core/tests/test_postgresql_search.py`. I would like to keep this test.

fowczarek (approved):

saleor/graphql/product/tests/test_product.py
tomaszszymanski129: If we have the test for search in particular im happy with that. I'm just not sure about the approach of testing two separate queries in one unit test which kinda violates the term and falls into pitfall of false-positive if we break the 1st mutation.

tomaszszymanski129 (approved):
Approve, since that one test concept is my only concern, otherwise looks good 👍🏻

saleor/graphql/translations/mutations.py
```diff
@@ -268,8 +268,7 @@ def perform_mutation(cls, _root, info, **data):
 
 class PageTranslationInput(SeoTranslationInput):
     title = graphene.String()
-    content = graphene.String()
-    content_json = graphene.JSONString()
+    content = graphene.JSONString()
```
korycins, line 271: @d-wysocki  This will break the dashboard. Do we already working on it?

saleor/product/migrations/0139_description_vector_search.py
```diff
@@ -0,0 +1,69 @@
...
+            if not text:
+                continue
+            string += text
+    return string
+
+
+def migrate_description_into_description_search(apps, schema):
+    Product = apps.get_model("product", "Product")
+    for product in Product.objects.iterator():
+        product.description_search = parse_description_json_to_string(
```
korycins, line 32: @maarcingebala @d-wysocki  Are you sure that the name `description_search` is the best one? Personally I would say that this is a raw description which we use to search(but we still can use to differen things).

korycins: Also, what is the difference between content from `Product.description_search` and `Product.description`?

d-wysocki: `description` is formatted content for editor js (JSON) and `description search` is parsed JSON content to a string to allow using SearchVector.

This field for now is internal field only for text search purpose.

maarcingebala: That's true the field could be used for different things. Maybe `description_plaintext`?

korycins: `description_search` would suggest that contains something related only for searching and it contains only the raw description text. `description_plaintext` or `description_raw` or `description_str`, all of them would be better imo.

d-wysocki: ok, I will change it

saleor/graphql/product/tests/test_product.py
```diff
@@ -2305,6 +2305,73 @@ def test_create_product(
...
```
korycins, line 2308: Correct me if I am wrong. 
You are checking here if `create product mutation` creates data required to search right? Then you're checking if `search query` returns the expected data. 
Don't you think that it looks like two separate tests?
 - check if product created with the proper search data (like if the data exists in DB etc)
 - check if product search query returns expected results?

saleor/graphql/product/mutations/products.py
```diff
@@ -650,6 +661,10 @@ class Meta:
     @transaction.atomic
     def save(cls, info, instance, cleaned_input):
         instance.save()
+        description_plaintext = cleaned_input.get("description_plaintext")
```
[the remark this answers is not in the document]
d-wysocki: SearchVector works only with update. So I cant add it to first save(). I will add comment in the code.

saleor/graphql/product/tests/test_product.py
d-wysocki: I will create two seperate tests :)

saleor/graphql/product/mutations/products.py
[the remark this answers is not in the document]
patrys: Can we create triggers to automatically populate the vectors when data changes? I don't think we should explicitly update each newly created row in the database.

http://rubyjacket.com/build/django-psql-fts.html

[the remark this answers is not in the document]
d-wysocki: Yes, we can. I thought my solution will be enough.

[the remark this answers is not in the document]
d-wysocki: Done :)

PR #6756, opened 2021-01-20
Use separate services for external integrations

patrys:
This introduces separate services for tracing external integrations

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/payment/gateways/braintree/__init__.py
+++ b/saleor/payment/gateways/braintree/__init__.py
@@ -1,6 +1,8 @@
 from typing import Dict, List, Optional
 
 import braintree as braintree_sdk
+import opentracing
+import opentracing.tags
 from django.core.exceptions import ImproperlyConfigured
 
 from ... import TransactionKind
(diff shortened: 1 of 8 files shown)


saleor/payment/gateways/braintree/__init__.py
```diff
@@ -107,10 +109,16 @@ def get_client_token(
     config: GatewayConfig, token_config: Optional[TokenConfig] = None
 ) -> str:
     gateway = get_braintree_gateway(**config.connection_params)
-    if not token_config:
-        return gateway.client_token.generate()
-    parameters = create_token_params(config, token_config)
-    return gateway.client_token.generate(parameters)
+    with opentracing.global_tracer().start_active_span(
```
korycins, line 112: Not sure if we should do that in this PR, but maybe we could create our function for opentracing, which could recieve the name and tags?
something like:
```
with start_active_span("braintree.client_token.generate", {opentracing.tags.COMPONENT: "payment", "service.name": "braintree"}):
    ...
```

korycins (approved):

saleor/payment/gateways/braintree/__init__.py
patrys: I want to switch from OpenTracing to OpenTelemetry (which adds the possibility to include other metrics) in the first half of 2021. I think that will present a good opportunity to improve the readability of our tracing code.

fowczarek:
We should add this tracking to authorize.net too.

fowczarek (approved):

PR #6785, opened 2021-01-25
Feature/6705 address metadata

sagital:

At the minimum, we need to store the VAT number for a company customer in order to emit an invoice. There might be other billing details that need to be stored so I am thinking the `address` entity could have its own metadata. 

https://github.com/mirumee/saleor/issues/6705

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] The changes are tested


sagital:
The code coverage for permissions.py was down with 12.45% but I am not sure how to improve it. 

I have added two new permissions, `private_address_permissions` and `public_address_permissions` and each of them has a check if the `address` is null or the corresponding `user` is null.

Neither of them could happen from my initial research, because if the address by pk is null  the code doesn't even get there (graphql will return an empty response) , and on the other hand only the addresses corresponding to an user seem to be retrieved so the `user` also  cannot be null.

sagital:
Hello, I got a request to reopen this PR, and I wanted to check if this will be used before I rebase and retest it.  Thanks

patrys:
@saleor/core, assemble! Let's get this reviewed and merged, as it's a valuable feature. The only part I'm unsure about is the one that checks address ownership, as it executes a database query every time it needs to.

PR #6793, opened 2021-01-26
SendGridPlugin

korycins:
Introduce new plugin responsible for sending the customer emails.


saleor/plugins/sendgrid/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+        UserNotifyEvent.ORDER_FULFILLMENT_UPDATE: send_fulfillment_update_email_task,
+        UserNotifyEvent.ORDER_PAYMENT_CONFIRMATION: (
+            send_payment_confirmation_email_task
+        ),
+        UserNotifyEvent.ORDER_CANCELED: send_order_canceled_email_task,
+        UserNotifyEvent.ORDER_REFUND_CONFIRMATION: send_order_refund_email_task,
+    }
+
+
+def get_dynamic_template_map():
```
IKarbowiak, line 60: Cannot we join `get_event_to_task_map` and `get_dynamic_template_map` to one mapping?

IKarbowiak (approved):

saleor/plugins/sendgrid/plugin.py
korycins: I think we can, but imo by the cost of readability:
```
return {
     UserNotifyEvent.ACCOUNT_CONFIRMATION: ("account_confirmation_template_id", send_account_confirmation_email_task)
     ...
}
```
@IKarbowiak  Do you think that this is the better way?

IKarbowiak: For me, it's a better way, as otherwise, you repeated every NotifyEvent. But the decision is up to you :)

saleor/plugins/sendgrid/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+        },
+        "sender_address": {
+            "type": ConfigurationTypeField.STRING,
+            "help_text": "Sender email which will be visible as 'from' email.",
+            "label": "Sender email",
+        },
+        "account_confirmation_template_id": {
+            "type": ConfigurationTypeField.STRING,
+            "help_text": "The dynamic template ID for account confirmation.",
+            "label": "Template ID for account confirmation",
```
maarcingebala, line 131: The text is duplicated in both fields, maybe instead in "label" we could name the template and in "help text" describe in what form the template should be provided, something like this:

label: "Account confirmation email template"
help text: "ID of the dynamic template in Sendgrid"

saleor/plugins/sendgrid/plugin.py
maarcingebala: Also, does it have to be a function instead of a plain dictionary?

saleor/plugins/sendgrid/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+        "order_refund_confirmation_template_id": {
+            "type": ConfigurationTypeField.STRING,
+            "help_text": "The dynamic template ID for order refund confirmation",
+            "label": "Template ID for order refund confirmation",
+        },
+        "api_key": {
+            "type": ConfigurationTypeField.SECRET,
+            "help_text": (
+                "This allows your application to authenticate to our API and send mail."
+            ),
```
maarcingebala, line 204: It allows our application (Saleor) to authenticate to Sendgrid API and send an email. But I think it would be enough to say: 

```suggestion
            "help_text": ("Your Sendgrid API key."),
```

saleor/plugins/sendgrid/tasks.py
```diff
@@ -0,0 +1,286 @@
...
+    message = Mail(from_email=from_email, to_emails=recipient_email)
+    message.dynamic_template_data = payload
+    message.template_id = template_id
+    sendgrid_client.send(message)
+
+
+@app.task(
+    autoretry_for=(SendGridException,),
+    retry_backoff=60,
+    retry_kwargs={"max_retries": 5},
```
maarcingebala, line 28: These could be constants since they are common to all tasks.

fowczarek (approved):

patrys:
Task linked: [SALEOR-2744 SendGridEmail Plugin](https://app.clickup.com/t/2549495/SALEOR-2744)

maarcingebala (approved):

PR #6799, opened 2021-01-28
Add interface for integrating the auth plugins

korycins:
This interface will allow us to integrate plugins responsible for authentication


fowczarek (approved):

saleor/core/jwt.py
```diff
@@ -115,8 +126,30 @@ def get_user_from_payload(payload: Dict[str, Any]) -> Optional[User]:
     return user
 
 
+def is_token_belong_to_saleor(token: str) -> bool:
```
maarcingebala, line 129: Maybe `is_saleor_token`? Sorry for nitpicking, but the current name sounds somewhat incorrect.

maarcingebala (approved):

saleor/core/jwt.py
korycins: You are right

PR #6800, opened 2021-01-28
Fix stripe

### LINKED ISSUE #5799: Stripe - A token may not be passed in as a PaymentMethod. Instead, use payment_method_data with type=card and card[token]
Complete a checkout with Stripe as the payment method.

1. Add keys to Stripe plugin
2. Retrieve token via StripeJS/Elements
3. Mutation createCheckoutPayment(...)
4. Mutation completeCheckout(...)

### What I expected to happen
Checkout to be completed, instead it throws an error

### Screenshots
image

**System information**
Operating system:
MacOS Cataline 10.15.5
Running Saleor with Docker

LeOndaz:
Fixed #5799

### DIFF
--- a/saleor/payment/gateways/stripe/__init__.py
+++ b/saleor/payment/gateways/stripe/__init__.py
@@ -53,7 +53,10 @@ def authorize(
             span.set_tag(opentracing.tags.COMPONENT, "payment")
             span.set_tag("service.name", "stripe")
             intent = client.PaymentIntent.create(
-                payment_method=payment_information.token,
+                payment_method_data={
+                    "type": "card",
+                    "card": {"token": payment_information.token},
+                },
                 amount=stripe_amount,
                 currency=currency,
                 confirmation_method="manual",


LeOndaz:
MB I forgot to commit the changes lol, anyway it's fixed now

LeOndaz:
I think I missed up the PR ~.~

LeOndaz:
@korycins what do you think? (I've tested and it's working, payments are accepted)

korycins:
@LeOndaz Can you share the stripe docs related to this change? I am trying to find something, but without any success :c
@maarcingebala  You're already working on the improvement of the stripe gateway. What do you think about these changes?

maarcingebala:
I couldn't reproduce the issue described in #6800, I tested Stripe recently and payments were created successfully in Stripe.

LeOndaz:
@korycins I don't know about the documentation part (I found nothing), I found this as a result of the error telling me I'm passing the token as payment method data, something like the image in the issue mentioned above, and I just edited the code to match what the error logs said, I got payments working with this 

image of successful payments

@maarcingebala Can you try with the latest version of both saleor & dashbaord? (the master branch), I'm using both and I update regularly, I'm neither using the storefront nor saleor sdk though, I'm sending all the data as expected (paymentData is not there in stripe, a token from stripe.createToken(...), checkoutId, no amount (uses checkout total))

LeOndaz:
@korycins @maarcingebala I think I'm getting my head wrapped around the issue (not sure), It's expected that a string token should be entered in `checkoutPaymentCreate`, however, if you pass a string, you get an error saying you can't use a string as a payment_method, instead use payment_method_data, I saw [this spectrum thread](https://spectrum.chat/saleor/development/issue-with-stripe-gateway-a-token-may-not-be-passed-in-as-a-paymentmethod-instead-use-payment-method-data-with-type-card-and-card-token-tok-1hpunhlf9gdrjvxqcqbxggbe~f100b0f4-e63f-4312-8fda-56befb11d303), so basically the token is being used as a payment method on the backend which should not happen, so the solution imo is either the token in createPayment should be paymentData: JSONString and we pass a full payment_method created by `stripe.createPaymentMethod()` and this will change the interface of usage so it's is a breaking change, or we can change it to payment_method_data and specify the token directly here and this is not a breaking change as the interface of usage will be the same and this is describing this PR, there's also a `Token.create` in stripe's python package

LeOndaz:
@korycins @maarcingebala Furthur more, the process_payment in stripe calls authorize with some `PaymentData`, the `PaymentData.token` is a string, but stripe expects `payment_method` not to be a string

LeOndaz:
@korycins any news?

korycins:
@LeOndaz Hey, just letting you know -  I am working on a new implementation of the Stripe payment gateway, which will use a custom payment flow described [here](https://stripe.com/docs/payments/integration-builder). It should be included in 3.0

LeOndaz:
Thank you, I'll be following, or maybe contributing

korycins:
I should open the PR in a couple of days (hope so 😄  )

korycins:
I beleive we can close this PR. New stripe plugin is placed here - https://github.com/mirumee/saleor/pull/7452 , docs are here - https://github.com/mirumee/saleor-docs/pull/277

PR #6801, opened 2021-01-28
Extracted require_active_plugin decorator

LeOndaz:


korycins (approved):

maarcingebala (approved):

maarcingebala:
Thanks for the PR! We spoke once that this function needs to be extracted but apparently, no one had time to do that earlier ;)

LeOndaz:
@maarcingebala Np, it was weird finding it in every file lol, anyway I have a question in mind, any ideas about when will the marketplace be available?

maarcingebala:
@korycins Is leading the Marketplace project so he can probably tell more. It depends on current priorities which sometimes change quickly, as we're also focusing on building some integrations around Saleor and Saleor Cloud.

LeOndaz:
@maarcingebala I understand, I'll wait for his response, thank you.

korycins:
@LeOndaz Hey, It is hard to say any exact date for now. 
We put all our resources in releases 2.11 and 3.0, but now we're planning to move with the marketplace also. At the end of Q1, we should be able to say more.

LeOndaz:
@korycins okay, one will have a nice collection of plugins by then I guess, thank you.

PR #6833, opened 2021-02-03
Price precision validation fix

### LINKED ISSUE #6820: Bug in validate_price_precision
Get sane validation obviously.

1. Try to create a voucher with the minimum order amount set to `2000`
2. It will throw an error.

### What I expected to happen
It shouldn't throw an error.

### Observation
Upon normalizing it converts the zeros to exponents.

```python
def validate_price_precision(value: Optional["Decimal"], currency: str = None):
    """Validate if price amount does not have too many decimal places.

    Price amount can't have more decimal places than currency allows to.
    Works only with decimal created from a string.
    """

    # check no needed when there is no value
    if not value:
        return

    currency_fraction = get_currency_fraction(currency or settings.DEFAULT_CURRENCY)
    value = value.normalize()
    if abs(value.as_tuple().exponent) > currency_fraction:
        raise ValidationError(
            f"Value cannot have more than {currency_fraction} decimal places."
        )
```
should be:

```python
def validate_price_precision(value: Optional["Decimal"], currency: str = None):
    """Validate if price amount does not have too many decimal places.

    Price amount can't have more decimal places than currency allows to.
    Works only with decimal created from a string.
    """

    # check no needed when there is no value
    if not value:
        return

    currency_fraction = get_currency_fraction(currency or settings.DEFAULT_CURRENCY)
    value = value.normalize()
    exp = value.as_tuple().exponent
    if exp < 0 and abs(value.as_tuple().exponent) > currency_fraction:
        raise ValidationError(
            f"Value cannot have more than {currency_fraction} decimal places."
        )
```
So that it doesn't misinterpret zeros from the right as values after decimal places.

IKarbowiak:
Apply #6572 fix to 2.11.

Resolves #6820

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable, unreleased changes to this project will be documented in this file.
 ## [Unreleased]
 
 - Deprecate `Shop.geolocalization` query - #6828 by @maarcingebala
+- Price precision validation fix - #6833 by @IKarbowiakg
 
 # 2.11.7
 
--- a/saleor/graphql/core/tests/test_validators.py
+++ b/saleor/graphql/core/tests/test_validators.py
@@ -16,6 +16,7 @@ from ..validators import validate_one_of_args_is_in_query, validate_price_precis
         (Decimal("1"), "ISK"),
         (Decimal("1.00"), "ISK"),
         (Decimal("5.12"), None),
+        (Decimal("1000"), "USD"),
     ],
 )
 def test_validate_price_precision(value, currency):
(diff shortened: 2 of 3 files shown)


d-wysocki (approved):

maarcingebala (approved):
@IKarbowiak We need to update the changelog as well.

maarcingebala (approved):

fowczarek (approved):

PR #6872, opened 2021-02-10
Add search rank to products

d-wysocki:
I want to merge this change because...I want to add search rank to searching products and I want to fix dependency of database default languade in search

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/product/filters.py
```diff
@@ -217,9 +218,17 @@ def product_search(phrase):
         phrase (str): searched phrase
 
     """
-    ft_in_description_or_name = Q(search_vector=phrase)
+    query = SearchQuery(phrase, config="english")
```
maarcingebala, line 221: Don't we want to determine the config from `LANGUAGE_CODE` in settings? We would have to obviously map it to values that are accepted by Postgres. That would improve the search for people who run open-source Saleor and use a different language than English.

fowczarek (approved):

saleor/graphql/product/filters.py
maarcingebala: We'll do it separately since currently English is also hardcoded in migrations and supporting more languages is a bit more work.

maarcingebala (approved):

IKarbowiak (approved):

saleor/graphql/product/tests/test_product.py
```diff
@@ -2427,6 +2430,15 @@ def test_search_product_by_description(user_api_client, product_list, channel_US
 
     assert len(content["data"]["products"]["edges"]) == 1
 
+    variables = {"filters": {"search": "red"}, "channel": channel_USD.slug}
```
korycins, line 2433: IMO these should be separate tests, responsible for testing the search when we need to take into consideration the weights of the fields. 
Also:
-  I would explicitly add the `red` word to the product description. Now, it is hard to understand, why the query returns two products. 
- I would use more objects for this test(remember about bulk_create). Product1 doesn't have `red` word, product 2 has `red` in the description, product 3 has two `red` words in the description, product 4 has `red` in name. The expected results should be - product4, product3, product2
- Also, DB returns products sorted by `slug`. I would also add different slugs to the products, to be fully sure that, returned products are sorted by search, and not by default slug sorting.

saleor/graphql/product/filters.py
korycins: @maarcingebala I think the first proposition is good. I mean, we could use Languade_code to determine the languade, for en we will use English, for different languages we will fallback to the simple config. 
@d-wysocki Is it possible?

d-wysocki: it's possible but as Marcin said we have hardcoded English in our migrations. For now we would like to stay with this solution as we agreed to hardcode english language at the beginning.

korycins: Just to confirm. Using simple requires an additional migration?

saleor/graphql/product/sorters.py
```diff
@@ -142,6 +143,9 @@ def description(self):
                 "collection. Note: "
                 "This option is available only for the `Collection.products` query."
             ),
+            ProductOrderField.RANK.name: (
+                "rank. Note: " "This option is available only with `search` filter."
```
maarcingebala, line 147:
```suggestion
                "rank. Note: This option is available only with the `search` filter."
```

saleor/graphql/product/filters.py
d-wysocki: Yes. For now we decided to hardcode English language for search vector.

Issue #5630, opened 2020-05-12
Bump mypy


maarcingebala:
Mypy has been locked to `0.740`. We should upgrade it to the newest version, but it requires fixing new errors in the codebase. Locking mypy also locked django-stubs, and we should upgrade it as well.

IKarbowiak:
We are blocked because of a lack of support for QuerySet.as_manager() in `django-stubs`. Bump mypy to 0.77 and django-stubs to 1.5 generate about 50 errors connected with this problem. Here is a ticket for that on `django-stubs`: https://github.com/typeddjango/django-stubs/issues/265 .

maarcingebala:
Mypy is now bumped to 0.910.

PR #6876, opened 2021-02-11
Bump ubuntu on gh action with pytest

fowczarek:
I want to merge this change because bumping ubuntu and python on gh action with pytest.

I Try to bump the python version from pytest action but we run pre-commit on the same env and mypy-0.740 doesn't work with python3.9. 

We have a separate task to bump mypy #5630 after this task we should bump the python version in this gh-action. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -16,7 +16,7 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
-    container: python:3.9
+    container: python:3.8
 
     services:
       postgres:
@@ -33,10 +33,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Setup Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: "3.9"
 
       - name: Install system dependencies
         run: apt-get install -y libpq-dev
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest.yml
```diff
@@ -15,8 +15,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-18.04
-    container: python:3.8
+    runs-on: ubuntu-latest
```
krzysztofwolski, line 18: please use exact version name to avoid issues when new ubuntu arrives

patrys: I'm against pinning this to a particular version as any version we put here will remain here for years.

krzysztofwolski: Wouldnt be a better solution to do a package upgrade chore task every other month? This would address python deps too.

fowczarek: We bump python to 3.9 a long time ago. In this CI config, I change python from 3.8 to 3.9 this week when CI stop working.

IKarbowiak (approved):

maarcingebala (approved):

PR #6897, opened 2021-02-16
Lock allocations to prevent negative available quantity

d-wysocki:
I want to merge this change because...I want to select for update allocations row while completing checkout to prevent negative available quantity

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/checkout/complete_checkout.py
```diff
@@ -342,6 +343,8 @@ def _create_order(
     """
     from ..order.utils import add_gift_card_to_order
 
+    Allocation.objects.select_for_update(of=("self",))
```
maarcingebala, line 346: Have you tried locking allocation inside `allocate_stocks`? This is the function that is supposed to handle that, so it would be nice to have this logic there.

d-wysocki: Yes, I tried but without any results. Only adding this lock at the beginning of completing checkout works fine.

korycins: Are you sure of this? I see that `allocate_stocks` raises the exception when there are not enough items in the stock.
Also, you're allocating here all allocations which means that you block the possibility to process more than one checkoutComplete at the same time (AFAIK).

d-wysocki: Yes, I'm sure. I added this lock into allocate_stocks and that doesn't prevent this negative quantity

fowczarek: But if you add lock in this way, it's lock read of any allocation(Any client can't check available quantity) when the server run  `_create_order` for any order.

korycins: imo, it is workaround for something that doesn't work as we expect. Locking all allocations will force Saleor to **process all checkouts one by one**, which is not the best way for handling this case. 
Have you investigated which place in code causes the issue?

IKarbowiak (approved):

saleor/warehouse/management.py
```diff
@@ -33,39 +36,54 @@ def allocate_stocks(order_lines_info: Iterable["OrderLineData"], country_code: s
         return
 
     variants = [line_info.variant for line_info in order_lines_info]
+    if None in variants:
+        raise ValidationError(
+            {
+                "variant": ValidationError(
+                    "Missing variant in order line. Cannot allocate the stock",
+                    code=StockErrorCode.INVALID.value,
+                )
```
maarcingebala, line 45: Missing variant shouldn't happen here, I assume this is also due to a race condition when a customer tries to buy a variant, that is suddenly removed from the system?

Maybe it would be better to change the line above:
```
variants = [line_info.variant for line_info in order_lines_info if line_info.variant]
```
And only create allocations for existing variants? What do you think @fowczarek?

maarcingebala: If we want to raise an error I'm not sure if this should be a validation error. A missing variant in this place is the same as InsufficientStock error from the user's perspective.

d-wysocki: I'm agree with that missing variant shouldn't happen here. I added this check to be sure because OrderLineInfo have variant as optional field.

saleor/warehouse/management.py
```diff
@@ -33,39 +36,54 @@ def allocate_stocks(order_lines_info: Iterable["OrderLineData"], country_code: s
...
+            }
+        )
 
-    stocks = (
+    stocks = list(
         Stock.objects.select_for_update(of=("self",))
```
korycins, line 50: Will `select_for_update` work when you convert the qs to the list?

d-wysocki: Yes, before my changes select_for_update was triggered too late (in lazy way), to avoid race condition.

saleor/warehouse/management.py
maarcingebala: So we should raise `InsufficientStock` in this case, since creating allocation can throw this error anyway, and it should be handled automatically at the mutation level. See how insufficient stock errors are returned in `_create_allocations`. Also we need to add a test for this case.

d-wysocki: I removed this validation. This should be discussed and fixed in other PR.

fowczarek (approved):

maarcingebala (approved):

PR #6906, opened 2021-02-19
Use checkout info in recalculate_checkout_discount

IKarbowiak:
Refactor `recalculate_checkout_discount` and related methods.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/fetch.py
+++ b/saleor/checkout/fetch.py
@@ -112,3 +112,14 @@ def fetch_checkout_info(
         shipping_method_channel_listings=shipping_channel_listings,
         valid_shipping_methods=valid_shipping_method,
     )
+
+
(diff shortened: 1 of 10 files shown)


maarcingebala (approved):
I see that in complex mutation like in Checkout we have to now synchronize `Checkout` and `CheckoutInfo` (e.g. `update_checkout_info_shipping_method`). We need to keep an eye on these places. I'm afraid that at some point it might be easy to miss the synchronization and this would lead to a new class of bugs when the model instance and the related data class instance store different data. Maybe we will have to come up with some universal solution to this problem in the future.

fowczarek (approved):

PR #6923, opened 2021-02-23
Fix logic to calculate prices with proper tax calculation

korycins:
I want to merge this change because it changes the way of applying the taxes to fit the case with discounted line/order.
Some fixmes section was just dropped as we had incorrect assumption.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


fowczarek (approved):
We should optimize event resolvers, but they should be separate tasks. I will create it.

IKarbowiak (approved):

PR #6931, opened 2021-02-24
Add partial product search

d-wysocki:
I want to merge this change because...I want to add partial product search

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/product/filters.py
+++ b/saleor/graphql/product/filters.py
@@ -218,7 +218,7 @@ def product_search(phrase):
         phrase (str): searched phrase
 
     """
-    query = SearchQuery(phrase, config="english")
+    query = SearchQuery(f"'{phrase}':*", config="english", search_type="raw")
     vector = F("search_vector")
     ft_in_description_or_name = Q(search_vector=query)
 
--- a/saleor/graphql/product/tests/test_product.py
+++ b/saleor/graphql/product/tests/test_product.py
@@ -2487,7 +2487,7 @@ def test_search_product_by_description_and_name(
     product.save()
 
     product_2 = product_list[1]
-    product_2.name = "red product"
+    product_2.name = "redproduct"
     product_2.save()
     product_1 = product_list[0]
     product_1.description_plaintext = "some red product"


patrys:
What's the use case? What if I don't want this behavior?

d-wysocki:
> What's the use case? What if I don't want this behavior?

@patrys For example if you had product with name `monsterblack` and you will type only `monster` into search field. You cannot turn off this functionality, we could add flag for this it will be necessary.

patrys:
Why not let the client build the query and document how operators work?

maarcingebala:
I'm not familiar with the operators, how would it look like in the API for example?

patrys:
https://www.postgresql.org/docs/current/textsearch-controls.html

Examples:

* `products(filter: { search: "foo & bar" }, ...)` - containing both "foo" and "bar"
* `products(filter: { search: "foo | bar" }, ...)` - containing "foo" or "bar"
* `products(filter: { search: "!foo" }, ...)` - not containing "foo"
* `products(filter: { search: "foo:A | bar:B" }, ...)` - title (weight A) containing "foo" or description (weight B) containing "bar"
* `products(filter: { search: "foo:*" }, ...)` - containing a word starting with "foo"
* `products(filter: { search: "foo & (bar | baz)" }, ...)` - using parenthesis to override operator precedence

maarcingebala:
Actually, that would be pretty awesome, but I have no idea how much work it would be. Seems like we should just pass the phrase from API directly to Postgres. There should be also some error handling when the user passes an invalid phrase, in which case we should raise some GraphQLError.

patrys:
It seems the syntax is already handled if we drop the `:*` change (current implementation won't work if you search using two words). Not sure how syntax errors are handled.

tomaszszymanski129 (approved):

maarcingebala:
We won't have time in the nearest future to improve this solution as discussed above. Should I close the PR for now?

patrys:
Task linked: [SALEOR-2495 Add partial search to SearchVector in product](https://app.clickup.com/t/2549495/SALEOR-2495)

PR #6952, opened 2021-03-02
update the way of calculating the order discount

korycins:
I want to merge this change because it introduces the proper way of calculating the percentage discount when the price has taxes

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/order/utils.py
+++ b/saleor/order/utils.py
@@ -1,7 +1,7 @@
 import copy
 from decimal import Decimal
 from functools import partial, wraps
-from typing import Iterable, List, Optional, Union
+from typing import Iterable, List, Optional, Tuple, Union
 
 from django.conf import settings
 from django.db import transaction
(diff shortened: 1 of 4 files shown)


d-wysocki (approved):

saleor/order/utils.py
```diff
@@ -576,15 +574,30 @@ def create_order_discount_for_order(
...
+        )
+        gross_total = apply_discount_to_value(
+            value, value_type, currency, current_total.gross
+        )
+    else:
+        new_total = apply_discount_to_value(value, value_type, currency, current_total)
+        net_total = new_total.net
+        gross_total = new_total.gross
+
+    new_amount = (current_total - gross_total).gross
```
maarcingebala, line 592: The same block is used below in `update_order_discount_for_order`, maybe it should be a function?

saleor/order/utils.py
```diff
@@ -76,7 +76,7 @@ def get_voucher_discount_assigned_to_order(order: Order):
     return order.discounts.filter(type=OrderDiscountType.VOUCHER).first()
 
 
-def recalculate_order_discounts(order: Order):
+def recalculate_order_discounts(order: Order, with_discount_event: bool = True):
```
maarcingebala, line 79: Are there cases that we want to skip the history log? Flag arguments are no good :(

saleor/order/utils.py
```diff
@@ -87,16 +87,14 @@ def recalculate_order_discounts(order: Order):
...
-            # lines price
-            already_applied_discount_amount=Money(Decimal(0), currency=order.currency),
-            save_order=False,
         )
         discount_value = order_discount.value
         discount_type = order_discount.value_type
         amount = order_discount.amount
 
+        if not with_discount_event:
+            continue
```
maarcingebala, line 96: I would say it should be:
```
if with_discount_event:
    # action
```

saleor/order/utils.py
korycins: Yes, I am aware, but after 2 hours of trying to find a better solution, I didn't find anything.
For order discount mutations, we call recalculate_order_discounts but we don't want to place the event as each of these mutations has its own specific event. Let me know if you have better way

saleor/order/utils.py
korycins: This will add one additional injection to the code :P This will cause that the next condition will be hard to understand but it is up to you

saleor/order/utils.py
patrys: What if the event was created by the caller and the function only returned a boolean, signaling whether anything changed?

maarcingebala (approved):

PR #6956, opened 2021-03-02
Allow to add url without TLD to install_app

d-wysocki:
I want to merge this change because...I want to allow install app with url without TLD

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/app/validators.py
+++ b/saleor/app/validators.py
@@ -1,0 +1,18 @@
+import re
+
+from django.core.validators import URLValidator
+from django.utils.regex_helper import _lazy_re_compile  # type: ignore
+
+
+class AppURLValidator(URLValidator):
+    validator = URLValidator
+    host_re = "(" + validator.hostname_re + validator.domain_re + "|localhost)"
+    regex = _lazy_re_compile(
+        r"^(?:[a-z0-9.+-]*)://"  # scheme is validated separately
+        r"(?:[^\s:@/]+(?::[^\s:@/]*)?@)?"  # user:pass authentication
+        r"(?:" + validator.ipv4_re + "|" + validator.ipv6_re + "|" + host_re + ")"
+        r"(?::\d{2,5})?"  # port
+        r"(?:[/?#][^\s]*)?"  # resource path
+        r"\Z",
+        re.IGNORECASE,
+    )
--- a/saleor/core/utils/url.py
+++ b/saleor/core/utils/url.py
@@ -11,7 +11,15 @@ def validate_storefront_url(url):
     Raise ValidationError if URL isn't in RFC 1808 format
     or it isn't allowed by ALLOWED_CLIENT_HOSTS in settings.
     """
-    domain = validate_url(url)
+    try:
+        parsed_url = urlparse(url)
+        domain, _ = split_domain_port(parsed_url.netloc)
+        if not parsed_url.netloc:
+            raise ValidationError(
+                "Invalid URL. Please check if URL is in RFC 1808 format."
+            )
+    except ValueError as error:
+        raise ValidationError(error)
     if not validate_host(domain, settings.ALLOWED_CLIENT_HOSTS):
         error_message = (
             f"{domain or url} is not allowed. Please check "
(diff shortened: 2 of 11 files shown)


patrys:
Task linked: [SALEOR-2236 `install_app` rejects valid URLs](https://app.clickup.com/t/2549495/SALEOR-2236)

saleor/core/utils/url.py
```diff
@@ -33,3 +25,16 @@ def prepare_url(params: str, redirect_url: str) -> str:
...
+    try:
+        parsed_url = urlparse(url)
+        domain, _ = split_domain_port(parsed_url.netloc)
+        if not parsed_url.netloc:
+            raise ValidationError(
+                "Invalid URL. Please check if URL is in RFC 1808 format."
+            )
+    except ValueError as error:
+        raise ValidationError(error)
+    return domain
```
IKarbowiak, line 40: Do we have tests for this?

IKarbowiak (approved):

d-wysocki:
Should we validate if this url it's in `ALLOWED_HOSTS`?

saleor/core/utils/url.py
korycins, line 40: Personally, it is strange that a function with the name  `validate_url` returns the domain.  I wouldn't expect that the validation will return me a part of url.

korycins: I don't think that this method does the same job as URLValidation. I understand that the issue was TLD, but why we didn't extend the URLValidator to handle the expected cases? Now, from a fully validated URL, we moved to parse URL. 
I would start with something like this:
```
valdiator = URLValidator()
validator.tld_re = "your regex"
valdiator.host_re = '(' + hostname_re + domain_re + tld_re + '|localhost)'
```

d-wysocki: Good point, I will change that.

saleor/core/utils/url.py
```diff
@@ -33,3 +25,16 @@ def prepare_url(params: str, redirect_url: str) -> str:
     split_url = urlsplit(redirect_url)
     split_url = split_url._replace(query=params)
     return split_url.geturl()
+
+
+def validate_url(url):
```
fowczarek, line 30: We should add docstring to this function.

fowczarek (approved):

korycins:
@d-wysocki  also one important thing. The App model uses URLField which uses a standard URLValidator for checking the value. We need to take this into account during applying this changes, afaik.

saleor/core/utils/url.py
```diff
@@ -33,3 +36,17 @@ def prepare_url(params: str, redirect_url: str) -> str:
     split_url = urlsplit(redirect_url)
     split_url = split_url._replace(query=params)
     return split_url.geturl()
+
+
+class APPURLValidator(URLValidator):
+    validator = URLValidator
+    host_re = "(" + validator.hostname_re + validator.domain_re + "|localhost)"
+    regex = _lazy_re_compile(
+        r"^(?:[a-z0-9.+-]*)://"  # scheme is validated separately
```
maarcingebala, line 45: Typo?
```suggestion
        r"^(?:[a-z0-9.+-]*)://"  # schema is validated separately
```

saleor/core/utils/url.py
```diff
@@ -33,3 +36,17 @@ def prepare_url(params: str, redirect_url: str) -> str:
     split_url = urlsplit(redirect_url)
     split_url = split_url._replace(query=params)
     return split_url.geturl()
+
+
+class APPURLValidator(URLValidator):
```
maarcingebala, line 41: I think it should say: `AppURLValidator`

maarcingebala: I also think this validator could be part of the `saleor/app` module. It's specific to apps; also it is used only in app-related modules.

saleor/core/utils/url.py
d-wysocki: `Scheme` its for URLs
https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_is_a_URL#scheme

maarcingebala: 👍

saleor/app/validators.py
```diff
@@ -0,0 +1,18 @@
...
+
+from django.core.validators import URLValidator
+from django.utils.regex_helper import _lazy_re_compile  # type: ignore
+
+
+class AppURLValidator(URLValidator):
+    validator = URLValidator
+    host_re = "(" + validator.hostname_re + validator.domain_re + "|localhost)"
+    regex = _lazy_re_compile(
+        r"^(?:[a-z0-9.+-]*)://"  # scheme is validated separately
```
fowczarek, line 11: Could we add the reason why the `scheme is validated separately` in this comment?

d-wysocki: It's took directly from django, that will be okay if I will add source of this piece of code?

maarcingebala (approved):

PR #6984, opened 2021-03-09
Handle allocations when editing order

tomaszszymanski129:
I want to merge this change because it handles allocations when order line items change.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


IKarbowiak (approved):

saleor/graphql/order/mutations/orders.py
```diff
@@ -670,6 +671,7 @@ def perform_mutation(cls, _root, info, **data):
         cls.validate_variants(order, variants)
 
         lines = cls.add_lines_to_order(order, lines_to_add)
+        allocate_lines(lines)
```
korycins, line 674: Correct me if I am wrong but you unify the logic for unconfirmed and draft orders. Am I right? I don't think that it is expected behaviour to allocate draft order in the same way as the unconfirmed order.

fowczarek:
When a Staff user tries to add more items than the available dashboard display an unreadable error.
image

saleor/warehouse/management.py
```diff
@@ -219,7 +219,21 @@ def increase_stock(
 
 
 @transaction.atomic
-def decrease_stock(order_lines_info: Iterable["OrderLineData"]):
+def increase_allocation(
+    line_info,
```
maarcingebala, line 223: Missing type annotation.

saleor/order/utils.py
```diff
@@ -367,9 +384,35 @@ def add_gift_card_to_order(order, gift_card, total_price_left):
...
+        return
+
+    if not get_order_lines_with_track_inventory([line_info]):
+        return
+
+    if old_quantity < new_quantity:
+        increase_allocation(line_info, new_quantity - old_quantity)
+    else:
+        line_info.quantity = old_quantity - new_quantity
+        decrease_allocations([line_info])
```
maarcingebala, line 407: I'm wondering if we could unify somehow `increase_allocation` and `descrease_allocations` - one is accepting a single line, while the other is accepting a list of lines. One updates `line_info.quantity` while the other does not and we do it in line 406 manually. Shouldn't `decrease_allocations` be moved to `warehouse/management.py` where other similar functions live?

saleor/order/utils.py
```diff
@@ -367,9 +384,35 @@ def add_gift_card_to_order(order, gift_card, total_price_left):
     return total_price_left
 
 
-def change_order_line_quantity(user, line, old_quantity, new_quantity):
+def decrease_allocations(lines_info):
+    """Decreate allocations for provided order lines."""
+    if not get_order_lines_with_track_inventory(lines_info):
+        return
+    decrease_stock(lines_info, update_stocks=False)
```
maarcingebala, line 391: Shouldn't we decrease stock only for lines with inventory tracking?

saleor/warehouse/management.py
```diff
@@ -266,9 +280,12 @@ def decrease_stock(order_lines_info: Iterable["OrderLineData"]):
             "quantity_allocated__sum"
         ]
 
-    _decrease_stocks_quantity(
-        order_lines_info, variant_and_warehouse_to_stock, quantity_allocation_for_stocks
-    )
+    if update_stocks:
```
maarcingebala, line 283: Maybe a comment would be useful to explain in what cases we want to enter this logic?

saleor/warehouse/management.py
```diff
@@ -219,7 +219,27 @@ def increase_stock(
 
 
 @transaction.atomic
-def decrease_stock(order_lines_info: Iterable["OrderLineData"]):
+def increase_allocation(
+    order_line,
+    quantity: int,
+):
+    """Increase allocation for order line with provided quantity."""
+    allocated = order_line.allocations.all().aggregate(Sum("quantity_allocated"))
```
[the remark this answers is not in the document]
tomaszszymanski129: What exactly should we lock in here? After further looking into it - we might have an even bigger problem.
test case:
trigger `increase_allocation` by dashboard call and stop after delete of allocations (before it calls `allocate_stocks`)
in different shell manually run `allocate_stocks` - the shell hangs when it eventually calls bulk_create
you release the dashboard call
shell process does it's thing, but the dashboard call ends with deadlock ☠️ 
```
django.db.utils.OperationalError: deadlock detected
DETAIL:  Process 2177 waits for ShareLock on transaction 178449; blocked by process 1992.
Process 1992 waits for ShareLock on transaction 178448; blocked by process 2177.
HINT:  See server log for query details.
CONTEXT:  while locking tuple (4,127) in relation "warehouse_stock"
```

saleor/warehouse/management.py
tomaszszymanski129: Added a docstring about what this param does

saleor/order/utils.py
tomaszszymanski129: Now both accept the lines and were moved to management, should be much cleaner now.

maarcingebala (approved):

saleor/warehouse/management.py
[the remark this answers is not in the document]
fowczarek: We should lock all stocks for which we changing allocations.

saleor/order/utils.py
```diff
@@ -346,6 +352,18 @@ def add_variant_to_order(order, variant, quantity, discounts=None):
             ]
         )
 
+    if line.order.is_unconfirmed():
```
korycins, line 355: What do you think to pass the flag explicitly as a function arg?
```
def add_variant_to_order(order, variant, quantity, discounts=None, allocate_stock=False):
```
IMO it should be more flexible than the current flow.

tomaszszymanski129: Yep, sounds good 👍🏻

fowczarek (approved):

PR #6997, opened 2021-03-11
change_currency: remove queries referencing deprecated currency field

sophie200:

these queries reference deprecated currency fields.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/core/management/commands/change_currency.py
+++ b/saleor/core/management/commands/change_currency.py
@@ -42,12 +42,8 @@ class Command(BaseCommand):
                 )
 
         Checkout.objects.update(currency=currency)
-        Voucher.objects.update(currency=currency)
         GiftCard.objects.update(currency=currency)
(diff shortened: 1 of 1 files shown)


maarcingebala (approved):

maarcingebala:
This change makes sense because it removes references to the currency field that was removed in some models, but on the other hand, the currency is now (in Saleor 3.0) stored in a few more models - "channel listings" - and it should be updated there as well. We should either fix the command and support other models as well. We can also discuss if this command is still useful in Saleor 3.0, where currency is defined per channel.

korycins (commented):
IMO, we should drop the command or fully extend it to be able to handle the channel slug as an input and cover the rest relations which uses the channel.currency.

maarcingebala:
Closing this PR for now, please reopen if there is capacity to work on it a bit more.

PR #6999, opened 2021-03-12
Update logic to use shipping zones available in given channel in order and shop module

IKarbowiak:
**Order:**
- Ensure that `draftOrderComplete` cannot be done, for the shipping method with shipping zone not available in order channel
- Ensure that `order.availableShippingMethods` returns only methods with shipping zones available in the order channel
- Update `_create_fulfillment_lines` method - get stocks for given channel
  - Ensure that only stocks for a given channel are used in `orderFulfill`
- Ensure that in `orderUpdateShipping` only shipping method available in the order channel can be used
- Ensure `get_valid_shipping_methods_for_order` returns only shipping methods with shipping zone available in the order channel

**Shop**
- Update `shop.availableShippingMethods` - return only shipping methods with shipping zones available in given channel

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/tests/test_fulfillment.py
+++ b/saleor/graphql/order/tests/test_fulfillment.py
@@ -330,6 +330,42 @@ def test_order_fulfill_zero_quantity(
     mock_create_fulfillments.assert_not_called()
 
 
+def test_order_fulfill_channel_without_shipping_zones(
+    staff_api_client,
+    order_with_lines,
+    permission_manage_orders,
+    warehouse,
+):
+    order = order_with_lines
+    order.channel.shipping_zones.clear()
+    query = ORDER_FULFILL_QUERY
+    order_id = graphene.Node.to_global_id("Order", order.id)
+    order_line = order.lines.first()
+    order_line_id = graphene.Node.to_global_id("OrderLine", order_line.id)
+    warehouse_id = graphene.Node.to_global_id("Warehouse", warehouse.pk)
+    variables = {
+        "order": order_id,
+        "input": {
+            "notifyCustomer": True,
+            "lines": [
+                {
+                    "orderLineId": order_line_id,
+                    "stocks": [{"quantity": 3, "warehouse": warehouse_id}],
+                },
+            ],
+        },
+    }
+    response = staff_api_client.post_graphql(
+        query, variables, permissions=[permission_manage_orders]
+    )
+    content = get_graphql_content(response)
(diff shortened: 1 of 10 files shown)


patrys:
Task linked: [SALEOR-2554 Update queries and logic to use shipping zones available in given channel](https://app.clickup.com/t/2549495/SALEOR-2554)

fowczarek (approved):

saleor/graphql/order/tests/test_fulfillment.py
```diff
@@ -330,6 +330,42 @@ def test_order_fulfill_zero_quantity(
...
+    }
+    response = staff_api_client.post_graphql(
+        query, variables, permissions=[permission_manage_orders]
+    )
+    content = get_graphql_content(response)
+    data = content["data"]["orderFulfill"]
+    assert len(data["orderErrors"]) == 1
+    error = data["orderErrors"][0]
+    assert error["field"] == "stocks"
+    assert error["code"] == OrderErrorCode.INSUFFICIENT_STOCK.name
```
korycins, line 366: Maybe it is hard to implement but currently, this can be misleading for someone who manages/implements the storefront/apps (I know this part is related to Dashboard, but the same approach we implement for checkout).
Shouldn't we have a separate error for missing shipping zone for channel or something? The current error will suggest me that the issue is with low stock or missing stock.
@maarcingebala @fowczarek  What do you think?

IKarbowiak: But it's a situation where we don't have stock for the chosen configuration. But I agree that it can be problematic and raising questions. Also, I'm not sure if we can apply a separate check for that situation without adding additional queries.

saleor/graphql/shop/resolvers.py
```diff
@@ -6,7 +6,8 @@
 
 def resolve_available_shipping_methods(info, channel_slug: str, address):
     available = ShippingMethod.objects.filter(
-        channel_listings__channel__slug=channel_slug
+        shipping_zone__channels__slug=channel_slug,
+        channel_listings__channel__slug=channel_slug,
```
maarcingebala, line 10: Maybe it would be useful to wrap this into a `for_channel` queryset? We have the same pattern for Stock model and it's easy to reuse and read.

maarcingebala (approved):

maarcingebala (approved):

PR #7005, opened 2021-03-13
Re-added query for fetching a specific gateway clientToken.

cipriancaba:
I want to merge this change because the payment gateway `get_client_token` generation was no longer exposed in the payments query. This also makes sure that the `user_id` is passed to the `get_client_token` even if there is no gateway specific user_id in the private meta

I've hit this issue while developing a payment gateway for [primer.io](https://primer.io) 
The use-case here is that the frontend needs to use a specific `client_token` that gets generated on the backend. This client token can be user-specific so it later allows fetching of user specific stored payment methods. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/payment/resolvers.py
+++ b/saleor/graphql/payment/resolvers.py
@@ -1,3 +1,5 @@
+import graphene
+
 from ...payment import gateway as payment_gateway
 from ...payment import models
 from ...payment.utils import fetch_customer_id
@@ -7,8 +9,10 @@ PAYMENT_SEARCH_FIELDS = ["id"]
 
 
 def resolve_client_token(user, gateway: str):
-    customer_id = fetch_customer_id(user, gateway)
-    return payment_gateway.get_client_token(gateway, customer_id)
+    gateway_customer_id = fetch_customer_id(user, gateway)
+    customer_id = gateway_customer_id if gateway_customer_id else graphene.Node.to_global_id(
+        "User", user.id) if user.id else None
+    return {'clientToken': payment_gateway.get_client_token(gateway, customer_id)}
 
 
 def resolve_payments(info, query):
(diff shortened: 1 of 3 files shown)


maarcingebala:
Hi @cipriancaba, there is currently a different way to return the client token. Example:

```graphql
{
  checkout(token: "example-token") {
    availablePaymentGateways {
      config {
        field
        value
      }
    }
  }
}
```
For each enabled payment gateway, you can fetch the `config` field which can return any key-value data. This config can be defined by overriding the `get_payment_config` function. Here is an example implementation for Braintree, that generates a new client token whenever the query is called:
 https://github.com/mirumee/saleor/blob/55bd0d1189f21554505703d51870201bfa526a3e/saleor/payment/gateways/braintree/plugin.py#L152

https://docs.saleor.io/docs/next/developer/checkout#creating-a-checkout-session

cipriancaba:
Hey @maarcingebala , that was the first thing I tried to use, but I need user-specific tokens. Eg, I need the user id of the customer initiating the payment that is not available in the get_payment_config. Also, the get_payment_config gets called on every payment query which means the token will get refreshed every single time, rather than reusing the existing one

maarcingebala:
`get_client_token` receives `token_config` param with `customer_id` inside, I think we should rather try to refactor the signature of the `get_payment_config` function and allow passing the config with customer ID there as well, maybe that would be the best solution?

cipriancaba:
I've tried to refactor the `get_payment_config` the you mentioned @maarcingebala, but I feel this adds a lot more complexity and needs changing lots of the tests as well.

Also, I've realised it doesn't really cover my usecase, where the payment gateway - in this case primer - needs to access a client specific token to list the use saved payment methods. They offer a "vault" as a service, where they basically manage the payment methods for you, and you identify them with your customer id and server api key. I need to list these payment methods in the user account page, so basically I would need a client specific token in an are that's not the actual checkout, and that wouldn't work with the change you suggested, unless I'm missing something?

PR #7044, opened 2021-03-22
Payment webhooks

maarcingebala:
This PR adds support for synchronous webhooks for payment events (payment process, authorize, capture, void, refund, confirm). This allows creating apps that can be used as payment gateways and handle payments outside of Saleor, via receiving a request from Saleor, executing the app's specific payment logic, and sending the result back in an HTTP response.

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/plugins/manager.py
```diff
@@ -31,19 +39,40 @@
...
+METHOD_NAME_REFUND_PAYMENT = "refund_payment"
+METHOD_NAME_VOID_PAYMENT = "void_payment"
+METHOD_NAME_CONFIRM_PAYMENT = "confirm_payment"
+METHOD_NAME_PROCESS_PAYMENT = "process_payment"
+
+
+def _get_webhook_for_method(webhooks_queryset, method_name):
+    """Get a webhook instance for a PluginManager's payment method."""
+    qs = webhooks_queryset.filter(is_active=True)
+    if method_name == METHOD_NAME_AUTHORIZE_PAYMENT:
```
korycins, line 59: Just visual improvment but if you create a dict map like:
```
{
 "process_payment": WebhookEventType.PAYMENT_PROCESS,
 ...
 }
 if method_name in payment_dict:
     qs = qs.filter(events__event_type=payment_dict[method_name])
 ```

File: saleor/plugins/webhook/utils.py [hunk 2]
```diff
@@ -0,0 +1,79 @@
...
+
+def webhook_response_to_gateway_response(
+    payment_information: "PaymentData",
+    response: "RequestsResponse",
+    transaction_kind: "str",
+) -> "GatewayResponse":
+    response_json = response.json()
+
+    error = response_json.get("error")
+    is_success = response.status_code == 200 and not error
```
korycins, line 56: is_succes should be taken from json response

saleor/plugins/webhook/utils.py
```diff
@@ -0,0 +1,79 @@
...
+
+    payment_method_info = PaymentMethodInfo(
```
korycins, line 58: I would consider to place all `PaymentMethodInfo` data to json separate object. Something like this:
```
response = {
  payment_method_info: {
      ...
  }
  ```

File: saleor/plugins/webhook/utils.py [hunk 4]
```diff
@@ -0,0 +1,79 @@
...
+
+    return GatewayResponse(
+        action_required=response_json.get("action_required", False),
+        action_required_data=response_json.get("action_required_data"),
+        amount=payment_information.amount,
+        currency=payment_information.currency,
+        customer_id=response_json.get("customer_id"),
+        error=error,
+        is_success=is_success,
+        kind=transaction_kind,
```
korycins, line 75: I would say that we should try to fetch the kind from the response. We can use `transaction_kind` as a fallback.

saleor/webhook/event_types.py
```diff
@@ -65,6 +73,13 @@ class WebhookEventType:
         PAGE_CREATED: "Page Created",
         PAGE_UPDATED: "Page Updated",
         PAGE_DELETED: "Page Deleted",
+        PAYMENT_AUTHORIZE: "Authorize payment",
```
korycins, line 76: We should consider new permission related to payment actions.

saleor/plugins/webhook/tasks.py
```diff
@@ -134,3 +160,27 @@ def send_webhook_request(webhook_id, target_url, secret, event_type, data):
         target_url,
         event_type,
     )
+
+
+def send_webhook_request_sync(target_url, secret, event_type, data):
```
korycins, line 165: IMO it is the same logic as `send_webhook_request`. Why we can't extend `send_webhook_request` to return response from request by default and use it in sync logic as a normal function?
```
def trigger_webhook_sync(event_type, data, webhooks_qs=None):
    ....
    response = send_webhook_request(
            webhook.pk, webhook.target_url, webhook.secret_key, event_type, data
        )
```

saleor/plugins/webhook/utils.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+    return gateways
+
+
+def webhook_response_to_gateway_response(
+    payment_information: "PaymentData",
+    response: "RequestsResponse",
+    transaction_kind: "str",
+) -> "GatewayResponse":
+    response_json = response.json()
```
korycins, line 53: I would wrap the response.json() in try-except. There is a big risk that it will return different structure than json

saleor/plugins/webhook/utils.py
korycins: Is it still valid?

saleor/plugins/webhook/utils.py
korycins: It is a PaymentGateway resposnibility to define proper transaction kind. On this lvl, Saleor doesn't know if the returned kind will be Auth, Capture etc.

saleor/webhook/event_types.py
```diff
@@ -92,6 +107,23 @@ class WebhookEventType:
...
+        PAYMENT_AUTHORIZE,
+        PAYMENT_CAPTURE,
+        PAYMENT_CONFIRM,
+        PAYMENT_LIST_GATEWAYS,
+        PAYMENT_PROCESS,
+        PAYMENT_REFUND,
+        PAYMENT_VOID,
     ]
 
     PERMISSIONS = {
```
korycins, line 129: Remember to assigning the permissions for payment webhook

saleor/plugins/webhook/utils.py
maarcingebala: It is a separate object, it is returned exactly as you proposed. Or I don't understand the comment.

saleor/plugins/webhook/utils.py
maarcingebala: I assumed that as long the `error` is empty and the requests was successful (200) than `is_success` is `True`.

We can change it to return `is_success` from the webhook response, but I think we also want to receive some error messages from the gateway. So if `error` is present, `is_success` is automatically false - no need to return two fields to know that IMO.

saleor/plugins/manager.py
```diff
@@ -574,32 +570,49 @@ def get_active_plugins(self, plugins=None) -> List["BasePlugin"]:
...
+        plugins = self.plugins
+        if active_only:
+            plugins = self.get_active_plugins()
+
+        # WebhookPlugin has the `process_payment` method defined but we don't want to
+        # list it as a payment gateway.
+        exclude_webhook_plugin = (
+            lambda plugin: plugin.PLUGIN_ID != WebhookPlugin.PLUGIN_ID
+        )
+        payment_plugins = [
```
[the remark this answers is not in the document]
maarcingebala: What exactly you want to simplify? Drop the lambda function or remove a condition from the `if` statement? We need to check two conditions:
- if the plugin has the "process_payment" method
- if it is not the WebhookPlugin

saleor/plugins/webhook/tasks.py
```diff
@@ -43,13 +46,35 @@ def trigger_webhooks_for_event(event_type, data):
...
     for webhook in webhooks:
         send_webhook_request.delay(
             webhook.pk, webhook.target_url, webhook.secret_key, event_type, data
         )
 
 
+def trigger_webhook_sync(event_type, data, webhooks_qs=None):
+    """Send a synchronous webhook request."""
+    webhooks = _get_webhooks_for_event(event_type, webhooks_qs)
+    webhook = webhooks.first()
```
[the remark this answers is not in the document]
maarcingebala: We spoke about it on a call - we agreed that using the first webhook in a row is fine, as long as we document it.

saleor/plugins/webhook/utils.py
korycins: Okay 👍

saleor/plugins/webhook/utils.py
korycins: I was meaning that now you have flat response object, like:
```
response = response.json()
assert response == {
...
kind: capture,
error: None, 
payment_method_brand: "something",
payment_method_exp_month: "something2",
payment_method_last_4:"something3"
...
}
```
And my proposition was that, maybe we could move all payment_method data in separate json data. So your response from payment gateway would be:
```
response = response.json()
assert response == {
...
kind: capture,
error: None,
payment_method:{ 
    brand: "something",
    exp_month: "something2",
    last_4:"something3"
}
...
}
```

saleor/plugins/manager.py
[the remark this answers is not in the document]
korycins: Nevermind. I think that I missed that the condition contains two ifs.

saleor/plugins/webhook/plugin.py
```diff
@@ -213,3 +219,120 @@ def page_deleted(self, page: "Page", previous_value: Any) -> Any:
...
+        previous_value,
+        **kwargs
+    ) -> "GatewayResponse":
+        if not self.active:
+            return previous_value
+
+        app = None
+        app_pk = kwargs.get("payment_app")
+        if app_pk is not None:
+            app = App.objects.for_event_type().filter(pk=app_pk).first()
```
korycins, line 239: afaik event_type missing for `for_event_type`

saleor/plugins/webhook/utils.py
maarcingebala: Ah, you meant the incoming response, not the payload - we can change it :)

saleor/plugins/base_plugin.py
```diff
@@ -91,6 +91,10 @@ def __init__(
     def __str__(self):
         return self.PLUGIN_NAME
 
+    def check_plugin_id(self, plugin_id: str) -> bool:
```
korycins, line 94: Don't you think that it would be good to convert this method to `classmethod`?  Sometimes we would like to check plugin_Id without creating a plugin  instance

saleor/plugins/webhook/plugin.py
```diff
@@ -37,6 +47,13 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.active = True
 
+    def check_plugin_id(self, plugin_id: str) -> bool:
+        is_webhook_plugin = super().check_plugin_id(plugin_id)
+        if not is_webhook_plugin:
```
korycins, line 52: Before we enter to this condition, we should check the structure of plugin_id. LIke - make sure that the plugin_id is in a structure like "app:<id>:<something>. Now we will call `from_payment_app_id` for all plugins different than WebhookPlugin.

saleor/plugins/webhook/plugin.py
```diff
@@ -214,3 +231,123 @@ def page_deleted(self, page: "Page", previous_value: Any) -> Any:
...
+        payment_app_data = from_payment_app_id(payment_information.gateway)
+
+        if payment_app_data is not None:
+            app = (
+                App.objects.for_event_type(event_type)
+                .filter(pk=payment_app_data.app_pk)
+                .first()
+            )
+
+        if not app:
```
korycins, line 258: It is only a suggestion but I would add here some logger or/and insert more details for this error. Now it doesn't give us any feedback. We could add some details like: App ID, event type etc.

saleor/plugins/webhook/plugin.py
```diff
@@ -214,3 +231,123 @@ def page_deleted(self, page: "Page", previous_value: Any) -> Any:
...
+
+    def get_payment_gateways(
+        self,
+        currency: Optional[str],
+        checkout: Optional["Checkout"],
+        previous_value,
+        **kwargs
+    ) -> List["PaymentGateway"]:
+        gateways = []
+        apps = App.objects.for_event_type("payment_list_gateways").prefetch_related(
```
korycins, line 275: Can we use defined const vars from saleor/webhook/event_type instead of str?

saleor/plugins/webhook/tasks.py
```diff
@@ -139,3 +167,32 @@ def send_webhook_request(webhook_id, target_url, secret, event_type, data):
...
+            )
+            response_data = response.json()
+        except RequestException as e:
+            logger.debug("[Webhook] Failed request to %r: %r.", target_url, e)
+        except JSONDecodeError as e:
+            logger.debug(
+                "[Webhook] Failed parsing JSON response from %r: %r.", target_url, e
+            )
+        else:
+            logger.warning("[Webhook] Success response from %r.", target_url)
```
korycins, line 195: Are you sure that we are going to WARNING all success response? 😄

fowczarek: :policeman:

saleor/payment/migrations/0029_alter_payment_options.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 3.2.2 on 2021-05-19 09:50
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
fowczarek, line 6: Should we add this permission to staff users?

saleor/plugins/webhook/plugin.py
```diff
@@ -214,3 +231,123 @@ def page_deleted(self, page: "Page", previous_value: Any) -> Any:
             return previous_value
         page_data = generate_page_payload(page)
         trigger_webhooks_for_event.delay(WebhookEventType.PAGE_DELETED, page_data)
+
+    # Webhook payment functions
+
```
fowczarek, line 236:
```suggestion
```

saleor/plugins/webhook/tasks.py
```diff
@@ -25,16 +31,19 @@ class WebhookSchemes(str, Enum):
...
+    required_permission = WebhookEventType.PERMISSIONS.get(event_type)
     if required_permission:
-        app_label, codename = required_permission.split(".")
+        app_label, codename = required_permission.value.split(".")
         permissions["app__permissions__content_type__app_label"] = app_label
         permissions["app__permissions__codename"] = codename
 
-    webhooks = Webhook.objects.filter(
+    if webhooks is None:
+        webhooks = Webhook.objects.all()
```
fowczarek, line 44: Should we filter out not active webhooks?

saleor/plugins/webhook/tasks.py
```diff
@@ -43,13 +52,31 @@ def trigger_webhooks_for_event(event_type, data):
...
+    webhooks = _get_webhooks_for_event(event_type)
     for webhook in webhooks:
         send_webhook_request.delay(
             webhook.pk, webhook.target_url, webhook.secret_key, event_type, data
         )
 
 
+def trigger_webhook_sync(event_type: str, data: str, app: "App"):
+    """Send a synchronous webhook request."""
+    webhooks = _get_webhooks_for_event(event_type, app.webhooks.all())
```
fowczarek, line 70: Should we filter out not active webhooks?

saleor/payment/migrations/0029_alter_payment_options.py
```diff
@@ -0,0 +1,20 @@
...
+    dependencies = [
+        ("payment", "0028_drop_searchable_key"),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name="payment",
+            options={
+                "ordering": ("pk",),
+                "permissions": (("manage_payments", "Manage payments"),),
```
IKarbowiak, line 17: Shouldn't we assign this perm to some permission groups?

fowczarek (approved):

saleor/plugins/webhook/tasks.py
maarcingebala: `_get_webhooks_for_event` filters webhooks by `is_active=True` internally.

saleor/plugins/webhook/tasks.py
maarcingebala: It's done in line 47.

saleor/payment/migrations/0029_alter_payment_options.py
maarcingebala: I could do that, but this permission is currently not used by users - no mutations or queries require it. It is only used by apps.

saleor/plugins/webhook/plugin.py
maarcingebala: We will not - all other plugins have a different implementation of this method where we only compare passed `plugin_ID` with `self.PLUGIN_ID`. Only WebhookPlugin has a custom method to check the ID and only this plugin can run `from_payment_app_id`.

korycins: 👍

IKarbowiak (approved):

korycins (approved):

fowczarek (approved):

fowczarek (approved):

PR #7051, opened 2021-03-23
Trigger produt_updated webhook for collections and categories

d-wysocki:
I want to merge this change because...I want to trigger `PRODUCT_UPDATED` webhook for

- `CollectionCreate`
- `CollectionDelete`
- `CollectionAddProducts`
- `CollectionRemoveProducts`
- `CollectionBulkDelete`
- `CategoryDelete`
- `CategoryBulkDelete`

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-2245 Removing and adding product in collections/categories does not trigger PRODUCT_UPDATE webhook](https://app.clickup.com/t/2549495/SALEOR-2245)

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -70,6 +81,25 @@ class Meta:
...
+            ids,
+            manager=info.context.plugins,
+            **data,
+        )
+        return response
+
+    @classmethod
+    def bulk_action(cls, queryset, manager):
+        collections_ids = queryset.values_list("id", flat=True)
+        products = list(models.Product.objects.filter(collections__in=collections_ids))
```
korycins, line 98: IMO, you need distinct here.
image

saleor/graphql/product/mutations/products.py
```diff
@@ -250,6 +256,10 @@ class Meta:
         error_type_class = CollectionError
         error_type_field = "collection_errors"
 
+    @classmethod
+    def post_save_action(cls, info, instance, cleaned_input):
```
korycins, line 260: Do you need this empty method?

saleor/graphql/product/mutations/products.py
```diff
@@ -270,7 +280,14 @@ class Meta:
 
     @classmethod
     def perform_mutation(cls, _root, info, **kwargs):
+        node_id = kwargs.get("id")
+
+        instance = cls.get_node_or_error(info, node_id, only_type=Collection)
+        products = list(instance.products.all())
```
korycins, line 286: Do you need to convert them to list?
What about prefetches?  variant, categories, collections,  attributes(?)

korycins (commented):
IMO Missing prefetches can cause a huge amount of DB queries here.

saleor/graphql/product/mutations/products.py
```diff
@@ -228,6 +228,12 @@ def save(cls, info, instance, cleaned_input):
         if cleaned_input.get("background_image"):
             create_collection_background_image_thumbnails.delay(instance.pk)
 
+    @classmethod
+    def post_save_action(cls, info, instance, cleaned_input):
+        products = instance.products.all()
```
fowczarek, line 233: Should we prefetch any data?

saleor/graphql/product/mutations/products.py
```diff
@@ -393,6 +410,11 @@ def perform_mutation(cls, _root, info, collection_id, products):
             update_products_discounted_prices_of_catalogues_task.delay(
                 product_ids=[pq.pk for pq in products]
             )
+        transaction.on_commit(
+            lambda: [
+                info.context.plugins.product_updated(product) for product in products
```
fowczarek, line 415: Should we make queryset with prefetched needed data from `products` list?

saleor/graphql/product/mutations/products.py
```diff
@@ -439,6 +461,8 @@ def perform_mutation(cls, _root, info, collection_id, products):
         )
         products = cls.get_nodes_or_error(products, "products", only_type=Product)
         collection.products.remove(*products)
+        for product in products:
```
fowczarek, line 464: Should we make queryset with prefetched needed data from products list?

saleor/product/utils/__init__.py
```diff
@@ -51,8 +51,12 @@ def delete_categories(categories_ids: List[str]):
     ProductChannelListing.objects.filter(product__in=products).update(
         is_published=False, publication_date=None
     )
-    product_ids = list(products.values_list("id", flat=True))
+    products = list(products)
```
fowczarek, line 54: Prefetch?

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -53,8 +53,19 @@ class Meta:
         error_type_field = "product_errors"
 
     @classmethod
-    def bulk_action(cls, queryset):
-        delete_categories(queryset.values_list("pk", flat=True))
+    def perform_mutation(cls, _root, info, ids, **data):
```
maarcingebala, line 56: Why do we override `perform_mutation` here? To pass the manager as an extra argument? It is available in the `info` object anyway.

d-wysocki: I override this method because, I wanted to pass manager to `bulk_action` method.
`perform_mutation` method in BaseBulkMutation takes `**data` parameter and pass it to `bulk_action`. So I added managet to `data` argument.

saleor/graphql/product/mutations/products.py
d-wysocki: I converted this into list because, this is delete mutation. Without casting to a list, after deleting collection, `instance` will doesn't had `products`.

I will add prefetch.

saleor/graphql/product/mutations/products.py
d-wysocki: Yes, I will add prefetch.

saleor/graphql/product/mutations/products.py
d-wysocki: Yes, I will add prefetch.

saleor/graphql/product/bulk_mutations/products.py
korycins: So maybe we should extend bulk_action to handle info by default? All future bulk mutations which use manager will require the same action,

saleor/graphql/product/mutations/products.py
d-wysocki: Yes, `CollectionUpdate` inherits from `CollectionCreate`, so I needed to clear this method to avoid triggering `PRODUCT_UPDATED` webhook when collection is updated.

saleor/graphql/product/mutations/products.py
korycins: > I converted this into list because, this is delete mutation. Without casting to a list, after deleting collection, `instance` will doesn't had `products`.

My bad, I missed the name of the mutation.

saleor/graphql/product/bulk_mutations/products.py
maarcingebala: Yep, was going to write the same as what Maciek suggests.

d-wysocki: Ok, I will change that.

saleor/graphql/product/mutations/products.py
korycins: So in that case I would call the product_updated inside a CollectionCreate.perform_mutation
or at least add the comment why we have empty method here <- which imho is not the best way.
@fowczarek  What do you think?

d-wysocki: Please correct me if im wrong, but `CollectionUpdate` also inherits `perform_mutation` from `CollectionCreate`.

saleor/graphql/product/bulk_mutations/products.py
IKarbowiak: I already added info to core `bulk_action` in #7052, it looks like we can merge it, so the problem will be solved.

saleor/graphql/product/mutations/products.py
d-wysocki: I added prefetch_related.

saleor/product/utils/__init__.py
d-wysocki: I added prefetch_related.

saleor/graphql/product/mutations/products.py
korycins: Oh, yes you're right. I missed thath!

saleor/graphql/product/mutations/products.py
```diff
@@ -385,14 +408,26 @@ def perform_mutation(cls, _root, info, collection_id, products):
         collection = cls.get_node_or_error(
             info, collection_id, field="collection_id", only_type=Collection
         )
-        products = cls.get_nodes_or_error(products, "products", Product)
+        products = cls.get_nodes_or_error(
+            products,
+            "products",
+            Product,
+            qs=models.Product.objects.prefetch_related(
+                "attributes", "collections", "variants", "category"
```
korycins, line 416: Please double check the `attributes` prefetch. I think that it misses some relations.

d-wysocki: Good catch, thanks. I will go thought all product_updated and product_variant_updated to add prefetch.

saleor/graphql/meta/mutations.py
```diff
@@ -114,6 +115,10 @@ def perform_model_extra_actions(cls, root, info, **data):
         """Run extra metadata method based on mutating model."""
         type_name, _ = graphene.Node.from_global_id(data["id"])
         if MODEL_EXTRA_METHODS.get(type_name):
+            if type_name == product_models.Product.__name__:
+                data[
+                    "qs"
+                ] = product_models.Product.objects.prefetched_product_for_webhook()
             instance = cls.get_instance(info, **data)
```
korycins, line 122: Why you can't call this prefetch inside the function responsible for Product extra actions?

saleor/graphql/product/mutations/products.py
korycins: But at least comment why we have empty method here

saleor/graphql/product/mutations/products.py
```diff
@@ -270,7 +280,14 @@ class Meta:
 
     @classmethod
     def perform_mutation(cls, _root, info, **kwargs):
+        node_id = kwargs.get("id")
+
+        instance = cls.get_node_or_error(info, node_id, only_type=Collection)
+        products = list(instance.products.prefetched_product_for_webhook().all())
```
korycins, line 286: I don't think that you need `.all()` here.

saleor/product/models.py
```diff
@@ -299,6 +299,16 @@ def sort_by_attribute(
             f"{ordering}name",
         )
 
+    def prefetched_product_for_webhook(self):
```
korycins, line 302: IMO `prefetched_for_webhook` has enough information. No need to duplicate the `product` word
`Product.objects. prefetched_for_webhook()`

saleor/product/models.py
```diff
@@ -299,6 +299,16 @@ def sort_by_attribute(
...
+        return self.prefetch_related(
+            "attributes__values",
+            "attributes__assignment__attribute",
+            "collections",
+            "variants__stocks",
```
korycins, line 307: `variants__stocks` is not needed here.

saleor/product/models.py
```diff
@@ -407,6 +417,15 @@ def annotate_quantities(self):
             ),
         )
 
+    def prefetch_variant_updated(self):
```
korycins, line 420: Would be good to keep some consistency here. You have `prefetched_for_webhook` and here `prefetch_variant_updated`. I don't think that everyone will connect that this method is prefetch for webhook with name  PRODUCT_VARIANT_UPDATED`

saleor/product/utils/__init__.py
```diff
@@ -44,15 +44,19 @@ def delete_categories(categories_ids: List[str]):
     categories = Category.objects.select_for_update().filter(pk__in=categories_ids)
     categories.prefetch_related("products")
 
-    products = Product.objects.none()
+    products = Product.objects.prefetched_product_for_webhook().none()
```
korycins, line 47: Are you aware that you're applying prefetches on the qs which is empty?

saleor/graphql/product/mutations/products.py
```diff
@@ -662,6 +697,7 @@ def save(cls, info, instance, cleaned_input):
 
     @classmethod
     def post_save_action(cls, info, instance, cleaned_input):
+        # TODO: add prefetch related
```
fowczarek, line 700: We should add this prefetch before the merge.

saleor/product/models.py
```diff
@@ -299,6 +299,12 @@ def sort_by_attribute(
             f"{ordering}name",
         )
 
+    def prefetched_for_webhook(self):
+        return self.prefetch_related(
+            "attributes__values",
+            "attributes__assignment__attribute",
```
fowczarek, line 305: Are you sure that's all the required fields for `generate_product_payload`?

d-wysocki: For single object we need only prefetch `attributes`, when we iterating over object then we need to prefetch everything related to that object.

maarcingebala:
From my perspective, the code looks good, but the number of DB queries that this PR adds (according to the benchmark) worries me. Are these expected results?

maarcingebala (approved):

IKarbowiak (approved):

maarcingebala:
@fowczarek @korycins This PR is waiting for your final review, can you take a look?

fowczarek (approved):

saleor/product/models.py
```diff
@@ -407,6 +421,12 @@ def annotate_quantities(self):
             ),
         )
 
+    def prefetched_for_webhook(self):
+        return self.prefetch_related(
+            "attributes__values",
+            "attributes__assignment__attribute",
```
korycins, line 427: Just to confirm: What about `media`? I see that payload generated for variant uses it.

saleor/graphql/product/tests/test_collection.py
```diff
@@ -920,6 +1018,40 @@ def test_remove_products_from_collection(
...
+                        totalCount
+                    }
+                }
+            }
+        }
+    """
+    collection.products.add(*product_list)
+    collection_id = to_global_id("Collection", collection.id)
+    product_ids = [to_global_id("Product", product.pk) for product in product_list]
+    no_products_before = collection.products.count()
```
korycins, line 1044: The name suggests that the value here is 0.  Not sure that it is correct as, in line 1051, we have subtraction from this value.

saleor/product/models.py
d-wysocki: Yes, I tested it.

korycins: We have N queries here, where N is the number of media assigned to variant. 
First case: Variant without prefetch and some media objects. For each media obj, it generates one query.
Second: Prefetch with const number of queries
image

d-wysocki: Fixed.

korycins (approved):

PR #7056, opened 2021-03-24
Fix doubling price in checkout for products without tax

IKarbowiak:
The problem was caused by `_calculate_checkout_subtotal` in avatax plugin. I decided to drop `calculate_checkout_subtotal` from plugins and left it only in `manager` as it was anyway calculated by the manager with the use of `calculate_checkout_line_total`. The plugin method `calculate_checkout_subtotal` was run only when all checkout lines have products with no charge taxes.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/plugins/manager.py
+++ b/saleor/plugins/manager.py
@@ -157,18 +157,10 @@ class PluginsManager(PaymentInterface):
             )
             for line_info in lines
         ]
-        default_value = base_calculations.base_checkout_subtotal(
-            line_totals, checkout_info.checkout.currency
-        )
+        currency = checkout_info.checkout.currency
+        total = sum(line_totals, zero_taxed_money(currency))
         return quantize_price(
-            self.__run_method_on_plugins(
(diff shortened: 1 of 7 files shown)


saleor/plugins/manager.py
```diff
@@ -157,18 +157,10 @@ def calculate_checkout_subtotal(
             )
             for line_info in lines
         ]
-        default_value = base_calculations.base_checkout_subtotal(
```
maarcingebala, line 160: Is `base_checkout_subtotal` needed after this change? I don't see that we use it anywhere else.

IKarbowiak: No, I forgot to drop it.

korycins (approved):

maarcingebala (approved):

fowczarek (approved):
Should we drop `calculate_checkout_subtotal` from `saleor/plugins/tests/sample_plugins.py`?

PR #7059, opened 2021-03-25
Rich text attribute input

piotrgrundas:
I want to merge this change because I would like to add simple text input attribute to the dashboard.

[Dashboard PR](https://github.com/mirumee/saleor-dashboard/pull/1023)

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/graphql/attribute/utils.py
```diff
@@ -370,20 +371,36 @@ def save(cls, instance: T_INSTANCE, cleaned_input: T_INPUT_MAP):
...
+                    attribute_values = cls._pre_save_values(attribute, attr_values)
+
+                associate_attribute_values_to_instance(
+                    instance, attribute, *attribute_values
                 )
-            elif attribute.input_type == AttributeInputType.REFERENCE:
-                attribute_values = cls._pre_save_reference_values(
-                    instance, attribute, attr_values
+
+            except DataError as e:
```
piotrgrundas, line 393: It can be thrown, if some1 passes string larger than the current `name` max_length.

saleor/graphql/attribute/utils.py
```diff
@@ -370,20 +371,36 @@ def save(cls, instance: T_INSTANCE, cleaned_input: T_INPUT_MAP):
...
+                errors.append(
+                    ValidationError(
+                        str(e),
+                        code=ProductErrorCode.INVALID.value,
+                        params={"attributes": [attr_values.global_id]},
+                    )
                 )
```
IKarbowiak, line 400: I don't see the reason to create the same error for every attribute value. I suggest using the list for collecting failing attribute values and then raising one error with the attribute list as params.

saleor/tests/fixtures.py
```diff
@@ -773,6 +773,22 @@ def color_attribute(db):  # pylint: disable=W0613
     return attribute
 
 
+@pytest.fixture
+def text_attribute(db):  # pylint: disable=W0613
+    attribute = Attribute.objects.create(
+        slug="text",
+        name="Text",
+        type=AttributeInputType.TEXT,
```
IKarbowiak, line 781: `type` defines if this is `PRODUCT` or `PAGE` attribute, it means if the attribute can be assigned to `product_type` or to `page_type`. So here you should use both fields: `type` and `input_type`.

Also, some more tests should be added, at least:
- tests for updating/creating product type and (maybe also for page type) with text attributes
- test for assigning text attribute to product and variant

saleor/tests/fixtures.py
```diff
@@ -773,6 +773,22 @@ def color_attribute(db):  # pylint: disable=W0613
     return attribute
 
 
+@pytest.fixture
+def text_attribute(db):  # pylint: disable=W0613
```
IKarbowiak, line 777: I don't think `# pylint: disable=W0613` comment is necessary.

saleor/graphql/attribute/utils.py
IKarbowiak: Sorry, I didn't notice that the message is the error message from the database, so it must be left like that.

saleor/graphql/attribute/utils.py
patrys, line 393: We need to validate data before passing it to the database. Database errors make the current database connection (or transaction) unusable until explicitly reset (or rolled back).

patrys (changes requested):
Validate data in input, don't rely on the database throwing an exception.

saleor/attribute/__init__.py
```diff
@@ -5,15 +5,17 @@ class AttributeInputType:
...
     CHOICES = [
         (DROPDOWN, "Dropdown"),
         (MULTISELECT, "Multi Select"),
         (FILE, "File"),
         (REFERENCE, "Reference"),
+        (TEXT, "Text"),
     ]
     # list of the input types that can be used in variant selection
-    ALLOWED_IN_VARIANT_SELECTION = [DROPDOWN]
+    ALLOWED_IN_VARIANT_SELECTION = [DROPDOWN, TEXT]
```
maarcingebala, line 18: Are we sure that we want to use "text" in variant selection? When an attribute is used in variant selection, it will be required to choose by the customer in the storefront to identify a variant; it will also appear in the variant creator in the dashboard and staff users will have to provide all text values for all variants to create them.

image

piotrgrundas: Good point 👍

maarcingebala (approved):

saleor/graphql/attribute/utils.py
```diff
@@ -531,6 +571,20 @@ def validate_standard_attributes_input(
...
+    variant_validation: bool,
+):
+    name_field = attribute.values.model.name.field  # type: ignore
+    attribute_id = attr_values.global_id
+    text = attr_values.values[0] if attr_values.values else ""
+
+    if len(text) > name_field.max_length:
+        attribute_errors[AttributeInputErrors.ERROR_MAX_LENGTH].append(attribute_id)
+
+
```
IKarbowiak, line 587: I think, that for `text` attribute validation you should use `validate_standard_attributes_input` and add there this new validation, as it could also regard another attribute, and validations from `validate_standard_attributes_input` will be also useful with `text` attributes.

saleor/tests/fixtures.py
```diff
@@ -773,6 +773,23 @@ def color_attribute(db):  # pylint: disable=W0613
     return attribute
 
 
+@pytest.fixture
+def text_attribute(db):
+    attribute = Attribute.objects.create(
+        slug="text",
+        name="Text",
+        type=AttributeInputType.TEXT,
```
IKarbowiak, line 781: Still `type` is incorrect, also there is a lack of a test for creating `ProductType` with `text_attribute`.

maarcingebala:
@piotrgrundas Please use meaningful commit names; you can also use `--amend` if making improvements to a previous commit for example.

piotrgrundas:
@maarcingebala obviously. That was just a transition commit, between devices.

saleor/graphql/product/tests/test_product.py
```diff
@@ -5496,67 +5556,45 @@ def test_product_type_query_only_variant_selections_value_set(
 """
 
 
-def test_product_type_create_mutation(
+def test_create_product_type_with_text_attribute(
```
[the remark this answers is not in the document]
piotrgrundas: Pycharm rebellion.. unintended :(

IKarbowiak (approved):

saleor/settings.py
```diff
@@ -263,6 +263,7 @@ def get_bool_from_env(name, default_value):
     "django_countries",
     "django_filters",
     "phonenumber_field",
+    "django_extensions",
```
IKarbowiak, line 266: Why?

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Validate discount value for percentage vouchers and sales - #7033 by @d-wysocki
 - Optimize children field on Category type - #7045 by @IKarbowiak
 - Added support for querying objects by metadata fields - #6683 by @LeOndaz
+- Add text attribute input - #7059 by @piotrgrundas
```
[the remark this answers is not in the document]
IKarbowiak: Together with PR name.

saleor/settings.py
piotrgrundas: Hmm git should ignore that :|

maarcingebala (approved):

IKarbowiak (approved):

IKarbowiak (changes requested):
`AttributeValueTranslate` mutation must be updated too.

patrys (approved):

PR #7085, opened 2021-04-01
Fix calculating total price on draft order

d-wysocki:
I want to merge this change because...I want to fix calculating total and weight for draft order while deleting variant or product And fix calculating order total while updating quantity of order line. Added validation for `draftOrderCreate` with negative quantity line.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-2645 Deleting Variant/Product removes order_line but doesn't recalculate order prices](https://app.clickup.com/t/2549495/SALEOR-2645)

saleor/graphql/product/mutations/products.py
```diff
@@ -688,14 +689,16 @@ def perform_mutation(cls, _root, info, **data):
...
+        lines = order_models.OrderLine.objects.select_related("order").filter(
+            variant__pk__id=variants_id, order__status=OrderStatus.DRAFT
         )
+        line_pks = [line.pk for line in lines]
+        orders = {line.order for line in lines}
         response = super().perform_mutation(_root, info, **data)
         # delete order lines for deleted variant
         order_models.OrderLine.objects.filter(pk__in=line_pks).delete()
+        for order in orders:
+            recalculate_order_prices(order, save=True)
```
korycins, line 701: IMO, you should call `recalculate_order` which recalculates the prices and also the weight of the order (which is also different now).

saleor/order/utils.py
```diff
@@ -290,15 +297,17 @@ def update_order_status(order):
 
 
 @transaction.atomic
-def add_variant_to_draft_order(order, variant, quantity, manager, discounts=None):
+def add_variant_to_draft_order(order, variant, quantity, info, discounts=None):
```
korycins, line 300: I am not convinced that the `info` is a better way for passing `manager` and `user`.  Even if this way would be better than explicit pass, the param should be a `info.context` instead of `info`. 
 I don't think that the utils method for `order` module should know anything about the graphql info object.

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -165,6 +165,15 @@ def clean_input(cls, info, instance, data):
...
             quantities = [line.get("quantity") for line in lines]
+            if not all(quantity > 0 for quantity in quantities):
+                raise ValidationError(
+                    {
+                        "quantity": ValidationError(
+                            "Ensure this value is greater than 0.",
+                            code=OrderErrorCode.ZERO_QUANTITY,
+                        )
+                    }
+                )
```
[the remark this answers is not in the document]
d-wysocki: Those lines aren't created yet, so we dont have IDS to return which line causing this error.

IKarbowiak (approved):

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -497,13 +499,11 @@ class Meta:
     @transaction.atomic
     def perform_mutation(cls, _root, info, ids, **data):
         _, pks = resolve_global_ids_to_primary_keys(ids, ProductVariant)
-        # get draft order lines for variants
```
fowczarek, line 500: In other places, you left this comment. In my opinion, we should remove all or don't remove any of this comment.

fowczarek (approved):

maarcingebala (approved):

maarcingebala (approved):

fowczarek (approved):

PR #7114, opened 2021-04-06
Catch unicode error

d-wysocki:
I want to merge this change because...I want to catch error when provided ID is invalid and raise ValidationError

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-2542 Catch can't decode ID.](https://app.clickup.com/t/2549495/SALEOR-2542)

maarcingebala (commented):
We already have two similar functions: `from_global_id_strict_type` and `get_database_id`. Maybe we could have instead one: `from_global_id_or_error` - it would accept optional `only_type` parameter; when provided we would check if the type of ID is the same as provided (just as `from_global_id_strict_type` does now). This would simplify our codebase.

I'm also wondering if instead of raising `ValidationError` we shouldn't use `GraphQLError`. `ValidationError` is for errors that were caused by the user and are a valid state of an application and the user can fix them in the UI. If you look at `from_global_id_strict_type`, it raises `ValidationError` when there is an issue with decoding ID from Base64. This shouldn't happen in UI in an application. I think GraphQLError could be more appropriate.

IKarbowiak (approved):

saleor/graphql/attribute/utils.py
```diff
@@ -118,7 +118,7 @@ def _resolve_attribute_nodes(
     @classmethod
     def _resolve_attribute_global_id(cls, error_class, global_id: str) -> int:
         """Resolve an Attribute global ID into an internal ID (int)."""
-        graphene_type, internal_id = from_global_id(global_id)  # type: str, str
+        graphene_type, internal_id = from_global_id_or_error(global_id)
         if graphene_type != "Attribute":
             raise ValidationError(
```
tomaszszymanski129, line 123: Maybe it could be handled by passing `only_type` to `from_global_id_or_error`?

saleor/graphql/checkout/mutations.py
```diff
@@ -892,7 +892,7 @@ def perform_mutation(cls, _root, info, checkout_id, store_source, **data):
                     field="checkout_id",
                 )
             except ValidationError as e:
-                checkout_token = from_global_id_strict_type(
+                _type, checkout_token = from_global_id_or_error(
                     checkout_id, Checkout, field="checkout_id"
```
tomaszszymanski129, line 896: inconsistent signature, here we pass Checkout as arg, elswere as `only_type` kwarg.

tomaszszymanski129 (approved):

saleor/graphql/core/mutations.py
```diff
@@ -157,10 +157,10 @@ def get_node_or_error(cls, info, node_id, field="id", only_type=None, qs=None):
 
         try:
             if only_type is not None:
-                pk = from_global_id_strict_type(node_id, only_type, field=field)
+                _type, pk = from_global_id_or_error(node_id, only_type, field=field)
```
korycins, line 160: IMO, this `if` is not needed here. You can handle this by single call `from_global_id_or_error(node_id, only_type=only_typre, field=field)`

korycins (approved):

saleor/graphql/core/utils/__init__.py
```diff
@@ -119,3 +97,24 @@ def validate_required_string_field(cleaned_input, field_name: str):
     else:
         raise ValidationError(f"{field_name.capitalize()} is required.")
     return cleaned_input
+
+
+def from_global_id_or_error(
+    id: str, only_type: Union[ObjectType, str] = None, field: str = "id"
+):
+    """Resolve id from global or raise ValidationError.
```
maarcingebala, line 105:
```suggestion
    """Resolve database ID from global ID or raise ValidationError.
```

saleor/graphql/core/utils/__init__.py
```diff
@@ -119,3 +97,24 @@ def validate_required_string_field(cleaned_input, field_name: str):
...
+
+    Optional validate object type.
```
maarcingebala, line 107:
```suggestion
    Optionally validate the object type, if `only_type` is provided.
```

maarcingebala (approved):
Suggested minor improvement in the docstring, but looks very good in general.

fowczarek (approved):

PR #7118, opened 2021-04-07
Feature/extend manager to use channel slug

korycins:
I want to merge this change because it adds logic to plugins manager responsible for distributing the hook calls for plugin assigned to the channels

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/plugins/mutations.py
+++ b/saleor/graphql/plugins/mutations.py
@@ -31,6 +31,10 @@ class PluginUpdate(BaseMutation):
 
     class Arguments:
         id = graphene.ID(required=True, description="ID of plugin to update.")
+        channel = graphene.String(
+            required=False,
+            description="Slug of a channel for which the data should be modified.",
+        )
         input = PluginUpdateInput(
             description="Fields required to update a plugin configuration.",
             required=True,
@@ -45,6 +49,7 @@ class PluginUpdate(BaseMutation):
     @classmethod
     def perform_mutation(cls, root, info, **data):
         plugin_id = data.get("id")
+        channel_slug = data.get("channel")
         data = data.get("input")
         manager = info.context.plugins
         plugin = manager.get_plugin(plugin_id)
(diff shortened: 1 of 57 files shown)


fowczarek (approved):

saleor/graphql/plugins/mutations.py
```diff
@@ -45,6 +49,7 @@ class Meta:
     @classmethod
     def perform_mutation(cls, root, info, **data):
         plugin_id = data.get("id")
+        channel_slug = data.get("channel")
         data = data.get("input")
         manager = info.context.plugins
         plugin = manager.get_plugin(plugin_id)
```
IKarbowiak, line 55: Shouldn't we have here the same channel validation as it was added to `PaymentInitialize`?

korycins: No. We still have some plugins which are configurable globally (like Emails, Webhooks). PaymentPlugins by default are configurable by channel. 
So here, it is fine if I will receive the channel as a None value

patrys: Slightly off-topic, but wouldn't it make sense to configure transactional emails on a per-channel basis?

saleor/graphql/shop/types.py
```diff
@@ -207,8 +216,12 @@ class Meta:
         )
 
     @staticmethod
-    def resolve_available_payment_gateways(_, info, currency: Optional[str] = None):
-        return info.context.plugins.list_payment_gateways(currency=currency)
+    def resolve_available_payment_gateways(
+        _, info, currency: Optional[str] = None, channel: Optional[str] = None
+    ):
+        return info.context.plugins.list_payment_gateways(
```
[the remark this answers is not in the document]
korycins: Probably I will drop currencies in separate PR

IKarbowiak (approved):

fowczarek (approved):

saleor/graphql/plugins/mutations.py
korycins: We can convert them to per channel but I would prefer to do that in separate PR

PR #7152, opened 2021-04-12
Allow passing metadata to `accountRegister` mutation

piotrgrundas:
I want to merge this change because I want to support passing public metadata to:
- accountUpdate
- accountRegister

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/account/mutations/account.py
+++ b/saleor/graphql/account/mutations/account.py
@@ -40,7 +40,9 @@ class AccountRegisterInput(graphene.InputObjectType):
         LanguageCodeEnum, required=False, description="User language code."
     )
     metadata = graphene.List(
-        MetadataInput, description="User public metadata.", required=False
+        graphene.NonNull(MetadataInput),
+        description="User public metadata.",
+        required=False,
     )
 
 
@@ -70,7 +72,7 @@ class AccountRegister(ModelMutation):
     @classmethod
     def clean_input(cls, info, instance, data, input_cls=None):
         data["metadata"] = {
-            item["key"]: item["value"] for item in data.get("metadata", [])
+            item["key"]: item["value"] for item in data.get("metadata") or []
         }
         if not settings.ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL:
             return super().clean_input(info, instance, data, input_cls=None)
@@ -131,7 +133,9 @@ class AccountInput(graphene.InputObjectType):
         LanguageCodeEnum, required=False, description="User language code."
     )
     metadata = graphene.List(
-        MetadataInput, description="User public metadata.", required=False
+        graphene.NonNull(MetadataInput),
+        description="User public metadata.",
+        required=False,
     )
 
 
@@ -155,9 +159,9 @@ class AccountUpdate(BaseCustomerCreate):
 
     @classmethod
     def clean_input(cls, info, instance, data):
-        data["metadata"] = {
-            item["key"]: item["value"] for item in data.get("metadata", [])
-        }
+        metadata = data.pop("metadata", [])
+        if metadata is not None:
+            data["metadata"] = {item["key"]: item["value"] for item in metadata}
         return super().clean_input(info, instance, data)
 
     @classmethod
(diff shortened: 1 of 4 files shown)


saleor/graphql/account/mutations/account.py
```diff
@@ -38,6 +39,9 @@ class AccountRegisterInput(graphene.InputObjectType):
     language_code = graphene.Argument(
         LanguageCodeEnum, required=False, description="User language code."
     )
+    metadata = graphene.List(
+        MetadataInput, description="User public metadata.", required=False
```
korycins, line 43: `graphene.NonNull(MetadataInput)`

saleor/graphql/account/mutations/account.py
```diff
@@ -123,6 +130,9 @@ class AccountInput(graphene.InputObjectType):
...
```
korycins, line 134: Should be NonNull

saleor/graphql/account/tests/test_account.py
```diff
@@ -1535,6 +1549,21 @@ def test_logged_customer_update_anonymous_user(api_client):
     assert_no_permission(response)
 
 
+def test_logged_customer_updates_metadata(user_api_client):
```
IKarbowiak, line 1552: It will be nice to also have for updating metadata in `AccountRegister` mutation. :)

korycins (approved):

saleor/graphql/account/tests/test_account.py
piotrgrundas: @IKarbowiak But there is 👀 . Test for accountRegister is updated to include that change.

IKarbowiak: True! Sorry :)

IKarbowiak (approved):

fowczarek (commented):
The current implementation of update metadata after by `accountUpdate`  loses data previously stored in metadata. 
eg. 
I have a user with metadata:
```
{
    "key": "k1",
    "value": "v1"
},
{
    "key": "k2",
    "value": "v2"
}
```
I want to update the value of k2 and add a new mete item. I should send:
```
input:{metadata:[{key:"k3", value:"v3"},{key:"k2", value:"v22"}]}
```
The current result is:
```
{
    "key": "k2",
    "value": "v22"
},
{
    "key": "k3",
    "value": "v3"
}
```

IMO expected result is:
```
{
    "key": "k1",
    "value": "v1"
},
{
    "key": "k2",
    "value": "v22"
},
{
    "key": "k3",
    "value": "v3"
}
```
@maarcingebala can you confirm?

piotrgrundas:
@fowczarek I've used same mechanism as in dashboard. 
Second thing - U won't be able to ever clear metadata items or particular one, using proposed way.

fowczarek:
But in the current implementation, you can reset your metadata accidentally. To delete metadata you should send `deleteMetadata` mutation. 
`updateMetadata` also doesn't clear existing metada.

korycins:
IMO, Update should not overwrite the previously stored fields in meta.
Also maybe I am missing something, but what is the difference between updating metadata by calling `UpdateMetadata` and `accountUpdate`? I understand that it is good to have a possibility to insert something in the metadata during the account registration, but not sure what is a benefit for extending `accountUpdate`

piotrgrundas:
@korycins
1) I followed dashboard pattern, but I won't argue here,
2) because it raises race condition in our integrations, and we talked about it (with U even) when discussing needed saleor dependencies.

korycins:
So can't you reverse the order of calling mutation? Like first call updateMetadata and then call update accountUpdate? 
We are going to separate events related to update metadata from standard object, maybe this could be helpful for you?

I don't know all your business cases and edge cases but I don't like the idea that we add action to mutation which will do the same thing as mutation already introduced. I would prefer to define the way for such cases as future projects can have similar issuse but related to different object, like Order etc.

patrys:
Why do we want multiple ways to manage metadata?

piotrgrundas:
As agreed with @patrys we will only allow it in `accountRegister`.

maarcingebala (approved):

PR #7170, opened 2021-04-15
Order total authorized performance

tomaszszymanski129:
I want to merge this change because it increases performance for order totalAuthorized field.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-2894 Optimize totalAuthorized resolver in Order type](https://app.clickup.com/t/2549495/SALEOR-2894)

saleor/order/models.py
```diff
@@ -342,10 +343,7 @@ def can_mark_as_paid(self):
 
     @property
     def total_authorized(self):
-        payment = self.get_last_payment()
```
maarcingebala, line 345: We should also look at other usages of `get_last_payment` from Order model and replace it with loading payments with dataloader and using the utility function that this PR adds. I'm not saying we have to do it in this PR but we should keep it on the radar.

maarcingebala: Btw is the `total_authorized` property needed anymore after migrating to dataloader in the Order type?

IKarbowiak (approved):

saleor/payment/utils.py
```diff
@@ -365,3 +367,14 @@ def is_currency_supported(currency: str, gateway_id: str, manager: "PluginsManag
     """Return true if the given gateway supports given currency."""
     available_gateways = manager.list_payment_gateways(currency=currency)
     return any([gateway.id == gateway_id for gateway in available_gateways])
+
+
+def get_last_payment(payments: List[Payment]):
```
korycins, line 372: It looks like it returns `last_payment` for the order but we never check if the payment is valid.  Not sure if this a proper way

fowczarek (approved):

saleor/order/models.py
tomaszszymanski129: @maarcingebala About property - It is used but on subsequent PRs i'll probably be able to get rid of it, i also have it in mind, it's just used in some other properties which will also be changed to use DLs.

maarcingebala (approved):

korycins (approved):

PR #7242, opened 2021-04-26
Extend notify method to handle channel slug as an input

korycins:
I want to merge this change because it:
- adds channel_slug parameter to manager.notify method.
- add input field - `channel` to account mutation, where channel slug is required to trigger proper Email plugins.
- add channel_slug to all manager.notify calls

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/account/mutations/base.py
+++ b/saleor/graphql/account/mutations/base.py
@@ -11,6 +11,9 @@ from ....account.notifications import (
     send_password_reset_notification,
     send_set_password_notification,
 )
+from ....channel.exceptions import ChannelNotDefined
+from ....channel.models import Channel
+from ....channel.utils import get_default_channel
 from ....core.exceptions import PermissionDenied
 from ....core.permissions import AccountPermissions
 from ....core.utils.url import validate_storefront_url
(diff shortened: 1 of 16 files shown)


saleor/graphql/account/mutations/base.py
```diff
@@ -149,8 +199,26 @@ def perform_mutation(cls, _root, info, **data):
...
+    def perform_mutation(cls, _root, info, **data):
+        email = data["email"]
+        redirect_url = data["redirect_url"]
+        channel_slug = data.get("channel_slug")
+        user = cls.clean_user(email, redirect_url)
+
+        if not user.is_staff:
+            channel_slug = clean_channel(channel_slug).slug
+        elif channel_slug is not None:
+            channel_slug = validate_channel(channel_slug).slug
```
IKarbowiak, line 214: Shouldn't be:
```suggestion
            channel_slug = clean_channel(channel_slug).slug
        else:
            channel_slug = validate_channel(channel_slug).slug
```
as we have a check for empty `channel_slug` there?

saleor/graphql/account/mutations/base.py
```diff
@@ -446,6 +520,14 @@ def save(cls, info, instance, cleaned_input):
             info.context.plugins.customer_updated(instance)
 
         if cleaned_input.get("redirect_url"):
+            channel_slug = cleaned_input.get("channel")
+            if not instance.is_staff:
+                channel_slug = clean_channel(channel_slug).slug
+            elif channel_slug is not None:
+                channel_slug = validate_channel(channel_slug).slug
```
IKarbowiak, line 527: Same here 👮‍♀️

saleor/graphql/account/mutations/base.py
korycins: For customer we clean the channel and validate it. We will raise exception when we have multiple channels and input.channel is None.

But for staff user we will proceed with validation only when the input.channel is provided. Staff user are not related with the channel so there is no need to require it as an input. Adding the `else` statement, we will require a channel slug for each staff user.

IKarbowiak: 👍  My mistake

IKarbowiak (approved):

fowczarek (approved):

PR #7249, opened 2021-04-27
Saleor 12918 fix image upload error

trannhuson:
fix image upload error


maarcingebala:
@trannhuson I believe this PR is open to the wrong repo, it looks like it should be opened to a Saleor fork.

PR #7333, opened 2021-05-12
Add traced_atomic_transaction decorator

tomaszszymanski129:
I want to merge this change because we had no way of tracing transaction times (only particular DB queries, but with no exact start time of transaction)
That PR introduces decorator to handle that and implements it across Saleor.

So now we see a transaction span with children being DB queries.

image

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-3325 Opentracing spans on transactions](https://app.clickup.com/t/2549495/SALEOR-3325)

fowczarek (commented):
IMO we should add tracing to all `with transaction.atomic():`  inside checkout/order flow and inside plugins.

saleor/core/tracing.py
```diff
@@ -14,3 +17,12 @@ def wrapper(*args, **kwargs):
             return func(*args, **kwargs)
 
     return wrapper
+
+
+@contextmanager
+def traced_atomic_transaction(x=None):
```
NyanKiyoshi, line 23: Non-used variable?

saleor/core/tracing.py
```diff
@@ -14,3 +17,12 @@ def wrapper(*args, **kwargs):
...
 
     return wrapper
+
+
+@contextmanager
+def traced_atomic_transaction():
+    with transaction.atomic():
+        with opentracing.global_tracer().start_active_span("transaction") as scope:
+            span = scope.span
+            span.set_tag(opentracing.tags.COMPONENT, "db")
```
fowczarek, line 27: I'm not sure that's the correct component.

patrys: Me neither. It will affect how much time is calculated to be spent "in the database."

NyanKiyoshi: Maybe a span tagged as component `orm`

fowczarek (approved):

saleor/core/transactions.py
```diff
@@ -8,11 +9,16 @@ def transaction_with_commit_on_errors():
...
     error = None
     with transaction.atomic():
-        try:
-            yield
-        except DatabaseError:
-            raise
-        except Exception as e:
-            error = e
+        with opentracing.global_tracer().start_active_span(
+            "transaction_with_commit_on_errors"
```
patrys, line 13: Does this need a separate span name? Tracing will show that this was not aborted if an exception is thrown.

tomaszszymanski129: Good point, changed it to just use `traced_atomic_transaction`.

GrzegorzKowalik (approved):

NyanKiyoshi (approved):

PR #7350, opened 2021-05-17
Add Kratos Authentication plugin

Sc0rpi0n101:


Sc0rpi0n101:
I'm sorry, that was not supposed to be there. My bad.

PR #7372, opened 2021-05-20
Change values to choices on AttributeType

d-wysocki:
I want to merge this change because...I want to change fields on AttributeType:
- values to choices

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-3071 Change values field on Attribute type to connection field](https://app.clickup.com/t/2549495/SALEOR-3071)

IKarbowiak (approved):

saleor/graphql/attribute/types.py
```diff
@@ -89,13 +89,13 @@ class Attribute(CountableDjangoObjectType):
...
     unit = MeasurementUnitsEnum(description=AttributeDescriptions.UNIT)
-    values = FilterInputConnectionField(
+    choices = FilterInputConnectionField(
         AttributeValue,
         filter=AttributeValueFilterInput(),
         description=AttributeDescriptions.VALUES,
     )
 
-    value_required = graphene.Boolean(
+    choice_required = graphene.Boolean(
```
maarcingebala, line 98: I'm not sure if we should change this field. Required value can be true for attributes such as rich-text, not only those that have choices. E.g. if you have a product with rich-text attr, you can set "Value required" to true, which means that this rich-text will be required to save a product. This is a bit separate thing and I wouldn't change it unless necessary.

saleor/graphql/attribute/types.py
```diff
@@ -128,12 +128,19 @@ class Meta:
 
     @staticmethod
     @traced_resolver
-    def resolve_values(root: models.Attribute, info, **_kwargs):
+    def resolve_choices(root: models.Attribute, _info, **_kwargs):
+        types_without_choices = [
+            AttributeInputType.REFERENCE,
+            AttributeInputType.NUMERIC,
+            AttributeInputType.RICH_TEXT,
+        ]
```
maarcingebala, line 136: This could be extracted to a constant and moved to `saleor/attribute/__init__.py`. This list may change only when adding a new input type (which happens in this init file), so it's good to have the two things as close to each other as possible.

saleor/graphql/attribute/types.py
d-wysocki: Okey, I will change it back.

saleor/attribute/__init__.py
```diff
@@ -22,6 +22,12 @@ class AttributeInputType:
     # list of the input types that can be used in variant selection
     ALLOWED_IN_VARIANT_SELECTION = [DROPDOWN]
 
+    TYPES_WITHOUT_CHOICES = [
+        REFERENCE,
+        NUMERIC,
+        RICH_TEXT,
+    ]
```
patrys, line 29: Shouldn't this work the other way around, listing the few types where we want to return choices?

maarcingebala: Fine by me

maarcingebala (approved):

d-wysocki:
Waiting on frontend part.

orzechdev (approved):

PR #7424, opened 2021-06-02
Fix performance of order events

tomaszszymanski129:
I want to merge this change because it adds missing dataloaders for order events
- `user` resolver on `OrderEvent` type
- `fulfilled_items` resolver on `OrderEvent` type (new dataloader - FulfillmentLinesByIdLoader)
- `order_line` resolver on `FulfillmentLine` type

test staff order details:

before: 86 queries
after: 46 queries

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested

### DIFF
--- a/saleor/graphql/order/types.py
+++ b/saleor/graphql/order/types.py
@@ -193,7 +193,7 @@ class OrderEvent(CountableDjangoObjectType):
                 or user.has_perm(AccountPermissions.MANAGE_STAFF)
             ):
                 return event_user
-            raise PermissionDenied()
+            return None
 
         if not root.user_id:
             return None
@@ -337,9 +337,6 @@ class FulfillmentLine(CountableDjangoObjectType):
     @staticmethod
     @traced_resolver
     def resolve_order_line(root: models.FulfillmentLine, info):
-        if not root.order_line_id:
-            return None
-
         return OrderLineByIdLoader(info.context).load(root.order_line_id)
 
 
--- a/saleor/graphql/order/types.py
+++ b/saleor/graphql/order/types.py
@@ -61,6 +61,7 @@ from ..shipping.types import ShippingMethod
 from ..warehouse.types import Allocation, Warehouse
 from .dataloaders import (
     AllocationsByOrderLineIdLoader,
+    FulfillmentLinesByIdLoader,
     FulfillmentsByOrderIdLoader,
     OrderByIdLoader,
     OrderEventsByOrderIdLoader,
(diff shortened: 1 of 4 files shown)


patrys:
Task linked: [SALEOR-2944 Optimize OrderEvent type resolvers](https://app.clickup.com/t/2549495/SALEOR-2944)

saleor/graphql/order/types.py
```diff
@@ -184,14 +185,20 @@ class Meta:
...
+            if (
+                user == event_user
+                or user.has_perm(AccountPermissions.MANAGE_USERS)
+                or user.has_perm(AccountPermissions.MANAGE_STAFF)
+            ):
+                return event_user
+            return None
+
+        if not root.user_id:
+            return None
```
IKarbowiak, line 199: I know it was like that before, but shouldn't we consider also an app requestor here?

maarcingebala (approved):

saleor/graphql/order/types.py
tomaszszymanski129: good point, will check other loaders as well.

fowczarek: I'm not sure about this permission, because if I'm staff user responsible only for order. I can't checkout who fulfill or process other actions on this order. 
IMO all staff users able to see order events should be able to see at least the email of staff who generate that event.
PS. Maybe we should do it in separate PR for all events. By implementing resolver like, `event_performer_email`

fowczarek: @pwgryglak 
What do you think?

fowczarek (approved):

saleor/graphql/order/types.py
```diff
@@ -184,14 +185,20 @@ class Meta:
...
-            or user.has_perm(AccountPermissions.MANAGE_STAFF)
-        ):
-            return root.user
-        raise PermissionDenied()
+        def _resolve_user(event_user):
+            user = info.context.user
+            if (
+                user == event_user
+                or user.has_perm(AccountPermissions.MANAGE_USERS)
+                or user.has_perm(AccountPermissions.MANAGE_STAFF)
```
fowczarek, line 193: In all user resolvers, we should use similar logic to `saleor/saleor/graphql/account/resolvers.py::resolve_user` 
We should check if the requested user is staff and then check proper permission.

saleor/graphql/order/types.py
pwgryglak: Users should know who created the event - that is the purpose of having a history in first place.

tomaszszymanski129:
Will be refactored in another ticket - this stays as is for performance (behavior remains unchanged).

PR #7468, opened 2021-06-12
Update Django Storages variables

ahmedaljazzar:
During my work with Google storages, I found that when trying to upload a picture in the dashboard (say `some_image_name.png`), everything goes smooth and fine. However, if I try to upload a different image with the same name `some_image_name.png` all previously uploaded images will be replaced with the new one.
I posted this issue on Spectrum chat [here](https://spectrum.chat/saleor/saleor-3-0/image-with-the-same-name-get-replaced-bug~419e8869-a069-4f93-a76d-db583a5f9854) and people suggested bucket versioning. However, this didn't work for me. 
After some digging, I found that [Django Storages](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html) control this behavior with `*_FILE_OVERWRITE` variable.
- `AWS_S3_FILE_OVERWRITE` for S3Boto3Storage
- `GS_BUCKET_NAME` for GoogleCloudStorage

I added these variables to the settings file and set them to `True`, which is the default behavior in the library (to override files with the same name).

On another note, Saleor uses `GS_STORAGE_BUCKET_NAME` which is going to throw errors when you upload a file using GS. This happens because Django Storages checks for `GS_BUCKET_NAME` not ~`GS_STORAGE_BUCKET_NAME`~. See [here](https://django-storages.readthedocs.io/en/latest/backends/gcloud.html) for documentation

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -419,16 +419,18 @@ AWS_S3_REGION_NAME = os.environ.get("AWS_S3_REGION_NAME", None)
 AWS_SECRET_ACCESS_KEY = os.environ.get("AWS_SECRET_ACCESS_KEY")
 AWS_STORAGE_BUCKET_NAME = os.environ.get("AWS_STORAGE_BUCKET_NAME")
 AWS_DEFAULT_ACL = os.environ.get("AWS_DEFAULT_ACL", None)
+AWS_S3_FILE_OVERWRITE = get_bool_from_env("AWS_S3_FILE_OVERWRITE", True)
 
 # Google Cloud Storage configuration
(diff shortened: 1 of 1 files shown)


maarcingebala (approved):
LGTM 👍  Issue with having unique names per file upload should be now fixed in Saleor - we now add a randomized string to each filename that lands in the storage. Setting `AWS_S3_FILE_OVERWRITE` to True shouldn't break anything if the default value is True anyway.

IKarbowiak (approved):

maarcingebala:
Thanks for contributing to Saleor @ahmedaljazzar 🎉  !

PR #7521, opened 2021-06-23
Introduce job_payload to Webhooks

d-wysocki:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-3583 Introduce JobPayloads for Webhooks tasks](https://app.clickup.com/t/2549495/SALEOR-3583)

saleor/plugins/webhook/tasks.py
```diff
@@ -144,27 +150,72 @@ def send_webhook_using_google_cloud_pubsub(
...
 )
-def send_webhook_request(webhook_id, target_url, secret, event_type, data):
+def send_webhook_request(
+    self, webhook_id, target_url, secret, event_type, event_payload_id
+):
+    try:
+        data = EventPayload.objects.get(id=event_payload_id)
+    except EventPayload.DoesNotExist:
+        task_logger.error(
+            "Cannot find payload related to webhook.",
```
korycins, line 181: What about updating status of `EventPayloadReference` here? Shouldn't we include info there that we were not able to send webhook?

saleor/plugins/webhook/tasks.py
```diff
@@ -144,27 +150,72 @@ def send_webhook_using_google_cloud_pubsub(
...
+    else:
+        raise ValueError("Unknown webhook scheme: %r" % (parts.scheme,))
+
+
 @app.task(
+    bind=True,
     autoretry_for=(RequestException,),
     retry_backoff=10,
     retry_kwargs={"max_retries": 5},
     compression="zlib",
```
korycins, line 172: Do we still need compression here?

saleor/plugins/webhook/tasks.py
```diff
@@ -144,27 +150,72 @@ def send_webhook_using_google_cloud_pubsub(
...
-        send_webhook_using_aws_sqs(target_url, message, domain, signature, event_type)
-    elif parts.scheme.lower() == WebhookSchemes.GOOGLE_CLOUD_PUBSUB:
-        send_webhook_using_google_cloud_pubsub(
-            target_url, message, domain, signature, event_type
+    with _catch_duration_time() as duration:
+        try:
+            _send_webhook_to_target(
+                parts, target_url, message, domain, signature, event_type
+            )
+        except Exception as exc:
```
korycins, line 200: Maybe we can convert these two calls into one?
I mean, almost all parameters are the same, you can assign an exception to error variable, and assign JobStatus based on this. Then if an error exists, just raise excpetion. What do you think?

saleor/plugins/webhook/tasks.py
```diff
@@ -144,27 +150,72 @@ def send_webhook_using_google_cloud_pubsub(
...
+            EventPayloadReference.objects.create(
+                task_id=self.request.id,
+                event_payload_id=event_payload_id,
+                error={"error_type": str(exc.__class__.__name__), "message": str(exc)},
```
maarcingebala, line 204: Do we need to store it as JSON? I think `str(exc)` could be enough here.

saleor/core/models.py
```diff
@@ -127,3 +127,20 @@ class Job(models.Model):
...
+
+class EventPayloadReference(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    event_payload = models.ForeignKey(
+        EventPayload, null=True, on_delete=models.SET_NULL
+    )
+    task_id = models.CharField(max_length=254)
+    status = models.CharField(max_length=50, choices=JobStatus.CHOICES)
+    error = JSONField(blank=True, null=True, default=dict, encoder=CustomJsonEncoder)
+    duration = models.CharField(max_length=254)
```
maarcingebala, line 144: I'm thinking about using a FloatField here. It should be easier to run queries against this model such as filtering by duration range.

saleor/core/models.py
```diff
@@ -127,3 +127,20 @@ class Job(models.Model):
...
+
+class EventPayload(models.Model):
+    payload = JSONField(default=dict, encoder=CustomJsonEncoder)
+
+
+class EventPayloadReference(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    event_payload = models.ForeignKey(
+        EventPayload, null=True, on_delete=models.SET_NULL
+    )
```
maarcingebala, line 140: A reverse relation could be useful here.

saleor/core/models.py
```diff
@@ -127,3 +127,20 @@ class Job(models.Model):
 
     class Meta:
         abstract = True
+
+
+class EventPayload(models.Model):
+    payload = JSONField(default=dict, encoder=CustomJsonEncoder)
+
+
+class EventPayloadReference(models.Model):
```
maarcingebala, line 136: This model represents a single Celery task that sends a webhook, so maybe a better name would be `EventTask`?

saleor/plugins/webhook/utils.py
```diff
@@ -104,3 +106,9 @@ def parse_payment_action_response(
             "transaction_already_processed", False
         ),
     )
+
+
+@contextmanager
+def _catch_duration_time():
```
maarcingebala, line 112: This function is not "private", since you don't use it in this module, but it is imported in some other place.

saleor/plugins/webhook/tasks.py
```diff
@@ -144,27 +150,75 @@ def send_webhook_using_google_cloud_pubsub(
...
+        EventTask.objects.create(
+            task_id=self.request.id,
+            event_payload_id=event_payload_id,
+            status=JobStatus.FAILED,
+            error=str(exc),
+            duration=0,
+            event_type=event_type,
+            webhook_id=webhook_id,
+        )
+        task_logger.error(
```
IKarbowiak, line 188: This function is quite long, maybe we can create something like `_get_event_payload` method for this part of the code?

IKarbowiak (approved):

saleor/core/tasks.py
```diff
@@ -1,8 +1,20 @@
...
 
 
 @app.task
 def delete_from_storage_task(path):
     default_storage.delete(path)
+
+
+@app.task
+def delete_event_payloads():
+    seven_days_ago = datetime.date.today() - datetime.timedelta(days=7)
```
korycins, line 16: IMO, this should be a const value or variable in the settings.py file.  We should be able to control this time range
You could do something like we did for jwt tokens:
```
settings.py
JWT_TTL_REFRESH = timedelta(seconds=parse(os.environ.get("JWT_TTL_REFRESH", "30 days")))
```

saleor/core/models.py
```diff
@@ -127,3 +127,23 @@ class Job(models.Model):
...
+
+class EventPayload(models.Model):
+    payload = JSONField(default=dict, encoder=CustomJsonEncoder)
+
+
+class EventTask(models.Model):
+    created_at = models.DateTimeField(auto_now_add=True)
+    event_payload = models.ForeignKey(
+        EventPayload,
+        related_name="event_payloads",
```
korycins, line 140: Are you sure that the related_name here is correct?
To fetch all EventPayload which are assigned to EventTaskID == 1, I will have to create query:
```
EventPayload.objects.filter(event_payloads__id=1)
```
related_name describes the connection from EventPayload -> EventTask, so it should be something related with event_task.
```
EventPayload.objects.filter(event_task__id=1)
```

For example ProductVariant.
It has field `product` which has `related_name=variants` (not `products`)
and you can do `Product.objects.filter(variants__id=1)`

saleor/settings.py
```diff
@@ -529,6 +530,13 @@ def get_host():
 CELERY_RESULT_SERIALIZER = "json"
 CELERY_RESULT_BACKEND = os.environ.get("CELERY_RESULT_BACKEND", None)
 
+CELERY_BEAT_SCHEDULE = {
+    "delete-event-payloads": {
+        "task": "core.tasks.delete_event_payloads",
+        "schedule": crontab(hour=4, day_of_week=6),
```
korycins, line 536: Why `day_of_week=6`?

saleor/core/tasks.py
```diff
@@ -1,8 +1,14 @@
 from django.core.files.storage import default_storage
 
 from ..celeryconf import app
+from .management.commands.delete_event_payloads import delete_event_payloads
```
korycins, line 4: Can we extract `delete_event_payloads` somewhere?  It could be even here.
IMHO import should be in the reverse direction. The command should import logic to execute from Saleor.

saleor/core/management/commands/delete_event_payloads.py
```diff
@@ -0,0 +1,27 @@
...
+
+class Command(BaseCommand):
+    help = (
+        "Delete EventPayloads and EventTasks from database "
+        "that are older than the value set "
+        "in EVENT_PAYLOAD_DELETE_PERIOD environment variable."
+    )
+
+    def handle(self, **options):
+        delete_event_payloads()
```
korycins, line 27: I would consider triggering a celery task here instead of direct calling.

maarcingebala (approved):

korycins (approved):

NyanKiyoshi (dismissed):
Could it be checked whether it adds a significant load to the database when a lot of requests a generating webhook events due to `JSONB` column type?

It could create huge/big resource usage for the database to handle, thus affecting the database capabilities and cost.

But also add significant response time overheads for requests generating webhook events

--- 

Requesting changes: test how it scales first and determine whether a non-JSON column type wouldn't be more adapted

fowczarek (approved):

maarcingebala:
@NyanKiyoshi Good catch, we changed JSON field to text field in the model and we dump the payload to string before saving it.

PR #7571, opened 2021-07-01
Fixed handling of App requesters in `saleor.order.events`

onecrayon:
This object was improperly assumed to be a User, when a requester can also be an App.

I ran across an AttributeError caused when an App makes a `orderDiscountAdd` request through GraphQL, and it was trivial to fix, so I did so.

However, I was not able to figure out where would be appropriate to add a test to ensure this doesn't suffer from regressions. I would love guidance on this (or information on where to find an overview of the testing strategy for Saleor).

### DIFF
--- a/saleor/graphql/order/mutations/discount_order.py
+++ b/saleor/graphql/order/mutations/discount_order.py
@@ -22,7 +22,6 @@ from ...core.mutations import BaseMutation
 from ...core.scalars import PositiveDecimal
 from ...core.types.common import OrderError
 from ...discount.enums import DiscountValueTypeEnum
-from ...utils import get_user_or_app_from_context
 from ..types import Order, OrderLine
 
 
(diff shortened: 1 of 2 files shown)


korycins:
Hey @onecrayon, thanks for raising this and creating a PR. To keep a consistency, instead of modifying condition is_user_valid, can you modify a call of `order_discount_added_event` in `saleor.graphql.order.mutations.discount_order.OrderDiscountAdd`?  It should take `user=info.context.user`. It is done in that way in all event calls (afair). In future releases, we will also cover events for app.

onecrayon:
@korycins Thanks for the feedback! I've updated my PR to explicitly pass `info.context.user` to all discount mutation events (including those for line items), since they all rely on the `_is_user_valid` check which fails when it gets an App instance.

korycins (approved):

ashocky (approved):

maarcingebala (approved):

IKarbowiak (approved):

maarcingebala:
@onecrayon Thank you for your contribution to Saleor 🎉

PR #7589, opened 2021-07-05
Reserve stock for checkout

rafalp:
This PR adds new `reservation` logic to the `saleor.warehouse` package. Reservations are closely inspired by stock allocation, except they associate stock with checkouts instead of orders and have expiration date after which they no longer affect available quantities count.

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/warehouse/availability.py
```diff
@@ -94,3 +125,49 @@ def is_product_in_stock(
...
+def get_reserved_quantity_bulk(
+    stocks: Iterable[Stock],
+    current_checkout_lines: Optional[List["CheckoutLine"]] = None,
+) -> Dict[int, int]:
+    reservations: Dict[int, int] = defaultdict(int)
+    if not stocks:
+        return reservations
+
+    result = (
+        Reservation.objects.filter(
```
korycins, line 156: it is only a suggestion, but you could have some helper function that will return `result` structure. You could use it here and inside `get_reserved_quantity`

saleor/warehouse/models.py
```diff
@@ -157,3 +173,43 @@ class Allocation(models.Model):
...
+    )
+    stock = models.ForeignKey(
+        Stock,
+        null=False,
+        blank=False,
+        on_delete=models.CASCADE,
+        related_name="reservations",
+    )
+    quantity_reserved = models.PositiveIntegerField(default=0)
+    reserved_until = models.DateTimeField()
```
korycins, line 205: Shouldn't accept some default value here? Like now+timedelta?

korycins (commented):
Don't we need a celery task that will remove expired reservation?

saleor/warehouse/reservations.py
```diff
@@ -0,0 +1,160 @@
...
+from ..product.models import ProductVariant
+from .models import Allocation, Reservation, Stock
+
+if TYPE_CHECKING:
+    from ..checkout.models import CheckoutLine
+
+StockData = namedtuple("StockData", ["pk", "quantity"])
+
+
+@transaction.atomic
```
fowczarek, line 17:
```suggestion
@traced_atomic_transaction()
```

saleor/warehouse/reservations.py
```diff
@@ -0,0 +1,160 @@
...
+        Allocation.objects.filter(
+            stock_id__in=stocks_id,
+            quantity_allocated__gt=0,
+        )
+        .values("stock")
+        .annotate(quantity_allocated_sum=Sum("quantity_allocated"))
+    )
+    quantity_allocation_for_stocks: Dict = defaultdict(int)
+    for allocation in quantity_allocation_list:
+        quantity_allocation_for_stocks[allocation["stock"]] += allocation[
```
fowczarek, line 52: We should cover this with tests. I think it's enough to add a test for creating reservations for lines that contain variants with non 0 allocations and non 0 reservations.

saleor/warehouse/reservations.py
```diff
@@ -0,0 +1,160 @@
...
+            quantity_reserved__gt=0,
+        )
+        .not_expired()
+        .exclude_checkout_lines(checkout_lines)
+        .values("stock")
+        .annotate(quantity_reserved_sum=Sum("quantity_reserved"))
+    )  # type: ignore
+    quantity_reservation_for_stocks: Dict = defaultdict(int)
+    for reservation in quantity_reservation_list:
+        quantity_reservation_for_stocks[allocation["stock"]] += reservation[
```
fowczarek, line 68: We should cover this with tests. I think it's enough to add a test for creating reservations for lines that contain variants with non 0 allocations and non 0 reservations.

saleor/warehouse/reservations.py
```diff
@@ -0,0 +1,160 @@
...
+    variant_to_stocks: Dict[str, List[StockData]] = defaultdict(list)
+    for stock_data in stocks:
+        variant = stock_data.pop("product_variant")
+        variant_to_stocks[variant].append(StockData(**stock_data))
+
+    insufficient_stock: List[InsufficientStockData] = []
+    reservations: List[Reservation] = []
+    for line in checkout_lines:
+        line.variant = cast(ProductVariant, line.variant)
+        stock_reservations = variant_to_stocks[line.variant.pk]
```
fowczarek, line 81:
```suggestion
        stock_reservations = variant_to_stocks[line.variant_pk]
```

fowczarek (commented):
There is any way to turn off this feature?

saleor/graphql/checkout/mutations.py
```diff
@@ -440,16 +448,28 @@ class Meta:
...
+            current_checkout_lines=current_checkout_lines,
+        )
 
     @classmethod
     def clean_input(
         cls, checkout, variants, quantities, checkout_info, manager, discounts, replace
     ):
         channel_slug = checkout_info.channel.slug
+        existing_lines = [
+            line_info.line for line_info in fetch_checkout_lines(checkout)
```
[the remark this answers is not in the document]
rafalp: We actually do this three times:

1. Pull pre-update checkout lines to exclude them from reservations check.
2. Pull post-update checkout lines to update `checkout_info.valid_shipping_methods` (this is in `clean_input`).
3. Pull post-update checkout lines to update shipping method (this is in `perform_mutation`).

Steps two and three are doing same thing, but in different places and step 3 also does shipping method update if current one is no longer valid. I've dropped second one and tests are passing so here's that.

saleor/warehouse/reservations.py
```diff
@@ -0,0 +1,160 @@
...
+                variant=line.variant, checkout_line=line  # type: ignore
+            )
+        )
+        return insufficient_stock, []
+
+    return [], []
+
+
+def get_checkout_lines_to_reserve(
+    checkout_lines: Iterable["CheckoutLine"],
```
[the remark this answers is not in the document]
rafalp: That was my initial approach to this, but I've also added reservations logic to the `add_variant_to_checkout` where there is no `CheckoutLineInfo`.

But I'll give those checkout queries a look to see if there's anything fishy going on anyway. #7633 also adds query count tests to checkout which should help out here.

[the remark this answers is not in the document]
rafalp: Add tests for query count for adding/updating checkout lines. Those found that number of queries is not constant for checkout lines update, but I've fixed that.

saleor/graphql/checkout/types.py
```diff
@@ -478,3 +485,18 @@ def with_product_types(product_types):
...
         return LanguageCodeEnum[str_to_enum(root.language_code)]
+
+    @staticmethod
+    @traced_resolver
+    def resolve_stock_reservation_expires(root: models.Checkout, info):
+        def get_oldest_stock_reservation_expiration_date(reservations):
+            if not reservations:
+                return None
+
+            return min(reservation.reserved_until for reservation in reservations)
```
korycins, line 496: Assuming that I have two lines:
1st - expires 12:00 29-07-2020
2nd - expires 12:05 29-07-2020
Now, I am adding a new line to checkout, which gives me 
3nd - expires 22:00 29-07-2020
calling `resolve_stock_reservation_expires` will return me `12:00 29-07-2020`. Is that correct? I though that adding new line to checkout will 'refresh' reservation time (maybe it is done in different place and I just missed that)

saleor/graphql/warehouse/dataloaders.py
```diff
@@ -179,3 +180,30 @@ class WarehouseByIdLoader(DataLoader):
     def batch_load(self, keys):
         warehouses = Warehouse.objects.in_bulk(keys)
         return [warehouses.get(UUID(warehouse_uuid)) for warehouse_uuid in keys]
+
+
+class StocksReservationsByCheckoutTokenLoader(DataLoader):
+    context_key = "stock_reservation_by_checkout_id"
```
korycins, line 186: `stock_reservation_by_checkout_token` - based on the name.

saleor/graphql/warehouse/dataloaders.py
```diff
@@ -179,3 +180,30 @@ class WarehouseByIdLoader(DataLoader):
...
+
+    def batch_load(self, keys):
+        def with_checkouts_lines(checkouts_lines):
+            checkouts_keys_map = {}
+            for i, key in enumerate(keys):
+                for checkout_line in checkouts_lines[i]:
+                    checkouts_keys_map[checkout_line.id] = key
+
+            reservations_map = defaultdict(list)
+            reservations = Reservation.objects.filter(
```
korycins, line 196: I believe this should also come from Dataloader.

saleor/site/models.py
```diff
@@ -65,6 +65,7 @@ class SiteSettings(models.Model):
     default_mail_sender_address = models.EmailField(blank=True, null=True)
     customer_set_password_url = models.CharField(max_length=255, blank=True, null=True)
     automatically_confirm_all_new_orders = models.BooleanField(default=True)
+    reserve_stock_duration_minutes = models.IntegerField(blank=True, null=True)
```
korycins, line 68: Is it ok that we by default assume that staff user will use minutes here? What if I would like to have 45 seconds for reservation (yes, i know that this is very short time to finish a checkout). What if user will want to have 5.5 minutes?

saleor/warehouse/reservations.py
```diff
@@ -0,0 +1,176 @@
...
+
+def _create_reservations(
+    line: "CheckoutLine",
+    variant: "ProductVariant",
+    stocks: List[StockData],
+    quantity_allocation_for_stocks: dict,
+    quantity_reservation_for_stocks: dict,
+    insufficient_stock: List[InsufficientStockData],
+    length_in_minutes: int,
+):
```
korycins, line 115: Can you also add a typing for response?

korycins (commented):
Looks good but I would also make sure that we cover all new actions by benchmarks tests (some part is already done but in standard test directory).

saleor/checkout/utils.py
```diff
@@ -88,8 +89,12 @@ def add_variant_to_checkout(
 
     If `replace` is truthy then any previous quantity is discarded instead
     of added to.
+
+    This function is not used outside of test suite.
```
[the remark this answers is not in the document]
rafalp: It only means that I've found that this function is only used in tests suite, with business logic using `add_variants_to_checkout` instead. I've added that bit to the comment to provoke the discussion, but IMHO whatever we decide to do next with it is out of scope of stock reservations.

saleor/checkout/utils.py
```diff
@@ -125,7 +133,14 @@ def calculate_checkout_quantity(lines: Iterable["CheckoutLineInfo"]):
 
 
 def add_variants_to_checkout(
-    checkout, variants, quantities, channel_slug, skip_stock_check=False, replace=False
+    site_settings,
```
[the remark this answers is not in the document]
rafalp: IMHO site settings are greater context than remaining arguments, hence they are first (same how `request` arg is usually passed first if function is expecting it in Django).

But I'm more worried that checkout and order logic is build around functions taking between 5 to 8 arguments. Checkout state seems to be very fragmented and would benefit from being combined in something level above `CheckoutInfo`.

saleor/site/models.py
rafalp: I've stopped doing granularity like that when I've realized scenarios like this only happen during design discussions. In 5.5m scenario user is more likely to write either 5 or 6 instead of going "I'll use different e-commerce solution that lets me enter 330s" 😄

saleor/graphql/checkout/types.py
rafalp: This was never named a requirement AFAIR. Reservations are supposed to increase your chances to complete checkout within given time, but its still on first come-first serve basis, so if you want to buy that pirate hoodie, better hurry up with picking those sneakers you've also wanted. But maybe this is worth revisiting? CC @patrys

saleor/checkout/utils.py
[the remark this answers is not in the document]
korycins: If you do not have enough resources to drop it in this PR or in separate. Make sure that proper task is created for this

saleor/checkout/utils.py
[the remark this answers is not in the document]
korycins: But greater context for whole Saleor or only this function? I would say that checkout, variants give this function more context than site_settings. In your case, you're using `site_settings` to fetch two bool flag to determine a configuration of Saleor. Personally I would say, that here `checkout` and `variants` are "the Django's `request`". You can't do anything without them, but the flags that you take from site_settings could be also passed as a optional parameters and this function woul work in that case.

saleor/graphql/checkout/types.py
korycins: @patrys  What do you thin?

saleor/site/models.py
korycins: 👍

saleor/graphql/checkout/types.py
patrys: How it's supposed to work is it should reset all line reservations every time you add something to your cart. In the above example, adding the third line should re-reserve all three lines. We don't want to punish people for buying multiple items.

korycins (approved):

rafalp:
I've rebased on master and finished checking if I've missed any place that should account for reservations but doesn't.

saleor/warehouse/availability.py
```diff
@@ -99,3 +129,75 @@ def check_stock_quantity_bulk(
...
+            quantity_reserved=Coalesce(Sum("quantity_reserved"), 0),
+        )
+    )  # type: ignore
+
+    return result["quantity_reserved"]
+
+
+def get_reserved_quantity_bulk(
+    stocks: Iterable[Stock],
+    checkout_lines: Iterable["CheckoutLineInfo"],
```
maarcingebala, line 179: Cosmetics - is it possible to have the same interface to both `get_reserved_quantity*` functions? Now one accepts `[CheckoutLineInfo]` and the other `[CheckoutLine]`. Both could be unified to `[CheckoutLine]` and a queryset as the first param. But if it requires to much effort, let's not bother.

saleor/graphql/shop/mutations.py
```diff
@@ -60,6 +60,18 @@ class ShopSettingsInput(graphene.InputObjectType):
     customer_set_password_url = graphene.String(
         description="URL of a view where customers can set their password."
     )
+    reserve_stock_duration_minutes_anonymous = graphene.Int(
+        description=(
+            "Default number of minutes stock will be reserved for checkout or 0 to "
```
maarcingebala, line 65: `ADDED_IN_31`

saleor/graphql/shop/mutations.py
```diff
@@ -60,6 +60,18 @@ class ShopSettingsInput(graphene.InputObjectType):
...
+            "disable stock reservations."
+        )
+    )
+    reserve_stock_duration_minutes_authenticated = graphene.Int(
+        description=(
+            "Default number of minutes stock will be reserved for checkout or 0 to "
```
maarcingebala, line 71: `ADDED_IN_31`

saleor/graphql/checkout/mutations.py
```diff
@@ -458,15 +478,21 @@ class Meta:
...
+            country,
+            channel_slug,
+            existing_lines=lines,
+            check_reservations=is_reservation_enabled(request.site.settings),
         )
 
     @classmethod
     def clean_input(
         cls,
+        request,
```
maarcingebala, line 495: Usually we pass the `info` object in a case like this, if you look at other mutations. To be consistent I would change `request` to `info` in these mutations.

saleor/graphql/checkout/mutations.py
```diff
@@ -803,7 +849,15 @@ def process_checkout_lines(
             ).prefetch_related("product__product_type")
         )  # FIXME: is this prefetch needed?
         quantities = [line_info.line.quantity for line_info in lines]
-        check_lines_quantity(variants, quantities, country, channel_slug)
+        check_lines_quantity(
+            variants,
+            quantities,
+            country,
+            channel_slug,
+            replace=True,
```
maarcingebala, line 857: Hmm maybe we should add some comment why we pass `replace=True` here.

saleor/warehouse/reservations.py
```diff
@@ -0,0 +1,191 @@
...
+        .annotate(quantity_allocated_sum=Sum("quantity_allocated"))
+    )
+    quantity_allocation_for_stocks: Dict = defaultdict(int)
+    for allocation in quantity_allocation_list:
+        quantity_allocation_for_stocks[allocation["stock"]] += allocation[
+            "quantity_allocated_sum"
+        ]
+
+    quantity_reservation_list = list(
+        Reservation.objects.filter(
```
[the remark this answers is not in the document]
rafalp: I've added select for update here.

maarcingebala (approved):
All good from my perspective, GraphQL Inspector is failing due but it seems unrelated to this PR, maybe rebasing it against master would help.

saleor/graphql/warehouse/dataloaders.py
```diff
@@ -206,6 +211,55 @@ def batch_load_stocks_by_country(
...
+                    checkouts_keys_map[checkout_line.id] = key
+
+            def with_lines_reservations(lines_reservations):
+                reservations_map = defaultdict(list)
+                for reservations in lines_reservations:
+                    for reservation in reservations:
+                        checkout_key = checkouts_keys_map[reservation.checkout_line_id]
+                        reservations_map[checkout_key].append(reservation)
+
+                return [reservations_map[key] for key in keys]
```
fowczarek, line 233: Dataloader should always return a list with the same length as the length of keys. 
```suggestion
                return [reservations_map.get(key) for key in keys]
```

saleor/graphql/warehouse/dataloaders.py
```diff
@@ -206,6 +211,55 @@ def batch_load_stocks_by_country(
...
+    def batch_load(self, keys):
+        reservations_by_checkout_line = defaultdict(list)
+        queryset = Reservation.objects.filter(
+            checkout_line_id__in=keys
+        ).not_expired()  # type: ignore
+        for reservation in queryset:
+            reservations_by_checkout_line[reservation.checkout_line_id].append(
+                reservation
+            )
+        return [reservations_by_checkout_line[key] for key in keys]
```
fowczarek, line 260:
```suggestion
        return [reservations_by_checkout_line.get(key) for key in keys]
```

fowczarek (commented):
Should we create a celery beat task for removing expired reservations?

saleor/graphql/shop/mutations.py
```diff
@@ -60,6 +60,18 @@ class ShopSettingsInput(graphene.InputObjectType):
     customer_set_password_url = graphene.String(
         description="URL of a view where customers can set their password."
     )
+    reserve_stock_duration_anonymous_user = graphene.Int(
+        description=(
+            f"{ADDED_IN_31} Default number of minutes stock will be reserved for "
+            "anonymous checkout or 0 to disable stock reservations."
+        )
+    )
```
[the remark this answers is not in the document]
rafalp: TBH any int < 1 or null will disable reservations for setting because I'm normalizing those (there are even tests for that).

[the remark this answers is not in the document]
rafalp: Also if we make this field require we will make partial setting updates through GraphQL API harder.

saleor/graphql/shop/types.py
```diff
@@ -217,6 +217,19 @@ class Shop(graphene.ObjectType):
         description="Enable automatic fulfillment for all digital products."
     )
 
+    reserve_stock_duration_anonymous_user = graphene.Int(
+        description=(
+            f"{ADDED_IN_31} Default number of minutes stock will be reserved for "
+            "anonymous checkout or null when stock reservation is disabled."
```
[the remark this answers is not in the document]
rafalp: We normalize <1 and null values to None in database.

saleor/graphql/warehouse/dataloaders.py
rafalp: `reservations_map` is a `defaultdict` that returns list for nonexistant keys (`.get(key, list())`) and `keys` here is argument passed to `batch_load` so it will always be list of reservations found for given checkout tokens (or empty lists if reservations weren't found for given checkout.

saleor/graphql/warehouse/dataloaders.py
rafalp: Ditto

saleor/graphql/shop/mutations.py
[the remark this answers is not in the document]
fowczarek: IMO, we should have consistent descriptions. Not all frontend developers know about backend implementation. The description should describe what means all possible values. 
@maarcingebala @krzysztofwolski what do you think?

saleor/graphql/schema.graphql
```diff
@@ -4356,7 +4357,8 @@ type OrderLine implements Node {
   id: ID!
   productName: String!
   variantName: String!
-  productSku: String!
+  productSku: String
+  productVariantId: String
```
maarcingebala, line 4361: Is this needed for frontend? Where does this change come from?

saleor/graphql/shop/mutations.py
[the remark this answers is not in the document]
rafalp: This is robustness principle in practice: be strict about what you send but tolerant in what you receive. I normalize all falsy values (and nonsensical negative ints) to disabled reservations, but to keep things simple in API docs side I don't describe edge cases.

I can reword this to "0 or null to disable".

saleor/graphql/shop/types.py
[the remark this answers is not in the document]
maarcingebala: Wouldn't it be easier for frontend to always return int here?

saleor/graphql/shop/mutations.py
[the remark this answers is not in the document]
maarcingebala: I think we can mention that `0` or `null` disables reservations. I think it's find to make `Int` fields in inputs optional, because as @rafalp said this makes partial updates easier.

fowczarek (approved):

saleor/graphql/schema.graphql
rafalp: It came from picking changes in this file from master. Interesting it shows up in git changes record, but not in github diff.

jbergstroem:
Would it make sense to allow the user to "extend" the reservation time by revisiting the checkout? Also; visiting an expired checkout – in the case of the stock still being available – could also automatically reserve it again. 

(Both scenarios would improve the client journey towards a successful purchase)

PR #7590, opened 2021-07-05
Product variant channel listing webhooks (stocks)

mstrumeck:
I want to merge this change because it adds two new webhooks:

- product variant out of stock
- product variant back in stock

Affected mutations:

- productVariantStocksDelete
- productVariantStocksCreate 
- draftOrderComplete 
- orderBulkCancel
- checkoutComplete 
- orderLinesCreate
- orderLineUpdate 
- orderFulfillmentRefundProducts
- productVariantStocksUpdate 
- orderFulfillmentReturnProducts
- orderLineDelete
- orderFullfill

Webhooks will trigger only in situations when all warehouses within a given channel will contain no stocks (or stocks with quantity equal to 0) for a given product variant.

Related with ticket SALEOR-2993

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


patrys:
Task linked: [SALEOR-2993 Product variant channel listing webhooks (stocks)](https://app.clickup.com/t/2549495/SALEOR-2993)

saleor/graphql/product/tests/test_product.py
```diff
@@ -9668,6 +9673,65 @@ def test_update_or_create_variant_stocks_empty_stocks_data(variant, warehouses):
...
+    info.context.plugins = get_plugins_manager()
+    stocks_data = [
+        {"quantity": 10, "warehouse": "123"},
+    ]
+
+    ProductVariantStocksUpdate.update_or_create_variant_stocks(
+        variant, stocks_data, warehouses, info
+    )
+
+    assert (
```
korycins, line 9695: You're testing here partially `PluginsManager` and `WebhookPlugin` functionality. 
It would be better if you would patch directly PluginsManager calls. Something like this:
`@patch("saleor.plugins.manager.PluginsManager.order_updated")`
You can use two patches for each method, then you can use `assert_called_once_with`.  Thanks to this, there is no need to worry about order of method calls (currently you're using call args list)

saleor/graphql/product/tests/test_product.py
```diff
@@ -9668,6 +9673,65 @@ def test_update_or_create_variant_stocks_empty_stocks_data(variant, warehouses):
...
+    )
+    assert (
+        mocked_webhook_trigger.call_args_list[1][0][0]
+        == "product_variant_stock_changed"
+    )
+    assert mocked_webhook_trigger.call_count == 2
+
+
+@patch("saleor.plugins.webhook.plugin.trigger_webhooks_for_event.delay")
+def test_update_or_create_variant_stocks_with_stock_changed_webhook_only(
```
korycins, line 9706: Same as above, we can mock method of manager

saleor/graphql/product/tests/test_product.py
```diff
@@ -9668,6 +9673,65 @@ def test_update_or_create_variant_stocks_empty_stocks_data(variant, warehouses):
...
+    mocked_webhook_trigger, settings, variant, warehouses, info
+):
+    Stock.objects.create(product_variant=variant, warehouse=warehouses[0], quantity=5)
+    from saleor.plugins.manager import get_plugins_manager
```
korycins, line 9710: Why do you have imports inside a test?

korycins (commented):
- For further PRs - It is much easier for the reviewer to get a context when PR has any description of changes.
- Missing changelog

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -705,22 +705,54 @@ def perform_mutation(cls, root, info, **data):
...
+                    "stock": stock,
+                    "created": created,
+                    "prev_quantity": stock.quantity,
+                    "current_quantity": stock_data["quantity"],
+                }
+            )
             stock.quantity = stock_data["quantity"]
-            stocks.append(stock)
-        warehouse_models.Stock.objects.bulk_update(stocks, ["quantity"])
+        stocks_to_update = [s["stock"] for s in stocks]
```
tomaszszymanski129, line 730: - vars should have meaningfull names, avoid using short ones like `s`.
- even if we name it `stock`, this line would be `[stock["stock"] for stock in stocks]` which is confusing
- `stocks` array should be renamed as it is no longer an array of stock objects, I would name it `stocks_data` and rename the incoming fuction arg to `stocks` to keep signature consistent.

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -705,22 +705,54 @@ def perform_mutation(cls, root, info, **data):
...
-            cls.update_or_create_variant_stocks(variant, stocks, warehouses)
+            cls.update_or_create_variant_stocks(variant, stocks, warehouses, info)
 
         variant = ChannelContext(node=variant, channel_slug=None)
         return cls(product_variant=variant)
 
     @classmethod
     @traced_atomic_transaction()
-    def update_or_create_variant_stocks(cls, variant, stocks_data, warehouses):
+    def update_or_create_variant_stocks(cls, variant, stocks_data, warehouses, info):
```
tomaszszymanski129, line 715: Inconsistent argument naming (`stocks_data` here, `stocks` in call)

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -705,22 +705,54 @@ def perform_mutation(cls, root, info, **data):
...
+                    "prev_quantity": stock.quantity,
+                    "current_quantity": stock_data["quantity"],
+                }
+            )
             stock.quantity = stock_data["quantity"]
-            stocks.append(stock)
-        warehouse_models.Stock.objects.bulk_update(stocks, ["quantity"])
+        stocks_to_update = [s["stock"] for s in stocks]
+        warehouse_models.Stock.objects.bulk_update(stocks_to_update, ["quantity"])
+        cls._run_product_variant_stock_exists_webhook(info, stocks)
```
tomaszszymanski129, line 732: We don't need to pass the entire `info` object as we only need the plugins manager. We could pass `info.context.plugins` instead of `info` and in  
`_run_product_variant_stock_exists_webhook`,
`_run_product_variant_stock_changed_webhook` 
receive `(cls, manager, updated_stocks)`

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -705,22 +705,54 @@ def perform_mutation(cls, root, info, **data):
...
+        cls._run_product_variant_stock_changed_webhook(info, stocks_to_update)
+
+    @classmethod
+    def _run_product_variant_stock_exists_webhook(cls, info, updated_stocks):
+        for stock in (
+            s["stock"]
+            for s in updated_stocks
+            if cls._product_variant_webhook_conditions(s)
```
tomaszszymanski129, line 740: please don't nest `for` loops, extract the inner part outside.

mstrumeck: The `_product_variant_webhook_conditions` return not loop but `bool`.

tomaszszymanski129: I've just meant to get the for out of for :) but maybe it's better this way to keep it compact.

maarcingebala: I agree with @tomaszszymanski129, nested for-loops make it harder to read.

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -693,34 +693,71 @@ class Meta:
...
+
+    @classmethod
+    def _run_product_variant_stock_changed_webhook(
+        cls, plugins_manager, updated_stocks
+    ):
+        for stock in updated_stocks:
+            plugins_manager.product_variant_stock_changed(stock.product_variant)
+
+    @classmethod
+    def _product_variant_webhook_conditions(cls, updated_stock):
```
maarcingebala, line 755: I would rename this function so that the name suggests what it checks, e.g. `_is_variant_back_in_stock`

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable, unreleased changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+- Product variant channel listing webhooks (stocks) - #SALEOR-2993 by @mstrumeck
```
maarcingebala, line 7: SALEOR-2993 is an internal task ID; in the changelog, we want to refer to PRs or commits that introduce described changes. Also, it may sounds too vague to the community; we want to describe as clearly as possible what has changed from their perspective. My suggestion:

```suggestion
- Add webhooks for stock changes: `PRODUCT_VARIANT_STOCK_CHANGED` and `PRODUCT_VARIANT_BACK_IN_STOCK`  - #7590 by @mstrumeck
```

maarcingebala:
Since we're adding new webhooks we should also update our docs: https://docs.saleor.io/docs/3.0/developer/extending/apps/asynchronous-webhooks#available-webhook-events

saleor/settings.py
```diff
@@ -21,6 +21,8 @@
 
 from . import patched_print_object
 
+PATTERNS_IGNORED_IN_QUERY_CAPTURES = ''
```
fowczarek, line 24: Why are you adding this to setttings?

korycins: 🚔

mstrumeck: My mistake, sorry ^^

mstrumeck:
PR with updated docs: https://github.com/mirumee/saleor-docs/pull/292

maarcingebala (approved):

saleor/webhook/payloads.py
```diff
@@ -406,11 +406,16 @@ def generate_product_variant_payload(product_variants: Iterable["ProductVariant"
             "channel_listings": lambda v: json.loads(
                 generate_product_variant_listings_payload(v.channel_listings.all())
             ),
+            "stocks": lambda v: generate_product_variant_stocks_payload(v),
```
korycins, line 409: I am not sure of this.  You return the quantity that is assigned to the stock in `stocks` field.  If we will want to extend variant's payload in the future by adding stocks data, we won't be able to extend this field as it stores int value currently. IMO it should be a custom field like `quantity_in_stocks` or something

tomaszszymanski129 (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable, unreleased changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+- Add webhooks for stock changes: `PRODUCT_VARIANT_OUT_OF_STOCK` and `PRODUCT_VARIANT_BACK_IN_STOCK`  - #7590 by @mstrumeck
```
tomaszszymanski129, line 7: newest changes goes on the bottom of unreleased section

saleor/settings.py
```diff
@@ -21,6 +21,8 @@
 
 from . import patched_print_object
 
+PATTERNS_IGNORED_IN_QUERY_CAPTURES = ""
```
korycins, line 24: Where do you sue it?

mstrumeck: My bad... again...

korycins (approved):

tomaszszymanski129 (approved):

CHANGELOG.md
mstrumeck: Done

eCzerniak (changes requested):
1. Request ProductVariantCahnnelListingUpdate is sent twice. 
2. Nothing is coming if an admin will create order and will add all available stocks.

mstrumeck:
> 1. Request ProductVariantCahnnelListingUpdate is sent twice.
> 2. Nothing is coming if an admin will create order and will add all available stocks.

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -693,67 +694,40 @@ class Meta:
...
         return cls(product_variant=variant)
 
     @classmethod
     @traced_atomic_transaction()
-    def update_or_create_variant_stocks(cls, variant, stocks_input, warehouses, info):
-        stocks_data = []
-        for stock_data, warehouse in zip(stocks_input, warehouses):
-            stock, created = warehouse_models.Stock.objects.get_or_create(
+    def update_or_create_variant_stocks(cls, variant, stocks_data, warehouses):
+        product_variant_stock_trigger = ProductVariantStockWebhookTrigger()
```
korycins, line 717: What does this class do for you? I don't see any prons that you achieved by converting logic to class.  You can do that the same by building standard functions.  If you want to have a separation between logic and API you can always add a file as we have for orders (actions.py)

saleor/warehouse/utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class ProductVariantStockWebhookTrigger:
+    _stock_data = namedtuple(
+        "_stock_data", ("stock", "is_created", "prev_quantity", "current_quantity")
+    )
+
+    def __init__(self):
+        self._stocks_data = []
+        self._plugins_manager = get_plugins_manager()
```
korycins, line 13: Always take manager from context.
Also, not sure if Warehouse/utils is a good place for this. I understand that webhook is called when we make changes in stock, but we raise an event dedicated to ProductVariant. IMHO the logic for raising these webhooks should be in the product module.
I really don't think that we get here any prons that this is a class

saleor/warehouse/management.py
```diff
@@ -198,12 +199,17 @@ def increase_stock(
...
     stock = (
         Stock.objects.select_for_update(of=("self",))
         .filter(warehouse=warehouse, product_variant=order_line.variant)
         .first()
     )
     if stock:
+        product_variant_stock_trigger.append_stock_data(
+            stock, False, stock.quantity, quantity
+        )
+        product_variant_stock_trigger.trigger_product_variant_back_in_stock_webhook()
```
korycins, line 212: - you're calling it before we increase stock
-  stock logic is called in a transaction which means, that there is a possibility that your hook will be called before we will update all data in DB.

fowczarek: We should trigger webhook in `on_commit`

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -713,14 +714,20 @@ def perform_mutation(cls, root, info, **data):
...
                 product_variant=variant, warehouse=warehouse
             )
+            product_variant_stock_trigger.append_stock_data(
+                stock, is_created, stock.quantity, stock_data["quantity"]
+            )
             stock.quantity = stock_data["quantity"]
             stocks.append(stock)
         warehouse_models.Stock.objects.bulk_update(stocks, ["quantity"])
+        product_variant_stock_trigger.trigger_product_variant_back_in_stock_webhook()
+        product_variant_stock_trigger.trigger_product_variant_out_of_stock_webhook()
```
fowczarek, line 730: We should trigger these actions on `on_commit` of this transaction.

fowczarek (changes requested):

saleor/graphql/product/bulk_mutations/products.py
mstrumeck: 1.The only reason why I've decided to use a class was to store somewhere all stocks with necessary data.  That class is nothing more than a simple container, with two public methods.
2. I have to apply my logic in many different places in our code. In that way, I can inject my logic everywhere where I need it within 4 -5 lines instead of 20-25.

korycins: We don't use classes in Saleor if we don't really need them. 
You can easily cover your case with two simple function, something like this:  
```

def _is_variant_back_to_stock(updated_stock):
    return (
        not updated_stock.is_created
        and updated_stock.prev_quantity <= 0
        and updated_stock.current_quantity > 0
    )

def trigger_product_variant_back_in_stock_webhook(stocks_data, manager):
    for data in stocks_data:
        if _is_variant_back_to_stock(data):
           manager.product_variant_back_in_stock(
                data.stock.product_variant
            )

def trigger_product_variant_out_of_stock_webhook(stocks_data, manager):
    for data in stocks_data:
        if data.current_quantity <= 0:
            manager.product_variant_out_of_stock(
                data.stock.product_variant
            )
```

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -713,14 +716,25 @@ def perform_mutation(cls, root, info, **data):
...
+            stock, is_created = warehouse_models.Stock.objects.get_or_create(
                 product_variant=variant, warehouse=warehouse
             )
+            product_variant_stock_webhooks.append_stock_data(
+                stock, is_created, stock.quantity, stock_data["quantity"]
+            )
             stock.quantity = stock_data["quantity"]
             stocks.append(stock)
         warehouse_models.Stock.objects.bulk_update(stocks, ["quantity"])
+        plugins_manager = get_plugins_manager()
```
fowczarek, line 731: We should avid fetching `plugins_manager` from the database. Could we get it from context?
(`manager = info.context.plugins`)

fowczarek (changes requested):

saleor/warehouse/management.py
```diff
@@ -198,12 +211,20 @@ def increase_stock(
...
     """
+    plugins_manager = get_plugins_manager()
+    product_variant_stock_trigger = ProductVariantStockWebhookTrigger()
     stock = (
         Stock.objects.select_for_update(of=("self",))
         .filter(warehouse=warehouse, product_variant=order_line.variant)
         .first()
     )
     if stock:
+        product_variant_stock_trigger.append_stock_data(
```
korycins, line 222: `back_in_stock` should be called for both case. If new stock is created for released variant, it means that it is returned to stock.

saleor/warehouse/management.py
```diff
@@ -198,12 +211,20 @@ def increase_stock(
...
+            stock, False, stock.quantity, quantity
+        )
+        product_variant_stock_trigger.trigger_product_variant_back_in_stock_webhook(
+            plugins_manager
+        )
         stock.increase_stock(quantity, commit=True)
     else:
         stock = Stock.objects.create(
```
korycins, line 230: Also 👇  ( I can't comment below lines but this comment is to 233-242.) there is a possibility that stock will be immediately allocated which means that back_in_stock is not true.

saleor/warehouse/management.py
```diff
@@ -348,14 +381,16 @@ def _decrease_stocks_quantity(
...
-
         stock.quantity = stock.quantity - line_info.quantity
         stocks_to_update.append(stock)
 
     if insufficient_stocks:
         raise InsufficientStock(insufficient_stocks)
 
     Stock.objects.bulk_update(stocks_to_update, ["quantity"])
+    plugins_manager = get_plugins_manager()
+    for v in variants_for_webhooks:
```
tomaszszymanski129, line 392:
```suggestion
    for variant in variants_for_webhooks:
```

saleor/warehouse/tests/test_stock_management.py
```diff
@@ -574,3 +593,229 @@ def test_deallocate_stock_for_order(
...
+        ],
+    )
+
+    flush_post_commit_hooks()
+
+    product_variant_out_of_stock_webhook_mock.assert_called_once()
+
+
+@mock.patch("saleor.plugins.manager.PluginsManager.product_variant_out_of_stock")
+def test_decrease_stock_with_out_of_stock_webhook_triggered_among_warehouses(
```
tomaszszymanski129, line 643: this test is way too long

tomaszszymanski129 (changes requested):

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -706,23 +711,40 @@ def perform_mutation(cls, root, info, **data):
...
             stock.quantity = stock_data["quantity"]
             stocks.append(stock)
         warehouse_models.Stock.objects.bulk_update(stocks, ["quantity"])
 
+        trigger_product_variant_back_in_stock_webhook(
+            variant_for_webhook, plugins_manager
+        )
+        trigger_product_variant_out_of_stock_webhook(
+            variant_for_webhook, plugins_manager
+        )
```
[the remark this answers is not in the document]
mstrumeck: We are triggering both webhooks in that method because we can create a new stock with a value bigger than 0 (case for back_in_stock) or update the existing stock object to 0 (case for out_of_stock).
But each trigger method is not equivalent to the manager method - under these functions we have some checks if webhook can be triggered or not.
I will change the names of these functions because they might missleading.

saleor/warehouse/tests/test_stock_management.py
mstrumeck: We need this test. We have to be sure for 100% that webhooks will recognize changes only for a given channel. But I've decided to conduct some refactor. Please, let me know if this is fine for you.

saleor/graphql/product/tests/test_product.py
```diff
@@ -9748,6 +9751,185 @@ def test_update_or_create_variant_stocks_empty_stocks_data(variant, warehouses):
...
+    settings,
+    variant,
+    warehouses,
+    info,
+    shipping_zone,
+):
+
+    for warehouse in warehouses:
+        warehouse.shipping_zones.set([shipping_zone])
+        warehouse.save()
```
korycins, line 9768: IMHO `.save()` is not needed here 🤔

saleor/graphql/product/tests/test_product.py
```diff
@@ -9748,6 +9751,185 @@ def test_update_or_create_variant_stocks_empty_stocks_data(variant, warehouses):
...
+        Stock.objects.create(product_variant=variant, warehouse=warehouse)
```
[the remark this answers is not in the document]
korycins: Make sure that you apply this comment to all tests where you call `.create` more than once.

saleor/graphql/product/tests/test_product.py
korycins: Make sure that you apply this comment to all tests where you call `.save` and `.set`

saleor/settings.py
tomaszszymanski129: Why did you mark it as resolved? Comments like that automatically resolves themselves after line gets changed. Do not resolve them manually unless you're sure you've changed related part and the code is pushed.

tomaszszymanski129 (changes requested):

saleor/product/actions.py
```diff
@@ -0,0 +1,88 @@
...
+        [
+            _get_product_variant_quantity(product_variant, channel_slug)
+            for channel_slug in channel_slugs
+        ]
+    )
+
+
+def _get_product_variant_quantity(product_variant: ProductVariant, channel_slug: str):
+    return (
+        product_variant.stocks.for_channel(channel_slug).aggregate(  # type: ignore
```
korycins, line 84: @fowczarek  What do you think? Can we improve this call? It looks like many DB calls

saleor/warehouse/management.py
```diff
@@ -198,17 +205,32 @@ def increase_stock(
...
+            order_line.variant,
+            warehouse.shipping_zones.values_list("channels__slug", flat=True),
+            prev_quantity=0,
+        )
+
         stock = Stock.objects.create(
             warehouse=warehouse, product_variant=order_line.variant, quantity=quantity
         )
+
     if allocate:
```
korycins, line 234: Maybe I am missing something, but Shouldn't we take into consideration that stock can be imedietly allocated here? So calling `back_in_stock` will not be true?
Related comment: https://github.com/mirumee/saleor/pull/7590/files#r675384459

saleor/warehouse/tests/utils.py
```diff
@@ -1,6 +1,8 @@
 from django.db.models import Sum
 from django.db.models.functions import Coalesce
 
+from saleor.warehouse.models import Stock, Warehouse
```
korycins, line 4: I believe it should be a realtive import

saleor/product/actions.py
tomaszszymanski129: definitely! We could annotate stocks qs with `for_channels` method, it's a matter of copying `for_channel` and changing `Channel.objects.filter(slug=channel_slug).values("pk")` to `Channel.objects.filter(slug__in=channel_slugs).values("pk")`
Then we can skip the comprehension in `_get_product_variant_total_quantity` which will save a lot of calls.

saleor/warehouse/management.py
mstrumeck: But allocation itself will not increase or decrease the total quantity of a given Stock object.  Sending webhooks after each allocation changing (for example, via orderLineCRUD mutations) can lead us to unwanted side effects.

saleor/product/actions.py
mstrumeck: I'm not a huge fan of copy-past so I've decided to small refactor - two public methods and a common private method. @korycins @tomaszszymanski129 please let me know if that is ok for you.

saleor/product/actions.py
```diff
@@ -0,0 +1,79 @@
...
+def _is_variant_back_to_stock(product_variant_data: product_variant_webhook_container):
+    variant = product_variant_data.product_variant
+    slugs = product_variant_data.channel_slugs
+    return _get_product_variant_total_quantity(variant, slugs) and (
+        product_variant_data.prev_quantity <= 0
+    )
+
+
+def _get_product_variant_total_quantity(
+    product_variant: ProductVariant, channel_slugs: typing.Iterable
```
tomaszszymanski129, line 72:
```suggestion
    product_variant: ProductVariant, channel_slugs: Iterable[str]
```

saleor/product/actions.py
tomaszszymanski129: Looks good 👍🏻

tomaszszymanski129 (approved):

saleor/warehouse/management.py
korycins: Maybe I am wrong but re-creating stock and then immediately create allocation means that the new stock which was just created is already allocated, which means that you don't have any free stock.

saleor/graphql/warehouse/mutations.py
```diff
@@ -165,3 +166,15 @@ class Meta:
...
+    @classmethod
+    def perform_mutation(cls, _root, info, **data):
+        manager = info.context.plugins
+        node_id = data.get("id")
+        model_type = cls.get_type_for_model()
+        instance = cls.get_node_or_error(info, node_id, only_type=model_type)
+        stocks = (stock for stock in instance.stock_set.only("product_variant"))
+        result = super(WarehouseDelete, cls).perform_mutation(_root, info, **data)
+        for stock in stocks:
+            trigger_out_of_stock_webhook(stock, manager)
```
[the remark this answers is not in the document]
fowczarek: We should also check that we receive a proper payload in delete mutation. If you generate payload after you remove stock. `stock.id` for this object will be `None`.

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -755,18 +775,26 @@ class Meta:
...
-        warehouse_models.Stock.objects.filter(
+        stocks_to_delete = warehouse_models.Stock.objects.filter(
             product_variant=variant, warehouse__pk__in=warehouses_pks
-        ).delete()
+        )
 
         variant = ChannelContext(node=variant, channel_slug=None)
+
+        for stock in stocks_to_delete:
+            trigger_out_of_stock_webhook(stock, manager)
```
fowczarek, line 794: We should also check that we receive a proper payload in delete mutation. If you generate payload after you remove stock. `stock.id` for this object will be `None`

saleor/product/actions.py
```diff
@@ -0,0 +1,21 @@
...
+    allocation, allocation_quantity_before, manager: PluginsManager
+):
+    available_stock_now = allocation.stock.available_quantity()
+
+    if allocation_quantity_before == 0 and available_stock_now > 0:
+        trigger_back_in_stock_webhook(allocation.stock, manager)
+
+
+def trigger_out_of_stock_webhook(stock: "Stock", manager: PluginsManager):
+    transaction.on_commit(lambda: manager.product_variant_out_of_stock(stock))
```
fowczarek, line 17: We shouldn't use `transaction` here, because someone may to you this webhook outside of a transaction.

saleor/product/actions.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def trigger_back_in_stock_webhook(stock: "Stock", manager: PluginsManager):
+    transaction.on_commit(lambda: manager.product_variant_back_in_stock(stock))
```
fowczarek, line 21: We shouldn't use `transaction` here, because someone may to you this webhook outside of a transaction.

saleor/order/tests/test_order_utils.py
```diff
@@ -4,6 +4,8 @@
 import pytest
 from prices import Money, TaxedMoney
 
+from saleor.plugins.manager import get_plugins_manager
```
fowczarek, line 7: Absolute import.

saleor/warehouse/availability.py
```diff
@@ -71,14 +71,16 @@ def check_stock_quantity_bulk(
...
             insufficient_stocks.append(
                 InsufficientStockData(
-                    variant=variant, available_quantity=available_quantity
+                    variant=variant,
+                    available_quantity=available_quantity,  # type: ignore
                 )
             )
         elif variant.track_inventory:
-            if quantity > available_quantity:
+            if quantity > available_quantity:  # type: ignore
```
fowczarek, line 79: Why are we adding here `# type: ignore`?

saleor/warehouse/management.py
```diff
@@ -84,6 +93,16 @@ def allocate_stocks(
     if allocations:
         Allocation.objects.bulk_create(allocations)
 
+        for allocation in allocations:
+            allocated_stock = (
+                allocation.stock.allocations.aggregate(Sum("quantity_allocated"))[
```
fowczarek, line 98:
```suggestion
                Allocations.object.filter(stock_id=allocation.stock_id).aggregate(Sum("quantity_allocated"))[
```

saleor/warehouse/management.py
```diff
@@ -178,8 +199,18 @@ def deallocate_stock(order_lines_data: Iterable["OrderLineData"]):
     if not_dellocated_lines:
         raise AllocationError(not_dellocated_lines)
 
+    available_stock_before = [
+        (allocation, allocation.stock.available_quantity())
```
fowczarek, line 203: Multiple database queries. 
We fetch stock and in the next query fetch `available_quantity` for this query. 
We should fetch it like `Allocations.object.filter(stock_id=allocation.stock_id).aggregate(Sum("quantity_allocated"))`

saleor/warehouse/management.py
```diff
@@ -307,14 +343,17 @@ def decrease_stock(order_lines_info: Iterable["OrderLineData"], update_stocks=Tr
...
-
     if update_stocks:
         _decrease_stocks_quantity(
             order_lines_info,
             variant_and_warehouse_to_stock,
             quantity_allocation_for_stocks,
         )
 
+        for stock in stocks:
+            if stock.available_quantity() <= 0:
```
fowczarek, line 354: In your soultion we execute databse query for each line separately. 
```suggestion
        for stock in stocks.annotate_available_quantity():
            if stock.available_quantity<= 0:
```

saleor/warehouse/management.py
```diff
@@ -370,9 +408,14 @@ def get_order_lines_with_track_inventory(
...
 @traced_atomic_transaction()
-def deallocate_stock_for_order(order: "Order"):
+def deallocate_stock_for_order(order: "Order", manager: PluginsManager):
     """Remove all allocations for given order."""
     allocations = Allocation.objects.filter(
         order_line__order=order, quantity_allocated__gt=0
     ).select_for_update(of=("self",))
+
+    for allocation in allocations:
+        if allocation.stock.available_quantity() <= 0:
```
fowczarek, line 418: We execute database queries for each allocation separately. We should do it in a single database query.

saleor/warehouse/models.py
```diff
@@ -66,24 +67,12 @@ def annotate_available_quantity(self):
...
-            )
-        ).values("warehouse_id")
-        return self.select_related("product_variant").filter(
-            Exists(
-                warehouse_shipping_zones.filter(warehouse_id=OuterRef("warehouse_id"))
-            )
-        )
+        return self._stock_from_channels(channels)
+
+    def for_channels(self, channel_slugs: typing.List[str]):
```
fowczarek, line 73: This function is not used in this PR. Should we revert those changes?

saleor/warehouse/models.py
```diff
@@ -139,6 +147,15 @@ def decrease_stock(self, quantity: int, commit: bool = True):
         if commit:
             self.save(update_fields=["quantity"])
 
+    def available_quantity(self):
```
fowczarek, line 150: IMO we should calculate it explicitly instead of using this function. 
Someone can accidently generate a lot of database queries when using this function.

saleor/warehouse/tests/test_stock_management.py
```diff
@@ -1,10 +1,15 @@
+from unittest import mock
+
 import pytest
 from django.db.models import Sum
 from django.db.models.functions import Coalesce
 
+from saleor.plugins.manager import get_plugins_manager
```
fowczarek, line 7: Absolut imports.

saleor/webhook/payloads.py
```diff
@@ -396,6 +396,23 @@ def generate_product_variant_media_payload(product_variant):
...
+    serializer = PayloadSerializer()
+    extra_dict_data = {
+        "product_id": lambda v: graphene.Node.to_global_id(
+            "Product", v.product_variant.product.id
+        ),
+        "product_variant_id": lambda v: graphene.Node.to_global_id(
+            "ProductVariant", v.product_variant.id
+        ),
+        "warehouse_id": lambda v: graphene.Node.to_global_id(
+            "Warehouse", v.warehouse.id
```
fowczarek, line 409:
```suggestion
            "Warehouse", v.warehouse_id
```

saleor/webhook/payloads.py
```diff
@@ -396,6 +396,23 @@ def generate_product_variant_media_payload(product_variant):
     ]
 
 
+def generate_product_variant_with_stock_payload(stocks: Iterable["Stock"]):
+    serializer = PayloadSerializer()
+    extra_dict_data = {
+        "product_id": lambda v: graphene.Node.to_global_id(
+            "Product", v.product_variant.product.id
```
fowczarek, line 403:
```suggestion
            "Product", v.product_variant.product_id
```

saleor/webhook/payloads.py
```diff
@@ -396,6 +396,23 @@ def generate_product_variant_media_payload(product_variant):
...
+        ),
+        "product_variant_id": lambda v: graphene.Node.to_global_id(
+            "ProductVariant", v.product_variant.id
```
fowczarek, line 406:
```suggestion
            "ProductVariant", v.product_variant_id
```

saleor/webhook/payloads.py
```diff
@@ -396,6 +396,23 @@ def generate_product_variant_media_payload(product_variant):
     ]
 
 
+def generate_product_variant_with_stock_payload(stocks: Iterable["Stock"]):
```
fowczarek, line 399: We should be sure that `Product` and `ProductVariant` are prefetched for this function.

saleor/webhook/payloads.py
```diff
@@ -408,11 +425,16 @@ def generate_product_variant_payload(product_variants: Iterable["ProductVariant"
             "channel_listings": lambda v: json.loads(
                 generate_product_variant_listings_payload(v.channel_listings.all())
             ),
+            "quantity_in_stocks": lambda v: generate_product_variant_stocks_payload(v),
```
fowczarek, line 428: This change is not mentioned in the changelog.

fowczarek: This field misses information about allocations. Are you sure that is proper behavior?

fowczarek: This field is also returning one value for all warehouses are we sure that is proper behavior?

saleor/graphql/warehouse/mutations.py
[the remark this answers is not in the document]
mstrumeck: 1. Body of `trigger_out_of_stock_webhook` function already contain `transaction.on_commit`
2. I've created a list of stocks to trigger before any mutation. I have a suitable test case for that.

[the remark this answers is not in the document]
fowczarek: 1. Tests are executed inside a transaction. This mutation doesn't use transactions. 
2. This test checks only `trigger_out_of_stock_webhook` is called. If you generate payload from a removed object you receive `object_graphql_id=base64("TypeName: None")` instead of `object_graphql_id=base64("TypeName: <Object.pk>")`

saleor/warehouse/availability.py
mstrumeck: Because I've got mypy errors:
`Argument "available_quantity" to "InsufficientStockData" has incompatible type "Union[Callable[[], Any], int]"; expected "Optional[int]"  [arg-type].`
I have no idea why. I didn't change any line of code within those files nor related files.

saleor/warehouse/management.py
mstrumeck: I can't just conduct annotations because `available_quantity` is a method of the `Stock` object.

saleor/graphql/product/bulk_mutations/products.py
```diff
@@ -646,9 +647,15 @@ def perform_mutation(cls, root, info, **data):
             warehouses = cls.clean_stocks_input(variant, stocks, errors)
             if errors:
                 raise ValidationError(errors)
-            create_stocks(variant, stocks, warehouses)
+            new_stocks = create_stocks(variant, stocks, warehouses)
+
+            for stock in new_stocks:
+                transaction.on_commit(
```
fowczarek, line 653: This mutation doesn't use transaction.

fowczarek (approved):

fowczarek (approved):

saleor/graphql/product/tests/test_product.py
```diff
@@ -10024,6 +10027,125 @@ def test_update_or_create_variant_stocks_empty_stocks_data(variant, warehouses):
...
+    ProductVariantStocksUpdate.update_or_create_variant_stocks(
+        variant, stocks_data, warehouses, info.context.plugins
+    )
+
+    assert variant.stocks.aggregate(Sum("quantity"))["quantity__sum"] == 0
+
+    flush_post_commit_hooks()
+    product_variant_back_in_stock_webhook.assert_not_called()
+    product_variant_stock_out_of_stock_webhook.assert_called_once_with(
+        Stock.objects.all()[1]
```
korycins, line 10102: Why not .first/.last? Would be more readble IMO

korycins (approved):

eCzerniak (approved):

PR #7618, opened 2021-07-12
Store users passwords in env for populatedb command

d-wysocki:
I want to merge this change because...I want to store users passwords in ENV for populatedb command

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-3040 Don't keep users password as plaintext in populatedb](https://app.clickup.com/t/2549495/SALEOR-3040)

IKarbowiak (approved):

saleor/settings.py
```diff
@@ -647,3 +647,16 @@ def get_host():
 
 assert hasattr(schema_printer, "_print_object")
 schema_printer._print_object = patched_print_object
+
+# Settings for populatedb command
+DEFAULT_USERS_PASSWORD = "password"
+
+POPULATE_SUPER_USER_PASSWORD = os.environ.get(
+    "POPULATE_SUPER_USER_PASSWORD", DEFAULT_USERS_PASSWORD
```
korycins, line 655: Is it expected that this will change our default password for `admin@example.com` from `admin` to `password`?
Can't we still use old good `admin` here?

maarcingebala: We shouldn't change the "admin" password. It is used for example in our demo where we automatically prefill it in the frontend, so changing it here will require us to also fix the demo.

Also, I think we only need at most two env variables:
1) for superuser - "admin"
2) for everyone else - "password"

fowczarek:
:police_officer: 
This PR don't fix assigned issue. 
image

fowczarek (changes requested):
In `create_fake_user` we should use `user = User.objects.create_user(` instead of `user = User(`

fowczarek (approved):

tomaszszymanski129 (approved):

saleor/settings.py
```diff
@@ -647,3 +647,14 @@ def get_host():
 
 assert hasattr(schema_printer, "_print_object")
 schema_printer._print_object = patched_print_object
+
+# Settings for populatedb command
```
korycins, line 651: Do we really need to have these envs for `populateDB` script inside a `settings.py` file?  It is something that is a useful only for local development or sandboxes but I am not sure that we should place it here. Maybe we can put them directly in the command file? And use there `os.environ` also?
@maarcingebala  What do you think? settings.py is growing and growing not sure that we should place here settings which are required only by a command line.

tomaszszymanski129: I agree, now if I think on it, I would consider going even further and drop the env var (use hardcoded constant). It feels unlikely to me that someone will ever gonna make a use of it, the command arg seems more appealing.

korycins (approved):

PR #7622, opened 2021-07-13
Add app support for events

IKarbowiak:
Add app support in `order`, `account`, `invoice` events.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-3735 Events do not support apps](https://app.clickup.com/t/2549495/SALEOR-3735)

tomaszszymanski129 (approved):
Nicely done! 👍🏻 For future - it would be cool to have it as separate PRs to feature branch - this one could've been splitted to `orders`, `accounts` and `invoices`, which would make it much easier to review.

saleor/account/events.py
```diff
@@ -75,39 +77,42 @@ def customer_downloaded_a_digital_link_event(
     )
 
 
-def staff_user_deleted_a_customer_event(
-    *, staff_user: UserType, deleted_count: int = 1
+def staff_or_app_deleted_a_customer_event(
```
korycins, line 80: Maybe instead of adding `staff_or_app`  in the name, we could just name it `customer_deleted_event`? 
Also, I would consider adding single args which will accept User or App (like requestor), then before we will create new event  (like we did in line 85), we can determine if the requestor is an app or user. This should simplify the usage of this function - like I don't need to fetch user and app separatly. I will just fetch the requestor which called this mutation, and pass it to the event function. The event function will handle the rest.

maarcingebala (approved):

PR #7633, opened 2021-07-16
Support empty SKUs for products variants

rafalp:
Allows product variants to have no SKU, eg for the case when product's SKU is populated by the integration with 3rd party service or store doesn't use SKU's.

Backend changes summary:

- Product variant can be created without SKU.
- SKU can be removed from product's variant.
- Variants without SKU function like ones with SKU: they are displayed in the storefront, can be added to checkout and ordered.
- SKU fields in GraphQL schema became nullable.
- Tax and payment gateway plugins receive either SKU or variant's GraphQL id for item code.
- Variant's GraphQL id is stored next to SKU on order lines.
- Variant's GraphQL id is included in order webhook payload.
- Variant's GraphQL id is passed to analytics instead of SKU if it's not set.
- Invoices don't display SKU when its not available.
- Product export includes optional "variant id" value that is GraphQL IDidof exported product variant.
- GraphQL id is used to set default name for variant if SKU is not provided.

# TODO

- E-mails, dashboard and example storefront templates need to be checked for SKU usages and updated for scenarios when SKU is not provided.
- Decide if `product_id` should be replaced with different name. I've used `product_id` in order line for both parity with `product_sku` that already exists and to avoid name conflict with `variant_id` already used by the ORM.

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


fowczarek (approved):

fowczarek (commented):
We should also update `ProductsQueryset.published_with_variants` to filter out products without available variants.

maarcingebala (approved):

korycins (approved):

eCzerniak:
Hi @rafalp
1. It should be possible to add variant without SKU to DraftOrder
2. It should be NOT possible to add variant without SKU to not Confirmed Order 
Now only second option is implemented.

rafalp:
@eCzerniak Adding products without SKU to draft orders is the opposite of original requirement:

> Its impossible to add product variant without SKU to draft order in dashboard

saleor/checkout/utils.py
```diff
@@ -685,11 +685,11 @@ def is_shipping_required(lines: Iterable["CheckoutLineInfo"]):
...
 
-    not_available_variants = [
-        variant_id
-        for variant_id, channel_listing in variants_listings_map.items()
-        if channel_listing is None or channel_listing.price is None
-    ]
+    not_available_variants = []
+    for variant_id, channel_listing in variants_listings_map.items():
+        if channel_listing is None or channel_listing.price is None:
+            not_available_variants.append(variant_id)
```
fowczarek, line 691: Can we revert those changes and left this file unchanged.

saleor/product/utils/variant_prices.py
```diff
@@ -13,7 +13,8 @@
 def _get_variant_prices_in_channels_dict(product):
     prices_dict = defaultdict(list)
     for variant_channel_listing in ProductVariantChannelListing.objects.filter(
-        variant__product_id=product, price_amount__isnull=False
+        variant__product_id=product,
+        price_amount__isnull=False,
```
fowczarek, line 17: Can we revert those changes and left this file unchanged.

fowczarek (approved):

saleor/order/models.py
```diff
@@ -418,7 +418,9 @@ class OrderLine(models.Model):
     variant_name = models.CharField(max_length=255, default="", blank=True)
     translated_product_name = models.CharField(max_length=386, default="", blank=True)
     translated_variant_name = models.CharField(max_length=255, default="", blank=True)
-    product_sku = models.CharField(max_length=255)
+    product_sku = models.CharField(max_length=255, null=True, blank=True)
+    # str with GraphQL ID used as fallback when product SKU is not available
+    product_id = models.CharField(max_length=255, null=True, blank=True)
```
korycins, line 423: Wouldn't be useful to have here a foreign key relation instead of graphql's product ID? You can always convert DB id into a graphql ID but relation with the product would be useful.

korycins: also (I can think wrongly here) django uses field like `variant_id` or `product_id` when you define new relation like `variant` or `product`. Would adding this field as `product_id`  won't break this approach?

rafalp: Changing this field to foreign key will defeat the point as it's supposed to preserve GraphQL ID of purchased variant even after this variant has been deleted from the database. But I'll rename this field to something else (perhaps `product_code`) so it won't potentially conflict with `product_id`.

rafalp: I've tried few names but I've settled on `productVariantId` because there are already few places in Saleor that use this approach.

korycins (approved):

PR #7668, opened 2021-07-21
Expose preorder data in API

Kwaidan00:
I want to merge this change because it exposes in API `ProductVariant` preorder settings and number of globally sold units / per channel. It also modifies `productVariant.quantityAvailable` behavior for preorder variants.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/product/types/channels.py
+++ b/saleor/graphql/product/types/channels.py
@@ -229,9 +229,26 @@ class ProductChannelListing(CountableDjangoObjectType):
         )
 
 
+class PreorderThreshold(graphene.ObjectType):
+    quantity = graphene.Int(
+        required=False,
+        description="Preorder threshold for product variant in this channel.",
+    )
+    sold_units = graphene.Int(
+        required=True,
+        description="Number of sold product variant in this channel.",
+    )
+
+    class Meta:
+        description = "Represents preorder variant data for channel."
+
+
 class ProductVariantChannelListing(CountableDjangoObjectType):
     cost_price = graphene.Field(Money, description="Cost price of the variant.")
     margin = graphene.Int(description="Gross margin percentage value.")
+    preorder_threshold = graphene.Field(
+        PreorderThreshold, required=False, description="Preorder variant data."
+    )
 
     class Meta:
         description = "Represents product varaint channel listing."
@@ -247,6 +264,17 @@ class ProductVariantChannelListing(CountableDjangoObjectType):
     @permission_required(ProductPermissions.MANAGE_PRODUCTS)
     def resolve_margin(root: models.ProductVariantChannelListing, *_args):
         return get_margin_for_variant_channel_listing(root)
+
+    @staticmethod
+    def resolve_preorder_threshold(
+        root: models.ProductVariantChannelListing, info, **_kwargs
+    ):
+        # This field is added through annotation when using the
+        # `resolve_channel_listings` resolver.
+        return PreorderThreshold(
+            quantity=root.preorder_quantity_threshold,
+            sold_units=getattr(root, "preorder_quantity_allocated", 0),
+        )
 
 
 class CollectionChannelListing(CountableDjangoObjectType):
--- a/saleor/graphql/product/types/channels.py
+++ b/saleor/graphql/product/types/channels.py
@@ -269,8 +269,8 @@ class ProductVariantChannelListing(CountableDjangoObjectType):
     def resolve_preorder_threshold(
         root: models.ProductVariantChannelListing, info, **_kwargs
     ):
-        # This field is added through annotation when using the
-        # `resolve_channel_listings` resolver.
+        # The preorder_quantity_allocated field is added through annotation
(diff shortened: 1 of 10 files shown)


patrys:
Task linked: [SALEOR-2998 Preorders](https://app.clickup.com/t/2549495/SALEOR-2998)

patrys:
Task linked: [SALEOR-3800 5. Expose preorder settings in API](https://app.clickup.com/t/2549495/SALEOR-3800)

Kwaidan00:
TBH I'm not sure why `mypy` has a problem here run in GitHub Actions; I have `pre-commits` installed and it passes on my local machine.

maarcingebala:
@Kwaidan00 Not so long ago we bumped mypy version, you have the same version locally?

saleor/product/models.py
```diff
@@ -629,6 +638,8 @@ class ProductVariantChannelListing(models.Model):
 
     preorder_quantity_threshold = models.IntegerField(blank=True, null=True)
 
+    objects = ProductVariantChannelListingQuerySet.as_manager()
```
maarcingebala, line 641: You can try to rewrite this definition to:
```suggestion
    objects = models.Manager.from_queryset(ProductVariantChannelListingQuerySet)()
```
This is a workaround for mypy to recognize custom queryset method, but in some cases, it doesn't help; I haven't figured out yet when exactly :/

saleor/graphql/schema.graphql
```diff
@@ -4083,6 +4083,18 @@ enum PostalCodeRuleInclusionTypeEnum {
...
 
+type PreorderData {
+  isPreorder: Boolean!
+  globalThreshold: Int
+  globalSoldUnits: Int
+  endDate: DateTime
+}
+
+type PreorderThreshold {
+  quantity: Int
```
maarcingebala, line 4094: Is this field nullable?

saleor/graphql/schema.graphql
```diff
@@ -4083,6 +4083,18 @@ enum PostalCodeRuleInclusionTypeEnum {
   EXCLUDE
 }
 
+type PreorderData {
+  isPreorder: Boolean!
+  globalThreshold: Int
+  globalSoldUnits: Int
+  endDate: DateTime
```
maarcingebala, line 4090: Are all of these fields nullable? Just making sure that we have correct types, it makes life of frontend devs easier ;)

saleor/graphql/product/types/channels.py
```diff
@@ -248,6 +265,17 @@ def resolve_channel(root: models.ProductVariantChannelListing, info, **_kwargs):
...
         return get_margin_for_variant_channel_listing(root)
 
+    @staticmethod
+    def resolve_preorder_threshold(
+        root: models.ProductVariantChannelListing, info, **_kwargs
+    ):
+        # This field is added through annotation when using the
+        # `resolve_channel_listings` resolver.
+        return PreorderThreshold(
+            quantity=root.preorder_quantity_threshold,
```
maarcingebala, line 275: Is there a way to query it from a different resolver, where this annotation is not available? If so, we should make sure that this resolver will returns None and doesn't fail trying to access a field that is not present.

saleor/graphql/schema.graphql
Kwaidan00: Yes, it is - by definition. User can set threshold to None - it means there is no threshold per channel and only the global threshold should be used (if it is set).

saleor/graphql/schema.graphql
Kwaidan00: - `globalThreshold` should be nullable - user can just not set it, according to spec.
- `endDate` should be nullable - user can just not set it and end preorder manually.
- `globalSoldUnits` - I'm not sure now, it's probably my typo during editing the tech spec. Number of sold units sounds like something that should not be nullable. Moreover, similar field for channel listing is not nullable. I will change it to `Int!`. cc @jwm0

saleor/product/models.py
Kwaidan00: Thanks ;) I've burned out the Python environment and I've set it up from scratch, now mypy works correctly on my local machine.
And your solution fixed that issue, thanks 🙇

saleor/graphql/product/types/channels.py
Kwaidan00: According to schema, the only way to query this type is `channelListings` field in `ProductVariant`. The only field added with annotation is the second one, with `getattr` (I've clarified the comment). I think it's ok to return 0 (`Int!`) with this one field if not available, because user can still see the preorder threshold; however, it should never happen.

fowczarek (approved):

maarcingebala (approved):
All good, but one test needs to be fixed.

Kwaidan00:
It was a flaky test (checking list elements, but there was no order defined); I've fixed two similar tests I've found.

PR #7661, opened 2021-07-20
Adjust amount of spans in resolvers

fowczarek:
I want to merge this change because of adjusting the amount of spans in resolvers.

- [x]  Accounts
- [x]  App
- [x]  Attribute
- [x]  Channel
- [x]  Checkout
- [x]  Core
- [x]  Csv
- [x]  Discount
- [x]  Giftcard
- [x]  Menu
- [x]  Meta
- [x]  Order
- [x]  Page
- [x]  Payment
- [x]  Plugins
- [x]  Product
- [x]  Shipping
- [x]  Shop
- [x]  Translations
- [x]  Warehouse
- [x]  Webhook

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #7673, opened 2021-07-22
[Saleor-2992] click and collect

kuchichan:
Related to: [SALEOR-2992 Click&Collect]( https://app.clickup.com/t/2549495/SALEOR-2992)
I want to merge this change because it adds click and collect functionality to Saleor

# Impact
* [x] New migrations
# Pull Request Checklist
* [ ] The changes are tested


patrys:
Task linked: [SALEOR-2992 Click&Collect](https://app.clickup.com/t/2549495/SALEOR-2992)

saleor/graphql/order/types.py
```diff
@@ -1029,6 +1038,23 @@ def wrap_shipping_method_with_channel_context(data):
...
+            return Order.resolve_shipping_method(root, info)
+        if root.collection_point_id:
+            collection_point = WarehouseByIdLoader(info.context).load(
+                root.collection_point_id
+            )
+            return collection_point
+        return None
+
+    @staticmethod
+    @traced_resolver
```
fowczarek, line 1054: In PR #7661 we drop tracing for simple resolvers. 
```suggestion
```

saleor/graphql/warehouse/dataloaders.py
```diff
@@ -178,4 +178,6 @@ class WarehouseByIdLoader(DataLoader):
 
     def batch_load(self, keys):
         warehouses = Warehouse.objects.in_bulk(keys)
-        return [warehouses.get(UUID(warehouse_uuid)) for warehouse_uuid in keys]
+        return [
+            warehouses.get(UUID(str(warehouse_uuid))) for warehouse_uuid in keys
+        ]  # psycopg converts these UUID's once, possible bug?
```
fowczarek, line 183: What should be done with this comment?

saleor/graphql/warehouse/filters.py
```diff
@@ -1,6 +1,10 @@
 import django_filters
 from graphene_django.filter import GlobalIDMultipleChoiceFilter
 
+from saleor.graphql.core.filters import EnumFilter
+from saleor.graphql.warehouse.enums import WarehouseClickAndCollectOptionEnum
+from saleor.warehouse import WarehouseClickAndCollectOption
```
fowczarek, line 6: It's should be relative imports.

saleor/webhook/tests/test_webhook_payloads.py
```diff
@@ -7,6 +7,8 @@
 
 import graphene
 
+from saleor.warehouse import WarehouseClickAndCollectOption
```
fowczarek, line 10: It's should be relative import.

fowczarek (commented):
I think we should add benchmarks for checkout flow with click and collect. In `saleor/saleor/graphql/checkout/tests/benchmark/test_checkout_mutations.py` you can look at sample checkout benchmarks.

fowczarek:
WE should also update populate db, to generate at least one warehouse with available as Click&Collect.

saleor/graphql/checkout/types.py
```diff
@@ -154,12 +159,28 @@ def is_shipping_required(product_type):
         )
 
 
+class DeliveryMethod(graphene.Union):
+    class Meta:
```
maarcingebala, line 163: We should a description that will explain what this union represents.

saleor/graphql/checkout/types.py
```diff
@@ -187,7 +211,14 @@ class Checkout(CountableDjangoObjectType):
     shipping_method = graphene.Field(
         ShippingMethod,
         description="The shipping method related with checkout.",
+        deprecation_reason="Use delivery method instead",
+    )
+
+    delivery_method = graphene.Field(
+        DeliveryMethod,
+        description="The shipping method related with checkout, or warehouse if C&C",
```
maarcingebala, line 219: If we add a description to the `DeliveryMethod` union, here we could simply say "The delivery method selected for this checkout".

saleor/graphql/order/types.py
```diff
@@ -657,6 +659,13 @@ class Order(CountableDjangoObjectType):
     is_shipping_required = graphene.Boolean(
         description="Returns True, if order requires shipping.", required=True
     )
+    is_click_and_collect = graphene.Boolean(
+        description="Returns True, if order is set as click and collect"
```
maarcingebala, line 663: "order is set as click and collect" - this could be more accurate, for example "Returns true if Click & Collect is selected as a delivery method for this order." Also, we should be consistent when referring to this feature in descriptions, docs, etc. Here we have "click and collect", below we have "C&C".

saleor/graphql/checkout/types.py
```diff
@@ -172,6 +193,9 @@ class Checkout(CountableDjangoObjectType):
     is_shipping_required = graphene.Boolean(
         description="Returns True, if checkout requires shipping.", required=True
     )
+    is_click_and_collect = graphene.Boolean(
+        description="Returns True, if checkout is click and collect"
```
maarcingebala, line 197: "Returns true, if click and collect is chosen as a delivery method for this checkout."

saleor/graphql/schema.graphql
```diff
@@ -928,13 +928,16 @@ type Checkout implements Node & ObjectWithMetadata {
   privateMetadata: [MetadataItem]!
   metadata: [MetadataItem]!
   availableShippingMethods: [ShippingMethod]!
+  availableCollectionPoints: [Warehouse]
```
maarcingebala, line 931: I think this should be `availableCollectionPoints: [Warehouse!]` - I guess we'll never have a list with nulls inside? Also, is this field nullable at all or should it return an empty list when there is no collection points available?

saleor/graphql/schema.graphql
```diff
@@ -928,13 +928,16 @@ type Checkout implements Node & ObjectWithMetadata {
...
   availablePaymentGateways: [PaymentGateway!]!
   email: String!
   isShippingRequired: Boolean!
+  isClickAndCollect: Boolean
```
maarcingebala, line 935: Does it ever return null?

saleor/graphql/schema.graphql
```diff
@@ -3023,6 +3035,8 @@ type Order implements Node & ObjectWithMetadata {
   totalBalance: Money!
   userEmail: String
   isShippingRequired: Boolean!
+  isClickAndCollect: Boolean
```
maarcingebala, line 3038: Is this field nullable?

saleor/graphql/schema.graphql
```diff
@@ -5926,9 +5940,17 @@ type Warehouse implements Node & ObjectWithMetadata {
   shippingZones(before: String, after: String, first: Int, last: Int): ShippingZoneCountableConnection!
   address: Address!
   email: String!
+  isPrivate: Boolean!
   privateMetadata: [MetadataItem]!
   metadata: [MetadataItem]!
   companyName: String! @deprecated(reason: "Use address.CompanyName. This field will be removed in Saleor 4.0.")
+  clickAndCollectOption: WarehouseClickAndCollectOptionEnum
```
maarcingebala, line 5947: Is this field nullable?

saleor/graphql/schema.graphql
```diff
@@ -2887,7 +2898,8 @@ type Mutation {
   checkoutRemovePromoCode(checkoutId: ID, promoCode: String!, token: UUID): CheckoutRemovePromoCode
   checkoutPaymentCreate(checkoutId: ID, input: PaymentInput!, token: UUID): CheckoutPaymentCreate
   checkoutShippingAddressUpdate(checkoutId: ID, shippingAddress: AddressInput!, token: UUID): CheckoutShippingAddressUpdate
-  checkoutShippingMethodUpdate(checkoutId: ID, shippingMethodId: ID!, token: UUID): CheckoutShippingMethodUpdate
+  checkoutShippingMethodUpdate(checkoutId: ID, shippingMethodId: ID!, token: UUID): CheckoutShippingMethodUpdate @deprecated(reason: "Use checkoutDeliveryMethodUpdate instead")
+  checkoutDeliveryMethodUpdate(checkoutId: ID, collectionPointId: ID, shippingMethodId: ID, token: UUID): CheckoutDeliveryMethodUpdate
```
maarcingebala, line 2902: `checkoutId: ID` is deprecated in existing mutations, in new checkout mutations we should use `token: UUID` instead.

saleor/order/models.py
```diff
@@ -130,9 +130,17 @@ class Order(ModelWithMetadata):
         related_name="orders",
         on_delete=models.SET_NULL,
     )
+    collection_point = models.ForeignKey(
+        "warehouse.Warehouse",
+        blank=True,
+        null=True,
+        related_name="orders",
+        on_delete=models.SET_NULL,
```
maarcingebala, line 138: I don't recall if we discussed it - do we somehow handle the case when a warehouse is deleted from the system, but it was selected as a collection point in some existing orders?

saleor/graphql/checkout/mutations.py
```diff
@@ -1027,6 +1068,157 @@ def perform_mutation(
...
+                        code=CheckoutErrorCode.SHIPPING_NOT_REQUIRED,
+                    )
+                }
+            )
+        if shipping_method_id is not None and collection_point_id is not None:
+            raise ValidationError(
+                {
+                    "delivery_method": ValidationError(
+                        "Only one delivery method should be picked - C&C or Shipping",
+                        code=CheckoutErrorCode.SHIPPING_METHOD_NOT_APPLICABLE,
```
maarcingebala, line 1207: This error code is related to a specific error when a selected shipping method cannot be used due to some business logic rules that are not satisfied. Here the error is caused just by invalid input passed to the mutation, maybe `INVALID` would be a better code?

saleor/warehouse/models.py
```diff
@@ -25,8 +30,66 @@ def for_country(self, country: str):
...
+            .annotate(line_quantity=F("available_quantity") - Subquery(lines_quantity))
+            .order_by("line_quantity")
+            .filter(
+                product_variant__id__in=lines.values("variant_id"),
+                line_quantity__gte=0,
+            )
+            .select_related("product_variant")
+        )
+
+        warehouses_qs = (
```
maarcingebala, line 76: This query is identical to the query in `applicable_for_click_and_collect_no_quantity_check`. Maybe extract it to a function? Could be a "private" function, only used locally in this queryset class.

saleor/order/models.py
kuchichan: No, we haven't discuss it, this could be a tough one.  Do you think that,  for instance `on_delete=models.PROTECT` will be better?

saleor/graphql/schema.graphql
kuchichan: Yes, probably even the most restrictive `[Warehouse!]!` could be used here, I would prefer an empty list when there is no collection points available.

saleor/checkout/fetch.py
```diff
@@ -50,6 +66,157 @@ def get_customer_email(self) -> str:
...
+            valid_delivery_methods and self.delivery_method in valid_delivery_methods
+        )
+
+
+@singledispatch
+def build_delivery_method(
+    delivery_method: Optional[Union["ShippingMethod", "Warehouse"]],
+    address=Optional["Address"],
+) -> EmptyDeliveryMethod:
+    raise NotImplementedError("Incompatible Type")
```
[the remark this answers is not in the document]
kuchichan: Yes, technically it would be the same, I just want to avoid `if else` ladder, and probably, the benefits would be more visible when creating these objects would be more complicated, than just simple `__init__`

saleor/checkout/complete_checkout.py
```diff
@@ -90,13 +95,16 @@ def _process_shipping_data_for_order(
...
             shipping_address = shipping_address.get_copy()
 
-    return {
+    result: Dict[str, Any] = {
         "shipping_address": shipping_address,
-        "shipping_method": checkout_info.shipping_method,
-        "shipping_method_name": smart_text(checkout_info.shipping_method),
+        "shipping_method_name": smart_text(
+            checkout_info.shipping_method
+        ),  # TODO: collection_point_name?
```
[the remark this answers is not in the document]
kuchichan: Done. Additional column was add to `order` as well as related properties in `DeliveryMethodInfo`

saleor/checkout/fetch.py
```diff
@@ -1,7 +1,16 @@
+import itertools
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Iterable, List, Optional
+from functools import singledispatch
+from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Union
 
-from ..shipping.models import ShippingMethodChannelListing
+from prices import TaxedMoney
+
+from saleor.warehouse import WarehouseClickAndCollectOption
```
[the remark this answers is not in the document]
kuchichan: Done.

saleor/checkout/fetch.py
```diff
@@ -50,6 +66,157 @@ def get_customer_email(self) -> str:
...
+@dataclass(frozen=True)
+class ShippingMethodInfo(EmptyDeliveryMethod):
+    delivery_method: "ShippingMethod"
+    shipping_address: Optional["Address"]
+    order_key: str = "shipping_method"
+
+    def calculate_checkout_shipping(self, checkout_info, lines=None) -> TaxedMoney:
+        """Return checkout shipping price."""
+        # FIXME: Optimize checkout.is_shipping_required
+        # Copied from base_calculation due to circular imports
```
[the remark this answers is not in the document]
kuchichan: Done.

saleor/graphql/order/types.py
```diff
@@ -8,6 +8,8 @@
 from graphene import relay
 from promise import Promise
 
+from saleor.graphql.checkout.types import DeliveryMethod
```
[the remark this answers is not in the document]
kuchichan: Done.

saleor/graphql/order/types.py
kuchichan: Done.

saleor/graphql/warehouse/dataloaders.py
kuchichan: I have removed this comment, however something sketchy is going on here. Probably there is a `UUID` type discrepancy in `load` and `batch_load`. `warehouse_uuid` is a UUID object itself,  but something like:
``` python
```

saleor/graphql/warehouse/filters.py
kuchichan: Done.

saleor/webhook/tests/test_webhook_payloads.py
kuchichan: Done.

saleor/checkout/fetch.py
```diff
@@ -50,6 +66,157 @@ def get_customer_email(self) -> str:
...
+
+        return quantize_price(
+            TaxedMoney(net=shipping_price, gross=shipping_price),
+            shipping_price.currency,
+        )
+
+    def is_valid_delivery_method(self) -> bool:
+        if not self.shipping_address:
+            return False
+        return True
```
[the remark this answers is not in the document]
kuchichan: 🤦  This method was modified few times, done.

saleor/checkout/fetch.py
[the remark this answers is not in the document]
kuchichan: I have changed to proposed version.

saleor/graphql/checkout/types.py
kuchichan: Done.

saleor/graphql/checkout/types.py
kuchichan: Done.

saleor/graphql/order/types.py
kuchichan: Done.

saleor/graphql/checkout/types.py
kuchichan: Done.

saleor/graphql/schema.graphql
kuchichan: Changed to non-nullable.

saleor/graphql/schema.graphql
kuchichan: Changed to non-nullable.

saleor/graphql/schema.graphql
kuchichan: Changed to non-nullable.

saleor/graphql/schema.graphql
kuchichan: Done.

saleor/graphql/checkout/mutations.py
kuchichan: Done.

saleor/warehouse/models.py
kuchichan: Done.

saleor/core/utils/random_data.py
```diff
@@ -22,6 +22,8 @@
 from measurement.measures import Weight
 from prices import Money, TaxedMoney
 
+from saleor.warehouse import WarehouseClickAndCollectOption
```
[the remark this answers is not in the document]
kuchichan: Done.

saleor/graphql/checkout/dataloaders.py
```diff
@@ -3,7 +3,9 @@
 from django.db.models import F
 from promise import Promise
 
-from ...checkout.fetch import CheckoutInfo, CheckoutLineInfo
+from saleor.graphql.warehouse.dataloaders import WarehouseByIdLoader
```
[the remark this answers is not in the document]
kuchichan: Done.

saleor/graphql/checkout/mutations.py
```diff
@@ -1,19 +1,22 @@
...
 import uuid
-from typing import TYPE_CHECKING, Iterable, List, Optional, Tuple
+from typing import TYPE_CHECKING, Iterable, List, Optional, Tuple, Union
 
 import graphene
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.db.models import Q
 
+from saleor.graphql.warehouse.types import Warehouse
```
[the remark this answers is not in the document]
kuchichan: Done.

saleor/graphql/checkout/mutations.py
```diff
@@ -1027,6 +1068,157 @@ def perform_mutation(
...
+    checkout = graphene.Field(Checkout, description="An updated checkout")
+
+    class Arguments:
+        checkout_id = graphene.ID(
+            description=(
+                "Checkout ID. "
+                "DEPRECATED: Will be removed in Saleor 4.0. Use token instead."
+            ),
+            required=False,
+        )
```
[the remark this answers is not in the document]
kuchichan: Done.

saleor/graphql/checkout/mutations.py
```diff
@@ -1027,6 +1068,157 @@ def perform_mutation(
...
+            CheckoutErrorCode, "checkout_id", checkout_id, "token", token
+        )
+
+        if token:
+            checkout = get_checkout_by_token(token)
+        # DEPRECATED
+        else:
+            checkout = cls.get_node_or_error(
+                info, checkout_id or token, only_type=Checkout, field="checkout_id"
+            )
```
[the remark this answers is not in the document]
kuchichan: Done.

saleor/graphql/checkout/schema.py
```diff
@@ -69,5 +70,8 @@ class CheckoutMutations(graphene.ObjectType):
     checkout_remove_promo_code = CheckoutRemovePromoCode.Field()
     checkout_payment_create = CheckoutPaymentCreate.Field()
     checkout_shipping_address_update = CheckoutShippingAddressUpdate.Field()
-    checkout_shipping_method_update = CheckoutShippingMethodUpdate.Field()
+    checkout_shipping_method_update = CheckoutShippingMethodUpdate.Field(
+        deprecation_reason="Use checkoutDeliveryMethodUpdate instead"
```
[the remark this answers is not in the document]
kuchichan: Done.

saleor/graphql/checkout/tests/test_checkout_lines.py
```diff
@@ -3,6 +3,9 @@
 from unittest import mock
 
 import graphene
+import pytest
+
+from saleor.warehouse import WarehouseClickAndCollectOption
```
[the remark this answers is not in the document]
kuchichan: Done.

saleor/graphql/checkout/types.py
```diff
@@ -445,6 +492,32 @@ def apply_price_to_shipping_method(channel_listings):
...
+        def get_available_collection_points(data):
+            address, lines, checkout_info = data
+            available = get_valid_collection_points_for_checkout(
+                lines, checkout_info, country_code=address.country.code
+            )
+            if not available:
+                return []
+
+            available_ids = available.values_list("id", flat=True)
+            return WarehouseByIdLoader(info.context).load_many(available_ids)
```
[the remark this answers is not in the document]
kuchichan: Done, it helped to simplify code in 2 other places.

saleor/checkout/fetch.py
```diff
@@ -36,10 +44,18 @@ class CheckoutInfo:
     channel: "Channel"
     billing_address: Optional["Address"]
     shipping_address: Optional["Address"]
-    shipping_method: Optional["ShippingMethod"]
+    shipping_method: Optional["ShippingMethod"]  # Will be deprecated
```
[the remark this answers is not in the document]
kuchichan: Done, also few functions related to `checkout_info.shipping_method` was removed / replaced by generic delivery_method

saleor/checkout/utils.py
```diff
@@ -612,11 +636,19 @@ def is_valid_shipping_method(checkout_info: "CheckoutInfo"):
 
     valid_methods = checkout_info.valid_shipping_methods
     if valid_methods is None or checkout_info.shipping_method not in valid_methods:
-        clear_shipping_method(checkout_info)
+        clear_shipping_method(checkout_info)  # TODO: Derides command query separation
```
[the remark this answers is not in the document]
kuchichan: Done,  this function  (`clear_shipping_method`) does not exist anymore

saleor/order/models.py
kuchichan: Based on our short conversation, I have add an `collection_point_name` character field, similar to `shipping_method_name`

saleor/core/utils/random_data.py
```diff
@@ -1201,16 +1202,49 @@ def create_shipping_zones():
     )
 
 
+def create_click_and_collect_related_fields() -> List[Tuple[bool, str]]:
```
korycins, line 1205: I am not sure of this function. It is super hard to understand what it is doing. 
It looks like it generates some list of tuple which you use randomly. It is used in one place so no need to return a list here. 
why not something like this:
```
def foo() -> Tuple[bool, str]:
    is_private = random.choice([True, False])
    c_and_c_option = random.choice(["WarehouseClickAndCollectOption with excluding local stock"]
    return is_private, c_and_c_option

...
is_private, cc_option = foo()

OR better
is_private = random.choice([True, False])
c_and_c_option = random.choice([Warehouse exlucding local stock])
```

saleor/graphql/checkout/dataloaders.py
```diff
@@ -187,13 +188,22 @@ def with_channel(channels):
                         self.context
                     ).load_many(shipping_method_ids_channel_slugs)
                 )
+                collection_point_ids = [
+                    checkout.collection_point_id
+                    for checkout in checkouts
+                    if checkout.collection_point
```
korycins, line 194: `if checkout.collection_point_id`

saleor/graphql/checkout/mutations.py
```diff
@@ -1032,6 +1046,139 @@ def perform_mutation(
...
+            update_fields=["shipping_method", "collection_point", "last_change"]
+        )
+        recalculate_checkout_discount(
+            manager, checkout_info, lines, info.context.discounts
+        )
+        manager.checkout_updated(checkout)
+        return CheckoutDeliveryMethodUpdate(checkout=checkout)
+
+    @classmethod
+    def perform_on_collection_point(
```
korycins, line 1106: `perform_on_shipping_method` and `perform_on_collection_point` are very similar. I don't see a lot of differences - Different params for `cls.get_node_or_error`, different error msg and different assigment to checkout. IMO we could try to extract common part of this methods as most of the logic is duplicated.

saleor/graphql/checkout/mutations.py
```diff
@@ -1032,6 +1046,139 @@ def perform_mutation(
...
+        if shipping_method_id is not None and collection_point_id is not None:
+            raise ValidationError(
+                {
+                    "delivery_method": ValidationError(
+                        "Only one delivery method should be picked - C&C or Shipping",
+                        code=CheckoutErrorCode.INVALID,
+                    )
+                }
+            )
+        return (
```
korycins, line 1171: For better readability, can we just call it inside normal `if` condition ?  It won't add a lot of additional lines but it will be easier to read.

saleor/checkout/fetch.py
```diff
@@ -50,6 +65,157 @@ def get_customer_email(self) -> str:
...
+class ShippingMethodInfo(EmptyDeliveryMethod):
+    delivery_method: "ShippingMethod"
+    shipping_address: Optional["Address"]
+    order_key: str = "shipping_method"
+
+    @property
+    def delivery_method_name(self) -> Dict[str, Optional[str]]:
+        return {"shipping_method_name": smart_text(self.delivery_method)}
+
+    def calculate_checkout_shipping(self, checkout_info, lines=None) -> TaxedMoney:
```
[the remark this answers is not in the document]
kuchichan: please note the circular dependency:  
`if lines is not None and all(isinstance(line, CheckoutLineInfo) for line in lines):`
We can't even import this method to `fetch.py`, This is the major reason why I have decided to put that code here. What is more the method of shipping calculation depends on shipping method, that is why i have decided to leverage polymorphism here. I have tried even different approach - injecting proper calculation method in DeliveryMethod builder, however due to two-way dependency this was really a daunting task :(

saleor/checkout/fetch.py
```diff
@@ -50,6 +65,157 @@ def get_customer_email(self) -> str:
         return self.user.email if self.user else self.checkout.email
 
 
+@dataclass(frozen=True)
+class EmptyDeliveryMethod:
```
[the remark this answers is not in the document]
kuchichan: Ofc we can but please note that is a kind of Null type - it helps me get rid of code like: `if something is not None...`. This class is instantiated a lot, that is why I did not use `Base` suffix (It resembles not-so-concrete classes, which hasn't their own instances)

saleor/graphql/checkout/mutations.py
kuchichan: 👍  Done.

saleor/graphql/checkout/mutations.py
kuchichan: Done, I have extracted 2 methods.

saleor/graphql/checkout/dataloaders.py
kuchichan: Done.

saleor/core/utils/random_data.py
kuchichan: Yes, I had have over engineered it, since I want a full collection before entering the loop ;d I have use your second proposal, done.

saleor/checkout/fetch.py
```diff
@@ -50,6 +65,157 @@ def get_customer_email(self) -> str:
...
+
+
+@singledispatch
+def get_delivery_method_info(
+    delivery_method: Optional[Union["ShippingMethod", "Warehouse"]],
+    address=Optional["Address"],
+) -> EmptyDeliveryMethod:
+    if delivery_method is None:
+        return EmptyDeliveryMethod()
+    elif isinstance(delivery_method, ShippingMethod):
```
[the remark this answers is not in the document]
kuchichan: 👍  Done.

saleor/checkout/fetch.py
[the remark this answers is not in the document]
kuchichan: Anyway, Done

saleor/graphql/order/types.py
kuchichan: I hope it's done now, my apologise ;( 🤦

saleor/checkout/utils.py
```diff
@@ -603,25 +604,34 @@ def get_valid_shipping_methods_for_checkout(
...
 
-def is_valid_shipping_method(checkout_info: "CheckoutInfo"):
-    """Check if shipping method is valid and remove (if not)."""
-    if not checkout_info.shipping_method:
-        return False
+def get_valid_collection_points_for_checkout(
+    lines: Iterable["CheckoutLineInfo"],
+    checkout_info: "CheckoutInfo",
+    country_code: Optional[str] = None,
+    quantity_check: bool = True,
```
maarcingebala, line 611: Could we add a docstring for this function? `quantity_check` requires explanation as it's not obvious why or when you should use it.

saleor/warehouse/models.py
```diff
@@ -25,8 +30,60 @@ def for_country(self, country: str):
             .order_by("pk")
         )
 
+    def applicable_for_click_and_collect_no_quantity_check(
```
maarcingebala, line 33: This function and the one below could also have some docstrings that would explain their assumptions or business logic used to resolve Warehouses. Ideally, there should be one function to do that, but we have two and it's not clear which should be used in which case.

saleor/checkout/utils.py
kuchichan: Docstrings has been added.

saleor/warehouse/models.py
kuchichan: Done.

saleor/checkout/fetch.py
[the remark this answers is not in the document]
kuchichan: Because, how we calculate shipping price depends on delivery method. Should I write another `calculate_shipping_method_for_c&c` or perform `if / else` inside base_checkout shipping price? If another `deliveryMethod` comes in, will we continue `if/else` scenario?  (Also, this method was hardcoded in AppManager)

saleor/checkout/utils.py
```diff
@@ -603,25 +604,41 @@ def get_valid_shipping_methods_for_checkout(
...
+
+    if not is_shipping_required(lines):
+        return []
+    if not checkout_info.shipping_address:
+        return []
+    line_ids = [line_info.line.id for line_info in lines]
+    lines = CheckoutLine.objects.filter(id__in=line_ids)
+
+    return (
+        Warehouse.objects.applicable_for_click_and_collect(lines, country_code)
```
korycins, line 628: Can't we have a single method `applicable_for_click_and_collect` which will accept optional flag which will determine if it should check stock quantity or not?

saleor/graphql/checkout/mutations.py
```diff
@@ -1035,6 +1053,178 @@ def perform_mutation(
         return CheckoutShippingMethodUpdate(checkout=checkout)
 
 
+class CheckoutDeliveryMethodUpdate(BaseMutation):
+    checkout = graphene.Field(Checkout, description="An updated checkout")
```
korycins, line 1057: `description="An updated checkout."` missing dot.

saleor/graphql/checkout/mutations.py
```diff
@@ -1035,6 +1053,178 @@ def perform_mutation(
...
+            info,
+            shipping_method_id,
+            only_type=ShippingMethod,
+            field="shipping_method_id",
+            qs=shipping_models.ShippingMethod.objects.prefetch_related(
+                "postal_code_rules"
+            ),
+        )
+
+        CheckoutDeliveryMethodUpdate._check_delivery_method(
```
korycins, line 1088: Why not `cls`? `cls._check_delivery_method`

saleor/graphql/checkout/mutations.py
```diff
@@ -1035,6 +1053,178 @@ def perform_mutation(
...
+            checkout_info, lines, shipping_method=shipping_method, collection_point=None
+        )
+
+        checkout_delivery_method_update = (
+            CheckoutDeliveryMethodUpdate._update_delivery_method(
```
korycins, line 1093: Why not `cls`?

saleor/graphql/checkout/mutations.py
```diff
@@ -1035,6 +1053,178 @@ def perform_mutation(
...
+        collection_point: Optional[Warehouse]
+    ) -> "CheckoutDeliveryMethodUpdate":
+        checkout.shipping_method = shipping_method
+        checkout.collection_point = collection_point
+        checkout.save(
+            update_fields=["shipping_method", "collection_point", "last_change"]
+        )
+        manager.checkout_updated(checkout)
+
+        return CheckoutDeliveryMethodUpdate(checkout=checkout)
```
korycins, line 1169: IMHO, just for readability and to keep up the common way of handling this. This method should not return anything or return checkout. 
`perform_mutation` should wrap `checkout` into `CheckoutDeliveryMethodUpdate`.
Now, the method does two jobs - update a delivery method details and prepare a response for perform_mutation.

saleor/graphql/checkout/mutations.py
```diff
@@ -1035,6 +1053,178 @@ def perform_mutation(
...
+
+    @staticmethod
+    def _resolve_delivery_method_id(id_) -> Optional[str]:
```
korycins, line 1172: Why does delviery_method_id return its type?

saleor/graphql/checkout/mutations.py
```diff
@@ -1035,6 +1053,178 @@ def perform_mutation(
...
+
+        delivery_method_is_valid = clean_delivery_method(
+            checkout_info=checkout_info, lines=lines, method=delivery_method
+        )
+        if not delivery_method_is_valid:
+            raise ValidationError(
+                {
+                    "delivery_method": ValidationError(
+                        error_msg,
+                        code=CheckoutErrorCode.SHIPPING_METHOD_NOT_APPLICABLE.value,
```
korycins, line 1149: Is it ok, that for both (shipping_method and c_andC) we have the same error code? The name of this suggest that error code is related to the shipping method.

saleor/graphql/checkout/types.py
```diff
@@ -266,6 +299,18 @@ def wrap_shipping_method_with_channel_context(data):
             wrap_shipping_method_with_channel_context
         )
 
+    @staticmethod
+    @traced_resolver
+    def resolve_delivery_method(root: models.Checkout, info):
+        if root.shipping_method_id:
+            return Checkout.resolve_shipping_method(root, info)
+        if root.collection_point_id:
+            collection_point = WarehouseByIdLoader(info.context).load(
```
korycins, line 308:
```
return WarehouseByIdLoader(info.context).load(
...
```

saleor/graphql/order/types.py
```diff
@@ -1031,6 +1046,17 @@ def wrap_shipping_method_with_channel_context(data):
 
     @staticmethod
     @traced_resolver
+    def resolve_delivery_method(root: models.Order, info):
+        if root.shipping_method_id:
+            return Order.resolve_shipping_method(root, info)
```
korycins, line 1051: I believe, you can add `@classmethod` instead of `staticmethod` and call `cls` here.

saleor/graphql/order/types.py
```diff
@@ -1031,6 +1046,17 @@ def wrap_shipping_method_with_channel_context(data):
...
+        if root.collection_point_id:
+            collection_point = WarehouseByIdLoader(info.context).load(
+                root.collection_point_id
+            )
+            return collection_point
+        return None
+
+    @staticmethod
     # TODO: We should optimize it in/after PR#5819
     def resolve_available_shipping_methods(root: models.Order, info):
```
korycins, line 1061: This method had wrapper `traced_resolver` but it is missing now. Is it expected?

saleor/graphql/order/types.py
```diff
@@ -1064,6 +1090,17 @@ def resolve_available_shipping_methods(root: models.Order, info):
...
         return instances
 
+    @staticmethod
+    def resolve_available_collection_points(root: models.Order, info):
+        def get_available_collection_points(data):
+            lines, address = data
+
+            return get_valid_collection_points_for_order(lines, address)
+
+        lines = Order.resolve_lines(root, info)
```
korycins, line 1100: Why not `@classmethod` and `cls`?

saleor/graphql/schema.graphql
```diff
@@ -5954,9 +5968,17 @@ type Warehouse implements Node & ObjectWithMetadata {
   shippingZones(before: String, after: String, first: Int, last: Int): ShippingZoneCountableConnection!
   address: Address!
   email: String!
+  isPrivate: Boolean!
```
korycins, line 5971: What does exactly describe this flag?

saleor/graphql/common/dataloaders.py
```diff
@@ -0,0 +1,50 @@
+from django.conf import settings
+
+from ...warehouse.models import Warehouse
+from ..account.dataloaders import AddressByIdLoader
+from ..channel.dataloaders import ChannelBySlugLoader
+from ..core.dataloaders import DataLoader
+
+
+class WarehouseCountryCodeByChannelLoader(DataLoader):
```
korycins, line 9: Why did we extract this Dataloader to a separate module?  It is directly related to Warehouse. 
cc @fowczarek

saleor/checkout/utils.py
kuchichan: I guess these queries are complex enough, and they not need another if/else clause.

korycins: I've spent more time finding the difference between them than understanding what they're doing

saleor/graphql/checkout/mutations.py
kuchichan: This is not the sake of readability. It's the CQS rule. I would agree with you, but why its also in `ShippingMethodUpdate` mutation and it's ok there? I Just want to reflect ShippingMethodUpdate 1:1, so i will leave it as it is.

saleor/graphql/checkout/mutations.py
kuchichan: Because we are checking to which type `Warehouse` or `Shipping` method id belongs.  It returns type name - "Warehouse" or "ShippingMethod".

saleor/graphql/common/dataloaders.py
kuchichan: Because of the circular dependency. I have used `WarehouseIdLoader` in Checkout. But now i cannot do that because Warehouse has dependency to -> Channel and Channel to Checkout.

saleor/graphql/schema.graphql
kuchichan: `As a User I want to set visibility of warehouse stock to Private so that Customer's won't be able to see the stock at specific warehouse`

saleor/checkout/utils.py
kuchichan: 🤷‍♂️

saleor/graphql/checkout/mutations.py
kuchichan: Done.

saleor/graphql/checkout/mutations.py
kuchichan: Done/

saleor/graphql/checkout/mutations.py
kuchichan: Done.

saleor/graphql/checkout/types.py
kuchichan: Done.

saleor/graphql/order/types.py
kuchichan: Yep, done.

saleor/graphql/checkout/mutations.py
kuchichan: I have an add additional Error code, done.

saleor/graphql/order/types.py
kuchichan: Good point. Done.

saleor/graphql/order/types.py
kuchichan: Done.

saleor/graphql/checkout/mutations.py
korycins: Not sure if I follow. I don't see `ShippingMethodUpdate`. Do you mean `CheckoutShippingMethodUpdate`? If yes, it creates response object in `perfom_mutation`. Maybe I am missing something.

saleor/graphql/checkout/mutations.py
korycins: So shouldn't name be `resolve_delivery_method_type` instead of `resolve_delivery_method_id`?

saleor/graphql/common/dataloaders.py
korycins: Wouldn't import inside a function/method be better idea?

saleor/checkout/fetch.py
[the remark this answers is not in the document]
korycins: base_calculation should have enough details to calculate prices. You can wrap logic to calculate shipping_price for shipping method in helper function and call it in `if/else` in `base_calculations.shipping_price` (like you said). `ShippingMethodInfo` should only "tell" a base_calculation that the given delivery method is a shipping_method then base_caluclation calculates prices in a proper way.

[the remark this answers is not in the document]
kuchichan: This is exactly the thing I want to avoid, but ok, I will do that.

saleor/graphql/common/dataloaders.py
kuchichan: This is considered as a bad practice AFAIK. Of course it will hide problem of huge co-dependency without visible hierarchy, but I will not do that, unless I have to.  This is exactly the same pattern as with `base_calculations.py` but there are bare functions that: acting pure functions (they are not), acting they are not holding dependency (they are). But they are in one place (what is the benefit, IDK). Later on, we have circular dependency and the only solution is to drag parameters across function calls (`if / else ` `isinstance` pattern) - to the next code explosion.

saleor/graphql/checkout/mutations.py
kuchichan: Probably you're right.

saleor/graphql/common/dataloaders.py
maarcingebala: Code is easier to maintain in the long run, when related things are in one place - that's the benefit.

saleor/graphql/checkout/mutations.py
```diff
@@ -1035,6 +1053,176 @@ def perform_mutation(
...
+
+    @classmethod
+    def perform_on_collection_point(
+        cls, info, collection_point_id, checkout_info, lines, checkout, manager
+    ):
+        collection_point = cls.get_node_or_error(
+            info,
+            collection_point_id,
+            only_type=Warehouse,
+            field="collection_point_id",
```
fowczarek, line 1111: We don't have this field in input. In `field` we should use only names from mutation input. This field is used to return errors in case of an invalid id.

fowczarek:
```suggestion
            field="delivery_method_id",
```

saleor/graphql/checkout/mutations.py
```diff
@@ -1035,6 +1053,176 @@ def perform_mutation(
...
+
+    @classmethod
+    def perform_on_shipping_method(
+        cls, info, shipping_method_id, checkout_info, lines, checkout, manager
+    ):
+        shipping_method = cls.get_node_or_error(
+            info,
+            shipping_method_id,
+            only_type=ShippingMethod,
+            field="shipping_method_id",
```
fowczarek, line 1082: We don't have this field in input. In `field` we should use only names from mutation input. This field is used to return errors in case of an invalid id.
```suggestion
            field="delivery_method_id",
```

saleor/graphql/checkout/mutations.py
```diff
@@ -1035,6 +1053,176 @@ def perform_mutation(
...
+            delivery_method = collection_point
+            error_msg = "This pick up point is not applicable."
+
+        delivery_method_is_valid = clean_delivery_method(
+            checkout_info=checkout_info, lines=lines, method=delivery_method
+        )
+        if not delivery_method_is_valid:
+            raise ValidationError(
+                {
+                    "delivery_method": ValidationError(
```
fowczarek, line 1145:
```suggestion
                    "delivery_method_id": ValidationError(
```

saleor/graphql/checkout/mutations.py
```diff
@@ -1035,6 +1053,176 @@ def perform_mutation(
...
+            return None
+
+        possible_types = ("Warehouse", "ShippingMethod")
+        type_, id_ = from_global_id_or_error(id_)
+        str_type = str(type_)
+
+        if str_type not in possible_types:
+            raise ValidationError(
+                {
+                    "delivery_method": ValidationError(
```
fowczarek, line 1181:
```suggestion
                    "delivery_method_id": ValidationError(
```

saleor/graphql/common/dataloaders.py
fowczarek: IMO we should use import inside function in this case because we have such import in a few other dataloaders. If the problem will grow, we should apply a different solution for all places where the problem occurs.

saleor/warehouse/models.py
```diff
@@ -57,6 +134,21 @@ def delete(self, *args, **kwargs):
...
         address.delete()
 
+    def clean(self) -> None:
+        if (
+            self.click_and_collect_option == WarehouseClickAndCollectOption.LOCAL_STOCK
+            and self.is_private
+        ):
+            raise ValidationError(
+                {
+                    "click_and_collect_option": (
```
fowczarek, line 144: We shouldn't build graphql error structure in `models.py`, Can we rise simple ValidationError here and catch it in mutations and wrap them in necessary information(filed).

The test for this error is also missing.

saleor/checkout/fetch.py
```diff
@@ -50,6 +65,157 @@ def get_customer_email(self) -> str:
...
+            == WarehouseClickAndCollectOption.LOCAL_STOCK
+        )
+
+    @property
+    def delivery_method_name(self) -> Dict[str, Optional[str]]:
+        return {"collection_point_name": smart_text(self.delivery_method)}
+
+    def get_warehouse_filter_lookup(self) -> Dict[str, Any]:
+        return (
+            {"warehouse__pk": self.delivery_method.pk}
```
fowczarek, line 181: Maybe?
This should allow us to remove additional JOIN from SQL queries.
```suggestion
            {"warehouse_id": self.delivery_method.pk}
```

saleor/graphql/checkout/mutations.py
kuchichan: Sorry for misunderstanding, done.

saleor/graphql/checkout/mutations.py
kuchichan: Done.

saleor/graphql/common/dataloaders.py
kuchichan: Done, I have reverted my changes and I have used import inside `batch_load`

saleor/graphql/checkout/mutations.py
kuchichan: Done.

saleor/graphql/checkout/mutations.py
kuchichan: Done.

saleor/graphql/checkout/mutations.py
kuchichan: Done.

saleor/graphql/checkout/mutations.py
kuchichan: Done.

saleor/warehouse/models.py
kuchichan: Based on different models and common practice in exising codebase,  I have decided to drop `clean` method and I have prepared appropriate logic inside `WarehouseMixin`

saleor/checkout/fetch.py
kuchichan: 👍  Done.

saleor/checkout/fetch.py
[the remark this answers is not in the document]
kuchichan: Done.

saleor/graphql/warehouse/dataloaders.py
kuchichan: Probably, I have found the cause.
In `order/types.py` we have:
``` python
    @staticmethod
    def resolve_warehouse(root: models.OrderEvent, info):
        if warehouse_pk := root.parameters.get("warehouse"):
            return WarehouseByIdLoader(info.context).load(warehouse_pk)
        return None
```
`parameters` is `JSONField`, thus, `warehouse_pk` will be `str`, on the other side when we pass warehouse_id (which is `UUIDField` and it is automatically casted to `UUID`). I can leave it as it is or good idea will be:
``` python
    @staticmethod
    def resolve_warehouse(root: models.OrderEvent, info):
        if warehouse_pk := root.parameters.get("warehouse"):
            return WarehouseByIdLoader(info.context).load(UUID(warehouse_pk))
        return None
```
and then we can write:
`return [warehouses.get(warehouse_uuid) for warehouse_uuid in keys]` where `keys: Iterable[UUID]`

What do you think?

korycins (approved):

maarcingebala (approved):

maarcingebala:
@kuchichan Please add a changelog entry in the 3.1 section.

saleor/graphql/checkout/mutations.py
```diff
@@ -1029,6 +1045,175 @@ def perform_mutation(
...
+            required=False,
+        )
+
+    class Meta:
+        description = (
+            "Updates the delivery method (shipping method or pick up point)"
+            "of the checkout."
+        )
+        error_type_class = CheckoutError
+        error_type_field = "checkout_error"
```
fowczarek, line 1064: We could `error_type_field` remove it. This field is responsible for adding deprecated errors to the mutaion. We could add new mutations without old errors.

fowczarek (approved):

saleor/graphql/checkout/mutations.py
kuchichan: Done.

saleor/graphql/warehouse/dataloaders.py
kuchichan: Done.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All notable, unreleased changes to this project will be documented in this file.
 
 # 3.1.0 [Unreleased]
 
+- Add Click&Collect feature - #SALEOR-2992 by @kuchichan
```
maarcingebala, line 7: In the public changelog, we want to refer to PRs that introduced the described changes.
```suggestion
- Add Click&Collect feature - #7673 by @kuchichan
```

kuchichan: 🥇  Done.

PR #7682, opened 2021-07-24
POC - validate webhook schema

korycins:
It is a proposition of usage a `pydantic` and `datamodel-code-generator` to prepare webhook payload schema.

- `saleor/webhook/schema.py ` contains payload models which define the structure of sending payload (for PoC, I added only Customer events). It is built with `pydantic`'s BaseModel which will help us to confirm the output payload that we're sending via webhooks.
- `test_generate_customer_payload` - checks if structure of payload is in expected schema format 
- `check_webhook_payload` - confirm that we are sending a payload in a proper format.

- benefits:
  - we are able to easily track changes in webhook payloads
  - one-line command to generate `jsonschema` file with full webhook description.
  - we can build new section in docs - webhook schema and automatically update them by using generated schema
  - tracking and updating Saleor Apps will be much more easier as we will treat webhook.schema as a contract.
  - I believe that generated webhook schema can be used to map jsonschema into  types/classes in any language. (for python we can use `datamodel-code-generator` to convert `jsonschema` file into a pydantic models.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


korycins:
@maarcingebala, @patrys, @krzysztofwolski  What do you think? ☝️

krzysztofwolski:
1. In this approach we are validating output - WDYT about using those datamodels to construct payloads instead of dict stitching we have now? IMO DX would go up and we would be sure that constructed data are valid with the model, without relying on test suite to catch every edge case.
2. I'm looking for documentation generators. Could you send me generated JSON Schemas? (https://pydantic-docs.helpmanual.io/usage/schema/).

pkucmus:
@krzysztofwolski last time I checked there was not much when it comes to documentation generation. But Redoc is a good one - it's important to use one that understands everything about the OAS3 spec (i.e. discriminator fields is something that is rarely supported)

Are you going for json schema or OAS3?
Having a schema / contract is a massive improvement. [FastAPI's OAS3 generator](https://github.com/tiangolo/fastapi/blob/master/fastapi/openapi/utils.py#L356) could be used here to generate the schema.

I would not validate the output but rather use those models to render the payload or go a step further and do what @krzysztofwolski is describing.

korycins:
@pkucmus @krzysztofwolski  I am going to use models to generate a payload but not in the first stage. The current approach will raise errors when the structure of the payload will be different than expected but Saleor will continue sending a webhook. IMHO it is safer in our case as there is a possibility that we could forget or miss some field during the implementation of models. I would rather make sure that our models fully cover our current payload and then change the logic for generating payload than receiving errors that we don't send a webhook because payload vs models are different.

PR #7720, opened 2021-07-28
First pass on translations webhook

rafalp:
This PR adds `TRANSLATION_CREATED` and `TRANSLATION_UPDATED` events and connects those events to web hooks, enabling integrations to sync translations on site.

This is draft for now because we would like translations to be more generic in the future, so this PR is more of RFC now that final work.

Approach taken is introduction of `Translation` base model for all translation models to extend. This model adds `language_code` field to models and two methods:

`get_translated_object`: getter for instance of translated object, used to build `id` that goes into webhook's payload. Currently this is reversed to `ID` via `generate_translation_payload` using following logic:

```python
translation_object = translation.get_translated_object()
graphene.Node.to_global_id(
    translation_object.meta_.object_name, translation_object.id
)
```

But this will run DB query if relation is not available. Perhaps its better to return tuple of name and id or change `get_translated_object` to `get_translated_object_id` and return final string? But that will add more boilerplate to our models we may not want.

`get_translated_keys`: getter for dict of translation key/value pairs. Delegates control on what translations to return in webhook to the model, but perhaps we would like to have list of translation keys instead and do `get_attr` in  `generate_translation_payload` for less boilerplate? Eg. `translated_keys = ["name", "description"]`. But this approach will require extra care for translation models extending `SeoModelTranslation` that includes its own translation fields.

Example payload with product translation:

```json
{
  "id": "UHJvZHVjdDoxMTU=",
  "language_code": "pl",
  "keys": [
    {"key": "seo_title", "value": null},
    {"key": "seo_description", "value": null}, {"key": "name", "value": "Bluza z kapturem czarna"}, {"key": "description", "value": {"time": 1627493627745, "blocks": [{"type": "paragraph", "data": {"text": "Oferta specjalna. Kup bluz\u0119 i otrzymaj za darmo czarny sweter. Woho! Lorem ipsum 125"}}], "version": "2.20.0"}}
  ]
}
```

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/translations/mutations.py
```diff
@@ -106,10 +106,16 @@ def perform_mutation(cls, _root, info, **data):
...
-        product.translations.update_or_create(
+        translation, created = product.translations.update_or_create(
             language_code=data["language_code"], defaults=data["input"]
         )
         product = ChannelContext(node=product, channel_slug=None)
+
+        if created:
+            info.context.plugins.translation_created(translation)
+        else:
+            info.context.plugins.translation_updated(translation)
```
maarcingebala, line 117: Maybe a bit unrelated to this PR, but this mutation should probably use transaction and these plugin calls should be made in transaction.on_commit. I don't know why it's not used already, as it is done in `ProductVariantTranslate`; maybe it's a good occasion to unify it.

maarcingebala:
@rafalp I think the changes look good and we can go in this direction. Regarding `But this will run DB query if the relation is not available.` - maybe we can add a comment in `get_translated_object` that would warn about this, and in the codebase assure that the related object is actually always prefetched or loaded?

rafalp:
We've discussed current approach internally and decided to delegate object ID resolution from generic implementation in payload function to translation models.

saleor/graphql/translations/mutations.py
```diff
@@ -106,10 +116,23 @@ def perform_mutation(cls, _root, info, **data):
...
         )
         product = ChannelContext(node=product, channel_slug=None)
+
+        def on_commit():
+            if created:
+                info.context.plugins.translation_created(translation)
+                print(
+                    "info.context.plugins.translation_created(translation)",
+                    info.context.plugins.translation_created,
+                )
```
maarcingebala, line 130: 🚔

saleor/plugins/manager.py
```diff
@@ -121,7 +122,7 @@ def __run_method_on_plugins(
         ):
             value = default_value
             plugins = self.get_plugins(channel_slug=channel_slug)
-
+            print(plugins)
```
maarcingebala, line 125: 🚔

saleor/plugins/manager.py
```diff
@@ -727,6 +728,20 @@ def list_payment_sources(
             )
         raise Exception(f"Payment plugin {gateway} is inaccessible!")
 
+    def translation_created(self, translation: "Translation"):
+        print("translation_created")
```
maarcingebala, line 732: 🚔

saleor/plugins/manager.py
```diff
@@ -727,6 +728,20 @@ def list_payment_sources(
...
+        default_value = None
+        return self.__run_method_on_plugins(
+            "translation_created", default_value, translation
+        )
+
+    def translation_updated(self, translation: "Translation"):
+        print("translation_updated")
```
maarcingebala, line 739: 🚔

fowczarek (commented):
Should we implement `TRANSLATION_DELETED` webhook?
Currently, we can't remove translation in Saleor.

maarcingebala (approved):

fowczarek (approved):

PR #7738, opened 2021-07-30
Refactor attributeValues resolvers in translations to use dataloaders

dexon44:
I want to merge this change because it refactor attributeValues resolvers in translations to use dataloaders.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/translations/types.py
+++ b/saleor/graphql/translations/types.py
@@ -17,6 +17,11 @@ from ..core.enums import LanguageCodeEnum
 from ..core.types import LanguageDisplay
 from ..core.utils import str_to_enum
 from ..decorators import permission_required
+from ..page.dataloaders import SelectedAttributesByPageIdLoader
+from ..product.dataloaders import (
+    SelectedAttributesByProductIdLoader,
+    SelectedAttributesByProductVariantIdLoader,
+)
 from .fields import TranslationField
 
 BASIC_TRANSLATABLE_FIELDS = ["id", "name"]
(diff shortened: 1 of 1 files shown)


saleor/graphql/translations/types.py
```diff
@@ -146,13 +151,20 @@ def resolve_product_variant(root: product_models.ProductVariant, info):
...
     @staticmethod
     def resolve_attribute_values(root: product_models.ProductVariant, _info):
-        translatable_values = []
-        for assignment in root.attributes.all():
-            attr = assignment.attribute
-            if attr.input_type in AttributeInputType.TRANSLATABLE_ATTRIBUTES:
-                value = assignment.values.first()
-                translatable_values.append(value)
-        return translatable_values
+        def func(attributes):
```
maarcingebala, line 154: `func` is a good name when you're building a draft of some changes; here it should be renamed to something more understandable by others. `get_translatable_attributes` for example?

maarcingebala: This function is identical in all three resolvers, we could actually extract it outside the class and reuse it. Adding a short docstring would be then useful as well.

tomaszszymanski129: Usually when dealing with single dataloader in resolvers, we compose func name like underscore + resolver name. For ex. `_resolve_attribute_values`. But in that case, yes, it should be extracted (and no underscore then).

saleor/graphql/translations/types.py
```diff
@@ -146,13 +151,20 @@ def resolve_product_variant(root: product_models.ProductVariant, info):
...
+            translatable_values = []
+            for assignment in attributes:
+                attr = assignment["attribute"]
+                if attr.input_type in AttributeInputType.TRANSLATABLE_ATTRIBUTES:
+                    value = assignment["values"][0]
+                    translatable_values.append(value)
+            return translatable_values
+
+        return (
+            SelectedAttributesByProductVariantIdLoader(_info.context)
```
tomaszszymanski129, line 164:
```suggestion
            SelectedAttributesByProductVariantIdLoader(info.context)
```
info is no longer unused so remove `_` from it

saleor/graphql/translations/types.py
```diff
@@ -29,6 +34,20 @@
 ]
 
 
+def get_translatable_attribute_values(attributes: list) -> list:
```
tomaszszymanski129, line 37: Can we do `List[type]` instead of generic `list`?

tomaszszymanski129 (approved):

tomaszszymanski129 (approved):

maarcingebala (approved):

PR #7781, opened 2021-08-06
Fix bug with Index out of range in ProductTranslatableContent

dexon44:
I want to merge this change because it fix bug with ProductTranslatable Content list out of range.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/translations/types.py
+++ b/saleor/graphql/translations/types.py
@@ -46,8 +46,7 @@ def get_translatable_attribute_values(attributes: list) -> List[AttributeValue]:
     for assignment in attributes:
         attr = assignment["attribute"]
         if attr.input_type in AttributeInputType.TRANSLATABLE_ATTRIBUTES:
-            value = assignment["values"][0]
-            translatable_values.append(value)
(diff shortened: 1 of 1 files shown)


saleor/graphql/translations/types.py
```diff
@@ -46,8 +46,9 @@ def get_translatable_attribute_values(attributes: list) -> List[AttributeValue]:
     for assignment in attributes:
         attr = assignment["attribute"]
         if attr.input_type in AttributeInputType.TRANSLATABLE_ATTRIBUTES:
-            value = assignment["values"][0]
-            translatable_values.append(value)
+            if assignment["values"]:
+                value = assignment["values"][0]
```
patrys, line 50: I see we assume there will always be one value here. Why not extend the `translatable_values` list with `assignment["values"]` then? It would work for empty lists, single element lists, and would even continue to work if we were to allow multi-value fields.

dexon44: Done 👍

maarcingebala (approved):

PR #7821, opened 2021-08-11
SALEOR-4122: ExternalNotificationTrigger mutation

mstrumeck:
I want to merge this change because contain new mutation called externalNotificattionTrigger.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


patrys:
Task linked: [SALEOR-4122 Create ExternalNotificationTrigger mutation that allows apps to trigger notification events](https://app.clickup.com/t/2549495/SALEOR-4122)

saleor/graphql/notifications/inputs.py
```diff
@@ -0,0 +1,11 @@
+import graphene
+
+
+class ExternalNotificationTriggerInput(graphene.InputObjectType):
+    ids = graphene.List(graphene.ID, required=True, description="List of customer ids.")
+    extra_payloads = graphene.JSONString(
+        required=True, description="Additionaly payload"
```
IKarbowiak, line 7: 👮‍♀️ 
```suggestion
        required=True, description="Additionaly payload."
```

saleor/graphql/notifications/inputs.py
```diff
@@ -0,0 +1,11 @@
...
+    )
+    external_event_type = graphene.String(
+        required=True, description="External event type description"
```
IKarbowiak, line 10: 👮‍♀️ 
```suggestion
        required=True, description="External event type description."
```

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,142 @@
...
+            generate_order_payload,
+            None,
+            PK__IN,
+            OrderPermissions.MANAGE_ORDERS,
+        ),
+    }
+
+    class Arguments:
+        input = ExternalNotificationTriggerInput(
+            required=True, description="Input for External Notification Trigger"
```
IKarbowiak, line 89: 👮‍♀️ 
```suggestion
            required=True, description="Input for External Notification Trigger."
```

saleor/graphql/notifications/tests/conftest.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@pytest.fixture
+def order_permission():
+    return Permission.objects.get(codename="manage_orders")
+
+
+@pytest.fixture
+def product_permission():
+    return Permission.objects.get(codename="manage_products")
```
IKarbowiak, line 43: We already have fixtures for these permissions in `saleor/tests/fixtures.py`

IKarbowiak (approved):

saleor/graphql/api.py
```diff
@@ -1,5 +1,7 @@
 from graphene_federation import build_schema
 
+from saleor.graphql.notifications.schema import ExternalNotificationMutations
```
fowczarek, line 3: We should use relative import here.

saleor/graphql/notifications/inputs.py
```diff
@@ -0,0 +1,11 @@
+import graphene
```
fowczarek, line 1: In other modules, we store inputs in `mutations.py` or `mutations\<file_name>.py`. IMO we should make it consistently. If we want to have inputs in a separate file we should do it for all modules. 
It could be a good idea if we want to reuse this input in multiple files. In this case, we use this input only in one mutations file.

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,142 @@
...
+from saleor.core.exceptions import PermissionDenied
+from saleor.core.permissions import (
+    CheckoutPermissions,
+    OrderPermissions,
+    ProductPermissions,
+)
+from saleor.graphql.core.types.common import ExternalNotificationError
+from saleor.invoice.models import Invoice
+from saleor.order.models import Fulfillment, Order
+from saleor.product.models import Product, ProductVariant
```
fowczarek, line 16: We shouldn't mix relative imports with absolute imports.

saleor/graphql/notifications/tests/test_external_notification_query.py
```diff
@@ -0,0 +1,508 @@
...
+            admin_email_plugin_notify,
+            webhook_plugin_notify,
+            staff_api_client,
+            external_notification_trigger_query,
+            sendgrid_email_plugin,
+            ids,
+            permission_manage_users,
+        )
+
+    def _common_part(
```
maarcingebala, line 475: If all tests have "the common part", shouldn't they be rewritten with the usage of `pytest.mark.parametrize`?

fowczarek:
Missing changelog.

saleor/graphql/notifications/error_codes.py
```diff
@@ -0,0 +1,7 @@
+from enum import Enum
+
+
+class ExternalNotificationErrorCodes(Enum):
+    CANNOT_ASSIGN_NODE = "cannot_assign_node"
```
korycins, line 5: Where do you sue this error?

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,142 @@
...
+    generate_product_payload,
+    generate_product_variant_payload,
+)
+from ...core.mutations import BaseMutation
+from ..inputs import ExternalNotificationTriggerInput
+
+
+class ExternalNotificationTrigger(BaseMutation):
+    PK__IN = "pk__in"
+    PAYLOAD_MAPPING = {
```
maarcingebala, line 35: It doesn't have to be a class variable, we could move it outside.

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,142 @@
...
+    class Meta:
+        description = "Send external notification to given customers."
+        error_type_class = ExternalNotificationError
+
+    @classmethod
+    def perform_mutation(cls, root, info, **data):
+        manager = info.context.plugins
+        ids, extra_payloads, external_event_type = cls.get_cleaned_data(data)
+        model_type, pks = resolve_global_ids_to_primary_keys(ids)
+        if payload_params := cls.PAYLOAD_MAPPING.get(model_type):
```
korycins, line 102: We shouldn't mix in mutation logic with business logic.  Mutation responsibility is to validate received data, clean up data,  and call a proper action.
All things related to prepare notification payload, mapping, and calling `notify` should be extracted to a separate file. I would consider if we can extract it to saleor/core/notification file.

saleor/graphql/notifications/inputs.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class ExternalNotificationTriggerInput(graphene.InputObjectType):
+    ids = graphene.List(graphene.ID, required=True, description="List of customer ids.")
+    extra_payloads = graphene.JSONString(
```
korycins, line 6: Shouldn't be `extra_payload`?

saleor/graphql/notifications/inputs.py
```diff
@@ -0,0 +1,11 @@
...
+        required=True, description="Additionaly payload."
+    )
```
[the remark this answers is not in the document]
maarcingebala: Also, I think the "extra payload" should be optional in the schema? The name "extra" suggests it.

saleor/graphql/notifications/inputs.py
```diff
@@ -0,0 +1,11 @@
+import graphene
+
+
+class ExternalNotificationTriggerInput(graphene.InputObjectType):
+    ids = graphene.List(graphene.ID, required=True, description="List of customer ids.")
```
korycins, line 5: description -> This can be IDs of customers or Orders

saleor/graphql/notifications/inputs.py
```diff
@@ -0,0 +1,11 @@
...
+    extra_payloads = graphene.JSONString(
+        required=True, description="Additionaly payload."
+    )
+    external_event_type = graphene.String(
+        required=True, description="External event type description."
```
maarcingebala, line 10: Can we describe this arg properly? What values does it expect? From what I see it can accept Sendgrid template IDs, what other values does it accept? We should explain it.

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,142 @@
...
+
+    class Meta:
+        description = "Send external notification to given customers."
+        error_type_class = ExternalNotificationError
+
+    @classmethod
+    def perform_mutation(cls, root, info, **data):
+        manager = info.context.plugins
+        ids, extra_payloads, external_event_type = cls.get_cleaned_data(data)
+        model_type, pks = resolve_global_ids_to_primary_keys(ids)
```
korycins, line 101: I don't think that PAYLOAD_MAPPING is needed here. AFAIR we are going to use this feature for `User` and `Order` type. No need to add the rest of them to mapping 🤔

saleor/graphql/notifications/error_codes.py
```diff
@@ -0,0 +1,7 @@
...
+    GRAPHQL_ERROR = "graphql_error"
+    WRONG_PLUGIN_ID = "wrong_plugin_id"
```
maarcingebala, line 7: Where is this error code used? Also I don't see usage of `CANNOT_ASSIGN_NODE`.

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,142 @@
...
+        if payload_params := cls.PAYLOAD_MAPPING.get(model_type):
+            (
+                model,
+                payload_function,
+                input_type,
+                pk_lookup_type,
+                permission_type,
+            ) = payload_params
+            if cls._is_user_contain_permission(info.context, permission_type):
+                if input_type == list:
```
korycins, line 111: Not sure what `input_type` means

korycins: I see that it is used in mapping for productVariant. I don't think that we will need this as we're going to prepare payload for orders and customers (afair)

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,142 @@
...
+                    payload_inputs = model.objects.filter(**{pk_lookup_type: pks})
+                    for payload_input in payload_inputs:
+                        payload = json.loads(payload_function(payload_input))
+                        payload = payload[0] if type(payload) == list else payload
+                        payload.update(extra_payloads)
+                        cls._notify(data, manager, external_event_type, payload)
+        return cls()
+
+    @classmethod
+    def _is_user_contain_permission(cls, context, permission_type):
```
maarcingebala, line 127: `_user_has_permission` - I think it sounds better.

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,142 @@
...
+                    payloads = json.loads(payload_function(payload_input))
+                    for payload in payloads:
+                        payload.update(extra_payloads)
+                        cls._notify(data, manager, external_event_type, payload)
+                else:
+                    payload_inputs = model.objects.filter(**{pk_lookup_type: pks})
+                    for payload_input in payload_inputs:
+                        payload = json.loads(payload_function(payload_input))
+                        payload = payload[0] if type(payload) == list else payload
+                        payload.update(extra_payloads)
```
korycins, line 122: The extra payload should be placed in a separate field in generated payload. (the risk of overwriting a payload fields)

saleor/graphql/notifications/tests/test_external_notification_dispatch.py
```diff
@@ -0,0 +1,187 @@
+from unittest.mock import patch
+
+from graphql_relay.node.node import to_global_id
+
+from ....core.notify_events import UserNotifyEvent
+from ....product.models import ProductVariant
+
+
+@patch("saleor.plugins.sendgrid.tasks.send_email_with_dynamic_template_id.delay")
```
korycins, line 9: We shouldn't mock any plugins from Saleor. Think of a plugin as something that doesn't belong to Saleor. 
Here you should mock the proper call of the manager and call `assert_called_once_with(..)`. 
Then you can add tests in sendgrid module to confirm that sendgrid Plugin triggered proper action.

saleor/graphql/notifications/tests/test_external_notification_dispatch.py
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    response = staff_api_client.post_graphql(
+        external_notification_trigger_query,
+        variables,
+        permissions=[permission_manage_products],
+    )
+
+    assert response.status_code == 200
+    webhook_plugin_notify.assert_called_once()
```
korycins, line 94: `assert_called_once_with()` will handle this ;)

saleor/graphql/notifications/tests/test_external_notification_query.py
```diff
@@ -0,0 +1,508 @@
...
+        check_no_permissions=False,
+    )
+    assert response.status_code == status_code
+
+
+@patch("saleor.plugins.webhook.plugin.WebhookPlugin.notify")
+@patch("saleor.plugins.admin_email.plugin.AdminEmailPlugin.notify")
+@patch("saleor.plugins.user_email.plugin.UserEmailPlugin.notify")
+@patch("saleor.plugins.sendgrid.plugin.SendgridEmailPlugin.notify")
+def test_notify_sendgrid_via_external_notification_trigger_for_all_plugins(
```
korycins, line 35: As mentioned earlier, no need to mock and include plugins in Saleor tests. You need to only confirm that we called a proper method on PluginsManager.

saleor/graphql/notifications/tests/test_external_notification_query.py
```diff
@@ -0,0 +1,508 @@
...
+    )
+
+    assert response.status_code == 200
+    payload = sendgrid_plugin_notify.call_args[1]["payload"]
+    assert "TEST" in payload.keys()
+    assert "TEST_LIST" in payload.keys()
+    assert ["GUEST1", "GUEST2"] == payload.get("TEST_LIST")
+
+
+class TestSetWithVariousDataTypes:
```
korycins, line 199: We don't use TestCases. If you have something which is similar for preparing data for many tests then you can create a fixture for this.
I see that you have common part where you have an asserts. Not sure of this. I prefer to have explicit assert in each test, to see all conditions that the test needs to pass. 
If you really need this, then you can wrap it in a simple function. No need to create a class for this.

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    class Arguments:
+        input = ExternalNotificationTriggerInput(
+            required=True, description="Input for External Notification Trigger."
+        )
+        pluginId = graphene.String(description="The ID of notification plugin.")
+
+    class Meta:
+        description = "Send external notification to given customers."
```
maarcingebala, line 94: 1. It's not only for customers.
2. Can we actually describe what the mutation does? I mean sth like: "Trigger sending a notification with the `notify` plugin method. Serializes nodes provided as `ids` parameter and includes this data in the notification payload."

saleor/graphql/notifications/inputs.py
maarcingebala, line 5: It's not a list of customer ids. It's a list of node IDs that will be serialized and included in the notification payload. It should be mentioned in the description.

saleor/tests/fixtures.py
```diff
@@ -4223,3 +4224,102 @@ def app_export_event(app_export_file):
         app=app_export_file.app,
         parameters={"message": "Example error message"},
     )
+
+
+@pytest.fixture
```
korycins, line 4229: As I explained above. No need to extract it to the global fixtures file.  (Yes I know that we did this with Vatlayer, but this is specific case which we're going to resolve in the future)

saleor/plugins/sendgrid/plugin.py
```diff
@@ -213,10 +214,17 @@ def notify(self, event: NotifyEventType, payload: dict, previous_value):
...
-            return previous_value
+        event_in_notify_event = event in UserNotifyEvent.CHOICES
+        event_in_event_map = event in EVENT_MAP
+
+        if not event_in_notify_event or not event_in_event_map:
+            logger.warning(f"Send email with event {event} as dynamic template ID.")
+            send_email_with_dynamic_template_id.delay(
+                payload, event, asdict(self.config)
+            )
+            return
```
korycins, line 225: you should return previous value

saleor/plugins/sendgrid/plugin.py
```diff
@@ -213,10 +214,17 @@ def notify(self, event: NotifyEventType, payload: dict, previous_value):
         if not self.active:
             return previous_value
 
-        if event not in UserNotifyEvent.CHOICES:
-            return previous_value
+        event_in_notify_event = event in UserNotifyEvent.CHOICES
+        event_in_event_map = event in EVENT_MAP
+
+        if not event_in_notify_event or not event_in_event_map:
+            logger.warning(f"Send email with event {event} as dynamic template ID.")
```
korycins, line 221: Why warning?

saleor/graphql/notifications/mutations/external_notification_trigger.py
mstrumeck: What's the difference? Only `ExternalNotificationTrigger` gonna use that mapping so IMO should be kept within that class.

saleor/graphql/notifications/mutations/external_notification_trigger.py
mstrumeck: 1. Frankly, I can't find any information for which models `externalNotificationTrigger` should be suitable. So I've decided to create a trigger for all with the payload function. 
2. Where can I find additional information about the types of the model? Did I miss something from https://roughtrade.atlassian.net/wiki/spaces/RTBD/pages/840499201/External+Notification+using+Sendgrid ? 
3. In the RT case, we would like to pass the `ProductVariant` payload. I will create a mapping for `User`, `Order` and `ProductVariant` models, agree?

maarcingebala: We don't have access to this board (and we shouldn't). If you're not sure about the scope, you should discuss it with whoever is leading the backend in the RT team - I have as much information as in the PBR document linked to the Clickup task.

saleor/graphql/notifications/tests/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+          ) {
+            errors {
+              message
+            }
+          }
+      }
+    """
+
+
+query_test_data = [
```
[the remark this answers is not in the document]
mstrumeck: Are you sure? I always thought that we should keep test data in a separate file.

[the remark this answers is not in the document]
maarcingebala: If we had to reuse this data in multiple test files - sure. Also for fixtures - that's the usual pattern to have them in a separate file. But `query_test_data` is not a fixture but a parameter to `parametrize`, it seems to be used only in one file, and if you look at other usages of Saleor, we always keep parameterize data next to tests.

saleor/graphql/notifications/mutations/external_notification_trigger.py
maarcingebala: Readability - the entire class gets longer for no specific reason. Also, this adds an additional indentation level to this dictionary.

saleor/graphql/notifications/mutations/external_notification_trigger.py
maarcingebala: I think we could leave it as it is. If there was more logic then yes, but in this case, this mutation has a very simple responsibility.

saleor/tests/fixtures.py
mstrumeck: AFAIK that almost 50 test cases are using this fixture. How should I refactor that? Move that fixture into each module conftests?

saleor/core/notification/mutation_handler.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def trigger_notifications(manager, external_event_type, payload, plugin_id=None):
+    if plugin_id:
+        manager.notify_in_single_plugin(plugin_id, external_event_type, payload)
+    else:
+        manager.notify(event=external_event_type, payload=payload)
+
+
+class ExternalNotificationTriggerPayload:
```
maarcingebala, line 56: You don't need this class. It doesn't do anything except for calling a few methods and passing class attributes as params. We can easily refactor it as simple functions. First a few notes:

1. `pks` and `model` - these args are passed only to create a queryset inside the class. We can reduce the number of arguments, if we extracted the queryset inside the mutation, and passed it as a single argument. See the examples below.
2. `payload_inputs` - this name suggests that this is some kind of data dictionary, while this are objects from Django ORM - I renamed it to `objects` (could be also `queryset`).

```python
def _get_extracted_payload_input(object, payload_function, extra_payload):
    payload = json.loads(payload_function(object))
    payload = payload[0] if type(payload) == list else payload
    payload.update({"extra_payload": extra_payload})
    return payload


def get_external_notification_payload(
    objects, payload_function, input_type, extra_payload
):
    if input_type == list:
        payloads = json.loads(payload_function(objects))
        for payload in payloads:
            payload.update({"extra_payload": extra_payload})
        return payloads

    return [
        _get_extracted_payload_input(object, payload_function, extra_payload)
        for object in objects
    ]
```

In the mutation, we can use it as:
```python
objects = model.objects.filter(pk__in=pks)
payload = get_external_notification_payload(
    objects, payload_function, input_type, extra_payload
)
```

Something like that should work. What do you think? IMO this is much simpler. You really don't need classes most of the time.

saleor/tests/fixtures.py
korycins: It should be used with sendgird's tests, so it should be placed there in the conftest

saleor/core/notification/mutation_handler.py
```diff
@@ -0,0 +1,86 @@
...
+from ...order.models import Order
+from ...product.models import ProductVariant
+from ...webhook.payloads import (
+    generate_customer_payload,
+    generate_order_payload,
+    generate_product_variant_payload,
+)
+
+PAYLOAD_MAPPING = {
+    "ProductVariant": (
```
korycins: The assumption of this mapping is to attach a defined object to the payload, which will be sent via webhook notify/ or by plugins. 
We want this mapping for User and for Order because these objects have enough details about customer which we can use to send notification (name, email, phone etc). ProductVariant doesn't have such details so there is no need to add it here.

saleor/plugins/webhook/plugin.py
```diff
@@ -211,12 +211,20 @@ def checkout_updated(self, checkout: "Checkout", previous_value: Any) -> Any:
...
-    def notify(self, event: "NotifyEventType", payload: dict, previous_value) -> Any:
+    def notify(self, event: NotifyEventType, payload: dict, previous_value) -> Any:
         if not self.active:
             return previous_value
+
+        notify_user_event = WebhookEventType.NOTIFY_USER
         data = {"notify_event": event, "payload": payload}
+
+        if event not in NotifyEventType.CHOICES:
+            logger.warning(
```
korycins, line 222: Why warning? This could produce a massive logs

saleor/plugins/manager.py
```diff
@@ -936,6 +936,23 @@ def notify(
             "notify", default_value, event, payload, channel_slug=channel_slug
         )
 
+    def notify_in_single_plugin(
+        self,
+        plugin_id: str,
+        event: "NotifyEventTypeChoice",
```
korycins, line 942: Are you sure that `NotifyEventTypeChoice` exsist? I don't see such obejct

saleor/core/notification/mutation_handler.py
```diff
@@ -0,0 +1,86 @@
...
+    if isinstance(payloads, list):
+        for payload in payloads:
+            trigger_notifications(manager, external_event_type, payload, plugin_id)
+    else:
+        trigger_notifications(manager, external_event_type, payloads, plugin_id)
+
+
+def trigger_notifications(manager, external_event_type, payload, plugin_id=None):
+    if plugin_id:
+        manager.notify_in_single_plugin(plugin_id, external_event_type, payload)
```
korycins, line 21: Currently `event` from `notify` expects a value from `NotifyEventType` (typing). This changes "extend" values which `notify`  will accept. Now it should be normal str OR `NotifyEventType`

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,82 @@
...
+    extra_payload = graphene.JSONString(
+        description=(
+            "Additional payload that will be merged with"
+            "the one based on the bussines object ID."
+        )
+    )
+    external_event_type = graphene.String(
+        required=True,
+        description=(
+            "External event type. In case of invalid type,"
```
maarcingebala, line 35:
```suggestion
            "External event type. In case of invalid type, "
```

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,82 @@
...
+            "Saleor will consider the content"
+            "of that field like a ID of dynamic template."
+        ),
+    )
+
+
+class ExternalNotificationTrigger(BaseMutation):
+    class Arguments:
+        input = ExternalNotificationTriggerInput(
+            required=True, description="Input for External Notification Trigger."
```
maarcingebala, line 45:
```suggestion
            required=True, description="Input for external notification trigger."
```

saleor/graphql/notifications/tests/test_external_notification_dispatch.py
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    response = staff_api_client.post_graphql(
+        external_notification_trigger_query,
+        variables,
+        permissions=[permission_manage_users],
+    )
+
+    assert response.status_code == 200
+    assert (
```
korycins, line 115: If you want to check the logs, then these tests should be separated by plugins. And placed in plugin's dir. 
Currently, if this test will fail, I don't know which of the plugin should I fix.

saleor/graphql/notifications/tests/test_external_notification_query.py
```diff
@@ -0,0 +1,136 @@
+import json
+from unittest.mock import patch
+
+import pytest
+from graphql_relay.node.node import to_global_id
+
+from ....account.models import User
+from ....core.notify_events import UserNotifyEvent
+from ....graphql.tests.utils import assert_no_permission
+from ....plugins.sendgrid.plugin import SendgridEmailPlugin
```
korycins, line 10: Use `saleor.plugins.tests.sample_plugins.PluginSample`

saleor/graphql/notifications/tests/test_external_notification_query.py
korycins: Can we rename the test to be more generic and related to manager's notify instead of Sendgrid?

saleor/graphql/notifications/tests/test_external_notification_query.py
```diff
@@ -0,0 +1,136 @@
...
+    plugin_manager_notify_mock,
+    settings,
+    staff_users,
+    external_notification_trigger_query,
+    staff_api_client,
+    permission_manage_users,
+):
+
+    settings.PLUGINS = [
+        "saleor.plugins.user_email.plugin.UserEmailPlugin",
```
korycins, line 90: As above. Use dedicated Sample plugins

saleor/graphql/notifications/tests/test_external_notification_query.py
```diff
@@ -0,0 +1,136 @@
...
+    staff_api_client,
+):
+
+    variables = {
+        "input": {
+            "ids": [to_global_id(User.__name__, user.id) for user in staff_users],
+            "extraPayload": "{}",
+            "externalEventType": UserNotifyEvent.ORDER_CANCELED,
+        },
+        "pluginId": SendgridEmailPlugin.PLUGIN_ID,
```
korycins, line 128: `saleor.plugins.tests.sample_plugins.PluginSample`

saleor/plugins/sendgrid/plugin.py
```diff
@@ -213,10 +214,17 @@ def notify(self, event: NotifyEventType, payload: dict, previous_value):
         if not self.active:
```
korycins, line 214: Typing for `notify` is incorrect here. It should be a `Union[NotifyEventType, str]`

saleor/plugins/sendgrid/plugin.py
```diff
@@ -213,10 +214,17 @@ def notify(self, event: NotifyEventType, payload: dict, previous_value):
...
+
+        if not event_in_notify_event or not event_in_event_map:
+            logger.info(f"Send email with event {event} as dynamic template ID.")
+            send_email_with_dynamic_template_id.delay(
+                payload, event, asdict(self.config)
+            )
             return previous_value
 
-        if event not in EVENT_MAP:
+        if not event_in_notify_event:
```
korycins, line 227: This condition will never be executed (see above `if` condition)

saleor/plugins/sendgrid/plugin.py
```diff
@@ -213,10 +214,17 @@ def notify(self, event: NotifyEventType, payload: dict, previous_value):
         if not self.active:
             return previous_value
 
-        if event not in UserNotifyEvent.CHOICES:
+        event_in_notify_event = event in UserNotifyEvent.CHOICES
+        event_in_event_map = event in EVENT_MAP
+
+        if not event_in_notify_event or not event_in_event_map:
```
korycins, line 220: I would say, that we shouldn't trigger a dynamic template for the events defined by Saleor but not defined in SendgridPlugin.
IMHO It should be:
`if not event_in_notify_event:`  instead of `if not event_in_notify_event or not event_in_event_map:`

saleor/plugins/sendgrid/tests/conftest.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 
-from ...manager import get_plugins_manager
-from ..plugin import SendgridEmailPlugin
+from ....plugins.manager import get_plugins_manager
```
korycins, line 3: I believe the previous import was correct.

saleor/plugins/webhook/plugin.py
```diff
@@ -211,12 +211,20 @@ def checkout_updated(self, checkout: "Checkout", previous_value: Any) -> Any:
             WebhookEventType.CHECKOUT_UPDATED, checkout_data
         )
 
-    def notify(self, event: "NotifyEventType", payload: dict, previous_value) -> Any:
+    def notify(self, event: NotifyEventType, payload: dict, previous_value) -> Any:
```
korycins, line 214: I beleive, it should be a `Union[NotifyEventType, str]` - Saleor's events + dynamic event [str]

saleor/core/notification/validation.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+
+def validate_and_get_payload_params(model_type):
+    if payload_params := PAYLOAD_MAPPING_FOR_CUSTOM_NOTIFICATION.get(model_type):
+        return payload_params
+    raise ValidationError(
+        f"Wrong type of model."
+        f" ExternalNotificationTrigger is suitable"
+        f" for {','.join(PAYLOAD_MAPPING_FOR_CUSTOM_NOTIFICATION.keys())} models.",
```
tomaszszymanski129, line 50:
```suggestion
        f"Wrong type of model. "
        f"ExternalNotificationTrigger is suitable "
        f"for {','.join(PAYLOAD_MAPPING_FOR_CUSTOM_NOTIFICATION.keys())} models.",
```

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+        plugin_id = graphene.String(description="The ID of notification plugin.")
+
+    class Meta:
+        description = (
+            "Trigger sending a notification with the notify plugin method."
+            " Serializes nodes provided as ids parameter and includes this data in"
+            " the notification payload."
```
tomaszszymanski129, line 53:
```suggestion
            "Trigger sending a notification with the notify plugin method. "
            "Serializes nodes provided as ids parameter and includes this data in "
            "the notification payload."
```

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,82 @@
...
+                payload = get_external_notification_payload(
+                    objects, extra_payload, payload_function
+                )
+                send_notification(
+                    manager, external_event_type, payload, plugin_id=plugin_id
+                )
+        return cls()
+
+    @classmethod
+    def _is_user_has_permission(cls, context, permission_type):
```
tomaszszymanski129, line 79: name seems confusing, I would simply call it `_requestor_has_permission` since context can have an app and basemutation implementation handles that.

saleor/plugins/sendgrid/plugin.py
```diff
@@ -209,14 +211,21 @@ def __init__(self, *args, **kwargs):
...
+
+        if not event_in_notify_event:
+            logger.info(f"Send email with event {event} as dynamic template ID.")
+            send_email_with_dynamic_template_id.delay(
+                payload, event, asdict(self.config)
+            )
             return previous_value
 
-        if event not in EVENT_MAP:
+        if not event_in_event_map:
```
tomaszszymanski129, line 228: IMO seems more confusing than it was

CHANGELOG.md
```diff
@@ -129,6 +129,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add app support for events - #7622 by @IKarbowiak
 - Add date & date time attributes - #7500 by @piotrgrundas
 - Add `withChoices` flag for Attribute type - #7733 by @dexon44
+- Add ExternalNotificationTrigger mutation - #7821 by @mstrumeck
```
maarcingebala, line 132: Please rebase the PR with master and move this changelog entry to the "3.1" section - these changes will be released in 3.1.

saleor/core/notification/validation.py
```diff
@@ -0,0 +1,65 @@
...
+def validate_and_get_payload_params(model_type):
+    if payload_params := PAYLOAD_MAPPING_FOR_CUSTOM_NOTIFICATION.get(model_type):
+        return payload_params
+    raise ValidationError(
+        {
+            "external_notification_trigger": ValidationError(
+                f"Wrong type of model."
+                f"ExternalNotificationTrigger is suitable"
+                f"for {','.join(PAYLOAD_MAPPING_FOR_CUSTOM_NOTIFICATION.keys())} "
+                f"models.",
```
maarcingebala, line 61: Can't we have instead:
```suggestion
                f"Wrong type of model - ExternalNotificationTrigger is suitable for "
                f"{','.join(PAYLOAD_MAPPING_FOR_CUSTOM_NOTIFICATION.keys())} models.",
```
Seems a bit simpler.

saleor/plugins/webhook/plugin.py
maarcingebala: What about this comment? How should we address it @korycins @mstrumeck?

saleor/core/notification/validation.py
mstrumeck: Flake8 will throw `saleor/core/notification/validation.py:60:89: E501 line too long (90 > 88 characters)` if i back to previous version.

maarcingebala: You sure? If I paste that to my editor it seems under 88 chars:
image

We should at least fix the spaces, now this part:
```
f"Wrong type of model."
f"ExternalNotificationTrigger is suitable"
```
will be printed as:
```
Wrong type of model.ExternalNotificationTrigger is suitable
```

patrys: Is this an error returned by GraphQL? If so, there are no "models" in GraphQL, it has "object types."

saleor/plugins/webhook/plugin.py
mstrumeck: Sorry, I thought that I've changed that into `logger.info`

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,82 @@
...
+    class Arguments:
+        input = ExternalNotificationTriggerInput(
+            required=True, description="Input for External Notification Trigger. "
+        )
+        plugin_id = graphene.String(description="The ID of notification plugin.")
+
+    class Meta:
+        description = (
+            "Trigger sending a notification with the notify plugin method."
+            "Serializes nodes provided as ids parameter and includes this data in"
```
maarcingebala, line 52:
```suggestion
            "Trigger sending a notification with the notify plugin method. "
            "Serializes nodes provided as ids parameter and includes this data in "
```
When constructing long strings like this you need to remember about spaces at the end of the lines. Otherwise, the entire message will be joined without them and it will be unreadable.

saleor/plugins/sendgrid/tests/test_plugin.py
```diff
@@ -14,6 +19,134 @@ def test_get_event_map():
...
+    }
+    mocked_event_map.__getitem__.side_effect = event_map.__getitem__
+    mocked_event_map.__contains__.return_value = True
+    mocked_event_map.get.side_effect = event_map.get
+
+    expected_payload = [get_user_custom_payload(user) for user in staff_users]
+    for payload in expected_payload:
+        payload["extra_payload"] = test_json
+
+    plugin.notify(UserNotifyEvent.ACCOUNT_PASSWORD_RESET, expected_payload, None)
```
korycins, line 65: This test doesn't test an external event. You pass an event that exists in EVENT_MAP (and in your mocked event map you also have it). No need to mock the event map, your function will exit before call of `if` related to event.  You should mock here a `send_email_with_dynamic_template_id.delay`  call.

saleor/plugins/sendgrid/tests/test_plugin.py
```diff
@@ -14,6 +19,134 @@ def test_get_event_map():
...
+    notify_single_plugin_mock,
+    settings,
+    staff_users,
+    staff_api_client,
+    permission_manage_users,
+    sendgrid_email_plugin,
+    product_with_single_variant,
+):
+
+    settings.PLUGINS = [
```
korycins, line 83: You're testing SendgridPlugin and activating PluginSample. I think that this settings is not needed here

saleor/plugins/sendgrid/tests/test_plugin.py
```diff
@@ -14,6 +19,134 @@ def test_get_event_map():
...
+
+    plugin.notify(UserNotifyEvent.ACCOUNT_PASSWORD_RESET, expected_payload, None)
+
+    mocked_event_task.delay.assert_called_once_with(
+        expected_payload, asdict(plugin.config)
+    )
+
+
+@patch("sendgrid.base_interface.BaseInterface.send")
+def test_send_notification_to_customers_with_product_variant_payload(
```
korycins, line 73: I think that we already talked about this. No need to add tests to the plugin which tests a Saleor's mutation. 
From the test perspective, you need to confirm that plugin logic works as expected with the given input.
Tests in sendgrid modules should use test only a expected plugin.notify calls.

saleor/plugins/sendgrid/tests/test_plugin.py
```diff
@@ -14,6 +19,134 @@ def test_get_event_map():
...
+        external_notification_trigger_query,
+        variables,
+        permissions=[permission_manage_users],
+    )
+
+    assert response.status_code == 200
+    assert notify_single_plugin_mock.call_count == 3
+
+
+def test_notify_sendgrid_via_external_notification_trigger_lack_of_logs(
```
korycins, line 113: I think that we already talked about this. No need to add tests to the plugin which tests a Saleor's mutation. 
From the test perspective, you need to confirm that plugin logic works as expected with the given input.
Tests in sendgrid modules should use test only a expected plugin.notify calls.

korycins (approved):

maarcingebala (approved):

eCzerniak (approved):

saleor/graphql/notifications/tests/test_external_notification_query.py
```diff
@@ -0,0 +1,156 @@
...
+                "channel": "c-pln",
+            }
+        },
+        400,
+    ),
+]
+
+
+@pytest.mark.parametrize("variables, status_code", query_test_data)
+def test_query(
```
korycins, line 83: Wouldn't be good to have `test_external_notification_trigger_query` ?

korycins: Also, I am not convinced about this test.  It tests response codes that with graphql are not really important (you can receive 200 with error from API). 
Instead of this, you should do something like this:
1st test for incorrect arguments and checking that API returned errors (something similar to this):
```
    data = content["data"]["appInstall"]

    errors = data["errors"]
    assert not data["appInstallation"]
    assert len(errors) == 1
    error = errors[0]
    assert error["field"] == "permissions"
    assert error["code"] == AppErrorCode.OUT_OF_SCOPE_PERMISSION.name
    assert error["permissions"] == [PermissionEnum.MANAGE_ORDERS.name]
```

2nd test - correct input - it is already covered by below tests

saleor/graphql/notifications/tests/test_external_notification_query.py
```diff
@@ -0,0 +1,156 @@
...
+            "channel": channel_PLN.slug,
+        },
+        "pluginId": PluginSample.PLUGIN_ID,
+    }
+
+    response = staff_api_client.post_graphql(
+        external_notification_trigger_query, variables
+    )
+    assert_no_permission(response)
+    assert response.status_code == 200
```
korycins, line 155: No need to check response.status_code. You should check if response from API returned an error field. For missing required permission - `assert_no_permission` is fine.

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    external_event_type = graphene.String(
+        required=True,
+        description=(
+            "External event type. In case of invalid type, "
+            "Saleor will consider the content"
+            "of that field like a ID of dynamic template."
+        ),
+    )
+    channel = graphene.String(
```
korycins, line 41: I would consider if this place is right. Your mutation has input: `input` and `plugin_id`. Plugin_id determines the configuration of running this mutation. `Chanel` does the same.  For me, it should be in the same place as `plugin_id`

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,95 @@
...
+            "Additional payload that will be merged with "
+            "the one based on the bussines object ID."
+        )
+    )
+    external_event_type = graphene.String(
+        required=True,
+        description=(
+            "External event type. In case of invalid type, "
+            "Saleor will consider the content"
+            "of that field like a ID of dynamic template."
```
korycins, line 38: `"Saleor will consider the content of that field like an ID of the dynamic template."` - it is not a truth. It depends on how each plugin handles that. For `sendgrid, it will be a dynamic template ID. For webhook it will be a header or something. for XX it will be something different. The description should only inform that this field is passed to a plugin as an event type. Our documentation should contain info on how this mutation works with each plugin.

saleor/graphql/notifications/tests/test_external_notification_query.py
```diff
@@ -0,0 +1,156 @@
...
+        external_notification_trigger_query,
+        variables,
+        permissions=[permission_manage_users],
+        check_no_permissions=False,
+    )
+    assert response.status_code == status_code
+
+
+@patch("saleor.plugins.manager.PluginsManager.notify")
+def test_notify_via_external_notification_trigger_for_all_plugins(
```
korycins, line 101: I think the name is not correct. It doesn't test mutation for all plugins. It just test the mutation and confirms that pluginmanager is called.

saleor/graphql/notifications/tests/test_external_notification_query.py
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    response = staff_api_client.post_graphql(
+        external_notification_trigger_query,
+        variables,
+        permissions=[permission_manage_users],
+    )
+
+    assert response.status_code == 200
```
korycins, line 128: I think it doesn't mean anything. If you want to be sure that your mutation doesn't returned an error, then you should confirm that returned payload doesn't have field `errors`. (You can find a lot of examples in Saleor's tests.)

saleor/core/notification/validation.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+
+def validate_and_get_external_event_type(data_input):
+    if external_event_type := data_input.get("external_event_type"):
+        return external_event_type
+    raise ValidationError(
+        {
+            "external_event_type": ValidationError(
+                "The obligatory param 'external_event_type' is missing or is empty.",
```
korycins, line 46: `external_event_type` -> `externalEventType` 
All var names are converted to camelCase, so in error msgs we have to update them.

korycins: Also, I don't think that this field can be missing. It is already covered by API and parameter `required=True`

saleor/core/notification/validation.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def validate_ids_and_get_model_type_and_pks(data_input):
+    if ids := data_input.get("ids"):
+        model_type, pks = resolve_global_ids_to_primary_keys(ids)
+        return model_type, pks
+    raise ValidationError(
+        {
+            "ids": ValidationError(
+                "The obligatory param 'ids' is missing or is empty.",
```
korycins, line 33: Also, I don't think that this field can be missing. It is already covered by API and parameter `required=True`

saleor/core/notification/validation.py
```diff
@@ -0,0 +1,98 @@
...
+                    "channel": ValidationError(
+                        "Cannot complete checkout with inactive channel.",
+                        code=ExternalNotificationTriggerErrorCode.CHANNEL_INACTIVE,
+                    )
+                }
+            )
+        raise ValidationError(
+            {
+                "channel": ValidationError(
+                    "The channel with given not exists.",
```
korycins, line 69: Is the msg correct?

saleor/core/notification/validation.py
```diff
@@ -0,0 +1,98 @@
...
+                "The obligatory param 'external_event_type' is missing or is empty.",
+                code=ExternalNotificationTriggerErrorCode.REQUIRED,
+            )
+        }
+    )
+
+
+def validate_and_get_channel(data_input):
+    if channel_slug := data_input.get("channel"):
+        if Channel.objects.filter(slug=channel_slug).exists():
```
korycins, line 55: Use the function dedicated to this `saleor.graphql.channel.utils.validate_channel`

saleor/core/notification/validation.py
```diff
@@ -0,0 +1,98 @@
...
+                    "The channel with given not exists.",
+                    code=ExternalNotificationTriggerErrorCode.REQUIRED,
+                )
+            }
+        )
+
+    raise ValidationError(
+        {
+            "channel": ValidationError(
+                "The obligatory param 'channel' is missing or is empty.",
```
korycins, line 78: Also, I don't think that this field can be missing. It is already covered by API and parameter required=True

korycins (approved):

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,95 @@
...
+            description=(
+                "Channel slug. "
+                "Saleor will send a notification within a provided channel. "
+                "Please, make sure that necessary plugins are active."
+            ),
+        )
+
+    class Meta:
+        description = (
+            "Trigger sending a notification with the notify plugin method. "
```
maarcingebala, line 59: Missing `ADDED_IN_31` note.

saleor/graphql/notifications/mutations/external_notification_trigger.py
```diff
@@ -0,0 +1,95 @@
...
+    extra_payload = graphene.JSONString(
+        description=(
+            "Additional payload that will be merged with "
+            "the one based on the bussines object ID."
+        )
+    )
+    external_event_type = graphene.String(
+        required=True,
+        description=(
+            "External event type. " "This field is passed to a plugin as an event type."
```
maarcingebala, line 37:
```suggestion
            "External event type. This field is passed to a plugin as an event type."
```

PR #7822, opened 2021-08-11
Let me urlencode ":" so I can run with Google Cloud SQL

matasar:

I can't use `DATABASE_URL` to connect to cloud SQL. Took this approach from https://github.com/CAVaccineInventory/vial/commit/316cd096b2fa1156d86c89be5ef118ab67b79415

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -72,6 +72,10 @@ DATABASES = {
     )
 }
 
+DATABASES["default"]["HOST"] = (
+    DATABASES["default"]["HOST"].replace("%3a", ":").replace("%3A", ":")
+)
+
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
 TIME_ZONE = "UTC"


saleor/settings.py
```diff
@@ -72,6 +72,10 @@ def get_bool_from_env(name, default_value):
     )
 }
 
+DATABASES["default"]["HOST"] = (
+    DATABASES["default"]["HOST"].replace("%3a", ":").replace("%3A", ":")
```
patrys, line 76: Does it have to be done by `replace` here or could we use `urllib.parse.unquote` instead? Are there any escape sequences we don't want to decode?

matasar: It's a good question. I initally thought that made sense too, but it seems from the conversation here that folks were worried about escaping too much: https://github.com/jacobian/dj-database-url/pull/67

matasar: That said, I'm happy to try it if you think it's better. Let me know.

patrys: Interestingly, the linked discussion points to a spec that explicitly disallows the colon from being used as part of the host. Now that I think about this, the socket path should not be a part of the hostname at all, it should look like the `file` schema:

```
cloudsql:///cloudsql/<project_id>:<region>:<instance_id>
```

(notice the three slashes, the hostname is empty and the path is used instead)

maarcingebala:
Closing due to no activity, please reopen if needed.

PR #7883, opened 2021-08-20
Fix checking existing lines quantity

CossackDex:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/checkout/mutations.py
+++ b/saleor/graphql/checkout/mutations.py
@@ -113,10 +113,10 @@ def update_checkout_shipping_method_if_invalid(
 def check_lines_quantity(
     variants,
     quantities,
-    existing_lines,
     country,
     channel_slug,
     allow_zero_quantity=False,
+    existing_lines=None,
 ):
     """Clean quantities and check if stock is sufficient for each checkout line.
 
@@ -159,7 +159,7 @@ def check_lines_quantity(
             )
     try:
         check_stock_quantity_bulk(
-            variants, country, quantities, existing_lines, channel_slug
+            variants, country, quantities, channel_slug, existing_lines
         )
     except InsufficientStock as e:
         errors = [
@@ -292,7 +292,7 @@ class CheckoutCreate(ModelMutation, I18nMixin):
         validate_variants_available_in_channel(
             variant_db_ids, channel.id, CheckoutErrorCode.UNAVAILABLE_VARIANT_IN_CHANNEL
         )
-        check_lines_quantity(variants, quantities, [], country, channel.slug)
+        check_lines_quantity(variants, quantities, country, channel.slug)
         return variants, quantities
 
     @classmethod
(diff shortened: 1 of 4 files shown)


patrys:
Task linked: [SALEOR-4178 checkoutLinesAdd - allows to add more variants than available for checkout.](https://app.clickup.com/t/2549495/SALEOR-4178)

saleor/graphql/checkout/mutations.py
```diff
@@ -440,16 +447,28 @@ class Meta:
         error_type_field = "checkout_errors"
 
     @classmethod
-    def validate_checkout_lines(cls, variants, quantities, country, channel_slug):
-        check_lines_quantity(variants, quantities, country, channel_slug)
+    def validate_checkout_lines(
```
korycins, line 450: Question -  Do we really need this method? It looks like it just run one function which can be executed inside a clean_input. IMHO I don't see any prons of having this here. 
cc @maarcingebala

saleor/warehouse/availability.py
```diff
@@ -62,7 +63,14 @@ def check_stock_quantity_bulk(
         variant_stocks[stock.product_variant_id].append(stock)
 
     insufficient_stocks: List[InsufficientStockData] = []
-    for variant, quantity in zip(variants, quantities):
+    for variant, input_quantity in zip(variants, quantities):
+
+        quantity = input_quantity
```
korycins, line 68: Maybe I am missing something but is there any prons to have two quantity variables here? `quantity` and `input_quantity` ? Looks like this assignment is not needed and we can take from `for` a normal `quantity` (as previously)

saleor/warehouse/availability.py
```diff
@@ -62,7 +63,14 @@ def check_stock_quantity_bulk(
...
+        if existing_lines:
+            for line_info in existing_lines:
```
korycins, line 70: Instead of calling `for`  for each variant, wouldn't be good to create a dict of lines (before the first for)- `{"variant.pk": "line"}` ? This should simplify the logic and complexity.

saleor/graphql/checkout/tests/test_checkout_lines.py
```diff
@@ -91,6 +91,24 @@ def test_checkout_lines_add_existing_variant(user_api_client, checkout_with_item
...
+def test_checkout_lines_add_existing_variant_over_allowed_stock(
+    user_api_client, checkout_with_item
+):
+    checkout = checkout_with_item
+    line = checkout.lines.first()
+    variant_id = graphene.Node.to_global_id("ProductVariant", line.variant.pk)
+
+    variables = {
+        "token": checkout.token,
+        "lines": [{"variantId": variant_id, "quantity": 9}],
```
maarcingebala, line 103: Some comments would be helpful somewhere in this test to describe the assumptions, e.g.:
- we have X quantity of the variant in stock
- we have Y quantity of this variant in checkout
- we try to add Z quantity to checkout, which should throw an error since Z + Y > X

Maybe instead of relying on stock quantities from fixtures, we could explicitly create a stock item at the beginning of the test, so that it never fails if someone ever changed the quantity in the fixture?

saleor/graphql/checkout/mutations.py
maarcingebala: If I'm correct we need to have control over arguments passed to `check_lines_quantity` in `CheckoutLinesUpdate` mutation that inherits from `CheckoutLinesAdd`.

korycins: Oh, right. I didn't think about this. Thanks! 👨‍🏫

PR #7884, opened 2021-08-20
Modify the checkout flow

dongduc:


maarcingebala:
Hey @dongduc, I'm closing the PR as we don't want to modify the checkout flow without a reason. Please describe what problem are you trying to solve using an issue or Github discussion and we'll decide if we need to adjust the code or it can be solved in a different way.

PR #7938, opened 2021-08-26
Fix fulfillments refund with waiting for approval state

tomaszszymanski129:
I want to merge this change because it fixes fulfillments waiting for approval not properly handled on refund option.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/mutations/fulfillments.py
+++ b/saleor/graphql/order/mutations/fulfillments.py
@@ -655,21 +655,48 @@ class FulfillmentRefundProducts(FulfillmentRefundAndReturnProductBase):
         Delete fulfillments themselves if metioned lines were only lines inside.
         """
         lines_to_remove = []
+        lines_to_update = []
         remaining_lines = []
+        quantity = 0
         for line_data in lines_info:
             if (
                 line_data.line.fulfillment.status
                 == FulfillmentStatus.WAITING_FOR_APPROVAL
             ):
-                lines_to_remove.append(line_data.line.id)
+                if line_data.quantity < line_data.line.quantity:
+                    _line = line_data.line
+                    _line.quantity = _line.quantity - line_data.quantity
+                    lines_to_update.append(_line)
+                    quantity += line_data.quantity
+                else:
+                    lines_to_remove.append(line_data.line.id)
+                    quantity += line_data.quantity
             else:
                 remaining_lines.append(line_data)
 
         lines_to_delete = order_models.FulfillmentLine.objects.filter(
             id__in=set(lines_to_remove)
         ).select_related("fulfillment")
+
         related_fulfillment_ids = [f.fulfillment.pk for f in lines_to_delete]
-        lines_to_delete.delete()
+        related_fulfillment_ids.extend([f.fulfillment.pk for f in lines_to_update])
+
+        order_line = None
+        if lines_to_update:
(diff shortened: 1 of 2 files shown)


tomaszszymanski129:
Re-worked the logic, now works also for partial refunds
- fulfillment lines are analyzed
- if line is waiting for approval, remove it and add quantity to order lines input of refund logic
- if quantity is lower than total - don't delete, lower the waiting fulfillment line quantity accordingly and add "to refund" quanity to order line as above

that way we use exactly same refund mechanism as before but treat "to refund" waiting amounts as part of order line unfulfilled quantity - waiting fulfillements in fact are unfulfilled in terms of warehouses and allocations.

korycins:
@tomaszszymanski129  Will a re-worked logic fix a case when I want to proceed with a refund and I see in the refund tab, only unfulfilled section, no matter how many fulfillments `waiting for approving` I have?

tomaszszymanski129:
@korycins frontend is now adjusted to display waiting ones separately, you can easily decide what exactly you wish to refund. Yes, works also for a mix of waiting / unfulfilled.

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -643,21 +643,105 @@ def clean_input(cls, info, order_id, input):
...
+                    )
+                )
+        return order_lines
+
     @classmethod
     def perform_mutation(cls, _root, info, **data):
         cleaned_input = cls.clean_input(info, data.get("order"), data.get("input"))
         order = cleaned_input["order"]
+
+        order_lines_map, remaining_fulfillment_lines = cls.remove_waiting_fulfillments(
```
korycins, line 730: Not sure of this (or maybe I am missing something).  Why do we have to do that - `cls.remove_waiting_fulfillments` and `cls.extend_order_lines`?
Wouldn't be good to treat fulfillment `Waiting for Approving` as a fulfillment without allocated stock, and extend a logic inside `create_refund_fulfillment` to properly handle it (especially `saleor.order.actions._move_fulfillment_lines_to_target_fulfillment`)?  
It should work by default.

tomaszszymanski129: That was my frist approach but there is already way more complicated logic there, which also touches stocks and stuff based on quantity allocated (which doesn't include waiting ones), it was hard to get it to even work and calculate things properly, plus it made a mess with extra `if`s in already complicated places, thats why it's now more "chained".

korycins: I understand that we don't want to allocate lines from `waiting for approving`, but is it OK that we don't update quantity_fulfilled after we move lines to the fulfillment - `waiting for approving`?
Personally, I would try to extend `_move_fulfillment_lines_to_target_fulfillment` even if this will include some ifs or complicates the logic.  This should also solve the issue with return/replace case. In your case `return/replace` will also require such additional actions. For me, it looks like we change the data in order and in fulfillments to fit in old logic for refunds. 
Also, the assumption of `create_refund_fulfillment` is that you can call it and it will cover for you all things related to refund, but it will not work well with the new fulfillment type.

maarcingebala (approved):
I think this code should be refactored as @korycins suggests, but for now we can merge it, as it needs to be released soon.

PR #7954, opened 2021-08-27
[SALEOR-3674] Add release unfinished payments feature

pzborow:

I want to call refund/void on all Payments which doesn't have an order.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/payment/utils.py
```diff
@@ -427,3 +430,49 @@ def price_to_minor_unit(value: Decimal, currency: str):
...
+    pass
+
+
+def release_payment(payment: Payment):
+    """Try release payment calling refund or void."""
+    if not is_payment_unfinished_and_ready_to_release(payment):
+        raise ReleasePaymentException(f"Payment '{payment}' is not ready to release")
+    from ..plugins.manager import get_plugins_manager
+
+    manager = get_plugins_manager()
```
maarcingebala, line 469: @korycins  Are we planning to trigger `release_payment` from other places in code as well? If so, I would recommend passing `manager` as a parameter and handle fetching it outside of the task. E.g. if this was called from a mutation, we already have the manager prefetched in the context and don't want to fetch it here.

saleor/payment/utils.py
```diff
@@ -427,3 +430,49 @@ def price_to_minor_unit(value: Decimal, currency: str):
     number_places = Decimal("10.0") ** precision
     value_without_comma = value * number_places
     return str(value_without_comma.quantize(Decimal("1")))
+
+
+def get_unfinished_payments():
```
maarcingebala, line 435: This function is purely related to querying data from DB - maybe we could rewrite it as a queryset method? It's not a big difference in how the code works, but a bit better organization of the code I think.

saleor/payment/tasks.py
```diff
@@ -0,0 +1,22 @@
...
+    payments = get_unfinished_payments()
+    for payment in payments.iterator():
+        release_dangling_unfinished_payment_task.apply(payment.pk)
+
+
+@app.task
+def release_dangling_unfinished_payment_task(pk):
+    payment = Payment.objects.get(pk=pk)
+
+    release_payment(payment)
```
maarcingebala, line 21: This function throws exceptions, how do we want to handle them?

saleor/payment/tasks.py
```diff
@@ -0,0 +1,22 @@
...
+from .models import Payment
+
+
+task_logger = get_task_logger(__name__)
+
+@app.task
+def release_unfinished_payments_task():
+    payments = get_unfinished_payments()
+    for payment in payments.iterator():
+        release_dangling_unfinished_payment_task.apply(payment.pk)
```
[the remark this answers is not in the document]
pzborow: OK

saleor/payment/tests/test_payment.py
```diff
@@ -550,3 +562,129 @@ def test_update_payment(gateway_response, payment_txn_captured):
...
+    payment = payment_dummy
+    if not order_exist:
+        payment.order = None
+        payment.save()
+
+    if not is_active:
+        payment.is_active = False
+        payment.save()
+
+    for date in dates:
```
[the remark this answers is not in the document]
korycins: Okay, but something like this will not work - https://github.com/spulec/freezegun#auto_tick_seconds-argument
```
@freeze_time("Jan 14th, 2020", auto_tick_seconds=15)
def test_nice_datetime():
    first_time = datetime.datetime.now()
    auto_incremented_time = datetime.datetime.now()
    assert first_time + datetime.timedelta(seconds=15) == auto_incremented_time
```

I think that we can also skip this comment if it is more effort that benefit

saleor/payment/utils.py
korycins: I thought that we want to keep such things outside a QuerySet class.  This function will be called in one place.

saleor/payment/utils.py
korycins: I think, that we could use it in the future in more places. for now it will only be used in one place

saleor/payment/utils.py
```diff
@@ -427,3 +430,49 @@ def price_to_minor_unit(value: Decimal, currency: str):
...
+def release_payment(payment: Payment):
+    """Try release payment calling refund or void."""
+    if not is_payment_unfinished_and_ready_to_release(payment):
+        raise ReleasePaymentException(f"Payment '{payment}' is not ready to release")
+    from ..plugins.manager import get_plugins_manager
+
+    manager = get_plugins_manager()
+    from saleor.payment.gateway import payment_refund_or_void
+
+    if payment.checkout is None:
```
korycins, line 472: The payment method should have a `checkout` word inside a name as we raise validation error for payments witch doesn't have a checkout.

saleor/payment/tasks.py
pzborow: Yes, fixed

saleor/payment/tests/test_payment.py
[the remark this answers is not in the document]
pzborow: OK, I've tried different methods, also using generator with freeze_time. It wan't work.

saleor/payment/tests/test_payment.py
```diff
@@ -550,3 +562,129 @@ def test_update_payment(gateway_response, payment_txn_captured):
...
+    has_checkout,
+):
+    payment = payment_dummy
+    if has_checkout:
+        payment.checkout = checkout_with_item
+        payment.save()
+
+    is_payment_unfinished_and_ready_to_release.return_value = should_be_released
+
+    if exepected_release_was_called:
```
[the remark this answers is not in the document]
pzborow: OK, Splited

saleor/payment/utils.py
pzborow: The first version was included in queryseet. Now i get out it into separete function. So how it should looks like?

saleor/payment/utils.py
```diff
@@ -427,3 +430,49 @@ def price_to_minor_unit(value: Decimal, currency: str):
...
     return str(value_without_comma.quantize(Decimal("1")))
+
+
+def get_unfinished_payments():
+    """Return payments older than specified time and without assigned order object."""
+    ttl = settings.UNFINISHED_PAYMENT_TTL
+    now = timezone.now()
+    day_before = now - ttl
+
+    newest = Transaction.objects.filter(payment=OuterRef("pk")).order_by("-created")
```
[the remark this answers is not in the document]
pzborow: OK, done with tests

saleor/payment/utils.py
```diff
@@ -427,3 +430,49 @@ def price_to_minor_unit(value: Decimal, currency: str):
...
+    """Exception occured on attempt to release payment."""
+
+    pass
+
+
+def release_payment(payment: Payment):
+    """Try release payment calling refund or void."""
+    if not is_payment_unfinished_and_ready_to_release(payment):
+        raise ReleasePaymentException(f"Payment '{payment}' is not ready to release")
+    from ..plugins.manager import get_plugins_manager
```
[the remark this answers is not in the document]
pzborow: OK, moved to global import

saleor/payment/utils.py
```diff
@@ -427,3 +430,49 @@ def price_to_minor_unit(value: Decimal, currency: str):
...
+    return get_unfinished_payments().filter(pk=payment.pk).exists()
+
+
+class ReleasePaymentException(Exception):
+    """Exception occured on attempt to release payment."""
+
+    pass
+
+
+def release_payment(payment: Payment):
```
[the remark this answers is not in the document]
pzborow: I agree, there are a least two options: Remove `is_payment_unfinished_and_ready_to_release` and depens on queryset caled in release_unfinished_payments_task. The second one is to repeat databases' sql logic and repeat in python way.

saleor/payment/utils.py
pzborow: Right, changed.

saleor/payment/utils.py
korycins: let's leave it for now as it is. In the worst scenario, we will move it to the Payment QS in final review.

saleor/plugins/manager.py
```diff
@@ -79,8 +79,12 @@ def _load_plugin(
 
         return PluginClass(configuration=plugin_config, active=active, channel=channel)
 
+    def __eq__(self, other):
```
korycins, line 82: Why do we need this?

saleor/payment/tests/test_payment.py
```diff
@@ -550,3 +562,237 @@ def test_update_payment(gateway_response, payment_txn_captured):
...
+):
+    payment = payment_dummy
+    if has_checkout:
+        payment.checkout = checkout_with_item
+        payment.save()
+
+    is_payment_unfinished_and_ready_to_release.return_value = should_be_released
+
+    with pytest.raises(ReleasePaymentException):
+        manager = get_plugins_manager()
```
korycins, line 758: IMHO, for better readability, you should place in `with pytest.rais...` only a part of code which can raise the error. 
Manager and mock should be excluded from it.

saleor/payment/utils.py
```diff
@@ -427,3 +430,55 @@ def price_to_minor_unit(value: Decimal, currency: str):
...
+
+class ReleasePaymentException(Exception):
+    """Exception occured on attempt to release payment."""
+
+    pass
+
+
+def release_checkout_payment(payment: Payment, manager: "PluginsManager"):
+    """Try release payment calling refund or void."""
+    if not is_payment_unfinished_and_ready_to_release(payment):
```
korycins, line 473: How many db queries this function will call? 
Also, I think that a raise msg is not correct.  Payment can not be ready to release, but it can be already relesed, or it can be assigned to Order.

saleor/plugins/manager.py
pzborow: Because test `test_release_dangling_unfinished_payment_task` faile on `release_checkout_payment.assert_called_once_with(payment_dummy, manager)`

saleor/payment/tests/test_payment.py
pzborow: OK

saleor/payment/utils.py
pzborow: 1. Jedno zapytanie per `payment` It calls `get_unfinished_payments` which create queryset and filter by payment.pk
```
SELECT ...
  (SELECT U0."created"
   FROM "payment_transaction" U0
   WHERE (U0."payment_id" = "payment_payment"."id"
          AND NOT U0."action_required"
          AND U0."is_success"
          AND U0."kind" IN (auth,
                            capture))
   ORDER BY U0."created" DESC
   LIMIT 1) AS "newest_trx_date"
FROM "payment_payment"
WHERE ("payment_payment"."is_active"
       AND "payment_payment"."order_id" IS NULL
       AND
         (SELECT U0."created"
          FROM "payment_transaction" U0
          WHERE (U0."payment_id" = "payment_payment"."id"
                 AND NOT U0."action_required"
                 AND U0."is_success"
                 AND U0."kind" IN (auth,
                                   capture))
          ORDER BY U0."created" DESC
          LIMIT 1) <= 2021-09-05 10:43:31.614135+00:00)
ORDER BY "payment_payment"."id" ASC
```

pzborow: 2. As above. It calls `get_unfinished_payments` which consider only payments being active and without order. See
```
.filter(is_active=True, order=None)
```

saleor/plugins/manager.py
korycins: Hhm... I don't think that this condition is correct. It will always return a true for managers with the same plugins, but the configuration of them can be different, right?  Manager stores plugins per channel, so one manager can have active two plugins in 3 channels, and second manager has two plugins in 1 channel, and this condition will return true. 
Also, I am not convinced that we should modify the base code, just for unittests.  I think that for your case you should mock a given action on the manager and then check it was called with proper values.  Something like this:
`@patch("saleor.plugins.manager.PluginsManager.notify")`. If payment should be refunded then you mock refund action, for void you mock manager.void action. You can also check if the function was called with an instance of PluginManager.

saleor/payment/utils.py
stnatic: Wouldn't it be enough to do:

```
Payment.objects.filter(
    is_active=True,
    charge_status__in=[AUTHORIZED, CAPTURED],
    modified__lt=now() - TTL,
)
```

`gateway_postprocess` is already denormalizing results of `Transaction`s for us and storing it in a `Payment` object

saleor/plugins/manager.py
pzborow: OK, I've removed changes in manager and use ANY. `release_checkout_payment.assert_called_once_with(payment_dummy, ANY)`

saleor/payment/utils.py
korycins: We can use a `payment.modified` but we need to make sure that we update it in all places where we operate on payment. 
Also, the proposed query probably needs to be modified to handle a Payment which we tried to refund/void but without any success.

korycins (approved):

saleor/payment/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+task_logger = get_task_logger(__name__)
+
+
+@app.task
+def release_unfinished_payments_task():
+    payments = get_unfinished_payments()
+
+    for payment in payments.iterator():
+        payment.is_active = False
+        payment.save(update_fields=["is_active", "modified"])
```
IKarbowiak, line 19: We should change `is_active` value only when `payment` is active.

saleor/payment/utils.py
```diff
@@ -432,3 +435,38 @@ def price_to_minor_unit(value: Decimal, currency: str):
...
+                TransactionKind.REFUND,
+            ]
+        )
+    )
+
+    return payments
+
+
+class ReleasePaymentException(Exception):
+    """Exception occured on attempt to release payment."""
```
IKarbowiak, line 472: It looks like an unfinished Exception. Also no blank line at the end.

saleor/payment/tasks.py
stnatic: `get_unfinished_payments` filters by `is_active=True`

saleor/payment/utils.py
stnatic: Good catch! It should have been the `PaymentError` that's raised by gateways

IKarbowiak (approved):

PR #7975, opened 2021-08-30
Allow using shipping methods from plugins.

bogdal:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/checkout/mutations.py
+++ b/saleor/graphql/checkout/mutations.py
@@ -26,6 +26,7 @@ from ...checkout.utils import (
     is_shipping_required,
     recalculate_checkout_discount,
     remove_promo_code_from_checkout,
+    set_app_shipping_id,
     validate_variants_in_checkout_lines,
 )
 from ...core import analytics
@@ -33,6 +34,7 @@ from ...core.exceptions import InsufficientStock, PermissionDenied, ProductNotPu
 from ...core.tracing import traced_atomic_transaction
 from ...core.transactions import transaction_with_commit_on_errors
 from ...order import models as order_models
+from ...plugins.webhook.utils import from_shipping_app_id
 from ...product import models as product_models
 from ...product.models import ProductChannelListing
 from ...shipping import models as shipping_models
@@ -995,33 +997,38 @@ class CheckoutShippingMethodUpdate(BaseMutation):
                 }
             )
 
-        shipping_method = cls.get_node_or_error(
-            info,
-            shipping_method_id,
-            only_type=ShippingMethod,
-            field="shipping_method_id",
-            qs=shipping_models.ShippingMethod.objects.prefetch_related(
-                "postal_code_rules"
-            ),
-        )
-
-        shipping_method_is_valid = clean_shipping_method(
-            checkout_info=checkout_info,
-            lines=lines,
-            method=shipping_method,
-        )
-        if not shipping_method_is_valid:
-            raise ValidationError(
-                {
-                    "shipping_method": ValidationError(
-                        "This shipping method is not applicable.",
-                        code=CheckoutErrorCode.SHIPPING_METHOD_NOT_APPLICABLE,
-                    )
-                }
+        app_shipping_id = from_shipping_app_id(shipping_method_id)
+        if app_shipping_id:
+            set_app_shipping_id(checkout=checkout, app_shipping_id=shipping_method_id)
+            checkout.save(update_fields=["private_metadata", "last_change"])
+        else:
+            shipping_method = cls.get_node_or_error(
+                info,
+                shipping_method_id,
+                only_type=ShippingMethod,
+                field="shipping_method_id",
+                qs=shipping_models.ShippingMethod.objects.prefetch_related(
+                    "postal_code_rules"
+                ),
             )
 
-        checkout.shipping_method = shipping_method
-        checkout.save(update_fields=["shipping_method", "last_change"])
+            shipping_method_is_valid = clean_shipping_method(
+                checkout_info=checkout_info,
+                lines=lines,
+                method=shipping_method,
+            )
+            if not shipping_method_is_valid:
+                raise ValidationError(
+                    {
+                        "shipping_method": ValidationError(
+                            "This shipping method is not applicable.",
+                            code=CheckoutErrorCode.SHIPPING_METHOD_NOT_APPLICABLE,
+                        )
+                    }
+                )
+
+            checkout.shipping_method = shipping_method
+            checkout.save(update_fields=["shipping_method", "last_change"])
         recalculate_checkout_discount(
             manager, checkout_info, lines, info.context.discounts
         )
--- a/saleor/graphql/shipping/types.py
+++ b/saleor/graphql/shipping/types.py
@@ -12,9 +12,11 @@ from ..channel.types import (
     ChannelContext,
     ChannelContextType,
     ChannelContextTypeWithMetadata,
+    ChannelContextTypeWithMetadataForObjectType,
 )
 from ..core.connection import CountableDjangoObjectType
 from ..core.fields import ChannelContextFilterConnectionField
+from ..core.scalars import WeightScalar
 from ..core.types import CountryDisplay, Money, MoneyRange
 from ..decorators import permission_required
 from ..meta.types import ObjectWithMetadata
@@ -69,7 +71,7 @@ class ShippingMethodPostalCodeRule(CountableDjangoObjectType):
         ]
 
 
-class ShippingMethod(ChannelContextTypeWithMetadata, CountableDjangoObjectType):
+class ShippingMethod(ChannelContextTypeWithMetadataForObjectType):
     type = ShippingMethodTypeEnum(description="Type of the shipping method.")
     translation = TranslationField(
         ShippingMethodTranslation,
@@ -100,23 +102,36 @@ class ShippingMethod(ChannelContextTypeWithMetadata, CountableDjangoObjectType):
         description="List of excluded products for the shipping method.",
     )
 
+    id = graphene.ID(required=True, description="Shipping method ID.")
+    name = graphene.String(required=True, description="Shipping method name.")
+    description = graphene.JSONString(description="Shipping method description (JSON).")
+    minimum_order_weight = WeightScalar(
+        description="Minimum order weight to use this shipping method."
+    )
+    maximum_order_weight = WeightScalar(
+        description="Maximum order weight to use this shipping method."
+    )
+    maximum_delivery_days = graphene.Int(
+        description="Maximum number of days for delivery."
+    )
+    minimum_delivery_days = graphene.Int(
+        description="Minimal number of days for delivery."
+    )
+
     class Meta:
         default_resolver = ChannelContextType.resolver_with_context
         description = (
             "Shipping method are the methods you'll use to get customer's orders to "
             "them. They are directly exposed to the customers."
         )
-        model = models.ShippingMethod
         interfaces = [relay.Node, ObjectWithMetadata]
-        only_fields = [
-            "id",
-            "maximum_order_weight",
-            "minimum_order_weight",
-            "maximum_delivery_days",
-            "minimum_delivery_days",
-            "name",
-            "description",
-        ]
+
+    @staticmethod
+    def resolve_id(root: ChannelContext[models.ShippingMethod], info, **_kwargs):
+        method_id = getattr(root.node, "pk", None)
+        if method_id:
+            return graphene.Node.to_global_id("ShippingMethod", root.node.pk)
+        return root.node.id
 
     @staticmethod
     def resolve_price(root: ChannelContext[models.ShippingMethod], info, **_kwargs):
--- a/saleor/plugins/manager.py
+++ b/saleor/plugins/manager.py
@@ -775,6 +775,41 @@ class PluginsManager(PaymentInterface):
             )
         return gateways
 
+    def list_shipping_methods(
+        self,
+        checkout: Optional["Checkout"] = None,
+        channel_slug: Optional[str] = None,
+        active_only: bool = True,
+    ) -> List["ShippingMethod"]:
+        channel_slug = checkout.channel.slug if checkout else channel_slug
+        plugins = self.get_plugins(channel_slug=channel_slug, active_only=active_only)
+        shipping_plugins = [
+            plugin
+            for plugin in plugins
+            if "get_shipping_methods" in type(plugin).__dict__
+        ]
+
+        shipping_methods = []
+        for plugin in shipping_plugins:
+            shipping_methods.extend(
+                plugin.get_shipping_methods(checkout=checkout, previous_value=None)
+            )
+        return shipping_methods
+
+    def get_shipping_method(
+        self,
+        shipping_method_id: str,
+        checkout: Optional["Checkout"] = None,
+        channel_slug: Optional[str] = None,
+    ):
+        methods = {
+            method.id: method
+            for method in self.list_shipping_methods(
+                checkout=checkout, channel_slug=channel_slug
+            )
+        }
+        return methods.get(shipping_method_id)
+
     def list_external_authentications(self, active_only: bool = True) -> List[dict]:
         auth_basic_method = "external_obtain_access_tokens"
         plugins = self.get_plugins(active_only=active_only)
--- a/saleor/plugins/webhook/plugin.py
+++ b/saleor/plugins/webhook/plugin.py
@@ -18,6 +18,7 @@ from ...webhook.payloads import (
     generate_product_deleted_payload,
     generate_product_payload,
     generate_product_variant_payload,
+    generate_shipping_methods_payload,
     generate_translation_payload,
 )
 from ..base_plugin import BasePlugin
@@ -25,6 +26,7 @@ from .tasks import trigger_webhook_sync, trigger_webhooks_for_event
 from .utils import (
     from_payment_app_id,
     parse_list_payment_gateways_response,
+    parse_list_shipping_methods_response,
     parse_payment_action_response,
 )
 
@@ -393,3 +395,23 @@ class WebhookPlugin(BasePlugin):
             previous_value,
             **kwargs,
         )
+
+    def get_shipping_methods(
+        self, checkout: Optional["Checkout"], previous_value, **kwargs
+    ) -> List["ShippingMethod"]:
+        methods = []
+        apps = App.objects.for_event_type(
+            WebhookEventType.SHIPPING_LIST_METHODS
+        ).prefetch_related("webhooks")
+        for app in apps:
+            response_data = trigger_webhook_sync(
+                event_type=WebhookEventType.SHIPPING_LIST_METHODS,
+                data=generate_shipping_methods_payload(checkout),
+                app=app,
+            )
+            if response_data:
+                shipping_methods = parse_list_shipping_methods_response(
+                    response_data, app
+                )
+                methods.extend(shipping_methods)
+        return methods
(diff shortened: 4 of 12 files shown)


saleor/graphql/core/mutations.py
```diff
@@ -495,10 +502,13 @@ def get_instance(cls, info, **data):
         object_id = data.get("id")
         qs = data.get("qs")
         if object_id:
-            model_type = cls.get_type_for_model()
-            instance = cls.get_node_or_error(
-                info, object_id, only_type=model_type, qs=qs
-            )
+            if cls._meta.model._meta.object_name == "ShippingMethod":
```
korycins, line 505: I am not convinced that specific `ShippingMethod` cases should be placed in base `Core` mutation. IMHO, get_instance should be overwrite for any non-standard cases (like ShippingMethod) but inside a mutation dedicated to this non-standard case, not inside a Base mutation.

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -103,6 +104,10 @@ def clean_input(cls, info, instance, data):
         redirect_url = data.pop("redirect_url", None)
         channel_id = data.pop("channel_id", None)
 
+        shipping_method = get_shipping_model_by_object_id(
```
korycins, line 107: IMHO, the name can be incorrect 🤔 .  This function returns an instance of ShippingMethod (if exists). Is it something dedicated for EP? I don't see any handler for shipping method from EP. Also if this case is specific only for DraftorderUpdate it should be placed there, not in DraftOrderCreate

saleor/graphql/order/types.py
```diff
@@ -9,6 +9,8 @@
 from graphene import relay
 from promise import Promise
 
+from saleor.checkout.utils import get_app_shipping_id
```
korycins, line 12: Please use relative imports as we use them everywhere

saleor/graphql/shipping/mutations/channels.py
```diff
@@ -232,9 +233,7 @@ def clean_add_channels(cls, shipping_method, input):
 
     @classmethod
     def perform_mutation(cls, _root, info, id, input):
-        shipping_method = cls.get_node_or_error(
-            info, id, only_type=ShippingMethod, field="id"
-        )
+        shipping_method = get_shipping_model_by_object_id(id)
```
korycins, line 236: This is mutation for managing the shipping_method. It is fine if this part of code will raise an error (it is even expected with inccorect ID). IMHO this part of code should work as previously - If shipping_method doesn't exist then we raise an error. If there is an issue with relation ShippingMethod type to ShippingMethod model, then I would overwrite the mutation method to handle such case. 
Currently we change the behaviour of Saleor, as your function can return None here, which were never happen before.

saleor/graphql/shipping/mutations/shippings.py
```diff
@@ -493,9 +494,7 @@ class Meta:
 
     @classmethod
     def perform_mutation(cls, _root, info, **data):
-        shipping_method = cls.get_node_or_error(
-            info, data.get("id"), only_type=ShippingMethod
-        )
+        shipping_method = get_shipping_model_by_object_id(data.get("id"))
```
korycins, line 497: as above ☝️  but this is better example.
```
def get_shipping_model_by_object_id(
    object_id: Optional[str],
) -> Optional[ShippingMethod]:
    if object_id:
        _, object_pk = from_global_id_or_error(object_id)
        return ShippingMethod.objects.filter(pk=object_pk).first()
    return None
...
497: shipping_method is None
498: shipping_method_id = shipping_method.id <-- shipping_method is None so it will raise an unexpected error. 
```

saleor/graphql/shipping/types.py
```diff
@@ -99,30 +100,42 @@ class ShippingMethod(ChannelContextTypeWithMetadata, CountableDjangoObjectType):
...
+    minimum_delivery_days = graphene.Int(
+        description="Minimal number of days for delivery."
+    )
 
     class Meta:
         default_resolver = ChannelContextType.resolver_with_context
         description = (
-            "Shipping method are the methods you'll use to get customer's orders to "
-            "them. They are directly exposed to the customers."
+            "External shipping method are the methods you'll use to get customer's "
```
korycins, line 122: Why `external`?  From the customer perspective/storefront perspective, it doesn't matter if this is `External` or `built-in`. It expects the list of shipping methods.

saleor/graphql/shipping/types.py
```diff
@@ -99,30 +100,42 @@ class ShippingMethod(ChannelContextTypeWithMetadata, CountableDjangoObjectType):
         "saleor.graphql.product.types.products.Product",
         description="List of excluded products for the shipping method.",
     )
+    id = graphene.ID(required=True, description="Shipping method ID.")
+    name = graphene.String(required=True, description="Shipping method name.")
+    description = graphene.JSONString(description="Shipping method description (JSON).")
```
korycins, line 105: I don't think that we need to add `(JSON)` here. We use `JSONString` type which informs you that this is a JSON field

saleor/plugins/manager.py
```diff
@@ -797,6 +798,41 @@ def list_payment_gateways(
...
+        ]
+
+        shipping_methods = []
+        for plugin in shipping_plugins:
+            shipping_methods.extend(
+                plugin.get_shipping_methods(checkout=checkout, previous_value=None)
+            )
+        return shipping_methods
+
+    def get_shipping_method(
```
korycins, line 822: I am not conviced that this is a proper way for handling this. 
Assuming that you have 3 external shipping methods provider. To fetch single shipping method, you will have to fetch **all** shipping methods for all shipping providers. 
Instead of calling all shipping methods, we should call a single Plugin/Webhook, to retrieve it.
We have a shipping_method_id, it shoud be parsable and should contain enough info to determine which Plugin/App should run this action.
cc @maarcingebala  what do you think? You're more familiar with sync webhook for payments and I believe the approach should be similar here.

saleor/plugins/manager.py
```diff
@@ -797,6 +798,41 @@ def list_payment_gateways(
             )
         return gateways
 
+    def list_shipping_methods(
```
korycins, line 801: IMHO we should have separate methods for `list_shipping_methods_for_order` and `list_shipping_methods_for_checkout`. The same thing for get_shipping_method (if `checkout` there is really important)

saleor/plugins/webhook/plugin.py
```diff
@@ -429,3 +432,23 @@ def process_payment(
...
+        self, checkout: Optional["Checkout"], previous_value, **kwargs
+    ) -> List["ExternalShippingMethod"]:
+        methods = []
+        apps = App.objects.for_event_type(
+            WebhookEventType.SHIPPING_LIST_METHODS
+        ).prefetch_related("webhooks")
+        for app in apps:
+            response_data = trigger_webhook_sync(
+                event_type=WebhookEventType.SHIPPING_LIST_METHODS,
+                data=generate_shipping_methods_payload(checkout),
```
korycins, line 446: It is the same checkout, I don't think that we need to re-generate payload  for each app call

saleor/graphql/checkout/mutations.py
```diff
@@ -36,8 +37,10 @@
 from ...core.tracing import traced_atomic_transaction
 from ...core.transactions import transaction_with_commit_on_errors
 from ...order import models as order_models
+from ...plugins.webhook.utils import from_shipping_app_id
```
korycins, line 40: We shouldn't have any depend relation between Saleor and Plugin. Saleor should not use any plugin's logic.

saleor/webhook/payloads.py
```diff
@@ -581,6 +581,39 @@ def generate_list_gateways_payload(
     return json.dumps(payload)
 
 
+def generate_shipping_methods_payload(checkout: Optional["Checkout"]):
```
korycins, line 584: Can't we use logic which generates payload for order and checkout? I think that it is better to extend them than adding new function for generating payload. Espeicially that we will need two functions. for order and for checkout

korycins:
@bogdal  Looks like, you didn't rebuild the schema. (Failing linters on CI).
`npm run build-schema`

saleor/graphql/order/mutations/draft_orders.py
bogdal: This logic is for both DraftorderUpdate and DraftOrderCreate mutations and because we don't have a connection between the ShippingMethod model and the type class we cannot use the build-in `get_node_or_error` function.

saleor/graphql/shipping/mutations/shippings.py
bogdal: This function throws an exception by default now.

saleor/graphql/shipping/types.py
bogdal: Good catch. Previously I have used this type only for external methods.

saleor/checkout/fetch.py
```diff
@@ -120,6 +128,26 @@ def update_channel_listings(self, checkout_info: "CheckoutInfo") -> None:
         )
 
 
+@dataclass(frozen=True)
+class ExternalShippingMethodInfo(DeliveryMethodBase):
```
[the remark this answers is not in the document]
bogdal: I would keep this class for now. `ExternalShippingMethodInfo` has some differences compared to `ShippingMethodInfo` e.g.: `update_channel_listings` and `order_key`. A single class will require multiple `if`s inside.

saleor/graphql/order/mutations/draft_orders.py
korycins: Instead of adding a special function for handling this case, what do you think about using `get_node_or_error` but with a QS from ShippingMethod?
`get_node_or_error(id, ..., qs=ShippingMethod.objects)` 
@maarcingebala  what do you think?

saleor/checkout/fetch.py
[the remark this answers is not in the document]
korycins: I mean that we could have a single type for shipping method, which will behave in the same manner for built-in and external shipping methods. The one thing which will need to be changed in that case would be separating `models.ShippingMethod` from `ShippingMethodInfo.delivery_method`, and converting `models.ShippingMethod` into a new dataclass which will work with external and built-in shipping methods. Then there is no need to store `ShippingMethodInfo.shipping_method_channel_listings` (it is used only to fetch shipping method's price) as we could use price saved in the dataclass. 
I will not insist on this change but future PRs on changes that we're working on will have to remove this type and replace with mentioned dataclass.
Also, @maarcingebala It would be good to have your opinion here.

saleor/graphql/shipping/utils.py
```diff
@@ -0,0 +1,32 @@
...
+                    "id": ValidationError(
+                        "Couldn't resolve to a node: %s" % object_id, code="not_found"
+                    )
+                }
+            )
+        return shipping_method
+    return None
+
+
+def get_instances_by_object_ids(object_ids: List[str]) -> List[ShippingMethod]:
```
maarcingebala, line 27: We have `get_nodes` and `resolve_global_ids_to_primary_keys`. Can't we use any of these functions instead of adding a new one?

saleor/plugins/webhook/utils.py
```diff
@@ -104,3 +128,25 @@ def parse_payment_action_response(
             "transaction_already_processed", False
         ),
     )
+
+
+def parse_list_shipping_methods_response(
```
korycins, line 133: This method is EasyPost specific case. I think that you return here all details which you received  from EP API. This is not something which we can easily extend in the future. The received response should be direclty mapped to the structure which this method returns. Currently ExternalShippingMethod contains below fields. And these fields should we look when we parse an response.  
We should not do any customization (like we do currently for name or ID, - it is fine that we add `app` to the beginning, but we shouldn't relay on custom fields). 

```
@dataclass
class ExternalShippingMethod:
    """Dataclass for storing information about a shipping method."""
    id: str
    name: str
    price: Money
```
See `saleor.plugins.webhook.utils.parse_payment_action_response` as an example of mapping. We map all fields from the response 1:1 to our dataclass. 
In your case EP should return a response aligned to Saleor's generic response structure, not an EP response which Saleor will map to own structure. In that case, we can't easily integrate any different type of Shipping provider because as a response we expect that we will get EP response.

saleor/webhook/event_types.py
```diff
@@ -65,6 +66,8 @@ class WebhookEventType:
     PAYMENT_CONFIRM = "payment_confirm"
     PAYMENT_PROCESS = "payment_process"
 
+    SHIPPING_LIST_METHODS = "shipping_list_methods"
```
korycins, line 69: As we should also have a webhook for the order shipping method, I would suggest renaming this webhook to
`SHIPPING_LIST_METHODS_FOR_CHECKOUT` and `SHIPPING_LIST_METHODS_FOR_ORDER`. We want to have a possibility to integrate shipping providers only for checkout or only for order OR for both. In current way we will have to use a single webhook to handle two different types, which will complicate the logic as we don't do that in any place.

saleor/webhook/payloads.py
```diff
@@ -311,11 +311,19 @@ def generate_checkout_payload(checkout: "Checkout"):
...
         "private_metadata",
         "metadata",
+        "channel",
     )
     user_fields = ("email", "first_name", "last_name")
     shipping_method_fields = ("name", "type", "currency", "price_amount")
     lines_dict_data = serialize_checkout_lines(checkout)
 
+    # todo use the most appropriate warehouse
+    warehouse = None
```
korycins, line 321: This should be changed when we merge the checkout reservation feature - https://github.com/mirumee/saleor/pull/7589

saleor/graphql/shipping/types.py
```diff
@@ -99,30 +100,42 @@ class ShippingMethod(ChannelContextTypeWithMetadata, CountableDjangoObjectType):
...
+        if method_id:
+            return graphene.Node.to_global_id("ShippingMethod", root.node.pk)
+        return root.node.id
 
     @staticmethod
     def resolve_price(root: ChannelContext[models.ShippingMethod], info, **_kwargs):
         # Price field are dynamically generated in available_shipping_methods resolver
         price = getattr(root.node, "price", None)
-        if price:
+        if price is not None:
```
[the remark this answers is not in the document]
bogdal: We want to handle Money(0) value as well.

saleor/graphql/order/mutations/draft_orders.py
maarcingebala: If possible I would really try to avoid adding new functions just to get items from DB by ID, we already have at least three functions for that.

saleor/checkout/fetch.py
```diff
@@ -64,14 +71,17 @@ def get_customer_email(self) -> str:
...
+    delivery_method: Optional[Union["ShippingMethodData", "Warehouse"]] = None
     shipping_address: Optional["Address"] = None
-    order_key: str = "shipping_method"
 
     @property
     def warehouse_pk(self) -> Optional[str]:
         pass
 
+    @property
+    def delivery_method_order_field(self) -> dict:
```
L3str4nge, line 82: It could be convert into class constant field instead of creating new dict each time

saleor/checkout/fetch.py
```diff
@@ -222,13 +242,31 @@ def fetch_checkout_info(
...
-    shipping_channel_listings = ShippingMethodChannelListing.objects.filter(
-        shipping_method=shipping_method, channel=channel
-    ).first()
-    delivery_method = checkout.collection_point or shipping_method
+    if shipping_method:
+        shipping_channel_listings = ShippingMethodChannelListing.objects.filter(
+            shipping_method=shipping_method, channel=channel
+        ).first()
+        delivery_method: Optional[
+            Union["ShippingMethodData", "Warehouse"]
```
L3str4nge, line 257: I don't think we need typehint here. It's quite huge and makes this block of code hard to read

saleor/checkout/fetch.py
```diff
@@ -222,13 +242,31 @@ def fetch_checkout_info(
 ) -> CheckoutInfo:
     """Fetch checkout as CheckoutInfo object."""
 
+    from .utils import get_app_shipping_id
+
     channel = checkout.channel
     shipping_address = checkout.shipping_address
+    shipping_channel_listings = None
+
```
L3str4nge, line 250: We can move `delivery_method = checkout.collection_point` here as a default and remove further `else`

saleor/checkout/tests/test_checkout.py
```diff
@@ -1238,3 +1247,18 @@ def test_chckout_without_delivery_method_creates_empty_delivery_method(
...
+    checkout.private_metadata = initial_private_metadata
+
+    set_app_shipping_id(checkout, app_shipping_id)
+    assert PRIVATE_META_APP_SHIPPING_ID in checkout.private_metadata
+
+    shipping_id = get_app_shipping_id(checkout)
+    assert shipping_id == app_shipping_id
+
+    delete_app_shipping_id(checkout)
+    checkout.private_metadata == initial_private_metadata
```
L3str4nge, line 1264: missing `assert` here?

saleor/graphql/checkout/tests/test_checkout.py
```diff
@@ -2685,6 +2735,10 @@ def test_checkout_delivery_method_update(
 ):
     checkout = checkout_with_item_for_cc
     old_delivery_method = getattr(checkout, attribute_name)
+    shipping_method_data = delivery_method
+    if attribute_name == "shipping_method":
```
L3str4nge, line 2739: I don't like `if statements` in tests

saleor/plugins/manager.py
```diff
@@ -834,6 +835,45 @@ def list_payment_gateways(
...
+        checkout: "Checkout",
+        channel_slug: Optional[str] = None,
+        active_only: bool = True,
+    ) -> List["ShippingMethodData"]:
+        channel_slug = checkout.channel.slug if checkout else channel_slug
+        plugins = self.get_plugins(channel_slug=channel_slug, active_only=active_only)
+        shipping_plugins = [
+            plugin
+            for plugin in plugins
+            if "get_shipping_methods_for_checkout" in type(plugin).__dict__
```
L3str4nge, line 849: Does the `hasattr` not fit here?

saleor/plugins/webhook/plugin.py
```diff
@@ -486,3 +488,24 @@ def process_payment(
...
+    def get_shipping_methods_for_checkout(
+        self, checkout: "Checkout", previous_value, **kwargs
+    ) -> List["ShippingMethodData"]:
+        methods = []
+        apps = App.objects.for_event_type(
+            WebhookEventType.SHIPPING_LIST_METHODS_FOR_CHECKOUT
+        ).prefetch_related("webhooks")
+        payload = generate_checkout_payload(checkout)
+        for app in apps:
+            response_data = trigger_webhook_sync(
```
L3str4nge, line 501: Will the triggering webhook sync for multiple apps be a bottleneck?

saleor/plugins/webhook/utils.py
```diff
@@ -38,6 +47,18 @@ def from_payment_app_id(app_gateway_id: str) -> Optional["PaymentAppData"]:
     return None
 
 
+def from_shipping_app_id(app_shipping_method_id: str) -> Optional["ShippingAppData"]:
+    splitted_id = app_shipping_method_id.split(":")
+    if len(splitted_id) == 3 and splitted_id[0] == APP_ID_PREFIX and all(splitted_id):
```
L3str4nge, line 52: Instead of repeating this if statement we could create function for that. In any case it will required only one place to change - easier to maintain in the future :)

saleor/plugins/manager.py
korycins: Also, make sure that this will work with the latest changes which we introduced to BasePlugin
https://github.com/mirumee/saleor/blob/master/saleor/plugins/base_plugin.py

PR #7992, opened 2021-09-01
Change hidden products access permissions

CossackDex:
I want to merge this change because it fixes hidden products access permissions.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/core/permissions.py
+++ b/saleor/core/permissions.py
@@ -157,3 +157,26 @@ def get_permissions_from_codenames(permission_codenames: List[str]):
         .prefetch_related("content_type")
         .order_by("codename")
     )
+
+
+def _permission_required(perms, context):
+    from saleor.account.models import User
+
+    if isinstance(context, User):
+        if context.has_perms(perms):
+            return True
+    else:
+        # for now MANAGE_STAFF permission for app is not supported
+        if AccountPermissions.MANAGE_STAFF in perms:
+            return False
+        return context.has_perms(perms)
+    return False
+
+
+def has_one_of_permissions(requestor, permissions=None):
+    if not permissions:
+        return True
+    for perm in permissions:
+        if _permission_required((perm,), requestor):
+            return True
+    return False
--- a/saleor/core/permissions.py
+++ b/saleor/core/permissions.py
@@ -1,6 +1,7 @@
 from enum import Enum
 from typing import Iterable, List
 
+from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Permission
 
 
@@ -160,8 +161,7 @@ def get_permissions_from_codenames(permission_codenames: List[str]):
 
 
 def _permission_required(perms, context):
-    from saleor.account.models import User
-
+    User = get_user_model()
     if isinstance(context, User):
         if context.has_perms(perms):
             return True
--- a/saleor/graphql/attribute/tests/queries/test_attribute_filter.py
+++ b/saleor/graphql/attribute/tests/queries/test_attribute_filter.py
@@ -554,7 +554,7 @@ def test_filter_attributes_in_category_not_published_by_staff_without_manage_pro
     )["data"]["attributes"]["edges"]
 
     # then
-    assert len(attributes) == attribute_count
+    assert len(attributes) == attribute_count - 1
 
 
 def test_filter_attributes_in_category_not_published_by_app_with_perm(
@@ -643,7 +643,7 @@ def test_filter_attributes_in_category_not_published_by_app_without_manage_produ
     )["data"]["attributes"]["edges"]
 
     # then
-    assert len(attributes) == attribute_count
+    assert len(attributes) == attribute_count - 1
 
 
 def test_filter_attributes_in_collection_invalid_category_id(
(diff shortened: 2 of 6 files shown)


patrys:
Task linked: [SALEOR-4302 All products shouldn't be visible for all staff users.](https://app.clickup.com/t/2549495/SALEOR-4302)

saleor/product/models.py
```diff
@@ -152,6 +158,13 @@ def __repr__(self) -> str:
 
 
 class ProductsQueryset(models.QuerySet):
+    ALL_PRODUCTS_PERMISSIONS = [
+        # One of listed permissions is required to see all products
+        OrderPermissions.MANAGE_ORDERS,
+        DiscountPermissions.MANAGE_DISCOUNTS,
+        ProductPermissions.MANAGE_PRODUCTS,
+    ]
```
maarcingebala, line 166:
```suggestion
    # List of permissions, where each of them allows viewing all products 
    # (including unpublished).
    ALL_PRODUCTS_PERMISSIONS = [
        OrderPermissions.MANAGE_ORDERS,
        DiscountPermissions.MANAGE_DISCOUNTS,
        ProductPermissions.MANAGE_PRODUCTS,
    ]
```

saleor/core/permissions.py
```diff
@@ -157,3 +157,26 @@ def get_permissions_from_codenames(permission_codenames: List[str]):
         .prefetch_related("content_type")
         .order_by("codename")
     )
+
+
+def _permission_required(perms, context):
+    from saleor.account.models import User
```
maarcingebala, line 163: If this import is here because of circular dependency, you can also try to import the user model using this Django helper:
```
# in the import section:
from django.contrib.auth import get_user_model

# in the function:
model = get_user_model()
```

saleor/graphql/attribute/tests/queries/test_attribute_filter.py
```diff
@@ -554,7 +554,7 @@ def test_filter_attributes_in_category_not_published_by_staff_without_manage_pro
     )["data"]["attributes"]["edges"]
 
     # then
-    assert len(attributes) == attribute_count
+    assert len(attributes) == attribute_count - 1
```
maarcingebala, line 557: We should change the way `attribute_count` is calculated. @fowczarek Do you have any idea what would be the best way to do it?

saleor/graphql/product/tests/test_product.py
```diff
@@ -10208,3 +10211,60 @@ def test_create_product_with_weight_input(
...
+    staff_api_client.user.user_permissions.remove(permission_manage_discounts)
+    response = staff_api_client.post_graphql(
+        QUERY_FETCH_ALL_PRODUCTS,
+        variables=variables,
+        permissions=(permission_manage_channels,),
+        check_no_permissions=False,
+    )
+    content = get_graphql_content(response)
+    total_count = content["data"]["products"]["totalCount"]
+    assert total_count == 2
```
IKarbowiak, line 10270: That should be 4 different tests.

maarcingebala (commented):
I don't see the changes in decorators that we talked about.

saleor/graphql/product/tests/test_product.py
maarcingebala: Yep, or one test with `parametrize`?

saleor/product/models.py
```diff
@@ -187,7 +200,8 @@ def published_with_variants(self, channel_slug: str):
         return published.filter(Exists(variants.filter(product_id=OuterRef("pk"))))
 
     def visible_to_user(self, requestor: Union["User", "App"], channel_slug: str):
-        if requestor_is_staff_member_or_app(requestor):
+
+        if has_one_of_permissions(requestor, self.ALL_PRODUCTS_PERMISSIONS):
```
fowczarek, line 204:
```suggestion
        if has_one_of_permissions(requestor, self.ALL_PRODUCTS_PERMISSIONS):
```

saleor/graphql/attribute/tests/queries/test_attribute_filter.py
fowczarek: Maybe we should introduce some helper for calculating `attribute_count` based on the products list? 
@maarcingebala  IMO we could do it in separate PR.

saleor/core/permissions.py
```diff
@@ -157,3 +157,26 @@ def get_permissions_from_codenames(permission_codenames: List[str]):
         .prefetch_related("content_type")
         .order_by("codename")
     )
+
+
+def _permission_required(perms, context):
```
korycins, line 162: I think the `context` can be an incorrect name. Based on your code I would say that this is a requestor: Union[User, App]

saleor/core/permissions.py
```diff
@@ -157,3 +157,26 @@ def get_permissions_from_codenames(permission_codenames: List[str]):
...
+    from saleor.account.models import User
+
+    if isinstance(context, User):
```
korycins, line 165: Wouldn't be better (instead of this `if/else`?:
```
if isinsance(requestor, "App") and AccountPermissions.MANAGE_STAFF in perms:
    return False
else:
    return requestor.has_perms(perms)
```

saleor/core/permissions.py
korycins: You're using this function to check single permission one by one. Maybe we could change the name `perms` to `perm` and use `requestor.has_perm` method (this will simplify and increase readability for calling this function)
`if _permission_required((perm,), requestor):` -> `if _permission_required(perm, requestor):`

maarcingebala (approved):

PR #7994, opened 2021-09-01
Update Stripe payment flow

mateuszgrzyb:

Extend payment source types to accept metadata [SALEOR-4239](https://app.clickup.com/t/wkwg51)
Update Stripe to propagate payment metadata [SALEOR-4243](https://app.clickup.com/t/wkwmyv)
Change the way of storing payment method in Stripe [SALEOR-4244](https://app.clickup.com/t/wkwrzw)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-4243 Update Stripe to propagate payment metadata](https://app.clickup.com/t/2549495/SALEOR-4243)

patrys:
Task linked: [SALEOR-4239 Extend payment source types to accept metadata](https://app.clickup.com/t/2549495/SALEOR-4239)

patrys:
Task linked: [SALEOR-4244 Change the way of storing payment method in Stripe](https://app.clickup.com/t/2549495/SALEOR-4244)

saleor/payment/gateways/stripe/plugin.py
```diff
@@ -177,13 +178,19 @@ def process_payment(
...
             setup_future_usage = data.get("setup_future_usage") if data else None
 
         off_session = data.get("off_session") if data else None
 
         payment_method_types = data.get("payment_method_types") if data else None
 
+        if not setup_future_usage:
+            store = payment_information.store_payment_method
+            if store != StorePaymentMethodEnum.NONE:
+                setup_future_usage = store.lower()
```
korycins, line 192: shouldn't be `store.value` ?

korycins: Also, I think the better way would be to map our enums to Stripe's values. Even if they're currently the same, it should be safer.

saleor/payment/gateways/stripe/plugin.py
```diff
@@ -294,8 +302,13 @@ def confirm_payment(
             kind, action_required = self._get_transaction_details_for_stripe_status(
                 payment_intent.status
             )
+
+            if payment_intent.get("setup_future_usage"):
```
korycins, line 306: Not sure if I understand. Is it required? I thought that stripe will return a payment_intent metadata so no need to copy them one more time. Am I missing something?

saleor/payment/gateways/stripe/tests/test_plugin.py
```diff
@@ -364,6 +364,160 @@ def test_process_payment_with_customer_and_future_usage(
     )
 
 
+@pytest.mark.parametrize(
+    "store_payment_method",
+    [StorePaymentMethodEnum.OFF_SESSION, StorePaymentMethodEnum.ON_SESSION],
+)
+@patch("saleor.payment.gateways.stripe.stripe_api.stripe.Customer.create")
+@patch("saleor.payment.gateways.stripe.stripe_api.stripe.PaymentIntent.create")
+def test_process_payment_with_customer_and_new_future_usage(
```
korycins, line 373: Instead of adding `new` to name I would add `depreceated` to the old tests. Currently, when we will remove the old approach we will have to remove `new` for each tests.

korycins: And the better way would a creating sepearate test file with deprecated word and place them all tests which operate on old approach.

mateuszgrzyb: There is a `saleor.payment.gateways.stripe.deprecated.tests` directory, maybe I should put all the old approach tests there?

korycins: There is a whole deprecated plugin, so I would propose not mixing it up as we want to keep tests together for each plugin.

saleor/payment/gateways/stripe/plugin.py
```diff
@@ -177,13 +188,19 @@ def process_payment(
         payment_method_id = data.get("payment_method_id") if data else None
 
         setup_future_usage = None
+        # DEPRECATED
```
korycins, line 191: Can you add here a note that this should be removed in Saleor 4.0? It will be easier to find this when we will going to remove all deprecated things for 4.0

saleor/payment/gateways/stripe/stripe_api.py
```diff
@@ -178,11 +178,18 @@ def create_payment_intent(
...
+    channel_slug: str,
+    metadata: Dict[str, str] = None,
+):
     with stripe_opentracing_trace("stripe.PaymentMethod.modify"):
         try:
             stripe.PaymentMethod.modify(
-                payment_method_id, api_key=api_key, metadata={"channel": channel_slug}
+                payment_method_id,
+                api_key=api_key,
+                metadata={**(metadata or {}), "channel": channel_slug},
```
korycins, line 192: You can drop a channel_slug from here. It is previous approach wich wasn't very well. If storefront would like to store channel slug, then it will pass it as a metadata in checkoutPaymentCreate

korycins: I think that you can also simplify this by: `metadata=metadata`. We can call this function only when we have some metadata in payment object.

saleor/payment/gateways/stripe/webhooks.py
```diff
@@ -229,6 +234,10 @@ def handle_authorized_payment_intent(
     if _channel_slug_is_different_from_payment_channel_slug(channel_slug, payment):
         return
 
+    _update_payment_method_metadata(
```
korycins, line 237: You can call `_update_payment_method_metadata` only when you have any `payment.metadata`

saleor/payment/gateways/stripe/stripe_api.py
```diff
@@ -236,6 +244,7 @@ def capture_payment_intent(
                 amount_to_capture=amount_to_capture,
                 api_key=api_key,
                 stripe_version=STRIPE_API_VERSION,
+                expand=["payment_method"],
```
korycins, line 247: Why do you need it?

saleor/payment/gateways/stripe/tests/test_plugin_deprecated.py
```diff
@@ -0,0 +1,84 @@
+from unittest.mock import Mock, patch
```
korycins, line 1: Use relative imports here instead of absolute. (we use relative almost in all places)

saleor/payment/gateways/stripe/stripe_api.py
mateuszgrzyb: Stripe normally returns PaymentIntent with PaymentMethod as its ID. To get the full StripeObject, we need to pass the expand parameter.

korycins: I understand but I don't know why you need this for the `capture payment intent` function. It is called in  `StripePlugin.capture_payment` so it seems that this change is not related to your feature

mateuszgrzyb: `capture_payment`, which calls `capture_payment_intent`, also calls `get_payment_method_details`, which assumes that `payment_method` is expanded. If `payment_method` isn't needed there, I suggest two alternative solutions: 
* check inside `get_payment_method_details` if `payment_method` is string and if so - return empty metadata
* split it to two functions: `get_payment_method_details` and `get_payment_method_details_with_metadata`

korycins (approved):

IKarbowiak (approved):

PR #8001, opened 2021-09-01
Adjust checkout flow to deal with multiple payments

stnatic:
This pull request:
- allows paymentCreate to create more than one active payment
- adds a new mutation checkoutPaymentComplete that authorises individual payments
- adjusts checkoutComplete so that it handles multiple payments

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] The changes are tested


patrys:
Task linked: [SALEOR-3992 Update CheckoutPaymentCreate  to not cancel previous assigned payments](https://app.clickup.com/t/2549495/SALEOR-3992)

saleor/checkout/complete_checkout.py
```diff
@@ -573,8 +667,24 @@ def complete_checkout(
     :raises ValidationError
     """
     checkout = checkout_info.checkout
-    channel_slug = checkout_info.channel.slug
-    payment = checkout.get_last_active_payment()
+
+    # For individual payments, checkoutComplete will process the payment,
+    # otherwise a specific paymentID has to be provided
+    incomplete_payments = [
+        p for p in get_active_payments(checkout) if not p.is_authorized
```
korycins, line 674: The condition shouldn't be if p.charge_status==NOT_CHARGED? This can return payments with statuses like capture etc. Right?

saleor/checkout/complete_checkout.py
```diff
@@ -573,8 +667,24 @@ def complete_checkout(
...
+    ]
+
+    if payment is None:
+        if len(incomplete_payments) > 1:
```
korycins, line 678: What about case:
As a customer, I've create a payment object, then changed something in checkoutLines, change addres etc, then again, I am moving to the payment step, which will create a second incomplete payment.  This condition will not allow me to finish a checkout. Am I right?

saleor/checkout/utils.py
```diff
@@ -688,3 +702,19 @@ def validate_variants_in_checkout_lines(lines: Iterable["CheckoutLineInfo"]):
...
+
+def call_payment_refund_or_void(
+    checkout_info: "CheckoutInfo",
+    payment: Optional[Payment],
+    manager: PluginsManager,
+):
+    if not payment:
+        return
+
+    if len(get_active_payments(checkout_info.checkout)) > 1:
```
korycins, line 715: Shouldn't be called or all payments?

korycins:
Question - Do we need to calculate whole checkout things when we just want to attach a payment to checkout?

saleor/checkout/utils.py
stnatic: This condition has been now changed not to cancel partial payments.

saleor/checkout/complete_checkout.py
stnatic: Creating a subsequent payment with partial=False (default) will cancel/refund/mark as inactive all previous payments.

Otherwise it's already an implementation that is aware of partial payments and will provide the necessary paymentID

saleor/checkout/complete_checkout.py
```diff
@@ -542,17 +547,105 @@ def _process_payment(
                 customer_id=customer_id,
                 store_source=store_source,
                 additional_data=payment_data,
-                channel_slug=channel_slug,
+                channel_slug=checkout_info.channel.slug,
             )
         payment.refresh_from_db()
+        payment.complete_order = complete_order
```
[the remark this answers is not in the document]
stnatic: Agreed, but in that case it feels like `get_checkout_by_token` should select it with select_for_update. Webhooks already do that

saleor/payment/migrations/0033_payment_complete_order.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.5 on 2021-08-30 08:09
```
korycins, line 1: Can we merge these migrations into one?

saleor/payment/models.py
```diff
@@ -118,6 +120,21 @@ def get_last_transaction(self):
     def get_total(self):
         return Money(self.total, self.currency)
 
+    def get_covered_amount(self):
+        """Return an amount that is covered by this payment (but not necessarily captured).
+
+        Partially refunded payments are included in the covered amount.
+        """
+        if self.charge_status == ChargeStatus.PARTIALLY_REFUNDED:
```
korycins, line 128: Wouldn't be easier:
```
if charge_status == AUTHORIZED:
    return self.total
return self.captured_amount
```
Also, we have method `get_authorized_amount`, shouldn't we use it here?

saleor/payment/utils.py
```diff
@@ -143,15 +140,12 @@ def create_payment(
...
-        "total": total,
-        "return_url": return_url,
-        "psp_reference": external_reference or "",
     }
 
-    payment, _ = Payment.objects.get_or_create(defaults=defaults, **data)
-    return payment
+    if payment_token:
+        return Payment.objects.get_or_create(token=payment_token, defaults=data)[0]
+    else:
```
korycins, line 147: `else` is not needed here. You're returning in `if` condition

saleor/tests/fixtures.py
```diff
@@ -280,6 +343,51 @@ def checkout_with_item(checkout, product):
...
+            manager=manager, checkout_info=checkout_info, lines=lines, address=address
+        )
+
+        amount = total.gross.amount / num_payments
+        charge_status = charge_status or ChargeStatus.AUTHORIZED
+        captured_amount = (
+            amount if charge_status != ChargeStatus.AUTHORIZED else Decimal("0")
+        )
+
+        for i in range(num_payments):
```
korycins, line 370: Can we do it in bulk? Looks like it should be easy

saleor/checkout/complete_checkout.py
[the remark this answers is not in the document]
stnatic: I added the lock in 1a67813. This way webhooks should no longer race condition with CheckoutComplete, e.g. for stripe we'll call `_process_payment_with_checkout` but `_get_checkout` should return None since another transaction should delete the checkout.

saleor/payment/utils.py
stnatic: I don't think I get it. This function has previously always called `get_or_create`. This caused a second 50% payment to be ignored as it was treated to be identical with the first one.

I decided to keep this function somewhat backwards compatible and update an existing payment by the token, while in most cases a new one will be created (the "else").

saleor/payment/models.py
stnatic: Re: `get_authorized_amount`. It's used once in conjunction with `get_last_payment` and my goal would be to remove both of them. I don't like `get_authorized_amount` as it refers to transactions which raises doubt whether `Payment.charge_status` and `Payment.amount` are correct and we should never allow them to be incorrect. 

Would you prefer:
```
if self.charge_status in {
    ChargeStatus.FULLY_CHARGED,
    ChargeStatus.PARTIALLY_CHARGED,
    ChargeStatus.PARTIALLY_REFUNDED
}:
    return self.captured_amount
if self.charge_status == ChargeStatus.AUTHORIZED:
    return self.total
return Decimal("0")
```
? I'm assuming that we want to return 0 for PENDING but we might want to check it when end-to-end testing

saleor/payment/migrations/0033_payment_complete_order.py
stnatic: Let's ignore and deal with it when merging `feature/partial-payments`. I'm expecting a dozen of conflicts with `3.0` / `master` which will have to be resolved anyway.

saleor/payment/gateways/stripe/tests/test_plugin.py
```diff
@@ -346,7 +345,7 @@ def test_process_payment_with_customer_and_future_usage(
         api_key=api_key,
         amount=price_to_minor_unit(payment_info.amount, payment_info.currency),
         currency=payment_info.currency,
-        capture_method=AUTOMATIC_CAPTURE_METHOD,
+        capture_method=MANUAL_CAPTURE_METHOD,
```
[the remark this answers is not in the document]
stnatic: Good catch 👍

saleor/checkout/utils.py
```diff
@@ -688,3 +702,20 @@ def validate_variants_in_checkout_lines(lines: Iterable["CheckoutLineInfo"]):
...
         )
+
+
+def call_payment_refund_or_void(
+    channel_slug: str,
+    payment: Optional[Payment],
+    manager: PluginsManager,
+    cancel_partial=False,
+):
+    if not payment:
```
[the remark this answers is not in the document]
stnatic: Would the purpose of this log be to catch any "unusual scenarios"? This particular if is reachable when calling checkoutComplete with all payments already completed.

saleor/checkout/complete_checkout.py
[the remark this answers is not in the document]
korycins: This needs to be tested well. Especially with enabled webhooks, to make sure that we don't have any deadlocks.

saleor/graphql/payment/mutations.py
```diff
@@ -76,18 +93,29 @@ class Meta:
...
                     )
                 }
             )
 
+        remaining = checkout_total.gross - get_covered_balance(checkout)
+        if amount > remaining.amount:
+            raise ValidationError(
+                {
+                    "amount": ValidationError(
+                        "Amount should not exceed checkout's total.",
```
korycins, line 113: Shouldn't be a `checkout's remaining amount`? Or something to emphasize, that exception is raised because provided amount is bigger that remaining amount for checkout

saleor/graphql/payment/mutations.py
```diff
@@ -198,6 +237,91 @@ def perform_mutation(cls, _root, info, checkout_id=None, token=None, **data):
...
+            "held until the payment is confirmed with second call of this mutation."
+        )
+        error_type_class = CheckoutError
+        error_type_field = "checkout_errors"
+
+    @classmethod
+    def perform_mutation(cls, _root, info, token, payment_id, **data):
+        tracking_code = analytics.get_client_id(info.context)
+        with transaction_with_commit_on_errors():
+            checkout = get_checkout_by_token(token)
```
korycins, line 291: Shouldn't you lock the checkout in the same way as you did for checkout_complete mutation?

saleor/payment/migrations/0033_payment_complete_order.py
korycins: 👍

saleor/payment/models.py
korycins: `saleor.payment.utils.gateway_postprocess` updates the capture_amount. afair It is done only in this place, so I believe we could relay the status &/OR capture_amount.

saleor/payment/utils.py
korycins: Maybe I wasn't clear enough. 
```
if payment_token:
    return Payment(...)
return Payment(...)
```
`else` is not needed as you have `return` inside an `if` condition. Just small cosmetic change

saleor/checkout/complete_checkout.py
```diff
@@ -626,7 +718,7 @@ def complete_checkout(
             checkout.delete()
         except InsufficientStock as e:
             release_voucher_usage(order_data)
-            gateway.payment_refund_or_void(payment, manager, channel_slug=channel_slug)
+            call_payment_refund_or_void(checkout_info.channel.slug, payment, manager)
```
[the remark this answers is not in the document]
stnatic: Can we leave it up to developer? `call_payment_refund_or_void` is implemented to be a noop for partial payments. After receiving "out of stock error" the UI could ask the customer whether to abandon the checkout and call PaymentRefund if needed.

saleor/checkout/utils.py
```diff
@@ -688,3 +702,20 @@ def validate_variants_in_checkout_lines(lines: Iterable["CheckoutLineInfo"]):
...
+        return
+
+    if payment.partial and not cancel_partial:
+        return
+
+    gateway.payment_refund_or_void(payment, manager, channel_slug=channel_slug)
+    payment.is_active = False
+    payment.save(update_fields=["is_active"])
```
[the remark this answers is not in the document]
stnatic: My goal was to target NOT_CHARGED payments with it (in case they are not processed by void/refund) in any way.

saleor/checkout/complete_checkout.py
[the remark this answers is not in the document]
korycins: It can't call payment refund mutation as it is dedicated to staff users.

saleor/checkout/utils.py
```diff
@@ -688,3 +702,20 @@ def validate_variants_in_checkout_lines(lines: Iterable["CheckoutLineInfo"]):
...
+    manager: PluginsManager,
+    cancel_partial=False,
+):
+    if not payment:
+        return
+
+    if payment.partial and not cancel_partial:
+        return
+
+    gateway.payment_refund_or_void(payment, manager, channel_slug=channel_slug)
```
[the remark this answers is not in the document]
stnatic: For next reviewers: our plan is to move the refund/void logic to celery in a separate task. This particular piece of code will only mark payments as `is_active=False`.

We're intending this flag set to False to mean: this payment does not count towards the order / checkout and should be refunded. Tasks and webhooks should keep processing the payment in order to keep its state up to date with the PSP.

saleor/checkout/checkout_cleaner.py
```diff
@@ -70,12 +75,19 @@ def clean_checkout_payment(
...
     error_code: Type[CheckoutErrorCode],
-    last_payment: Optional[payment_models.Payment],
+    current_payment: Optional[payment_models.Payment],
 ):
     clean_billing_address(checkout_info, error_code)
-    if not is_fully_paid(manager, checkout_info, lines, discounts):
-        gateway.payment_refund_or_void(
-            last_payment, manager, channel_slug=checkout_info.channel.slug
+
+    if len(get_active_payments(checkout_info.checkout)) == 0:
```
L3str4nge, line 82: Would be better to call `.count()` on db level instead of retrieving all objects and calling `len`

saleor/graphql/payment/mutations.py
```diff
@@ -198,6 +237,96 @@ def perform_mutation(cls, _root, info, checkout_id=None, token=None, **data):
...
+                "URL of a view where users should be redirected to "
+                "see the order details. URL in RFC 1808 format."
+            ),
+        )
+
+    class Meta:
+        description = (
+            f"{ADDED_IN_31} Completes an individual payment as part of the checkout. "
+            "This mutation does not create the order and any webhooks related "
+            "to this payment will neither create the order."
```
IKarbowiak, line 279:
```suggestion
            "to this payment will neither create the order. "
```

saleor/graphql/payment/mutations.py
```diff
@@ -198,6 +237,96 @@ def perform_mutation(cls, _root, info, checkout_id=None, token=None, **data):
...
+            "In case an additional confirmation step such as 3D secure is required "
+            "confirmationNeeded flag will be set to True and no authorization will be "
+            "held until the payment is confirmed with second call of this mutation."
+        )
+        error_type_class = CheckoutError
+        error_type_field = "checkout_errors"
```
IKarbowiak, line 285: The `error_type_field` is not needed. It adds a deprecated error field.

saleor/graphql/payment/types.py
```diff
@@ -75,6 +76,12 @@ class Payment(CountableDjangoObjectType):
         ),
         required=True,
     )
+    partial = graphene.Boolean(
+        description=(
+            f"{ADDED_IN_31} Indicates whether this payment will"
```
IKarbowiak, line 81:
```suggestion
            f"{ADDED_IN_31} Indicates whether this payment will "
```

saleor/payment/migrations/0034_payment_partial.py
```diff
@@ -0,0 +1,18 @@
...
+        ("payment", "0033_payment_complete_order"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="payment",
+            name="partial",
+            field=models.BooleanField(default=False),
+        ),
+    ]
```
IKarbowiak, line 18: You can merge this migration with the previous one.

saleor/checkout/complete_checkout.py
```diff
@@ -545,14 +550,93 @@ def _process_payment(
...
+    by this function as part of checkoutPaymentComplete mutation.
+    """
+    customer_id = None
+    if not payment:
+        return {}, False
+
+    if user.is_authenticated:
+        customer_id = fetch_customer_id(user=user, gateway=payment.gateway)
+
+    txn = _process_payment(
```
L3str4nge, line 587: What does `txn` mean?

IKarbowiak: It's a shortcut for transaction.

saleor/checkout/complete_checkout.py
```diff
@@ -545,14 +550,93 @@ def _process_payment(
...
+        payment,
+        manager,
+        payment_data,
+        store_source,
+        user,
+        order_data=None,
+        complete_order=False,
+    )
+
+    return action_required, action_data
```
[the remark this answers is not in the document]
stnatic: I'd personally keep it as it's the public interface that's called directly in the mutation and `_complete_checkout_payment` is the private piece of logic that is called in `complete_checkout` as well.

saleor/checkout/tests/test_checkout.py
```diff
@@ -1183,27 +1169,14 @@ def test_is_fully_paid_partially_paid(checkout_with_item, payment_dummy):
...
     checkout = checkout_with_item
     manager = get_plugins_manager()
     lines = fetch_checkout_lines(checkout)
     checkout_info = fetch_checkout_info(checkout, lines, [], manager)
-    is_paid = is_fully_paid(manager, checkout_info, lines, None)
+    is_paid = is_fully_covered(manager, checkout_info, lines, None)
     assert not is_paid
-
-
-def test_cancel_active_payments(checkout_with_payments):
```
L3str4nge, line 1199: Why is this test removed?  Does `cancel_active_payments` still exist?

saleor/payment/migrations/0034_payment_partial.py
stnatic: Let's leave this out until the whole epic is done. This entire feature branch will have to be eventually rebased to 3.0, so I'm expecting a full migration history rewrite prior to making a PR to master / 3.0.

saleor/checkout/complete_checkout.py
stnatic: I don't want to excuse myself, but it's an extracted method (https://github.com/mirumee/saleor/pull/8001/files/2ccb78ffba77d98d8364fbe443340c10bfbb6f40#diff-27b3f5d7b64efa6b9211eb632292fa7d6f6e031c8a5a38581da6391a901b5d0eL598) and it's used throughout Saleor already (e.g. in pytest fixture names) :bowtie: I wouldn't change the term right now even if it could be confusing to some people

saleor/graphql/payment/mutations.py
stnatic: Thanks for flagging! Changed

saleor/checkout/checkout_cleaner.py
stnatic: I added prefetch at the top level: https://github.com/mirumee/saleor/pull/8001/commits/b53df6cccf0786f92164d8db540ac6d8418fb71d.

`get_active_payments` is being called several times during processing of this request and it's probably cheaper to refer to that qs rather than hitting the database once again.

korycins (approved):

L3str4nge (approved):

PR #8051, opened 2021-09-07
[SALEOR-3677] Add ability to capture all active payments

pzborow:

I need that OrderConfirm mutation trigger capture action on all active payments

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-3677 OrderConfirm mutation should triggers capture action on all active payments](https://app.clickup.com/t/2549495/SALEOR-3677)

saleor/graphql/order/mutations/orders.py
```diff
@@ -677,16 +678,14 @@ def get_instance(cls, info, **data):
...
     @classmethod
-    @traced_atomic_transaction()
-    def perform_mutation(cls, root, info, **data):
-        order = cls.get_instance(info, **data)
-        order.status = OrderStatus.UNFULFILLED
-        order.save(update_fields=["status"])
-        payment = order.get_last_payment()
-        manager = info.context.plugins
+    def process_payment(cls, order, payment, info, manager, to_pay):
         if payment and payment.is_authorized and payment.can_capture():
```
korycins, line 682: You can also drop `payment.is_authorized` as it is covered by `can_capture`

saleor/graphql/order/mutations/orders.py
```diff
@@ -677,16 +678,14 @@ def get_instance(cls, info, **data):
...
+            amount = min(to_pay, Money(payment.get_charge_amount(), payment.currency))
             gateway.capture(
-                payment, info.context.plugins, channel_slug=order.channel.slug
+                payment,
+                info.context.plugins,
+                channel_slug=order.channel.slug,
+                amount=amount,
             )
             order_captured(
```
korycins, line 690: Proably this wasn't included in DoD, but this should be raised only once for mutation, not for all payments. It is an event for capturing order.

saleor/graphql/order/mutations/orders.py
pzborow: OK

saleor/graphql/order/mutations/orders.py
pzborow: Hm. as I image, it was the result of many payment instead one in this mutation. It uses payment to send event that payment captured order. Now there could be may payments. Which one should I used to sent event?

korycins: We should do that in the same way as we're going to do for Order refund action - https://app.clickup.com/t/u9ejzr

pzborow: OK

saleor/graphql/order/mutations/orders.py
```diff
@@ -1,7 +1,9 @@
 import graphene
 from django.core.exceptions import ValidationError
+from prices import Money
 
 from ....account.models import User
+from ....checkout.capture_payment import capture_payments
```
korycins, line 6: Why do you place it in the checkout module? It definitely looks like logic related to order.

saleor/checkout/capture_payment.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def capture_payments(order: Order, manager: PluginsManager):
+    order.status = OrderStatus.UNFULFILLED
+    order.save(update_fields=["status"])
+    to_pay = order.missing_amount_to_be_paid()
+    payments_to_notify = []
+    for payment in order.get_active_payments():  # move it to dedicated file
+        if to_pay > Money(0, to_pay.currency):
+            to_pay -= capture_payment(order, payment, manager, to_pay)
```
korycins, line 28: Keep in mind that (as each action on payment gateway) this can raise an exception. You don't have any error handler here so it can break the whole action. I believe we should call capture as you call. If any fail we should include an failed event for this.

saleor/checkout/capture_payment.py
```diff
@@ -0,0 +1,33 @@
...
+            manager,
+            channel_slug=order.channel.slug,
+            amount=amount,
+        )
+        return amount
+
+
+def capture_payments(order: Order, manager: PluginsManager):
+    order.status = OrderStatus.UNFULFILLED
+    order.save(update_fields=["status"])
```
IKarbowiak, line 23: IMO the status should be updated at the end of this method. I suppose we do not want to update order status when capture failed.

stnatic:
@IKarbowiak @korycins pinging you as I can't edit the original PR. The changes from the original are now:
- I used a dataclass instead of dicts in order to transmit information amount payment + amount (actually, I think it's already called a `Transaction` and that's how it usually gets created 🤔 )
- Capture is called universally when an order is being confirmed
- Tests are improved to check the order of triggered events
- The same logic is used in the `OrderCapture` mutation
- There are some more changes coming that are related to handling multiple payments, eg. in refund https://github.com/SaleorIntegrations/saleor/pull/27/files - I'm happy to push it for review as well, but I didn't want to make this PR extra large. This subsequent PR also makes some code style improvements to the functions used in this PR (e.g. `clean_payment(payments)`). It also restores a logic where order mutations throw payment errors instead of silent failing

saleor/order/actions.py
```diff
@@ -83,31 +87,68 @@ def order_created(
...
+        ]
+        if captured_payments:
+            order_captured(
                 order=order,
                 user=user,
                 app=app,
-                amount=payment.total,
-                payment=payment,
+                amounts_and_payments=[
+                    OrderPaymentAction(
```
korycins, line 120: Do we need this structure here? Looks like we could pass a list of `payment`s here and inside a body use a `payment.capture_amount` (not sure if this is possble).

saleor/order/capture_payment.py
```diff
@@ -0,0 +1,74 @@
+from decimal import Decimal
+from typing import Optional
+
+from saleor.account.models import User
```
korycins, line 4: Absolute

saleor/graphql/order/mutations/orders.py
```diff
@@ -9,11 +10,10 @@
 from ....core.tracing import traced_atomic_transaction
 from ....order import OrderLineData, OrderStatus, events, models
 from ....order.actions import (
+    _capture_payments,
```
korycins, line 13: Shouldn't remove the underscore from the function? I would assume that the functions with `_` will be used only in local file.

saleor/order/interface.py
```diff
@@ -0,0 +1,12 @@
+from dataclasses import dataclass
+from decimal import Decimal
+
+from saleor.payment.models import Payment
```
korycins, line 4: 🚔

saleor/order/notifications.py
```diff
@@ -4,6 +4,8 @@
 from django.forms import model_to_dict
 from graphql_relay import to_global_id
 
+from saleor.order.interface import OrderPaymentAction
```
korycins, line 7: 🚔

saleor/order/tests/test_notifications.py
```diff
@@ -6,6 +6,8 @@
 from measurement.measures import Weight
 from prices import Money, fixed_discount
 
+from saleor.order.interface import OrderPaymentAction
```
korycins, line 9: 🚔

saleor/order/tests/test_order_actions.py
```diff
@@ -5,8 +5,10 @@
 import pytest
 from prices import Money, TaxedMoney
 
+from saleor.order.interface import OrderPaymentAction
```
korycins, line 8: 🚔

korycins (commented):
Please change added imports from absolute to relative

saleor/order/interface.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+@dataclass
+class OrderPaymentAction:
```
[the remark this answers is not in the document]
stnatic: IMO no. The reason is that this dataclass is passed to functions that are already operating in the context of a particular payment action. It would make it extra annoying to use that class and duplicate the action type every single time.

saleor/order/actions.py
```diff
@@ -83,31 +87,68 @@ def order_created(
...
                 manager=manager,
             )
-        elif order.is_pre_authorized():
-            order_authorized(
+
+        captured_payments = [
+            p
+            for p in payments
+            if p.charge_status
+            in [
```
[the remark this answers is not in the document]
maarcingebala: It would be useful to have a helper like `get_captured_payments(payments)` that we can easy reuse and it encapsulates the logic of checking proper statuses, do we have something like that?

saleor/order/actions.py
```diff
@@ -83,31 +87,68 @@ def order_created(
...
+            ]
+        ]
+        if captured_payments:
+            order_captured(
                 order=order,
                 user=user,
                 app=app,
-                amount=payment.total,
-                payment=payment,
+                amounts_and_payments=[
```
maarcingebala, line 119: In some places this is called payments and here it's amounts_and_payments. We should settle on one consistent name for that, e.g. `payments` (but this could suggest that it expects a list of Payment instances) or `payment_actions` etc.

saleor/order/capture_payment.py
stnatic: I'll put all of the logic under `actions.py` in the subsequent PR and fix import / remove underscores on the public interface.

PR #8057, opened 2021-09-09
Fix bug with handling orders containing deleted variants in 3.1

jakubkuc:
I want to merge this change because user should be able to manage orders (adding tracking number, deleting) containing variants that were deleted after the order was created.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-4253 Errors when handling orders containing deleted variants](https://app.clickup.com/t/2549495/SALEOR-4253)

IKarbowiak (approved):

korycins:
I believe this should be merged to master instead of 3.1 branch

saleor/order/tests/test_notifications.py
```diff
@@ -160,6 +160,46 @@ def test_get_order_line_payload(order_line):
...
+    order_line.variant = None
+    payload = get_order_line_payload(order_line)
+    unit_tax_amount = (
+        order_line.unit_price_gross_amount - order_line.unit_price_net_amount
+    )
+    total_gross = order_line.unit_price_gross * order_line.quantity
+    total_net = order_line.unit_price_net * order_line.quantity
+    total_tax = total_gross - total_net
+    assert payload == {
+        "variant": None,
```
korycins, line 173: IMHO the standard payload is covered in different test. I would say that this is enough to only check if a variant field exists in the payload. (No need to twice check the same payload)

jakubkuc: It was also my concern. But I think other tests do not cover the payload with deleted variant so we can't be sure if any change made in code won't change anything in the whole payload in this specific case. What do you think?

korycins (approved):

jakubkuc:
Invalid pull request base

PR #8061, opened 2021-09-09
[SALEOR-3782] Update OrderEvents for payments

pzborow:

I need update OrderEvents for payments

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-3782 Update OrderEvents for payments](https://app.clickup.com/t/2549495/SALEOR-3782)

saleor/order/migrations/0114_alter_orderevent_type.py
```diff
@@ -0,0 +1,77 @@
...
+PAYMENT_REFUND_FAILED = "payment_refund_failed"
+PAYMENT_VOID_FAILED = "payment_void_failed"
+
+
+def change_fail_event_type_up(apps, schema_editor):
+    order_event = apps.get_model("order", "OrderEvent")
+
+    refund_events = order_event.objects.filter(
+        type=PAYMENT_FAILED,
+        parameters__message__icontains="refund"
```
korycins, line 17: I am not sure if from performance perspective it is okay. Parameters is a json field, and we don't have any index applied to this field that will cause a lot of operation on DB side. Personally, I would recommend to iterate over failed payment events and convert all of them into all new events. You can create 3 lists (void_events, refund_events and capture_events) and placed them each of the event. At the end you can call bulk action to update all of them

pzborow: Dividing into three lists probably won't do the trick. You will have to go through all the records anyway. The whole problem is that, first, everything has to be done in one transaction. Second, you have to query this field for the value in json field anyway. I see two possibilities. Or, to optimize the transaction, divide it into three migrations, and thus into three smaller transactions. Or you can filter the query only by type and set the value using SQL case.
In my opinion, it is a fight for a constant computational complexity, we will not change the algorithm to a more efficient one.

korycins: It is fine to go through all records if we use an iterator for that. The lists could contain only an ID of the payments which can be used to call .filter(id__in=list).update(type=new_status). We don't want to call `icontains`, in the loop we could just take the message from object and based on this propagate it to the proper list. 
Wouldn't that work?

pzborow: I get it, so you think that looking into json will be more efficient on the python side than on the postgres side? Sequence scan has to be done inside type==PAYMENT_FAILED anyway

korycins: You need to apply changes for each event object, so I would prefer to iterate over all these events only once than calling three times filters over single json fields. In similar project, we had a situation when such a query broke the migration process because the data set was huge, and we couldn't finish it. Probably this migration is not a case where this can happen but I would prefer to be sure than testing it on production environments

saleor/graphql/order/mutations/orders.py
```diff
@@ -139,20 +139,34 @@ def clean_refund_payment(payment):
...
-        return result
-    except (PaymentError, ValueError) as e:
-        message = str(e)
-        events.payment_failed_event(
-            order=order, user=user, app=app, message=message, payment=payment
-        )
-        raise ValidationError(
-            {"payment": ValidationError(message, code=OrderErrorCode.PAYMENT_ERROR)}
-        )
+def try_payment_action_factory(payment_failed_event_func):
```
korycins, line 142: I think that it would be good to fully remove `try_payment_action` function. 
We should have separate functions for capture, refund, void etc. We want to trigger there a different OrderEvents or custom logic. Using generic logic for these actions complicates it as it is now in master codebase.

stnatic: @korycins Let's merge what we have and have Anatoly resolve this point in #8082. Otherwise we'll run into some pretty nasty conflicts.

korycins: @stnatic merged

PR #8102, opened 2021-09-15
Improve draft orders and orders webhooks 

### LINKED ISSUE #7620: Editing the quantity of a line item in a draft order does not trigger the `order_updated` plugin method
There doesn't appear to be any methods in the BasePlugin class that can be triggered by editing a draft order's line item's quantity. As far as I can tell this is the only element of an order that can be changed without triggering at least one of the built in plug in methods. I'm not sure if this is by design or just an oversight. I would expect to see a plugin method called in the save method of the OrderLineUpdate mutation class. Either the generic `order_updated` method, or a specific method for line item quantity would seem appropriate. 

Saleor version:
- [ ] dev (current master)
- [x] 3.0
- [ ] 2.11
- [ ] 2.10

jakubkuc:
I want to merge this change because there were missing draft order webhooks and order_update webhook was not called in some mutations.

Additionally in separate commit mocks for invoice tests are changed from BasePlugin to PluginManager

Fixes #7620

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-4008 Improve draft order webhooks](https://app.clickup.com/t/2549495/SALEOR-4008)

saleor/graphql/order/mutations/orders.py
```diff
@@ -781,6 +785,8 @@ def perform_mutation(cls, _root, info, **data):
         )
 
         recalculate_order(order)
+        transaction.on_commit(lambda: info.context.plugins.order_updated(order))
```
korycins, line 788: I believe that this mutation can be also used for draft order, which will raise an incorrect webhook event

saleor/webhook/event_types.py
```diff
@@ -18,6 +18,10 @@ class WebhookEventType:
     ORDER_CANCELLED = "order_cancelled"
     ORDER_FULFILLED = "order_fulfilled"
 
+    DRAFT_ORDER_CREATED = "draft_order_created"
```
korycins, line 21: You didn't add a label and permission for this webhook
See below 👇  a `DISPLAY_LABELS`, `CHOICES` and `PERMISSIONS`

saleor/graphql/order/tests/test_order.py
```diff
@@ -3597,6 +3630,7 @@ def test_invalid_order_when_updating_lines(
     content = get_graphql_content(response)
     data = content["data"]["orderLineUpdate"]
     assert data["errors"]
+    assert not order_update_webhook_mock.called
```
IKarbowiak, line 3633: I see, that in code, we mostly use `assert_not_called()` so it will be nice to unify this. But I also see that is some places where we check this in the way you did it. So I don't press for change, it's up to you.

IKarbowiak (approved):

saleor/graphql/order/mutations/orders.py
```diff
@@ -386,6 +388,7 @@ def perform_mutation(cls, _root, info, **data):
             app=info.context.app,
             message=cleaned_input["message"],
         )
+        transaction.on_commit(lambda: info.context.plugins.order_updated(order))
```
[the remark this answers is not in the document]
IKarbowiak: We discussed it with @patrys, and decided to do it in this way, as it's updating the order history.

saleor/graphql/order/mutations/orders.py
```diff
@@ -898,6 +908,9 @@ def save(cls, info, instance, cleaned_input):
                 code=OrderErrorCode.INSUFFICIENT_STOCK,
             )
         recalculate_order(instance.order)
+        transaction.on_commit(
```
korycins, line 911: Are we sure that we want to trigger a webhook here? We can add/remove lines only when the order is in `DRAFT` and `UNCONFIRMED` status. This will trigger one webhook for any change in quantity number in a single line for `DRAFT` and for `UNCONFIRMED` order. The same situation is for `OrderLineCreate` and `OrderLineDelete`. 
cc @maarcingebala, @patrys

korycins: @jakubkuc  Okay, everything is clear for me. The one thing which I think needs to be changed is that we should not call `order_updated` when this mutation is called for DRAFT order.  (And it is related to all mutations called on order lines create/update/delete).

maarcingebala: We could call the right plugin method based on order's status, couldn't we? I mean something like:
```
if order.status == OrderStatus.DRAFT:
    func = info.context.plugins.draft_order_updated
else:
    func = info.context.plugins.order_updated
```

Because now I believe we use the same mutation for updating order lines in drafts and regular orders.

korycins: Yes, sorry,  I forgot to mention how we should solve it. As @maarcingebala  said, I think we can create simple `if else`

saleor/graphql/order/mutations/orders.py
jakubkuc: changed to raise webhook event only for order lines other than draft

saleor/webhook/event_types.py
maarcingebala: @jakubkuc We need to update this as well.

saleor/graphql/order/mutations/orders.py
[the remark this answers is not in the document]
maarcingebala: This mutation can be also triggered for draft order, so we should move this logic here as well:
```
        if order.status == OrderStatus.DRAFT:
            func = info.context.plugins.draft_order_updated
        else:
            func = info.context.plugins.order_updated
```

I'm thinking that we could actually use this logic for all order mutations, so that we are sure that we'll always call the right webhook. Maybe we should create a small helper function so that we don't have to copy these four lines in every mutation.

saleor/graphql/order/mutations/orders.py
```diff
@@ -781,6 +785,12 @@ def perform_mutation(cls, _root, info, **data):
         )
 
         recalculate_order(order)
+        if order.status == OrderStatus.DRAFT:
```
L3str4nge, line 788: This block of code is repeated a few times, maybe create function like `get_function_by_order_status`?

saleor/graphql/order/tests/test_order.py
```diff
@@ -3430,11 +3456,25 @@ def test_order_lines_create_with_existing_variant(
     data = content["data"]["orderLinesCreate"]
     assert data["orderLines"][0]["productSku"] == variant.sku
     assert data["orderLines"][0]["quantity"] == old_quantity + quantity
+    if status == OrderStatus.DRAFT:
```
L3str4nge, line 3459: The same thing like in logic, we could have function for assertion.

saleor/webhook/event_types.py
jakubkuc: fixed it

saleor/graphql/order/mutations/orders.py
jakubkuc: fixed it. Is it acceptable now? It could be done all in one line in lambda but I though this way was more readable.

saleor/graphql/order/tests/test_order.py
jakubkuc: fixed it

saleor/graphql/order/mutations/orders.py
```diff
@@ -155,6 +156,13 @@ def try_payment_action(order, user, app, payment, func, *args, **kwargs):
         )
 
 
+def get_function_by_order_status(status, info):
```
maarcingebala, line 159: Maybe `get_webhook_handler_by_order_status` to make the name more descriptive?

maarcingebala (approved):

L3str4nge:
@jakubkuc resolve conflicts please

PR #8105, opened 2021-09-15
Change PluginsManager to call methods only on active plugins

L3str4nge:
Fix for `PluginsManager` to call methods only on active plugins. Flag `active_only` is passed into `get_plugins` method in `__run_method_on_plugins` because after investigation it looks like there is no situation when this method should be called on `inactive` plugin.  Actions like that are handled by other methods like `save_plugin_configuration` or `list_external_authentications`.

I've also added missing tests for `__run_method_on_plugins` and `__run_method_on_single_plugin` to be sure that we have proper output from them because for now they are not tested directly - just by functions called from the upper level. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/plugins/tests/sample_plugins.py
+++ b/saleor/plugins/tests/sample_plugins.py
@@ -221,6 +221,9 @@ class PluginSample(BasePlugin):
     def get_order_shipping_tax_rate(self, order: "Order", previous_value: Decimal):
         return Decimal("0.080").quantize(Decimal(".01"))
 
+    def sample_not_implemented(self, previous_value):
+        return NotImplemented
+
 
 class ChannelPluginSample(PluginSample):
     PLUGIN_ID = "channel.plugin.sample"
--- a/saleor/plugins/tests/sample_plugins.py
+++ b/saleor/plugins/tests/sample_plugins.py
@@ -324,3 +324,19 @@ class InactivePaymentGateway(BasePlugin):
 
     def process_payment(self, payment_information, previous_value):
         pass
+
+
+ACTIVE_PLUGINS = (
+    __name__ + "." + ChannelPluginSample.__name__,
+    __name__ + "." + ActivePaymentGateway.__name__,
+    __name__ + "." + ActivePlugin.__name__,
+    __name__ + "." + ActiveDummyPaymentGateway.__name__,
+)
+
+INACTIVE_PLUGINS = (
+    __name__ + "." + InactivePaymentGateway.__name__,
+    __name__ + "." + PluginInactive.__name__,
+    __name__ + "." + InactiveChannelPluginSample.__name__,
+)
+
+ALL_PLUGINS = ACTIVE_PLUGINS + INACTIVE_PLUGINS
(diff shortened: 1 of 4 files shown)


patrys:
Task linked: [SALEOR-2831 PluginsManager should call action only on active plugin](https://app.clickup.com/t/2549495/SALEOR-2831)

saleor/plugins/tests/test_manager.py
```diff
@@ -875,3 +876,92 @@ def test_list_external_authentications_active_only(channel_USD):
...
+def test_run_method_on_plugins_only_on_active_ones(
+    mocked_method, channel_USD, all_plugins_manager
+):
+    all_plugins_manager._PluginsManager__run_method_on_plugins(
+        method_name="test_method_name",
+        default_value="default_value",
+    )
+
+    assert len(all_plugins_manager.all_plugins) == 7
+    assert len([p for p in all_plugins_manager.all_plugins if p.active]) == 4
```
korycins, line 927: Would be good to confirm that these 4 plugins are our active plugins, but it can be also in current format if this would require too much effort

korycins (approved):

tomaszszymanski129 (approved):

saleor/plugins/tests/test_manager.py
L3str4nge: I could sort out these sample plugins in `sample_plugins.py`  like `inactive_plugins = [...]` etc and use it in fixtures. Would be more readable in tests. WDYT? @korycins

IKarbowiak (approved):

saleor/plugins/tests/sample_plugins.py
```diff
@@ -321,3 +324,19 @@ class InactivePaymentGateway(BasePlugin):
...
+    __name__ + "." + ActivePaymentGateway.__name__,
+    __name__ + "." + ActivePlugin.__name__,
+    __name__ + "." + ActiveDummyPaymentGateway.__name__,
+)
+
+INACTIVE_PLUGINS = (
+    __name__ + "." + InactivePaymentGateway.__name__,
+    __name__ + "." + PluginInactive.__name__,
+    __name__ + "." + InactiveChannelPluginSample.__name__,
+)
```
tomaszszymanski129, line 340: we commonly use f-strings for concentrating, however in this case maybe it's more readable as is.

tomaszszymanski129 (approved):

saleor/plugins/tests/sample_plugins.py
L3str4nge: I've moved this concating to fixture directly.

saleor/plugins/tests/test_manager.py
L3str4nge: I've added checking these ids.

korycins (approved):

fowczarek (approved):

eCzerniak (approved):

PR #8165, opened 2021-09-22
Extend FulfillmentRefundProducts mutation to be able to handle paymentID

solartune:

This PR addresses the SALEOR-3669 issue, and brings these changes:

1. The `amount_to_refund` and `include_shipping_costs` attributes of the `OrderRefundProductsInput` are now optional.
2. There is a new optional argument called payments_to_refund that accepts a list of payment IDs mapped with amounts to refund and it's possible to specify one `include_shipping_costs` per list.
3. Appropriate functions and methods which were accepting a value of the `amount_to_refund` argument before now were refactored and accept and deal with a list of payments to refund mapped with appropriate amounts.

NOTE:
I expect some parts of the code to be refactored when working on the issue `SALEOR-3670`. Some other parts can be refactored when the deprecated fields will be removed and no longer supported.

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] The changes are tested


patrys:
Task linked: [SALEOR-3669 Extend FulfillmentRefundProducts mutation to be able to handle paymentID](https://app.clickup.com/t/2549495/SALEOR-3669)

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -622,22 +645,124 @@ class Meta:
...
+
+    @classmethod
+    def _get_payments_to_refund(
+        cls, info, order, payments_to_refund, amount_to_refund, include_shipping_costs
+    ):
+        if payments_to_refund:
+            cls._check_shipping_costs(payments_to_refund)
+            payments = []
+            for item in payments_to_refund:
+                payment = cls.get_node_or_error(info, item["payment_id"])
```
korycins, line 709: Can we fetch payments in bulk? `.objects.in_bulk(keys)`

saleor/order/actions.py
```diff
@@ -910,19 +908,36 @@ def create_refund_fulfillment(
...
     with transaction_with_commit_on_errors():
+        refund_shipping_costs = False
+        # Amount to be eventually specified in the Fulfillment object.
+        counted_shipping_amount = None
+        for item in payments:
+            if item["include_shipping_costs"]:
+                if not item["amount"]:
+                    counted_shipping_amount = order.shipping_price_gross_amount
+                refund_shipping_costs = True
+                shipping_refund_amount = __get_shipping_refund_amount(
```
korycins, line 920: Will it work as expected? `__get_shipping_refund_amount` sets amount to refund only when `item["amount"]` will be None and you build a payment_to_refund structure you set `"amount": amount_to_refund or Decimal("0"),` 
fulfillments.py line 775
https://github.com/mirumee/saleor/pull/8165/files#diff-bb17901778348f1cf6d54c9c0b102b8e7ea967d9df62385315c45c786d93ffc9R735

saleor/order/actions.py
```diff
@@ -1361,37 +1380,51 @@ def _process_refund(
...
+
+    # Happens when neither payments_to_refund nor amount_to_refund were specified.
     if amount is None:
         amount = refund_amount
         # we take into consideration the shipping costs only when amount is not
         # provided.
         if refund_shipping_costs:
             amount += order.shipping_price_gross_amount
+        # At this point there can be only one payment.
+        payments[0]["amount"] = min(payments[0]["payment"].captured_amount, amount)
```
korycins, line 1392: Not sure of this. (probably I am missing something) Why do we do this? What in the case,  when we are not able to include shipping_price to amount of first payment because captured_amount of first payment is less than amount+shipping_price?
`refund_shipping_costs` was the flag that determines if we want to refund shipping costs for payment, when we have a single payment. Shouldn't we operate on `item["include_shipping_costs"]` and drop this arguments?

saleor/order/actions.py
```diff
@@ -1361,37 +1380,51 @@ def _process_refund(
...
-            )
-        transaction.on_commit(
-            lambda: events.payment_refunded_event(
-                order=order,
-                user=user,
-                app=app,
-                amount=amount,  # type: ignore
-                payment=payment,
+        for item in payments:
+            try:
```
korycins, line 1396: Assuming that you have 3 payments. 
1st will be processed with success, the second also, then the 3rd one will raise an error, which will raise a ValidationError, in that case I think, that we will not have any events about success payment refund (which are called in transaction.on_commit.

saleor/graphql/order/mutations/fulfillments.py
solartune: Done

saleor/order/actions.py
solartune: Done

saleor/order/actions.py
korycins: Still, I don't understand it. `proccess_refund` should call refund action for each payment. The total  amount of refund should be taken from automatically calculated refund OR from `payment['amount']` + `payment['include_shipping_costs']` determined for each payment.
In case when we don't have `payment['amount']` value, then we should use `min(payment.captured,total_amound_of_refund)`. The cases which we need to cover:

- _user doesn't provide amount of returned value but needs to provide a list of paymentIDs in proper order, which will be used to call refund. (In case when amount to return is greater than amount from single payment. For amount lower than amount of single payment the list can have a single paymentID).  Saleor calculates refund amount, call refund for max amount of payment and move to next payment from provided list_ -  This will use automatically calculated total of refund and it will iterate over a payment and process a refund one by one until we will cover `total of refund`. 

- _user provides amount to refund for each payment. 
example when order has two payments to refund:
15 USD for 1st payment
10 USD for 2nd payment_ - we call try_refund for 'amount' defined for each `payment' object
- _user provides amount only for part of refunded payments - [{paymentID: "ID1"}, {"paymentID: "ID2", amount: 15}]
First call refund for payments with defined amount, then call refund for payments which don't have provided amount_ - `_process_refund` should call refund for objects with determined `amount` value, then we should call refund for the rest of the payments, where amount to refund is == `total_amount_of_refund - sum(payments['amount'] and is not None)`

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -622,22 +646,137 @@ class Meta:
...
+                    )
+                }
+            )
+
+    @classmethod
+    def _check_order_has_single_payment(cls, order):
+        if order.payments.count() > 1:
+            raise ValidationError(
+                {
+                    "amount_to_refund": ValidationError(
```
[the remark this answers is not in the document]
korycins: @solartune  I think I am wrong here. Sorry for that! The `amount_to_refund` is Ok.

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -447,14 +462,23 @@ class OrderRefundProductsInput(graphene.InputObjectType):
...
+            f"{DEPRECATED_IN_3X_INPUT} Use paymentsToRefund instead."
+        ),
     )
     include_shipping_costs = graphene.Boolean(
         description=(
-            "If true, Saleor will refund shipping costs. If amountToRefund is provided"
-            "includeShippingCosts will be ignored."
+            "If true, Saleor will refund shipping costs. If amountToRefund is provided "
+            "includeShippingCosts will be ignored. "
+            f"{DEPRECATED_IN_3X_INPUT} Use paymentsToRefund instead."
```
stnatic, line 481: @korycins Shouldn't we keep it where it is? Dashboard designs have only one "include_shipping_costs" checkbox. It hard or me to imagine the logic of handling it per payment.

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -622,22 +646,137 @@ class Meta:
         error_type_class = OrderError
         error_type_field = "order_errors"
 
+    @classmethod
+    def _check_shipping_costs(cls, payments_to_refund):
```
stnatic, line 650: This wouldn't be needed at all if we decided to keep `include_shipping_costs` where it is.

saleor/graphql/order/mutations/fulfillments.py
korycins: Hhm... Yes, you're right. We don't need it, we ignore `includeShippingCosts` when `amountToRefund` is provided, so we could just add it to the total refund amount, but not sure how we should handle the case, when part of payments to refund has provided a refund amount, and the rest of the payments doesn't have it.  (last point from this comment:https://github.com/mirumee/saleor/pull/8165#discussion_r716547245)

saleor/order/actions.py
solartune: All the cases described here don't apply to the line above. When a list of payments is provided we include the shipping costs in this line:
https://github.com/mirumee/saleor/pull/8165/files#diff-45007f7469800f0df7ae204bece8aa9303d0b7261e52ac1fdd9a5d710d089368R916

I'd consider this as a transitioning step that we can refactor later.

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -622,22 +646,137 @@ class Meta:
...
             info, order_id, field="order", only_type=Order, qs=qs
         )
-        payment = order.get_last_payment()
-        cls.clean_order_payment(payment, cleaned_input)
-        cls.clean_amount_to_refund(amount_to_refund, payment, cleaned_input)
+        payments = cls._get_payments_to_refund(
+            info, order, payments_to_refund, amount_to_refund, include_shipping_costs
+        )
+        for item in payments:
+            cls.clean_order_payment(item["payment"], cleaned_input)
```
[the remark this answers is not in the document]
korycins: Sounds good 👍

korycins (approved):

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -622,22 +646,138 @@ class Meta:
...
+            improper_payments_global_ids = [
+                graphene.Node.to_global_id("Payment", payment_id)
+                for payment_id in improper_payments_ids
+            ]
+            code = OrderErrorCode.PAYMENTS_DO_NOT_BELONG_TO_ORDER
+            raise ValidationError(
+                {
+                    "payments_to_refund": ValidationError(
+                        "These payments do not belong to the order.",
+                        code=code,  # type: ignore
```
IKarbowiak, line 684: Instead of adding `# type: ignore` just get `value` from code.
`code = OrderErrorCode.PAYMENTS_DO_NOT_BELONG_TO_ORDER.value`

solartune: Thanks, added.
Any idea why it didn't complain about other similar lines? As in other cases, the value isn't passed.

IKarbowiak: It seems like it checks types only in methods where the type hints are defined.

IKarbowiak: And, I didn't see the change 🤔

IKarbowiak (approved):

korycins:
@solartune  Let's solve a conflicts and we are good to merge it

solartune:
> @solartune Let's solve a conflicts and we are good to merge it

@korycins done. One test failed but I cannot reproduce it locally. Seems to be a nondeterministic one.

saleor/order/actions.py
```diff
@@ -1361,37 +1370,42 @@ def _process_refund(
...
+        for item in payments:
+            try_refund(
                 order=order,
                 user=user,
                 app=app,
-                amount=amount,  # type: ignore
-                payment=payment,
+                payment=item["payment"],
+                manager=manager,
+                channel_slug=order.channel.slug,
```
maarcingebala, line 1392: `order.channel` is a relationship, not sure if it wouldn't run a database query in each iteration? we could assign the channel slug to a variable before this for-loop and use the variable here.

solartune: Django uses a cached value for later calls.
From [the docs](https://docs.djangoproject.com/en/3.2/topics/db/queries/#one-to-many-relationships):

> Forward access to one-to-many relationships is cached the first time the related object is accessed. Subsequent accesses to the foreign key on the same object instance are cached.

maarcingebala (approved):

PR #8181, opened 2021-09-24
[3.1] Convert ShippingMethod type into ShippingMethodType

kczan:
[SALEOR-4410](https://app.clickup.com/t/2549495/SALEOR-4410)

Port of changes for 3.0 - https://github.com/mirumee/saleor/pull/8212
# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-4410 Convert ShippingMethod type into ShippingMethodType](https://app.clickup.com/t/2549495/SALEOR-4410)

saleor/graphql/schema.graphql
```diff
@@ -4712,7 +4712,7 @@ type PageTranslatableContent implements Node {
 type PageTranslate {
   translationErrors: [TranslationError!]! @deprecated(reason: "This field will be removed in Saleor 4.0. Use `errors` field instead.")
   errors: [TranslationError!]!
-  page: PageTranslatableContent
+  page: Page
```
korycins, line 4715: Looks like unexpected change.

korycins:
@kczan  black is failing

saleor/graphql/schema.graphql
kczan: This change comes from here https://github.com/mirumee/saleor/blob/2e7b5aa1b312e1792ad3ff7930e0217d60be5629/saleor/graphql/translations/mutations.py#L44
I'm not sure why that is the case though, because it seems to me that it should've been "Page" already before, since that's the value of `page_models.Page._meta.object_name`.
I think that it should stay as `Page` in the schema, since if I'm not mistaken, we want to point to the actual model here.

korycins: Not sure if we should change the schema in that way. 
@maarcingebala are you able to help here? I am not sure if the PageTranslatableContent is something expected here.

fowczarek (approved):

saleor/graphql/translations/mutations.py
```diff
@@ -25,37 +31,23 @@
...
-    str(
-        translation_types.SaleTranslatableContent
-    ): discount_models.Sale._meta.object_name,
-    str(
-        translation_types.VoucherTranslatableContent
-    ): discount_models.Voucher._meta.object_name,
-    str(
-        translation_types.MenuItemTranslatableContent
-    ): menu_models.MenuItem._meta.object_name,
+    str(translation_types.ProductTranslatableContent): str(product_types.Product),
```
maarcingebala, line 34: This should map type names to model names. Currently in all cases the names are the same but I'm not sure if it always stay like that so I think it's safer to keep the mapping as it was before.

saleor/graphql/schema.graphql
maarcingebala: It looks like in other translation types like this (e.g. ProductTranslation) we don't use the TranslatableContent type, so it seems that it should be `page: Page` before 🤔

saleor/graphql/translations/schema.py
```diff
@@ -58,7 +58,7 @@ class TranslatableKinds(graphene.Enum):
     PAGE = "Page"
     PRODUCT = "Product"
     SALE = "Sale"
-    SHIPPING_METHOD = "ShippingMethod"
+    SHIPPING_METHOD = "ShippingMethodType"
```
maarcingebala, line 61: This should stay as `SHIPPING_METHOD` or should we rename it asa well?

saleor/graphql/schema.graphql
korycins: Yes, it looks like a bug and it should be a `page`. 
This type is only used in:
`pageTranslate(id: ID!, input: PageTranslationInput!, languageCode: LanguageCodeEnum!): PageTranslate`
So I assume that the dashboard also uses `PageTranslatableContent` instead of `Page`.
Not sure what should we do with this now. I am worrying that this change can break any integration which operates on page: PageTranslatableContent. 
cc @maarcingebala

saleor/graphql/translations/mutations.py
stnatic: @maarcingebala This mapping is being passed to `graphene.Node.to_global_id` which expects a graphene type 🤔 
image

maarcingebala: You're right, it should map translatable content types to their corresponding "base" types, so `PageTranslatableContent` to `Page`. So the changes are correct and the mapping could be actually renamed to something like `TRANSLATABLE_CONTENT_TO_TYPE`.

saleor/graphql/schema.graphql
stnatic: @maarcingebala @korycins 

- django-graphene keeps a registry (1:1 function) between graphene.DjangoObjectTypes and django.db.Model
- this 1:1 mapping is crucial - DjangoObjectType automatically converts from model instances returned by resolvers to object types
- there are several models with a duplicated ObjectType (I can send you a list if you ping me on Slack)
- if there are multiple ObjectTypes with the same model, the latest one to be imported ends up in the registry

This is an example commit, that "reverses" this change in schema by delaying the import of the "Page object type":
https://github.com/mirumee/saleor/commit/8f6754666292494cfd066c7e5087f6e1f5df7d47

patrys: We want to get rid of `django-graphene` as it is blocking us from using `asyncio`. Can we change both to regular `ObjectType`s?

stnatic: @patrys we'll implementing the new `ShippingMethod` as a plain `ObjectType`. I'll let you know how complex does it make things.

korycins: Not sure if @patrys  and @maarcingebala  agree with me but I wouldn't change the returned type for the page (from `PageTranslatableContent` to `Page`).  I agree that we need to change it as it is a bug but I would like to do that in separate PR after we will confirm that we won't break any integration/storefront.

stnatic: @maarcingebala @patrys @korycins 

For now what we'll do is:
```
    @classmethod
    def get_type_for_model(cls):
        return PageTranslatableContent
```

This ensures that the type in `PageTranslate` won't change in this PR. This could be even a long term fix for Saleor's `ModelMutation` as we don't seem to be using other Django-graphene features that touch the registry. It should be pretty straightforward to implement project-wise as we can raise NotImplemented in `BaseTranslateMutation` and fix tests

saleor/graphql/translations/schema.py
kczan: I think that those keys more accurately refer to the model, and not the type. So in my opinion it should stay as `SHIPPING_METHOD`

saleor/graphql/translations/mutations.py
kczan: I've changed it to `TRANSLATABLE_CONTENT_TO_TYPE`

korycins (approved):

PR #8217, opened 2021-09-27
Extend Adyen's and Stripe's webhooks to create an order when total_paid covers a checkout.total

solartune:

This PR addresses the SALEOR-3751 and SALEOR-3752 issues, and brings these changes:
1. A new field called `create_order` has been added to the Payment model
2. If a payment (along with previous payments) covers the checkout's total amount the field's value is set to True.
3. When a notification arrives we now check if an order should be created with this payment.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-3751 Extend Adyen's webhooks to create an order when total_paid covers a checkout.total](https://app.clickup.com/t/2549495/SALEOR-3751)

patrys:
Task linked: [SALEOR-3752 Extend Stripe's webhook to create an order when total_paid covers a checkout.total](https://app.clickup.com/t/2549495/SALEOR-3752)

saleor/payment/models.py
```diff
@@ -61,6 +61,14 @@ class Payment(models.Model):
     order = models.ForeignKey(
         "order.Order", null=True, related_name="payments", on_delete=models.PROTECT
     )
+    create_order = models.BooleanField(
```
stnatic, line 64: It seems to duplicate `complete_order` that I added earlier on?

saleor/payment/gateways/adyen/tests/webhooks/test_handle_notifications.py
```diff
@@ -1010,3 +1011,37 @@ def test_handle_cancel_or_refund_action_cancel_invalid_payment_id(
...
+
+    handle_not_created_order(
+        notification,
+        payment,
+        payment.checkout,
+        TransactionKind.ACTION_TO_CONFIRM,
+        manager,
+    )
+
+    if return_value is True:
```
L3str4nge, line 1044: I think it's not a good idea to create any logic in the tests. It's hard to mantain and understand. How about to create to separate tests one for ` mock_create_order.assert_called_once_with(payment, payment.checkout, manager)` and
second one for `mock_create_order.assert_not_called()`?

saleor/payment/gateways/stripe/tests/test_webhooks.py
```diff
@@ -554,3 +555,54 @@ def test_handle_webhook_events(
...
+
+    _process_payment_with_checkout(
+        payment,
+        payment_intent,
+        TransactionKind.ACTION_TO_CONFIRM,
+        payment.total,
+        channel_USD.slug,
+    )
+
+    if return_value is True:
```
L3str4nge, line 597: The same as above.

saleor/payment/models.py
```diff
@@ -120,6 +128,14 @@ def get_last_transaction(self):
     def get_total(self):
         return Money(self.total, self.currency)
 
+    def can_create_order(self):
+        """Indicate whether an order can be created with this payment.
+
+        If the value is None it still should be possible
+        for supporting the legacy approach.
+        """
+        return True if self.create_order or self.create_order is None else False
```
L3str4nge, line 137: Why not just `return True`? 😄

saleor/graphql/payment/mutations.py
```diff
@@ -160,6 +160,15 @@ def validate_return_url(cls, input_data):
                 {"redirect_url": error}, code=PaymentErrorCode.INVALID
             )
 
+    @classmethod
+    def check_covered_amount(cls, checkout, checkout_total, amount):
```
korycins, line 164: Can we add typing here? Example `checkout_total`  doesn't tell me if I should pass the value as a `Money` or `TaxedMoney`.

saleor/payment/utils.py
```diff
@@ -99,6 +99,7 @@ def create_payment(
     order: Order = None,
     return_url: str = None,
     external_reference: Optional[str] = None,
+    is_amount_fully_covered: bool = None,
```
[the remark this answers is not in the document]
solartune: With the changes, I merged recently the `complete_order` field appeared in the Payment model with some logic around it that happens in the `CheckoutComplete`. Now it seems to me that we can use that field in the hooks to determine whether an order should be created. And the logic here, around the field `created_order` isn't really necessary and can be removed altogether with the field.

Can you confirm it?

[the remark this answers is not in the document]
korycins: I think complete_order should handle this situation. 
Logic which handles a webhook from Stripe or Adyen should relay on this flag. If the given payment has this flag set to True we assume that this was the last payment and we can create an order.
This flag should only be set to True in checkoutComplete action. If we will process all payments via checkoutPaymentComplete, checkoutComplete will only create order (not webhook notifications which can be delivered before we will call checkoutComplete). If we process a payment inside a checkoutComplete we assume that this payment is the last payment so we can immediately create an order via webhook notification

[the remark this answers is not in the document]
solartune: Switched to using the `complete_order`

saleor/payment/gateways/adyen/tests/webhooks/test_handle_notifications.py
solartune: I decided to go that way as the difference in the tests is tiny and the logic isn't complicated. But no problem with splitting it into multiple tests. Done.

saleor/payment/models.py
solartune: Removed

PR #8246, opened 2021-09-30
Add gift card fulfillment line relation

IKarbowiak:
Add `GiftCard` - `FulfillmentLine` relation.
When a gift card is created assign a corresponding fulfillment line.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/giftcard/utils.py
```diff
@@ -165,6 +165,7 @@ def gift_cards_create(
     non_shippable_gift_cards = []
     expiry_date = calculate_expiry_date(settings)
     for order_line in gift_card_lines:
+        fulfillment_line = order_line.fulfillment_lines.last()
```
tomaszszymanski129, line 168: maybe we should fetch related lines once rather than calling these .last()s on each iteration? I'm assuming this are not cached by django and create extra queries, but that would have to be checked.
Quick solution would be, before the loop:
```
fulfillment_lines = FulfillmentLine.objects.filter(order_line_pk__in=gift_card_lines).order_by("-pk")
```
and inside something like
```
fulfillment_line = next(filter(lambda f: f.order_line_id == order_line.id, fulfillment_lines), None)
```
On the other hand maybe thats not an issue at all, seems unlikely that user will ever buy lots of these at once, right?

tomaszszymanski129 (approved):
It would be nice to have in PR description why we'll need this extra relation.

saleor/giftcard/utils.py
IKarbowiak: Yes, I think that in most cases it will be only one gift card line. But maybe it's a good idea to introduce a solution with a constant query amount.

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -277,6 +266,17 @@ def perform_mutation(cls, _root, info, order, **data):
             errors = prepare_insufficient_stock_order_validation_errors(exc)
             raise ValidationError({"stocks": errors})
 
+        if approved:
```
korycins, line 269: What do you think about moving this to saleor/order/actions.py file. There is a method `saleor.order.actions.order_fulfilled` Which is triggered by this mutation. It contains actions like this one so I would consider moving this action there.

korycins (approved):

PR #8265, opened 2021-10-04
Fix webhooks logging of excessive events

tomaszszymanski129:
I want to merge this change because it limits logging of send_webhook_request - it no longer logs error, it handles failure and logs on debug level that it happened.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/plugins/webhook/tasks.py
+++ b/saleor/plugins/webhook/tasks.py
@@ -159,21 +159,30 @@ def send_webhook_request(webhook_id, target_url, secret, event_type, data):
     message = data.encode("utf-8")
     signature = signature_for_payload(message, secret)
     if parts.scheme.lower() in [WebhookSchemes.HTTP, WebhookSchemes.HTTPS]:
-        send_webhook_using_http(target_url, message, domain, signature, event_type)
+        try:
+            send_webhook_using_http(target_url, message, domain, signature, event_type)
+        except RequestException as e:
+            logger.debug("[Webhook] Failed request to %r: %r.", target_url, e)
     elif parts.scheme.lower() == WebhookSchemes.AWS_SQS:
         send_webhook_using_aws_sqs(target_url, message, domain, signature, event_type)
+        task_logger.debug(
+            "[Webhook ID:%r] Payload sent to %r for event %r",
+            webhook_id,
+            target_url,
+            event_type,
+        )
     elif parts.scheme.lower() == WebhookSchemes.GOOGLE_CLOUD_PUBSUB:
         send_webhook_using_google_cloud_pubsub(
             target_url, message, domain, signature, event_type
         )
+        task_logger.debug(
+            "[Webhook ID:%r] Payload sent to %r for event %r",
+            webhook_id,
+            target_url,
+            event_type,
+        )
     else:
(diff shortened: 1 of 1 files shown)


saleor/plugins/webhook/tasks.py
```diff
@@ -159,21 +159,30 @@ def send_webhook_request(webhook_id, target_url, secret, event_type, data):
     message = data.encode("utf-8")
     signature = signature_for_payload(message, secret)
     if parts.scheme.lower() in [WebhookSchemes.HTTP, WebhookSchemes.HTTPS]:
-        send_webhook_using_http(target_url, message, domain, signature, event_type)
+        try:
+            send_webhook_using_http(target_url, message, domain, signature, event_type)
+        except RequestException as e:
+            logger.debug("[Webhook] Failed request to %r: %r.", target_url, e)
```
korycins, line 165: I don't think that this should be a debug. We will lose information that someone has failed webhook integration.  Currently it is the only way to check this. With debug, we will lose this possibility. 
Also, the celery task uses autoretry_for `RequestException`, which you're handling here, I am not sure if such task will be retried in that case.

NyanKiyoshi (approved):

saleor/plugins/webhook/tasks.py
```diff
@@ -149,31 +151,49 @@ def send_webhook_using_google_cloud_pubsub(
...
+    }
+
+    if methods := scheme_matrix.get(parts.scheme.lower()):
+        send_method, send_exception = methods
+        try:
+            send_method(target_url, message, domain, signature, event_type)
+        except send_exception as e:
+            task_logger.debug("[Webhook] Failed request to %r: %r.", target_url, e)
+            try:
+                self.retry(countdown=10)
```
korycins, line 182: Will this also handle a configuration from `@app.task`?  You provided here a countdown and I am not sure if this will not overwrite a current configuration (especially a retry_backoff)

saleor/plugins/webhook/tasks.py
```diff
@@ -149,31 +151,49 @@ def send_webhook_using_google_cloud_pubsub(
...
+            pubsub_v1.publisher.exceptions.MessageTooLargeError,
+        ),
+    }
+
+    if methods := scheme_matrix.get(parts.scheme.lower()):
+        send_method, send_exception = methods
+        try:
+            send_method(target_url, message, domain, signature, event_type)
+        except send_exception as e:
+            task_logger.debug("[Webhook] Failed request to %r: %r.", target_url, e)
```
korycins, line 180: Should it be a `debug` here? We will not have any logs for log lvl > INFO

tomaszszymanski129: there is an info if it fails to deliver after retries (on `MaxRetriesExceededError`). I think it's fine.

saleor/plugins/webhook/tasks.py
tomaszszymanski129: yes, it ignored what we've passed in backoff and kwargs. Sadly retry func cannot accept backoff as arg, but I've implemented countdown calculation the same way celery does (based on passed backoff). Also kwargs are properly passed now so it seems all covered. I've tested it both with both task_always_eager settings.

PR #8273, opened 2021-10-04
Batch entites by type in federated queries

rafalp:
When Apollo-Federation executes federated query, it queries `_enties` field on `Query` type with a list of entities to resolve. Currently each of entities is resolved separately from rest, so if there's 100 entities queried, this will call `__resolve_entity` 100 times which in turn will run 100 database queries.

This PR replaces default `_entities` resolver with new one that batches loads per entity type. `graphene-federation`'s `__resolve_entity` is replaced by `__resolve_entities` that is expected to return list of items.

`FEDERATED_QUERY_MAX_ENTITIES` setting is included that allows imposing hard limit on number of entities that may be queried for at single time. Defaults to 300, but can be disabled by setting `FEDERATED_QUERY_MAX_ENTITIES=0` in env, or increasing accordingly.

_Note:_ I will provide documentation for federation implementation in Saleor as a separate task.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/graphql/account/types.py
```diff
@@ -378,28 +389,40 @@ def resolve_wishlist(root: models.User, info, **_kwargs):
         return resolve_wishlist_items_from_user(root)
 
     @staticmethod
-    def __resolve_reference(root: "User", info, **_kwargs):
-        User = get_user_model()
+    def __resolve_references(roots: List["User"], info, **_kwargs):
+        requestor = get_user_or_app_from_context(info.context)
+        requestor_has_access_to_all = has_one_of_permissions(
```
L3str4nge, line 394: Has access to all but checking if has only one of given?

rafalp:
@L3str4nge entity resolver for federation quietly skips entities that don't define `__entities_resolver`. This is up to federation gateway that's level above to resolve and handle missing entities.

rafalp:
I'm sending this back to draft, I'll need to update this PR to include two more federated types that for some reason don't have any entity resolvers (`PageType` and `Group`). Should be quick fix for tomorrow.

saleor/settings.py
```diff
@@ -528,6 +528,12 @@ def SENTRY_INIT(dsn: str, sentry_opts: dict):
     ],
 }
 
+# Max number entities that can be requested in single query by Apollo Federation
+# Federation protocol implements no securities on its own part - malicious actor
+# may build a query that requests for potentially few thousands of entities.
+# Set FEDERATED_QUERY_MAX_ENTITIES=0 in env to disable (not recommended)
+FEDERATED_QUERY_MAX_ENTITIES = int(os.environ.get("FEDERATED_QUERY_MAX_ENTITIES", 300))
```
maarcingebala, line 535: Maybe this should be configured to the same value as `RELAY_CONNECTION_MAX_LIMIT`? If we allow max 100 to be returned there, I think it's a reasonable value here as well.

saleor/graphql/product/types/products.py
```diff
@@ -612,6 +591,39 @@ def calculate_global_sold_units(variant_channel_listings):
...
+            )
+            if root_id:
+                roots_ids.append(f"{root.channel}_{root_id}")
+                channels[root.channel].add(root_id)
+
+        variants = {}
+        for channel, ids in channels.items():
+            visible_products = models.Product.objects.visible_to_user(
+                requestor, channel
+            ).values_list("pk", flat=True)
```
maarcingebala, line 614: Could we use `resolve_product_variants` here somehow? If possible we shouldn't duplicate the resolver logic here, especially the logic responsible for limiting visibility of products/variants to users based on their permission/ staff status etc.

saleor/graphql/page/types.py
```diff
@@ -108,3 +111,13 @@ def resolve_has_pages(root: models.PageType, info, **kwargs):
...
             .then(lambda pages: bool(pages))
         )
+
+    @staticmethod
+    def __resolve_references(roots: List["PageType"], info, **_kwargs):
+        ids = [
+            int(from_global_id_or_error(root.id, PageType, raise_error=True)[1])
+            for root in roots
+        ]
+        qs = models.PageType.objects.filter(id__in=ids)
```
maarcingebala, line 121: Pages visibility depends on user's permission, see the `resolve_pages` resolver -it uses `visible_for_user` method. Could we reuse the resolver here? If needed I would adjust the resolver to accept an optional list of IDs. Ideally those resolver function should be the only places where we can easily maintain this user-specific visibility logic.

maarcingebala (commented):
It looks good overall but I think we should try to reuse existing resolvers in __resolve_references methods. If needed I would adjust existing resolvers so that we can reuse them.

saleor/graphql/page/types.py
maarcingebala: Nevermind, I was thinking about Page model and not PageType. This resolver is fine.

fowczarek:
Could we add benchmarks for those new queries?

PR #8285, opened 2021-10-05
Add basic implementation for NP Atobarai gateway

mateuszgrzyb:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,209 @@
...
+
+def _format_name(ad: AddressData):
+    """Follow the japanese name guidelines."""
+    return f"{ad.first_name} {ad.last_name}".lstrip().rstrip()
+
+
+def _format_address(ad: AddressData):
+    """Follow the japanese address guidelines."""
+    # example: "東京都千代田区麹町４－２－６　住友不動産麹町ファーストビル５階"
+    with Posuto() as pp:
```
stnatic, line 66: @mateuszgrzyb @korycins maybe we should put it behind a feature flag in plugin configuration for now?

saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,209 @@
...
+                    price_to_minor_unit(
+                        payment_information.amount, payment_information.currency
+                    )
+                ),
+                "customer": {
+                    "customer_name": billing.first_name,
+                    "company_name": billing.company_name,
+                    "zip_code": billing.postal_code,
+                    "address": _format_address(billing),
+                    "tel": billing.phone.replace("+81", "0"),
```
stnatic, line 116: Is phone mandatory? I think we should have an extra validation higher up that ensures that:
1. The phone number is provided
2. It is a Japanese phone number

saleor/payment/interface.py
```diff
@@ -82,6 +90,15 @@ class PaymentData:
...
     data: Optional[dict] = None
     graphql_customer_id: Optional[str] = None
+    # Optional, lazy-evaluated gateway arguments
+    _resolve_lines: InitVar[Callable] = None
+
+    def __post_init__(self, _resolve_lines: Callable):
+        self.__resolve_lines = _resolve_lines
+
+    @cached_property
+    def lines(self) -> List[PaymentLineData]:
```
stnatic, line 100: FYI - this is not visible to `dataclasses.asdict` so this field does not appear in webhook payloads. Maybe it would be worth leaving a comment about this.

saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+    if not shipping:
+        raise PaymentError(
+            "Shipping address is required for transaction in NP Atobarai."
+        )
+
+    data = {
+        "transactions": [
+            {
+                "shop_transaction_id": payment_information.payment_id,
```
stnatic, line 103: We should check whether NP does not mention this information to the end customer - in that case we should use a more meaningful reference (e.g. the one that is used as a credit card statement for Stripe, while being unique at the same time)

saleor/payment/gateways/np_atobarai/api.py
mateuszgrzyb: checkoutCreate is already validating the phone number

saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,209 @@
...
+    if "results" in response_data:
+        transaction = response_data["results"][0]
+        status = transaction["authori_result"]
+        transaction_id = transaction["np_transaction_id"]
+        errors = []
+
+        if status == PaymentStatus.PENDING:
+            cancel_error_codes = _cancel(config, transaction_id)
+            if cancel_error_codes:
+                logger.error(
```
korycins, line 154: Is it something that should be logged as an error? This will be sent to the sentry. Not sure if this is something that we should log here. Maybe warning would be better 🤔  as this is an expected case that some transactions can be canceled.

saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,209 @@
...
+        error_messages = get_error_messages_from_codes(
+            error_codes, TransactionCancellationResultError
+        )
+
+        return PaymentResult(
+            status=PaymentStatus.FAILED,
+            psp_reference=psp_reference,
+            errors=error_messages,
+        )
+    else:
```
korycins, line 205: Else is not needed here. `If` call `return`

saleor/payment/gateways/np_atobarai/errors.py
```diff
@@ -0,0 +1,294 @@
...
+
+    E0100132 = "Please confirm that the payment method is as expected."
+
+
+def get_error_messages_from_codes(
+    error_codes: Iterable[str], error_enum_cls: Type[Enum]
+) -> List[str]:
+    error_messages = []
+    for code in error_codes:
+        try:
```
korycins, line 215: Wouldn't be better to instead of having error cls, we could have a dict map? This could drop this try/except as you could use .get(code, default_error)

saleor/payment/gateways/np_atobarai/plugin.py
```diff
@@ -0,0 +1,149 @@
...
+
+GATEWAY_NAME = "NP後払い"
+
+if TYPE_CHECKING:
+    # flake8: noqa
+    from ....plugins.models import PluginConfiguration
+    from . import GatewayResponse, PaymentData
+
+
+__all__ = ["NPAtobaraiGatewayPlugin"]
```
korycins, line 19: Why do you need this?

saleor/payment/gateways/np_atobarai/tests/test_plugin_configuration.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@mock.patch("saleor.payment.gateways.np_atobarai.api.requests.request")
+def test_validate_plugin_configuration_missing_data(mocked_request, np_atobarai_plugin):
+    # given
+    plugin = np_atobarai_plugin()
+    response = Mock()
+    response.status_code = 200
+    mocked_request.return_value = response
+    plugin_configuration = PluginConfiguration.objects.get()
```
korycins, line 56: Do you need this? You could set None values via `np_atobarai_plugin`

saleor/payment/utils.py
```diff
@@ -1,31 +1,93 @@
...
+                    gross=unit_gross,
+                )
+            )
+
+    elif order:
+        for order_line in order.lines.all():
+            line_items.append(
+                PaymentLineData(
+                    quantity=order_line.quantity,
+                    description=order_line.translated_product_name,
```
korycins, line 83: Is it expected that we have two different values which can be passed to this field? 
`f"{line_info.variant.product.name}, {line_info.variant.name}"`
and 
`order_line.translated_product_name`

saleor/payment/gateways/np_atobarai/api.py
mateuszgrzyb: This error is logged only if NP doesn't cancel the payment that is in pending state. Pending payment needs to be resolved either by cancelling it, or by contacting NP customer support. Leaving that payment in pending state indefinitely is an error, so I think it's good idea to log it as error.

saleor/payment/gateways/np_atobarai/errors.py
```diff
@@ -0,0 +1,233 @@
+from typing import Dict, Iterable, List
+
+UNKNOWN_ERROR = "Unknown error while processing the payment."
+
+
+TRANSACTION_REGISTRATION_RESULT_ERRORS = {
+    "E0100048": "Please make sure that the your (Kanji) has been entered.",
```
maarcingebala, line 7: "that the your" - is this correct?

saleor/payment/gateways/np_atobarai/errors.py
```diff
@@ -0,0 +1,233 @@
...
+    "E0100049": (
+        "Please confirm that your name(Kanji) does not include prohibited characters."
```
maarcingebala, line 9:
```suggestion
        "Please confirm that your name (Kanji) does not include prohibited characters."
```

saleor/payment/gateways/np_atobarai/errors.py
```diff
@@ -0,0 +1,233 @@
...
+        "does not include prohibited characters."
+    ),
+    "E0100057": "Please check if the customer’s ZIP code has been entered.",
+    "E0100058": (
+        "Please confirm that the customer’s ZIP code "
+        "does not contain prohibited characters."
+    ),
+    "E0100136": (
+        "Please check if the customer’s ZIP code "
+        "format is correct.(Example: 1070052, 107 - 0052)"
```
maarcingebala, line 40:
```suggestion
        "format is correct. (Example: 1070052, 107 - 0052)"
```

saleor/payment/gateways/np_atobarai/errors.py
```diff
@@ -0,0 +1,233 @@
...
+        "Please confirm that your name(Kanji) does not include prohibited characters."
+    ),
+    "E0100050": (
+        "Please confirm that your name(Kanji) is within 21 full - width characters."
+    ),
+    "E0100051": (
+        "Please confirm that your name(Kana) is within 25 full - width characters."
+    ),
+    "E0100052": (
+        "Please confirm that your name(kana) does not contain prohibited characters."
```
maarcingebala, line 18: Not sure if it should be "Kana" or "kana" but it should be consistent.

saleor/payment/gateways/np_atobarai/errors.py
```diff
@@ -0,0 +1,233 @@
...
+    "E0100062": (
+        "Please make sure that the Customer’s address "
+        "is within 55 full - width characters."
+    ),
+    "E0100063": "Please check if the customer’s phone number has been entered.",
+    "E0100064": (
+        "Please confirm that the customer’s phone number "
+        "does not contain prohibited characters."
+    ),
+    "E0100065": "Please confirm that the formant of the phone number is valid.",
```
maarcingebala, line 57:
```suggestion
    "E0100065": "Please confirm that the format of the phone number is valid.",
```

maarcingebala: I would recommend checking these error messages with Grammarly or some other tool like this to fix grammar issues.

saleor/payment/utils.py
```diff
@@ -1,31 +1,94 @@
...
 
 
+def create_payment_lines_information(
+    payment: Payment,
+) -> List[PaymentLineData]:
+    checkout = payment.checkout
+    order = payment.order
+    line_items = []
+    if checkout:
+        manager = get_plugins_manager()
```
maarcingebala, line 43: Can we pass the manager as argument to this function? We would have to also pass it to `create_payment_information`, but I see that functions that call `create_payment_information` already have the manager in their signatures.

saleor/payment/utils.py
```diff
@@ -1,31 +1,94 @@
...
+        lines = fetch_checkout_lines(checkout)
+        discounts = fetch_active_discounts()
+        checkout_info = fetch_checkout_info(checkout, lines, discounts, manager)
```
maarcingebala, line 46: Do we know how will this affect checkout performance (in terms of database query performance)? For example in `checkoutComplete` mutation we already have this data fetched at the beginning of the mutation, and here we would duplicate these queries I think.

saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,237 @@
...
+def health_check(config: ApiConfig) -> bool:
+    try:
+        response = _request(config, "post", "/authorizations/find")
+        return response.status_code not in [401, 403]
+    except PaymentError:
+        return False
+
+
+def _format_name(ad: AddressData):
+    """Follow the japanese name guidelines."""
```
maarcingebala, line 70:
```suggestion
    """Follow the Japanese name guidelines."""
```

saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,237 @@
...
+    """Follow the japanese address guidelines."""
+    # example: "東京都千代田区麹町４－２－６　住友不動産麹町ファーストビル５階"
+    if not config.fill_missing_address:
+        return f"{ad.country_area}" f"{ad.street_address_2}" f"{ad.street_address_1}"
+    with Posuto() as pp:
+        try:
+            jap_ad = pp.get(ad.postal_code)
+        except KeyError:
+            raise PaymentError(
+                "Valid japanese address is required for transaction in NP Atobarai."
```
maarcingebala, line 84:
```suggestion
                "Valid Japanese address is required for transaction in NP Atobarai."
```

saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+
+def inject_api_config(fun):
+    def inner(payment_information: PaymentData, config: GatewayConfig):
+        return fun(payment_information, get_api_config(config.connection_params))
+
+    return inner
+
+
+def parse_errors(errors: List[str]) -> str:
+    # FIXME: better solution?
```
maarcingebala, line 20: What about this? If it's not going to be fixed in this PR do we have a task for it?

mateuszgrzyb: It's not going to be fixed in this PR, I've added a task in Net Protections task list.

L3str4nge:
Please fix the mypy errors as well :)

PR #8289, opened 2021-10-05
Allow creating checkout without email

IKarbowiak:
Do not require `email` for `checkoutCreate` mutation.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-3143 Allow to create checkouts without email](https://app.clickup.com/t/2549495/SALEOR-3143)

saleor/checkout/models.py
```diff
@@ -46,7 +46,7 @@ class Checkout(ModelWithMetadata):
         related_name="checkouts",
         on_delete=models.CASCADE,
     )
-    email = models.EmailField()
+    email = models.EmailField(blank=True, default="")
```
maarcingebala, line 49: Maybe instead of `default=""` we should use `null=True` so that we have explicit null in the database when email is not set, and we don't return an empty string in API, but null.

saleor/checkout/utils.py
```diff
@@ -492,9 +492,10 @@ def add_promo_code_to_checkout(
             manager, checkout_info, lines, promo_code, discounts
         )
     elif promo_code_is_gift_card(promo_code):
+        use_email = cast(str, checkout_info.get_customer_email())
```
fowczarek, line 495: :police_officer: 
```suggestion
        user_email = cast(str, checkout_info.get_customer_email())
```

saleor/graphql/checkout/mutations.py
```diff
@@ -989,13 +989,27 @@ class Meta:
         error_type_class = CheckoutError
         error_type_field = "checkout_errors"
 
+    @staticmethod
+    def clean_email(email):
```
fowczarek, line 993: Should we add similar validation in checkout complete? Maybe It's planed for separate PR.

fowczarek (approved):

saleor/graphql/checkout/mutations.py
IKarbowiak: It will be done in separate PR :)

korycins (approved):

PR #8334, opened 2021-10-08
Improve product_variant filters performance

CossackDex:
I want to merge this change because it improves product_variant filters performance.

search/name filter
		before -> https://explain.dalibo.com/plan/I8a
	        after -> https://explain.dalibo.com/plan/Thy
	
search/product_name 
		before -> https://explain.dalibo.com/plan/DJ3
		after -> https://explain.dalibo.com/plan/uqZW
	
search/sku
		before -> https://explain.dalibo.com/plan/CYy
		after -> https://explain.dalibo.com/plan/Ggc
	
sku/sku
		before -> https://explain.dalibo.com/plan/TUw
		after -> https://explain.dalibo.com/plan/RbU

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


tomaszszymanski129:
New gin index is not being used at all even after running the migrations ⚠️ 

Query which is produced by this PR: https://explain.dalibo.com/plan/Ety
usage of the subquery (or rather condition inside) convinces postgres to do the sequence scan.
even if we modify the query manually to use inner join instead of distinct, it still don't use the index (and is much slower on multiple runs that i've tested):
https://explain.dalibo.com/plan/fKm
If we skip the subquery / inner join, only then the index is being used:
https://explain.dalibo.com/plan/z6oh

This concludes the index is useless in our case and should be dropped.

tomaszszymanski129:
Dropped unused index since postgres uses different ones anyway, rewriting to subquery gains up to 30% on my test DB.

stnatic:
Some of my thoughts on it, the test queries are eventually executing:
```
ORDER BY
       product_productvariant.sort_order DESC,
       product_productvariant.sku DESC
```

The database is going to sort all of the results in memory, unless we have a multi column, that matches all of:
- the search fields
- `sort_order`
- `sku`

I can expect that there will be use cases to sort using a different method, e.g. by the price or by the stock quantity and we probably won't be providing an index for each of those.

So I'm expecting a perfect database plan to look like this:
- we index-filter all variants matching the query by the * product * attribute
- we index-filter all variants matching the query by the * variant * attribute
- hopefully this returns a relatively small number of rows (let's say a 1000, rather than 30k which can still happen if we are searching for "product_name" where all product names were generated by the sequence such as "product_name_{idx}")
- the database sorts the result and returns first X rows

This makes me think that for restrictive queries our best bet would be to keep the gin index, but run the products and the variants queries in an `union` statement as that's almost surely going to use an index. It might be an overkill, but I thought I'll give you some ideas ;)

tomaszszymanski129:
Indexes will be added as a part of separate ticket.

PR #8399, opened 2021-10-19
Exclude shipping methods for 3.0

korycins:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/plugins/webhook/plugin.py
```diff
@@ -417,3 +426,71 @@ def process_payment(
...
+    def excluded_shipping_methods_for_order(
+        self,
+        order: "Order",
+        available_shipping_methods: List[ShippingMethod],
+        previous_value: List[ExcludedShippingMethod],
+    ) -> List[ExcludedShippingMethod]:
+        payload = generate_excluded_shipping_methods_for_order_payload(
+            order,
+            available_shipping_methods,
+        )
```
fowczarek, line 475: @kczan 
We shouldn't generate payload if there is no active exclude shipping method webhook.

saleor/plugins/webhook/plugin.py
```diff
@@ -417,3 +426,71 @@ def process_payment(
...
+            payload=payload,
+        )
+
+    def excluded_shipping_methods_for_checkout(
+        self,
+        checkout: "Checkout",
+        available_shipping_methods: List[ShippingMethod],
+        previous_value: List[ExcludedShippingMethod],
+    ) -> List[ExcludedShippingMethod]:
+        payload = generate_excluded_shipping_methods_for_checkout_payload(
```
fowczarek, line 488: @kczan 
We shouldn't generate payload if there is no active exclude shipping method webhook.

saleor/webhook/payloads.py
```diff
@@ -535,6 +537,42 @@ def generate_list_gateways_payload(
...
+    payload = asdict(method)
+    payload["id"] = graphene.Node.to_global_id("ShippingMethod", payload["id"])
+    return payload
+
+
+def generate_excluded_shipping_methods_for_order_payload(
+    order: "Order",
+    available_shipping_methods: List[ShippingMethod],
+):
+    order_data = json.loads(generate_order_payload(order))[0]
```
fowczarek, line 550: @kczan 
Are you sure that all required data are prefetched? 
We should add a benchmark for this case(Remember to use multiline order).

saleor/webhook/payloads.py
```diff
@@ -535,6 +537,42 @@ def generate_list_gateways_payload(
...
+        ],
+    }
+    return json.dumps(payload, cls=CustomJsonEncoder)
+
+
+def generate_excluded_shipping_methods_for_checkout_payload(
+    checkout: "Checkout",
+    available_shipping_methods: List[ShippingMethod],
+):
+    checkout_data = json.loads(generate_checkout_payload(checkout))[0]
```
fowczarek, line 565: @kczan 
Are you sure that all required data are prefetched?
We should add a benchmark for this case(Remember to use multiline order).

saleor/plugins/webhook/plugin.py
fowczarek: We shoud alse add tracing fot that. 
`import opentracing`
```suggestion
        with opentracing.global_tracer().start_active_span("generate_excluded_shipping_methods_for_order_payload"):
            payload = generate_excluded_shipping_methods_for_order_payload(
                order,
                available_shipping_methods,
            )
```

saleor/plugins/webhook/plugin.py
```diff
@@ -417,3 +426,71 @@ def process_payment(
...
+            checkout,
+            available_shipping_methods,
+        )
```
fowczarek, line 491: We shoud alse add tracing fot that. 
`import opentracing`
```suggestion
        with opentracing.global_tracer().start_active_span("generate_excluded_shipping_methods_for_checkout_payload"):
            payload = generate_excluded_shipping_methods_for_checkout_payload(
                checkout,
                available_shipping_methods,
            )
```

patrys (dismissed):
Please address the performance concerns raised by @fowczarek

stnatic:
@kczan please start by preparing the benchmark test first. We might reuse some of the code introduced in 3.1 that fetches available shipping methods in a dataloader

fowczarek (approved):

saleor/graphql/shipping/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    shipping_methods: List[shipping_models.ShippingMethod],
+    excluded_methods: List[ExcludedShippingMethod],
+):
+    for instance in shipping_methods:
+        instance.active = True  # type: ignore
+        instance.message = ""  # type: ignore
+        for method in excluded_methods:
+            if str(instance.id) == str(method.id):
+                instance.active = False  # type: ignore
+                instance.message = method.reason  # type: ignore
```
maarcingebala, line 63: @stnatic Assigning fields to an instance on the fly is not a practice that we do in Saleor - this will have to be refactored later - please make sure that you have it written down in a task with things to refactor. CC @patrys just to make sure that you saw this.

maarcingebala (approved):

saleor/graphql/shipping/utils.py
stnatic: @maarcingebala This is already changed on 3.1 where all delivery method roots are instances of the ShippingMethodData. I agree that it would be a good idea backport some of that functionality to 3.0.

Please note that Saleor has been already doing this to add fields of `ShippingMethodChannelListing` to a `ShippingMethod` when displaying it in a context of a specific channel, so we didn't introduce a new convention here:
https://github.com/saleor/saleor/pull/8399/files#diff-473baa60c7b2166e11d93921bf9fee5243a14adb9198e5ddc452d1cc8776ec6cL407

maarcingebala: It doesn't change the fact that this is a bad convention and should be avoided; we should probably change the place you linked as well.

stnatic: Fully agreed! My recommendation would be to first approve the solution on 3.1 and backport the `ShippingMethodData` interface onto 3.0 as well. This way we will make sure that the shipping logic in both doesn't diverge even further

korycins: As we discussed a couple of times. The main priority is to finish the changes from 3.0, and apply the requested changes from the previous PRs which we converted to task.  We don't think about 3.1 yet as we have limited time to finish 3.0.

PR #8440, opened 2021-10-25
Add permission_manage_orders to customer query

CossackDex:
I want to merge this change because it adds `permission_manage_orders` to the `customer` query. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/account/schema.py
+++ b/saleor/graphql/account/schema.py
@@ -1,6 +1,6 @@
 import graphene
 
-from ...core.permissions import AccountPermissions
+from ...core.permissions import AccountPermissions, OrderPermissions
 from ..core.fields import FilterInputConnectionField
 from ..core.types import FilterInputObjectType
 from ..core.utils import from_global_id_or_error
@@ -157,7 +157,9 @@ class AccountQueries(graphene.ObjectType):
             city_area=city_area,
         )
 
-    @permission_required(AccountPermissions.MANAGE_USERS)
+    @one_of_permissions_required(
+        [OrderPermissions.MANAGE_ORDERS, AccountPermissions.MANAGE_USERS]
+    )
     def resolve_customers(self, info, **kwargs):
         return resolve_customers(info, **kwargs)
 
--- a/saleor/graphql/account/schema.py
+++ b/saleor/graphql/account/schema.py
@@ -181,7 +181,11 @@ class AccountQueries(graphene.ObjectType):
         return resolve_staff_users(info, **kwargs)
 
     @one_of_permissions_required(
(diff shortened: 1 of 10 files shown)


saleor/graphql/account/schema.py
```diff
@@ -157,7 +157,9 @@ def resolve_address_validation_rules(
             city_area=city_area,
         )
 
-    @permission_required(AccountPermissions.MANAGE_USERS)
+    @one_of_permissions_required(
+        [OrderPermissions.MANAGE_ORDERS, AccountPermissions.MANAGE_USERS]
+    )
```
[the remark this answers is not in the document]
maarcingebala: We don't have a dedicated endpoint for a single "customer".

saleor/graphql/order/types.py
```diff
@@ -212,7 +212,9 @@ def _resolve_user(event_user):
     @staticmethod
     def resolve_app(root: models.OrderEvent, info):
         requestor = get_user_or_app_from_context(info.context)
-        if requestor_has_access(requestor, root.user, AppPermission.MANAGE_APPS):
+        if requestor_has_access(
+            requestor, root.user, AppPermission.MANAGE_APPS
+        ) | requestor_has_access(requestor, root.user, OrderPermissions.MANAGE_ORDERS):
```
maarcingebala, line 217: This will give access to the entire App object to a user with `MANAGE_ORDERS`. Instead, we should allow the "order manager" to query apps from order events, but limit the fields they can ask for by adding permission checks inside the `App` type. I think order manager (or even all staff users) should have access to fields like `name`, `appUrl`, `homepageUrl`, `supportUrl`, `aboutApp`. Fields like `permissions`, `tokens`, `webhooks` etc should require `MANAGE_APPS` permission.

maarcingebala: @patrys Do you agree? We should do the same for `Order.user` field.

IKarbowiak (approved):

saleor/graphql/account/utils.py
```diff
@@ -463,7 +463,7 @@ def look_for_permission_in_users_with_manage_staff(
 
 
 def requestor_has_access(
-    requestor: Union["User", "App"], owner: Optional["User"], perm
+    requestor: Union["User", "App"], owner: Optional["User"], *perms
```
maarcingebala, line 466: Please also update the docstring `perm` -> `perms`. We should also add some description that this function expects **one of** permissions to be assigned to return true.

fowczarek (approved):

maarcingebala (approved):

eCzerniak (approved):

PR #8477, opened 2021-10-29
Prevent excluded shipping method being assigned to order

kczan:
This task fixes an unwanted behaviour when orderUpdateShipping would allow an excluded shipping method to be assigned to the order.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/mutations/orders.py
+++ b/saleor/graphql/order/mutations/orders.py
@@ -334,11 +334,17 @@ class OrderUpdateShipping(EditableOrderValidationMixin, BaseMutation):
                 "postal_code_rules"
             ),
         )
-
-        shipping_price = info.context.plugins.calculate_order_shipping(order)
-        clean_order_update_shipping(order, method, shipping_price, info.context.plugins)
+        shipping_channel_listing = (
+            shipping_models.ShippingMethodChannelListing.objects.filter(
+                shipping_method=method, channel=order.channel
+            ).first()
+        )
+        clean_order_update_shipping(
+            order, method, shipping_channel_listing.price, info.context.plugins
+        )
 
         order.shipping_method = method
+        shipping_price = info.context.plugins.calculate_order_shipping(order)
         order.shipping_price = shipping_price
         order.shipping_tax_rate = info.context.plugins.get_order_shipping_tax_rate(
             order, shipping_price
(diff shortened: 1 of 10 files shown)


korycins (approved):

saleor/graphql/order/mutations/orders.py
```diff
@@ -314,10 +335,10 @@ def perform_mutation(cls, _root, info, **data):
             ),
         )
 
-        clean_order_update_shipping(order, method)
+        shipping_price = info.context.plugins.calculate_order_shipping(order)
```
IKarbowiak, line 338: We cannot calculate the shipping price and assign it to the `order.shipping_price` field before assigning `shipping_method`, because some plugins check if `shipping_method` is assigned to order (e.g. vatlayer).

So if the shipping method is assigned to order for the first time, we will get the wrong `shipping_price` value.

IKarbowiak (changes requested):

saleor/graphql/order/mutations/orders.py
kczan: @IKarbowiak 
Ok, I've switched it back to how it was. Also added a test.

IKarbowiak (approved):

kczan:
I've added an additional test and switched plugin mocks to a mock on PluginsManager.

saleor/graphql/order/tests/test_order.py
```diff
@@ -13,6 +13,8 @@
 from measurement.measures import Weight
 from prices import Money, TaxedMoney
 
+from saleor.plugins.base_plugin import ExcludedShippingMethod
```
korycins, line 16: Absolute.

saleor/graphql/order/mutations/orders.py
```diff
@@ -50,14 +53,19 @@
 ORDER_EDITABLE_STATUS = (OrderStatus.DRAFT, OrderStatus.UNCONFIRMED)
 
 
-def clean_order_update_shipping(order, method):
+def clean_order_update_shipping(
+    order: Order,
+    method: ShippingMethod,
```
korycins, line 58: Wrong definition of typing. It points to graphql's type instead of db model.

saleor/graphql/order/mutations/orders.py
```diff
@@ -50,14 +53,19 @@
 ORDER_EDITABLE_STATUS = (OrderStatus.DRAFT, OrderStatus.UNCONFIRMED)
 
 
-def clean_order_update_shipping(order, method):
+def clean_order_update_shipping(
+    order: Order,
```
korycins, line 57: Wrong typing

kczan:
@korycins I've fixed issues mentioned in your comments and added validation in DraftOrderComplete.

korycins (approved):

PR #8496, opened 2021-11-03
Fix CheckoutLinesUpdate fails to remove lines unavailable for purchase.

IKarbowiak:
Allow deleting checkout line with the use of  `CheckoutLinesUpdate` mutation for unavailable for purchase products.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -146,10 +146,11 @@ def add_variants_to_checkout(
     channel_listings_by_product_id = {cl.product_id: cl for cl in channel_listings}
 
     # check if variants are published
-    for variant in variants:
-        product_channel_listing = channel_listings_by_product_id[variant.product_id]
-        if not product_channel_listing or not product_channel_listing.is_published:
-            raise ProductNotPublished()
+    for variant, quantity in zip(variants, quantities):
+        if quantity > 0:
+            product_channel_listing = channel_listings_by_product_id[variant.product_id]
+            if not product_channel_listing or not product_channel_listing.is_published:
+                raise ProductNotPublished()
 
     variant_ids_in_lines = {line.variant_id: line for line in checkout.lines.all()}
     to_create = []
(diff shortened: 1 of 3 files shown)


patrys:
Task linked: [SALEOR-4829 `checkoutLinesUpdate` fails to remove lines unavailable for purchase](https://app.clickup.com/t/2549495/SALEOR-4829)

saleor/checkout/utils.py
```diff
@@ -146,10 +146,11 @@ def add_variants_to_checkout(
     channel_listings_by_product_id = {cl.product_id: cl for cl in channel_listings}
 
     # check if variants are published
-    for variant in variants:
-        product_channel_listing = channel_listings_by_product_id[variant.product_id]
-        if not product_channel_listing or not product_channel_listing.is_published:
-            raise ProductNotPublished()
+    for variant, quantity in zip(variants, quantities):
```
maarcingebala, line 149: It was here before but since we are touching this - could you check if we could move this entire validation to `clean_input`? `add_variants_to_checkout` is supposed to only save things and validation should happen earlier.

maarcingebala: If we move it, we should be able to remove `try..except` block from `clean_input` that catches  `ProductNotPublished` errors.

korycins (approved):

saleor/checkout/utils.py
```diff
@@ -134,23 +134,6 @@ def add_variants_to_checkout(
     Otherwise, quantity will be added or replaced (if replace argument is True).
     """
 
-    # check quantities
-    country_code = checkout.get_country()
-    if not skip_stock_check:
```
maarcingebala, line 139: `skip_stock_check` becomes unnecessary after this change and could be removed from the signature.

maarcingebala (approved):

PR #8526, opened 2021-11-05
Add GraphQL query cost validation

rafalp:
Adds GraphQL query cost validation. Limit is configurable via `GRAPHQL_QUERY_MAX_COMPLEXITY` setting/env variable, and complexity values via `saleor.graphql.query_cost_map.COST_MAP` dict.

Currently query complexity validation is disabled outside of few tests. When I was testing it locally, only one test failed, due to assert for error message from GraphQL's query validation stack, which changes a little when query cost is included.

Currently cost limit validation is disabled. We should set complexity costs on schema before enabling it.

When validation is enabled, current and max query costs are included in query result, under `extensions.cost` path:

```json
  "extensions": {
    "cost": {
      "requestedQueryCost": 0,
      "maximumAvailable": 100
    }
  }
```

Needs documenting `GRAPHQL_QUERY_MAX_COMPLEXITY` in settings reference.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


IKarbowiak (approved):

saleor/graphql/core/validators/query_cost.py
```diff
@@ -0,0 +1,300 @@
...
+        self.cost = 0
+        self.operation_multipliers: List[Any] = []
+
+    def __call__(self, context: ValidationContext):
+        self.context = context
+        return self
+
+    def compute_node_cost(self, node: CostAwareNode, type_def, parent_multipliers=None):
+        if parent_multipliers is None:
+            parent_multipliers = []
```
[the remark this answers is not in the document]
rafalp: This is a mistake in python and results in function's argument being singleton passed around:

[Common Gotchas: Mutable Default Arguments](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)

saleor/settings.py
```diff
@@ -532,6 +532,9 @@ def SENTRY_INIT(dsn: str, sentry_opts: dict):
     ],
 }
 
+# Set GRAPHQL_QUERY_MAX_COMPLEXITY=0 in env to disable (not recommended)
+GRAPHQL_QUERY_MAX_COMPLEXITY = int(os.environ.get("GRAPHQL_QUERY_MAX_COMPLEXITY", 0))
```
maarcingebala, line 536: Since 0 is not recommended, we could set it to some meaningful value by default (probably the same value that would be set in the cloud anyway).

saleor/graphql/core/tests/test_query_cost_validation.py
```diff
@@ -0,0 +1,133 @@
...
+
+    response = api_client.post_graphql(query, variables)
+    json_response = response.json()
+    assert "data" not in json_response
+    query_cost = json_response["extensions"]["cost"]["requestedQueryCost"]
+    assert query_cost > 1
+    assert len(json_response["errors"]) == 1
+    assert json_response["errors"][0]["message"] == (
+        f"The query exceeds the maximum cost of 1. Actual cost is {query_cost}"
+    )
```
maarcingebala, line 39: Are we able to also return some error code here? For example, Github API returns a "type" field for this kind of error; when exceeding their max limit you would get:

```
{
  "errors": [
    {
      "type": "MAX_NODE_LIMIT_EXCEEDED",
      "locations": [
        {
          "line": 19,
          "column": 21
        }
      ],
      "message": "By the time this query traverses to the issues connection, it is requesting up to 1,000,000 possible nodes which exceeds the maximum limit of 500,000."
    }
  ]
}
```
I'm wondering if we can achieve something similar.

saleor/settings.py
rafalp: It's too early for that. Devs should first set costs on some of fields based on known usage patterns and then set default query limit.

saleor/graphql/query_cost_map.py
```diff
@@ -0,0 +1,6 @@
+COST_MAP = {
```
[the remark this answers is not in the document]
rafalp: This task was about porting query validation mechanism from Ariadne. We still should do follow up PR that actually configures this feature for real-world™ usage.

I can add documentation for how cost map works, but I'm wondering what you think about making this a docstring in this module, seeing how its internal thing that only Saleor devs and authors of forks will have to deal with it.

saleor/settings.py
maarcingebala: OK sounds good.

saleor/graphql/query_cost_map.py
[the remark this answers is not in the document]
maarcingebala: I meant a docstring for us (Saleor devs and authors) so that we always understand how to use this dict.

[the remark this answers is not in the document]
rafalp: Allright.

rafalp:
Extend costs map to give costs to more fields, and added default query limit of 250, but we should finetune this in future based on our knowledge of slow queries that people are doing.

saleor/graphql/core/tests/test_query_cost_validation.py
rafalp: Not sure. Error from cost validator is reported via `ValidationContext.report_error` which mangles `GraphQLError` instances. I'll give it a second look later today after I'm back from running errands.

fowczarek (approved):

saleor/graphql/core/tests/test_query_cost_validation.py
rafalp: Discussed this on slack and there will be dedicated error code `QueryCostError` in exceptions.

maarcingebala (approved):

PR #8533, opened 2021-11-08
Introduce anonymous checkout

IKarbowiak:
- Allow creating anonymous checkouts - without email
- Add email validation to:
  - addPromoCode 
  - checkoutComplete 
  - checkoutPaymentCreate 
- Wipes old checkouts

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/checkout/complete_checkout.py
```diff
@@ -70,7 +71,8 @@ def _get_voucher_data_for_order(checkout_info: "CheckoutInfo") -> dict:
     if voucher.usage_limit:
         increase_voucher_usage(voucher)
     if voucher.apply_once_per_customer:
-        add_voucher_usage_by_customer(voucher, checkout_info.get_customer_email())
+        customer_email = cast(str, checkout_info.get_customer_email())
```
fowczarek, line 74: What if this checkout is anonymous and we add a promo code with `voucher.apply_once_per_customer=True` to this checkout?

saleor/payment/error_codes.py
```diff
@@ -16,3 +16,4 @@ class PaymentErrorCode(Enum):
     NOT_SUPPORTED_GATEWAY = "not_supported_gateway"
     CHANNEL_INACTIVE = "channel_inactive"
     BALANCE_CHECK_ERROR = "balance_check_error"
+    EMAIL_NOT_SET = "email_not_set"
```
fowczarek, line 19: Maybe we should use `CHECKOUT_EMAIL_NOT_SET`. It's could be confusing when Payment mutation responds that "Email not set" and someone thinks that they should set this email on payment object instead of checkout.

saleor/settings.py
```diff
@@ -470,6 +471,11 @@ def get_host():
     "saleor.core.auth_backend.PluginBackend",
 ]
 
+# Expired checkouts settings - defines after what time checkouts will be deleted
+ANONYMOUS_CHECKOUTS_TIMEDELTA = {"days": 30}
+USER_CHECKOUTS_TIMEDELTA = {"days": 90}
+EMPTY_CHECKOUTS_TIMEDELTA = {"hours": 6}
```
fowczarek, line 477: In other time delta settings we use:
`JWT_TTL_REFRESH = timedelta(seconds=parse(os.environ.get("JWT_TTL_REFRESH", "30 days")))`
In my opinion, we should store it in the same way.

fowczarek (commented):
@IKarbowiak Did you test anonymous checkout flow with active checkout webhooks?

saleor/checkout/complete_checkout.py
IKarbowiak: You cannot add a promo code to checkout without email. You need to assign an email to checkout before. There is also email validation in `CheckoutAddPromoCode` mutation.

eCzerniak (changes requested):
1. addPromoCode removes added shippingMethod
2. if avalara is activated then checkoutComplete cannot be finished.

fowczarek (approved):

IKarbowiak:
@fowczarek Yes, it works correctly.
@eCzerniak, the first problem is a global issue. The problem with the avalara doesn't exist anymore.

tomaszszymanski129 (approved):

PR #8515, opened 2021-11-05
Search by order and checkout lines metadata

iprlic:
I want to merge this change because...

added search by checkout and order line metadata (I'll rebase when previous PR is merged)

# Impact

* [x] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #8512, opened 2021-11-04
Add search by metadata to orders and checkouts

iprlic:
I want to merge this change because...

Add search by metadata to orders and checkouts

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #8543, opened 2021-11-09
Search accounts by address metadata

iprlic:

added searching accounts by address metadata
# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


korycins:
Hey @iprlic  thanks for the PRs that you've created. Can you explain more what do you want to achieve by this? We do not surely understand the case.  The current approach doesn't use any db index which we have on meta fields which means that from a performance perspective this is something that we need to handle in a different way as it will be costly and slowly with bigger DBs.  We treat a `search` as something where staff/user directly provides in the filter so not sure that extending search in this way is the best solution. I would say that we should try to extend a metadata filter not a search filter.
Linking the rest of the PRs, as the answer is related to all of them. https://github.com/saleor/saleor/pull/8515 https://github.com/saleor/saleor/pull/8512

maarcingebala:
Closing for now as we don't want to extend `search` filter with metadata filtering. A separate PR will be opened for adding metadata support to addresses.

PR #8544, opened 2021-11-09
Webhooks improvements for partial payments

stnatic:
- manager from request is being passed onto webhooks (query optimisation)
- fixed order authorize event amount when creating an order (it was captured amount always equal to 0 instead of the authorized amount)
- resolve conflicts with zombie payment solution (changed unit tests)
- count payments with pending status towards order total
- call process_gateway in all payment-related webhooks prior to finalising checkout

Adyen tests:
image

image

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/payment/gateways/stripe/webhooks.py
+++ b/saleor/payment/gateways/stripe/webhooks.py
@@ -158,6 +158,7 @@ def _create_transaction_if_not_exists(
         token=transaction_id,
         kind=kind,
         is_success=True,
+        action_required=False,
     ).exists():
         return None
 
--- a/saleor/payment/gateways/stripe/webhooks.py
+++ b/saleor/payment/gateways/stripe/webhooks.py
@@ -15,7 +15,7 @@ from ....discount.utils import fetch_active_discounts
 from ....order import events
 from ....order.actions import order_captured, order_voided
 from ....order.interface import OrderPaymentAction
-from ....plugins.manager import get_plugins_manager
+from ....plugins.manager import PluginsManager
 from ... import ChargeStatus, TransactionKind
 from ...interface import GatewayConfig, GatewayResponse
 from ...models import Payment
@@ -81,7 +81,12 @@ def handle_webhook(
             "Processing new Stripe webhook",
             extra={"event_type": event.type, "event_id": event.id},
         )
-        webhook_handlers[event.type](event.data.object, gateway_config, channel_slug)
+        webhook_handlers[event.type](
+            event.data.object,
+            gateway_config,
+            channel_slug,
+            request.plugins,  # type: ignore
+        )
     else:
         logger.warning(
             "Received unhandled webhook events", extra={"event_type": event.type}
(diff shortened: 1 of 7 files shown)


saleor/payment/gateways/stripe/webhooks.py
```diff
@@ -158,6 +130,15 @@ def _finalize_checkout(
...
+def _get_checkout(payment_id: int) -> Optional[Checkout]:
+    return (
+        Checkout.objects.prefetch_related("payments")
+        .select_for_update(of=("self",))
+        .filter(payments__id=payment_id, payments__is_active=True)
+        .first()
+    )
+
+
 def _update_payment_with_new_transaction(
```
korycins, line 142: I believe a name should be changed. Previously we always upade a payment with new transaction. Currently only when condition inside a function is true

saleor/payment/gateways/stripe/webhooks.py
```diff
@@ -158,6 +130,15 @@ def _finalize_checkout(
     )
 
 
+def _get_checkout(payment_id: int) -> Optional[Checkout]:
+    return (
+        Checkout.objects.prefetch_related("payments")
+        .select_for_update(of=("self",))
+        .filter(payments__id=payment_id, payments__is_active=True)
```
korycins, line 137: For processing a notification we should also take into account the payments which are not active

saleor/payment/gateways/stripe/webhooks.py
```diff
@@ -172,117 +153,134 @@ def _update_payment_with_new_transaction(
...
+    manager: PluginsManager,
 ):
     checkout = _get_checkout(payment.id)
 
     if checkout and payment.can_create_order():
-        _finalize_checkout(checkout, payment, payment_intent, kind, amount, currency)
+        _finalize_checkout(checkout, payment, manager)
 
 
 def handle_authorized_payment_intent(
```
korycins, line 184: Please double check that you've changed all definitions used for handling a notifications. Just want to be sure that we will not get an exception because we are trying to call function with incorrect definition

saleor/payment/gateways/stripe/webhooks.py
```diff
@@ -172,6 +152,15 @@ def _update_payment_with_new_transaction(
         raw_response=stripe_object.last_response,
         psp_reference=stripe_object.id,
     )
+
+    transaction_id = stripe_object.id
+    if payment.transactions.filter(
+        token=transaction_id,
+        kind=kind,
+        is_success=True,
```
korycins, line 160: What about `action_required=False`?

saleor/payment/gateways/stripe/webhooks.py
stnatic: This should be fine - there are both no gaps in coverage and no complaints from mypy

saleor/payment/gateways/stripe/webhooks.py
stnatic: https://github.com/saleor/saleor/pull/8544/commits/b27e7c2d537b51ee4f379f6926766fa545a6d559

korycins (approved):

maarcingebala (approved):

IKarbowiak (approved):

PR #8561, opened 2021-11-10
Refund/void inactive payments in Adyen/Stripe webhooks handling

L3str4nge:
Added refund/void functionality in Adyen/Stripe webhooks handlers for inactive payments. 

This is because we have payment methods that do not require user's confirmation (so even if user doesn't finish redirect then payment will be auth/captured). While such a payment is during processing (with kinda delay eg. Ideal payment method) we can have situation where user has second tab opened and complete checkout with different payment. By doing that we are cancelling all active payments before creating new one. In fact we have two payments for single checkout but we can't handle first one which is actually inactive but has to be refunded or voided.

Decorator `@require_active_payment` is no longer valid for `refund` and `void` methods thats because unnecessary tests were modified or deleted.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-4887 Refund non active payments](https://app.clickup.com/t/2549495/SALEOR-4887)

saleor/payment/utils.py
```diff
@@ -433,3 +433,29 @@ def price_to_minor_unit(value: Decimal, currency: str):
     number_places = Decimal("10.0") ** precision
     value_without_comma = value * number_places
     return str(value_without_comma.quantize(Decimal("1")))
+
+
+def get_channel_slug_from_payment(payment: Payment) -> Union[str, None]:
```
IKarbowiak, line 438:
```suggestion
def get_channel_slug_from_payment(payment: Payment) -> Optional[str]:
```

saleor/payment/utils.py
```diff
@@ -433,3 +433,29 @@ def price_to_minor_unit(value: Decimal, currency: str):
...
+    if payment.checkout:
+        channel_slug = payment.checkout.channel.slug
+    elif payment.order:
+        channel_slug = payment.order.channel.slug
+    else:
+        channel_slug = None  # type: ignore
```
IKarbowiak, line 444: Why `# type: ignore` is needed here?

saleor/payment/gateways/adyen/tests/webhooks/test_handle_notifications.py
```diff
@@ -239,6 +239,48 @@ def test_handle_authorization_with_adyen_auto_capture(
...
+        value=price_to_minor_unit(payment.total, payment.currency),
+    )
+
+    plugin = adyen_plugin(adyen_auto_capture=True)
+    handle_authorization(notification, plugin.config)
+
+    payment.refresh_from_db()
+
+    assert payment.charge_status == ChargeStatus.FULLY_CHARGED
+    assert refund_mock.called
```
IKarbowiak, line 260: Could you use blank lines or `given, when and then` annotation in code? Right now is hard to find what are you testing.

saleor/payment/gateways/stripe/webhooks.py
```diff
@@ -217,6 +218,19 @@ def handle_authorized_payment_intent(
...
+
+    if not payment.is_active:
+        transaction = _update_payment_with_new_transaction(
+            payment,
+            payment_intent,
+            TransactionKind.AUTH,
+            payment_intent.amount,
+            payment_intent.currency,
+        )
+        manager = get_plugins_manager()
```
maarcingebala, line 230: In PR #8544 this function is refactored to receive plugins manager as an argument; it also changes the logic here: https://github.com/saleor/saleor/pull/8544/files#diff-2213dac925da7e225ab80a2e22732c5bfd480e7cf86072990f990409d3b1ef70R190

saleor/payment/utils.py
```diff
@@ -433,3 +433,29 @@ def price_to_minor_unit(value: Decimal, currency: str):
...
+):
+    """Handle refund or void inactive payments.
+
+    In case when we have open multiple payments for single checkout but only one is
+    active. Some payment methods don't required confirmation so we can receive delayed
+    webhook when we have order already paid.
+    """
+    if transaction.is_success:
+        update_payment_charge_status(payment, transaction)
+        channel_slug = get_channel_slug_from_payment(payment)
```
maarcingebala, line 460: I see that channel_slug is available in `handle_authorized_payment_intent` as an argument, so maybe we don't have to fetch it again? Maybe we could receive the channel easily in other places as well? I would check if we can get rid of extra DB queries here.

saleor/payment/gateways/stripe/webhooks.py
L3str4nge: Agreed but It's refactored on the feature branch (partial payments) but we need it now on `3.0`.

saleor/payment/utils.py
L3str4nge: It's a common utility for Adyen and Stripe and with Stripe it will work like a charm but with Adyen not really because we don't receive channel in handlers. Only way to get channel is get it from payment.order or payment.checkout. It's a good point overall because less queries is better but in this case I will need separate functions...

saleor/payment/utils.py
L3str4nge: `saleor/payment/utils.py:444: error: Incompatible types in assignment (expression has type "None", variable has type "str")` because of mypy :( anyway, refactored  and type ignore not needed anymore

saleor/payment/gateways/adyen/tests/webhooks/test_handle_notifications.py
IKarbowiak: Could you update also other tests? I see similar situations in other tests in `saleor/payment/gateways/adyen/tests/webhooks/test_handle_notifications.py` and `saleor/payment/gateways/stripe/tests/test_webhooks.py`.

IKarbowiak (approved):

saleor/payment/utils.py
```diff
@@ -433,3 +433,29 @@ def price_to_minor_unit(value: Decimal, currency: str):
...
+        gateway.payment_refund_or_void(payment, manager, channel_slug=channel_slug)
```
stnatic, line 461: This can raise a PaymentError which will result in the webhook failing. On the partial payments branch we have a task called `refund_or_void_inactive_payment` that retries on PaymentError

saleor/payment/gateways/stripe/webhooks.py
stnatic: I merged this PR into partial payments here: https://github.com/SaleorIntegrations/saleor/commit/d6cd634721e456e64fe0619abe8cec2e967478a7. They are on their way to 3.0 next week so I think it would be good to test those 2 features together

saleor/payment/utils.py
stnatic: I tested this by running this code by our task. I created an iDEAL payment but didn't press the button to redirect back to Saleor and then changed my mind and paid with a VISA card. At this moment Saleor was only aware of the VISA transaction:
image

Later on, after the payment got captured, it got refunded both in Saleor and Adyen (and appeared on the list of payments):
image

image

korycins (approved):

jakubkuc (approved):

eCzerniak (approved):

PR #8583, opened 2021-11-16
Clean up shipping method handling

kczan:
This PR allows for further steps in porting changes introduced in `exclude-shipping-methods-3-0` to 3.1 version.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-4842 [Stage 0] Clean up shipping method handling in upstream](https://app.clickup.com/t/2549495/SALEOR-4842)

saleor/shipping/utils.py
```diff
@@ -44,4 +46,5 @@ def convert_to_shipping_method_data(
         minimum_delivery_days=shipping_method.minimum_delivery_days,
         metadata=shipping_method.metadata,
         private_metadata=shipping_method.private_metadata,
+        minimum_order_price=getattr(listing, "minimum_order_price", None),
```
stnatic, line 49: Why getattr? Isn't it a regular model field?

saleor/shipping/interface.py
```diff
@@ -30,6 +30,8 @@ class ShippingMethodData:
     minimum_delivery_days: Optional[int] = None
     metadata: Dict[str, str] = field(default_factory=dict)
     private_metadata: Dict[str, str] = field(default_factory=dict)
+    active: bool = True
+    message: Optional[str] = ""
```
stnatic, line 34: Does it make sense to allow None here? Won't an empty string be sufficient?

saleor/graphql/shipping/types.py
```diff
@@ -313,6 +279,10 @@ def resolve_shipping_methods(
         root: ChannelContext[models.ShippingZone], info, **_kwargs
     ):
         def wrap_shipping_method_with_channel_context(shipping_methods):
+            # TODO: ShippingMethodType will have no price attribute
```
stnatic, line 282: This shouldn't be a TODO anymore

saleor/graphql/order/resolvers.py
```diff
@@ -58,3 +59,14 @@ def resolve_order_by_token(token):
         .filter(token=token)
         .first()
     )
+
+
+# TODO: We should optimize it in/after PR#5819
+def resolve_order_shipping_methods(root: models.Order, info):
+    available = get_valid_shipping_methods_for_order(root)
+    if available is None:
```
stnatic, line 67: Does this still return None?

saleor/graphql/checkout/mutations.py
```diff
@@ -113,7 +113,8 @@ def clean_delivery_method(
         )
 
     valid_methods = checkout_info.valid_delivery_methods
-    return method in valid_methods
+    # TODO: use me to check for price mismatches
```
stnatic, line 116: Restore the original version and check if this change is still needed.

saleor/graphql/checkout/dataloaders.py
```diff
@@ -245,14 +254,11 @@ def with_checkout_info(results):
                             delivery_method = collection_points_map.get(
                                 checkout.collection_point_id
                             )
-                        shipping_address = (
-                            address_map.get(checkout.shipping_address_id),
-                        )
+                        shipping_address = address_map.get(checkout.shipping_address_id)
```
stnatic, line 257: This was previously a list and now is an instance - was it a bug until now?

saleor/core/taxes.py
```diff
@@ -22,6 +22,10 @@ def zero_taxed_money(currency: str) -> TaxedMoney:
     return TaxedMoney(net=zero, gross=zero)
 
 
+def identical_taxed_money(money: Money) -> TaxedMoney:
```
stnatic, line 25: Maciek suggested renaming it, let's look for other options / ask others for feedback.

stnatic (changes requested):
One general question;
Is `manager.apply_taxes_to_shipping` still called anywhere in the code? If not, should we deprecate it?

saleor/graphql/checkout/mutations.py
kczan: It isn't actually, thanks

saleor/graphql/checkout/dataloaders.py
kczan: all the methods that use `shipping_address` below seem to be expecting an `Address` instance, not a one element tuple. I think before it was just a typo or maybe a linter messed up somehow

saleor/core/taxes.py
kczan: I went with Maciek's suggestion, maybe it actually is clearer

saleor/graphql/order/resolvers.py
kczan: it doesn't, empty lists will be returned from `get_valid_shipping_methods_for_order` 👍

saleor/shipping/interface.py
kczan: I believe you're right, allowing None when assigning an empty string either way seems redundant

saleor/shipping/utils.py
kczan: It's because `listing` is optional here but I can move the assignment to the condition check above if you prefer it that way

PR #8595, opened 2021-11-18
Add metadata to checkout and order lines

iprlic:

adding metadata to checkout and order lines

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


maarcingebala:
We discussed internally that in order to support metadata for order lines and checkout lines, we should first migrate database IDs to UUID for these models. This is due to the fact that public metadata mutations are not restricted with any permissions for these models. Switching to UUIDs would make the GraphQL IDs harder to guess and thus safer in the public API. We'll add a task to our internal roadmap for that.

maarcingebala:
For now I'm closing the PR and we will get back to it when database IDs are sorted out. Thanks for your contribution anyway @iprlic!

PR #8640, opened 2021-11-22
Support for Net protections in 3.0

korycins:
I want to merge this change because it adds support for new payment gateway: Net protections

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+    return inner
+
+
+def parse_errors(errors: List[str]) -> str:
+    # FIXME: better solution?
+    #  Transaction.error in database has max_length of 256
+    return os.linesep.join(errors)[:256]
+
+
+@inject_api_config
```
L3str4nge, line 34: Do we really need that decorator? Why not just call `get_api_config` in function body?

saleor/payment/gateways/np_atobarai/api_helpers.py
```diff
@@ -0,0 +1,258 @@
...
+    with np_atobarai_opentracing_trace("np-atobarai.utilities.request"):
+        response = requests.request(
+            method=method,
+            url=get_url(config, path),
+            timeout=REQUEST_TIMEOUT,
+            json=json or {},
+            auth=HTTPBasicAuth(config.merchant_code, config.sp_code),
+            headers={"X-NP-Terminal-Id": config.terminal_id},
+        )
+        if 400 < response.status_code <= 600:
```
L3str4nge, line 55: How about `raise_for_status()` ?

saleor/payment/gateways/np_atobarai/api_helpers.py
```diff
@@ -0,0 +1,258 @@
...
+    message = f"Payment #{transaction_id} {action.capitalize()} Unrecoverable Error"
+    logger.error("%s: %s", message, ", ".join(error_codes))
+    if order:
+        notify_dashboard(order, message)
+
+
+def health_check(config: "ApiConfig") -> bool:
+    try:
+        _request(config, "post", "/authorizations/find")
+        return True
```
L3str4nge, line 89: Does timeout raise `requestException` ?

saleor/payment/gateways/np_atobarai/api_helpers.py
```diff
@@ -0,0 +1,258 @@
...
+def format_address(config: "ApiConfig", ad: AddressData) -> Optional[str]:
+    """Follow the Japanese address guidelines."""
+    # example: "東京都千代田区麹町４－２－６　住友不動産麹町ファーストビル５階"
+    if not config.fill_missing_address:
+        return f"{ad.country_area}{ad.street_address_1}{ad.street_address_2}"
+    with Posuto() as pp:
+        try:
+            jap_ad = pp.get(ad.postal_code)
+        except KeyError:
+            return None
```
L3str4nge, line 108: Would be good to add log here

saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+def void(payment_information: PaymentData, config: ApiConfig) -> GatewayResponse:
+    raise NotImplementedError
+
+
+@inject_api_config
+def tracking_number_updated(fulfillment: Fulfillment, config: ApiConfig) -> None:
+    from .plugin import NPAtobaraiGatewayPlugin
```
korycins, line 71: You're importing this to only get a plugin_id. Can't we move plugin ID do `const.py` file and import it normaly in this file and in the `plugin.py`? 
In `NPAtobaraiGatewayPlugin` you could just use simple assigment `PLUGIN_ID=const.PLUGIN_ID`

saleor/payment/gateways/np_atobarai/api_helpers.py
```diff
@@ -0,0 +1,258 @@
...
+                f"{ad.country_area}"
+                f"{jap_ad.city}"
+                f"{jap_ad.neighborhood}"
+                f"{ad.street_address_1}"
+                f"{ad.street_address_2}"
+            )
+
+
+def format_price(price: Decimal, currency: str) -> int:
+    return int(price_to_minor_unit(price, currency))
```
L3str4nge, line 120: Is casting decimal to int is safe for really big amounts?

saleor/payment/gateways/np_atobarai/errors.py
```diff
@@ -0,0 +1,32 @@
...
+TRANSACTION_CHANGE = "CH"
+FULFILLMENT_REPORT = "FR"
+
+
+def add_action_to_code(action: str, error_code: str) -> str:
+    return f"{action}#{error_code}"
+
+
+def get_error_messages_from_codes(action: str, error_codes: Iterable[str]) -> List[str]:
+    return [add_action_to_code(action, code) for code in error_codes]
```
L3str4nge, line 14: Couldn't `add_action_to_code` be inline here?

saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+
+        if already_captured:
+            logger.warning("%s was already captured", payment_name.capitalize())
+            notify_dashboard(
+                order, f"Error: {payment_name.capitalize()} was already captured"
+            )
+        elif errors:
+            error = ", ".join(errors)
+            logger.warning(
+                "Could not capture %s in NP Atobarai: %s", payment_name, error
```
L3str4nge, line 96: Why not `f-string` here?

saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,237 @@
...
+                    )
+                error_messages = result["authori_hold"]
+                return errors_payment_result(error_messages)
+
+            return PaymentResult(
+                status=PaymentStatus.SUCCESS,
+            )
+
+        if PRE_FULFILLMENT_ERROR_CODE in error_codes:
+            return None
```
L3str4nge, line 147: Log here?

saleor/payment/gateways/np_atobarai/api_helpers.py
```diff
@@ -0,0 +1,258 @@
...
+if TYPE_CHECKING:
+    from . import ApiConfig
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_url(config: "ApiConfig", path: str = "") -> str:
+    """Resolve test/production URLs based on the api config."""
+    return f"{NP_TEST_URL if config.test_mode else NP_URL}{path}"
```
L3str4nge, line 37: Wouldn't be better to move URLs to settings and get it directly depends on current environment? Checking environment each call isn't good idea tbh

L3str4nge, line 37: Checking config mode each call isn't good idea tbh. Could we set URL explicit?

saleor/payment/gateways/np_atobarai/plugin.py
```diff
@@ -0,0 +1,204 @@
...
+                            f"Shipping company code is invalid",
+                            code=PluginErrorCode.INVALID.value,
+                        )
+                    }
+                )
+
+            cls.validate_authentication(plugin_configuration)
+
+    def get_payment_config(self, previous_value):
+        return []
```
L3str4nge, line 204: Is it proper to return empty list here?

saleor/payment/gateways/np_atobarai/plugin.py
```diff
@@ -0,0 +1,204 @@
...
+    def validate_plugin_configuration(cls, plugin_configuration: "PluginConfiguration"):
+        """Validate if provided configuration is correct."""
+        if not plugin_configuration.active:
+            return
+
+        configuration = plugin_configuration.configuration
+        configuration = {item["name"]: item["value"] for item in configuration}
+        missing_fields = [
+            field
+            for field in [MERCHANT_CODE, SP_CODE, TERMINAL_ID, SHIPPING_COMPANY]
```
L3str4nge, line 176: Could be `tuple`

saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+
+@inject_api_config
+def process_payment(
+    payment_information: PaymentData, config: ApiConfig
+) -> GatewayResponse:
+    payment_id = payment_information.payment_id
+    payment = Payment.objects.filter(pk=payment_id).first()
+
+    if not payment:
+        raise PaymentError(f"Payment with id {payment_id} does not exist.")
```
maarcingebala, line 42: I think we should add an error log here that would include the failed payment ID and not include it in the message. If we want to include the ID, it should be the GraphQL node ID.

saleor/payment/utils.py
```diff
@@ -1,38 +1,178 @@
...
 
 
+def create_payment_lines_information(
+    payment: Payment,
+    manager: PluginsManager,
+) -> List[PaymentLineData]:
+    checkout = payment.checkout
+    order = payment.order
+    line_items = []
+    if checkout:
```
maarcingebala, line 46: Minor thing but this function is pretty big, it would be a bit more readable if we added some separate helpers for the checkout case and the order case below.

saleor/payment/gateways/np_atobarai/api_helpers.py
mateuszgrzyb: `raise_for_status()` raises on status 400, which is a valid response from NP Atobarai

saleor/payment/gateways/np_atobarai/api_helpers.py
mateuszgrzyb: It raises `Timeout` exception, which inherits from `RequestException`

saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+def inject_api_config(fun):
+    def inner(payment_information: PaymentData, config: GatewayConfig):
+        return fun(payment_information, get_api_config(config.connection_params))
+
+    return inner
+
+
+def parse_errors(errors: List[str]) -> str:
+    # FIXME: better solution?
+    #  Transaction.error in database has max_length of 256
```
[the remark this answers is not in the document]
mateuszgrzyb: How about `os.linesep.join(errors[:20])`? Max length of error code is 11, so we will never get the result here longer than 256. It's better solution than the current one, because it will not split the last error code in half

saleor/payment/gateways/np_atobarai/api.py
mateuszgrzyb: Why should we log this result? It's not an exceptional flow. We're just asking here NP for transaction status. If it's after fulfilment report, we reregister transaction.

L3str4nge: It's silent `None` when we have actually `error_codes`

L3str4nge: Logging isn't for exceptional flow :o

L3str4nge (approved):

saleor/payment/gateways/np_atobarai/errors.py
mateuszgrzyb: It's also used in `api.py` module

maarcingebala (approved):
Approving the current state of the PR because we need to release it for further testing. @stnatic @mateuszgrzyb Please make sure that all review comments are written down as a task and will be fixed in a separate PR.

korycins:
@stnatic @mateuszgrzyb Remember about extending Saleor's docs and adding a docstring to all functions/methods which executes some bussiness logic.

PR #8644, opened 2021-11-23
Webhook payloads

jakubkuc:
I want to merge this change because it changes way saleor handles webhooks. It will allow us to monitor and retry failed event deliveries. 

Clickup list: https://app.clickup.com/2549495/v/l/6-144350552-1

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/webhook/types.py
```diff
@@ -45,3 +109,7 @@ class Meta:
     @staticmethod
     def resolve_events(root: models.Webhook, *_args, **_kwargs):
         return root.events.all()
+
+    @staticmethod
+    def resolve_deliveries(root: models.Webhook, *_args, **_kwargs):
+        return core_models.EventDelivery.objects.filter(webhook_id=root.pk)
```
korycins, line 115: Shouldn't we have here a dataloader?

saleor/plugins/webhook/utils.py
```diff
@@ -138,3 +149,71 @@ def parse_list_shipping_methods_response(
             )
         )
     return shipping_methods
+
+
+@contextmanager
+def catch_duration_time():
+    start = datetime.now()
+    yield lambda: datetime.now() - start
```
przlada, line 157:
```suggestion
from time import time
@contextmanager
def catch_duration_time():
    start = time()
    yield lambda: time() - start
```
Maybe we should stick to `time` module instead of `datetime` to prevent problems regarding counting time delta during a change to daylight saving time?

saleor/core/migrations/0003_auto_20211122_0927.py
```diff
@@ -0,0 +1,21 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0002_initial"),
+    ]
+
+    operations = [
```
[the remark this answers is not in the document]
jakubkuc: yes it will be the best to do that. As soon as all PRs with migrations will be marged to this feature branch I will do that.

saleor/graphql/webhook/types.py
jakubkuc: since it is FilterInputConnectionField with filtering and sorting we can't use data loaders at the moment.

saleor/graphql/schema.graphql
```diff
@@ -7256,6 +7332,7 @@ type Webhook implements Node {
   id: ID!
   events: [WebhookEvent!]!
   app: App!
+  deliveries(sortBy: EventDeliverySortingInput, filter: EventDeliveryFilterInput, before: String, after: String, first: Int, last: Int): EventDeliveryCountableConnection
```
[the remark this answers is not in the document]
jakubkuc: done

tomaszszymanski129:
please rebase @jakubkuc

saleor/graphql/webhook/tests/mutations/test_delivery_retry.py
```diff
@@ -0,0 +1,102 @@
+from unittest.mock import patch
+
+import graphene
+
+from saleor.graphql.tests.utils import assert_no_permission, get_graphql_content
```
korycins, line 5: Just styling thing, but I see that you use relative import in different files, and here an absolute. Maybe we could align it to one approach?

saleor/graphql/webhook/tests/mutations/test_delivery_retry.py
```diff
@@ -0,0 +1,102 @@
...
+@patch("saleor.plugins.manager.PluginsManager.event_delivery_retry")
+def test_delivery_retry_mutation(
+    mocked_send_request_async,
+    app_api_client,
+    permission_manage_apps,
+    event_delivery,
+    settings,
+):
+    # given
+    settings.PLUGINS = ["saleor.plugins.webhook.plugin.WebhookPlugin"]
```
korycins, line 31: Do we need to enable webhook plugin here? Looks like we mock a PluginsManager's method here (which is great) so, no need to include here a webhook Plugin.

saleor/plugins/webhook/tasks.py
```diff
@@ -160,15 +190,205 @@ def send_webhook_using_google_cloud_pubsub(
...
+    else:
+        delivery_update(delivery, EventDeliveryStatus.FAILED)
+        raise ValueError("Unknown webhook scheme: %r" % (parts.scheme,))
+    delivery_update(delivery, response.status)
+    clear_successful_delivery(delivery)
+    return response_data
+
+
+# DEPRECATED
+# to be removed in task: #1q2x7xw
```
korycins, line 375: I don't think that adding task id is a good way for this. Maybe we should add a number of version or time period? Something more "public"

korycins (approved):

saleor/plugins/webhook/tasks.py
maarcingebala: This will be changed right after releasing to cloud, the sooner the better. We just need to wait for existing Celery tasks to execute and we are free to drop this code.

PR #8654, opened 2021-11-25
price values validation for ShippingMethodChannelListingUpdate mutation

jakubkuc:
I want to merge this change because ShippingMethodChannelListingUpdate mutation needed additional validation for decimal precision fields.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/core/validators.py
+++ b/saleor/graphql/core/validators.py
@@ -54,7 +54,7 @@ def validate_price_precision(value: Optional["Decimal"], currency: str):
         )
 
 
-def validate_price_max_value(value: Optional["Decimal"], max_value=10 ** 9):
+def validate_decimal_max_value(value: Optional["Decimal"], max_value=10 ** 9):
     """Validate if price amount is not higher than the limit for precision field.
 
     Decimal fields have value limits.
--- a/saleor/graphql/core/validators.py
+++ b/saleor/graphql/core/validators.py
@@ -54,15 +54,12 @@ def validate_price_precision(value: Optional["Decimal"], currency: str):
         )
 
 
-def validate_decimal_max_value(value: Optional["Decimal"], max_value=10 ** 9):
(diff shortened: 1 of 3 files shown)


patrys:
Task linked: [SALEOR-4488 Mutation ShippingMethodChannelListingUpdate does not validate input fields max values](https://app.clickup.com/t/2549495/SALEOR-4488)

saleor/graphql/core/validators.py
```diff
@@ -54,6 +54,19 @@ def validate_price_precision(value: Optional["Decimal"], currency: str):
         )
 
 
+def validate_price_max_value(value: Optional["Decimal"], max_value=10 ** 9):
```
korycins, line 57: I would rename function to decimal instead of price. It could be use for both cases as you're operating here on Decimal which could be related to something else than price

fowczarek (approved):

saleor/graphql/core/validators.py
L3str4nge: Why value is Optional?

jakubkuc: @korycins function renamed
@L3str4nge as we discussed - simplified function and dropped unnecessary checks

saleor/graphql/core/validators.py
```diff
@@ -54,6 +54,16 @@ def validate_price_precision(value: Optional["Decimal"], currency: str):
         )
 
 
+def validate_decimal_max_value(value: "Decimal", max_value=10 ** 9):
+    """Validate if price amount is not higher than the limit for precision field.
```
IKarbowiak, line 58:
```suggestion
    """Validate if amount is not higher than the limit for precision field.
```

IKarbowiak (approved):

korycins (approved):

tomaszszymanski129 (approved):

PR #8665, opened 2021-11-26
fix: plugins can now query the API without content-security-policy re…

LeOndaz:
This is something tricky, plugins can now query the /graphql/ API. This update introduces the following pattern:

```
#  saleor/plugins/my_special_plugin/plugin.py
from .graphql.schema import schema  # schema created for the plugin 

class MPlugin(BasePlugin):
    ...

    def webhook(self, request, path, previous_value):
        view = GraphQLView.as_view(schema=schema)
        return view(request)
```
And plugins can have their endpoints on /plugins/plugin_id/

Then we can federate the plugins with the base GraphQL API.


LeOndaz:
Kindly note that I wanted to access the plugin configuration in the schema of the GraphQLView above, and the only 2 way I could use are, either extend the GraphQLView and let it accept new parameters, or monkey-patch it on the request on the fly 

```
    def webhook(self, request, path, previous_value):
        view = GraphQLView.as_view(schema=schema)
        request.plugin_config = self.configuration   # or whatever needed
        return view(request)
```

And this lets me think of, why don't we return the `request.app` that was removed in an issue before and set it to the requestor plugin instance, or app URL?

LeOndaz:
@korycins @patrys @maarcingebala 

Not sure I know someone else here, can we discuss this?

IslamMesha (approved):

patrys:
@LeOndaz, could you start by describing the use case/problem you're trying to solve? I'm not sure I understand why that change is beneficial.

LeOndaz:
> @LeOndaz, could you start by describing the use case/problem you're trying to solve? I'm not sure I understand why that change is beneficial.

Hello @patrys , it's nice to hear from you again, let me elaborate

Plugins extend the core functionality, but webhook are restful, we can make the plugins expose a GraphQL endpoint as well, but we do need the plugin to use the sae playground instead of copying it around just to edit the content security policy, plugins at the moment can't use playground.html because content security policy doesn't allow urls other than `api_url` and this update adds support for plugin urls as well.

patrys:
@LeOndaz, I get it. That makes sense.

LeOndaz:
@patrys So basically, we can't render the template from any urls other than {{ api_uri }}, but plugins need to render it because some plugins may want to add a GraphQL playground on their own then we federate all of them at the end, this is the first step towards allowing plugins to be GraphQL services, another step was suggested above and here's a demo:

The `request.app` is not related to this specific PR, but here's a use case

```
/plugins/my.plugin/

initiateOAuth2(provider: Google) {
    url
    state
}

OAuth2Callback(provider: Google, code: "code", state: "statex") {
     user {
        id
     }
}
```

and the implementation is 

```
class initiateOAuth2Mutation(...):
    def perform_mutation(self, ...)
        # how can I know get the plugin's configuration here? I want client_id and client_secret of google to be able to use oauth2 
        # here lies the benefit of request.app again, 
```

LeOndaz:
@patrys can you also give me your opinion about the `request.app` thing?

saleor/graphql/views.py
```diff
@@ -120,7 +120,10 @@ def render_playground(self, request):
         return render(
             request,
             "graphql/playground.html",
-            {"api_url": request.build_absolute_uri(str(API_PATH))},
+            {
+                "api_url": request.build_absolute_uri(str(API_PATH)),
+                "plugins_url": f"{request.build_absolute_uri('/')}plugins/",
```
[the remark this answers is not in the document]
LeOndaz: Actually yes, I probably assumed it won't because I didn't see plugins in urlpatterns directly (without <plugin_id>), but I will commit atm

LeOndaz:
@patrys not sure why it is failing, I'll appreciate some help and how can this be merged

saleor/graphql/views.py
```diff
@@ -120,7 +120,10 @@ def render_playground(self, request):
         return render(
             request,
             "graphql/playground.html",
-            {"api_url": request.build_absolute_uri(str(API_PATH))},
+            {
+                "api_url": request.build_absolute_uri(str(API_PATH)),
+                "plugins_url": request.build_absolute_uri("/plugins/"),
```
maarcingebala, line 125: Looks good but shouldn't we have this URL defined in `urls.py` and use the `reverse` function here? I'm not sure as there is no view `/plugins/`; there is the `plugins/(?P<plugin_id>[.0-9A-Za-z_\-]+)/` named URL but it requires the ID.

LeOndaz: reverse won't work because it's not defined in url.py, but build_absolute_uri works because it just takes a path and returns it relative to the current host

P.S: This is tested

maarcingebala: 👍

maarcingebala (approved):

LeOndaz:
@maarcingebala 
Should I PR adding request.app again?

look closely at my example above, the schema is rendered correctly but it has no access to the plugin configuration

PR #8680, opened 2021-11-29
Fix order search

IKarbowiak:
- Introduce `search_document` order field, that collect values from all fields used in searching
- Add index for `search_document` user field
- Update `search_document` whenever any of the fields are changed
- Make search case insensitive
- Update random data - generate search value for created users

Explain (49221 orders):
https://explain.dalibo.com/plan/Yqe

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-4967 Improve order search](https://app.clickup.com/t/2549495/SALEOR-4967)

saleor/graphql/order/mutations/orders.py
```diff
@@ -842,6 +850,7 @@ def perform_mutation(cls, _root, info, **data):
         )
 
         recalculate_order(order)
+        update_order_search_document(order)
```
tomaszszymanski129, line 853: Did you consider moving this method call to `Order` model `save` method? This way we wouldn't have to track every single place order gets updated. Currently it is simple and explicit but might turn out to be hard to maintain.

saleor/order/search.py
```diff
@@ -0,0 +1,78 @@
...
+    return lines_data
+
+
+def search_orders(qs, value):
+    if value:
+        lookup = Q()
+        for val in value.split():
+            if val.startswith("#"):
+                val = val[1:]
+            lookup &= Q(search_document__ilike=val.lower())
```
tomaszszymanski129, line 74: `.lower()` is probably redundant as ILIKE operator on postgres is case insensitive anyway.

saleor/graphql/order/mutations/orders.py
patrys: Please don't put logic in the `save` method. When `save` is called there is no way to tell why it was called and it's impossible to give it enough context to make use of any data loaders.

saleor/order/search.py
patrys: Did you check if the proper index is used for this lookup?

saleor/graphql/order/mutations/orders.py
IKarbowiak: No, because we do not want to always update `search_document`, but only when field used for searching are changed.

saleor/order/search.py
IKarbowiak: Good point. I will change that

IKarbowiak: I attached explain in PR description. It uses `Bitmap Index Scan`.

korycins (commented):
@IKarbowiak
- the `prepare_order_search...` uses an order object and its relations, to generate a search document. Shouldn't we confirm, that for all places where we call this method we have all required prefetches? Or maybe a solution which will handle prefetches on its own?

IKarbowiak:
@korycins Yes we getting fields values from order relations, but to prefetch this relations we'd have to get order from db ones again with prefetch. I'm not sure if it will help as we are using only `values_list` on relations to get fields values.

patrys:
@IKarbowiak You can use [`prefetch_related_objects`](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#django.db.models.prefetch_related_objects) to prefetch data on already fetched instances.

IKarbowiak:
Thanks @patrys. I'll check if this reduces number of queries.

maarcingebala (approved):

fowczarek (approved):

korycins (approved):

PR #8684, opened 2021-11-29
Port NP Atobarai gateway to 3.1

stnatic:
I want to merge this change because I want to port changes from #8640 to 3.1.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/payment/gateways/np_atobarai/api_helpers.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+def get_refunded_goods(
+    config: "ApiConfig",
+    refund_data: Dict[int, int],
+    payment_information: PaymentData,
+) -> List[dict]:
+    return [
+        {
+            "goods_name": line.product_sku if config.sku_as_name else line.product_name,
```
maarcingebala, line 130: Will this work when `product_sku` is not known? SKUs are optional in 3.1.

saleor/order/tests/test_order_actions_refund_products.py
```diff
@@ -218,7 +230,11 @@ def test_create_refund_fulfillment_custom_amount(
     for line in fulfillment_lines:
         assert line.quantity == original_quantity.get(line.pk) - 2
     mocked_refund.assert_called_once_with(
-        payment_dummy, ANY, amount=amount, channel_slug=fulfilled_order.channel.slug
+        payment_dummy,
+        ANY,
+        amount=amount,
+        channel_slug=fulfilled_order.channel.slug,
+        refund_data=ANY,
```
IKarbowiak, line 237: Why any of these tests do not check what is provided in the `refund_data` parameter?

saleor/payment/gateways/np_atobarai/api_helpers.py
mateuszgrzyb: `sku_as_name` is a special (it's off by default) option for users that specifically use SKUs and want them in NP transactions.

saleor/order/tests/test_order_actions_refund_products.py
mateuszgrzyb: Fixed in current commit.

korycins (commented):
@mateuszgrzyb  IMHO we should add more docstrings. Especially for functions in `__init__` and `api` files. Some of them are doing business logic which can be hard to understand without a proper context.  Docstring of the function should give us such context.

saleor/payment/gateways/np_atobarai/api_helpers.py
stnatic: I am assuming that it's going to fail when we pass `null` - something has to be printed on an invoice. On the other hand, the requirement from the customer was that SKU is displayed as the invoice line description, therefore it was added as a plugin configuration option.

We need to make a judgement call here, should we leave it up to customer to take responsibility for changing the plugin configuration to use SKUs or fall back to the product name if SKU is unset? I'd probably vote for the latter.

korycins: In similar cases, we assumed that we will use .id when .sku is missing. Currently, I think that we need to be sure that this will not raise an exception that we will not handle.

saleor/graphql/order/mutations/fulfillments.py
```diff
@@ -683,6 +683,7 @@ def clean_lines(cls, lines_data, cleaned_input):
                     line.pk,
                     "order_line_id",
                 )
+            variant = line.variant
```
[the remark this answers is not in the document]
mateuszgrzyb: Yes, we pass prefetched queryset in the `get_node_or_error` method at the beginning of the function.

IKarbowiak (approved):

saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    order = fulfillment.order
+    payments = order.payments.filter(gateway=NP_PLUGIN_ID, is_active=True)
+
+    if payments:
+        results = [
+            api.report_fulfillment(config, payment, fulfillment) for payment in payments
+        ]
+    else:
+        results = [("", ["No active payments for this order"], False)]
```
L3str4nge, line 86: Consider just 
```
notify_dashboard(order, "No active payments for this order")
return
```
Instead of force creating data for `loop` below. Notice that you produce `""`  payment_id so in line 98/99 the error msg will be lack of payment name information.

saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,263 @@
...
+
+def register_transaction(
+    order: Optional[Order], config: ApiConfig, payment_information: "PaymentData"
+) -> PaymentResult:
+    """Create a new transaction in NP Atobarai.
+
+    On pending status from NP the transaction is cancelled and
+    reason for pending is returned as error message.
+    """
+    with np_atobarai_opentracing_trace("np-atobarai.checkout.payments.register"):
```
korycins, line 48: Maybe I am wrong but I would say that the tracing should be only on the request, not request+processing and resopnse. 
@fowczarek  Can you confirm?

saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+                payment_information,
+                shipping_company_code,
+                tracking_number,
+                refund_data,
+            )
+
+    else:
+        result = api.cancel_transaction(config, payment_information)
+
+    if psp_reference := result.psp_reference:
```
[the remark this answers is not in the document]
mateuszgrzyb: Yes, without updated psp reference subsequent updates would not be possible as we would refer to the deleted transaction in NP.

saleor/payment/gateways/np_atobarai/__init__.py
mateuszgrzyb: Resolved in recent commi

saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+
+    if payments:
+        results = [
+            api.report_fulfillment(config, payment, fulfillment) for payment in payments
+        ]
+    else:
+        results = [("", ["No active payments for this order"], False)]
+
+    for payment_id, errors, already_captured in results:
+        payment_name = get_payment_name(payment_id)
```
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit

saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+    A new psp reference may be issued,
+    which will be saved in payment object.
+    """
+    payment_id = payment_information.payment_id
+    payment = Payment.objects.filter(pk=payment_id).first()
+
+    if not payment:
+        raise PaymentError(f"Payment with id {payment_id} does not exist.")
+
+    if payment_information.amount < payment.captured_amount:
```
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit

saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+                config,
+                payment,
+                payment_information,
+                shipping_company_code,
+                tracking_number,
+                refund_data,
+            )
+
+    else:
+        result = api.cancel_transaction(config, payment_information)
```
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit

saleor/payment/utils.py
```diff
@@ -25,27 +32,161 @@
...
+) -> Optional[PaymentLineData]:
+    discount_amount = -checkout.discount_amount
+    return create_voucher_payment_line_data(discount_amount)
+
+
+def create_order_voucher_payment_line_data(
+    order: Order,
+) -> Optional[PaymentLineData]:
+    discount_amount = order.total_gross_amount - order.undiscounted_total_gross_amount
+    return create_voucher_payment_line_data(discount_amount)
```
korycins, line 166: Why do we need this? I am not sure that we should place them as a payment line. Personally I would say that this is something related to NP.

saleor/payment/utils.py
```diff
@@ -25,27 +32,161 @@
...
+    return line_items
+
+
+# Values are outside of model's pk range to resolve
+# any collision with actual product variant pk
+VOUCHER_PAYMENT_LINE_ID = 0
+SHIPPING_PAYMENT_LINE_ID = -1
+
+
+def create_shipping_payment_line_data(amount: Decimal) -> PaymentLineData:
```
korycins, line 145: I am not convinced, that placing shipping and voucher as a payment line is the best way to pass it to NP plugin. Especially that we have to use `ID` field with values like `0` and `-1`.
@maarcingebala  What do you think?

saleor/payment/utils.py
```diff
@@ -107,8 +248,89 @@ def create_payment_information(
...
+    fulfillment_lines_to_refund: List[FulfillmentLineData],
+    refund_shipping_costs: bool,
+) -> Dict[int, int]:
+    order_lines = {line.variant_id: line.quantity for line in order.lines.all()}
+
+    refund_lines = _prepare_refund_lines(
+        order, order_lines_to_refund, fulfillment_lines_to_refund
+    )
+
+    summed_refund_lines: Dict[int, int] = defaultdict(int)
```
korycins, line 314: Looks like, you're calling `_prepare_refund_lines` function only to get a data to iterate over them. 
What about moving this for to `_prepare_refund_lines`? Instead of creating `previous_refund_lines`, `current_order_refund_lines` and `current_fulfillment_refund_lines`, you could immedietly create `refund_lines: Dict[int, int] = defaultdict(int)` and fulfill it with all data passed to `_prepare_refund_lines`.

saleor/payment/gateways/np_atobarai/api_helpers.py
```diff
@@ -0,0 +1,267 @@
...
+    with np_atobarai_opentracing_trace("np-atobarai.utilities.request"):
+        response = requests.request(
+            method=method,
+            url=get_url(config, path),
+            timeout=REQUEST_TIMEOUT,
+            json=json or {},
+            auth=HTTPBasicAuth(config.merchant_code, config.sp_code),
+            headers={"X-NP-Terminal-Id": config.terminal_id},
+        )
+        if 400 < response.status_code <= 600:
```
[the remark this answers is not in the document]
mateuszgrzyb: There is no custom handler for `status_code` == 400. NP returns error codes with this status code. We want to process it as a normal response to pass them to the UI.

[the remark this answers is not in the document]
korycins: So this is a info that we should place as a comment 👍

patrys:
Task linked: [SALEOR-5872 Port Net Protection from 3.0 to 3.1](https://app.clickup.com/t/2549495/SALEOR-5872)

saleor/payment/gateways/np_atobarai/api_helpers.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+def format_address(config: "ApiConfig", ad: AddressData) -> Optional[str]:
+    """Follow the Japanese address guidelines."""
+    # example: "東京都千代田区麹町４－２－６　住友不動産麹町ファーストビル５階"
+    if not config.fill_missing_address:
+        return f"{ad.country_area}{ad.street_address_1}{ad.street_address_2}"
+    with Posuto() as pp:
+        try:
+            jap_ad = pp.get(ad.postal_code)
```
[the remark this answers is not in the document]
stnatic: @fowczarek this library seems to be internally using a sqlite database to keep the postal codes, I'm flagging this one as it can mess up with your APM?

saleor/payment/gateways/np_atobarai/const.py
```diff
@@ -0,0 +1,45 @@
+# This integration is currently supporting the settlement type 02 - NP Atobarai
+NP_ATOBARAI = "02"
+NP_ATOBARAI_WIZ = "03"
+
+REQUEST_TIMEOUT = 30
```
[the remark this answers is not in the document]
stnatic: NP recommends it to be around 30s, so I'd stick with that:
image

I left an inline comment explaining the value

fowczarek (approved):

saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,237 @@
...
+        return PaymentResult(
+            status=PaymentStatus.SUCCESS,
+        )
+
+    if PRE_FULFILLMENT_ERROR_CODE in error_codes:
+        logger.info(
+            "Fulfillment for payment with id %s was reported",
+            payment_information.graphql_payment_id,
+        )
+        return None
```
korycins, line 146: I am not sure if I follow why we return here a `None`.  It looks like returning `None` here means that the expected case just happened. Why we don't use `PaymentResult` here? I believe we should have a separate status for handling such cases.

saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,237 @@
...
+        config,
+        payment_information,
+        billed_amount,
+        goods,
+    )
+
+    if not error_codes:
+        new_psp_reference = result["np_transaction_id"]
+
+        report(config, shipping_company_code, new_psp_reference, tracking_number)
```
korycins, line 204: Isn't  it a risk of false positive here? `report` can return error status but we always return `SUCCESS` here

saleor/payment/gateways/np_atobarai/api_types.py
```diff
@@ -0,0 +1,70 @@
...
+class PaymentStatus(str, Enum):
+    SUCCESS = "00"
+    PENDING = "10"
+    FAILED = "20"
+
+
+@dataclass
+class PaymentResult:
+    status: PaymentStatus
+    psp_reference: str = ""
```
korycins, line 47: I think that this is incorrect. If we don't always have `psp_reference` then I think this field should be marked as `Optional[str]` as it is more meaningful that setting empty `str` as a default one

saleor/payment/gateways/np_atobarai/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+    payment = Payment.objects.filter(pk=payment_id).first()
+
+    if not payment:
+        logger.error(
+            "Payment with id %s does not exist",
+            payment_information.graphql_payment_id,
+        )
+        raise PaymentError("Payment does not exist.")
+
+    result = api.register_transaction(payment.order, config, payment_information)
```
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit.

saleor/payment/gateways/np_atobarai/api_types.py
mateuszgrzyb: Resolved in recent commit.

saleor/payment/utils.py
mateuszgrzyb: Resolved in recent commit.

saleor/payment/gateways/np_atobarai/api_helpers.py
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit.

saleor/payment/gateways/np_atobarai/api.py
mateuszgrzyb: Resolved in recent commit.

saleor/payment/gateways/np_atobarai/api.py
```diff
@@ -0,0 +1,250 @@
...
+            )
+
+        if cancel_error_codes := cancel(config, psp_reference).error_codes:
+            error_messages = get_error_messages_from_codes(
+                action=TRANSACTION_CANCELLATION, error_codes=cancel_error_codes
+            )
+            return errors_payment_result(error_messages)
+
+        if refund_data:
+            goods = get_refunded_goods(config, refund_data, payment_information)
```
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit.

saleor/payment/gateways/np_atobarai/api_helpers.py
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit.

saleor/payment/gateways/np_atobarai/api_helpers.py
```diff
@@ -0,0 +1,267 @@
...
+            raise requests.HTTPError
+        return response
+
+
+def np_request(
+    config: "ApiConfig", method: str, path: str = "", json: Optional[dict] = None
+) -> NPResponse:
+    try:
+        response = _request(config, method, path, json)
+        response_data = response.json()
```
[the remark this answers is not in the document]
mateuszgrzyb: Yes. Status 400 is used exclusively for NP error codes.

saleor/payment/utils.py
mateuszgrzyb: Resolved in recent commit.

korycins (approved):

saleor/payment/utils.py
```diff
@@ -277,9 +277,15 @@ def _create_refund_manual_amount(
 
     Takes into account previous refunds and current refund mutation parameters.
     """
-    previous_manual_amount_to_refund = order.fulfillments.filter(
-        lines__order_line__variant_id__isnull=True, shipping_refund_amount__isnull=True
-    ).aggregate(discount=Sum("total_refund_amount"))["discount"]
+    previous_manual_amount_to_refund = (
+        Fulfillment.objects.prefetch_related("lines__order_line")
```
korycins, line 281: Are you sure that `.prefetch_related("lines__order_line")` is required here?

saleor/payment/utils.py
```diff
@@ -277,9 +277,15 @@ def _create_refund_manual_amount(
...
+        .filter(
+            order_id=order.pk,
+            lines__order_line__variant_id__isnull=True,
+            shipping_refund_amount__isnull=True,
```
korycins, line 285: I don't think that we should rely on the shipping_refund_amount as a value that tells us that it was manual refund.

saleor/payment/gateways/np_atobarai/api_helpers.py
```diff
@@ -165,29 +165,49 @@ def _get_voucher_and_shipping_goods(
...
 
 
-def get_refunded_goods(
+def get_goods_with_refunds(
     config: "ApiConfig",
-    refund_data: RefundData,
     payment_information: PaymentData,
 ) -> List[dict]:
-    """Convert payment lines and refund lines into NP Atobarai goods lines.
+    goods_lines = []
```
korycins, line 177: This method previously had a docstring. Why you didn't add it now?

mateuszgrzyb: Resolved in recent commit.

korycins (approved):

korycins (approved):

saleor/payment/gateways/np_atobarai/plugin.py
```diff
@@ -36,7 +36,7 @@ class NPAtobaraiGatewayPlugin(BasePlugin):
     PLUGIN_ID = NP_PLUGIN_ID
     PLUGIN_NAME = GATEWAY_NAME
     CONFIGURATION_PER_CHANNEL = True
-    SUPPORTED_CURRENCIES = "PLN"
```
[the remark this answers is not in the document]
stnatic: It's supposed to be a comma separated list, see:
image

PR #8688, opened 2021-11-30
Use frozen order data to return shipping method

bogdal:
This PR uses `order.shipping_method_name` and `order.shipping_price` fields to resolve the shipping method.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/types.py
+++ b/saleor/graphql/order/types.py
@@ -43,6 +43,7 @@ from ...payment.model_helpers import (
 from ...product import ProductMediaTypes
 from ...product.models import ALL_PRODUCTS_PERMISSIONS
 from ...product.product_images import get_product_image_thumbnail
+from ...shipping.interface import ShippingMethodData
 from ..account.dataloaders import AddressByIdLoader, UserByUserIdLoader
 from ..account.types import User
 from ..account.utils import requestor_has_access
@@ -73,7 +74,6 @@ from ..product.dataloaders import (
     ProductVariantByIdLoader,
 )
 from ..product.types import ProductVariant
-from ..shipping.dataloaders import ShippingMethodByIdLoader
 from ..shipping.types import ShippingMethod
 from ..warehouse.types import Allocation, Warehouse
 from .dataloaders import (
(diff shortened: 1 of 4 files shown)


saleor/graphql/order/types.py
```diff
@@ -1042,39 +1042,25 @@ def _resolve_user(user):
...
-                return ChannelContext(
-                    node=shipping_method, channel_slug=root.channel.slug
-                )
+    def resolve_shipping_method(root: models.Order, _info):
+        order_shipping_method_id = root.shipping_method_id or get_external_shipping_id(
+            root
+        )
 
-        if not root.shipping_method_id:
+        if not order_shipping_method_id:
```
L3str4nge, line 1050: This check is redundant because. We could just `return root.shipping_method_id or get_external_shipping_id(root)`

L3str4nge (approved):

saleor/graphql/order/types.py
```diff
@@ -1042,39 +1042,25 @@ def _resolve_user(user):
...
 
-        def wrap_shipping_method_with_channel_context(data):
-            shipping_method, channel = data
-            return ChannelContext(node=shipping_method, channel_slug=channel.slug)
-
-        shipping_method = ShippingMethodByIdLoader(info.context).load(
-            root.shipping_method_id
+        method = ShippingMethodData(
+            id=str(order_shipping_method_id),
+            name=root.shipping_method_name or "",
```
L3str4nge, line 1055: Could we get shipping method name in the same way like `get_external_shipping_id`. Leaving it empty would be confusing.

saleor/graphql/order/types.py
```diff
@@ -1042,39 +1042,25 @@ def _resolve_user(user):
...
+        if not order_shipping_method_id:
             return None
 
-        def wrap_shipping_method_with_channel_context(data):
-            shipping_method, channel = data
-            return ChannelContext(node=shipping_method, channel_slug=channel.slug)
-
-        shipping_method = ShippingMethodByIdLoader(info.context).load(
-            root.shipping_method_id
+        method = ShippingMethodData(
```
korycins, line 1053: we should use the previous way of fetching shipping method here. In case when we use `external_shipping` then we should dynamically generate ShippingMethodData, but only for external shipping.

fowczarek (approved):

IKarbowiak (approved):

stnatic:
Please add a test that selects a price attribute. You are assigning `TaxedMoney` to a field that is supposed to be `Money` which is going to result in an AttributeError.

korycins (approved):

maarcingebala (approved):

PR #8696, opened 2021-12-01
Refactor connection fields

rafalp:
This refactor implements following changes:

- Moves filtering/slicing logic from `Field`s instances to `create_connection_slice` and `filter_connection_queryset` utils, making it explicit and visible in resolvers.
- Replaces connection types and fields abstractions based on `graphene-django` with simpler and far less magic logic based on `graphene.relay`.

Purpose of those changes are two-fold:

- Reduce reliance on `graphene-django` to bring us closer to the goal of removing it as dependency
- Make filtering/pagination logic explicit and visible in resolvers, so we can make it async in future and make it easier to pick places that should use dataloaders for loading items.

Schema and business logic is not changed.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


fowczarek (approved):

saleor/graphql/core/fields.py
```diff
@@ -36,231 +26,51 @@ def patch_pagination_args(field: DjangoConnectionField):
     )
 
 
-class BaseConnectionField(graphene.ConnectionField):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+class ConnectionField(graphene.Field):
+    def __init__(self, type_, *args, **kwargs):
+        kwargs.setdefault("before", graphene.String())
```
maarcingebala, line 31: If we are defining these fields here, can we also set the descriptions here? Then we could drop `patch_pagination_args`.

maarcingebala (approved):

PR #8733, opened 2021-12-06
Add `fetch_order_prices_if_expired`

mateuszgrzyb:
I want to merge this change because I want to add `fetch_order_prices_if_expired` function, which is a new method of calculating and storing taxes for orders.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-4364 Add function for handling order prices](https://app.clickup.com/t/2549495/SALEOR-4364)

saleor/order/calculations.py
```diff
@@ -0,0 +1,176 @@
+from decimal import Decimal
```
korycins, line 1: Missing docstring for all new functions/methods

saleor/order/calculations.py
```diff
@@ -0,0 +1,176 @@
...
+    lines: Optional[Iterable[OrderLine]] = None,
+    force_update: bool = False,
+) -> Order:
+    if not force_update and order.price_expiration_for_unconfirmed < timezone.now():
+        return order
+
+    if lines is None:
+        lines = list(order.lines.all())
+
+    _apply_tax_data_from_plugins(manager, order, lines)
```
korycins, line 82: Is it expected that we don't handle potential exceptions which can be raised here?

saleor/order/calculations.py
```diff
@@ -0,0 +1,176 @@
...
+def order_line_unit(
+    order: Order,
+    line: OrderLine,
+    manager: PluginsManager,
+    order_lines: Optional[Iterable[OrderLine]] = None,
+    force_update: bool = False,
+) -> TaxedMoney:
+    return (
+        fetch_order_prices_if_expired(order, manager, order_lines, force_update)
+        .lines.get(pk=line.pk)
```
korycins, line 124: Using `.get` here will call immediately a DB call.

saleor/order/calculations.py
```diff
@@ -0,0 +1,176 @@
...
+def order_line_total(
+    order: Order,
+    line: OrderLine,
+    manager: PluginsManager,
+    order_lines: Optional[Iterable[OrderLine]] = None,
+    force_update: bool = False,
+) -> TaxedMoney:
+    return (
+        fetch_order_prices_if_expired(order, manager, order_lines, force_update)
+        .lines.get(pk=line.pk)
```
korycins, line 138: Same as above. We can't call again DB only to fetch the line which we already have fetched from DB.

saleor/order/calculations.py
```diff
@@ -0,0 +1,176 @@
...
+def order_line_tax_rate(
+    order: Order,
+    line: OrderLine,
+    manager: PluginsManager,
+    order_lines: Optional[Iterable[OrderLine]] = None,
+    force_update: bool = False,
+) -> Decimal:
+    return (
+        fetch_order_prices_if_expired(order, manager, order_lines, force_update)
+        .lines.get(pk=line.pk)
```
korycins, line 152: As above ☝️

saleor/graphql/order/mutations/orders.py
```diff
@@ -339,11 +347,9 @@ def perform_mutation(cls, _root, info, **data):
         clean_order_update_shipping(order, method)
 
         order.shipping_method = method
-        shipping_price = info.context.plugins.calculate_order_shipping(order)
+        shipping_price = calculations.order_shipping(order, manager)
```
korycins, line 350: AFAIR we wanted to delay calculations of new prices, I think this should be also done for shipping. This is why we have a flag expiration for prices. The calculation should be done, when storefront/dashboard calls an expected resolver that needs to have calculated prices.

saleor/order/calculations.py
```diff
@@ -0,0 +1,212 @@
...
+    lines: Optional[Iterable[OrderLine]] = None,
+    force_update: bool = False,
+) -> Tuple[Order, Iterable[OrderLine]]:
+    """Fetch order prices with taxes.
+
+    First calculate and apply all order prices with taxes separately,
+    then apply tax data as well if we receive one.
+
+    Prices can be updated only if force_update == True, or if time elapsed from the
+    last price update is greater than settings.ORDER_PRICES_TTL.
```
korycins, line 82: Is `last price update is greater than settings.ORDER_PRICES_TTL.` correct? I see that you update a prices when expiration time is older than timezone.now() 🤔

saleor/order/calculations.py
```diff
@@ -0,0 +1,212 @@
...
+def order_line_unit(
+    order: Order,
+    order_line: OrderLine,
+    manager: PluginsManager,
+    lines: Optional[Iterable[OrderLine]] = None,
+    force_update: bool = False,
+) -> TaxedMoney:
+    """Return the unit price of provided line, taxes included.
+
+    Use it instead of calling manager methods directly.
```
korycins, line 143: I don't think that you need to explictly write here `Use it instead...` 
IMO you should just write that we call expected methods on plugins and store the price in the model if prices are expired etc.

saleor/order/calculations.py
```diff
@@ -0,0 +1,212 @@
...
+) -> Tuple[Order, Iterable[OrderLine]]:
+    """Fetch order prices with taxes.
+
+    First calculate and apply all order prices with taxes separately,
+    then apply tax data as well if we receive one.
+
+    Prices can be updated only if force_update == True, or if time elapsed from the
+    last price update is greater than settings.ORDER_PRICES_TTL.
+    """
+    if lines is None:
```
korycins, line 84: Also, I would consider adding here an `if` which will protect us from re-calculating prices for orders with different types than `DRAFT` or `UNCONFIRMED`.

saleor/order/utils.py
```diff
@@ -187,14 +187,11 @@ def recalculate_order_weight(order):
 def update_taxes_for_order_line(
```
korycins, line 187: Is the name still valid? We have now logic with expiration which means, that we could call this method and we won't update anything.

saleor/settings.py
```diff
@@ -605,6 +605,11 @@ def get_host():
     seconds=parse(os.environ.get("CHECKOUT_PRICES_TTL", "5 minutes"))
 )
 
+ORDER_PRICES_TTL = timedelta(
+    seconds=parse(os.environ.get("CHECKOUT_PRICES_TTL", "5 minutes"))
```
korycins, line 609: wrong env.

saleor/order/utils.py
```diff
@@ -236,11 +231,8 @@ def update_order_prices(order: Order, manager: "PluginsManager", tax_included: b
     update_taxes_for_order_lines(order.lines.all(), order, manager, tax_included)
 
     if order.shipping_method:
-        shipping_price = manager.calculate_order_shipping(order)
-        order.shipping_price = shipping_price
-        order.shipping_tax_rate = manager.get_order_shipping_tax_rate(
-            order, shipping_price
-        )
+        order.shipping_price = calculations.order_shipping(order, manager)
```
korycins, line 234: If I am not wrong, you're calling  `.save()` twice for these values. Inside a `fetch_if_expired` and here. Keep an eye on such things as we can't allow for duplicating DB requests as it has huge impacts on performance overall.

saleor/order/calculations.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def _apply_tax_data_from_plugins(
+    manager: PluginsManager, order: Order, lines: Iterable[OrderLine]
+):
+    for line in lines:
+        variant = line.variant
+        if not variant:
+            continue
+        product = variant.product
```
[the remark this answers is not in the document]
fowczarek: To resolve that you add prefetch in `fetch_order_prices_if_expired`  when lines are not passed. 
Did you check that all calls of `fetch_order_prices_if_expired` with `lines` passed contain this prefetch?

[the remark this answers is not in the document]
mateuszgrzyb: Right now, I think it will be better to just always prefetch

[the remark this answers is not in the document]
fowczarek: Maybe we could use `prefetch_related_objects`?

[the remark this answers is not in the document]
mateuszgrzyb: I've already pushed the changes with new prefetch, please check

fowczarek (approved):

saleor/order/utils.py
```diff
@@ -234,25 +227,18 @@ def update_taxes_for_order_lines(
...
-            update_fields=[
-                "shipping_price_net_amount",
-                "shipping_price_gross_amount",
-                "shipping_tax_rate",
-                "currency",
-            ]
-        )
+        # recalculate order.shipping_price and order.shipping_tax_rate
+        fetch_order_prices_if_expired(order, manager)
+        order.save(update_fields=["currency"])
```
maarcingebala, line 241: Is this save needed? A save is already done in `fetch_order_prices_if_expired`. Also, what's the use-case with saving the currency, can it change here?

saleor/order/calculations.py
```diff
@@ -0,0 +1,261 @@
...
+) -> Optional[TaxData]:
+    prefetch_related_objects(lines, "variant__product")
+    currency = order.currency
+
+    tax_lines_data = []
+    subtotal = zero_taxed_money(currency)
+    try:
+        for line in lines:
+            variant = line.variant
+            if not variant:
```
korycins, line 27: I think that you should include the line prices to subtotal. Even if you're not able to update prices for given line.

saleor/order/utils.py
```diff
@@ -218,24 +216,19 @@ def update_taxes_for_order_line(
...
+        )
     OrderLine.objects.bulk_update(
         lines,
         [
-            "tax_rate",
-            "unit_price_net_amount",
-            "unit_price_gross_amount",
-            "total_price_net_amount",
-            "total_price_gross_amount",
             "undiscounted_unit_price_gross_amount",
```
korycins, line 232: What is a benefit of having part update of prices fields in order.calculations file and the rest of them here? Wouldn't be good to extend functions in order.calculations and include them a rest of the methods responsible for returning prices? Fields like `undiscounted_unit_price...`.

saleor/order/utils.py
```diff
@@ -198,17 +199,14 @@ def recalculate_order_weight(order):
     order.save(update_fields=["weight"])
 
 
-def update_taxes_for_order_line(
+def update_order_line_prices_if_expired(
```
korycins, line 202: I don't think that this method will be needed if we move the rest of the calculations of prices to `order.calculations` file. I believe, all of this value will be already updated as we will just call `fetch_order_prices_if_expired`.

saleor/order/calculations.py
```diff
@@ -0,0 +1,261 @@
...
+from ..core.taxes import TaxData, TaxError, TaxLineData, zero_taxed_money
+from ..plugins.manager import PluginsManager
+from . import ORDER_EDITABLE_STATUS
+from .models import Order, OrderLine
+
+
+def _combine_prices_from_manager_into_tax_data(
+    manager: PluginsManager, order: Order, lines: Iterable[OrderLine]
+) -> Optional[TaxData]:
+    prefetch_related_objects(lines, "variant__product")
```
korycins, line 19: Is it expected that this will work only partially? For some methods that calls `update_order_prices_if_expired` you don't pass a lines, which means that it is None, and then you're calling `lines = list(order.lines.all())`, which means that we will not use any benefit of prefetching a data.

saleor/order/calculations.py
```diff
@@ -0,0 +1,261 @@
...
+        order_line.total_price = qp(
+            net=tax_line.total_net_amount, gross=tax_line.total_gross_amount
+        )
+        order_line.tax_rate = tax_line.tax_rate
+
+
+def fetch_order_prices_if_expired(
+    order: Order,
+    manager: PluginsManager,
+    lines: Optional[Iterable[OrderLine]] = None,
```
korycins, line 106: I am not conviced that this parameter should be optional. IMHO I would consider treat it as a mandatory parameter, as in many places we already have a prefetched data.

saleor/order/utils.py
```diff
@@ -244,30 +237,20 @@ def update_taxes_for_order_lines(
     )
 
 
-def update_order_prices(
+def update_order_prices_if_expired(
```
korycins, line 240: Why do we call `update_order_lines_prices_if_expired`  and `fetch_order_prices_if_expired`? I thought that new way of fetching prices will always calculate all prices from a beginning (if expired). Can't we just call `fetch_order_prices_if_expired` once here? 
Also, maybe I am missing something but I am not convinced that we should mix in a calculation of order prices with invalidating them. 
I would say that current flow is incorrect, especially with flag invalidate_prices=True:
1. call update_order_lines_prices_if_expired
2. call fetch_order_prices_if_expired (which I believe calls the same logic as in 1st point)
3. call `recalculate_order` with invalidate_prices=True which means that this will overwrite and force to re-calculate all prices once again. 

Also, I am not sure that invalidating prices is something that belongs to `recalculate_order`.

korycins: I don't think that we still need functions like this one. There is no need to fetch prices in advance.  If we need to have a prices we can use a interface to just take them like `order.calculation.unit_price`. 
Mutations for draf/unconfirmed orders should invalidate a price expiration but nothing more.

saleor/order/calculations.py
mateuszgrzyb: Resolved in a recent commit.

saleor/order/calculations.py
```diff
@@ -0,0 +1,268 @@
...
+
+from ..core.prices import quantize_price
+from ..core.taxes import TaxData, TaxError, zero_taxed_money
+from ..plugins.manager import PluginsManager
+from . import ORDER_EDITABLE_STATUS
+from .interface import OrderTaxedPricesData
+from .models import Order, OrderLine
+
+
+def _apply_tax_data_from_manager(
```
fowczarek, line 17: Should we add dosctring for this function what mean if they return False or True?

saleor/order/calculations.py
```diff
@@ -0,0 +1,268 @@
...
+
+            order_line.tax_rate = price_line["tax_rate"]
+
+        return False
+
+
+def _apply_tax_data(
+    order: Order, lines: Iterable[OrderLine], tax_data: TaxData
+) -> None:
+    def qp(net: Decimal, gross: Decimal) -> TaxedMoney:
```
fowczarek, line 66: Maybe? This code should be more readable.
```suggestion
    def _quantize_price(net: Decimal, gross: Decimal) -> TaxedMoney:
```

saleor/order/calculations.py
```diff
@@ -0,0 +1,268 @@
...
+    if lines is None:
+        lines = list(order.lines.prefetch_related("variant__product"))
+    else:
+        prefetch_related_objects(lines, "variant__product")
+
+    if order.status not in ORDER_EDITABLE_STATUS:
+        return order, lines
+
+    if not force_update and order.price_expiration_for_unconfirmed < timezone.now():
+        return order, lines
```
fowczarek, line 127: This check should be performed before fetching lines for line 119.

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -287,18 +283,10 @@ def _refresh_lines_unit_price(cls, info, instance, cleaned_input, new_instance):
...
-            )
+            invalidate_order_prices(instance, save=True)
         billing_address = cleaned_input.get("billing_address")
         if billing_address and not instance.is_shipping_required():
-            update_order_prices(
-                instance,
-                info.context.plugins,
-                info.context.site.settings.include_taxes_in_prices,
-            )
+            invalidate_order_prices(instance, save=True)
```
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit.

saleor/graphql/order/mutations/orders.py
```diff
@@ -247,12 +247,9 @@ def save(cls, info, instance, cleaned_input):
...
         )
 
-        update_order_prices(
-            instance,
-            info.context.plugins,
-            info.context.site.settings.include_taxes_in_prices,
-            invalidate_prices,
-        )
+        if invalidate_prices:
+            invalidate_order_prices(instance, save=True)
```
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit.

saleor/order/migrations/0118_order_price_expiration_for_unconfirmed.py
```diff
@@ -1,5 +1,3 @@
-# Generated by Django 3.2.6 on 2021-12-06 18:40
-
```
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit.

saleor/order/calculations.py
mateuszgrzyb: Resolved in recent commit.

saleor/order/calculations.py
mateuszgrzyb: Resolved in recent commit.

saleor/order/calculations.py
mateuszgrzyb: Resolved in recent commit.

saleor/order/calculations.py
```diff
@@ -0,0 +1,281 @@
...
+            )
+
+            price_lines.append({"tax_rate": tax_rate, "unit": unit, "total": total})
+
+        subtotal = sum((line.total_price for line in lines), zero_taxed_money(currency))
+        shipping_price = manager.calculate_order_shipping(order)
+        shipping_tax_rate = manager.get_order_shipping_tax_rate(order, shipping_price)
+        total = shipping_price + subtotal
+    except TaxError:
+        return True
```
L3str4nge, line 50: I don't understand. Why we are returning `True` when something goes wrong and `False` otherwise? Does it make sense?

saleor/order/calculations.py
```diff
@@ -0,0 +1,281 @@
...
+
+def _apply_tax_data_from_manager(
+    manager: PluginsManager, order: Order, lines: Iterable[OrderLine]
+) -> bool:
+    """Apply tax data from manager methods into order and lines.
+
+    If any of the manager order tax methods raises `TaxError`,
+    order and lines stay unchanged.
+
+    :return: True if `TaxError` was raised, False otherwise.
```
L3str4nge, line 25: I can't compile this statement in my mind :D

IKarbowiak (approved):

fowczarek (approved):

saleor/order/calculations.py
maarcingebala: Can you please try to rewrite this with smaller functions maybe? The current solution is confusing as the function returns unexpected values, according to its name.

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -285,28 +282,23 @@ def _refresh_lines_unit_price(cls, info, instance, cleaned_input, new_instance):
         if new_instance:
             # It is a new instance, all new lines have already updated prices.
             return
+
         shipping_address = cleaned_input.get("shipping_address")
```
korycins, line 286: I think that there is a little bit messy here. We drop calculation of prices but we still have methods like `_refresh_lines_unit_price`, which can be a little bit confusing as we don't refresh anything, only invalidating a flag that determines if prices will need a new fetch.

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -285,28 +282,23 @@ def _refresh_lines_unit_price(cls, info, instance, cleaned_input, new_instance):
...
 
     @classmethod
     def invalidate_prices(cls, instance, cleaned_input) -> bool:
-        return False
+        return cls._invalidate_prices
 
     @classmethod
     @traced_atomic_transaction()
     def save(cls, info, instance, cleaned_input):
+        cls._invalidate_prices = False
```
korycins, line 301: line - 337 `recalculate_order` calculates prices. Why do we need it? It will be fully replaced when someone asks about any order price, right?

saleor/order/calculations.py
```diff
@@ -0,0 +1,291 @@
...
+
+            unit = manager.calculate_order_line_unit(order, line, variant, product)
+            total = manager.calculate_order_line_total(order, line, variant, product)
+            tax_rate = manager.get_order_line_tax_rate(
+                order, product, variant, None, unit.undiscounted_price
+            )
+
+            price_lines.append({"tax_rate": tax_rate, "unit": unit, "total": total})
+
+        subtotal = sum((line.total_price for line in lines), zero_taxed_money(currency))
```
korycins, line 42: Maybe I am missing something, but in that way, you will not update taxes for subtotal. (you're using `lines` list, and `price_lines` contains the taxes data).

saleor/order/calculations.py
```diff
@@ -0,0 +1,291 @@
...
+        shipping_price = manager.calculate_order_shipping(order)
+        shipping_tax_rate = manager.get_order_shipping_tax_rate(order, shipping_price)
+        total = shipping_price + subtotal
+    except TaxError:
+        return None
+    else:
```
korycins, line 48: I think, else is not needed here

saleor/order/calculations.py
```diff
@@ -0,0 +1,291 @@
...
+            "total": total,
+            "subtotal": subtotal,
+            "shipping_price": shipping_price,
+            "shipping_tax_rate": shipping_tax_rate,
+            "lines": price_lines,
+        }
+
+
+def _apply_tax_data_from_manager(
+    tax_data: dict, order: Order, lines: Iterable[OrderLine]
```
L3str4nge, line 59: Where is manager? I don't see it in parameters and function body as well.

saleor/order/calculations.py
```diff
@@ -0,0 +1,291 @@
...
+        order_line.undiscounted_total_price = (
+            order_line.undiscounted_unit_price * order_line.quantity
+            if order_line.unit_discount
+            else order_line.total_price
+        )
+
+        order_line.tax_rate = tax_line.tax_rate
+
+
+def fetch_order_prices_if_expired(
```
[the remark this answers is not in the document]
korycins: I think that `update_order_prices` is already removed here. This method should be a single way of updating an order's prices

saleor/order/calculations.py
```diff
@@ -0,0 +1,291 @@
...
+
+    if lines is None:
+        lines = list(order.lines.prefetch_related("variant__product"))
+    else:
+        prefetch_related_objects(lines, "variant__product")
+
+    manager_tax_data = _get_tax_data_from_manager(manager, order, lines)
+
+    if manager_tax_data is None:
+        return order, lines
```
korycins, line 149: I think that this is incorrect. With this `if` we will not call a `manager.get_taxes_for_order` in the case the old approach returns an exception.

saleor/order/calculations.py
```diff
@@ -0,0 +1,291 @@
...
+
+    _apply_tax_data_from_manager(manager_tax_data, order, lines)
```
korycins, line 151: What is a reason for splitting logic between `_get_tax_data_from_manager` and `_apply_tax_data_from_manager`? 
Can 't we have a single method that will iterate over lines, call old tax methods for each line and immiedietly assign it to Order object?  Then we could call the new tax webhook. At the end call save on order and lines.

korycins (commented):
Please double check the files related with the orders. 
I think that methods like - `recalculate_order_prices`, should not be needed anymore.  All calculations should be go through the function that this PR introduces.

saleor/order/utils.py
```diff
@@ -410,19 +339,13 @@ def add_variant_to_order(
...
             variant=variant,
         )
-        unit_price_data = manager.calculate_order_line_unit(
-            order, line, variant, product
-        )
-        total_line_price_data = manager.calculate_order_line_total(
-            order, line, variant, product
-        )
+        unit_price_data = calculations.order_line_unit(order, line, manager)
+        total_line_price_data = calculations.order_line_total(order, line, manager)
```
korycins, line 343: Are you sure that this is still needed? This function calls a .save for prices.

saleor/order/utils.py
```diff
@@ -977,13 +890,4 @@ def remove_discount_from_order_line(
...
-    order_line.save(
-        update_fields=[
-            "unit_price_gross_amount",
-            "unit_price_net_amount",
-            "total_price_net_amount",
-            "total_price_gross_amount",
-            "tax_rate",
-        ]
-    )
+    fetch_order_prices_if_expired(order, manager)
```
korycins, line 893: All calculations of new prices should go thourhg the fetch_order_prices (see lines above 880-891). Also `fetch_order_prices_if_expired` has been build to delay a price calculation, and call it on the resolver and palces where we really need a new prcies. It is not 1:1 a replacment for `update_order_prices` method, so we can't replace one with the other. If we change something in order, we reset an expiration time.

saleor/order/calculations.py
```diff
@@ -0,0 +1,318 @@
...
+
+            subtotal += line_total.price_with_discounts
+
+        order.shipping_price = manager.calculate_order_shipping(order)
+        order.shipping_tax_rate = manager.get_order_shipping_tax_rate(
+            order, order.shipping_price
+        )
+        order.total = order.shipping_price + subtotal
+
+    except TaxError:
```
korycins, line 59: I am not sure of this. We will break tax calculations even if a single line will return an exception.  What values will be placed under these fields in case of calling this function for the newly created order?

saleor/order/calculations.py
```diff
@@ -0,0 +1,318 @@
...
+    """Fetch order prices with taxes.
+
+    First calculate and apply all order prices with taxes separately,
+    then apply tax data as well if we receive one.
+
+    Prices can be updated only if force_update == True,
+    or if order price expiration time was exceeded
+    (which is settings.ORDER_PRICES_TTL if the prices are not invalidated).
+    """
+    with transaction.atomic():
```
maarcingebala, line 130: Since the entire logic is in the context manager, maybe it would be more readable to decorate the function with `transaction.atomic()`?

saleor/graphql/order/mutations/utils.py
```diff
@@ -1,24 +1,19 @@
-from typing import List
-
 from django.utils import timezone
 
 from ....order import ORDER_EDITABLE_STATUS
 from ....order.models import Order
 
 
-def invalidate_order_prices(order: Order, *, save: bool) -> List[str]:
+def invalidate_order_prices(order: Order, *, save: bool) -> None:
```
maarcingebala, line 7: Could this function be potentially useful outside of mutations? I'm wondering if we shouldn't move it to `saleor/order/utils.py`.

maarcingebala (approved):
In general, these changes look much better now, I'm happy that we managed to drop old calculation functions and now we have them in one place 👍

saleor/order/calculations.py
stnatic: We should push it down. Starting a transaction everytime this function is called makes a db connection which can add up to ~20-30 redundant queries per request

stnatic: I've done this in #9242 so let's leave it out of this PR

saleor/graphql/order/mutations/discount_order.py
```diff
@@ -200,7 +185,7 @@ def perform_mutation(cls, root, info, **data):
         order_discount.value_type = value_type
         order_discount.save()
 
-        cls.recalculate_order(order)
+        fetch_order_prices_if_expired(order, info.context.plugins, force_update=True)
```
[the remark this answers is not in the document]
stnatic: I think we need to take order webhook serialization into account (e.g. order_updated). This means that the order serializer should actually use the calculations module instead of relying on the model attributes

saleor/graphql/checkout/tests/test_checkout_price_expiration.py
```diff
@@ -6,8 +6,9 @@
 from django.utils import timezone
 from freezegun import freeze_time
 
+from saleor.checkout.utils import invalidate_checkout_prices
```
korycins, line 9: Shouldn't be relative import here?

saleor/webhook/payloads.py
```diff
@@ -287,7 +301,7 @@ def generate_order_payload(
     payments = order.payments.all()
     discounts = order.discounts.all()
 
-    quantize_price_fields(order, ORDER_PRICE_FIELDS, order.currency)
+    # quantize_price_fields(order, ORDER_PRICE_FIELDS, order.currency)
```
korycins, line 304: Is it expected?

korycins (approved):

saleor/graphql/order/mutations/discount_order.py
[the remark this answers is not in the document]
stnatic: Yes

saleor/order/calculations.py
mateuszgrzyb: I've changed exception handling to be more granular, so that `order.total` is always calculated.

PR #8738, opened 2021-12-06
Port ShippingMethodData to 3.0

stnatic:
I want to merge this change because I would like to stop annotating ShippingMethod instances with custom attributes such as price

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/checkout/fetch.py
```diff
@@ -150,71 +224,69 @@ def update_checkout_info_shipping_address(
...
-    checkout_info.shipping_method_channel_listings = (
-        (
-            ShippingMethodChannelListing.objects.filter(
-                shipping_method=shipping_method, channel=checkout_info.channel
-            ).first()
-        )
-        if shipping_method
-        else None
-    )
+    checkout_info.delivery_method_info = get_delivery_method_info(None)
```
stnatic, line 292: This looks like a mistake

saleor/graphql/shipping/types.py
```diff
@@ -248,7 +249,7 @@ def resolve_channels(root: ChannelContext[models.ShippingZone], info, **_kwargs)
         return ChannelsByShippingZoneIdLoader(info.context).load(root.node.id)
 
 
-class ShippingMethod(ChannelContextType, MetadataMixin):
+class ShippingMethod(ObjectType):
```
korycins, line 252: Maybe I am missing something, but you dropped a MetadataMixin, which means that you will not have a metadata fields here. It is strange as I don't see an any scheme change 🤔

stnatic: Metadata fields are being added by the interface defined down below. MetadataMixin was there to unwrap the ChannelContext which is now obsolete

saleor/checkout/fetch.py
```diff
@@ -56,6 +54,66 @@ def get_country(self) -> str:
...
+
+    def is_valid_delivery_method(self) -> bool:
+        return False
+
+    def is_method_in_valid_methods(self, checkout_info: "CheckoutInfo") -> bool:
+        return False
+
+
+@dataclass(frozen=True)
+class ShippingMethodInfo(DeliveryMethodBase):
```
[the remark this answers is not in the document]
stnatic: We need the base as it's used in the business logic:
```
    if delivery_method is None:
        return DeliveryMethodBase()
```

I agree that we could drop some unnecessary attributes here

saleor/checkout/base_calculations.py
```diff
@@ -36,12 +52,12 @@ def base_checkout_shipping_price(
...
-    shipping_price = shipping_method.channel_listings.get(
-        channel_id=checkout_info.checkout.channel_id,
-    ).get_total()
 
-    return quantize_price(
-        TaxedMoney(net=shipping_price, gross=shipping_price), shipping_price.currency
+    # Base price does not yet contain tax information,
+    # which can be later applied by tax plugins
+    return TaxedMoney(
+        net=shipping_method.price,
```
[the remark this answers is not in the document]
stnatic: We're not making any operations that could break the precision, but maybe this is the "right level" where quantization should be applied for safety reasons

saleor/graphql/shipping/dataloaders.py
```diff
@@ -111,6 +111,24 @@ def batch_load(self, keys):
         ]
 
 
+class ShippingMethodChannelListingByChannelSlugLoader(DataLoader):
+    context_key = "shippingmethodchannellisting_by_channel"
+
+    def batch_load(self, keys):
+        shipping_method_channel_listings = ShippingMethodChannelListing.objects.filter(
```
fowczarek, line 118: We should add `.using(self.database_connection_name)` for all db queries in dataloaders. I sill working on docs for this database read replicas. 
```suggestion
        shipping_method_channel_listings = ShippingMethodChannelListing.objects.using(
            self.database_connection_name
        ).filter(
```

saleor/graphql/translations/resolvers.py
```diff
@@ -28,6 +29,9 @@
     shipping_models.ShippingMethod: (
         dataloaders.ShippingMethodTranslationByIdAndLanguageCodeLoader
     ),
+    shipping_interface.ShippingMethodData: (
+        dataloaders.ShippingMethodTranslationByIdAndLanguageCodeLoader
+    ),
```
fowczarek, line 34: Should we add new tests for that?

stnatic: The tests are already there, we just changed the root type from `ShippingMethod` to `ShippingMethodData`. Commenting out this line results in:

```
TypeError: No dataloader found to <class 'saleor.shipping.interface.ShippingMethodData'>
========================================================================= short test summary info ==========================================================================
FAILED saleor/graphql/checkout/tests/test_checkout.py::test_checkout_selected_shipping_method - AssertionError: [{'message': "No dataloader found to <class 'saleor.shipp...
FAILED saleor/graphql/checkout/tests/test_checkout.py::test_checkout_available_shipping_methods_with_weight_based_shipping_method[0] - AssertionError: [{'message': "No d...
FAILED saleor/graphql/checkout/tests/test_checkout.py::test_checkout_available_shipping_methods_with_weight_based_shipping_method[None] - AssertionError: [{'message': "N...
FAILED saleor/graphql/checkout/tests/test_checkout.py::test_checkout_available_shipping_methods_with_weight_based_shipping_method[2] - AssertionError: [{'message': "No d...
FAILED saleor/graphql/checkout/tests/test_checkout.py::test_checkout_available_shipping_methods_weight_method_with_higher_minimal_weigh - AssertionError: [{'message': "N...
FAILED saleor/graphql/checkout/tests/test_checkout.py::test_checkout_available_shipping_methods[shippingMethods] - AssertionError: [{'message': "No dataloader found to <...
FAILED saleor/graphql/checkout/tests/test_checkout.py::test_checkout_available_shipping_methods_with_price_displayed - AssertionError: [{'message': "No dataloader found ...
FAILED saleor/graphql/checkout/tests/test_checkout.py::test_checkout_shipping_methods_with_price_based_shipping_method_and_discount - AssertionError: [{'message': "No da...
FAILED saleor/graphql/checkout/tests/test_checkout.py::test_checkout_available_shipping_methods[availableShippingMethods] - AssertionError: [{'message': "No dataloader f...
========================================================== 9 failed, 5774 passed, 10 skipped in 63.98s (0:01:03) ===========================================================
```

saleor/checkout/fetch.py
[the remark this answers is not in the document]
stnatic: My intention was do do the most of braking changes in the `CheckoutInfo` interface in 3.0. We are doing "some" at least, i.e. dropping the channel listing and changing types from `ShippingMethod` to `ShippingMethodData`. Having a delivery method base in 3.0 (just for shipping methods, without collection points) would result in less conflict overall.

saleor/graphql/shipping/types.py
```diff
@@ -305,60 +306,51 @@ class Meta:
         )
 
     @staticmethod
-    def resolve_type(root: ChannelContext[models.ShippingMethod], info, **_kwargs):
-        return root.node.type
+    def resolve_type(root: ShippingMethodData, info, **_kwargs):
```
korycins, line 309: Do you still need these resolvers? Maybe I am missing something, but default graphene logic should be able to handle simple resolvers if the field name matches the field in the object.

korycins (approved):

saleor/graphql/shipping/types.py
stnatic: You are right, previously unit tests were failing, but this could have been due to some legacy ChannelContext logic. I've removed all those obsolete resolvers

saleor/checkout/fetch.py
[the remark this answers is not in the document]
maarcingebala: While "deliveries" are a concept from 3.1, I think that it makes sense to start using this name in 3.0 - it will make the codebase more similar to what we have in 3.1 which is a good thing.

saleor/checkout/utils.py
```diff
@@ -571,30 +570,45 @@ def get_valid_shipping_methods_for_checkout(
...
-        country_code=country_code,  # type: ignore
+        country_code=country_code,
         lines=lines,
     )
 
+    channel_listings_map = {
+        listing.shipping_method_id: listing for listing in shipping_channel_listings
+    }
+
+    saleor_methods = []
```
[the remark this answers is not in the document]
stnatic: I dropped the "saleor" prefix right now, as the difference is apparent only on 3.1. There it was previously called "local" which I think had a bad connotation with being a city-wide same-day delivery option.

maarcingebala (approved):

PR #8744, opened 2021-12-06
Remove repeated shipping API calls from checkout lines mutations

rafalp:
Reduces number of shipping API calls performed by Checkout Lines mutations from 4 to 1 by removing redundant API calls from `fetch_checkout_info`.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/checkout/tests/benchmark/test_checkout_mutations.py
+++ b/saleor/graphql/checkout/tests/benchmark/test_checkout_mutations.py
@@ -706,11 +706,7 @@ def test_update_checkout_lines_with_reservations(
         reservation_length=5,
     )
 
-<<<<<<< HEAD
-    with django_assert_num_queries(60):
-=======
-    with django_assert_num_queries(56):
->>>>>>> 95bb73108... Remove repeated shipping API calls from checkout lines mutations
+    with django_assert_num_queries(57):
         variant_id = graphene.Node.to_global_id("ProductVariant", variants[0].pk)
         variables = {
             "token": checkout.token,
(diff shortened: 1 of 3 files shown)


fowczarek (approved):

fowczarek:
@rafalp We also have similar logic in `CheckoutInfoByCheckoutTokenLoader` dataloader. Should we introduce any change in them?

rafalp:
@fowczarek I haven't heard of any performance issues happening there, but IMHO we should have a task to cleanup changes introduced in this PR and propagating them to across similar places.

stnatic:
This looks like a massive conflict with #8737

rafalp:
@stnatic C'est la vie ;) I think we should consider either merging this and then adapting #8737 so it also doesn't cause extra webhook calls, or skipping this PR and adapting #8737 in its place. But I don't know how much time we have, I've got impression that getting rid of those extra API calls was urgent for us.

stnatic:
I made an alternative PR (#8772) that uses `SimpleLazyObject` in order to add lazy evaluation to the attributes in question, it seems to pass the latest test added by Rafał and it's the same solution that's used in #8737.

rafalp:
I've added asserts on webhook calls count to checkout benchmarks for two scenarios:

- Add lines to checkout without shipping
- Add lines to checkout with external shipping

stnatic:
Thanks @rafalp for pointing out the differences between our solutions and adding an extra test. I've made your tests pass on my original branch (#8737).

maarcingebala (approved):

stnatic:
Since the performance issue is mitigated in `main`, let's assume that we'll clean up the code as part of #8737. I'm already working on merging changes from Rafał and me

saleor/graphql/checkout/tests/benchmark/test_checkout_mutations.py
```diff
@@ -813,6 +912,10 @@ def test_add_checkout_lines(
         api_client.post_graphql(MUTATION_CHECKOUT_LINES_ADD, variables)
     )
     assert not response["data"]["checkoutLinesAdd"]["errors"]
+    # Two API calls:
+    # - mutate() logic
+    # - dataloader for lines totalPrice
```
stnatic, line 917: This API call should be avoided - totalPrice of an individual line should not depend on the selected shipping method. I believe that I've already covered it on my cleanup branch

PR #8103, opened 2021-09-15
Update checkout taxes by webhooks

mateuszgrzyb:
### I want to merge this change because...

Update Checkout and CheckoutLine models [SALEOR-4329](https://app.clickup.com/t/10hrrv3)
Add function for handling checkout prices [SALEOR-4336](https://app.clickup.com/t/118nt90)
Extend webhook checkout payload [SALEOR-4341](https://app.clickup.com/t/118ppp3)
Invalidate checkout prices [SALEOR-4343](https://app.clickup.com/t/118pr5y)

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #8747, opened 2021-12-06
Add tax rates to checkout prices calculation

mateuszgrzyb:
I want to merge this change because I want to integrate shipping tax rate and line tax rate into new implementation of checkout prices calculation.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


mateuszgrzyb:
Depends on #8103

IKarbowiak (approved):

saleor/checkout/calculations.py
```diff
@@ -110,24 +146,22 @@ def checkout_line_total(
     checkout_info: "CheckoutInfo",
     lines: Iterable["CheckoutLineInfo"],
     checkout_line_info: "CheckoutLineInfo",
+    address: Optional["Address"] = None,
```
korycins, line 149: Do we need it? Previously we were taking an address from checkout_info, which looked like an ok solution.

saleor/checkout/calculations.py
```diff
@@ -179,7 +235,7 @@ def fetch_checkout_prices_if_expired(
     address: Optional["Address"] = None,
     discounts: Optional[Iterable["DiscountInfo"]] = None,
     force_update: bool = False,
-) -> "Checkout":
+) -> Tuple["CheckoutInfo", Iterable["CheckoutLineInfo"]]:
     """Fetch checkout prices with taxes.
 
     Apply checkout prices with taxes from plugins and
```
korycins, line 241: I wouldn't say explicitly from webhooks. First we try the "old" approach, where any tax lines are calculated separatly, then we call a "new" approach, where we call single method which will return all prices fully calculated.

korycins (commented):
Missing docstring in some new functions.

fowczarek (approved):

PR #8823, opened 2021-12-17
Remove django objects

rafalp:
This PR replaces `DjangoObjectType` and `CountableDjangoObject` type in our schema with `ModelObjectType` that brings no special logic or magic with it, but it knows that it represents Django Model, so we can still scan schema's types for GraphQL types supposed to represent given model.

All fields are now explicitly declared on GraphQL types (vs being populated automagically by `only_fields`.

Reliance on global type -> model registry is also removed by this change. I'm not 100% sure because Graphene has it's own types registry I've haven't investigated in detail, but it looks like this brings us step closer to splitting our GraphQL schema into Admin and Public schema.

There's still few places that import different utils from `graphene_django` packages (mostly for filtering), those will need to be extracted to Saleor in next PR.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


korycins (approved):

saleor/graphql/translations/schema.py
```diff
@@ -43,6 +43,33 @@ class Meta:
...
+        if isinstance(instance, ShippingMethod):
+            return translation_types.ShippingMethodTranslatableContent
+        if isinstance(instance, Sale):
+            return translation_types.SaleTranslatableContent
+        if isinstance(instance, Voucher):
+            return translation_types.VoucherTranslatableContent
+        if isinstance(instance, MenuItem):
+            return translation_types.MenuItemTranslatableContent
+
+        return super(TranslatableItem, cls).resolve_type(instance, info)
```
fowczarek, line 71: Maybe we could create a map and simplify this function?

fowczarek (approved):
Missing changelog :)

saleor/graphql/translations/schema.py
rafalp: @fowczarek I've added mapping.

fowczarek (approved):

saleor/graphql/account/types.py
```diff
@@ -55,10 +50,21 @@ class AddressInput(graphene.InputObjectType):
 
 
 @key(fields="id")
-class Address(CountableDjangoObjectType):
+class Address(ModelObjectType):
```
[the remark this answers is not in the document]
maarcingebala: I think Owczar meant descriptions in the below fields - almost all of them are missing it.

maarcingebala (approved):

rafalp:
@maarcingebala don't worry about that, task is up already ;)

maarcingebala:
These changes look great but I have one concern - if we merged it today it would potentially make any ports from 3.0 to 3.1 more difficult, as all types are changed here. I'm wondering if we shouldn't wait with merging this one until stable 3.0 is released. Will discuss it with @fowczarek and @korycins.

PR #8880, opened 2021-12-30
skip update taxes for lines with deleted variant

jakubkuc:
I want to merge this change because it fixes problem with updating orders with deleted variants

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/order/tests/test_order_utils.py
+++ b/saleor/order/tests/test_order_utils.py
@@ -1,3 +1,4 @@
+import copy
 from decimal import Decimal
 from unittest.mock import Mock
 
@@ -13,6 +14,7 @@ from ..utils import (
     change_order_line_quantity,
     get_valid_shipping_methods_for_order,
     match_orders_with_new_user,
+    update_taxes_for_order_line,
     update_taxes_for_order_lines,
 )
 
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [SALEOR-5477 Changing address doesn't work in orders with deleted variants](https://app.clickup.com/t/2549495/SALEOR-5477)

saleor/order/utils.py
```diff
@@ -195,6 +195,8 @@ def update_taxes_for_order_line(
     line: "OrderLine", order: "Order", manager, tax_included
 ):
     variant = line.variant
+    if not variant:
+        return
```
IKarbowiak, line 199: Could you add a test for the situation which is fixed?

jakubkuc: added

IKarbowiak (approved):

saleor/order/tests/test_order_utils.py
```diff
@@ -326,3 +328,41 @@ def test_add_variant_to_order(order, customer_user, variant, site_settings):
...
+
+def test_update_taxes_for_order_line_deleted_variant(order_with_lines):
+    # given
+    order_line = order_with_lines.lines.first()
+    order_line_unchanged_copy = copy.deepcopy(order_line)
+    unit_discount_amount = Decimal("2.00")
+
+    unit_price = TaxedMoney(net=Money("10.23", "USD"), gross=Money("15.80", "USD"))
+    total_price = TaxedMoney(net=Money("30.34", "USD"), gross=Money("36.49", "USD"))
+    tax_rate = Decimal("0.23")
```
maarcingebala, line 341: Where are these values coming from, are they hardcoded? Will the test pass if we make an unrelated change in `order_with_lines` fixture and it would return different values?

jakubkuc: Those values are different from values in the fixture (they are copied from other tests from `test_order.utils.py`). `update_taxes_for_order_line` method should not update prices in line with those values. So changing anything in fixture should not interfere with the test. The test only checks if no changes in prices were made by `update_taxes_for_order_line` in line when variant is `None`.

maarcingebala: 👍

maarcingebala (approved):

PR #8897, opened 2022-01-03
Add in_shipping_zones attribute to shop countries query

jakubkuc:
I want to merge this change because  API query for fetching countries with or without shipping zones was needed.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/shop/filters.py
+++ b/saleor/graphql/shop/filters.py
@@ -2,4 +2,9 @@ import graphene
 
 
 class CountryFilterInput(graphene.InputObjectType):
-    in_shipping_zones = graphene.Boolean()
+    in_shipping_zones = graphene.Boolean(
+        description="Boolean for filtering countries by having shipping zone assigned."
+        "If 'true', return countries with shipping zone assigned."
+        "If 'false', return countries without any shipping zone assigned."
+        "If the argument is not provided (null), return all countries."
+    )
--- a/saleor/graphql/shop/filters.py
+++ b/saleor/graphql/shop/filters.py
@@ -1,0 +1,5 @@
+import graphene
+
+
+class CountryFilterInput(graphene.InputObjectType):
+    in_shipping_zones = graphene.Boolean()
--- a/saleor/graphql/shop/filters.py
+++ b/saleor/graphql/shop/filters.py
@@ -2,7 +2,7 @@ import graphene
 
 
 class CountryFilterInput(graphene.InputObjectType):
-    in_shipping_zones = graphene.Boolean(
+    attached_to_shipping_zones = graphene.Boolean(
         description="Boolean for filtering countries by having shipping zone assigned."
         "If 'true', return countries with shipping zone assigned."
         "If 'false', return countries without any shipping zone assigned."
--- a/saleor/graphql/shop/resolvers.py
+++ b/saleor/graphql/shop/resolvers.py
@@ -1,8 +1,14 @@
+from django.utils import translation
+from django_countries import countries
+from django_prices_vatlayer.models import VAT
+
 from ...account.models import Address
 from ...core.tracing import traced_resolver
 from ...shipping.models import ShippingMethod, ShippingMethodChannelListing
 from ...shipping.postal_codes import filter_shipping_methods_by_postal_code_rules
 from ...shipping.utils import convert_to_shipping_method_data
+from ..core.types import CountryDisplay
+from .utils import get_countries_codes_list
 
 
 @traced_resolver
@@ -23,6 +29,22 @@ def resolve_available_shipping_methods(info, channel_slug: str, address):
         convert_to_shipping_method_data(method, shipping_mapping[method.pk])
         for method in available
     ]
+
+
+def resolve_countries(**kwargs):
+    countries_filter = kwargs.get("filter", {})
+    in_shipping_zones = countries_filter.get("in_shipping_zones")
+    language_code = kwargs.get("language_code")
+    taxes = {vat.country_code: vat for vat in VAT.objects.all()}
+    # DEPRECATED: translation.override will be dropped in Saleor 4.0
+    with translation.override(language_code):
+        return [
+            CountryDisplay(
+                code=country[0], country=country[1], vat=taxes.get(country[0])
+            )
+            for country in countries
+            if country[0] in get_countries_codes_list(in_shipping_zones)
+        ]
 
 
 def get_shipping_method_to_shipping_listing_mapping(shipping_methods, channel_slug):
--- a/saleor/graphql/shop/resolvers.py
+++ b/saleor/graphql/shop/resolvers.py
@@ -33,7 +33,7 @@ def resolve_available_shipping_methods(info, channel_slug: str, address):
 
 def resolve_countries(**kwargs):
     countries_filter = kwargs.get("filter", {})
-    in_shipping_zones = countries_filter.get("in_shipping_zones")
+    attached_to_shipping_zones = countries_filter.get("attached_to_shipping_zones")
     language_code = kwargs.get("language_code")
     taxes = {vat.country_code: vat for vat in VAT.objects.all()}
     # DEPRECATED: translation.override will be dropped in Saleor 4.0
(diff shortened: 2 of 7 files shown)


patrys:
Task linked: [SALEOR-5075 Add query to get countries not selected in any shipping zone](https://app.clickup.com/t/2549495/SALEOR-5075)

saleor/graphql/shop/types.py
```diff
@@ -128,6 +129,10 @@ class Shop(graphene.ObjectType):
                 "A language code to return the translation for."
             ),
         ),
+        in_shipping_zones=graphene.Boolean(
+            description="Only countries that have shipping zones assigned",
```
maarcingebala, line 133: I think we can document all cases as each has a different result, for example: "When `true`, return X; when `false`, return Y; when the argument is not provided (`null`), return Z."

saleor/graphql/shop/types.py
```diff
@@ -222,16 +227,16 @@ def resolve_available_shipping_methods(_, info, channel, address=None):
...
         taxes = {vat.country_code: vat for vat in VAT.objects.all()}
-
         # DEPRECATED: translation.override will be dropped in Saleor 4.0
         with translation.override(language_code):
             return [
                 CountryDisplay(
                     code=country[0], country=country[1], vat=taxes.get(country[0])
                 )
                 for country in countries
+                if country[0] in get_countries_codes_list(in_shipping_zones)
```
maarcingebala, line 239: Can't we have something like this? Now we iterate over `django_countries.countries` twice- in the resolver and in the utility function while I think we can do it once.

```suggestion
                for country in get_countries_codes_list(in_shipping_zones)
```

saleor/graphql/shop/utils.py
```diff
@@ -0,0 +1,22 @@
+from typing import Optional
+
+from django_countries import countries
+
+from ...shipping.models import ShippingZone
+
+
+def get_countries_codes_list(in_shipping_zones: Optional[bool] = None):
+    """Return set of countries codes."""
```
maarcingebala, line 9: We should also describe the parameter (it could have the same description as in the schema).

maarcingebala (commented):
Now we only have a unit test for the utility function (which is good), but we should also have a test that uses the new parameter in some API call. In general, all API fields should be used in some tests; otherwise, we wouldn't know that we accidentally broke a field or resolver, while the utility function would keep passing.

saleor/graphql/shop/types.py
L3str4nge, line 133: "." 🚓

saleor/graphql/shop/types.py
```diff
@@ -222,16 +227,16 @@ def resolve_available_shipping_methods(_, info, channel, address=None):
         return resolve_available_shipping_methods(info, channel, address)
 
     @staticmethod
-    def resolve_countries(_, _info, language_code=None):
+    def resolve_countries(_, _info, language_code=None, in_shipping_zones=None):
```
L3str4nge, line 230: Shouldn't be `in_shupping_zones=False` by default?

saleor/graphql/shop/utils.py
```diff
@@ -0,0 +1,22 @@
+from typing import Optional
+
+from django_countries import countries
+
+from ...shipping.models import ShippingZone
+
+
+def get_countries_codes_list(in_shipping_zones: Optional[bool] = None):
```
L3str4nge, line 8: I am not sure `Optional[bool]` is good approach. I think if we want to pass `bool` then it should be True/False by default.

saleor/graphql/shop/types.py
```diff
@@ -128,6 +129,10 @@ class Shop(graphene.ObjectType):
                 "A language code to return the translation for."
             ),
         ),
+        in_shipping_zones=graphene.Boolean(
```
korycins, line 132: Isn't it something that could be placed under `filter` field? At the first moment I wasn't sure what this field exactly does.

saleor/graphql/shop/types.py
jakubkuc: I added better description to this field. It can be None - in this case it returns all of the countries and that is what we want by default.

saleor/graphql/shop/types.py
jakubkuc: I added better description. I hope it will be clear now

saleor/graphql/shop/utils.py
jakubkuc: I added better description. We want to be able to not pass this argument at all (as we were up till now) - which will cause all countries codes to be returned.

saleor/graphql/shop/tests/test_shop_utils.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def test_get_countries_codes_list(shipping_zones):
+    # given
+    fixture_countries_code_set = {zone.countries[0].code for zone in shipping_zones}
+
+    # when
+    countries_list_all = get_countries_codes_list()
+    countries_list_false = get_countries_codes_list(False)
+    countries_list_true = get_countries_codes_list(in_shipping_zones=True)
```
IKarbowiak, line 11: There should be 3 different tests.

saleor/graphql/shop/types.py
jakubkuc: It will not work because translation.override works only when iteration is made on djangos `Countries` object. I can't see the way to do it with less iterations than 2 at the moment. I might miss something though.

saleor/graphql/shop/filters.py
```diff
@@ -0,0 +1,5 @@
+import graphene
+
+
+class CountryFilterInput(graphene.InputObjectType):
+    in_shipping_zones = graphene.Boolean()
```
maarcingebala, line 5: Missing description.

saleor/graphql/shop/filters.py
```diff
@@ -0,0 +1,10 @@
+import graphene
+
+
+class CountryFilterInput(graphene.InputObjectType):
+    in_shipping_zones = graphene.Boolean(
```
korycins, line 5: Wouldn't be better to have something like: `attached_to_shipping_zones`? `in_shipping_zones` for first look sugested me that I can provide IDs of shipping zones. 
Just want to know what do you think

saleor/graphql/shop/resolvers.py
```diff
@@ -25,6 +31,22 @@ def resolve_available_shipping_methods(info, channel_slug: str, address):
     ]
 
 
+def resolve_countries(**kwargs):
+    countries_filter = kwargs.get("filter", {})
+    in_shipping_zones = countries_filter.get("in_shipping_zones")
+    language_code = kwargs.get("language_code")
+    taxes = {vat.country_code: vat for vat in VAT.objects.all()}
+    # DEPRECATED: translation.override will be dropped in Saleor 4.0
```
korycins, line 39: Make sure that you created a task for this and attached it to Saleor4.0 list

saleor/graphql/shop/filters.py
jakubkuc: now that you mention it I can see that it might be confusing. I'll change it.

maarcingebala (approved):

saleor/graphql/shop/tests/test_shop.py
```diff
@@ -1334,3 +1334,55 @@ def test_get_shop_limit_info_returns_null_by_default(staff_api_client):
...
+COUNTRY_FILTER_QUERY = """
+    query($filter: CountryFilterInput!) {
+        shop {
+            countries(filter: $filter){
+            code
+            }
+        }
+    }
+
+"""
```
IKarbowiak, line 1348: I will suggest adding two additional tests for empty "filter" option and for `{"attachedToShippingZones": None}`.

IKarbowiak (approved):

fowczarek (approved):

jakubkuc:
> Now we only have a unit test for the utility function (which is good), but we should also have a test that uses the new parameter in some API call. In general, all API fields should be used in some tests; otherwise, we wouldn't know that we accidentally broke a field or resolver, while the utility function would keep passing.

Tests added. Will add some more as @IKarbowiak suggested

saleor/graphql/shop/tests/test_shop.py
jakubkuc: empty filter option is already covered in old tests, the 'None' option I will add

PR #8909, opened 2022-01-06
Added support for metadata in address

LeOndaz:
Added support for metadata in address, my use case was because I wanted to name addresses 

for instance: home address

Or should I add a name for it? metadata seems more generic


LeOndaz:
The problem is

```
TypeError: __init__() got an unexpected keyword argument 'private_metadata'
```
Because `Address.as_data()` returns a copy that gets passed `private_metadata`

1 solution is to exclude Metadata from being on the copy address (done), but I'm not sure if this is the right one

LeOndaz:
not sure why it's failing now

IslamMesha (approved):

korycins:
@LeOndaz 
The changes committed by you, extend DB models but what about API? How do you want to change metadata for address when you don't have any access to it?

Some thoughts in case if we want to update address's metadata from API
- Enabling metadata in the Address object requires assigning proper permissions - to protect private metadata. This can be problematic as an Address is shared between many different objects. (like Checkout.shipping_address, User.addresses, Order.shipping_address etc).
- The second issue is defining how we want to protect public metadata from unauthorized access. Currently Address ID is base64("Address:1") where 1 is a DB ID. This means that we need to figure out how to protect it, update metadata mutation can return an updated object, which means, that we could have unauthorized access to Address object.

LeOndaz:
@korycins yes I should have converted this to draft as I'm partially working on it, please have a look

IslamMesha (approved):

IslamMesha (approved):

LeOndaz:
@korycins can you review now?

LeOndaz:
@korycins @patrys @maarcingebala 
Could I have a review here?

saleor/graphql/meta/tests/test_meta_mutations.py
```diff
@@ -63,7 +64,7 @@ def execute_update_public_metadata_for_item(
         variables,
         permissions=[permissions] if permissions else None,
     )
-    response = get_graphql_content(response)
+    response = get_graphql_content(response, ignore_errors=True)
```
fowczarek, line 67: This change affects that `get_graphql_content` will not check permission in all meta tests.

saleor/graphql/meta/tests/test_meta_mutations.py
```diff
@@ -87,7 +88,7 @@ def execute_update_public_metadata_for_multiple_items(
...
```
fowczarek, line 91: This change affects that `get_graphql_content` will not check permission in all meta tests.

saleor/graphql/meta/tests/test_meta_mutations.py
```diff
@@ -934,7 +934,7 @@ def execute_clear_public_metadata_for_item(
...
```
fowczarek, line 937: This change affects that `get_graphql_content` will not check permission in all meta tests.

saleor/graphql/meta/tests/test_meta_mutations.py
```diff
@@ -1714,7 +1713,7 @@ def execute_update_private_metadata_for_item(
...
```
fowczarek, line 1716: This change affects that `get_graphql_content` will not check permission in all meta tests.

saleor/graphql/meta/tests/test_meta_mutations.py
```diff
@@ -2578,7 +2576,7 @@ def execute_clear_private_metadata_for_item(
...
```
fowczarek, line 2579: This change affects that `get_graphql_content` will not check permission in all meta tests.

saleor/account/models.py
```diff
@@ -56,7 +56,7 @@ def annotate_default(self, user):
         )
 
 
-class Address(models.Model):
+class Address(ModelWithMetadata):
```
fowczarek, line 59: We should add indexing for metadata.

fowczarek (commented):
We should add a new entry to `MODEL_TO_TYPE_MAP` in `saleor/saleor/graphql/meta/resolvers.py`.
Missing test for all address meta operations.

PR #8915, opened 2022-01-09
No need for .nvmrc, since we mention engine into packge.json

HazemKhaled:
We have different versions for node here, one into `.nvmrc` which version 10.x and another one into `package.json` version 14 to 15 ... I prefer to keep `package.json`

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


HazemKhaled:
Sure, what do you think about `lts/*` instead of specifying a specific version since we don't care much here about the version?

ref https://github.com/nvm-sh/nvm#nvmrc @maarcingebala

maarcingebala:
Sounds good, but the current LTS is 16.13.2, so we should also update versions of `node` and `npm` defined in `package.json`.

HazemKhaled:
You don't need both, `.nvmrc` file and `engine` in `package.json` ... just decide to use any of them instead of changing both every time

https://stackoverflow.com/questions/61403815/js-nvmrc-vs-package-json-engines

I've no problem to update both by the way

maarcingebala:
OK, so let's stick with one version defined in `package.json`, anyone actually needing to use any of the npm commands with this repo should also understand how to set the correct version of Node.

maarcingebala:
@HazemKhaled Could you rebase the PR against the `main` branch? We push all changes related to Saleor 3.1 to main, later during releases we move them to the `3.1` branch.

crcastle:
I just stumbled over this a bit. It's unclear if this project requires [Node v10.15.2](https://github.com/saleor/saleor/blob/main/.nvmrc#L1) or [Node >=16 and <17](https://github.com/saleor/saleor/blob/main/package.json#L5).

PR #8916, opened 2022-01-09
Allow filtering staff users and permission groups by ids

stnatic:
I want to merge this change because it's going to make it easier to fetch multiple objects by third party apps.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/account/filters.py
+++ b/saleor/graphql/account/filters.py
@@ -1,10 +1,12 @@
 import django_filters
 from django.db.models import Count, Exists, OuterRef, Q
+from graphene_django.filter import GlobalIDMultipleChoiceFilter
 
 from ...account.models import Address, User
 from ..core.filters import EnumFilter, MetadataFilterBase, ObjectTypeFilter
 from ..core.types.common import DateRangeInput, IntRangeInput
-from ..utils.filters import filter_range_field
+from ..utils.filters import filter_by_id, filter_range_field
+from . import types as account_types
 from .enums import StaffMemberStatus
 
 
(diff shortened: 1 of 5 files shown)


saleor/graphql/account/filters.py
```diff
@@ -81,12 +83,17 @@ class Meta:
...
 class PermissionGroupFilter(django_filters.FilterSet):
     search = django_filters.CharFilter(method=filter_search)
+    ids = GlobalIDMultipleChoiceFilter(method=filter_by_id(account_types.Group))
 
 
 class StaffUserFilter(django_filters.FilterSet):
     status = EnumFilter(input_class=StaffMemberStatus, method=filter_staff_status)
     search = django_filters.CharFilter(method=filter_user_search)
-
+    ids = GlobalIDMultipleChoiceFilter(
```
korycins, line 92: Maybe I am wrong but I would say that it doesn't allow you to return `staff users` but the all users. You don't validate anywhere that given ID is a `staff user`.

stnatic: It's a mean of filtering the existing list, rather than fetching the base queryset. `filter_user_search` is also shared between "customers" and "staffUsers" queries

jakubkuc (approved):

saleor/graphql/utils/filters.py
```diff
@@ -60,3 +60,15 @@ def filter_range_field(qs, field, value):
         lookup = {f"{field}__lte": lte}
         qs = qs.filter(**lookup)
     return qs
+
+
+def filter_by_id(object_type):
+    from saleor.graphql.utils import resolve_global_ids_to_primary_keys
```
maarcingebala, line 66: Relative import 🚓

IKarbowiak (approved):

saleor/graphql/utils/filters.py
```diff
@@ -60,3 +60,15 @@ def filter_range_field(qs, field, value):
         lookup = {f"{field}__lte": lte}
         qs = qs.filter(**lookup)
     return qs
+
+
+def filter_by_id(object_type):
```
fowczarek, line 65: Should we use this function in all `IDs` filters?

maarcingebala: @stnatic Can you please open a ticket on your list for this? I can merge it as it is, but since this is rather an easy thing to reuse, we should do it as soon as we can. Please share the ticket (or add me as a watcher) when it's added.

maarcingebala:
@stnatic Please port these changes to 3.1 as well.

PR #8964, opened 2022-01-18
Observability reporter POC

przlada:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/plugins/webhook/tasks.py
+++ b/saleor/plugins/webhook/tasks.py
@@ -19,7 +19,6 @@ from ...core import EventDeliveryStatus
 from ...core.models import EventDelivery, EventPayload
 from ...core.tracing import webhooks_opentracing_trace
 from ...payment import PaymentError
-from ...plugins.manager import get_plugins_manager
 from ...settings import WEBHOOK_SYNC_TIMEOUT, WEBHOOK_TIMEOUT
 from ...site.models import Site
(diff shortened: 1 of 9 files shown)


saleor/plugins/webhook/tasks.py
```diff
@@ -367,6 +374,7 @@ def send_webhook_request_sync(app_name, delivery):
         delivery_update(delivery, EventDeliveryStatus.FAILED)
         raise ValueError("Unknown webhook scheme: %r" % (parts.scheme,))
     delivery_update(delivery, response.status)
+    report_event_delivery_attempt(delivery.event_type, attempt)
```
NyanKiyoshi, line 377: Shouldn't we provide the response status code and contents?

przlada: Response headers and content is stored in the attempt, which is passed here. This response data is then used to report this webhook delivery attempt. That is an excellent point because we skipped storing response status codes in a delivery attempt. Keeping a status code can be helpful in the future, and maybe we should add it to the `EventDeliveryAttempt` data model @maarcingebala @jakubkuc

maarcingebala: Yep, I think it would be helpful to store it. We could use a new field in the attempt model for that.

PR #8970, opened 2022-01-18
Sync webhooks for filtering shipping methods (port to 3.1)

stnatic:
I want to merge this change because I want to port changes from #8399 to 3.1

Depends on #8737.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/shipping/utils.py
```diff
@@ -53,3 +54,17 @@ def convert_to_shipping_method_data(
...
     )
+
+
+def annotate_active_shipping_methods(
+    shipping_methods: List["ShippingMethodData"],
+    excluded_methods: List["ExcludedShippingMethod"],
+):
+    reason_map = {str(method.id): method.reason for method in excluded_methods}
+    for instance in shipping_methods:
+        instance.active = True
```
L3str4nge, line 65: I am confusing, method name tells me that we annotate object with something but actually we are changing `active` state either.

patrys:
Task linked: [SALEOR-5871 Port of changes for shipping filter methods from 3.0 to 3.1](https://app.clickup.com/t/2549495/SALEOR-5871)

saleor/graphql/order/utils.py
```diff
@@ -27,7 +31,35 @@ def validate_total_quantity(order: "Order", errors: T_ERRORS):
...
+def get_shipping_method_availability_error(
+    order: "Order",
+    method: Optional["ShippingMethodData"],
+    manager: "PluginsManager",
+):
+    """Validate whether shipping method is still available for the order."""
+    valid_methods_ids = (
+        {
+            method.id
+            for method in get_valid_shipping_methods_for_order(
```
korycins, line 43: Can we revert this part of code a little bit? 
There is no need to call `get_valid_shipping_methods_for_order` when `method` is None, right? Can we check this as a first thing?

saleor/graphql/order/utils.py
```diff
@@ -27,7 +31,35 @@ def validate_total_quantity(order: "Order", errors: T_ERRORS):
...
+        return ValidationError(
+            "Shipping method cannot be used with this order.",
+            code=OrderErrorCode.SHIPPING_METHOD_NOT_APPLICABLE.value,
+        )
+
+
+def validate_shipping_method(
+    order: "Order", errors: T_ERRORS, manager: "PluginsManager"
+):
     error = None
```
korycins, line 63: It is not needed.

saleor/plugins/webhook/shipping.py
```diff
@@ -0,0 +1,156 @@
...
+        excluded_methods.append(ExcludedShippingMethod(id=method_id, reason=reason))
+    return excluded_methods
+
+
+def get_excluded_shipping_methods_from_response(
+    response_data: dict,
+) -> List[dict]:
+    excluded_methods = []
+    for method_data in response_data.get("excluded_methods", []):
+        try:
```
korycins, line 130: Have you tested it with external shipping methods?

saleor/shipping/utils.py
korycins: Agree, looks like this function just update the status of all shipping methods based on the list of excluded shipping methods, recieved from webhook.
Can we change the name?

saleor/webhook/deprecated_event_types.py
```diff
@@ -164,64 +155,11 @@ class WebhookEventType:
...
         ),
+        (ORDER_FILTER_SHIPPING_METHODS, DISPLAY_LABELS[ORDER_FILTER_SHIPPING_METHODS]),
+        (
+            CHECKOUT_FILTER_SHIPPING_METHODS,
+            DISPLAY_LABELS[CHECKOUT_FILTER_SHIPPING_METHODS],
+        ),
         (TRANSLATION_CREATED, DISPLAY_LABELS[TRANSLATION_CREATED]),
         (TRANSLATION_UPDATED, DISPLAY_LABELS[TRANSLATION_UPDATED]),
     ]
-
```
korycins, line 170: Is it expeceted?

saleor/webhook/deprecated_event_types.py
```diff
@@ -67,6 +54,8 @@ class WebhookEventType:
     PAYMENT_PROCESS = "payment_process"
 
     SHIPPING_LIST_METHODS_FOR_CHECKOUT = "shipping_list_methods_for_checkout"
+    CHECKOUT_FILTER_SHIPPING_METHODS = "checkout_filter_shipping_methods"
```
fowczarek, line 57: Should we add it to `deprecated_event_types.py` we also adding it to `event_types.py`

saleor/webhook/payloads.py
```diff
@@ -995,3 +998,47 @@ def generate_translation_payload(
...
+        "name": method.name,
+        "maximum_order_weight": method.maximum_order_weight,
+        "minimum_order_weight": method.minimum_order_weight,
+        "maximum_delivery_days": method.maximum_delivery_days,
+        "minimum_delivery_days": method.minimum_delivery_days,
+    }
+    return payload
+
+
+def generate_excluded_shipping_methods_for_order_payload(
```
fowczarek, line 1017:
```suggestion
@traced_payload_generator
def generate_excluded_shipping_methods_for_order_payload(
```

saleor/webhook/payloads.py
```diff
@@ -995,3 +998,47 @@ def generate_translation_payload(
...
+        "order": order_data,
+        "shipping_methods": [
+            _generate_payload_for_shipping_method(shipping_method)
+            for shipping_method in available_shipping_methods
+        ],
+    }
+    return json.dumps(payload, cls=CustomJsonEncoder)
+
+
+def generate_excluded_shipping_methods_for_checkout_payload(
```
fowczarek, line 1032:
```suggestion
@traced_payload_generator
def generate_excluded_shipping_methods_for_checkout_payload(
```

saleor/webhook/deprecated_event_types.py
stnatic: It's on 3.0 so I think we need to keep it in deprecated event types even though we are adding it to 3.1 only now

saleor/webhook/deprecated_event_types.py
stnatic: I believe that I removed it since it's unused

saleor/webhook/payloads.py
```diff
@@ -996,3 +999,49 @@ def generate_translation_payload(
         translation_data.update(context)
 
     return json.dumps(translation_data)
+
+
+def _generate_payload_for_shipping_method(method: ShippingMethodData):
+    payload = {
+        "id": graphene.Node.to_global_id("ShippingMethod", method.id),
```
stnatic, line 1006: @korycins should we return just `method.id` if external?

saleor/shipping/utils.py
stnatic: I used the verb `initialize` instead in order to suggest that accessing this field makes no sense prior to initialisation

saleor/plugins/webhook/shipping.py
```diff
@@ -0,0 +1,161 @@
...
+            raw_id = method_data["id"]
+            typename, _id = graphene.Node.from_global_id(raw_id)
+            if typename == "app":
```
korycins, line 133: Shouldn't be `APP_ID_PREFIX` ?

korycins (approved):

saleor/plugins/webhook/shipping.py
stnatic: It's the same value. However, in some parts of the code it's referenced to as a string, e.g:
image

Should we refactor this everywhere? The standard object types are also always represented as strings

fowczarek (approved):

saleor/graphql/order/utils.py
```diff
@@ -27,8 +31,36 @@ def validate_total_quantity(order: "Order", errors: T_ERRORS):
         )
 
 
-def validate_shipping_method(order: "Order", errors: T_ERRORS):
-    error = None
+def get_shipping_method_availability_error(
+    order: "Order",
+    method: Optional["ShippingMethodData"],
+    manager: "PluginsManager",
+):
```
L3str4nge, line 38: It's really hard to read. How about this version:
```
is_valid = False
if method:
    valid_methods_ids = get_valid_methods_ids(order, manager)
    is_valid = method.id in valid_methods_ids
```

saleor/plugins/webhook/shipping.py
```diff
@@ -0,0 +1,161 @@
...
+        method_id = shipping_method_data.get("id")
+        method_name = shipping_method_data.get("name")
+        method_amount = shipping_method_data.get("amount")
+        method_currency = shipping_method_data.get("currency")
+        method_maximum_delivery_days = shipping_method_data.get("maximum_delivery_days")
+
+        shipping_methods.append(
+            ShippingMethodData(
+                id=to_shipping_app_id(app, method_id),
+                name=method_name,
```
L3str4nge, line 44: In case when we have an `App` with > 2 methods without 'id' (reading the code it's possible because of `.get()`) result from `to_shipping_app_id` is not unique. Consider use `enumerate` and return something unique in `get()` instead of `None`.

saleor/plugins/webhook/shipping.py
```diff
@@ -0,0 +1,161 @@
...
+    excluded_methods = []
+    # Gather responses from webhooks
+    for webhook in webhooks:
+        response_data = trigger_webhook_sync(
+            event_type,
+            payload,
+            webhook.app,
+            EXCLUDED_SHIPPING_REQUEST_TIMEOUT,
+        )
+        # TODO handle a case when we didn't receive a proper response
```
L3str4nge, line 75: Do we have a task for this TODO?

saleor/plugins/webhook/shipping.py
L3str4nge: I would change everywhere to APP_ID_PREFIX. I can imagine we can change this prefix.

saleor/plugins/webhook/shipping.py
```diff
@@ -0,0 +1,161 @@
...
+                method_id = raw_id
+            elif typename == "ShippingMethod":
+                method_id = _id
+            else:
+                raise ValueError(
+                    f"Invalid type received. Expected ShippingMethod, got {typename}"
+                )
+        except (KeyError, ValueError, TypeError) as e:
+            logger.warning(f"Malformed ShippingMethod id was provided: {e}")
+            continue
```
L3str4nge, line 143: Lol we are raising Value error to catch it 3 lines later to call `continue` :D

saleor/shipping/interface.py
```diff
@@ -39,3 +40,9 @@ def is_external(self) -> bool:
             return str_type == "app"
 
         return False
+
+    @property
+    def graphql_id(self):
+        if self.is_external:
+            return self.id
```
L3str4nge, line 47: So it's actually not `graphql` one?

stnatic: For externals it is:
```
def to_shipping_app_id(app: "App", shipping_method_id: str) -> "str":
    return base64.b64encode(
        str.encode(f"{APP_ID_PREFIX}:{app.pk}:{shipping_method_id}")
    ).decode("utf-8")
```

This piece of code is extracted from the original `resolve_id` method of the `ShippingMethod` object type hence the name

saleor/plugins/webhook/shipping.py
stnatic: That felt like the cleanest way to keep a single logger statement. Suggestions are welcome ;)

saleor/plugins/webhook/shipping.py
stnatic: I don't think it's a TODO anymore, as the section that made you laugh is already covering pretty much all possible exceptions

saleor/plugins/webhook/shipping.py
stnatic: That's not a new piece of code, but rather me moving it between files to avoid import spaghetti. I'd prefer not to change the implementation without a dedicated task.

saleor/plugins/webhook/shipping.py
L3str4nge: So I would remove this TODO

saleor/graphql/order/utils.py
stnatic: We still need to check the `active` flag. I'd keep this one as it is unless someone else complains about this. Extracting this to a function that's going to have only one occurrence doesn't seem to be worth it

saleor/plugins/webhook/shipping.py
L3str4nge: Why not just:
```
else:
    log.warning(f"Invalid type received. Expected ShippingMethod, got {typename}")
    continue
```

saleor/plugins/webhook/shipping.py
korycins: Agree, we should use `APP_ID_PREFIX` everywhere

saleor/graphql/order/utils.py
```diff
@@ -27,8 +31,36 @@ def validate_total_quantity(order: "Order", errors: T_ERRORS):
...
+            for m in get_valid_shipping_methods_for_order(
+                order,
+                order.channel.shipping_method_listings.all(),
+                manager,
+            )
+            if m.active
+        }
+        is_valid = method.id in valid_methods_ids
+    else:
+        is_valid = False
```
IKarbowiak, line 52:
```suggestion
    is_valid = False
    if method:
        valid_methods_ids = {
            m.id
            for m in get_valid_shipping_methods_for_order(
                order,
                order.channel.shipping_method_listings.all(),
                manager,
            )
            if m.active
        }
        is_valid = method.id in valid_methods_ids
```

saleor/graphql/shipping/types.py
```diff
@@ -345,17 +345,13 @@ class ShippingMethod(graphene.ObjectType):
     class Meta:
         interfaces = [relay.Node, ObjectWithMetadata]
         description = (
-            (
-                "Shipping methods that can be used as means of shipping"
-                "for orders and checkouts."
-            ),
+            "Shipping methods that can be used as means of shipping"
```
IKarbowiak, line 348: Lack of space:
```suggestion
            "Shipping methods that can be used as means of shipping "
```

IKarbowiak (approved):

jakubkuc (approved):

L3str4nge (approved):

PR #8978, opened 2022-01-20
Fix incorrect handling of unavailable products in checkout

IKarbowiak:
Raise a `ValidationError` in `CheckoutCreate` and `CheckoutPaymentCreate` when any of the checkout line variant is unavailable for purchase. In other words, any of the following is True:  
- `ProductChannelListing` does not exist 
- `ProductChannelListing`'s `available_for_purchase` value is  None or is in feature 
- `ProductVariantChannelListing` does not exist
- `ProductVariantChannelListing` does not have price data

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-5026 Incorrect handling of unavailable products in checkout](https://app.clickup.com/t/2549495/SALEOR-5026)

fowczarek (approved):

saleor/checkout/fetch.py
```diff
@@ -137,6 +154,17 @@ def fetch_checkout_lines(checkout: "Checkout") -> Iterable[CheckoutLineInfo]:
...
+
+    if validate_variants and unavailable_variant_pks:
+        not_available_variants_ids = {
+            graphene.Node.to_global_id("ProductVariant", pk)
+            for pk in unavailable_variant_pks
+        }
+        raise ValidationError(
+            "Unavailable variants.",
+            params={"variants": not_available_variants_ids},
+        )
```
maarcingebala, line 166: I would remove the flag argument and instead return two lists from this function (one with available and the other with unavailable lines). The mutations should handle raising the validation error.

saleor/checkout/fetch.py
```diff
@@ -147,35 +175,61 @@ def fetch_checkout_lines(checkout: "Checkout") -> Iterable[CheckoutLineInfo]:
...
+            if channel_listing.channel_id == channel_id:
+                product_channel_listing = channel_listing
+        product_channel_listing_mapping[product.id] = product_channel_listing
+    return product_channel_listing
+
+
+def _apply_voucher_on_product(
+    voucher: "Voucher", checkout: "Checkout", lines_info: Iterable[CheckoutLineInfo]
+):
+    from .utils import get_discounted_lines
```
L3str4nge, line 205: Is any reason for local import here? I mean circular import etc?

L3str4nge (approved):

saleor/checkout/fetch.py
IKarbowiak: Yes, circular import is the reason.

maarcingebala (approved):

fowczarek (approved):

saleor/checkout/fetch.py
```diff
@@ -99,32 +100,37 @@ def is_method_in_valid_methods(self, checkout_info: "CheckoutInfo") -> bool:
         )
 
 
-def fetch_checkout_lines(checkout: "Checkout") -> Iterable[CheckoutLineInfo]:
+def fetch_checkout_lines(
+    checkout: "Checkout",
+) -> Tuple[Iterable[CheckoutLineInfo], Iterable[int]]:
```
korycins, line 105: https://github.com/saleor/saleor-plugin-avatax-excise/search?q=fetch_checkout_lines

saleor/graphql/checkout/mutations.py
```diff
@@ -1278,8 +1293,30 @@ def perform_mutation(
                 raise e
 
             manager = info.context.plugins
-            lines = fetch_checkout_lines(checkout)
-            validate_variants_in_checkout_lines(lines)
+            lines, unavailable_variant_pks = fetch_checkout_lines(checkout)
```
korycins, line 1296: I see at least 4 places where we're using fetch_checkout_lines to check if variants are unavailable. So my question is - wouldn't be good to create two separate functions:
- fetch_checkout_info - function that fetches all data as it was done previously OR function with new input + changes in many files
- fetch+validation - this can be called in different places and will give us, a single point for handling a validation of checkout's lines.  Then we could call this function in all places where valid variants are important to correctly process a checkout

IKarbowiak: @maarcingebala @fowczarek

fowczarek: IMO we could merge this PR and add a task with your proposition idea.

maarcingebala: Agree with @fowczarek here.

korycins: @fowczarek @maarcingebala Okay.  But we need to keep in mind that we have now 4 places where this validation is made. Any change or extending it will require adding/changing it in 4 places.

PR #8847, opened 2021-12-20
Taxes by sync webhooks

korycins:
I want to merge this change because...

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #8987, opened 2022-01-20
Add tax codes handling by webhooks

mateuszgrzyb:
I want to add tax codes handling (fetching all tax codes from API, saving tax code for a product/product type, fetching previously saved tax code) to sync webhooks.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/plugins/webhook/tasks.py
```diff
@@ -71,7 +70,7 @@ def _get_webhooks_for_event(event_type, webhooks=None):
     if webhooks is None:
         webhooks = Webhook.objects.all()
 
-    webhooks = webhooks.filter(
+    webhooks = webhooks.order_by("app_id", "pk").filter(
```
stnatic, line 73: Just a comment here: we discussed with Mateusz that webhooks should be processed foremost in the order of app installation. Some webhooks (such as tax webhooks) return first relevant result rather than iterating over all potential plugins.

This means that, e.g. orders and checkouts could be processed by different plugins or that an app could "turn itself off" by replacing the existing webhook

saleor/core/taxes.py
```diff
@@ -82,3 +91,125 @@ class TaxData:
...
+
+
+WEBHOOK_TAX_CODES_CACHE_KEY = "webhook_tax_codes"
+
+DEFAULT_TAX_CODE = "O9999999"
+DEFAULT_TAX_DESCRIPTION = "Unmapped Other SKU - taxable default"
+
+
+def _get_current_tax_app() -> App:
+    """Return currently used tax app or None, if there aren't any."""
```
L3str4nge, line 103: If we can return `None` I think we should change the function signature to `-> Optional[App]`

saleor/core/tests/test_taxes.py
```diff
@@ -0,0 +1,270 @@
...
+    # then
+    assert expected_app == app
+
+
+def test_get_current_tax_app_no_app():
+    # when
+    app = _get_current_tax_app()
+
+    # then
+    assert app is None
```
L3str4nge, line 114: Like mention above it could be one liner.

saleor/webhook/tests/test_webhook_payloads.py
```diff
@@ -253,6 +253,8 @@ def test_order_lines_have_all_required_fields(order, order_line_with_one_allocat
     global_warehouse_id = graphene.Node.to_global_id(
         "Warehouse", allocation.stock.warehouse_id
     )
+    product = line.variant.product
```
L3str4nge, line 256: I think `variant` could be `None` so It will raise here or am I wrong?

saleor/plugins/webhook/tasks.py
```diff
@@ -112,7 +111,12 @@ def trigger_webhook_sync(event_type: str, data: str, app: "App"):
     return send_webhook_request_sync(app.name, delivery)
 
 
-def trigger_tax_webhook_sync(event_type: str, data: str):
+R = TypeVar("R")
```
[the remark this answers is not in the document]
mateuszgrzyb: Any response type. I wanted to keep PEP 8 naming convention for `TypeVar`.

saleor/core/taxes.py
mateuszgrzyb: Resolved in recent commit.

saleor/core/tests/test_taxes.py
```diff
@@ -0,0 +1,270 @@
...
+
+def test_get_current_tax_app(tax_app):
+    # given
+    expected_app = tax_app
+
+    # when
+    app = _get_current_tax_app()
+
+    # then
+    assert expected_app == app
```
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit.

saleor/core/tests/test_taxes.py
mateuszgrzyb: Resolved in recent commit.

saleor/plugins/webhook/utils.py
```diff
@@ -222,6 +222,23 @@ def parse_list_shipping_methods_response(
...
+    try:
+        return [
+            TaxType(
+                code=tax["code"],
+                description=tax["description"],
+            )
+            for tax in response_tax_data
+        ]
+    except KeyError:
+        return None
```
L3str4nge, line 239: I will be a badass here but it looks like hieroglyph :D Do we really need that list comprehensions in `try...except` ?  Consider classic old schoold for loop.

L3str4nge (approved):

IKarbowiak (approved):

saleor/core/taxes.py
```diff
@@ -82,3 +91,125 @@ class TaxData:
...
     lines: List[TaxLineData]
+
+
+WEBHOOK_TAX_CODES_CACHE_KEY = "webhook_tax_codes"
+
+DEFAULT_TAX_CODE = "O9999999"
+DEFAULT_TAX_DESCRIPTION = "Unmapped Other SKU - taxable default"
+
+
+def _get_current_tax_app() -> Optional[App]:
```
maarcingebala, line 102: We have `App.for_event_type` which is used to return apps for specific event types.

saleor/plugins/webhook/tasks.py
maarcingebala: Is it tested? I think we should provide a test for that as this is something that might be easily broken by accident.

saleor/core/taxes.py
```diff
@@ -82,3 +90,121 @@ class TaxData:
...
+    return (
+        App.objects.order_by("pk")
+        .for_event_type(WebhookEventSyncType.CHECKOUT_CALCULATE_TAXES)
+        .for_event_type(WebhookEventSyncType.ORDER_CALCULATE_TAXES)
+        .last()
+    )
+
+
+def _get_meta_code_key(app: App) -> str:
+    return f"{app.name}.code"
```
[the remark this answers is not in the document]
mateuszgrzyb: It's not really building tax code per app, but rather per API provider. So when installing two different apps from the same provider, Saleor will use those keys. Also removing and adding the same app will not break integration, because API provider does not change.

I think using `App.identifier` here instead of `App.name` is a better choice.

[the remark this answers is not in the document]
korycins: Right, sorry, I don't know why I've focused that `name` will behave in the same manner as `id` 😄

saleor/plugins/webhook/utils.py
```diff
@@ -222,6 +223,32 @@ def parse_list_shipping_methods_response(
...
+def parse_tax_codes(
+    response_tax_data: Any,
+) -> Optional[Dict[str, str]]:
+    if not isinstance(response_tax_data, list):
+        return None
+
+    tax_types = {}
+
+    for tax in response_tax_data:
+        print(tax)
```
korycins, line 235: 🚔 👮

saleor/plugins/webhook/utils.py
```diff
@@ -293,3 +320,27 @@ def delivery_update(delivery: "EventDelivery", status: str):
 def clear_successful_delivery(delivery: "EventDelivery"):
     if delivery.status == EventDeliveryStatus.SUCCESS:
         delivery.delete()
+
+
+WEBHOOK_TAX_CODES_CACHE_KEY = "webhook_tax_codes"
+DEFAULT_TAX_CODE = "SA0000"
```
korycins, line 326: Wouldn't be better to convert this to something like `STANDARD` or `UNMAPPED`?

korycins (approved):

saleor/plugins/webhook/utils.py
```diff
@@ -293,3 +320,27 @@ def delivery_update(delivery: "EventDelivery", status: str):
...
+DEFAULT_TAX_DESCRIPTION = "Unmapped Product/Product Type"
+
+
+def get_current_tax_app() -> Optional[App]:
+    """Return currently used tax app or None, if there aren't any."""
```
fowczarek, line 331: What in situations that we have 3 different apps for taxes? 
The last app can be not active.

fowczarek (commented):
We should add benchmarks for checkout or order with taxes by webhook.
We should extend webhook docs.

saleor/plugins/webhook/utils.py
korycins: Assigning a tax codes doesn't support more than one enabled tax App. `is_active` is checked in for_event_type.

korycins (approved):

stnatic:
@fowczarek I think it makes sense to benchmark only when the final PR is assembled (#8847). Updating new checkout line fields is going to affect the number of queries and #8733 might affect that

PR #8990, opened 2022-01-21
Add unit test exposing potential duplicate webhooks

stnatic:
While reviewing another pull request, I've noticed this anti pattern - filtering by a reverse foreign key without `distinct`
```
webhooks = webhooks.filter(
        is_active=True,
        app__is_active=True,
        events__event_type__in=[event_type, WebhookEventAsyncType.ANY],
        **permissions,
)
```

I'd recommend to either add distinct or to add a comment why is it not needed. It doesn't seem like the `createWebhook` mutation is protected against it.

The unit test shows en example scenario where `_get_webhooks_for_event` could return a single webhook twice, since it was configured to be triggered by both `event_type` and `ANY`.

```
>       assert len(webhooks) == 1
E       assert 2 == 1
E        +  where 2 = len(<QuerySet [<Webhook: Simple webhook>, <Webhook: Simple webhook>]>)
```

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/plugins/webhook/tests/test_webhook.py
+++ b/saleor/plugins/webhook/tests/test_webhook.py
@@ -42,6 +42,7 @@ from ....webhook.payloads import (
 from ...manager import get_plugins_manager
 from ...webhook.tasks import (
     WebhookResponse,
+    _get_webhooks_for_event,
     send_webhook_request_async,
     trigger_webhooks_async,
 )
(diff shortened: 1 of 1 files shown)


L3str4nge:
@stnatic Your test is failing. Could you take a look?

stnatic:
@L3str4nge That's the point - it shows how a single webhook is returned twice by that helper function. The code can be easily adjusted by adding `distinct` but first I want to make sure that it wasn't omitted on purpose (performance reasons? although I think in this case it's negligible)

L3str4nge:
Why we can't just fix that? :D

maarcingebala:
I think it should be rewritten with subqueries - a separate query to get active apps, then another to get the right event types, and the third one combining it all together.

stnatic:
I would suggest to consider this together with https://github.com/saleor/saleor/issues/9138 as this is also the piece of code that makes ANY_EVENT apply to sync webhooks

PR #9006, opened 2022-01-26
Workaround for incoming partial payments from adyen

korycins:
This PR is one big workaround for handling Adyen partial payments in Saleor3.0.
For some Adyen's merchant account (especially the old ones), Adyen moves a customer to a redirect page, where for payments like GIVEX, Adyen can handle partial payments. We already support a notification related to this: `order_opened`, `order_closed` but we didn't track anything about partial payments that have been made. From Saleor perspective we store only info about a single payment, where on Adyen side it is two(or more) payments.

This PR introduces a logic to:
-  store the info about partial payments
- send partial payments objects via webhooks.
- handling a refund of partial payment in a case when we are not able to create an order
- place a note in order history about partial refunds

This code will be rewritten in 3.1, as we are going to introduce their full support for partial payments. 
Also, I will handle the task of rewriting this (for 3.1) in the meantime.

Also, the workaround was the only solution for fixing this without porting a whole partial payment feature to 3.0

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/order/models.py
```diff
@@ -277,7 +277,11 @@ def __str__(self):
         return "#%d" % (self.id,)
 
     def get_last_payment(self):
-        return max(self.payments.all(), default=None, key=attrgetter("pk"))
+        # Skipping a partial payment is a temporary workaround for storing a basic data
+        # about partial payment from Adyen plugin. This is something that will removed
+        # in 3.1 by introducing a partial payments feature.
+        payments = [payment for payment in self.payments.all() if not payment.partial]
```
L3str4nge, line 283: Would be a good option to filter active payments only or do we need all types? If so maybe consider to `filter`?

saleor/payment/gateways/adyen/utils/common.py
```diff
@@ -32,6 +32,16 @@
 AUTH_STATUS = "authorised"
 
 
+def initialize_adyen_client(config: GatewayConfig) -> Adyen.Adyen:
+    api_key = config.connection_params["api_key"]
+
+    live_endpoint = config.connection_params["live"] or None
```
L3str4nge, line 38: Could be just `.get("live")` but it's just cosmetic.

saleor/payment/gateways/adyen/webhooks.py
```diff
@@ -616,10 +623,113 @@ def handle_order_opened(notification: Dict[str, Any], gateway_config: GatewayCon
...
+
+
+def get_adyen_partial_payment_keys(additiona_data):
+    """Get keys that contains adyen partial data.
+
+    The data received from Adyen has strange structure. In middle of dict key, there is
+    a int number pointing to the given payment method. So we need to convert it and
+    match somehow a paymentMethod, paymentAmount and payment psp reference
+    """
+    for i in range(1, 99):
```
L3str4nge, line 707: I'm wondering about this range here and it's values ?

saleor/payment/gateways/adyen/webhooks.py
```diff
@@ -616,10 +623,113 @@ def handle_order_opened(notification: Dict[str, Any], gateway_config: GatewayCon
...
+        if payment_method_key not in additiona_data:
+            return
+        yield (payment_method_key, payment_amount_key, psp_reference_key)
+
+
+def refund_partial_payments(payments, config):
+    for payment in payments:
+        adyen_client = initialize_adyen_client(config)
+        merchant_account = config.connection_params["merchant_account"]
+        logger.info("Calling refund for partial payment: %s", payment.psp_reference)
```
L3str4nge, line 722: I see logzzzz <3

saleor/payment/migrations/0030_payment_partial.py
```diff
@@ -0,0 +1,18 @@
...
+        ("payment", "0029_alter_payment_options"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="payment",
+            name="partial",
+            field=models.BooleanField(default=False),
+        ),
+    ]
```
L3str4nge, line 18: I believe that we have a huge number of payments on the client's production so I think this migration could produce some downtimes.

L3str4nge (approved):

IKarbowiak (approved):

jakubkuc (approved):

fowczarek (approved):

maarcingebala (approved):

saleor/order/models.py
korycins: I was thinking about this and also I tried. The issue is action like void, refund etc mark payment as inactive. With the filter we will not return a payment data.

saleor/payment/gateways/adyen/webhooks.py
korycins: 🤦  right...

saleor/order/models.py
stnatic: Another reason to use `all` instead of filter is that `filter` will not respect prefetches

saleor/payment/migrations/0030_payment_partial.py
korycins: @cmiacz  What do you think?

patrys:
Task linked: [SALEOR-5604 Split payments using Givex do not show correct information](https://app.clickup.com/t/2549495/SALEOR-5604)

saleor/payment/migrations/0030_payment_partial.py
cmiacz: In PostgreSQL11 adding a field with constant default value is fast no matter how big is the table.
See the docs: https://www.postgresql.org/docs/11/ddl-alter.html#DDL-ALTER-ADDING-A-COLUMN

tomaszszymanski129 (approved):

eCzerniak (approved):

PR #9024, opened 2022-01-28
Fix breaking calculate_order_shipping avatax method

IKarbowiak:
Fix breaking `calculate_order_shipping` avatax method. The `price` field is no longer part of the `ShippingMethod` model.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/mutations/draft_orders.py
+++ b/saleor/graphql/order/mutations/draft_orders.py
@@ -314,6 +314,14 @@ class DraftOrderCreate(ModelMutation, I18nMixin):
                 "Unable to calculate taxes - %s" % str(tax_error),
                 code=OrderErrorCode.TAX_ERROR.value,
             )
+        except Exception as e:
+            raise ValidationError(
+                {
+                    "shipping_method": ValidationError(
+                        str(e), code=OrderErrorCode.SHIPPING_METHOD_NOT_APPLICABLE.value
+                    )
+                }
+            )
 
         if new_instance:
             transaction.on_commit(
--- a/saleor/graphql/order/mutations/draft_orders.py
+++ b/saleor/graphql/order/mutations/draft_orders.py
@@ -314,14 +314,6 @@ class DraftOrderCreate(ModelMutation, I18nMixin):
                 "Unable to calculate taxes - %s" % str(tax_error),
                 code=OrderErrorCode.TAX_ERROR.value,
             )
-        except Exception as e:
-            raise ValidationError(
-                {
-                    "shipping_method": ValidationError(
-                        str(e), code=OrderErrorCode.SHIPPING_METHOD_NOT_APPLICABLE.value
-                    )
-                }
-            )
 
         if new_instance:
             transaction.on_commit(
--- a/saleor/plugins/avatax/plugin.py
+++ b/saleor/plugins/avatax/plugin.py
@@ -636,10 +636,14 @@ class AvataxPlugin(BasePlugin):
                     gross = Money(amount=net + tax, currency=currency)
                     net = Money(amount=net, currency=currency)
                 return TaxedMoney(net=net, gross=gross)
+
+        # Ignore typing checks because it is checked in _validate_order
+        price = order.shipping_method.channel_listings.get(  # type: ignore
+            channel_id=order.channel_id
+        ).price
         return TaxedMoney(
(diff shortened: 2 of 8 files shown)


patrys:
Task linked: [SALEOR-5715 Avatax calculate_order_shipping hook breaks under some conditions](https://app.clickup.com/t/2549495/SALEOR-5715)

L3str4nge (approved):

jakubkuc (approved):

fowczarek (approved):

tomaszszymanski129 (approved):

saleor/plugins/avatax/plugin.py
```diff
@@ -636,10 +636,14 @@ def calculate_order_shipping(
                     gross = Money(amount=net + tax, currency=currency)
                     net = Money(amount=net, currency=currency)
                 return TaxedMoney(net=net, gross=gross)
+
+        # Ignore typing checks because it is checked in _validate_order
+        price = order.shipping_method.channel_listings.get(  # type: ignore
```
maarcingebala, line 641: Are we sure that this channel always listing exists in the database at this point? I'm thinking about a race condition when it is deleted while this function is processing.

IKarbowiak: The same is in [manager.py](https://github.com/saleor/saleor/blob/3.0/saleor/plugins/manager.py#L270), so maybe we should move validation there.

saleor/graphql/order/mutations/draft_orders.py
```diff
@@ -314,6 +314,14 @@ def save(cls, info, instance, cleaned_input):
                 "Unable to calculate taxes - %s" % str(tax_error),
                 code=OrderErrorCode.TAX_ERROR.value,
             )
+        except Exception as e:
+            raise ValidationError(
+                {
+                    "shipping_method": ValidationError(
```
korycins, line 320: Is it safe enough that we assume, that `DraftOrderCreate` mutation, when raises a generic exception, means - inccorect shipping_method? Assuming that we will change anything, or break something inside this `try,except`, then DraftOrderCreate will raise incorrect exception.

saleor/plugins/avatax/plugin.py
```diff
@@ -636,10 +636,17 @@ def calculate_order_shipping(
                     gross = Money(amount=net + tax, currency=currency)
                     net = Money(amount=net, currency=currency)
                 return TaxedMoney(net=net, gross=gross)
+
+        # Ignore typing checks because it is checked in _validate_order
```
korycins, line 640: Do we really need this part? Maybe we could simply extend `saleor.plugins.avatax._validate_order` to return False when `channel_listing ` is not available? - As we can say: 'we don't have proper shipping method assigned. This order doesn't have enough info to fetch taxes for it`
Also, do we need to raise an exception? I thought that in case of removing channel_listing for a given method, Saleor will block a possibility to finalize an order, no need to double-check it inside a Tax method, right?

saleor/plugins/manager.py
```diff
@@ -267,9 +267,12 @@ def calculate_checkout_shipping(
     def calculate_order_shipping(self, order: "Order") -> TaxedMoney:
         if not order.shipping_method:
             return zero_taxed_money(order.currency)
-        shipping_price = order.shipping_method.channel_listings.get(
+        channel_listing = order.shipping_method.channel_listings.filter(
             channel_id=order.channel_id
-        ).price
+        ).first()
+        if not channel_listing:
```
korycins, line 273: Do we also want to raise an exception here? (If it is really must-have, I don't think that generic `Exception` is the best way for this). 
IMHO, I would recommend returning in that case `zero_taxed_money` as it is done as above. We can just simply assume that not having `channel_listing` is equal to not having shipping_method assigned to order. 
I believe that we are not able to finalize an order in case of missing `channel_listing` which I think it is enough  for forcing user to update a shipping.

saleor/plugins/avatax/plugin.py
IKarbowiak: But in this situation, the `previous_value` will be returned and an exception will be not raised, I am not sure if it's the correct solution.

maarcingebala (approved):

fowczarek (approved):

PR #9081, opened 2022-02-04
Query cost in graphql spans tracing

tomaszszymanski129:
I want to merge this change because it adds a new tag to `graphql` tracing span - `graphql.query_cost`.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/views.py
+++ b/saleor/graphql/views.py
@@ -279,17 +279,17 @@ class GraphQLView(View):
                 except GraphQLError as e:
                     return ExecutionResult(errors=[e], invalid=True)
 
-                if settings.GRAPHQL_QUERY_MAX_COMPLEXITY:
-                    query_cost, cost_errors = validate_query_cost(
-                        schema,
-                        document,
-                        variables,
-                        COST_MAP,
-                        settings.GRAPHQL_QUERY_MAX_COMPLEXITY,
-                    )
-                    if cost_errors:
-                        result = ExecutionResult(errors=cost_errors, invalid=True)
-                        return set_query_cost_on_result(result, query_cost)
+                query_cost, cost_errors = validate_query_cost(
+                    schema,
+                    document,
+                    variables,
+                    COST_MAP,
+                    settings.GRAPHQL_QUERY_MAX_COMPLEXITY,
+                )
+                span.set_tag("graphql.query_cost", query_cost)
+                if settings.GRAPHQL_QUERY_MAX_COMPLEXITY and cost_errors:
+                    result = ExecutionResult(errors=cost_errors, invalid=True)
+                    return set_query_cost_on_result(result, query_cost)
 
             extra_options: Dict[str, Optional[Any]] = {}
 


patrys:
Task linked: [SALEOR-5060 Add Query Cost to to GraphQL span](https://app.clickup.com/t/2549495/SALEOR-5060)

IKarbowiak (approved):

saleor/graphql/views.py
```diff
@@ -279,17 +279,17 @@ def execute_graphql_request(self, request: HttpRequest, data: dict):
...
-                        schema,
-                        document,
-                        variables,
-                        COST_MAP,
-                        settings.GRAPHQL_QUERY_MAX_COMPLEXITY,
-                    )
-                    if cost_errors:
-                        result = ExecutionResult(errors=cost_errors, invalid=True)
-                        return set_query_cost_on_result(result, query_cost)
+                query_cost, cost_errors = validate_query_cost(
```
korycins, line 282: In settings.py, the `GRAPHQL_QUERY_MAX_COMPLEXITY` has an additional comment: `# Set GRAPHQL_QUERY_MAX_COMPLEXITY=0 in env to disable (not recommended)`. It looks like this PR changes this, am I right? How it will behave when this value will be set to 0?

tomaszszymanski129: @korycins only change is that now we always calculate the query cost and set it on tag. Error logic remains the same, we "raise" it when `GRAPHQL_QUERY_MAX_COMPLEXITY` is > 0 and complexity was over that value.

L3str4nge (approved):

fowczarek (approved):

saleor/graphql/views.py
korycins: But now, you can't disable a max complexity from settings, right? You can't use `0` as it was previously. So it means that the comment from settings is not true now, right?

korycins: Also, what will happen when this value will be set `0`? Assuming that you already set it up this to `0` in Saleor3.0 - You wanted to turn it off as you didn't need it. But from 3.1, we are changing the behavior, right?
3.0 -> == `0` turned off
3.1 -> == `0` complexity equl to `0`, which means (i beleive at least) that any query will raise an exception

tomaszszymanski129: This feature is not present in 3.0 but I think I got the point. If someone sets that to 0, it no longer disables entire thing (calculation stays on, error raising disabled). We should be good to change it to set the tag only when the setting is > 0 - if you don't want the complexity caclculation at all, you most likely don't want the traces either. This way nothing gets changed - we only trace the complexity if it's actually being calculated.
CC @fowczarek

korycins: My bad. It looks like I've changed the codebase for 3.0 incorrectly, as I was sure that this flag exists in 3.0 😄 
Sorry!

fowczarek: IMO we should have complexity always. For sure on start, because we need to tune complexity calculation. Maybe after some time, we could turn off it.

PR #9102, opened 2022-02-07
Fix invalid available shipping methods calculation for checkout

IKarbowiak:
**Issue description**

Given:
- checkout with shipping method price based set

Situation:
- shipping voucher is applied, the voucher discount reduces the price, so the minimal price condition for the current shipping method is not fulfilled

Previous behavior:
- the current shipping method was removed

Current solution
- the current method is not removed, as the voucher is applied on shipping, which conditions were already fulfilled;
(i.e. somebody applied free shipping voucher - we should ensure that the shipping is not removed)
- available shipping methods are updated when discount concerns entire orders or specific products

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -429,8 +429,7 @@ def recalculate_checkout_discount(
     applicable.
     """
     checkout = checkout_info.checkout
-    voucher = get_voucher_for_checkout_info(checkout_info)
-    if voucher is not None:
+    if voucher := checkout_info.voucher:
         address = checkout_info.shipping_address or checkout_info.billing_address
         try:
             discount = get_voucher_discount_for_checkout(
@@ -438,6 +437,7 @@ def recalculate_checkout_discount(
             )
         except NotApplicable:
             remove_voucher_from_checkout(checkout)
+            checkout_info.voucher = None
         else:
             subtotal = calculations.checkout_subtotal(
                 manager=manager,
@@ -552,6 +552,7 @@ def add_voucher_to_checkout(
             "last_change",
         ]
     )
+    checkout_info.voucher = voucher
 
 
 def remove_promo_code_from_checkout(checkout_info: "CheckoutInfo", promo_code: str):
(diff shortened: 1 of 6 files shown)


saleor/checkout/complete_checkout.py
```diff
@@ -58,7 +57,11 @@ def _process_voucher_data_for_order(checkout_info: "CheckoutInfo") -> dict:
     :raises NotApplicable: When the voucher is not applicable in the current checkout.
     """
     checkout = checkout_info.checkout
-    voucher = get_voucher_for_checkout_info(checkout_info, with_lock=True)
+    voucher = (
+        Voucher.objects.all().select_for_update().get(code=checkout_info.voucher.code)
```
L3str4nge, line 61: Do we need `.all()` here?

L3str4nge (approved):

saleor/checkout/utils.py
```diff
@@ -429,15 +429,20 @@ def recalculate_checkout_discount(
     applicable.
     """
     checkout = checkout_info.checkout
-    voucher = get_voucher_for_checkout_info(checkout_info)
-    if voucher is not None:
+    if checkout_info.voucher is not None:
+        voucher = (
+            Voucher.objects.all()
+            .select_for_update()
```
korycins, line 435: I don't think that `select_for_update` is needed here. It will lock the voucher for whole checkout processing.

saleor/checkout/complete_checkout.py
korycins: Also, not sure why we don't use the previous function. It has much more logic than one simple query.

korycins: for example, see how lock is working there. It doesn't call select_for_update always. It lock row only for specific case.

saleor/checkout/utils.py
korycins: Also, do we need to re-fetch the voucher here? We already have it in checkout_info

IKarbowiak: It was done like that before with a voucher so I reassigned it. If we do not want the voucher to be selected for update `checkout_info.voucher` is enought.

korycins (approved):

PR #9104, opened 2022-02-07
Observability reporter

przlada:
I want to merge this change because it adds an observability reporter to Saleor. 

- Activate Observability reporter by setting `OBSERVABILITY_BROKER_URL`. It can be an URL to any broker supported by the `kombu` messaging library. 
- The reporter gathers two observability event types: API calls and event delivery attempts. It collects API calls only from apps by default, but the `OBSERVABILITY_REPORT_ALL_API_CALLS` setting can change this (useful for debugging).
- Observability events are buffered, and the buffer size is configurable using the `OBSERVABILITY_BUFFER_SIZE_LIMIT` setting. Events put to the full buffer are skipped. In addition, the payload of the observability event is truncated during creation to the maximum size specified by `OBSERVABILITY_MAX_PAYLOAD_SIZE`. 
- A Celery background task periodically empties the buffer. `OBSERVABILITY_REPORT_PERIOD` can regulate the frequency of emptying the buffer (every 20s by default). Reporter sends observability events to all webhooks, listening for `OBSERVABILITY` event types. Events are sent in batches witch size is configured by `OBSERVABILITY_BUFFER_BATCH_SIZE`.  

During the observability event payload creation, all data coming from the Graphql backend is anonymized. Right now, the presence of any sensitive field in the query results in anonymizing the whole response. It will be improved in future PR

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

https://app.clickup.com/t/2549495/CLOUD-1195


saleor/plugins/webhook/utils.py
```diff
@@ -222,3 +227,20 @@ def delivery_update(delivery: "EventDelivery", status: str):
...
     if delivery.status == EventDeliveryStatus.SUCCESS:
         delivery.delete()
+
+
+def report_event_delivery_attempt(
+    event_type: str,
+    attempt: "EventDeliveryAttempt",
+    next_retry: Optional[datetime] = None,
+):
+    if event_type not in WebhookEventAsyncType.REPORTER_EVENTS:
```
korycins, line 237: [question] Wouldn't be more readable to explicitly specify the events here, instead of creating a specific variable in EventTypes to use it here?

saleor/webhook/payloads.py
```diff
@@ -996,3 +999,83 @@ def generate_translation_payload(
...
+        return text
+    return text.encode(encoding)[:limit].decode(encoding, "ignore")
+
+
+def generate_api_call_payload(request, response):
+    content_length = int(request.headers.get("Content-Length", 0))
+    response_body = response.content.decode(response.charset)
+    request_body = None
+    if request.POST:
+        request_body = json.dumps(dict(request.POST))
```
korycins, line 1020: Can' we use here `response.json()`?

saleor/webhook/payloads.py
```diff
@@ -996,3 +999,83 @@ def generate_translation_payload(
...
+        try:
+            request_body = request.body.decode("utf-8")
+        except ValueError:
+            pass
+    request_id = None
+    if hasattr(request, "request_uuid") and request.request_uuid:
+        request_id = str(request.request_uuid)
+
+    payload = {
+        "request_id": request_id or str(uuid.uuid4()),
```
korycins, line 1031: What exactly is achieved by adding `request_id`?  I mean, assuming that someone will come to you as an admin, and will send you an ID of a failed request. We don't track it anywhere on our side.

saleor/webhook/payloads.py
```diff
@@ -996,3 +999,83 @@ def generate_translation_payload(
...
+        "request_time": request.request_time.timestamp(),
+        "request_headers": dict(request.headers),
+        "request_body": truncate_str_to_byte_limit(request_body),
+        "request_content_length": content_length,
+        "response_status_code": response.status_code,
+        "response_reason_phrase": response.reason_phrase,
+        "response_headers": dict(response.headers),
+        "response_content": truncate_str_to_byte_limit(response_body),
+    }
+    if hasattr(request, "app") and request.app:
```
korycins, line 1041: `getattr(request, "app")`?

saleor/plugins/webhook/utils.py
przlada: That was my initial implementation, and then I mowed it to EventTypes. So I fall back on your opinion and will move it back here 😄

saleor/webhook/payloads.py
przlada: The idea was to be universal and get data to report even if returned content isn't proper JSON. We later send and process it as a string, so there is no need to JSON parse it.

saleor/webhook/payloads.py
przlada: Because of transport channels, the report receiver can receive the same report multiple times. `request_id` is added to identify them on the receiver side and skip duplicates.

cmiacz (approved):

saleor/graphql/views.py
```diff
@@ -197,30 +200,33 @@ def handle_query(self, request: HttpRequest) -> JsonResponse:
...
+            execution_result = self.execute_graphql_request(request, data)
+            status_code = 200
+            if execution_result:
+                response = {}
+                if execution_result.errors:
+                    response["errors"] = [
+                        self.format_error(e) for e in execution_result.errors
+                    ]
+                if execution_result.invalid:
+                    status_code = 400
```
korycins, line 221: Do you really need a status_code here? Looks like you have  200 when everything is ok, and 400 when something is wrong. Doesn't `response["errors"]` have enough information that something went wrong?
Also, do we really want to use similar http status_code  here? 200/400 could suggest this, but this could be a simple `invalid:true/false` value, right?

saleor/plugins/webhook/tasks.py
```diff
@@ -474,11 +484,70 @@ def send_webhook_request_sync(
...
+            )
+        if response.status == EventDeliveryStatus.SUCCESS:
+            logger.debug(
+                "Observability successful delivered %s of type %r to %r.",
+                len(events),
+                event_type,
+                webhook.target_url,
+            )
+        else:
+            task_logger.info(
```
korycins, line 530: Is `info` enough? I would consider `warning`

saleor/plugins/webhook/tasks.py
```diff
@@ -474,11 +484,70 @@ def send_webhook_request_sync(
...
+            for event in events:
+                response = send_webhook_using_scheme_method(
+                    webhook.target_url,
+                    domain,
+                    webhook.secret_key,
+                    event_type,
+                    json.dumps(event),
+                )
+                if response.status == EventDeliveryStatus.FAILED:
+                    break
```
korycins, line 513: Don't we need any log here?

saleor/plugins/webhook/tasks.py
```diff
@@ -474,11 +484,70 @@ def send_webhook_request_sync(
...
     clear_successful_delivery(delivery)
 
     return response_data if response.status == EventDeliveryStatus.SUCCESS else None
 
 
+@app.task
+def observability_send_events():
+    events = buffer.get_events()
+    if not events:
+        return 0
```
korycins, line 497: It is celery task. Is it expected that you return some value here and below? `len(events)`?

saleor/webhook/event_types.py
```diff
@@ -190,6 +195,7 @@ class WebhookEventAsyncType:
         PAGE_DELETED: PagePermissions.MANAGE_PAGES,
         TRANSLATION_CREATED: SitePermissions.MANAGE_TRANSLATIONS,
         TRANSLATION_UPDATED: SitePermissions.MANAGE_TRANSLATIONS,
+        OBSERVABILITY: AppPermission.MANAGE_APPS,
```
korycins, line 198: Are you sure of this? Don't you think that better way, would be having separate permission for observability? I am not conviced that app that, gathers a whole communication logs should use `manage_apps`

saleor/webhook/observability_reporter/__init__.py
```diff
@@ -0,0 +1,139 @@
...
+        ):
+            return False
+        tracer = opentracing.global_tracer()
+        with tracer.start_active_span("observability_report_api_call") as scope:
+            scope.span.set_tag(opentracing.tags.COMPONENT, "observability")
+            if self.response is None:
+                logger.info("[Observability] HttpResponse not provided")
+                return False
+            event_type = WebhookEventAsyncType.OBSERVABILITY
+            if webhooks_for_event_exists(event_type):
```
korycins, line 60: Wouldn't be good to create a cache or something that could reduce the amount of DB queries? Like, this function will add 2-3 queries to each request.

saleor/webhook/observability_reporter/payloads.py
```diff
@@ -0,0 +1,318 @@
...
+) -> Tuple[GraphQLOperationPayload, int]:
+    bytes_limit -= GQL_OPERATION_PLACEHOLDER_SIZE
+    if bytes_limit < 0:
+        raise ValueError()
+    anonymize_gql_operation_response(operation, SENSITIVE_GQL_FIELDS)
+    name: Optional[JsonTruncText] = None
+    operation_type: Optional[str] = None
+    query: Optional[JsonTruncText] = None
+    result: Optional[JsonTruncText] = None
+    if operation.name:
```
korycins, line 148: Not sure if this is already added, but wouldn't be good to include a some const string like or something that would give you info, that te response that you see was limited.

saleor/webhook/observability_reporter/sensitive_data.py
```diff
@@ -0,0 +1,36 @@
...
+    "Address": {
+        "firstName",
+        "lastName",
+        "companyName",
+        "streetAddress1",
+        "streetAddress2",
+        "phone",
+    },
+    "AppTokenCreate": {"authToken"},
+    "AppToken": {"authToken"},
```
korycins, line 35: "App": "accessToken"

korycins: "Order": "userEmail"

korycins: "Payment": "creditCard"

korycins: "Checkout": "email"

korycins: This is after a few minutes search. I think we need to double check this

cmiacz:
https://app.clickup.com/t/2549495/CLOUD-1195

patrys:
Task linked: [CLOUD-1195 Run observability app performance test and measure costs](https://app.clickup.com/t/2549495/CLOUD-1195)

saleor/plugins/webhook/tasks.py
przlada: Done

saleor/plugins/webhook/tasks.py
przlada: Done

saleor/plugins/webhook/tasks.py
przlada: You are right. That was an artifact from an abandoned idea. Removed

saleor/graphql/views.py
przlada: Agree, fixed

saleor/webhook/observability_reporter/sensitive_data.py
przlada: Thank you, all added. I will try to gather more

saleor/webhook/observability_reporter/payloads.py
przlada: Yes, that's already added if I understand the question correctly. `JsonTruncText` object serializes to `text: str, truncated: bool`

saleor/webhook/observability_reporter/__init__.py
przlada: That's the last thing on my list of reporter optimizations, but I decided to skip it. It should add one DB query per request coming from apps only ([I tried to confirm it with this test](https://github.com/saleor/saleor/pull/9104/files#diff-67ea1df1ee3831a253d7ad11e34162acecd7f06af48b31a1ff00f2a9b397d440R38-R50)). But yes, that is still a lot of additional DB queries, which we can be reduced by adding caching. If you recommend it, I will do that.

patrys: Caching is not free. If we want to replace an SQL query with a cache query, make sure it's faster.

saleor/webhook/observability_reporter/utils.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class CustomJsonEncoder(DjangoJSONEncoder):
+    def default(self, o):
+        if isinstance(o, JsonTruncText):
+            return {"text": o.text, "truncated": o.truncated}
+        return super().default(o)
+
+
+def webhooks_for_event_exists(event_type, webhooks=None) -> bool:
```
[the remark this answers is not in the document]
korycins: Agree, looks like the same logic, but with exists() at the end.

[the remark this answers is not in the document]
przlada: Yes, they are the same. Great catch! Here is a PR #9505 that moves `_get_webhooks_for_event` to webhook utils, making it more reusable in the whole codebase. When it is accepted, I will rebase observability to use it. I should make that earlier, sorry.

saleor/webhook/observability_reporter/__init__.py
przlada: I will try to measure it, and we will decide

saleor/webhook/event_types.py
przlada: Ok, working on that

saleor/plugins/webhook/tasks.py
```diff
@@ -474,11 +484,88 @@ def send_webhook_request_sync(
...
+    events = buffer.get_events()
+    if not events:
+        return
+    domain = Site.objects.get_current().domain
+    event_type = WebhookEventAsyncType.OBSERVABILITY
+    for webhook in _get_webhooks_for_event(event_type):
+        scheme = urlparse(webhook.target_url).scheme.lower()
+        try:
+            if scheme in [WebhookSchemes.AWS_SQS, WebhookSchemes.GOOGLE_CLOUD_PUBSUB]:
+                for event_num, event in enumerate(events):
```
korycins, line 504: This is only a question: Would be good (and more readable) to wrap this `for` (same for http) inside a helper function that will send webhooks? `observability_send_events` looks like a looong function

saleor/webhook/observability_reporter/obfuscation.py
```diff
@@ -0,0 +1,193 @@
...
+                        child_node.type_condition.name.value
+                    )
+                self.contain_sensitive_field(child_node, inline_fragment_type)
+        return False
+
+    def enter_operation_definition(
+        self, node, key, parent, path, ancestors
+    ):  # pylint: disable=unused-argument
+        validate_sensitive_fields_map(self.sensitive_fields, self.context.get_schema())
+        if node.operation == "query":
```
korycins, line 108: Looks like you call `contain_sensitive_field` for all `if` but with different condition. Maybe we could use `if` to gather specific params (like `.get_query_type()`) and call `contain_sensitive_field` only once after processing the `if`s?

saleor/webhook/observability_reporter/payloads.py
```diff
@@ -0,0 +1,317 @@
...
+    initial_dump = _json_serialize(payload)
+    remaining_bytes = bytes_limit - len(initial_dump)
+    if remaining_bytes < 0:
+        raise ValueError(f"Payload too big. Can't truncate to {bytes_limit}")
+    try:
+        payload["gql_operations"]["operations"] = serialize_gql_operation_results(
+            gql_operations, remaining_bytes
+        )
+    except ValueError:
+        pass
```
korycins, line 230: Don't we need any logger here?

saleor/webhook/observability_reporter/payloads.py
```diff
@@ -0,0 +1,317 @@
...
+        raise ValueError(f"Payload too big. Can't truncate to {bytes_limit}")
+    trunc_resp_body = JsonTruncText.truncate(response_body, remaining_bytes // 3)
+    remaining_bytes -= trunc_resp_body.byte_size
+    data["response"]["body"] = trunc_resp_body
+
+    subscription_query = attempt.delivery.webhook.subscription_query
+    data["webhook"]["subscription_query"] = None
+    if subscription_query:
+        trunc_subscription_query = JsonTruncText.truncate(
+            subscription_query, remaining_bytes // 2
```
korycins, line 302: Wouldn't be good to wrap `2` inside a var with the proper name?

maarcingebala:
Closing in favor of https://github.com/saleor/saleor/pull/9803.

PR #9121, opened 2022-02-08
Fix Avalara crash with shipping methods webhook

rafalp:
This PR removes plugins call from checkout's subtotal calculation done for shipping methods webhook. This plugins call resulted in infinite recursion when both Avalara and webhook plugin are enabled.

This PR also adds test case that reproduced this exact crash before fix.

This change is in line with what we are planning to do for discounts, and from my testing it doesn't affect tax amounts on checkout.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ All notable, unreleased changes to this project will be documented in this file.
 `CheckoutAddPromoCode`, `CheckoutPaymentCreate` will raise a ValidationError when product in the checkout is
 unavailable - #8978 by @IKarbowiak
 - Fix disabled warehouses appearing as valid click and collect points when checkout contains only preorders - #9052 by @rafalp
+- Fix crash when Avalara plugin was used together with Webhooks plugin for shipping methods - #9121 by @rafalp
 
 
 # 3.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ All notable, unreleased changes to this project will be documented in this file.
 - Do no allow using `id` for updating checkout and order metadata - #8906 by @IKarbowiak
   - Use `token` instead
 - Remove `graphene-django` dependency - #9170 by @rafalp
+- Don't run plugins when calculating checkout's total price for available shipping methods resolution - #9121 by @rafalp
+  - Use either net or gross price depending on store configuration.
 
 ## Other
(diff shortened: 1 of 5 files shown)


patrys:
Task linked: [SALEOR-5816 Assign shipping method crash with Avalara](https://app.clickup.com/t/2549495/SALEOR-5816)

saleor/checkout/fetch.py
```diff
@@ -457,16 +458,25 @@ def get_valid_internal_shipping_method_list_for_checkout_info(
...
 ) -> List["ShippingMethodData"]:
+    from . import base_calculations
     from .utils import get_valid_internal_shipping_methods_for_checkout
 
     country_code = shipping_address.country.code if shipping_address else None
-    subtotal = manager.calculate_checkout_subtotal(
-        checkout_info, lines, checkout_info.shipping_address, discounts
-    )
+
+    line_totals = [
```
korycins, line 468: The one thing that I would consider is wrapping this loop it into a function and placed it in calculation.py file

fowczarek (approved):

IKarbowiak (approved):

korycins (approved):

CHANGELOG.md
```diff
@@ -48,6 +48,9 @@ All notable, unreleased changes to this project will be documented in this file.
 - Change metadata mutations to use token for order and checkout as identifier - #8426 by @IKarbowiak
   - After changes, using the order `id` for changing order metadata is deprecated
 - Add `withChoices` flag for Attribute type - #7733 by @dexon44
+`CheckoutAddPromoCode`, `CheckoutPaymentCreate` will raise a ValidationError when product in the checkout is
+unavailable - #8978 by @IKarbowiak
+- Fix crash when Avalara plugin was used together with Webhooks plugin for shipping methods - #9121 by @rafalp
```
stnatic, line 53: If I understand correctly, this PR changes the business logic behind calculating available shipping methods. Previously, I think always the gross amount was used? Now it uses the "base price" which can mean net or gross according to SiteSettings.

I think it would be worth to indicate the change in behaviour in the changelog on top of fixing the infinite recursion error.

rafalp: @stnatic this if fair point, I'll add this to changelog.

PR #9126, opened 2022-02-09
Fix consuming to much memory when updating product search document values

IKarbowiak:
Change `btach_size` for `set_product_search_document_values` task. I got information that for batch size `1000` it was still crashing sometimes, so I reduced it even more, to be sure that will not crash again.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/core/search_tasks.py
+++ b/saleor/core/search_tasks.py
@@ -63,9 +63,12 @@ def set_order_search_document_values(total_count, updated_count):
 
 @app.task
 def set_product_search_document_values(total_count, updated_count):
+    # set lower batch size as it was crashing for products with
+    # lots of attributes because out of memory issues
+    batch_size = 500
     qs = Product.objects.filter(search_document="").prefetch_related(
         *PRODUCT_FIELDS_TO_PREFETCH
-    )[:BATCH_SIZE]
+    )[:batch_size]
     if not qs:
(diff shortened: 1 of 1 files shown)


patrys:
Task linked: [SALEOR-5844 Updating Product.search_document consumes too much memory.](https://app.clickup.com/t/2549495/SALEOR-5844)

saleor/core/search_tasks.py
```diff
@@ -63,9 +63,12 @@ def set_order_search_document_values(total_count, updated_count):
 
 @app.task
 def set_product_search_document_values(total_count, updated_count):
+    # set lower batch size as it was crashing for products with
+    # lots of attributes because out of memory issues
+    batch_size = 500
```
korycins, line 68: Question: Don't we want to reduce te global batch_size defined in this file? Looks like the logic for `Order` could also fetch a lot of data from related objects.

korycins (approved):

saleor/core/search_tasks.py
IKarbowiak: The problem was only for `products` so I reduced it only there. 
Maybe we should consider reducing batch size for orders too, but I do not think it should be reduced as much as for products.

maarcingebala (approved):

PR #9135, opened 2022-02-10
Fix invalid available shipping methods calculation for checkout

IKarbowiak:
Port changes from #9102

**Issue description**

Given:
- checkout with shipping method price based set

Situation:
- shipping voucher is applied, the voucher discount reduces the price, so the minimal price condition for the current shipping method is not fulfilled

Previous behavior:
- the current shipping method was removed

Current solution
- the current method is not removed, as the voucher is applied on shipping, which conditions were already fulfilled;
(i.e. somebody applied free shipping voucher - we should ensure that the shipping is not removed)
- available shipping methods are updated when discount concerns entire orders or specific products

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/fetch.py
+++ b/saleor/checkout/fetch.py
@@ -61,6 +61,7 @@ class CheckoutInfo:
     delivery_method_info: "DeliveryMethodBase"
     all_shipping_methods: List["ShippingMethodData"]
     valid_pick_up_points: List["Warehouse"]
+    voucher: Optional["Voucher"] = None
 
     @property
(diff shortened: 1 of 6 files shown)


patrys:
Task linked: [SALEOR-4946 Adding free shipping voucher to checkout with subtotal price lower than shipping price removes shipping method](https://app.clickup.com/t/2549495/SALEOR-4946)

maarcingebala (approved):

PR #9141, opened 2022-02-11
Update exported files names and remove old exported files periodically

IKarbowiak:
- Use `uuid4` for generating hash in export file names
- Add periodic Celery task that removes exports older than specified time - default `30 days`

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/csv/tasks.py
+++ b/saleor/csv/tasks.py
@@ -1,11 +1,20 @@
 from typing import Dict, Union
 
+from celery.utils.log import get_task_logger
+from django.conf import settings
+from django.core.files.storage import default_storage
+from django.db.models import Q
+from django.db.models.expressions import Exists, OuterRef
+from django.utils import timezone
+
 from ..celeryconf import app
 from ..core import JobStatus
-from . import events
-from .models import ExportFile
+from . import ExportEvents, events
+from .models import ExportEvent, ExportFile
 from .notifications import send_export_failed_info
 from .utils.export import export_gift_cards, export_products
+
+task_logger = get_task_logger(__name__)
 
 
 def on_task_failure(self, exc, task_id, args, kwargs, einfo):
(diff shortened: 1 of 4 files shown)


saleor/csv/tasks.py
```diff
@@ -59,3 +68,30 @@ def export_gift_cards_task(
...
+    events = ExportEvent.objects.filter(
+        type=ExportEvents.EXPORT_PENDING,
+        date__lte=now - settings.EXPORT_FILES_TIMEDELTA,
+    ).values("export_file_id")
+    export_files = ExportFile.objects.filter(
+        Q(events__isnull=True) | Exists(events.filter(export_file_id=OuterRef("id")))
+    )
+
+    if not export_files:
+        return
```
L3str4nge, line 86: Shouldn't queryset require `exists ()` to check for empty?

saleor/csv/tasks.py
```diff
@@ -59,3 +68,30 @@ def export_gift_cards_task(
...
+
+    paths_to_delete = list(export_files.values_list("content_file", flat=True))
+
+    for path in paths_to_delete:
+        if path:
+            default_storage.delete(path)
+
+    count = export_files.count()
```
L3str4nge, line 94: I think we should count in loop. In case when `path is None` we have +1 undeleted item.

saleor/settings.py
```diff
@@ -534,6 +539,10 @@ def get_host():
         "task": "saleor.giftcard.tasks.deactivate_expired_cards_task",
         "schedule": crontab(hour=0, minute=0),
     },
+    "delete-old-export-files": {
+        "task": "saleor.csv.tasks.delete_old_export_files",
+        "schedule": crontab(hour=1, minute=0),
```
tomaszszymanski129, line 544: wouldn't once a day be enough?

saleor/csv/tasks.py
```diff
@@ -59,3 +68,30 @@ def export_gift_cards_task(
...
+    )
+
+    if not export_files:
+        return
+
+    paths_to_delete = list(export_files.values_list("content_file", flat=True))
+
+    for path in paths_to_delete:
+        if path:
+            default_storage.delete(path)
```
tomaszszymanski129, line 92: I'm wondering if we should secure it a bit in case it raises `FileNotFoundError` / `OSError`? might there be a case when the path doesn't actually exists?

saleor/csv/tasks.py
IKarbowiak: No, it'll work too :)

saleor/settings.py
IKarbowiak: It is once per day, at a specified time.

saleor/csv/tasks.py
IKarbowiak: I checked with S3 and default django storage and it's not raising an error when path does not exist. But I will consider adding it as I will be updating the counter.

IKarbowiak: And IMO there is no reason for raising an error in such cases.

L3str4nge (approved):

maarcingebala (approved):

PR #9093, opened 2022-02-07
Add attributes delete to page and page_type bulk delete mutation

CossackDex:
I want to merge this change because it adds missing associated attributes delete function to page and page_Type bulk delete mutation.

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #9162, opened 2022-02-14
Reference is not removed when user will remove object mvp

CossackDex:
I want to merge this change because it changes add a relation to reference. This PR also contain port https://github.com/saleor/saleor/pull/9093/files from 3.0 to 3.1 

This is only MVP, stay tuned for Part 2 with post MVP changes :D (Not ready for merge yet even after MVP review and QA)  

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-5087 Reference is not removed when user will remove object](https://app.clickup.com/t/2549495/SALEOR-5087)

saleor/attribute/migrations/0020_auto_20220214_1027.py
```diff
@@ -0,0 +1,50 @@
...
+def import_attribute_values(apps, schema):
+    AttributeValue = apps.get_model("attribute", "AttributeValue")
+    Page = apps.get_model("page", "Page")
+    Product = apps.get_model("product", "Product")
+
+    for attribute_value in AttributeValue.objects.iterator():
+        if attribute_value.attribute.entity_type == "Page":
+            page_id = Page.objects.filter(
+                pk=fetch_reference_id_from_slug(attribute_value.slug)
+            )[0]
```
IKarbowiak, line 20: It will raise an error when `Page` with given `id` will be not found.

saleor/attribute/migrations/0020_auto_20220214_1027.py
```diff
@@ -0,0 +1,50 @@
...
+            if page_id:
+                attribute_value.reference_page = page_id
+                attribute_value.save(update_fields=["reference_page"])
+            else:
+                attribute_value.delete()
+        elif attribute_value.attribute.entity_type == "Product":
+            product_id = Product.objects.filter(
+                pk=fetch_reference_id_from_slug(attribute_value.slug)
+            )[0]
```
IKarbowiak, line 29: Same here.

saleor/graphql/attribute/utils.py
```diff
@@ -257,17 +257,33 @@ def _pre_save_reference_values(
...
+                    attribute=attribute,
+                    reference_product=reference,
+                    slug=slugify(
+                        f"{instance.id}_{reference.id}",  # type: ignore
+                        allow_unicode=True,
+                    ),
+                    defaults={"name": getattr(reference, field_name)},
+                )[0]
+                for reference in attr_values.references
+            )
```
IKarbowiak, line 286: We can simplify this. You can create dict with values  that are the same for `page` and `product` references and then update it with `reference_page` or `reference_product`. And at the end unpack it in the `get_or_create` method.

saleor/graphql/product/tests/test_bulk_delete.py
```diff
@@ -658,6 +659,14 @@ def test_delete_products_with_file_attributes(
     ) == {v.file_url for v in values}
 
 
+def test_delete_products_delete_product_reference():
+    return False
```
L3str4nge, line 663: `assert True == True` would be better :D

saleor/attribute/migrations/0020_auto_20220214_1027.py
```diff
@@ -0,0 +1,122 @@
...
+
+def import_attribute_values(attr_values_qs, Model):
+    attr_values_to_update = []
+    attr_values_to_delete = set()
+
+    attr_value_map = defaultdict(list)
+    for attr_value in attr_values_qs:
+
+        ref_id = fetch_reference_id_from_slug(attr_value.slug)
+        attr_value_map[ref_id].append(attr_value)
```
IKarbowiak, line 22: In your solution, you iterate through all `AttributeValues` two times. I will suggest firstly getting all attribute values slugs with the use of `values_list`, then preparing a dict with the attribute value `slug` as a key and `ref_id` as a value. In the next step, we should get all reference objects with the use of the `in_bulk` method. Next, we can iterate through every value and check if the given attribute value is in our slug - reference mapping. If it is, we should assign reference object to `reference_page` or `reference_product` field and add this attr value to `attr_values_to_update` list (it should be also set by the way). Otherwise, add  attr value to `attr_values_to_delete` set.

saleor/attribute/migrations/0020_auto_20220214_1027.py
```diff
@@ -0,0 +1,122 @@
...
+            )
+        )
+    )
+    for batch_pks in queryset_in_batches(queryset):
+        batch = AttributeValue.objects.filter(pk__in=batch_pks)
+        attr_values_to_update, attr_values_to_delete = import_attribute_values(
+            batch, Product
+        )
+        AttributeValue.objects.bulk_update(attr_values_to_update, ["reference_product"])
+        AttributeValue.objects.filter(pk__in=attr_values_to_delete).delete()
```
IKarbowiak, line 89: These two parts look similar. IMO we can create one generic function for that, or at least separate it into two different methods.

L3str4nge (approved):

saleor/csv/utils/products_data.py
```diff
@@ -437,9 +441,15 @@ def prepare_attribute_value(attribute_data: AttributeData):
...
-        value = f"{attribute_data.entity_type}_{reference_id}"
+    elif input_type == AttributeInputType.REFERENCE and (
+        attribute_data.reference_page or attribute_data.reference_product
+    ):
+        if attribute_data.reference_page:
+            reference_id = attribute_data.reference_page
+            value = f"{attribute_data.entity_type}_{reference_id}"
+        else:
+            reference_id = attribute_data.reference_product
+            value = f"{attribute_data.entity_type}_{reference_id}"
```
IKarbowiak, line 452:
```suggestion
        if attribute_data.reference_page:
            reference_id = attribute_data.reference_page
        else:
            reference_id = attribute_data.reference_product
        value = f"{attribute_data.entity_type}_{reference_id}"
```

IKarbowiak (approved):
👏

saleor/webhook/serializers.py
```diff
@@ -46,11 +46,16 @@ def serialize_product_or_variant_attributes(
...
-    def _prepare_reference(attribute, attr_slug):
+    def _prepare_reference(attribute, attr_value):
         if attribute.input_type != AttributeInputType.REFERENCE:
             return
+        if attribute.entity_type == AttributeEntityType.PAGE:
+            reference_pk = attr_value.reference_page_id
+        elif attribute.entity_type == AttributeEntityType.PRODUCT:
+            reference_pk = attr_value.reference_product_id
+        else:
+            reference_pk = attr_value.slug.split("_")[1]
```
maarcingebala, line 57: In which case would this logic be used? We shouldn't have to split the slugs anymore. I think we should rather return None.

saleor/graphql/attribute/utils.py
```diff
@@ -257,17 +259,36 @@ def _pre_save_reference_values(
...
+        reference_list = []
+        for ref in attr_values.references:
+            reference_page = None
+            reference_product = None
+
+            if attribute.entity_type == AttributeEntityType.PAGE:
+                if isinstance(ref, Page):
+                    reference_page = ref
+                else:
+                    raise GraphQLError
```
maarcingebala, line 272: Is this GraphQLError handled anywhere in the mutation? I think the validation should happen earlier at the "cleaning" step and not here in the "saving" step. We have for example `_validate_references` which might be a good place to move this validation.

saleor/graphql/attribute/utils.py
```diff
@@ -435,9 +448,14 @@ def _validate_references(
...
-        values.references = ref_instances
-        return values
+        try:
+            ref_instances = get_nodes(
+                references, attribute.entity_type, model=entity_model
+            )
+            values.references = ref_instances
+            return values
+        except GraphQLError as e:
+            raise ValidationError({ValidationError(str(e), code="graphql_error")})
```
maarcingebala, line 458: Did you mean (1):
```
raise ValidationError(str(e), code="graphql_error")
```
or (2)
```
raise ValidationError({"references": ValidationError(str(e), code="graphql_error")})
```
We use the 1 when we are returning a general error that isn't related to any specific input field and 2 when we are referring to a specific input field causing the error. Please also add a unit test for this error.

maarcingebala: Also, the code is not correct as it is not a graphql error anymore. The code should be rather `INVALID`. Don't hardcode the error code string, there is the `error_class` param that provides you with an error class to get the constant from (see other usages of `error_class` in this file).

saleor/webhook/serializers.py
```diff
@@ -46,11 +46,16 @@ def serialize_product_or_variant_attributes(
 ) -> List[Dict]:
     data = []
 
-    def _prepare_reference(attribute, attr_slug):
+    def _prepare_reference(attribute, attr_value):
```
maarcingebala, line 49: Is this function tested?

saleor/graphql/attribute/utils.py
CossackDex: 👍

PR #9187, opened 2022-02-18
Handle ProductMedia delete in task

jakubkuc:
I want to merge this change because we want to handle product media deletion in task to prevent 
long ProductMediaDelete mutation executions.
Deleting ProductMedia will switch new BooleanField `to_remove` into `True` and run a task to clear out all connected media files from s3 and delete ProductMedia object from db.
Deleting Product will send pre_delete signal that will switch flag and run the task. (pre delete will be triggered as well when Product will be deleted by CASCADE)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-4783 Long time to execute queries in ProductMediaDelete](https://app.clickup.com/t/2549495/SALEOR-4783)

saleor/graphql/product/mutations/products.py
```diff
@@ -1700,8 +1701,19 @@ class Meta:
...
+                        "Media not found.",
+                        code=ProductErrorCode.NOT_FOUND,
+                    )
+                }
+            )
         media_id = media_obj.id
-        media_obj.delete()
+        media_obj.to_remove = True
+        media_obj.save(update_fields=["to_remove"])
+        delete_product_media_task.delay(media_id)
```
L3str4nge, line 1716: I would run the task only if `ProductMedia` exists. Basically move line this line from task here `product_media = ProductMedia.objects.filter(pk=media_id, to_remove=True).first()`

IKarbowiak: AFAIK we cannot pass media object to the celery task.

IKarbowiak (approved):

Qasem-h (approved):

saleor/graphql/product/mutations/products.py
jakubkuc: Yes, just like @IKarbowiak mentioned. That is why I passed media id media can be fetched by task itself

maarcingebala (approved):

saleor/core/tasks.py
```diff
@@ -21,3 +24,16 @@ def delete_event_payloads_task():
     EventPayload.objects.filter(
         deliveries__isnull=True, created_at__lte=event_payload_delete_period
     ).delete()
+
+
+@app.task(
+    autoretry_for=(ClientError,),
```
fowczarek, line 30: Should we auto-retry only on `ClientError`?
@korycins

fowczarek (approved):

saleor/core/tasks.py
korycins: Do we support only S3 for storing images? It looks like we took an AWS error from the lower layer used in versatile.

PR #9193, opened 2022-02-18
Add possibility for plugins to execute code before each mutation

NyanKiyoshi:
This would allow plugins to execute custom logic whenever a mutation is invoked without needing to override the core logic.
The call would only happen once the permissions are checked.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/account/mutations/base.py
+++ b/saleor/graphql/account/mutations/base.py
@@ -12,6 +12,7 @@ from ....account.notifications import (
 )
 from ....account.search import prepare_user_search_document_value
 from ....checkout import AddressType
+from ....core.db.utils import set_mutation_flag_in_context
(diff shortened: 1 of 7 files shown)


korycins (approved):

korycins:
@NyanKiyoshi  Should manager check if the result is in expected type of Exception?

IKarbowiak (approved):

L3str4nge (approved):

jakubkuc (approved):

NyanKiyoshi:
> @NyanKiyoshi Should manager check if the result is in expected type of Exception?

Graphene will silently override the returned value if it's invalid, thus I would say we should check it. I will handle this tomorrow

maarcingebala (approved):

PR #9225, opened 2022-02-22
Catch update_fields did not affect any rows errors and return response with message

L3str4nge:
~~Possible solution for race conditions such as `Save with update_fields did not affect any rows`~~

~~Documentation: https://docs.djangoproject.com/en/3.2/ref/models/querysets/#select-for-update.~~

~~There are not many issues on sentry but it does happen sometimes. For now we have around 3 entries a week.~~
~~I think we need to monitor system behaviour after applying the fix as it does not guarantee a 100% solution.~~

This fix catches `update_fields did not affect any rows` errors and returns response with message. Covered two most common places in sentry (about 300 events at all).

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-4478 DatabaseError (save with update_fields did not affect any rows) on various mutations](https://app.clickup.com/t/2549495/SALEOR-4478)

saleor/graphql/checkout/mutations.py
```diff
@@ -987,19 +987,23 @@ def perform_mutation(
         if token:
             checkout = get_checkout_by_token(
                 token,
-                qs=models.Checkout.objects.prefetch_related(
-                    "lines__variant__product__product_type"
-                ),
+                qs=models.Checkout.objects.select_for_update()
```
korycins, line 990: Why `select_for_update` twice?

L3str4nge: My bad

saleor/graphql/meta/mutations.py
```diff
@@ -60,8 +60,15 @@ def get_instance(cls, info, **data):
             # ShippingMethodType represents the ShippingMethod model
             if type_name == "ShippingMethodType":
                 qs = shipping_models.ShippingMethod.objects
+
+            if qs:
+                qs = qs.select_for_update(of=("self",))
```
fowczarek, line 65: Are we sure that we want to lock objects during medatada mutation?

mzawad: TBH not sure if but there are some event on sentry about that: https://sentry.io/organizations/saleor/issues/2625281955/?project=5543592

L3str4nge: Not sure tbh bot there are some event on sentry about that: https://sentry.io/organizations/saleor/issues/2625281955/?project=5543592

fowczarek: Maybe we shouldn't execute save if there is no metadata to change?

L3str4nge: Agreed but this is not the main issue. The main problem is the situation where during `save` we have `None` object (kinda race condition).

fowczarek: Maybe we should catch this error and return msg that this object does not exist?

L3str4nge: Yeah thought about it. Sound good for meta but I am wondering about checkout. It might be good to handle this in the same way instead of locking the object? Wdyt?

fowczarek: Yup, we should lock objects only in necessary places.

L3str4nge: @fowczarek Refactored at all.

L3str4nge:
Added small lib for unit testing race conditions: https://github.com/c-oreills/before_after

fowczarek (approved):

IKarbowiak (approved):

saleor/graphql/meta/mutations.py
```diff
@@ -32,7 +36,25 @@ def _save_instance(instance, metadata_field: str):
...
+        instance.save(update_fields=fields)
+    except DatabaseError as e:
+        msg = (
+            "Cannot update metadata for instance: %s. "
+            "Possible updating not existing object. "
+            "Details: %s.",
+            instance,
+            str(e),
+        )
+        logger.warning(msg)
```
korycins, line 49: You wrote: `Possible updating not existing object.`. What kind of different situation can cause this issue? If missing object is the only one, then we could return code `NOT_FOUND` in `ValidationError`

saleor/graphql/product/utils.py
```diff
@@ -111,3 +117,28 @@ def clean_variant_sku(sku: Optional[str]) -> Optional[str]:
...
+                "Possible updating not existing object. "
+                "Details: %s.",
+                media,
+                str(e),
+            )
+            logger.warning(msg)
+
+            raise ValidationError(
+                {
+                    "order": ValidationError(
```
korycins, line 139: Are you sure, that the `order` field is OK here? I would say, that we should include here a `media`. Also it happens in the loop: 
- Shouldn't we return a list of exception here?
- Don't we need a transaction here? To make sure that the whole operation finished with success or fail? Currently, there is a risk of partial update.

saleor/graphql/meta/mutations.py
L3str4nge: I am not sure about the other options tbh, thats why I am putting `e` in the msg to track it in the future.

saleor/graphql/product/utils.py
L3str4nge: Good point!

saleor/graphql/meta/mutations.py
L3str4nge: [fixed](https://github.com/saleor/saleor/pull/9225/commits/237af39e573412557186f5d8febf2a63e845d270)

saleor/graphql/product/utils.py
L3str4nge: [fixed](https://github.com/saleor/saleor/pull/9225/commits/237af39e573412557186f5d8febf2a63e845d270)

saleor/graphql/product/mutations/products.py
```diff
@@ -73,8 +74,11 @@
     get_draft_order_lines_data_for_variants,
     get_used_attribute_values_for_variant,
     get_used_variants_attribute_values,
+    update_ordered_media,
 )
 
+logger = logging.getLogger(__name__)
```
maarcingebala, line 80: Do we need this logger?

L3str4nge: If we don't want to `logger.warning(msg)` I would say no :)

maarcingebala: We don't have `logger.warning(msg)` here or I am blind.

maarcingebala: So can we remove it? :)

L3str4nge: We have it [here](https://github.com/saleor/saleor/pull/9225/files#diff-ff3d47ae262b3250574ecd74a7750c0b04ddf0d0d84498605810753a4ba094dfR49) actually. Do we need this log, if not I would remove it :)

L3str4nge: LOL, I looked at the wrong file. Removed :D

maarcingebala (approved):

PR #9231, opened 2022-02-23
Better API for fetching metadata

patrys:
This exploration was inspired by discussion https://github.com/saleor/saleor/discussions/9223

It makes working with metadata easier in languages like JS where unmarshalling a JSON structure happens automatically.

New API (we likely need a better field name but `metadata` is already taken):

```graphql
query {
  product(id: ...) {
    metafields
  }
}
```

Response:

```json
{
  "data": {
    "product": {
      "metafields": {
        "myKey": "myValue",
        "anotherKey": "anotherValue"
      }
    }
  }
}
```

You can also ask for only particular keys to be returned:

```graphql
query {
  product(id: ...) {
    metafields(keys: ["foo", "myKey"])
  }
}
```

Or ask for a single key:

```graphql
query {
  product(id: ...) {
    fooValue: metafield(key: "foo")
  }
}
```

I declared a new scalar type so strictly typed clients can provide a custom type override like `Map<string, string>`.

For now, a write/update API is out of scope.

If we like that, we should also implement a similar interface for all Editor.js content.

Tracked under SALEOR-6043

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/meta/types.py
+++ b/saleor/graphql/meta/types.py
@@ -1,9 +1,12 @@
 import graphene
+from graphene.types.generic import GenericScalar
 
 from ...core.models import ModelWithMetadata
 from ..channel import ChannelContext
+from ..core.descriptions import ADDED_IN_33, PREVIEW_FEATURE
 from ..core.types import NonNullList
 from .resolvers import (
+    check_private_metadata_privilege,
     resolve_metadata,
     resolve_object_with_metadata_type,
     resolve_private_metadata,
@@ -15,13 +18,50 @@ class MetadataItem(graphene.ObjectType):
     value = graphene.String(required=True, description="Value of a metadata item.")
 
 
+class Metadata(GenericScalar):
+    """Metadata is a map of key-value pairs, both keys and values are `String`.
+
+    Example:
+    ```
+    {
+        "key1": "value1",
+        "key2": "value2"
+    }
+    ```
+
+    """
+
+
+def _filter_metadata(metadata, keys):
+    if keys is None:
+        return metadata
+    return {key: value for key, value in metadata.items() if key in keys}
+
+
 class ObjectWithMetadata(graphene.Interface):
     private_metadata = NonNullList(
         MetadataItem,
         required=True,
         description=(
-            "List of private metadata items."
-            "Requires proper staff permissions to access."
+            "List of private metadata items. Requires staff permissions to access."
+        ),
+    )
+    private_metafield = graphene.String(
+        args={"key": graphene.NonNull(graphene.String)},
+        description=(
+            "A single key from private metadata. "
+            "Requires staff permissions to access.\n\n"
+            "Tip: Use GraphQL aliases to fetch multiple keys.\n\n"
+            f"{ADDED_IN_33}\n\n{PREVIEW_FEATURE}"
+        ),
+    )
+    private_metafields = Metadata(
+        args={"keys": NonNullList(graphene.String)},
+        description=(
+            "Private metadata. Requires staff permissions to access. "
+            "Use `keys` to control which fields you want to include. "
+            "The default is to include everything.\n\n"
+            f"{ADDED_IN_33}\n\n{PREVIEW_FEATURE}"
         ),
     )
     metadata = NonNullList(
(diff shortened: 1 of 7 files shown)


patrys:
Related tracker item: SALEOR-6043

patrys:
CC @benoj

korycins:
(for anyone who doesn't remember current metadata structure)
So, instead of representing it below format:
`metadata: [{key: "1", value: "4", __typename: "MetadataItem"}]`
we will have:
`meta: {"1": "4"}`

Personally, I like the idea, looks much simpler.

benoj:
LGTM :shipit:

patrys:
If we all like it, do you think it's worth merging as is (we still need tests, of course)? Any suggestions for a better field name?

timuric:
@patrys I think `meta` is slightly ambiguous if I you want to be pedantic about the naming. However I can't think of decent alternatives. 
Here are few ideas:
`meta_fields` - implies that there are only fields?
`meta_values` - implies that there are only values?
`metaJSON` - not sure if it is accurate though

jwm0:
Looks good, I'd say just call it `metafields` and it seems descriptive enough

maarcingebala (commented):
`metafields` / `privateMetafields` sound good. To make the PR mergable we should add tests and probably mark the existing metadata fields as deprecated?

timuric:
@patrys could you please also document `metafield`

```
query {
  product(id: ...) {
    metafield(key: "foo")
  }
}
```

patrys:
@timuric I would need to implement that first.

IKarbowiak (approved):

saleor/graphql/meta/types.py
```diff
@@ -15,13 +17,48 @@ class MetadataItem(graphene.ObjectType):
...
-            "Requires proper staff permissions to access."
+            "List of private metadata items. Requires staff permissions to access."
+        ),
+    )
+    private_metafield = graphene.String(
+        args={"key": graphene.NonNull(graphene.String)},
+        description=(
+            "A single key from private metadata. "
+            "Requires staff permissions to access. "
+            "Use GraphQL aliases to fetch multiple keys."
```
maarcingebala, line 53: All new fields should be marked with ADDED_IN_32 label, also should we mark it as preview?

patrys: Good catch!

patrys: @maarcingebala I've added the version string to the end of field descriptions because otherwise, the schema becomes a long list of "New in Saleor 3.2...". I've also reworded the "New in Saleor X.Y" to say "Added in Saleor X.Y" because those strings will remain in place until we release 4.0. It's not a feature, so I'm unsure about the preview status. @timuric?

patrys: I've talked to @timuric about it. We'll keep it marked as Feature Preview until 3.3.

korycins (approved):

saleor/graphql/meta/types.py
```diff
@@ -1,9 +1,13 @@
 import graphene
+from graphene.types.generic import GenericScalar
+
+from saleor.graphql.core.descriptions import ADDED_IN_32, PREVIEW_FEATURE
```
maarcingebala, line 4: For consistency we should use relative imports here.

saleor/graphql/meta/types.py
```diff
@@ -15,13 +19,50 @@ class MetadataItem(graphene.ObjectType):
...
+            "List of private metadata items. Requires staff permissions to access."
+        ),
+    )
+    private_metafield = graphene.String(
+        args={"key": graphene.NonNull(graphene.String)},
+        description=(
+            "A single key from private metadata. "
+            "Requires staff permissions to access.\n\n"
+            "Tip: Use GraphQL aliases to fetch multiple keys.\n\n"
+            f"{ADDED_IN_32}\n\n{PREVIEW_FEATURE}"
```
maarcingebala, line 56: `ADDED_IN_33`

maarcingebala:
@eCzerniak I don't see a Clickup task so I'm assigning you here. You should get how to test it from the PR description, but if needed please ask @patrys for instructions.

eCzerniak (approved):

PR #9235, opened 2022-02-23
Optimize assign_quantity_allocated_to_stocks to work in batches

tomaszszymanski129:
I want to merge this change because it optimizes assign_quantity_allocated_to_stocks migration func to work in batches. Might result in long SQL execution times otherwise.

# Optimization checklist

* [x] The changes are tested (migration works)
* [x] Number of queries is consistent
* [x] Memory usage is low
* [x] Memory usage does not increase between chunk iterations
* [x] Single query execution time is relatively low on large dataset (to be checked)

### DIFF
--- a/saleor/warehouse/migrations/0016_stock_quantity_allocated.py
+++ b/saleor/warehouse/migrations/0016_stock_quantity_allocated.py
@@ -8,17 +8,21 @@ from django.db.models.functions import Coalesce
 def assign_quantity_allocated_to_stocks(apps, schema_editor):
     Stock = apps.get_model("warehouse", "Stock")
 
-    Stock.objects.update(
-        quantity_allocated=Subquery(
-            Stock.objects.filter(id=OuterRef("id"))
-            .annotate(
-                allocations_allocated=Coalesce(
-                    Sum("allocations__quantity_allocated"), 0
+    batch_size = 1000
+    stocks_ids = Stock.objects.values_list("id", flat=True)
+    for idx in range(0, len(stocks_ids), batch_size):
+        ids_to_update = stocks_ids[idx : idx + batch_size]  # noqa: E203
+        Stock.objects.filter(id__in=ids_to_update).update(
+            quantity_allocated=Subquery(
+                Stock.objects.filter(id=OuterRef("id"))
+                .annotate(
+                    allocations_allocated=Coalesce(
+                        Sum("allocations__quantity_allocated"), 0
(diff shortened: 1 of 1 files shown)


patrys:
Task linked: [SALEOR-6057 Optimize 0016_stock_quantity_allocated migration](https://app.clickup.com/t/2549495/SALEOR-6057)

saleor/warehouse/migrations/0016_stock_quantity_allocated.py
```diff
@@ -8,17 +8,21 @@
...
-    Stock.objects.update(
-        quantity_allocated=Subquery(
-            Stock.objects.filter(id=OuterRef("id"))
-            .annotate(
-                allocations_allocated=Coalesce(
-                    Sum("allocations__quantity_allocated"), 0
+    batch_size = 1000
+    stocks_ids = Stock.objects.values_list("id", flat=True)
+    for idx in range(0, len(stocks_ids), batch_size):
+        ids_to_update = stocks_ids[idx : idx + batch_size]  # noqa: E203
```
patrys, line 14: That is going to be very costly for the SQL server as it relies on offsets.

patrys (changes requested):
Please use an approach that does not rely on offsets:

```python
last_id = 0

while True:
  stock_ids = list(Stock.objects.filter(id__gt=last_id).order_by('id').values_list('id', flat=True)[:batch_size])
  if not stock_ids:
    break
  last_id = stock_ids[-1]
  ...  # perform the update

File: saleor/warehouse/migrations/0016_stock_quantity_allocated.py [hunk 2]
```diff
@@ -8,17 +8,28 @@
...
+        last_id = stock_ids[-1]
+        Stock.objects.filter(id__in=stock_ids).update(
+            quantity_allocated=Subquery(
+                Stock.objects.filter(id=OuterRef("id"))
+                .annotate(
+                    allocations_allocated=Coalesce(
+                        Sum("allocations__quantity_allocated"), 0
+                    )
                 )
+                .values("allocations_allocated")[:1]
```
IKarbowiak, line 30: IMO, we should separate part for getting batches to another function, as it's hard to ready right now.

IKarbowiak (approved):

korycins (approved):

maarcingebala (approved):

PR #9242, opened 2022-02-24
Add benchmarks / fix checkout invalidation for taxes

stnatic:
Contents of this PR:
- added benchmark for tax calculation requests made
- fixed invalidation issue (the condition for checkout prices expiry was flipped the wrong way, which got discovered by the above benchmark)
- added missing invalidations in checkout mutations
- moved discount recalculation into the prices invalidation cycle
- reduced impact of the PR on the existing benchmark tests

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


CHANGELOG.md
```diff
@@ -52,7 +52,6 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add `withChoices` flag for Attribute type - #7733 by @dexon44
 `CheckoutAddPromoCode`, `CheckoutPaymentCreate` will raise a ValidationError when product in the checkout is
 unavailable - #8978 by @IKarbowiak
-- Fix crash when Avalara plugin was used together with Webhooks plugin for shipping methods - #9121 by @rafalp
```
korycins, line 55: 🚔

saleor/checkout/calculations.py
```diff
@@ -257,23 +241,32 @@ def fetch_checkout_prices_if_expired(
...
     last price update is greater than settings.CHECKOUT_PRICES_TTL.
     """
-    checkout = checkout_info.checkout
+    from .utils import recalculate_checkout_discount
 
-    if not force_update and checkout.price_expiration < timezone.now():
+    checkout = checkout_info.checkout
+    if (
+        not force_update
+        and checkout.price_expiration
```
korycins, line 249: when do we have a case that this value will be None?

saleor/checkout/fetch.py
```diff
@@ -227,7 +227,7 @@ def fetch_checkout_lines(
                 "variant__attributes__values",
             ]
         )
-    lines = checkout.lines.prefetch_related(*prefetched_fields)
+    lines = checkout.lines.select_related().prefetch_related(*prefetched_fields)
```
korycins, line 230: Why it is needed?

saleor/checkout/models.py
```diff
@@ -138,7 +138,7 @@ class Checkout(ModelWithMetadata):
         max_digits=5, decimal_places=4, default=Decimal("0.0")
     )
 
-    price_expiration = models.DateTimeField(default=timezone.now)
+    price_expiration = models.DateTimeField(null=True, default=None, blank=True)
```
korycins, line 141: What is a reason for giving a possibility to set this to None? Would be good to describe this change at least in PR's describtion

CHANGELOG.md
stnatic: This was a bad merge, it's already in line 58

saleor/checkout/calculations.py
stnatic: I need to make a final decision on it. I thought using `None` instead of `timezone.now()` might be easier for forced invalidation (and especially testing it) but I think I might go back to `timezone.now()`

saleor/graphql/order/dataloaders.py
```diff
@@ -63,6 +63,9 @@ class OrderLinesByOrderIdLoader(DataLoader):
     def batch_load(self, keys):
         lines = (
             OrderLine.objects.using(self.database_connection_name)
+            # Products are prefetched, as they are needed in all
+            # financial calculations related to order lines
+            .select_related("variant__product")
```
maarcingebala, line 68: Not sure about this, this will load the products and variants always, even if you're just loading plain order lines, which adds additional queries. Can't we instead use the product/variants dataloader where we are sure that we need products/variants? We can easily chain results of multiple dataloaders which we do in many resolvers.

maarcingebala: Can't you load products with a data loader to a separate queryset and just pass it down to where it's needed?

saleor/checkout/fetch.py
korycins: Can't we use `prefetched_fields` to explicitly provide a list of fields that we want to prefetch? I am not sure that fetching all relations is correct solution here. 
From django docs:


> it is possible to call select_related() with no arguments. This will follow all non-null foreign keys it can find - nullable foreign keys must be specified. This is not recommended in most cases as it is likely to make the underlying query more complex, and return more data, than is actually needed.

[docs](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#select-related)

stnatic: Good point! I'm now we should just define those select_related fields explicitly, my bad for using something that had unwanted consequences.

saleor/graphql/order/dataloaders.py
stnatic: I'll just revert back to prefetching in `fetch_order_prices_if_expired` - this function will be called only once for every single unconfirmed order change, so while the benchmark might look like a yellow flag, in reality this is going to be cached pretty much permanently.

saleor/checkout/models.py
stnatic: It was much easier to deal with it in tests, but eventually I refactored tests so that timezone.now is used for invalidation

stnatic:
There are 2 unit tests failing, as voucher code isn't detached until prices are recalculated. Is hiding the "voucher code" attribute behind calculations the right way to handle it?

saleor/checkout/fetch.py
```diff
@@ -227,7 +227,9 @@ def fetch_checkout_lines(
                 "variant__attributes__values",
             ]
         )
-    lines = checkout.lines.prefetch_related(*prefetched_fields)
+    lines = checkout.lines.select_related(
+        "variant", "variant__product"
```
korycins, line 231: Lets create similar list as we have for prefetched_fields

saleor/checkout/utils.py
```diff
@@ -542,15 +544,6 @@ def recalculate_checkout_discount(
...
             )
-            checkout.save(
-                update_fields=[
-                    "translated_discount_name",
-                    "discount_amount",
-                    "discount_name",
-                    "currency",
-                    "last_change",
-                ]
-            )
```
fowczarek, line 553: Are we sure that we should remove this save? Whole `else` statements change `checkout` objects and don't save them.

saleor/graphql/checkout/mutations/checkout_shipping_address_update.py
```diff
@@ -163,7 +162,6 @@ def perform_mutation(
                 manager,
                 shipping_channel_listings,
             )
-        recalculate_checkout_discount(manager, checkout_info, lines, discounts)
```
fowczarek, line 166: IMO we should recalculate discounts. When someone changes the shipping address, then the shipping method could change. Afterr the shipping method changed voucher could be inapplicable.

saleor/graphql/checkout/tests/benchmark/test_homepage.py
```diff
@@ -193,3 +194,74 @@ def test_user_checkout_details_with_external_shipping_method(
...
+
+
+@patch("saleor.plugins.webhook.tasks.send_webhook_request_sync")
+def test_user_checkout_details_with_tax_app(
+    mock_send_request,
+    user_api_client,
+    customer_checkout,
+    tax_app,
+    settings,
+):
```
fowczarek, line 206:
```suggestion
@pytest.mark.django_db
@pytest.mark.count_queries(autouse=False)
@patch("saleor.plugins.webhook.tasks.send_webhook_request_sync")
def test_user_checkout_details_with_tax_app(
    mock_send_request,
    user_api_client,
    customer_checkout,
    tax_app,
    settings,
    count_queries
):
```

saleor/graphql/checkout/tests/test_checkout_digital.py
```diff
@@ -45,11 +45,14 @@ def test_create_checkout(
...
-            MUTATION_CHECKOUT_CREATE, {"checkoutInput": checkout_input}
+    print(
+        get_graphql_content(
+            api_client.post_graphql(
+                MUTATION_CHECKOUT_CREATE, {"checkoutInput": checkout_input}
+            )
         )
-    )["data"]["checkoutCreate"]
+    )
+    # ["data"]["checkoutCreate"]
```
fowczarek, line 55: We should remove it.

saleor/checkout/utils.py
stnatic: It was moved to: https://github.com/saleor/saleor/pull/9242/files#diff-3a4d5975bee96cf58c1c7f362360641d5cf6030a1817e0b14ea92561d98e28eeR274

All discount/price recalculation will happen inside `fetch_checkout_prices_if_expired`

saleor/graphql/checkout/mutations/checkout_shipping_address_update.py
stnatic: This happens inside `fetch_checkout_prices_if_expired`

saleor/checkout/fetch.py
```diff
@@ -214,6 +214,7 @@ def fetch_checkout_lines(
     """Fetch checkout lines as CheckoutLineInfo objects."""
     from .utils import get_voucher_for_checkout
 
+    selected_fields = ["variant__product__product_type"]
```
L3str4nge, line 217: I would rename it to `related_fields`

saleor/plugins/webhook/tasks.py
```diff
@@ -130,8 +132,11 @@ def trigger_all_webhooks_sync(
     this function returns None.
     """
     webhooks = _get_webhooks_for_event(event_type)
-    event_payload = EventPayload.objects.create(payload=data)
+    event_payload = None
     for webhook in webhooks:
+        event_payload = event_payload or EventPayload.objects.create(
```
korycins, line 137: Why do you create EventPayload for each webhook? Is it expected? If yes, why don't you use bulk action here?

saleor/graphql/checkout/tests/benchmark/test_checkout_mutations.py
```diff
@@ -912,9 +912,10 @@ def test_add_checkout_lines_with_external_shipping(
         api_client.post_graphql(MUTATION_CHECKOUT_LINES_ADD, variables)
     )
     assert not response["data"]["checkoutLinesAdd"]["errors"]
-    # One api call:
+    # Two api calls:
     # - post-mutate() logic used to validate currently selected method
-    assert mock_send_request.call_count == 1
+    # - fetch_checkout_prices_if_expired - calculating all prices for checkout
+    assert mock_send_request.call_count == 2
```
fowczarek, line 918: Could we add checks for these calls?

fowczarek (approved):

saleor/plugins/webhook/tasks.py
maarcingebala: If the payload is the same there should be only one EventPayload object created in the database.

patrys:
@stnatic Black is failing during linter checks.

saleor/graphql/checkout/tests/benchmark/test_checkout_mutations.py
fowczarek: Are we sure that we should have duplicated external calls in this mutation?

stnatic: Full background here:

Those are actually duplicated calls and their number was already 2 in this optimization pull request.
https://github.com/saleor/saleor/pull/8744/files#diff-732cfdb4bfc0be8786c912bc299bb2ea95e56326ca5c8d62907c82117886e057R915-R917

In a subsequent pull request calls to external shipping methods have been wrapped with lazy initialization which reduced the number of calls to 1, since the shipping method wasn't queried in particular benchmark test. The reason for the number of calls on this pull request going back to 2 is that `fetch_checkout_prices_if_expired` requires evaluating the shipping price.

The real reason for the duplicated call here is the fact that `ChecokutInfo` is evaluated twice:
- mutations call `fetch_checkout_info`
- `checkout_info.checkout` is then passed as the root object to the GraphQL checkout type
- finally another `CheckoutInfo` instance is constructed inside of `CheckoutInfoByCheckoutTokenLoader`

There are 2 possible solutions:
- add cache for external shipping methods (there's a draft PR somewhere that addresses this)
- use CheckoutInfo as the root object for Checkout queries, as this will avoid double evaluation in all checkout mutations

saleor/graphql/checkout/mutations/checkout_remove_promo_code.py
```diff
@@ -135,5 +143,7 @@ def remove_promo_code_by_id(
                 )
             if checkout.voucher_code == node.code:
                 remove_voucher_from_checkout(checkout)
+                return True
```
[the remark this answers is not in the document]
stnatic: I added a docstring and added `return False`

fowczarek:
@stnatic 2 tests fail on CI.

saleor/plugins/webhook/plugin.py
```diff
@@ -616,17 +616,19 @@ def process_payment(
...
         )
 
     def get_taxes_for_order(
         self, order: "Order", previous_value
     ) -> Optional["TaxData"]:
-        payload = generate_order_payload(order, self.requestor)
         return trigger_all_webhooks_sync(
-            WebhookEventSyncType.ORDER_CALCULATE_TAXES, payload, parse_tax_data
+            WebhookEventSyncType.ORDER_CALCULATE_TAXES,
+            lambda: generate_order_payload(order, self.requestor),
```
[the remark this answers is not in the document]
stnatic: I'd try to address `trigger_all_webhooks_sync` on the final PR (I think we don't need the additional to complexity introduced by this method as compared to other sync webhooks). There's a separate PR that is going to add dedicated methods for generating payloads for different checkout/order lifecycle hooks so, I wouldn't touch it here either.

maarcingebala (approved):

maarcingebala (approved):

korycins (approved):

fowczarek (approved):

PR #9263, opened 2022-02-28
Add order/checkout serializers for payloads with and without taxes

mateuszgrzyb:
# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/plugins/webhook/plugin.py
```diff
@@ -636,8 +634,9 @@ def get_shipping_methods_for_checkout(
         apps = App.objects.for_event_type(
             WebhookEventSyncType.SHIPPING_LIST_METHODS_FOR_CHECKOUT
         ).prefetch_related("webhooks")
+        print(apps)
```
stnatic, line 637: 👮‍♂️

saleor/plugins/webhook/plugin.py
```diff
@@ -14,26 +14,24 @@
...
     generate_product_deleted_payload,
     generate_product_payload,
     generate_product_variant_payload,
     generate_product_variant_with_stock_payload,
-    generate_requestor,
     generate_sale_payload,
     generate_translation_payload,
 )
+from ...webhook.payloads_utils import generate_meta, generate_requestor
+from ...webhook.taxed_payloads import generate_checkout_payload, generate_order_payload
```
stnatic, line 34: I wouldn't call the module `taxed_payloads` if there's an option do `taxed=False`

saleor/webhook/payloads.py
```diff
@@ -1080,3 +835,530 @@ def generate_excluded_shipping_methods_for_checkout_payload(
...
+
+
+def _generate_checkout_prices_data_without_taxes(
+    checkout: Checkout,
+) -> Dict[str, Decimal]:
+    def qp(amount: Decimal) -> Decimal:
+        return quantize_price(amount, checkout.currency)
+
+    return {
+        "subtotal_net_amount": qp(checkout.subtotal.net.amount),
```
stnatic, line 1225: Why net amount? Shouldn't this be called "base"?

saleor/webhook/payloads.py
```diff
@@ -1080,3 +835,530 @@ def generate_excluded_shipping_methods_for_checkout_payload(
...
+        "price_with_discounts_gross_amount": unit_price_with_discounts.gross.amount,
+    }
+
+
+def _get_line_prices_data_without_taxes(
+    checkout: "Checkout",
+    line_info: CheckoutLineInfo,
+) -> Dict[str, Decimal]:
+    return {
+        "price_net_amount": quantize_price(
```
stnatic, line 1318: Why net amount instead of "base"?

mateuszgrzyb: We already have a "base" price field in checkout lines, it's the amount specified in channel listing.

stnatic: Isn't that exactly the one that we need (plus a discount)?

saleor/core/prices.py
```diff
@@ -1,16 +1,18 @@
...
 from babel.numbers import get_currency_precision
 
 if TYPE_CHECKING:
     from django.db.models import Model
-    from prices import Money, TaxedMoney, TaxedMoneyRange
+    from prices import Money, TaxedMoney, TaxedMoneyRange  # noqa
 
-MoneyTypes = Union["TaxedMoney", "Money", Decimal, "TaxedMoneyRange"]
+MoneyType = TypeVar(
+    "MoneyType", "TaxedMoney", "Money", Decimal, "TaxedMoneyRange"  # noqa
```
korycins, line 11: Why do you include new type as a `type` that belongs to?
`MoneyType = TypeVar("MoneyType")`

saleor/plugins/webhook/tests/test_webhook.py
```diff
@@ -618,6 +619,39 @@ def test_checkout_updated(
...
+    mocked_get_webhooks_for_event,
+    mocked_fetch,
+    any_webhook,
+    settings,
+    checkout,
+    app,
+    permission_manage_shipping,
+):
+    app.permissions.add(permission_manage_shipping)
+    app.save()
```
korycins, line 637: Don't think that you need to call `.save()` when you're adding external  relations.

saleor/plugins/webhook/tests/test_webhook.py
```diff
@@ -618,6 +619,39 @@ def test_checkout_updated(
     )
 
 
+@freeze_time("1914-06-28 10:50")
+@mock.patch("saleor.checkout.calculations.fetch_checkout_prices_if_expired")
+@mock.patch("saleor.plugins.webhook.plugin._get_webhooks_for_event")
+@mock.patch("saleor.plugins.webhook.plugin.trigger_webhook_sync")
+def test_get_shipping_methods_for_checkout_untaxed(
```
korycins, line 626: What exacly this test is testing? I see that the assert at the end is that we expect that we don't make any call, but the name of the test doesn't say that this is expected.

saleor/plugins/webhook/tests/test_webhook.py
```diff
@@ -618,6 +619,39 @@ def test_checkout_updated(
...
+):
+    app.permissions.add(permission_manage_shipping)
+    app.save()
+    webhook = Webhook.objects.create(
+        name="webhook",
+        app=app,
+    )
+    WebhookEvent.objects.create(
+        webhook=webhook,
+        event_type=WebhookEventSyncType.SHIPPING_LIST_METHODS_FOR_CHECKOUT,
```
korycins, line 644: IMHO you could use `shipping_app` fixture

saleor/core/prices.py
mateuszgrzyb: I've had problems with mypy using `quantize_price` function. When passing `Decimal`, inferred return value was of type `TaxedMoney`. The return type should be exactly the same as parameter type and `TypeVar` is better suited for that case.

saleor/webhook/tests/test_webhook_payloads.py
```diff
@@ -103,7 +112,6 @@ def test_generate_order_payload(
     assert payload["collection_point_name"] is None
     assert payload.get("token") == order_with_lines.token
     assert payload.get("shipping_method")
-    assert payload.get("shipping_tax_rate")
```
korycins, line 106: Is it expected?

saleor/webhook/payloads.py
```diff
@@ -1080,3 +1082,304 @@ def generate_excluded_shipping_methods_for_checkout_payload(
...
+    }
+
+
+def _generate_order_prices_data_without_taxes(
+    order: "Order",
+    included_taxes_in_price: bool,
+) -> Dict[str, Decimal]:
+    def quantize_base_price(price: TaxedMoney) -> Decimal:
+        return quantize_price(
+            _get_base_price(price, included_taxes_in_price), order.currency
```
korycins, line 1121: I would name it something like `untaxed_price_amount`. Untaxed because we're using a value stored in Saleor DB. `price_amount` because you're returning a decimal value

saleor/webhook/tests/test_webhook_payloads.py
```diff
@@ -231,76 +349,43 @@ def test_order_lines_have_all_required_fields(order, order_line_with_one_allocat
...
-        "undiscounted_total_price_net_amount": str(undiscounted_total_price_net_amount),
-        "undiscounted_total_price_gross_amount": str(
-            undiscounted_total_price_gross_amount
-        ),
-        "product_metadata": product.metadata,
-        "product_type_metadata": product_type.metadata,
-        "voucher_code": line.voucher_code,
-        "sale_id": line.sale_id,
-    }
+    assert line_payload["id"] == line_id
```
korycins, line 365: The previous assert was checking that payload doesn't have any additional fields. Now you can check 10 fields where `line_payload` will have 20 fields.

saleor/plugins/webhook/tests/test_webhook.py
mateuszgrzyb: I've changed the name of this test function, it should be clearer now what is it testing

saleor/webhook/tests/test_webhook_payloads.py
mateuszgrzyb: Yes, we're testing it in `test_generate_order_payload_prices`, as it isn't a field that will be in both taxed and untaxed payloads.

saleor/webhook/payloads.py
mateuszgrzyb: I don't think that's too big of a problem. `quantize_base_price` doesn't add any new logic, it just combines two calls of existing functions to decrease repetition and increase readability.

saleor/webhook/serializers.py
```diff
@@ -1,59 +1,114 @@
...
 
 
+def serialize_checkout_lines_without_taxes(
+    checkout: "Checkout",
+    lines: Iterable["CheckoutLineInfo"],
+    included_taxes_in_price: bool,
+) -> List[dict]:
+    def untaxed_price_amount(price: TaxedMoney) -> Decimal:
+        return quantize_price(
+            get_base_price(price, included_taxes_in_price), checkout.currency
```
L3str4nge, line 98: Method name says we are serializing lines without taxes but we have `included_taxes_in_price` to control taxes.

saleor/webhook/tests/test_webhook_serializers.py
```diff
@@ -73,52 +74,46 @@ def test_serialize_product_attributes(
...
 
+ATTRIBUTES = sentinel.ATTRIBUTES
+
+
+@patch(
+    "saleor.webhook.serializers.serialize_product_or_variant_attributes",
+    new=Mock(return_value=ATTRIBUTES),
+)
+@pytest.mark.parametrize("taxes_included", [True, False])
+# TODO: split this test by test function
```
L3str4nge, line 85: Please split this test instead of leaving `TODO`

mateuszgrzyb: Resolved in recent commit.

saleor/webhook/tests/test_webhook_serializers.py
```diff
@@ -73,52 +74,46 @@ def test_serialize_product_attributes(
...
-            "currency",
-            "unit_price_net_amount",
-            "unit_price_gross_amount",
-        ],
-    )
 
     # when
-    checkout_lines_data = serialize_checkout_lines(checkout)
+    if serialize_checkout_lines == serialize_checkout_lines_with_taxes:
+        checkout_lines_data = serialize_checkout_lines(checkout_info, manager, lines)
```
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit.

saleor/webhook/tests/test_webhook_payloads.py
mateuszgrzyb: Resolved in recent commit.

saleor/webhook/tests/test_webhook_payloads.py
```diff
@@ -43,108 +43,424 @@
...
-            payment_txn_captured.captured_amount.quantize(Decimal("0.01"))
+
+@freeze_time()
+@pytest.mark.parametrize("taxes_included", [True, False])
+@mock.patch("saleor.webhook.payloads._generate_order_lines_payload_with_taxes")
+@mock.patch("saleor.webhook.payloads.generate_fulfillment_lines_payload")
+def test_generate_order_payload(
+    mocked_fulfillment_lines,
+    mocked_order_lines,
+    mocked_fetch_order,
```
korycins, line 109: do you still needed it?

saleor/core/prices.py
fowczarek: Maybe we should better name that `MoneyType`.

saleor/core/prices.py
```diff
@@ -1,16 +1,18 @@
 from decimal import Decimal
-from typing import TYPE_CHECKING, Iterable, Union
+from typing import TYPE_CHECKING, Iterable, TypeVar
 
 from babel.numbers import get_currency_precision
 
 if TYPE_CHECKING:
     from django.db.models import Model
-    from prices import Money, TaxedMoney, TaxedMoneyRange
+    from prices import Money, TaxedMoney, TaxedMoneyRange  # noqa
```
fowczarek, line 8: Why `# noqa`  here?

saleor/webhook/payloads.py
```diff
@@ -1080,3 +1159,178 @@ def generate_excluded_shipping_methods_for_checkout_payload(
...
+    }
+
+
+def generate_checkout_payload(
+    checkout: "Checkout",
+    requestor: Optional["RequestorOrLazyObject"] = None,
+):
+    manager = get_plugins_manager()
+    lines, _ = fetch_checkout_lines(checkout, prefetch_variant_attributes=True)
+    checkout_info = fetch_checkout_info(checkout, lines, [], manager)
```
fowczarek, line 1302: Shouldn't we pass discounts to this function?

mateuszgrzyb: Resolved in recent commit

saleor/core/prices.py
mateuszgrzyb: `flake8` failed there, as it claimed that those types are unused, because they were used as forward references. I think we can just import them normally and remove those `# noqa`

saleor/core/prices.py
mateuszgrzyb: How about `PriceType`?

saleor/webhook/serializers.py
mateuszgrzyb: I've changed the name from `included_taxes_in_price` to `use_gross_as_base_price` where appropriate

saleor/webhook/tests/test_webhook_payloads.py
mateuszgrzyb: Yes, I want to make sure that we're sending fresh prices in order payload

saleor/webhook/payloads.py
```diff
@@ -137,61 +135,128 @@ def _charge_taxes(order_line: OrderLine) -> Optional[bool]:
...
+def _generate_order_lines_payload_with_taxes(
+    order: Order,
+    manager: PluginsManager,
+    lines: Iterable[OrderLine],
+):
+    def qp(price: TaxedMoney) -> TaxedMoney:
+        return quantize_price(price, order.currency)
+
+    def get_unit_price(line: OrderLine) -> TaxedMoney:
+        return qp(
```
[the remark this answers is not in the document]
mateuszgrzyb: Resolved in recent commit

fowczarek (approved):

saleor/webhook/payloads.py
```diff
@@ -1080,3 +1150,176 @@ def generate_excluded_shipping_methods_for_checkout_payload(
...
+    }
+
+
+def generate_order_payload(
+    order: "Order",
+    requestor: Optional["RequestorOrLazyObject"] = None,
+    with_meta: bool = True,
+):
+    manager = get_plugins_manager()
+    lines = OrderLine.objects.prefetch_related("variant__product__product_type")
```
korycins, line 1204: I think that, You can do here select_related. This should reduce amount of queries

saleor/webhook/payloads.py
```diff
@@ -1080,3 +1150,176 @@ def generate_excluded_shipping_methods_for_checkout_payload(
...
+        included_taxes_in_prices=include_taxes_in_prices(),
+    )
+
+
+def generate_order_payload_without_taxes(
+    order: "Order",
+    requestor: Optional["RequestorOrLazyObject"] = None,
+    with_meta: bool = True,
+):
+    lines = OrderLine.objects.prefetch_related("variant__product__product_type")
```
korycins, line 1223: I think that, You can do here select_related. This should reduce amount of queries

saleor/webhook/tests/test_webhook_serializers.py
```diff
@@ -73,52 +75,112 @@ def test_serialize_product_attributes(
...
+            product, collections, checkout.channel, variant_channel_listing
+        )
+        currency = checkout.currency
+        assert data == {
+            "id": graphene.Node.to_global_id("CheckoutLine", line.pk),
+            "sku": variant.sku,
+            "quantity": line.quantity,
+            "charge_taxes": product.charge_taxes,
+            "base_price": quantize_price(base_price.amount, currency),
+            "price_net_amount": ANY,
```
korycins, line 127: Can we change it to real values? It is important for us, to send values after quantitze them, so would be good to check this. (DB  stores values with 3 number after coma: Decimal(10.000), webhook payload should send Decimal(10.00)

korycins: This comment is related to all tests related to price fields, where we used `ANY`

korycins (approved):

maarcingebala (approved):

PR #9281, opened 2022-03-03
Add migration to change order id from int to UUID

IKarbowiak:
Change order `id` from `int` to `UUID`.
These changes add only the migration and changes required to make tests pass.

⚠️ 
This PR is to the task branch, as these changes require additional actions, which will be done in separate PRs.

What is changing
- the order `id` change type from `int` to `UUID`
- extend `order` model with field `number`
- all old ids has been rewritten to the new `number` field
- the `number` field is auto-incrementing
- extend `order` model with `use_old_id` field
- set `use_old_id` to `True` for all existing orders
- all relations have been rewritten to refer to `UUID` instead of `int`

The rewriting relations for many to one relation contains the following steps:
- Before changing order `id` type:
  - introduce new field `order_token`
  - fulfill `order_token` with corresponding token values from assigned orders
  - change `to_field` in existing `order` relations to pointed on `number` instead of `id` order field
- After change:
  - change `order_token` fields into `ForeginKeys`
  - remove old `order` relation
  - rename `order_token` to `order`
 
For many to many relations:
- Before changing order `id` type:
  - add a new field `ordr_token` in the junction table
  - fill `order_token` with corresponding token values from assigned orders
  - drop old FK constraint for `order_id` field that was pointed to order `id`
  - add a new FK constraint for `order_id` field that point to order `number`
- After change:
  - drop the FK constraint for `order_id`
  - add the FK for `order_token`
  - drop `order_id`
  - rename `order_token` to `order_id`

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-6139 Add migration for changing order id from int to uuid](https://app.clickup.com/t/2549495/SALEOR-6139)

L3str4nge (approved):

saleor/plugins/webhook/tests/test_shippping_exclude_cache.py
```diff
@@ -13,7 +13,7 @@
 @mock.patch(
     "saleor.plugins.webhook.plugin.generate_excluded_shipping_methods_for_order_payload"
 )
-def test_excluded_shipping_methods_for_order_use_cache(
+def _use_cache(
```
fowczarek, line 16: :police_officer:

fowczarek (approved):

saleor/order/models.py
```diff
@@ -82,7 +82,18 @@ def ready_to_confirm(self):
         return self.filter(status=OrderStatus.UNCONFIRMED)
 
 
+def get_order_number():
+    with connection.cursor() as cursor:
+        cursor.execute("SELECT nextval('order_order_number_seq')")
+        result = cursor.fetchone()
+        return result[0]
```
NyanKiyoshi, line 89: Isn't there a chance that this will cause races? I would say we should have postgres do auto-increment rather than making a SQL query to get the next value

NyanKiyoshi, line 89: This may cause races, surely there is a better way than doing this. We should leave it to postgres to do the auto-increment rather than making a SQL query to get the next value.

Maybe `BigAutoField` for `bigserial`? Hopefully it supports non-primary key fields https://github.com/django/django/blob/fac3dd7f390d372736e05974cc5c3ef1a3768fbb/django/db/backends/postgresql/base.py#L75-L76

NyanKiyoshi (approved):

maarcingebala (approved):

PR #9300, opened 2022-03-04
Allow plugins to create their custom error code

LeOndaz:
Plugins which expose graphql playgrounds using their webhooks couldn't create their own error codes because of this check.

Note that plugins can create graphql playgrounds since #8665


LeOndaz:
This is tested 

Please tell me what you think guys: @patrys @maarcingebala @korycins 

Thank you.

LeOndaz:
This is just to notify you about the existence of this PR: @patrys @maarcingebala @korycins or at least tell me about someone who's regularly active in here that I contact because I'm regularly active in the repo.

Thank you.

saleor/graphql/core/utils/error_codes.py
```diff
@@ -96,10 +96,12 @@ def get_error_code_from_error(error) -> str:
...
+    if (
+        code is None
+        or code in DJANGO_VALIDATORS_ERROR_CODES
+        or code in DJANGO_FORM_FIELDS_ERROR_CODES
+    ):
         return "invalid"
     if isinstance(code, Enum):
         code = code.value
-    if code not in saleor_error_codes:
-        return "invalid"
```
maarcingebala, line 104: I think the changes look good but I'm wondering if we need to remove this part. Can you provide an example of how you raise those custom errors in your plugin? I think if you used an enum to represent errors, the code would be handled in the previous condition and removing this one wouldn't be necessary? I mean for example this usage:

```
code=AccountErrorCode.DELETE_OWN_ACCOUNT
```

LeOndaz: @maarcingebala A plugin exposes a playground at /plugins/plugin-id/ and has a `ModelMutation` with an `error_type_class` of a custom error created with a custom enum created, always says "expected a value of type `error_type_class_here`, received invalid"

LeOndaz: This part restricts the codes to be from the internal saleor error codes, we want to add custom ones in this case

LeOndaz: The workaround currently is to use any random error code from saleor itself, but this does not help when extending the dashboard because you usually show a message depending on the code

LeOndaz: So yes, your example works perfectly, because AccountErrorCode is in `saleor_error_codes` but `MyPluginErrorCode` is not

maarcingebala: I've checked and you're right. We can remove this part then, but in this case, we should also remove `saleor_error_codes` and `SALEOR_ERROR_CODE_ENUMS` which won't be needed anymore.

LeOndaz: Great then, I haven't seen if they were used elsewhere, I'll check and if so, I'll remove them in a few minutes

LeOndaz: I've currently pushed the commit, if it passes, you're good to go

maarcingebala:
It seems we have a broken import in Vatlayer:
```
File "/__w/saleor/saleor/saleor/plugins/vatlayer/plugin.py", line [21](https://github.com/saleor/saleor/runs/5445815716?check_suite_focus=true#step:7:21), in <module>
    from ...graphql.core.utils.error_codes import PluginErrorCode
ImportError: cannot import name 'PluginErrorCode' from 'saleor.graphql.core.utils.error_codes' (/__w/saleor/saleor/saleor/graphql/core/utils/error_codes.py)
```

It's importing `PluginErrorCode` from the wrong file.

LeOndaz:
@maarcingebala yes I'm currently aware of this and trying to see anything that imports from error_codes atm

LeOndaz:
Okay I think we don't have anything other than that, I ran `pre-commit` and it went fine

maarcingebala:
@LeOndaz Black fails in CI, can you reformat files once again? Also, I upgraded Black to the newest version yesterday.

saleor/plugins/vatlayer/plugin.py
```diff
@@ -39,12 +32,8 @@
...
     from ...discount import DiscountInfo
     from ...order.models import Order, OrderLine
-    from ...product.models import (
-        Collection,
-        Product,
-        ProductVariant,
-        ProductVariantChannelListing,
-    )
+    from ...product.models import (Collection, Product, ProductVariant,
+                                   ProductVariantChannelListing)
```
maarcingebala, line 36: I think this formatting is incorrect now.

maarcingebala (approved):

jakubkuc (approved):

LeOndaz:
@maarcingebala @jakubkuc can I get this merged?

maarcingebala:
@LeOndaz Something it's still failing in CI.

maarcingebala:
@LeOndaz We could merge this one but need to fix the linters first, would you be able to handle this?

PR #9312, opened 2022-03-07
Bump dependencies

maarcingebala:
Bump dependencies.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


LeOndaz:
@maarcingebala any possibilities of bumping graphene as well? because all graphql libraries are using graphql core v3 except graphene v2 (which saleor is using), this will allow the possibility of not using graphene django in plugins and use any alternatives, because currently, the only thing that supports graphql core v2 is graphene django (same version as graphene)

maarcingebala:
@LeOndaz We would love to do that but I think we are still blocked. Some work has been done recently to unblock it, but i need to double-check with the team where we are with this. Here is a PR with more details: https://github.com/saleor/saleor/pull/8503

PR #9326, opened 2022-03-09
Custom prices API preview

IKarbowiak:
This is a preview of the custom prices feature in Saleor.

The feature will allow defining your own prices for product variants added to the checkout. Below you can find main assumptions:
- The price changes will be allowed only for apps with new `HANDLE_CHECKOUT` permission. Other users will get the `PermissionDenied` error when trying to provide the custom price.
- The new permission will be granted to all apps that already have `MANAGE_CHECKOUT`. 
- The `CheckoutLineInput` will be extended with the optional price field `price: PositiveDecimal`, so the custom price could be set in the following mutations:
  - `checkoutCreate`
  - `checkoutLinesAdd` - when adding a variant that is already added to checkout, don't create any new line, but override the already existing one - increment the quantity and update the price.
  - `checkoutLinesUpdate` - will override the existing line with the price provided in the mutation
- Whether the given price is net or gross should depend on the shop settings (the same settings that decide whether variant.get_price is considered net or gross.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/schema.graphql
+++ b/saleor/graphql/schema.graphql
@@ -1191,6 +1191,7 @@ type CheckoutLineDelete {
 input CheckoutLineInput {
   quantity: Int!
   variantId: ID!
+  price: PositiveDecimal
 }
 
 type CheckoutLinesAdd {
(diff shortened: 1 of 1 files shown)


patrys:
@IKarbowiak Do we need to grant the new permission to any existing Apps? It does not seem like they risk losing any existing powers.

IKarbowiak:
@patrys not to any existing apps, only to those that have `manage_checkouts`.

IKarbowiak:
The implementation of the feature can be found here: https://github.com/saleor/saleor/pull/9393.

PR #9329, opened 2022-03-09
Save images to product media from external URLs

krzysztofwolski:
I want to merge this change because I would like to upload images from external URLs using productMediaCreate. I would use it to import images together with other data from external services.

How productMediaCreate work right now?
- if an image file is attached in input.file, then it will be saved into saleor media files
- in case if using input.media_url, no file is saved into media. ProductMedia object will contain only the URL to the external source (there's a restricted list of supported services, so users cannot use their own hosting). Example: user adds media from URL https://flic.kr/p/J2EuC1. Object media will return URL pointing to https://live.staticflickr.com/816/27585811288_797a384807_b.jpg

How productMediaCreate work with my modifications?
There are no breaking changes - when using media services like youtube or flickr, ProductMedia will contain only the URL to an external source. 
If URL is pointing to the image file (I'm determining it based on file extension and response headers), the file will be downloaded and put into media files.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/core/tests/test_core.py
+++ b/saleor/graphql/core/tests/test_core.py
@@ -1,7 +1,7 @@
 import enum
 import os
 from io import BytesIO
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
 import django_filters
 import graphene
@@ -29,8 +29,11 @@ from ..utils import (
     add_hash_to_file_name,
     clean_seo_fields,
     get_duplicated_values,
+    get_filename_from_url,
+    is_image_mimetype,
     snake_to_camel_case,
     validate_image_file,
+    validate_image_url,
     validate_slug_and_generate_if_needed,
 )
 from . import ErrorTest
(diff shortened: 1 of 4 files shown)


patrys:
Task linked: [SALEOR-6183 Uploading pictures to product media from URLs](https://app.clickup.com/t/2549495/SALEOR-6183)

maarcingebala:
@krzysztofwolski What are the next steps with this PR?

saleor/graphql/core/tests/test_core.py
```diff
@@ -319,6 +322,111 @@ def test_validate_image_file_invalid_file_extension():
...
+
+def test_is_image_mimetype_valid_mimetype():
+    # given
+    valid_mimetype = "image/jpeg"
+
+    # when
+    result = is_image_mimetype(valid_mimetype)
+
+    # then
+    assert result
```
L3str4nge, line 348: Could be just `assert is_image_mimetype("image/jpeg")`

saleor/graphql/core/tests/test_core.py
```diff
@@ -319,6 +322,111 @@ def test_validate_image_file_invalid_file_extension():
...
+
+def test_is_image_mimetype_invalid_mimetype():
+    # given
+    invalid_mimetype = "application/javascript"
+
+    # when
+    result = is_image_mimetype(invalid_mimetype)
+
+    # then
+    assert not result
```
L3str4nge, line 359: Could be just `assert not is_image_mimetype"application/javascript")`

saleor/graphql/product/mutations/products.py
```diff
@@ -1519,14 +1524,29 @@ def perform_mutation(cls, _root, info, **data):
...
-                oembed_data=oembed_data,
-            )
+        if media_url:
+            # Remote URLs can point to the images or oembed data.
+            # In case of images, file is downloaded. Otherwise we keep only
+            # URL to remote media.
+            if is_image_url(media_url):
+                validate_image_url(media_url, "media_url", ProductErrorCode.INVALID)
+                filename = get_filename_from_url(media_url)
+                image_data = requests.get(media_url, stream=True)
```
L3str4nge, line 1534: I am wondering about the security. In this case we don't have any validation against malware/suspicious files so we can upload anything and I am not sure that checking mime type and content type is enough.  Correct me if I am wrong.
 @maarcingebala @patrys

maarcingebala: Maybe we could just add validation of the mime types and support these? https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#image_types

maarcingebala: Ah nevermind, `is_image_mimetype` already does it. I'm not sure if we can check anything else here.

maarcingebala (approved):

saleor/graphql/product/mutations/products.py
krzysztofwolski: Can I mark it as resolved?

saleor/graphql/core/tests/test_core.py
krzysztofwolski: Talked with Marcin - wanted to keep consistency with the other tests and it can stay

saleor/graphql/core/tests/test_core.py
krzysztofwolski: As in previous comment

IKarbowiak (approved):

PR #9332, opened 2022-03-09
Fix S3 and GS implementation of file urls 

LeOndaz:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/core/types/common.py
```diff
@@ -406,6 +406,20 @@ class File(graphene.ObjectType):
 
     @staticmethod
     def resolve_url(root, info):
+        if settings.AWS_MEDIA_BUCKET_NAME:
+            s3_url = "http://{bucket_name}.s3.amazonaws.com{path}"
```
jbergstroem, line 410: I'm not sure that hardcoding the paths here is the way.. it seems to ignore `AWS_S3_ENDPOINT_URL` which is used for custom s3 setups (we lean on https://min.io for instance)

L3str4nge: Imho it's not a good idea. We should define base url in SETTINGS

LeOndaz: @jbergstroem I agree, we can check in line 410 if AWS_S3_ENDPOINT_URL is supplied, provide the file from it and if not, provide it from the base `s3.amazonaws.com`, what do you think?

LeOndaz: > Imho it's not a good idea. We should define base url in SETTINGS

No need to define more variables in this case, what do you think?

HazemKhaled: give a look here https://docs.saleor.io/docs/3.x/developer/running-saleor/s3#environment-variables

This one ignoring `AWS_STATIC_CUSTOM_DOMAIN` and/or `AWS_MEDIA_CUSTOM_DOMAIN`

Don't forget to care about STORAGE & MEDIA cases `AWS_STORAGE_BUCKET_NAME`

HazemKhaled (changes requested):

PR #9340, opened 2022-03-10
Always cast database ID to int in dataloader

maarcingebala:
Shipping methods ID can be of two types:
- int - when a shipping method from Saleor is used 
- string - when an eternal shipping method is used.

Internally string is used to pass the ID between functions, but when this happens the `ShippingMethodByIdLoader` fails to resolve the shipping method, as it uses integers internally. This quick-fix casts IDs to int inside the data loader, which should always resolve shipping methods from Saleor DB.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/fetch.py
+++ b/saleor/checkout/fetch.py
@@ -134,7 +134,7 @@ class ShippingMethodInfo(DeliveryMethodBase):
     @property
     def delivery_method_order_field(self) -> dict:
         if not self.delivery_method.is_external:
-            return {"shipping_method_id": self.delivery_method.id}
+            return {"shipping_method_id": int(self.delivery_method.id)}
         return {}
 
     def is_valid_delivery_method(self) -> bool:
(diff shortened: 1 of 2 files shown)


stnatic:
Hey @maarcingebala,

I believe you're fixing the issue where checkoutComplete isn't returning the deliveryMethod attribute correctly on the first send?

This actually came up in QA as an issue that was present on 3.1 since introducing external shipping methods. The main culprit is the second line here:
image

We're assigning a dataclass instance to a model field and it magically saved to the database due to some default casts, but maybe we should fix the origin of the error, rather than patching the data loader. The fact that a dataclass is assigned instead of the shipping method instance could have other, yet unidentified consequences in the system

maarcingebala:
Yep I saw it but wasn't sure if I can cast to int in that place, as I'm not sure how it works with external shipping methods, which doesn't have numeric IDs if I'm correct? If we are sure that we will always have a numeric value there, I'm happy to change it.

stnatic:
For external shipping methods `checkout.shiping_method is None` and the string id is kept in private metadata

fowczarek (approved):

IKarbowiak (approved):

jakubkuc (approved):

PR #8734, opened 2021-12-06
Improve product search

IKarbowiak:
- Introduce `search_document` product field, that collect values from all fields used in searching
- Drop `search_vector` field and related triggers
- Add index for `search_document` product field
- Update `search_document` whenever any of the fields are changed
- Make search case insensitive
- Update random data - generate search value for created products
- Move commands for updating search documents to migrations

Explain (5000 products):
https://explain.dalibo.com/plan/c4K

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #8158, opened 2021-09-21
Fix cursor in product search results

rafalp:
Fixes `RANK` based sorting for products so `cursor` value returned for each edge is correct. Also fixes `before` and `after` pagination for same sorting.

Fixes #8011 

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [x] Privileged queries and mutations are guarded by proper permission checks
* [x] Database queries are optimized and the number of queries is constant
* [x] Database migration files are up to date
* [x] The changes are tested
* [x] GraphQL schema and type definitions are up to date
* [x] Changes are mentioned in the changelog

PR #9344, opened 2022-03-11
Full-text search for products

patrys:
Unlike our previous search, this approach does not support substring matches. However, it gives us proper result ranking with different weights assigned to product names, attributes, and descriptions.

`RANK` is no longer deprecated as the sorting option.

It also supports websearch expressions:
- `"foo bar"` for word distance (proximity)
- `foo -bar` for excluding terms
- `foo OR bar` for alternatives

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/product/sorters.py
```diff
@@ -123,7 +123,7 @@ class Meta:
 
 class ProductOrderField(graphene.Enum):
     NAME = ["name", "slug"]
-    RANK = ["name", "slug"]
+    RANK = ["search_rank"]
```
fowczarek, line 126: We should add a unique field as the last order field. Without it, pagination couldn't work. 
```suggestion
    RANK = ["search_rank", "slug"]
```

saleor/product/search.py
```diff
@@ -51,7 +58,28 @@ def prepare_product_search_document_value(
     return search_document.lower()
 
 
-def generate_product_fields_search_document_value(product: "Product"):
+def prepare_product_search_vector_value(
+    product: "Product", *, already_prefetched=False
+) -> SearchVector:
+    if not already_prefetched:
+        prefetch_related_objects([product], *PRODUCT_FIELDS_TO_PREFETCH)
+    search_vector = SearchVector(Value(product.name), weight="A") + SearchVector(
```
stnatic, line 66: Are you providing the search config? It's useful for languages other than English, as it would apply stemming, etc.
https://docs.djangoproject.com/en/4.0/ref/contrib/postgres/search/#changing-the-search-configuration

The downside is that AWS RDS supports only a limited number of languages and there's no option to upload a dictionary without filesystem access

saleor/core/search_tasks.py
```diff
@@ -104,3 +109,32 @@ def set_search_document_values(
...
+        )
+        instances.append(instance)
+    Model.objects.bulk_update(instances, ["search_document", "search_vector"])
+
+    updated_count += len(instances)
+    progress = round((updated_count / total_count) * 100, 2)
+
+    task_logger.info(
+        f"Updated {updated_count} from {total_count} {Model.__name__.lower()}s - "
+        f"{progress}% done."
```
fowczarek, line 138: Shouldn't we pass values as params?

fowczarek (commented):
In #8734 we drop fixes from #8158 about cursors builds from ranks. Should we revert it?

saleor/product/search.py
patrys: I explicitly don't provide a config so we don't have to deal with stop words and stemming for now.

saleor/core/search_tasks.py
patrys: This is not new code, I created a vector version of all search_document functions so we can compare them while this is in draft.

patrys: > In #8734 we drop fixes from #8158 about cursors builds from ranks. Should we revert it?

Good catch, we should.

patrys:
> In #8734 we drop fixes from #8158 about cursors builds from ranks. Should we revert it?

Good catch, we should.

saleor/core/management/commands/update_search_indexes.py
```diff
@@ -15,7 +15,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         # Update products
-        products_total_count = Product.objects.filter(search_document="").count()
+        products_total_count = Product.objects.filter(search_vector=None).count()
```
cmiacz, line 18: This can cause memory issues on big databases. Please consider to reduce BATCH size in saleor/core/search_tasks.py

patrys: There is no counting any more.

fowczarek (approved):

IKarbowiak (approved):

saleor/graphql/product/schema.py
```diff
@@ -336,8 +350,31 @@ def resolve_product(self, info, id=None, slug=None, channel=None, **_kwargs):
...
+            and kwargs["filter"].get("search") is not None
+            and kwargs["filter"]["search"].strip()
+        ):
+            # default to sorting by rank unless specified otherwise
+            if "sort_by" not in kwargs:
+                product_type = info.schema.get_type("ProductOrder")
+                kwargs["sort_by"] = product_type.create_container(
+                    {"direction": "-", "field": ["search_rank", "id"]}
+                )
+
```
maarcingebala, line 377: Maybe we could wrap it with some function to make the resolver a bit easier to read?

maarcingebala (approved):

saleor/graphql/product/schema.py
patrys: Extracted two helper functions for the `if` logic.

maarcingebala (approved):

eCzerniak (approved):
PR is 👍  but dashboard allows to search for white characters which causes an error. 
cc @dominik-zeglen

saleor/order/migrations/0130_save_order_token_in_relation_models.py
```diff
@@ -147,20 +147,9 @@ class Migration(migrations.Migration):
...
-                    (
-                        select constraint_name from information_schema.table_constraints
-                        WHERE table_name = 'order_order_gift_cards'
-                            and constraint_type = 'FOREIGN KEY'
-                            and constraint_name like '%order_order_id'
-                    )
-                );
-                EXECUTE query;
-            END $$;
+            ALTER TABLE order_order_gift_cards
```
cmiacz, line 150: What shall we do with 3.2 environments which already have this migration applied?

cmiacz: cc @NyanKiyoshi

NyanKiyoshi: It's just a simplification of the original migration: nothing to be done about it

saleor/graphql/product/tests/test_product.py
```diff
@@ -4572,22 +4550,23 @@ def test_search_product_by_description_and_name(
...
     }
     response = user_api_client.post_graphql(SEARCH_PRODUCTS_QUERY, variables)
-    content = get_graphql_content(response)
-    data = content["data"]["products"]["edges"]
-    assert len(data) == product_count
+    content = get_graphql_content(response, ignore_errors=True)
+    assert "errors" in content
+    assert (
+        content["errors"][0]["message"]
+        == "Sorting by Rank is available only with searching."
```
cmiacz, line 4564: Tests are failing on cloud build. This error message doesn't match:
https://github.com/saleor/saleor/blob/cfdd78cf7fb4a6814be8266fd42276aebcf424f0/saleor/graphql/product/schema.py#L370

cmiacz (changes requested):

patrys:
@cmiacz Fixed the test and rebased on current main.

PR #9363, opened 2022-03-15
Better permissions

maarcingebala:
- All `PermissionDenied` errors should mention which permissions are required to perform the given action
- Required permissions should be mentioned in the GraphQL description
- To represent all permissions in the same way, new enum values are introduced. They represent permission checks that are based on functions instead of named admin permission scopes.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/core/permissions.py
+++ b/saleor/core/permissions.py
@@ -123,7 +123,7 @@ class InternalPermissions(BasePermissionEnum):
     # Grants access to the owner of the related object. This rule doesn't come with any
     # permission function, as the ownership needs to be defined individually in each
     # case.
-    OWNER = "internal_permissions.owner"
+    IS_OWNER = "internal_permissions.owner"
 
 
 def resolve_internal_permission_fn(perm):
--- a/saleor/core/permissions.py
+++ b/saleor/core/permissions.py
@@ -110,27 +110,27 @@ def is_staff_user(context):
     return is_user(context) and context.user.is_staff
 
 
-class InternalPermissions(BasePermissionEnum):
+class PermissionFunctions(BasePermissionEnum):
     # Grants access to any authenticated app.
-    IS_AUTHENTICATED_APP = "internal_permissions.is_authenticated_app"
+    IS_AUTHENTICATED_APP = "permission_functions.is_authenticated_app"
 
     # Grants access to any authenticated staff user.
-    IS_AUTHENTICATED_STAFF_USER = "internal_permissions.is_authenticated_staff_user"
+    IS_AUTHENTICATED_STAFF_USER = "permission_functions.is_authenticated_staff_user"
 
     # Grants access to any authenticated user.
-    IS_AUTHENTICATED_USER = "internal_permissions.is_authenticated_user"
+    IS_AUTHENTICATED_USER = "permission_functions.is_authenticated_user"
 
     # Grants access to the owner of the related object. This rule doesn't come with any
     # permission function, as the ownership needs to be defined individually in each
     # case.
-    IS_OWNER = "internal_permissions.owner"
+    IS_OWNER = "permission_functions.owner"
 
 
 def resolve_internal_permission_fn(perm):
     fn_map = {
-        InternalPermissions.IS_AUTHENTICATED_APP: is_app,
-        InternalPermissions.IS_AUTHENTICATED_USER: is_user,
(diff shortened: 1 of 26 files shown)


patrys:
Task linked: [SALEOR-6024 Better permissions](https://app.clickup.com/t/2549495/SALEOR-6024)

jbergstroem:
It would be interesting to hear how you see this progressing. I think moving closer to a RBAC-style permission model where different attributes/functioinalities/data models expresses their permission bits and the saleor store admin can assign these to roles as they see fit. I know that allowing clients to update profile pictures has come up more than once as a request (in discussions) that would be much simplified to achieve this way!

Qasem-h (approved):

IKarbowiak (approved):

saleor/core/permissions.py
```diff
@@ -98,6 +98,43 @@ class SitePermissions(BasePermissionEnum):
...
+    AUTHENTICATED_USER = "permission_functions.authenticated_user"
+
+    # Grants access to the owner of the related object. This rule doesn't come with any
+    # permission function, as the ownership needs to be defined individually in each
+    # case.
+    OWNER = "permission_functions.owner"
+
+
+def resolve_internal_permission_fn(perm):
+    fn_map = {
```
L3str4nge, line 130: I would not create dict every time I will call this function. Will be better to move this mapping outside.

saleor/graphql/core/mutations.py
```diff
@@ -98,9 +107,22 @@ class BaseMutation(graphene.Mutation):
...
+    def _validate_permissions(cls, permissions):
+        if not permissions:
+            return
+        if not isinstance(permissions, tuple):
+            raise ImproperlyConfigured(
+                f"Permissions should be a tuple in Meta class: {permissions}"
+            )
+        for p in permissions:
+            if not isinstance(p, Enum):
+                raise ImproperlyConfigured(f"Permission should be an enum: {p}.")
```
L3str4nge, line 120:
```suggestion
                raise ImproperlyConfigured(f"Permissions should be an enum: {p}.")
```

L3str4nge (approved):

fowczarek (approved):

korycins (approved):

saleor/graphql/core/mutations.py
```diff
@@ -316,24 +342,52 @@ def check_permissions(cls, context, permissions=None):
...
-            return True
-        if context.user.has_perms(permissions):
+        all_permissions = permissions or cls._meta.permissions
+        if not all_permissions:
             return True
+
+        permission_fns = [
+            p for p in all_permissions if isinstance(p, PermissionFunctions)
+        ]
+        admin_permissions = [
```
korycins, line 352: Do these names are enough meaningful? 
- `permission_fns` - it is a list of permission that determines access to resources by a different states, right? Like `user is authorized`, `app is authorized`, `customer is authorized`
- `admin_permissions` -  Does it contain a list of our permission enums, like `manage_orders`? Does `admin_` gives us any context here? Personally I thought that `admin_permissions` will be used only for users that are staff but I see that we also check if `app` has `admin_permissions`. Not sure if using `admin_` is ok here, as `App` cannot be an admin, right?


Maybe we could give more context for both of these values?

maarcingebala: You guessed right about both names and their purpose. I didn't have any better idea how to generally call the fact of using "manage_" and "handle_" enums for permission checks and I called them "admin", but you're right this may suggest a physical user while it also applies to apps. I'm happy to accept better ideas for names here.

korycins: `permission_fns` -> maybe something like `access_permissions`? As it contains all possible access permissions for the requestor.
`admin_permissions` -> `system_permissions` or something that will tell us that it contains a list of our built-in permissions.

saleor/core/permissions.py
maarcingebala: Fixed

saleor/graphql/core/mutations.py
maarcingebala: Changed:
permission_fns -> authorization_filters
admin_permissions -> permissions

korycins (approved):

PR #9364, opened 2022-03-15
Cast shipping_method_id to int

L3str4nge:
Cast `root.shipping_method_id` to `int` because passing it as `str` causes unexpected returning `None` from `ShippingMethodByIdLoader`.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/types.py
+++ b/saleor/graphql/order/types.py
@@ -1108,7 +1108,7 @@ class Order(ModelObjectType):
             return listing.then(calculate_price)
 
         shipping_method = ShippingMethodByIdLoader(info.context).load(
-            root.shipping_method_id
+            int(root.shipping_method_id)
         )
         channel = ChannelByIdLoader(info.context).load(root.channel_id)
 


patrys:
Task linked: [SALEOR-5961 [local] checkoutComplete an error is shown](https://app.clickup.com/t/2549495/SALEOR-5961)

tomaszszymanski129 (approved):

maarcingebala:
Isn't it the same issue that I was fixing in this PR https://github.com/saleor/saleor/pull/9340/? It was reported on Slack and I fixed it without opening a task but it seems the task was already there.

L3str4nge:
> Isn't it the same issue that I was fixing in this PR #9340? It was reported on Slack and I fixed it without opening a task but it seems the task was already there.

Seems to be the same. Probably there are other places in the codebase to fix i guess.

maarcingebala:
I was initially adding the fix inside the dataloader but then in the PR we agreed it would be better to fix the issue in some lower layer and to always have an integer instead of string in `root.shipping_method_id` when it refers to a shipping method from Saleor. I thought my fix helps but if not maybe we could dig deeper and see where the string is coming from.

L3str4nge:
The fun fact is that `root.shipping_method_id` is `str` type, but `root.shipping_method.id` is actually `int`. How about to pass `root.shipping_method.id` instead of `shipping_method_id`?

maarcingebala (approved):

IKarbowiak (approved):

PR #9388, opened 2022-03-22
Switch payload ids from DB's to Graphql 

L3str4nge:
Switches all DB ids in email payload to graphql.  For creating events we still need db ids so there is a function to get db id from graphql (kinda revert).

[DOCS PR](https://github.com/saleor/saleor-docs/pull/387)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable, unreleased changes to this project will be documented in this file.
 # Unreleased
 
 ### Breaking changes
+- Convert IDs from DB to Graphql format in Email Payload - #9388 by @L3str4nge
 
 - Migrate order id from int to UUID - #9324 by @IKarbowiak
   - Changed the order `id` changed from `int` to `UUID`, the old ids still can be used
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable, unreleased changes to this project will be documented in this file.
 # Unreleased
 
 ### Breaking changes
-- Convert IDs from DB to Graphql format in Email Payload - #9388 by @L3str4nge
+- Convert IDs from DB to Graphql format in all notification payloads (email plugins and webhook.NOTIFY)- #9388 by @L3str4nge
 
 - Migrate order id from int to UUID - #9324 by @IKarbowiak
   - Changed the order `id` changed from `int` to `UUID`, the old ids still can be used
--- a/saleor/core/utils/__init__.py
+++ b/saleor/core/utils/__init__.py
@@ -3,7 +3,6 @@ import socket
 from typing import TYPE_CHECKING, Optional, Type, Union
 from urllib.parse import urljoin
 
-import graphene
 from babel.numbers import get_territory_currencies
 from celery.utils.log import get_task_logger
 from django.conf import settings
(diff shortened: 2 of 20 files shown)


saleor/core/utils/__init__.py
```diff
@@ -156,3 +157,19 @@ def generate_unique_slug(
...
+
+
+def graphql_id(instance):
+    class_name = instance.__class__.__name__
+    if instance.id is None:
+        return None
+    return graphene.Node.to_global_id(class_name, instance.id)
+
+
+def get_id_from_graphql_id(_id):
```
IKarbowiak, line 169: Why not `id` instead of `_id`?

IKarbowiak: And do we really need this method? Cannot we use `from_global_id_or_error` from `saleor/saleor/graphql/core/utils/__init__.py`?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable, unreleased changes to this project will be documented in this file.
 # Unreleased
 
 ### Breaking changes
+- Convert IDs from DB to Graphql format in Email Payload - #9388 by @L3str4nge
```
korycins, line 9: It is not exacly Email payload. It is more like generic payload for any type of notification. You can have an app that sends a text messages, or sends notification on any type of comunicator.

korycins: (I see that task has `Email Payload` but maybe it would be good to rename it to something more generic, as it has an impact on all email plugins, and Notify webhook)

saleor/core/utils/__init__.py
L3str4nge: Fixed but still need additional function. It uses `from_global_id_or_error` but I need to return None in some scenarios. To avoid checking payload XX times in the logic that's why I've created such a function. Also moved to `saleor/saleor/graphql/core/utils/__init__.py` as they are part of grapqhl. [code](https://github.com/saleor/saleor/pull/9388/commits/723baa75e4df8f222e86ec5895acf1995f57cbba#diff-49b45f224daeaed23cfc4a19f956c588071fae3f97494be88617f04894be099bR186)

saleor/order/notifications.py
```diff
@@ -243,7 +244,8 @@ def get_default_order_payload(order: "Order", redirect_url: str = ""):
     order_payload = model_to_dict(order, fields=ORDER_MODEL_FIELDS)
     order_payload.update(
         {
-            "number": order.id,
+            "id": to_global_id_or_none(order),
+            "number": to_global_id_or_none(order),
```
korycins, line 248: This is incorrect. `number` field should return a number of a given order.

mzawad: This is the same thing actually:
```
@staticmethod
    def resolve_number(root: models.Order, _info):
        return str(root.pk)
```

L3str4nge: This is the same thing actually:
```
@staticmethod
    def resolve_number(root: models.Order, _info):
        return str(root.pk)
```
in `saleor.graphql.order.types.Order`

korycins: It works because we didn't merge conversion from int to UUID yet.

For your case, it should be:
```
id: base64("Order", order.id)
number: order.id (as it is still int value)

Currently it doesn't make sense as both fields return the same value.

maarcingebala: I merged the ID migration today so this PR can be updated and we can see if it works as expected now.

maarcingebala (approved):

PR #9393, opened 2022-03-23
Custom prices

IKarbowiak:
Allow defining checkout items custom prices.

Docs: https://github.com/saleor/saleor-docs/pull/390

- Add `HANDLE_CHECKOUTS` permission.
  - The new permission is granted to all apps and groups that already have `MANAGE_CHECKOUT`.
- The `CheckoutLineInput` is extended with the optional price field `price: PositiveDecimal`, so the custom price could be set in the following mutations:
  - `checkoutCreate`
  - `checkoutLinesAdd` - when adding a variant that is already added to checkout, the already existing line is overridden - the quantity is incremented and the price is updated.
  - `checkoutLinesUpdate` - overrides the existing line with the price provided in the mutation
- Update all places where the checkout price is calculated to use `price_override` if set.
- Extend populate db with checkout with custom prices.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-6089 Custom prices when adding items to checkout](https://app.clickup.com/t/2549495/SALEOR-6089)

Qasem-h (approved):

L3str4nge (approved):

korycins (commented):
@IKarbowiak Question. How do we want to handle such a case? (Or just confirm that we are able to do that)
- User creates checkout with item A
- User adds an item B
- A+B means that for some reason we want to reduce the value of each of these lines
- App with `Handle_checkouts` recieve webhook checkout_updated (OR storefront triggers notification to app, that we want to have override prices), and because of the above condition it wants to change prices for these lines but without touching their quantity.  Is it possible? (Maybe I am wrong, but not sure if this is possible currently).

IKarbowiak:
@korycins Thank you for mentioning this case. The `CheckoutLinesUpdate` is for that, but I need to make some changes as it requires the `CheckoutLineInput` where `quantity` is required. Because of that, I'm changing the PR into a draft.

saleor/graphql/checkout/mutations/checkout_lines_update.py
```diff
@@ -1,15 +1,44 @@
...
 from ..types import Checkout
+from .checkout_create import CheckoutLineInput
 from .checkout_lines_add import CheckoutLinesAdd
 from .utils import check_lines_quantity
 
 
+class CheckoutLineUpdateInput(CheckoutLineInput):
+    quantity = graphene.Int(
+        required=False,
+        description="The number of items purchased. Required for not app requestor.",
```
maarcingebala, line 18: This might sound more natural:

The number of items purchased. Optional for apps, required for any other users.

korycins (approved):

PR #9394, opened 2022-03-23
Add Webhook payload via graphql subscriptions

jakubkuc:
I want to merge this change because it is first phase of adding subscription payload feature. 
It will enable creating webhook payloads using subscription query, this will let user define what payload will be send using graphl subscription - which can be build via graphql API. 

This is first phase - so at the moment only async webhooks are updated. Sync events and notify events will be done in next phases.
It does not affect he current way of generating payloads.
I tried to make the code in a way that will allow us to easily let go of the old way of generating payloads in the future saleor versions. 

Here is a small example steps with snippets that will show how the whole process looks like for subscription payloads.
I'll use simple ProductUpdated event for those examples.

1. Create subscription using graphql API (graphql playground, insomnia or any other graphql interface can be used to do it):
```
subscription{
  event{
    ...on ProductUpdated{
      product{
        id
        name
      }
    }
  }
}
```
2. Create new or upgrade existing webhook using our mutations with subscription passed in `query` field as string : 
```
mutation {
  webhookCreate(
    input: {
      app: "QXBwOjE="
      name: "test_product_updated_subscription"
      targetUrl: "<webhook_url>"
      asyncEvents: [PRODUCT_UPDATED]
      query: "subscription{event{...on ProductUpdated{product{id name}}}}"
    }
  ) {
    errors {
      field
      message
    }
    webhook {
      id
    }
  }
}
```
3. The payload in webhook will look as follows: 
```
[
  {
    "product": {
      "id": "UHJvZHVjdDo3Ng==",
      "name": "Coconut Juice"
    },
    "meta": {
      "issued_at": "2022-03-23T14:37:27.084275+00:00",
      "version": "dev",
      "issuing_principal": {
        "id": "<id>",
        "type": "user"
      }
    }
  }
]
```
Meta field is always generated additionally.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/plugins/webhook/subscription_webhooks/fixtures.py
```diff
@@ -0,0 +1,880 @@
...
+            id
+          }
+        }
+      }
+    }
+"""
+
+
+@pytest.fixture
+def subscription_product_variant_back_in_stock_webhook(app):
```
korycins, line 195: Looks like alls fixtures look very similar. You could do something like this:
```
@pytest.fixture
def subscription_webhook(app):
    def fun(query, event_type, name="Subscription"):
        webhook = Webhook.objects.create(
            name=name,
            app=app,
            target_url="http://www.example.com/any",
            subscription_query=query,
        )
        webhook.events.create(
            event_type=event_type
        )
        return webhook
@pytest.fixture
def subscription_product_variant_back_in_stock_webhook(subscription_webhook):
    return subscription_webhook(PRODUCT_VARIANT_BACK_IN_STOCK_SUBSCRIPTION_QUERY, WebhookEventAsyncType.PRODUCT_VARIANT_BACK_IN_STOCK)

``` 

This should be re-usable between all events without duplicate a lot of code.

saleor/webhook/subscription_payload.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    graphql_middleware = graphene_settings.MIDDLEWARE
+    graphql_middleware = MiddlewareManager(
+        instantiate_middleware(graphql_middleware), wrap_in_promise=False
+    )
+    app_id = app.pk if app else None
+
+    context.app = app  # type: ignore
+
+    results = document.execute(
```
korycins, line 83: @fowczarek  can you confirm? Should we somehow define the replica that we want to use it here?

saleor/graphql/webhook/mutations.py
```diff
@@ -40,6 +56,10 @@ class WebhookCreateInput(graphene.InputObjectType):
         description="The secret key used to create a hash signature with each payload.",
         required=False,
     )
+    query = graphene.String(
+        description="Subscription query used to define a webhook payload.",
```
korycins, line 60: This should be marked as a preview_feature.

saleor/graphql/webhook/subscription_types.py
```diff
@@ -0,0 +1,407 @@
...
+            WebhookEventAsyncType.DRAFT_ORDER_CREATED: DraftOrderCreated,
+            WebhookEventAsyncType.DRAFT_ORDER_UPDATED: DraftOrderUpdated,
+            WebhookEventAsyncType.DRAFT_ORDER_DELETED: DraftOrderDeleted,
+            WebhookEventAsyncType.PRODUCT_CREATED: ProductCreated,
+            WebhookEventAsyncType.PRODUCT_UPDATED: ProductUpdated,
+            WebhookEventAsyncType.PRODUCT_DELETED: ProductDeleted,
+            WebhookEventAsyncType.PRODUCT_VARIANT_CREATED: ProductVariantCreated,
+            WebhookEventAsyncType.PRODUCT_VARIANT_UPDATED: ProductVariantUpdated,
+            WebhookEventAsyncType.PRODUCT_VARIANT_OUT_OF_STOCK: ProductVariantOutOfStock,  # noqa
+            WebhookEventAsyncType.PRODUCT_VARIANT_BACK_IN_STOCK: ProductVariantBackInStock,  # noqa
```
fowczarek, line 373: IMO we shouldn't use `noqa` to mute line length errors.
@maarcingebala what do you think?

saleor/webhook/subscription_payload.py
```diff
@@ -0,0 +1,116 @@
...
+def initialize_context() -> HttpRequest:
+    """Prepare a request object for webhook subscription.
+
+    It creates a dummy request object and initialize middleware on it. It is required
+    to process a request in the same way as API logic does.
+    return: HttpRequest
+    """
+    handler = BaseHandler()
+    context = RequestFactory().request()
+    handler.load_middleware()
```
fowczarek, line 39: Should we load middleware for generating payloads?

saleor/webhook/subscription_payload.py
```diff
@@ -0,0 +1,116 @@
...
+    graphql_backend = get_default_backend()
+    ast = parse(subscription_query)  # type: ignore
+    document = graphql_backend.document_from_string(
+        schema,
+        ast,
+    )
+    graphql_middleware = graphene_settings.MIDDLEWARE
+    graphql_middleware = MiddlewareManager(
+        instantiate_middleware(graphql_middleware), wrap_in_promise=False
+    )
```
fowczarek, line 78: I'm not sure we should load graphene middleware here. As far as I know, we want to don't use them. 
@maarcingebala @patrys

saleor/webhook/subscription_payload.py
fowczarek: Nope. Replicas are not related to GraphQLView.

saleor/webhook/subscription_payload.py
```diff
@@ -0,0 +1,116 @@
...
+        logger.warning(
+            "Subscription did not return a payload.",
+            extra={"query": subscription_query, "app": app_id},
+        )
+
+    payload_instance = payload[0]
+
+    event_payload = payload_instance.data.get("event")
+
+    # TODO Make sure that we handle missing permissions in proper way.
```
fowczarek, line 109: Should we resolve it or remove it?

saleor/webhook/subscription_payload.py
```diff
@@ -0,0 +1,116 @@
...
+
+def validate_subscription_query(query: str) -> bool:
+    from ..graphql.api import schema
+
+    graphql_backend = get_default_backend()
+    try:
+        graphql_backend.document_from_string(schema, query)
+    except (ValueError, GraphQLSyntaxError):
+        return False
+    return True
```
fowczarek, line 27: We should also validate that the query contains subscriptions only.

saleor/graphql/webhook/subscription_types.py
maarcingebala: Why? There is no easy way to format this particular line.

saleor/graphql/webhook/subscription_types.py
```diff
@@ -0,0 +1,407 @@
...
+        return types.get(object_type)
+
+    @classmethod
+    def resolve_type(cls, instance, info):
+        type_str, _ = instance
+        return cls.get_type(type_str)
+
+
+class Subscription(ObjectType):
+    event = graphene.Field(Event)
```
maarcingebala, line 403: Missing description?

saleor/webhook/subscription_payload.py
maarcingebala: Why do we load it? Isn't it old code from when we have Graphene middlewares?

saleor/plugins/webhook/tasks.py
```diff
@@ -82,17 +91,101 @@ def _get_webhooks_for_event(event_type, webhooks=None):
...
+            subscribable_object=subscribable_object,
+            subscription_query=webhook.subscription_query,
+            context=context,
+            app=webhook.app,
+        )
+        if not data:
+            # FIXME build better way of handling such cases
+            continue
+        if len(data) == 1 and "__typename" in data:
+            continue
```
maarcingebala, line 127: Maybe a line comment here why we add this `if`, to explain what case is this.

saleor/webhook/subscription_payload.py
```diff
@@ -0,0 +1,116 @@
...
+
+    results = document.execute(
+        allow_subscriptions=True,
+        root=(event_type, subscribable_object),
+        context=context,
+        middleware=graphql_middleware,
+    )
+    if hasattr(results, "errors"):
+        logger.warning(str(results.errors))
+        logger.warning(
```
L3str4nge, line 91: I would wrap it into single warning

saleor/webhook/subscription_payload.py
```diff
@@ -0,0 +1,116 @@
...
+    payload = []  # type: ignore
+    results.subscribe(payload.append)
+
+    if not payload:
+        logger.warning(
+            "Subscription did not return a payload.",
+            extra={"query": subscription_query, "app": app_id},
+        )
+
+    payload_instance = payload[0]
```
L3str4nge, line 105: It will raise error here after logged warning

saleor/webhook/subscription_payload.py
patrys: We don't use middleware, it's expensive.

saleor/graphql/webhook/subscription_types.py
jakubkuc: Actually there is. @fowczarek suggested that putting key in parenthesis will let us put it in new line. I tried and it works.

maarcingebala: 👍

saleor/webhook/subscription_payload.py
korycins: All our resolvers assume that Django middlewares have been called and we have access to data like `request.discounts`, `request.plugins` etc. We load middlewares here to have the same input resolver data as we have from API calls. The second way would be manually attaching all dependencies that are added by middlewares.

saleor/webhook/subscription_payload.py
korycins: This should be dropped. Subscription logic was added at the moment when we were using graphql middleware also.

jakubkuc: True, I'll drop it with nearest commit

saleor/plugins/webhook/subscription_webhooks/fixtures.py
jakubkuc: brilliant. Fixed it

saleor/graphql/webhook/mutations.py
jakubkuc: done.

.isort.cfg
```diff
@@ -1,2 +1,2 @@
...
+known_third_party = Adyen,PIL,authorizenet,babel,before_after,boto3,botocore,braintree,celery,cryptography,dateutil,dj_database_url,dj_email_url,django,django_cache_url,django_countries,django_filters,django_measurement,django_prices,django_prices_openexchangerates,django_prices_vatlayer,draftjs_sanitizer,faker,freezegun,google,google_measurement_protocol,graphene,graphene_django,graphql,graphql_relay,html2text,html_to_draftjs,i18naddress,jaeger_client,jwt,kombu,lxml,markdown,measurement,micawber,mptt,oauthlib,openpyxl,opentracing,petl,phonenumber_field,phonenumbers,pkg_resources,posuto,prices,promise,pybars,pytest,pythonjsonlogger,pytimeparse,pytz,razorpay,requests,rx,sendgrid,sentry_sdk,storages,stripe,urllib3,uvicorn,versatileimagefield,weasyprint
```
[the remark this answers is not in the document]
jakubkuc: yes, mistake during rebase. fixed.

saleor/graphql/webhook/subscription_types.py
```diff
@@ -0,0 +1,407 @@
+import graphene
+from graphene import AbstractType, ObjectType, Union
+from rx import Observable
+
+from ...webhook.event_types import WebhookEventAsyncType
+from ..channel import ChannelContext
+
+
+class OrderBase(AbstractType):
+    order = graphene.Field("saleor.graphql.order.types.Order")
```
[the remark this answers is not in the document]
jakubkuc: added

saleor/webhook/subscription_payload.py
jakubkuc: done.

saleor/webhook/subscription_payload.py
jakubkuc: additional validation and tests added.

saleor/graphql/webhook/subscription_types.py
jakubkuc: added

saleor/plugins/webhook/tasks.py
jakubkuc: after some investigation and consultations with @korycins we decided to add some tests and drop it as no longer viable

saleor/webhook/subscription_payload.py
jakubkuc: done

saleor/webhook/subscription_payload.py
jakubkuc: yes, good catch. It should've return after logging an error. Fixed

saleor/webhook/subscription_payload.py
jakubkuc: we decided to remove it from MVP, added a task for to deal with it in the future development of this feature

saleor/plugins/webhook/tasks.py
korycins: Yes, I think that this is something that I left accidentally after cleaning up the implementation of POC

saleor/graphql/webhook/mutations.py
```diff
@@ -40,6 +56,11 @@ class WebhookCreateInput(graphene.InputObjectType):
         description="The secret key used to create a hash signature with each payload.",
         required=False,
     )
+    query = graphene.String(
+        description=f"Subscription query used to define a webhook payload. "
+        f"{PREVIEW_FEATURE}",
```
maarcingebala, line 61: We also need ADDED_IN_32.

saleor/graphql/webhook/subscription_types.py
```diff
@@ -0,0 +1,476 @@
...
+        return types.get(object_type)
+
+    @classmethod
+    def resolve_type(cls, instance, info):
+        type_str, _ = instance
+        return cls.get_type(type_str)
+
+
+class Subscription(ObjectType):
+    event = graphene.Field(Event, description="Look up subscription event.")
```
maarcingebala, line 472: ADDED_IN_32 + preview.

saleor/graphql/webhook/mutations.py
```diff
@@ -147,6 +171,10 @@ class WebhookUpdateInput(graphene.InputObjectType):
     secret_key = graphene.String(
         description="Use to create a hash signature with each payload.", required=False
     )
+    query = graphene.String(
+        description="Subscription query used to define a webhook payload.",
```
maarcingebala, line 175: ADDED_IN_32 + preview

saleor/graphql/webhook/mutations.py
```diff
@@ -4,13 +4,29 @@
...
 
 
+def validate_query(query):
+    if not query:
+        return
+    is_valid = validate_subscription_query(query)
+    if not is_valid:
+        raise ValidationError(
+            {
+                "query": ValidationError(
```
korycins, line 22: The one thing which I am worried about is that we don't provide a lot of info. We just say that something was wrong and you need to figure it out. We could somehow wrap a msg from graphql validator (something similar that we get when we will try to send an incorrect query).
Maybe it is not the scope of this PR but it is worth mentioning it.
@maarcingebala  what do you think?

saleor/plugins/webhook/tasks.py
```diff
@@ -82,17 +91,102 @@ def _get_webhooks_for_event(event_type, webhooks=None):
...
+    for webhook in webhooks:
+        data = generate_payload_from_subscription(
+            event_type=event_type,
+            subscribable_object=subscribable_object,
+            subscription_query=webhook.subscription_query,
+            context=context,
+            app=webhook.app,
+        )
+        if not data:
+            logger.warning(
```
korycins, line 124: Instead of single warning, don't you think, that we should build a EventDelivery object, that would allow user to see, that something was wrong with the query? Currently, they won't know that something was wrong before checking logs..

saleor/graphql/webhook/mutations.py
maarcingebala: Definitely would be useful to give more info but don't have to be done in this PR.

saleor/plugins/webhook/tasks.py
jakubkuc: You think about only single EventDelivery object that will have FAILED status - that will indicate that something went wrong but with no more info? Because creating related EventDeliveryAttempt that will have some more info about failure seems a bit off to me - cause it would not be a proper delivery attempt (no response and request headers etc.). And we need to keep in mind that creating EventDelivery would allow us to use `EventDeliveryRetry` mutation on that delivery.
I'll try to think of a better way to signal to user that payload generation went wrong.

saleor/webhook/subscription_payload.py
```diff
@@ -0,0 +1,131 @@
+from typing import Any, Dict, Optional
```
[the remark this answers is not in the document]
maarcingebala: Agree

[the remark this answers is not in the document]
jakubkuc: I agree.moved.

saleor/plugins/webhook/tasks.py
jakubkuc: As we agreed - this is something to be further discussed and done in future PRs.

saleor/graphql/webhook/mutations.py
jakubkuc: As we agreed - this is something to be further discussed and done in future PRs.

saleor/webhook/subscription_payload.py
jakubkuc: @fowczarek django

saleor/plugins/webhook/subscription_webhooks/fixtures.py
```diff
@@ -0,0 +1,868 @@
+import pytest
```
IKarbowiak, line 1: It seems like the whole `subscription_webhooks` directory should be in `tests` as there is only test-related stuff.

saleor/graphql/webhook/subscription_payload.py
```diff
@@ -0,0 +1,131 @@
...
+        return False
+    if not check_document_is_single_subscription(document):
+        return False
+    return True
+
+
+def check_document_is_single_subscription(document: GraphQLDocument) -> bool:
+    """Check if document contains only a single subscription definition.
+
+    # Only fragments and single subscription definition are allowed.
```
IKarbowiak, line 33: `#` not needed.

saleor/graphql/webhook/subscription_payload.py
```diff
@@ -0,0 +1,131 @@
...
+            return False
+    if len(subscriptions) != 1:
+        result = False
+    return result
+
+
+def initialize_context() -> HttpRequest:
+    """Prepare a request object for webhook subscription.
+
+    # It creates a dummy request object and initialize middleware on it. It is required
```
IKarbowiak, line 55: Same here: `#` not needed

saleor/plugins/webhook/subscription_webhooks/fixtures.py
jakubkuc: It is indeed. Since `subscription_payload.py` was moved there are only test related files here. I'll move them/

IKarbowiak (approved):

maarcingebala (approved):
@jakubkuc Please add changelog.

saleor/graphql/webhook/mutations.py
```diff
@@ -1,16 +1,33 @@
 import graphene
 from django.core.exceptions import ValidationError
 
+from saleor.graphql.webhook.subscription_payload import validate_subscription_query
```
maarcingebala, line 4: 🚔   Use relative imports. 🚔

saleor/graphql/webhook/subscription_payload.py
```diff
@@ -0,0 +1,131 @@
...
+        elif isinstance(definition, OperationDefinition):
+            if definition.operation == "subscription":
+                subscriptions.append(definition)
+            else:
+                return False
+        else:
+            return False
+    if len(subscriptions) != 1:
+        result = False
+    return result
```
tomaszszymanski129, line 49: `result` var introduces more complexity than it's worth imo.
```suggestion
    subscriptions = []
    for definition in document.document_ast.definitions:
        if isinstance(definition, FragmentDefinition):
            pass
        elif isinstance(definition, OperationDefinition):
            if definition.operation == "subscription":
                subscriptions.append(definition)
            else:
                return False
        else:
            return False

    return len(subscriptions) == 1
```

tomaszszymanski129 (approved):

saleor/graphql/webhook/subscription_payload.py
jakubkuc: nice, I like it :) fixed

tomaszszymanski129:
@jakubkuc conflicts in changelog

fowczarek (approved):

Issue #7856, opened 2021-08-17
Building Saleor API container fails on collectstatic.


vetostream:
### What I'm trying to achieve
Build saleor platform on local machine.

### Steps to reproduce the problem
```
 => ERROR [stage-1 9/9] RUN SECRET_KEY=dummy STATIC_URL=/static/ python3 manage.py collectstatic --no-input                                                                                          0.6s
------                                                                                                                                                                                                    
 > [stage-1 9/9] RUN SECRET_KEY=dummy STATIC_URL=/static/ python3 manage.py collectstatic --no-input:                                                                                                     
#19 0.577 Traceback (most recent call last):                                                                                                                                                              
#19 0.577   File "manage.py", line 10, in <module>                                                                                                                                                        
#19 0.577     execute_from_command_line(sys.argv)                                                                                                                                                         
#19 0.577   File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line                                                                      
#19 0.577     utility.execute()
#19 0.577   File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 377, in execute
#19 0.577     django.setup()
#19 0.577   File "/usr/local/lib/python3.8/site-packages/django/__init__.py", line 24, in setup
#19 0.577     apps.populate(settings.INSTALLED_APPS)
#19 0.577   File "/usr/local/lib/python3.8/site-packages/django/apps/registry.py", line 91, in populate
#19 0.577     app_config = AppConfig.create(entry)
#19 0.577   File "/usr/local/lib/python3.8/site-packages/django/apps/config.py", line 116, in create
#19 0.577     mod = import_module(mod_path)
#19 0.577   File "/usr/local/lib/python3.8/importlib/__init__.py", line 127, in import_module
#19 0.577     return _bootstrap._gcd_import(name[level:], package, level)
#19 0.577   File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
#19 0.577   File "<frozen importlib._bootstrap>", line 991, in _find_and_load
#19 0.577   File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
#19 0.577   File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
#19 0.577   File "<frozen importlib._bootstrap_external>", line 843, in exec_module
#19 0.577   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
[...]

Issue #7431, opened 2021-06-04
ERROR: Service 'api' failed to build : The command '/bin/sh -c SECRET_KEY=dummy STATIC_URL=${STATIC_URL} python3 manage.py collectstatic --no-input' returned a non-zero code: 1


ahmadoji01:
### What I'm trying to achieve
Install Saleon based on the provided documentation

### Steps to reproduce the problem

1. I run the `docker-compose run --rm api python3 manage.py migrate`

### What I expected to happen
I expect the installation to be successful

### Screenshots and logs
Here is the log I received after running that command
```
Step 16/21 : ARG STATIC_URL
 ---> Running in b2c51524839b
Removing intermediate container b2c51524839b
 ---> ec3e96b7fa07
Step 17/21 : ENV STATIC_URL ${STATIC_URL:-/static/}
 ---> Running in ceec7df66b6b
Removing intermediate container ceec7df66b6b
 ---> c434fbc6f03a
Step 18/21 : RUN SECRET_KEY=dummy STATIC_URL=${STATIC_URL} python3 manage.py collectstatic --no-input
 ---> Running in ec65e2dd4120
Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 377, in execute
    django.setup()
  File "/usr/local/lib/python3.8/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/python3.8/site-packages/django/apps/registry.py", line 91, in populate
    app_config = AppConfig.create(entry)
  File "/usr/local/lib/python3.8/site-packages/django/apps/config.py", line 116, in create
    mod = import_module(mod_path)
  File "/usr/local/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
[...]

PR #9397, opened 2022-03-24
fix: use non-binary psycopg2 distribution

jbergstroem:
# Impact

In aarch64 distributions the bundled version of `libpq` in `psycopg2-binary` is too old,
leading to connection / auth issues, made visible on M1 Macs during last year or so (found out while my new teammates joining with these laptops):
> pg_connect(): Unable to connect to PostgreSQL server: SCRAM authentication requires libpq version 10 or above.

Thi PR fixes the (upstream) issue by compiling against the native `libpq` bundled with debian
(all architectures). Since its already available in the base distribution of `python:3.9`
and subsequently `python:3.10`, all we have to do is let it compile.

Impact:
- a newer libpq will be used / linked against
- a few seconds of additional compile time while building dependencies

Refs: https://github.com/psycopg/psycopg2/issues/1286

# Pull Request Checklist

* [ ] The changes are tested


jbergstroem:
Hoping this can put a close to #7431 and #7856.

maarcingebala:
We agreed that it's a useful change and we want to merge it, we'll have to update the docs for Windows users probably.

maarcingebala:
Thanks for your contribution @jbergstroem 🥇

PR #9406, opened 2022-03-25
Make OIDC plugin public

korycins:
This PR makes our OIDC plugin public. It is a port of changes from private repo. 
Docs for the plugin can be found here: https://docs.saleor.io/docs/3.x/developer/available-plugins/openid-connect

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/plugins/base_plugin.py
```diff
@@ -137,7 +137,7 @@ def __str__(self):
     #  Authenticate user which should be assigned to the request.
     #
     #  Overwrite this method if the plugin handles authentication flow.
-    authenticate_user: Callable[[WSGIRequest], Union["User", NoneType]]
+    authenticate_user: Callable[[WSGIRequest, Any], Union["User", NoneType]]
```
patrys, line 140: Any?

korycins: All definitions of methods in BasePlugin have `previous_value` declared as `Any`. I added it here also as it was missing previously.

patrys: I don't think `Any` is the correct type in any of those cases? Isn't previous value always the same type as returned value?

Qasem-h (approved):

saleor/plugins/openid_connect/utils.py
```diff
@@ -0,0 +1,571 @@
...
+    if not keys:
+        logger.warning("List of JWKS keys is empty")
+    return keys
+
+
+def get_jwks_keys_from_cache_or_fetch(jwks_url: str) -> dict:
+    jwks_keys = cache.get(JWKS_KEY)
+    if jwks_keys is None:
+        jwks_keys = fetch_jwks(jwks_url)
+        cache.set(JWKS_KEY, jwks_keys, JWKS_CACHE_TIME)
```
L3str4nge, line 96: I don't think it's good place to set keys into cache. `fetch_jwsk(url, cache=True)` might be better. I think getter method should not setting anything.

patrys (approved):

cmiacz (approved):

L3str4nge (approved):

saleor/plugins/openid_connect/plugin.py
```diff
@@ -0,0 +1,451 @@
...
+
+from authlib.common.errors import AuthlibBaseError
+from authlib.integrations.requests_client import OAuth2Session
+from django.core import signing
+from django.core.exceptions import ValidationError
+from django.core.handlers.wsgi import WSGIRequest
+from jwt import ExpiredSignatureError, InvalidTokenError
+from requests import HTTPError, PreparedRequest
+
+from saleor.account.models import User
```
maarcingebala, line 12: 🚔  Use relative imports. 🚔

saleor/plugins/openid_connect/utils.py
```diff
@@ -0,0 +1,577 @@
...
+            user, refresh_token, csrf_token, owner
+        )
+        tokens["csrf_token"] = csrf_token
+    return tokens
+
+
+def validate_refresh_token(refresh_token, data):
+    csrf_token = data.get("csrfToken")
+    if not refresh_token:
+        raise ValidationError(
```
maarcingebala, line 451: IMO this shouldn't return an ValidationError but some other class, but I guess we don't have time for this now.

saleor/plugins/openid_connect/plugin.py
```diff
@@ -0,0 +1,451 @@
...
+
+        user = get_or_create_user_from_payload(
+            parsed_id_token, self.config.authorization_url
+        )
+
+        user_permissions = []
+        if self.config.use_scope_permissions:
+            scope = token_data.get("scope")
+            user_permissions = self._use_scope_permissions(user, scope)
+            if not user.is_staff and bool("saleor:staff" in scope or user_permissions):
```
maarcingebala, line 275: `saleor:staff` could be moved to some constant. It's being used in a number of places and it's a hardcoded string everywhere.

manojLondhe:
In which Saleor release this will be out?

patrys:
@manojLondhe It should be part of the upcoming 3.2.

toastedtoast:
3.2 is not mentioned in the roadmap yet. [https://github.com/orgs/saleor/projects/3/views/1](https://github.com/orgs/saleor/projects/3/views/1)
Any estimate when the first release for 3.2 may happen?

PR #9428, opened 2022-03-29
Include permissions in schema descriptions of protected fields

maarcingebala:
- Add `PermissionsField` which accepts a list of `permissions` and automatically generates descriptions for them.
- From now on the `ConnectionField` inherits from `PermissionsField`, so you can pass `permissions` directly in the connection field declaration.
- Split `saleor/core/permissions.py` into multiple modules.
- No need to use permission decorators, except for a few custom cases where permission checks involve more logic or involve AND operator for conditions, e.g: "is app" and "has specific permission".

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/core/permissions/__init__.py
+++ b/saleor/core/permissions/__init__.py
@@ -1,0 +1,111 @@
+# flake8: noqa
+
+from django.contrib.auth import get_user_model
+
+from .auth_filters import (
+    AuthorizationFilters,
+    is_app,
+    is_staff_user,
+    is_user,
+    resolve_authorization_filter_fn,
+)
+from .enums import (
+    PERMISSIONS_ENUMS,
+    AccountPermissions,
+    AppPermission,
+    BasePermissionEnum,
+    ChannelPermissions,
+    CheckoutPermissions,
+    DiscountPermissions,
+    GiftcardPermissions,
+    MenuPermissions,
+    OrderPermissions,
+    PagePermissions,
+    PageTypePermissions,
+    PaymentPermissions,
+    PluginsPermissions,
+    ProductPermissions,
+    ProductTypePermissions,
+    ShippingPermissions,
+    SitePermissions,
+    get_permission_names,
+    get_permissions,
+    get_permissions_codename,
+    get_permissions_enum_dict,
+    get_permissions_enum_list,
+    get_permissions_from_codenames,
+    get_permissions_from_names,
+    split_permission_codename,
+)
+
+
+def one_of_permissions_or_auth_filter_required(context, permissions):
+    """Determine whether user or app has rights to perform an action.
+
+    The `context` parameter is the Context instance associated with the request.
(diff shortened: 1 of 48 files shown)


IKarbowiak (approved):

saleor/core/permissions/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+
+    authorization_filters = [
+        p for p in permissions if isinstance(p, AuthorizationFilters)
+    ]
+    permissions = [p for p in permissions if not isinstance(p, AuthorizationFilters)]
+
+    granted_by_permissions = False
+    granted_by_authorization_filters = False
+
+    # TODO: move this function from graphql to core
```
korycins, line 58: Should be solved in this PR?

saleor/core/permissions/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+                auth_filters_results.append(bool(res))
+        granted_by_authorization_filters = any(auth_filters_results)
+
+    return granted_by_permissions or granted_by_authorization_filters
+
+
+def permission_required(requestor, perms):
+    User = get_user_model()
+    if isinstance(requestor, User):
+        if requestor.has_perms(perms):
```
korycins, line 90: `return  requestor.has_perms(perms)` ?

korycins: Then, you will also not need `return False` at the end of function

saleor/graphql/schema.graphql
```diff
@@ -223,7 +255,9 @@ type Query {
     last: Int
   ): CollectionCountableConnection
 
-  """Look up a product by ID."""
+  """
+  Look up a product by ID. Requires one of the following permissions to include the unpublished items: MANAGE_ORDERS, MANAGE_DISCOUNTS, MANAGE_PRODUCTS.
```
korycins, line 259: `to include the unpublished items` what does it mean? Is it ok? I thought that this perm could be required to get an access to `product`, but maybe I am missing something

saleor/core/permissions/__init__.py
maarcingebala: I'll do it separately as this PR is already large.

saleor/graphql/schema.graphql
maarcingebala: If a product is published it will be returned. If not, you need those perms to get it.

saleor/core/permissions/__init__.py
maarcingebala: It was like that before but we can change it as you suggest. There is `return False` at the end.

maarcingebala: Fixed

korycins (approved):

eCzerniak (approved):

PR #9442, opened 2022-03-31
Don't rely on counting objects when reindexing

patrys:
Also try to limit memory consumption in Celery eager mode.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/core/search_tasks.py
+++ b/saleor/core/search_tasks.py
@@ -1,3 +1,5 @@
+from typing import List
+
 from celery.utils.log import get_task_logger
 
 from ..account.models import User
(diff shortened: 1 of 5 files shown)


saleor/core/search_tasks.py
```diff
@@ -13,94 +15,101 @@
...
+BATCH_SIZE = 1000
 
 
 @app.task
-def set_user_search_document_values(total_count, updated_count):
-    qs = User.objects.filter(search_document="").prefetch_related("addresses")[
-        :BATCH_SIZE
-    ]
+def set_user_search_document_values(updated_count: int = 0) -> None:
+    qs = list(
```
mzawad, line 23: I think we should rename `qs` to something different as it's not queryset anymore after `list` casting. I was a little bit confused few lines below when I saw `if len(qs) < BATCH_SIZE:`. I wanted to write that we should use `.count()` instead but then I saw it's not a queryset :)

L3str4nge, line 23: I think we should rename qs to something different as it's not queryset anymore after list casting. I was a little bit confused few lines below when I saw if len(qs) < BATCH_SIZE:. I wanted to write that we should use .count() instead but then I saw it's not a queryset :)

IKarbowiak (approved):

fowczarek (approved):

saleor/core/search_tasks.py
patrys: I'll rename it but I'd also never use `.count()` on a QuerySet that is going to be evaluated as it would hit the database with an additional `COUNT()` query.

L3str4nge (approved):

maarcingebala (approved):

PR #9444, opened 2022-04-01
Add main logic for new handler in manager and webhook plugin

korycins:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


Qasem-h (approved):

saleor/webhook/payloads.py
```diff
@@ -1069,3 +1069,55 @@ def generate_excluded_shipping_methods_for_checkout_payload(
...
+    action_value = (
+        quantize_price(payment_data.action_value, payment.currency)
+        if payment_data.action_value
+        else None
+    )
+
+    order_id = payment.order_id
+    graphql_order_id = (
+        graphene.Node.to_global_id("Order", order_id) if order_id else None
+    )
```
maarcingebala, line 1090: On `main` we recently added `to_global_id_or_none`, you may consider using it here (or we can add later before merging feature branch to main).

saleor/webhook/payloads.py
```diff
@@ -1069,3 +1069,55 @@ def generate_excluded_shipping_methods_for_checkout_payload(
...
+
+    checkout_id = payment.checkout_id
+    graphql_checkout_id = (
+        graphene.Node.to_global_id("Checkout", checkout_id) if checkout_id else None
+    )
+
+    payload = {
+        "action": {
+            "type": payment_data.action_requested,
```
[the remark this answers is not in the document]
korycins: You're right. I will change it.

saleor/webhook/payloads.py
korycins: @maarcingebala, 
Hhm, `to_global_id_or_none` takes an instance. In that case, I operate only on IDs. `payment.order_id`. If I will switch to this approach, I will have to call `payment.order` which means, that this could add an additional query to fetch an order/checkout instance. Are you ok with this?

maarcingebala: OK so ignore my comment then, we can keep your solution in this case and save the database query.

fowczarek (approved):

PR #9451, opened 2022-04-01
Ensure initialization of the dummy context succeeded

NyanKiyoshi:
Saleor tests introduced by aca6418d6c36956bc1ab530e6ef7e146ec9df90c were not picking up if unexpected error happened when creating the dummy context (e.g. middleware caused HTTP 500 or HTTP 404) which can cause unexpected results or behaviors

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/webhook/subscription_payload.py
+++ b/saleor/graphql/webhook/subscription_payload.py
@@ -56,7 +56,8 @@ def initialize_context() -> HttpRequest:
     handler = BaseHandler()
     context = RequestFactory().request()
     handler.load_middleware()
-    handler.get_response(context)
+    response = handler.get_response(context)
+    assert response.status_code == 200
     return context
 
 


saleor/graphql/webhook/subscription_payload.py
```diff
@@ -56,7 +56,8 @@ def initialize_context() -> HttpRequest:
     handler = BaseHandler()
     context = RequestFactory().request()
     handler.load_middleware()
-    handler.get_response(context)
+    response = handler.get_response(context)
+    assert response.status_code == 200
```
maarcingebala, line 60: How does the assert fixes the issue? Shouldn't we somehow handle this error?

NyanKiyoshi: The issue is if the error happens during the fixture, we never get to notice it. Thus we can get unexpected test results, where with this change we can see clearly that the status code error wasn't 200.

There isn't an issue with Saleor but the issue can happen in forks where there is additional logic

maarcingebala (approved):

fowczarek (approved):

IKarbowiak (approved):

PR #9472, opened 2022-04-04
Fix shipping methods caching

tomaszszymanski129:
I want to merge this change because it fixes caching of excluded shipping methods.

Broken caching was caused by:
order payload being compared, this contained two inconsistent fields:
1) invalid `meta` added to order payload which contained `issued_at` info - that was always different value because of timezone.now() usage
2) weight field. Uses `MeasurementField` which sometimes gives `1kg` and sometimes `1000g`, will be investigated and fixed in subsequent PR which will improve caching even further (check all other payloads). For now it's simply ignored in comparison so we still benefit from caching.

⚠️ There might be more caching things to check and fix, however these will be done in separate PR.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/plugins/webhook/shipping.py
+++ b/saleor/plugins/webhook/shipping.py
@@ -1,7 +1,8 @@
 import base64
+import json
 import logging
 from collections import defaultdict
-from typing import TYPE_CHECKING, Any, Callable, Dict, List
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional
 
 from django.core.cache import cache
 from django.db.models import QuerySet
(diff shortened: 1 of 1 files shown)


patrys:
Task linked: [SALEOR-6427 orderUpdateShipping calls order_filter_shipping_methods multiple times](https://app.clickup.com/t/2549495/SALEOR-6427)

saleor/plugins/webhook/shipping.py
```diff
@@ -51,6 +52,27 @@ def parse_list_shipping_methods_response(
     return shipping_methods
 
 
+def _compare_payloads(
```
korycins, line 55: Is this method generic? I can assume that based on the function name, but inside you're using `order` keys. Maybe it would be more safer to name it `compare_order_payloads`? 
Also, if we have method for comparing an order's payloads and it is dedicated to order, then shouldn't we have declared excluded_keys inside a function, instead of passing them as an input?

tomaszszymanski129: good point, will change in a moment

fowczarek (approved):

IKarbowiak (approved):

korycins (approved):

PR #9483, opened 2022-04-05
Enable JWT expiration by default

maarcingebala:
Enable JWT expiration by default.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -674,8 +674,7 @@ if REDIS_URL:
 CACHES = {"default": django_cache_url.config()}
 CACHES["default"]["TIMEOUT"] = parse(os.environ.get("CACHE_TIMEOUT", "7 days"))
 
-# Default False because storefront and dashboard don't support expiration of token
-JWT_EXPIRE = get_bool_from_env("JWT_EXPIRE", False)
+JWT_EXPIRE = True
 JWT_TTL_ACCESS = timedelta(seconds=parse(os.environ.get("JWT_TTL_ACCESS", "5 minutes")))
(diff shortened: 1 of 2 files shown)


saleor/settings.py
```diff
@@ -674,8 +674,7 @@ def SENTRY_INIT(dsn: str, sentry_opts: dict):
 CACHES = {"default": django_cache_url.config()}
 CACHES["default"]["TIMEOUT"] = parse(os.environ.get("CACHE_TIMEOUT", "7 days"))
 
-# Default False because storefront and dashboard don't support expiration of token
-JWT_EXPIRE = get_bool_from_env("JWT_EXPIRE", False)
+JWT_EXPIRE = True
```
korycins, line 677: Shouldn't we still have `get_bool_from_env` here but with the default value set to `True`?

maarcingebala: I think it should be always enabled. If you want to disable it locally, use `local_settings.py`.

L3str4nge (approved):

saleor/settings.py
IKarbowiak: I agree with @korycins, the environment variables aren't exactly for such a case like this when you want to override existing settings?

maarcingebala: Having it disabled is a security issue so I'm not sure it should be even configurable.

IKarbowiak (approved):

maarcingebala:
Docs: https://github.com/saleor/saleor-docs/pull/401

PR #9494, opened 2022-04-07
Add newsletter

karolkielecki:
I want to merge this change because it adds newsletter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@
   Join our active, engaged community: <br>
   <a href="https://saleor.io/">Website</a>
   <span> | </span>
-  <a href="https://medium.com/saleor">Blog</a>
+  <a href="https://saleor.io/blog/">Blog</a>
+  <span> | </span>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,7 +18,9 @@
   Join our active, engaged community: <br>
   <a href="https://saleor.io/">Website</a>
   <span> | </span>
-  <a href="https://medium.com/saleor">Blog</a>
+  <a href="https://saleor.io/blog/">Blog</a>
+  <span> | </span>
+  <a href="https://saleor.typeform.com/to/JTJK0Nou">Newsletter</a>
```
korycins, line 23: Shouldn't be something like `Subscribe to newsletter`? Personally, I would think that this link will redirect me to a page with all news, but maybe I am wrong 😄

krzysztofwolski: I agree, it would be more clear that way 👍

karolkielecki: Done

patrys (approved):

maarcingebala (approved):

PR #9503, opened 2022-04-08
Handle all Braintree errors

L3str4nge:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-6167 Uncatched error if Braintree is not available](https://app.clickup.com/t/2549495/SALEOR-6167)

saleor/payment/gateways/braintree/errors.py
```diff
@@ -117,3 +127,28 @@ def get_error_type(error_code):
...
+    elif isinstance(exc, UpgradeRequiredError):
+        msg = "Upgrade required error"
+    elif isinstance(exc, TooManyRequestsError):
+        msg = "Too many requests error"
+    elif isinstance(exc, ServerError):
+        msg = "Server error"
+    elif isinstance(exc, ServiceUnavailableError):
+        msg = "Server unavailable error"
+    elif isinstance(exc, GatewayTimeoutError):
+        msg = "Gateway timeout error"
```
IKarbowiak, line 150: IMO, the better and clearer solution is using mapping here, where keys are errors and values are messages. Then we can just check if `exc` is in the given mapping and return the correct msg if is, or default if isn't.

saleor/payment/gateways/braintree/errors.py
```diff
@@ -117,3 +127,23 @@ def get_error_type(error_code):
...
+    TooManyRequestsError: "Too many requests error",
+    ServerError: "Server error",
+    ServiceUnavailableError: "Server unavailable error",
+    GatewayTimeoutError: "Gateway timeout error",
+}
+
+
+def handle_braintree_error(exc):
+    msg = BRAINTREE_ERROR_MESSAGES.get(
+        type(exc), "Unexpected http response " + str(exc)
```
korycins, line 147: `Unexpected http response` - is it correct? I would say that we had unexcpected paymet/braintree error, not sure if we have enough context here to say that the problem was related with http 🤔

korycins (approved):

saleor/payment/gateways/braintree/errors.py
L3str4nge: Got it from braintree - https://github.com/braintree/braintree_python/blob/master/braintree/util/http.py#L56

korycins: Yeah, but if Braintree will add a new exception there, then we will raise an exception `Unexpected http response` as we will not have this error in the map, right? So not sure if the `default` error, shouldn't be more generic

PR #9504, opened 2022-04-08
Fix payment flow

IKarbowiak:
### First problem:
Sometimes we can have situations when something went wrong during order creation (i.e. tax plugin raises an error)  and the refund process is started. In the meantime, we got a notification from the payment gateway with confirmation that the payment was captured. At this step, we usually have an order, but if there isn't we call the method for creating the order. As a result, we're ending with an order, which is refunded just after creation. 
Fix: before order creation, I'm checking if there is any ongoing refund for a given payment.

### Second problem:
The order has two payments:
- first was created, and we get a proper response from the gateway, but again something went wrong, during order creation, so we started the refund process. 
- because of the problems somebody run the payment creation process again, this time everything goes well, and the order was created. The first payment was deactivated
- then we got some notification from getaway regarding the first payment, but this payment was deactivated, so again we run the refund process. But before that, we handled the notification and doubled captured amount
- after that, the first refund was properly performed, and we received a notification about that, so the captured amount was reduced, but just before the amount was doubled, so we ended with two payments, both of them with a captured amount equal to the total value of the order. 

So in the second scenario, we have two problems:
1. we shouldn't captured the payment again, if it was already captured
2. the refund should be started if there is any already ongoing

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/complete_checkout.py
+++ b/saleor/checkout/complete_checkout.py
@@ -44,7 +44,7 @@ from ..order.fetch import OrderInfo, OrderLineInfo
 from ..order.models import Order, OrderLine
 from ..order.notifications import send_order_confirmation
 from ..order.search import prepare_order_search_document_value
-from ..payment import PaymentError, gateway
+from ..payment import PaymentError, TransactionKind, gateway
 from ..payment.models import Payment, Transaction
 from ..payment.utils import fetch_customer_id, store_customer_id
 from ..product.models import ProductTranslation, ProductVariantTranslation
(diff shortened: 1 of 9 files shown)


patrys:
Task linked: [SALEOR-6467 Auto refund investigation](https://app.clickup.com/t/2549495/SALEOR-6467)

L3str4nge (approved):

korycins (approved):

PR #9560, opened 2022-04-19
Generic clean up for new payment flow

korycins:
- adds permission for handling metadata related to TransactionItem
- added permission requirement for `order.transactions` field
- cleanup related to mutation/queries labels:
   -  remove deprecation notes as we don't offer a new approach as it is still a PREVIEW_FEATURE
   - add missing labels for 3.2 and preview_feature

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/tests/fixtures.py
```diff
@@ -4129,8 +4129,6 @@ def draft_order_with_preorder_lines(order_with_preorder_lines):
     return order_with_preorder_lines
 
 
-# FIXME This fixtures should be renamed to legacy_ when we will convert current payment
-# flow to depreceated
 @pytest.fixture
 def payment_txn_preauth(order_with_lines, payment_dummy):
```
L3str4nge, line 4133: But why not deprecated?

L3str4nge (approved):

saleor/tests/fixtures.py
korycins: We changed an approach here. Previously I assumed that we will convert model `Payment` to `LegacyPayment` but we don't get any benefits about renaming it for now.
This fixme was added to keep in mind about this, but with separate model that we created `TransactionItem`. It is not needed anymore.

fowczarek (approved):

IKarbowiak (approved):

PR #9570, opened 2022-04-20
New events for changes related to channels

SzymJ:
I want to merge this change because it adds new subscription base webhooks for channel create, update, delete, activate and deactivate events.

New webhook events:

* `CHANNEL_CREATED`
* `CHANNEL_UPDATED`
* `CHANNEL_DELETED`
* `CHANNEL_STATUS_CHANGED`

Docs PR [here](https://github.com/saleor/saleor-docs/pull/406)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-6482 Add channel webhook events](https://app.clickup.com/t/2549495/SALEOR-6482)

saleor/graphql/webhook/subscription_types.py
```diff
@@ -57,6 +57,41 @@ class CategoryDeleted(ObjectType, CategoryBase):
...
+    ...
+
+
+class ChannelStatusChanged(ObjectType, ChannelBase):
+    status = graphene.String(
+        description="New status of a channel.",
+    )
+
+    @staticmethod
+    def resolve_status(root, info):
```
korycins, line 90: What do we achieve by adding this field? Looks like `channel.status` returns ACTIVE/INACTIVE based on value `isActive`. The same result I can achieve by fetching `channel.isActive`, right?

SzymJ: I thought that this will be more descriptive but now I think that you are probably right and `isActive` should be enough.

korycins (approved):

saleor/plugins/webhook/plugin.py
```diff
@@ -114,6 +115,49 @@ def category_deleted(self, category: "Category", previous_value: None) -> None:
...
+        event_type = WebhookEventAsyncType.CHANNEL_STATUS_CHANGED
+        if webhooks := get_webhooks_for_event(event_type):
+            payload = {
+                "id": graphene.Node.to_global_id("Channel", channel.id),
+                "is_active": channel.is_active,
+            }
+            trigger_webhooks_async(
+                payload, event_type, webhooks, channel, self.requestor
+            )
+
```
L3str4nge, line 160: Looking at this module we have a lot of duplications. I think 
```
if not self.active:
    return previous_value
``` 
could be decorated for example `@active_required`

and 
```
if webhooks := get_webhooks_for_event(event_type):
    payload = {"id": graphene.Node.to_global_id("Channel", channel.id)}
    trigger_webhooks_async(
        payload, event_type, webhooks, channel, self.requestor
     )
``` 
would be `trigger_webhooks_for_event(WebhookEventAsyncType.CHANNEL_STATUS_CHANGED)` with building payload etc inside. It's just an idea but for additional statuses there will be more duplications and mess in general. It's out of the scope ofc.

wdyt? @korycins @SzymJ @maarcingebala

L3str4nge (approved):

tomaszszymanski129 (approved):

saleor/plugins/webhook/plugin.py
korycins: @L3str4nge  IMO The better way would be to handle this on plugins manager side. And... 🥁 🥁 🥁 
It is already done 😄  (at least for the "default" way of calling plugins)
`saleor.plugins.manager.PluginsManager.__run_method_on_plugins`.

L3str4nge: So if it's already done it would be awesome to get rid of that mess :D

maarcingebala: Sounds good, but we should then rework all these methods, not only the new ones from this PR. So maybe for now a maintenance task and we would do it in a separate PR?

maarcingebala (approved):

PR #9586, opened 2022-04-21
Allow applications to subscribe to events automatically using the manifest

magul:
I want to merge this change because we want to be able to define webhooks inside app's manifest file.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested


patrys:
Task linked: [SALEOR-6404 Expanding app manifest](https://app.clickup.com/t/2549495/SALEOR-6404)

saleor/app/manifest_validations.py
```diff
@@ -4,9 +4,12 @@
 
 from django.contrib.auth.models import Permission
 from django.core.exceptions import ValidationError
+from django.core.validators import URLValidator
 from django.db.models import Value
 from django.db.models.functions import Concat
 
+from saleor.graphql.webhook.mutations import validate_query
```
maarcingebala, line 11: Use relative import for consistency.

maarcingebala (approved):

maarcingebala:
@magul Linters are failing (I recommend using pre-commit locally to avoid that). Also, can we add some tests for the new code?

saleor/app/manifest_validations.py
```diff
@@ -206,3 +213,16 @@ def validate_required_fields(manifest_data, errors):
                     code=AppErrorCode.REQUIRED.value,
                 )
             )
+
+    webhooks = manifest_data.get("webhooks", [])
+    for webhook in webhooks:
+        webhook_fields = set(webhook.keys())
+        missing_fields = webhook_required_fields.difference(webhook_fields)
+        if missing_fields:
```
tomaszszymanski129, line 221:
```suggestion
        if missing_fields := webhook_required_fields.difference(webhook_fields):
```

tomaszszymanski129 (approved):

jakubkuc (approved):

patrys (changes requested):
It needs a docs PR before it's ready to be merged. We've recently merged several changes that didn't come with nearly enough documentation to make them useful.

saleor/graphql/app/types.py
```diff
@@ -124,6 +125,26 @@ class Meta:
         node = AppExtension
 
 
+class AppManifestWebhook(graphene.ObjectType):
+    name = graphene.String(required=True)
+    events = NonNullList(WebhookEventTypeEnum, required=True)
+    query = graphene.String(required=True)
+    target_url = graphene.String(required=True)
+    is_active = graphene.Boolean(required=True)
```
IKarbowiak, line 133: Could we add the descriptions to the fields here?

saleor/app/installation_utils.py
```diff
@@ -62,6 +64,24 @@ def install_app(
         )
         extension.permissions.set(extension_data.get("permissions", []))
 
+    for webhook_data in manifest_data.get("webhooks", []):
+        with transaction.atomic():
```
korycins, line 68: Can we do it in bulk_create?

saleor/app/manifest_validations.py
```diff
@@ -12,6 +12,7 @@
     get_permissions_enum_list,
     split_permission_codename,
 )
+from ..graphql.webhook.mutations import validate_query
```
korycins, line 15: If you need to import logic from graphql's mutation file then I would say, that the `validate_query` should be placed in a different file. (Previously it was used only in mutations, if we want to use them in second place, let's move it to more common file.)

saleor/app/manifest_validations.py
```diff
@@ -178,31 +180,43 @@ def clean_extensions(manifest_data, app_permissions, errors):
         _clean_extension_permissions(extension, app_permissions, errors)
 
 
+def clean_webhooks(manifest_data, app_permissions, errors):
+    webhooks = manifest_data.get("webhooks", [])
+    for webhook in webhooks:
+        validate_query(webhook["query"])
```
korycins, line 186: You don't use `errors` here. 
IMO, you should iterate over a webhooks and gather all errors that you got. Currently, it will raise an exception after discovering first incorrect query.

saleor/app/manifest_validations.py
```diff
@@ -178,31 +180,43 @@ def clean_extensions(manifest_data, app_permissions, errors):
         _clean_extension_permissions(extension, app_permissions, errors)
 
 
+def clean_webhooks(manifest_data, app_permissions, errors):
```
korycins, line 183: What about validations for:
"targetUrl", "isActive", "events" ?

saleor/graphql/app/types.py
magul: @maarcingebala 
* how would I know if that's preview feature? who can provide answer to that question?
* assuming, that I'm providing the whole `webhooks` as `NonNullList` in this version, should I mark `ADDED_IN_34` just that list? Should I mark also nested fields? Should I mark just nested fields, leaving list unmarked?

maarcingebala: - Usually, an owner of the feature should be able to provide the answer; I don't know who owns this feature, maybe @timuric?
- Since we are adding an entirely new type here, we could add ADDED_IN_34 at the type description level. To add a description to the type you can add:
```
class Meta:
    description = ""
```

saleor/app/manifest_validations.py
magul: As we decided on a call I'm going to use `validate_subscription_query` (so it will requre it to be non-empty), and there's no reason to move `validate_subscription_query` from `graphql.webhook.subscription_payload`

maarcingebala:
@magul What's the status here? There are a few unresolved comments.

magul:
As we encountered an issue with core not triggering webhoooks, that have subscriptions defined we're decided to make subscription queries optional in manifest (that will also allow us to support both sync and async events).

I will deliver that, but currently I'm swamped in something else (give me 2 days to return here).

patrys:
@magul Aren't we effectively building a legacy feature by enabling the subscription to be skipped? I want to deprecate non-subscription webhooks as soon as possible, to me it does not make sense to expand its featureset.

saleor/app/manifest_validations.py
```diff
@@ -178,31 +182,93 @@ def clean_extensions(manifest_data, app_permissions, errors):
...
+
+    target_url_validator = AppURLValidator(
+        schemes=["http", "https", "awssqs", "gcpubsub"]
+    )
+
+    for webhook in webhooks:
+        if not validate_subscription_query(webhook["query"]):
+            errors["webhooks"].append(
+                ValidationError(
+                    "Subscription query is not valid.",
```
korycins, line 203: Would be good to provide here more context. Currently, we know that the subscription query is invalid but we don't know which webhook declaration has an issue.

magul: I agree with that, but at the same time I wanted to make it coherent with how we validate extensions - we also do not provide mean to identify, which one is causing validation error (see: https://github.com/saleor/saleor/blob/main/saleor/app/manifest_validations.py#L160)

* should I implement that in this pull request? should I also change the way `extensions` errors are defined?
* is changing shape of `errors` object considered breaking change?

korycins (approved):

saleor/graphql/app/types.py
```diff
@@ -264,7 +300,7 @@ class App(ModelObjectType):
     )
     extensions = NonNullList(
         AppExtension,
-        description="App's dashboard extensions." + ADDED_IN_31 + PREVIEW_FEATURE,
+        description=f"App's dashboard extensions.{ADDED_IN_31}{PREVIEW_FEATURE}",
```
maarcingebala, line 303: Why are we changing this? The previous format was the one commonly used.

maarcingebala (approved):

maarcingebala:
@magul It looks ready to merge from my perspective, I would just fix the little thing with the description format mentioned above and we're good to go.

saleor/graphql/app/types.py
magul: done, reverted.

magul:
Unleashing @eCzerniak !

eCzerniak (approved):

PR #9626, opened 2022-04-26
Limit transaction mutations to be used only by apps

korycins:
I want to merge this change because the `transactionCreate`, `transactionUpdate` should be only accessible by App.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


fowczarek (approved):
Should we add a custom message for this case? 
Currently, when a staff user wants to perform this mutation he receives generic permission denied information. IMO we should inform it in the message that this action is available only for apps.

korycins:
@fowczarek I already added a message that the mutation can be called only by app 🤔

saleor/graphql/payment/mutations.py
```diff
@@ -640,11 +640,21 @@ class Arguments:
...
         permissions = (PaymentPermissions.HANDLE_PAYMENTS,)
 
+    @classmethod
+    def check_permissions(cls, context, permissions=None):
+        """Determine whether app has rights to perform this mutation."""
+        permissions = permissions or cls._meta.permissions
+        app = getattr(context, "app", None)
+        if app:
+            return app.has_perms(permissions)
+        return False
```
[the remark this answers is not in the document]
maarcingebala: `AUTHENTICATED_APP` grants access to ANY app that is authenticated, but it doesn't check permissions. In this case if HANDLE_PAYMENTS is required, we need to do it this way.

saleor/graphql/payment/mutations.py
```diff
@@ -640,11 +640,21 @@ class Arguments:
...
         )
         error_type_class = common_types.TransactionCreateError
         permissions = (PaymentPermissions.HANDLE_PAYMENTS,)
 
+    @classmethod
+    def check_permissions(cls, context, permissions=None):
+        """Determine whether app has rights to perform this mutation."""
+        permissions = permissions or cls._meta.permissions
+        app = getattr(context, "app", None)
+        if app:
```
tomaszszymanski129, line 654:
```suggestion
        if app := getattr(context, "app", None):
```

tomaszszymanski129 (approved):

tomaszszymanski129:
@korycins build is failing

saleor/graphql/checkout/mutations/order_create_from_checkout.py
```diff
@@ -51,7 +51,8 @@ class Arguments:
 
     class Meta:
         description = (
-            f"{ADDED_IN_32} Create new order from existing checkout. {PREVIEW_FEATURE}"
+            f"{ADDED_IN_32} Create new order from existing checkout. Can be used only"
+            f" by apps. {PREVIEW_FEATURE}"
```
maarcingebala, line 55: We could also do something like this:

```
class Meta:
    auto_permission_message = False
    description = (
        f"{ADDED_IN_32} Create new order from existing checkout. Requires the "
        f"following permissions: AUTHENTICATED_APP and HANDLE_CHECKOUTS."
        f"{PREVIEW_FEATURE}"
    )
```

This way we override the autogenerated message about required permissions, and instead of writing our own description we use AUTHENTICATED_APP to represent that it requires an authenticated app.

IKarbowiak (approved):

saleor/graphql/checkout/mutations/order_create_from_checkout.py
korycins: Thanks <3

PR #9699, opened 2022-05-05
Add fix for multiplied prices on Avatax side

korycins:
I want to merge this change because it fixes an issue with multiplied values passed to Avatax.
Avatax for UK's addresses propagates shipping taxes over checkout/order lines. Previously we stored returned net/gross amounts and used them to prepare a payload for the next call to Avatax. Each next call was using a net/gross amount returned by Avatax so each new call had a bigger value than previously. 
This PR introduces:
-  base_price fields for Order, to store price saved during the order creation. These fields are used to re-calculate taxes
- drop logic for sending Avatax prices for undiscounted_price, price_with_sale, price_with_discounts. We send only one price now. The undiscounted price returned by AvataxPlugin doesn't have taxes anymore. This change was required as Avatax propagates shipping taxes over lines, so we were never receiving the correct amount for a single line we send 3 separate lines in Avatax's payload
- voucher of type specific_products or apply_once_per_order will not be included in `checkout.discount` as all voucher's discount amount is propagated over a checkout's lines.

**PS. Don't panic about number of lines, I have to re-generate all Avatax's cassettes 😄** 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-6684 Multiple calls to Avatax increases a price of lines](https://app.clickup.com/t/2549495/SALEOR-6684)

saleor/discount/tests/test_discounts.py
```diff
@@ -367,57 +364,6 @@ def test_voucher_queryset_active_in_other_channel(voucher, channel_PLN):
...
-        ([10], 10, DiscountValueType.FIXED, True, 10),
-        ([5], 10, DiscountValueType.FIXED, True, 5),
-        ([5, 5], 10, DiscountValueType.FIXED, True, 5),
-        ([2, 3], 10, DiscountValueType.FIXED, True, 2),
-        ([10, 10], 5, DiscountValueType.FIXED, False, 10),
-        ([5, 2], 5, DiscountValueType.FIXED, False, 7),
-        ([10, 10, 10], 5, DiscountValueType.FIXED, False, 15),
-    ],
-)
-def test_specific_products_voucher_checkout_discount(
```
korycins, line 382: I've deleted this test as the same we have in checkout module.

saleor/checkout/fetch.py
```diff
@@ -342,7 +356,10 @@ def _apply_voucher_on_product(
         if cheapest_line:
             discounted_lines_by_voucher.append(cheapest_line)
     else:
-        discounted_lines_by_voucher.extend(get_discounted_lines(lines_info, voucher))
+        voucher_info = fetch_voucher_info(voucher)
```
L3str4nge, line 359: question: Why are we fetching voucher info if we already have it? `voucher_info: "VoucherInfo",` Maybe it's just a naming issue?

saleor/graphql/discount/dataloaders.py
```diff
@@ -150,6 +151,57 @@ def batch_load(self, keys):
...
+        for voucher_id, variant_id in voucher_variants:
+            variant_pks_map[voucher_id].append(variant_id)
+        for voucher_id, category_id in voucher_categories:
+            category_pks_map[voucher_id].append(category_id)
+        for voucher_id, collection_id in voucher_collections:
+            collection_pks_map[voucher_id].append(collection_id)
+        voucher_infos = []
+        for code in keys:
+            voucher = vouchers_map.get(code)
+            if not voucher:
```
fowczarek, line 191: This could break the dataloader. Dataloader should always return a list of the same length as `keys`.

saleor/plugins/avatax/__init__.py
```diff
@@ -365,25 +353,11 @@ def get_order_lines_data(
...
-        else:
-            append_line_to_data(
-                **append_line_to_data_kwargs,
-                amount=price_with_discounts_amount,
-                ref1=line.variant.sku,
-            )
+        append_line_to_data(
+            **append_line_to_data_kwargs,
+            amount=price_with_discounts_amount,
+            # ref1=line.variant.sku,
```
fowczarek, line 359: :police_car: :police_officer:

saleor/graphql/discount/dataloaders.py
```diff
@@ -150,6 +151,57 @@ def batch_load(self, keys):
...
+        ).values_list("voucher_id", "product_id")
+        voucher_variants = Voucher.variants.through.objects.filter(
+            voucher__in=vouchers
+        ).values_list("voucher_id", "productvariant_id")
+        voucher_collections = Voucher.collections.through.objects.filter(
+            voucher__in=vouchers
+        ).values_list("voucher_id", "collection_id")
+        voucher_categories = Voucher.categories.through.objects.filter(
+            voucher__in=vouchers
+        ).values_list("voucher_id", "category_id")
```
fowczarek, line 175: We should add `.using(self.database_connection_name)` between `objects` and `filter`.

saleor/graphql/discount/dataloaders.py
korycins: I used `saleor.graphql.checkout.dataloaders.CheckoutInfoByCheckoutTokenLoader` as an example. `CheckoutInfoByCheckoutTokenLoader` uses `VoucherByCodeLoader` inside the load methods. `VoucherByCodeLoader` fetches a vouchers based on codes taken from checkout:
```
                voucher_codes = {
                    checkout.voucher_code
                    for checkout in checkouts
                    if checkout.voucher_code
                }
                vouchers = VoucherByCodeLoader(self.context).load_many(voucher_codes)
```
In that moment, I don't think that we have an order that could match to `checkouts`. 
After that, we create `voucher_map`:
```
voucher_map = {voucher.code: voucher for voucher in vouchers}
...
voucher = voucher_map.get(checkout.voucher_code)
```
and use it to fetch correct voucher to checkout with promo code. 

I used this flow as an example. @fowczarek What do you think? Do you think this is bad pattern or incorrect usage of dataloader?

korycins:
> check the db-queries, something is definitely wrong here.

@tomaszszymanski129 Thanks for raising this <3 I've spent half of the day to find the issue 😄

saleor/graphql/discount/dataloaders.py
```diff
@@ -150,6 +151,71 @@ def batch_load(self, keys):
         ]
 
 
+class VoucherInfoByVoucherCodeLoader(DataLoader):
+    context_key = "voucher_info_by_voucher_code"
+
+    def batch_load(self, keys):
+        vouchers_map = (
+            Voucher.objects.using(self.database_connection_name)
+            # FIXME dataloader should not operate on prefetched data. The channel
```
maarcingebala, line 160: If not yet done, we should open a task for SRE/core to fix it.

maarcingebala (approved):

saleor/graphql/discount/dataloaders.py
korycins: The task is already created

saleor/checkout/tests/test_checkout.py
```diff
@@ -351,13 +348,83 @@ def test_get_discount_for_checkout_specific_products_voucher(
...
+
+@pytest.mark.parametrize(
+    "discount_value, discount_type, total_discount_amount",
+    [
+        (5, DiscountValueType.FIXED, 5),
+        (10000, DiscountValueType.FIXED, 10),
+        (10, DiscountValueType.PERCENTAGE, 1),
+    ],
+)
+def test_get_discount_for_checkout_specific_products_voucher_applu_only_once(
```
IKarbowiak, line 384: Typo, should by `apply` instead of `applu`

IKarbowiak (approved):

PR #9718, opened 2022-05-09
Add benchmark for fulfillment warehouse

8r2y5:
I want to merge this change because I want to add benchmark for fulfillment warehouse resolvers

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/tests/benchmark/fixtures.py
+++ b/saleor/graphql/order/tests/benchmark/fixtures.py
@@ -6,7 +6,7 @@ from prices import Money, TaxedMoney
 
 from .....account.models import User
 from .....order import OrderEvents, OrderStatus
-from .....order.models import Fulfillment, Order, OrderEvent, OrderLine
+from .....order.models import Fulfillment, FulfillmentLine, Order, OrderEvent, OrderLine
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [SALEOR-3729 Optimize Fulfillment type resolvers](https://app.clickup.com/t/2549495/SALEOR-3729)

maarcingebala (approved):

IKarbowiak (changes requested):
We should also extend the fixture `test_fulfillment_query` as there is only one `Warehouse` so we do not test the performance of fetching warehouses.

tomaszszymanski129:
@IKarbowiak we actually do, the resolver fires for each order so in this case query count will drop anyway as the old resolver added up to 3 extra queries for each. But I agree, more objects is better when it comes to benchmarks, we should add some.

IKarbowiak:
@tomaszszymanski129 I checked locally the test that uses this query and there was only one warehouse instance.

IKarbowiak (approved):

maarcingebala (approved):

PR #9748, opened 2022-05-11
Fix file upload URLs when S3 is configured

### LINKED ISSUE #8106: Using cloud storage for upload media doesn't work well with file type attributes.
Use cloud storage to store media. For products it goes well but for file types attributes, API doesn't return the correct media URL.

1. Configure Saleor to use a Google Cloud Storage bucket.
2. Generate a new attribute selecting in the dropdown "Catalog Input type for Store Owner"  File type.
3. Create a page using that attribute and uploading an image.
4. When saving the page, the URL of the image it's not the cloud storage URL (can see it in the screenshot), it's an URL with the IP of the machine where the API is deployed. In the cloud storage bucket configured we can see the image uploaded.
### What I expected to happen
The API must return the image url of the attribute with the correct url referencing to cloud storage.

### Screenshots
saleor-bug

**System information**

Saleor version:
- [ ] dev (current master)
- [X] 3.0
- [ ] 2.11
- [ ] 2.10

Operating system:
- [ ] Windows
- [X] Linux
- [ ] MacOS
- [ ] Other

### LINKED ISSUE #9421: file urls are not valid if using S3 or GS
as mentioned in here: https://github.com/saleor/saleor/pull/9332

Any file URL is parsed from the server IP even if the server isn't serving the media files

dougli:
Fixes #8106. Fixes #9421. where file uploads on page attributes (and page attributes only -- product images are working) don't return the correct URL if AWS S3 or Google Cloud Storage is configured.

Test plan:
* Added an additional test case. Shimmed out the .save() function on the underlying S3 storage so the test doesn't actually save anything or require an S3 account.
* Ran all existing test cases.
* Configured a saleor instance to use S3.
* In a saleor admin dashboard, created a page that uses file attributes.
* Uploaded a file through the attribute. Clicked save.
* Verified that the file preview now pointed to S3.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/graphql/attribute/types.py
+++ b/saleor/graphql/attribute/types.py
@@ -1,6 +1,8 @@
 from typing import cast
+from urllib.parse import urljoin
 
 import graphene
+from django.conf import settings
 from django.db.models import QuerySet
 
 from ...attribute import AttributeEntityType, AttributeInputType, models
@@ -80,7 +82,10 @@ class AttributeValue(ModelObjectType):
     def resolve_file(root: models.AttributeValue, _info):
         if not root.file_url:
             return
-        return File(url=root.file_url, content_type=root.content_type)
+        return File(
+            url=urljoin(settings.MEDIA_URL, root.file_url),
+            content_type=root.content_type,
+        )
 
     @staticmethod
     def resolve_reference(root: models.AttributeValue, info):


dougli:
Thanks. We are using CloudFront to serve our files and using AWS_MEDIA_CUSTOM_DOMAIN as [the 3.x docs](https://docs.saleor.io/docs/3.x/developer/running-saleor/s3) described. Saleor was disregarding the setting though for page attributes and returning URLs with the incorrect domain name, hence this PR.

patrys:
You can look at changing the way URLs are resolved on output instead of how they are stored in the database. We don't want to store the full domain name in each image field as it would make it impossible to change that domain (for example, when promoting your staging environment to production).

dougli:
Good point. I'm abandoning this change then, setting MEDIA_URL worked for us on output. Thanks for the help!

PR #9526, opened 2022-04-12
Add support for calculating taxes in Saleor Apps

fowczarek:
I want to merge this change because continue working on #9220

Our goal is to be able to dynamically calculate taxes during the checkout/order process. To achieve this goal we use Saleor Apps system. 

# 1. General Assumptions
- Prices are denormalized (indefinitely for orders, or for settings.CHECKOUT_PRICES_TTL for checkouts) until they are invalidated or expired
- When there are more than one app subscribed to CHECKOUT_CALCULATE_TAXES or ORDER_CALCULATE_TAXES we iterate over them until we get a proper response
- New permission required to handle sync tax webhooks is `HANDLE_TAXES` 
- When apps and plugins don't return any tax data we return the error on the CheckoutComplete or OrderCreateFromCheckout step. 

# 2. Terms
- Sale - price reduction is applied automatically when all conditions are met
- Voucher - price reduction applied when the user adds promo code to checkout and all conditions are met
- Discounts - sales and vouchers

# 3. General flow
```mermaid
sequenceDiagram
    User ->>+ Saleor: Query requesting model's prices
    opt if prices are invalid
        Saleor->>+Manager: Calculate taxes for model <br>with old approach
        loop iterate over tax plugins
          Manager ->>+ Tax Plugin: Fetch prices by manager.calculate_*
          Tax Plugin -->>- Manager: Return prices (with optional taxes)
        end
        Manager -->>- Saleor: Apply prices to model
        Saleor ->>+ Manager: Calculate taxes <br>with new approach
          Manager ->>+ Webhook Plugin: Fetch taxes by <br>manager.get_taxes_for_*
          loop iterate over tax apps
            Webhook Plugin ->>+ Tax App: Send sync webhook
            Tax App -->>- Webhook Plugin: Send tax data
            Webhook Plugin ->> Webhook Plugin: Exit when received valid tax data<br>or no more apps
          end
          Webhook Plugin -->>- Manager: Return Optional[TaxData]
        opt if received tax data
          Manager-->>-Saleor: Apply taxes to model
        end
        Saleor ->> Saleor: Save all model changes to DB
    end
    Saleor -->>- User: Return model's prices
```

# 4. API changes
## 4.1 Extend PermissionEnum
```graphql
enum PermissionEnum {
    …
     HANDLE_TAXES
}
```
## 4.2 Extend WebhookEventTypeEnum
```graphql
enum WebhookEventTypeEnum {
    …
    CHECKOUT_CALCULATE_TAXES
[...]

PR #9749, opened 2022-05-12
Refactor checkout tax interface in plugins 

fowczarek:
I want to merge this change because of refactoring the checkout tax interface in plugins.

1. Drop the concept of `undiscounted_price`, `price_with_sale`, and `price_with_discounts` instead of that taxes should be calculated on the final price. 
2. Refactor SyncTax input interface to be consistent with description from #9526 

:warning: 
This PR is to feature branch not to main. 
Migrations will be refacotred in #9526.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/checkout/base_calculations.py
```diff
@@ -26,19 +25,8 @@ def calculate_base_line_unit_price(
...
-    Voucher is added to 'price_with_discounts' when line's product matches to products
-    applicable for voucher.
-    For voucher with 'apply once per order', the price will be reduced in calculation
-    of total price.
-    'price' includes discount from sale if any valid exists.
-    'price_with_discounts' includes voucher discount and sale discount if any valid
-    exists.
-    'undiscounted_price' is a price without any sale and voucher.
-    """
+) -> Money:
```
korycins, line 28: Would be good to have some description here. Especially that we're applying here voucher

korycins: Somethin like:
```
"""Calculate line unit price.
Unit price includes discount from sale and voucher
A voucher is added to the unit price when line's product matches to products applicable for the voucher.
"""
```

saleor/webhook/serializers.py
```diff
@@ -87,7 +87,9 @@ def untaxed_price_amount(price: TaxedMoney) -> Decimal:
     return [
         {
             **_get_checkout_line_payload_data(line_info),
-            "unit_amount": untaxed_price_amount(line_info.line.unit_price),
+            "unit_amount": untaxed_price_amount(
+                line_info.line.total_price / line_info.line.quantity
```
korycins, line 91: Not sure if `untaxed_price_amount` already does this, but keep in mind that you should quantize_price here

fowczarek: image
`quantize_price` is used inside `untaxed_price_amount`.
Do you want to extract it from this function?

korycins: IMO we can extract it

IKarbowiak (approved):

maarcingebala (approved):
We should remember about updating Avatax Excise plugin.

PR #9753, opened 2022-05-12
Observability webhook event type and permission

przlada:
I want to merge this change because it adds:
- `app.manage_observability` permission
- `OBSERVABILITY` async webhook event type

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/webhook/enums.py
```diff
@@ -82,6 +82,7 @@
     WebhookEventAsyncType.VOUCHER_UPDATED: "A voucher is updated.",
     WebhookEventAsyncType.VOUCHER_DELETED: "A voucher is deleted.",
     WebhookEventAsyncType.ANY: "All the events.",
+    WebhookEventAsyncType.OBSERVABILITY: "A observability event is created.",
```
jakubkuc, line 85:
```suggestion
    WebhookEventAsyncType.OBSERVABILITY: "An observability event is created.",
```

jakubkuc:
@magul 
This PR adds only event type - No event of this type exists yet, so I don't think adding subscription payload definition or adding this type to SUBSCRIBABLE_EVENTS is necessary ATM. 
Though it doesn't break anything, I don't get why would we add event type and permission if it's not yet used anywhere.

magul:
@jakubkuc 
when you wrote

> This PR adds only event type - No event of this type exists yet,

did you mean events being triggered?

I would assume, that subscription etc. should be delivered alongside event type definition, not when it's going to be triggered for the first time.

jakubkuc:
@magul  Well... I think you're actually right. Since we are actively working on covering all the events with subscriptions maybe we shouldn't allow to add new event types without covering them

przlada:
Sorry, I should have made it more clear. Observability webhook events will not be subscribable (at least for now), but there was discussion that a new permission type is required. I will follow up this PR with observability reporter PR making use of it. 
I separated the creation of event type and permissions to ease the review and naming discussion, but maybe it was a wrong move.

maarcingebala (approved):

PR #9759, opened 2022-05-13
Meta fields added to subscription webhooks event types.

SzymJ:
I want to merge this change because it adds `meta fields` to subscription webhooks types. Now users can ask for those fields in the subscription query. Additionally `meta` will always be returned for regular webhooks.  

New types:
- `Event` - changed from Union to Interface and used in subscription webhook types
- `IssuingPrincipal` - Union type of App and User

Lots of files are changed but it's mostly a `meta fields` check added to the events test. You can find the main changes in:
- `saleor/graphql/webhook/subscription_types.py`
- `saleor/graphql/webhook/subscription_payload.py`
- `saleor/plugins/webhook/tasks.py`

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-6763 Webhook's meta fields should be available in types](https://app.clickup.com/t/2549495/SALEOR-6763)

saleor/graphql/webhook/subscription_types.py
```diff
@@ -38,7 +44,80 @@
...
+    app = graphene.Field(
+        "saleor.graphql.app.types.App",
+        description="Look up a recipient of the webhook."
+        + ADDED_IN_34
+        + PREVIEW_FEATURE,
+    )
+
+    @staticmethod
+    def resolve_issued_at(_root, _info):
+        return timezone.now()
```
L3str4nge, line 93: Is it correct? If we wan't to resolve older events it will be always current datetime.

patrys: This needs to be timestamped when the event happens, not when it's delivered.

SzymJ: `create_deliveries_for_subscriptions` is called when event happens and `EventPayload` is created. So `issued_at =/= delivery time` or I'm missing something?

patrys: We want all payloads to have the exact same timestamp (and the timeline events for orders should also share the same timestamp).

saleor/graphql/webhook/subscription_types.py
```diff
@@ -38,7 +44,80 @@
 }
 
 
-class AppBase(AbstractType):
+class IssuingPrincipalType(Enum):
+    APP = "app"
+    USER = "user"
+
+
+class IssuingPrincipal(ObjectType):
```
maarcingebala, line 52: IssuingPrincipal could be implemented as union, see usages of `graphene.Union` and [the docs here](https://graphql.org/learn/schema/#union-types).

saleor/graphql/webhook/subscription_types.py
```diff
@@ -38,7 +44,80 @@
...
+    @staticmethod
+    def resolve_type(_root, info):
+        if isinstance(info.context.requestor, (User, AnonymousUser)):
+            return IssuingPrincipalType.USER.value
+        return IssuingPrincipalType.APP.value
+
+
+class Meta(ObjectType):
+    issued_at = graphene.DateTime(
+        description="Time of generating the event." + ADDED_IN_34 + PREVIEW_FEATURE,
```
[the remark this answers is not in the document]
patrys: Does it even need to be an embedded model? Why not create a base Event interface that already has those fields and have other events extend it? We could then also create base interfaces for different groups of events with common payload shapes: ProductEvent, OrderEvent, etc., and allow clients to say:

```graphql
subscription {
  event {
    type
    issuedAt
    ... on OrderEvent {
       order {
         id
       }
    }
  }
}
```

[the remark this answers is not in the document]
SzymJ: `Event` field is a `Union` and we can't add extra fields to it, so we are not able to build a query like this one above. But we can use an interface or create `AbstractType` with meta fields and use it in all event types but the query will look like this:
```
subscription {
  event {
    ... on OrderEvent {
       type
       issuedAt
       order {
         id
       }
    }
  }
}
```

Is that ok for you @patrys @maarcingebala or maybe you have other ideas?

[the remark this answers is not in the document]
patrys: It does not have to be a union, see https://docs.saleor.io/docs/3.x/developer/api-reference/interfaces/object-with-metadata

maarcingebala (approved):

saleor/graphql/webhook/subscription_types.py
```diff
@@ -38,6 +43,133 @@
...
+
+    @staticmethod
+    def resolve_recipient(_root, info):
+        return info.context.app
+
+    @staticmethod
+    def resolve_issuing_principal(_root, info):
+        if not info.context.requestor:
+            return None
+        return info.context.requestor
```
IKarbowiak, line 170: Cannot be just:
```suggestion
        return info.context.requestor
```
?

IKarbowiak (approved):

PR #9763, opened 2022-05-13
GraphQL subscription support for synchronous webhook events

jakubkuc:
I want to merge this change because it adds subscription payloads to synchronous webhooks.

Names of subscription event types are slightly changed (added "Event" at the end) since schema could not be generated for duplicated names (we had mutations with the same names). I'm not sure if we want to keep it that way or change the way we name it? 
Also keep in mind that It is done so it doesn't break current flow and synchronous events will have to be further changed when we will want to drop the old way of generating payload. Unfortunately, it was not possible to make the changes so they are easily switched to just subscriptions like with the async webhooks. 

## How should it work:
Old sync events should work as they were. New approach should not brake then in any way. 
Sync Events that should be able to be created using `subscription_query` with access to data that was used in old events:

Those events should have access to `payment` data:
- [ ] PaymentAuthorize
- [ ] PaymentCaptureEvent
- [ ] PaymentRefundEvent
- [ ] PaymentVoidEvent
- [ ] PaymentConfirmEvent
- [ ] PaymentProcessEvent

This one should have access to `checkout` data:
- [ ] PaymentListGateways

This one should have access to `checkout` data:
- [ ] ShippingListMethodsForCheckout

This one should have access to `checkout` data and `shipping_methods`
- [ ] CheckoutFilterShippingMethods

This one should have access to `order` data and `shipping_methods`
- [ ] OrderFilterShippingMethods

## Edge cases: 
⚠️  For sync events there are fields that are blocked to prevent circular calls. Those fields should return errors in payload if they are defined in `subscription_query`: 
`checkout.shipping_methods`
`checkout.available_shipping_methods`
`checkout.available_payment_gateways`
`order.shipping_methods`
`order.available_shipping_methods`

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-6294 Handle sync webhooks](https://app.clickup.com/t/2549495/SALEOR-6294)

IKarbowiak (approved):

SzymJ (approved):

saleor/graphql/webhook/subscription_types.py
```diff
@@ -995,6 +1008,68 @@ def resolve_warehouse(root, _info):
...
+    class Meta:
+        interfaces = (Event,)
+
+
+class ShippingListMethodsForCheckout(ObjectType, CheckoutBase):
+    class Meta:
+        interfaces = (Event,)
+
+
+class CheckoutFilterShippingMethods(ObjectType, CheckoutBase):
```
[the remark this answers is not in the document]
fowczarek: Same situation for `PAYMENT_LIST_GATEWAYS`, `ORDER_FILTER_SHIPPING_METHODS`, `SHIPPING_LIST_METHODS_FOR_CHECKOUT`

saleor/graphql/webhook/subscription_types.py
```diff
@@ -995,6 +1008,68 @@ def resolve_warehouse(root, _info):
         return warehouse
 
 
+class PaymentBase(AbstractType):
+    payment = graphene.Field(
+        "saleor.graphql.payment.types.Payment",
+        description="Look up a payment." + ADDED_IN_34 + PREVIEW_FEATURE,
```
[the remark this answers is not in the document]
jakubkuc: fixed

saleor/graphql/webhook/subscription_types.py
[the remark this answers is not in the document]
jakubkuc: - added `sync_event` flag to `create_deliveries_for_subscriptions` to block circular calls
- fixed subscription_types fields for `ORDER_FILTER_SHIPPING_METHODS` and `SHIPPING_LIST_METHODS_FOR_CHECKOUT`

saleor/graphql/checkout/types.py
```diff
@@ -549,6 +550,10 @@ def with_checkout_info(checkout_info):
     @classmethod
     @traced_resolver
     def resolve_shipping_methods(cls, root: models.Checkout, info):
+        if hasattr(info.context, "sync_event") and info.context.sync_event:
```
korycins, line 553: I see that we use this in many places. Maybe we could create a single decorator that would be used in all resolvers where we need it?

saleor/core/exceptions.py
```diff
@@ -73,3 +75,7 @@ def __init__(self, message: str):
         super().__init__(message)
         self.message = message
         self.code = CheckoutErrorCode.GIFT_CARD_NOT_APPLICABLE.value
+
+
+class CircularQuery(GraphQLError):
```
korycins, line 80: Is `Query` correct? Without checking where it is used I am not sure that I would guess what is this.
The issue is related only to `Subscription` + `Sync Event`, right? So maybe something like `CircularSubscriptionQuery` or `CircularSubscriptionSyncEvent`? Or something else that will give more  context than only `Query`.

saleor/graphql/webhook/subscription_types.py
```diff
@@ -1077,6 +1103,125 @@ def resolve_warehouse(root, _info):
...
+class ShippingListMethodsForCheckout(ObjectType, CheckoutBase):
+    class Meta:
+        interfaces = (Event,)
+
+
+class CheckoutFilterShippingMethods(ObjectType, CheckoutBase):
+    shipping_methods = NonNullList(
+        ShippingMethod,
+        description="Shipping methods that can be used with this checkout."
+        + ADDED_IN_34
```
korycins, line 1162: It should be 35 right?

saleor/graphql/webhook/subscription_types.py
```diff
@@ -1077,6 +1103,125 @@ def resolve_warehouse(root, _info):
...
+
+    class Meta:
+        interfaces = (Event,)
+
+
+class OrderFilterShippingMethods(ObjectType, OrderBase):
+    shipping_methods = NonNullList(
+        ShippingMethod,
+        description="Shipping methods that can be used with this checkout."
+        + ADDED_IN_34
```
korycins, line 1199: 35

saleor/order/utils.py
```diff
@@ -674,6 +674,8 @@ def get_valid_shipping_methods_for_order(
     order: Order,
     shipping_channel_listings: Iterable["ShippingMethodChannelListing"],
     manager: "PluginsManager",
+    *,
+    exclude_methods: bool = True
```
korycins, line 678: Not sure of this. Like we have a helper method that can return us a `valid_shipping_methods` but now we can get `valid_shipping_methods` (as name function) and the methods that are not exactly valid (by setting `exclude_methods=False`). 
Maybe we could extract the first part of the function's body to separate function and call it inside `get_valid_shipping_methods_for_order(`? Then you could use this newly created function and use it in your scenario, then you will not need `exclude_methods` here

saleor/plugins/webhook/plugin.py
```diff
@@ -1099,7 +1100,10 @@ def __run_payment_webhook(
             )
 
         webhook_payload = generate_payment_payload(payment_information)
-        response_data = trigger_webhook_sync(event_type, webhook_payload, app)
+        payment = Payment.objects.get(id=payment_information.payment_id)
```
korycins, line 1103: It is a detail but wouldn't be better to use `filter().first()` here and confirm that `payment` != None? I know that this is almost not possible but  this would safe us from case when `.get` raises an exception

saleor/plugins/webhook/tasks.py
```diff
@@ -66,7 +66,7 @@ class WebhookResponse:
 
 
 def create_deliveries_for_subscriptions(
-    event_type, subscribable_object, webhooks, requestor=None
+    event_type, subscribable_object, webhooks, requestor=None, sync_event=False
```
korycins, line 69: Do you need this flag? 
Maybe you could just do something like this:
```
request=initialize_request(requestor, webhook.event in WebhookEventSyncType.ALL`),
```
Then you don't need to pass `sync_event` 🤗

saleor/plugins/webhook/tests/subscription_webhooks/test_create_deliveries_for_subscription.py
```diff
@@ -1546,6 +1547,227 @@ def test_voucher_deleted(voucher, subscription_voucher_deleted_webhook):
     assert deliveries[0].webhook == webhooks[0]
 
 
+def test_payment_authorize(payment, subscription_payment_authorize_webhook):
```
korycins, line 1550: IMHO this file is growing very fast. 
I would propose to put all tests related to payments into single file.
And for the rest of the sync webhooks - one file per one event, this should help us in keeping readability.

saleor/plugins/webhook/tests/subscription_webhooks/test_create_deliveries_for_subscription.py
```diff
@@ -1644,3 +1866,30 @@ def test_generate_payload_from_subscription_return_permission_errors_in_payload(
...
+def test_generate_sync_event_payload_with_circular_call(
+    checkout_ready_to_complete,
+    subscription_checkout_filter_shipping_methods_webhook_with_circular_fields,
+):
+
+    # given
+    webhooks = [
+        subscription_checkout_filter_shipping_methods_webhook_with_circular_fields
+    ]
+    event_type = WebhookEventSyncType.CHECKOUT_FILTER_SHIPPING_METHODS
```
korycins, line 1880: IMHO, each resolver that should raise such an exception should have a separate test that confirms that we handle this correctly. This test only checks that query with `checkout.shippingMethods` doesn't go into circular calls.

saleor/tests/fixtures.py
```diff
@@ -440,6 +441,19 @@ def checkout_with_item_and_shipping_method(checkout_with_item, shipping_method):
...
+@pytest.fixture
+def checkout_with_item_shipping_method_and_shipping_address(
+    checkout_with_item, shipping_method
+):
+    checkout = checkout_with_item
+    checkout.shipping_method = shipping_method
+    checkout.save()
+    return checkout
+
+    checkout.shipping_address = address.get_copy()
```
SzymJ, line 453: there is `return` before those lines

saleor/graphql/checkout/types.py
jakubkuc: great idea. fiuxed

saleor/tests/fixtures.py
jakubkuc: this fixture was supposed to be deleted. good catch. thx

saleor/core/exceptions.py
jakubkuc: 'CircularSubscriptionSyncEvent' sounds good :)

saleor/plugins/webhook/plugin.py
jakubkuc: good point, fixed

saleor/plugins/webhook/tasks.py
jakubkuc: brilliant, thanks!

saleor/plugins/webhook/tests/subscription_webhooks/test_create_deliveries_for_subscription.py
jakubkuc: I totally agree - though I would leave it for another PR to make review easier and shiorter ATM

saleor/plugins/webhook/tests/subscription_webhooks/test_create_deliveries_for_subscription.py
jakubkuc: Agree. Fixed

saleor/graphql/webhook/subscription_types.py
jakubkuc: Good catch we should look into this

saleor/order/utils.py
jakubkuc: fixed

saleor/graphql/webhook/subscription_types.py
```diff
@@ -1077,6 +1103,123 @@ def resolve_warehouse(root, _info):
         return warehouse
 
 
+class PaymentBase(AbstractType):
+    payment = graphene.Field(
+        "saleor.graphql.payment.types.Payment",
+        description="Look up a payment." + ADDED_IN_35 + PREVIEW_FEATURE,
```
korycins, line 1109: I know that we have a mess in the rest of the events but I don't see any reason why we couldn't clean it up for new events. I mean:
`ADDED_IN_35 + PREVIEW_FEATURE` (Should be 36), should be added on each event lvl, not on a field lvl. 
Base shouldn't have ANY labels like `ADDED_XX` or `PREVIEW_FEATURE` as this type is commonly shared between different events that can be introduced in a different version of Saleor.

saleor/order/utils.py
```diff
@@ -702,7 +697,22 @@ def get_valid_shipping_methods_for_order(
...
+
+
+def get_valid_shipping_methods_for_order(
+    order: Order,
+    shipping_channel_listings: Iterable["ShippingMethodChannelListing"],
+    manager: "PluginsManager",
+) -> List[ShippingMethodData]:
+    """Return a list of shipping methods according to Saleor's own business logic.
+
+    The resulting methods are not yet filtered by plugins.
```
korycins, line 711: Is it correct? See lines 717 and 718. Look like we already filter out them

korycins (commented):
@jakubkuc 
What kind of the feedback user will get when sync subsription with not allowed fields will be triggered?
Maybe I am missing something but I don't see any changes inside `saleor.graphql.webhook.subscription_payload.generate_payload_from_subscription`, which means that we don't handle an circular exception in specific way. Which means that, we will skip building event payload and we will only raise a generic log.warning. Also we won't create any `EventPayload`/`EventDeliveries` so at the end, user will not get any information that we tried to trigger his webhook event.

saleor/order/utils.py
jakubkuc: I haven't noticed since I didn't change the logic and left the docstring as it was. I agree, the description is confusing since it filters excluded methods.

jakubkuc:
@korycins At the moment if there is a field that is not allowed in sync event subscription it will raise a `CircularSubscriptionSyncEvent(GraphQLError)` during payload generation and will return `errors` in payload (like any other error with subscription query). As far as I remember there was an agreement that this solution was enough for now. But in the future if there will be anything wrong with the subscription query it will be validated during webhook/event creation process and not when we generate payload.

saleor/graphql/webhook/subscription_types.py
jakubkuc: fixed it of new events

fowczarek (approved):

korycins (approved):

saleor/graphql/checkout/utils.py
```diff
@@ -13,3 +16,15 @@ def prepare_insufficient_stock_checkout_validation_error(exc):
         code=exc.code.value,
         params={"variants": variant_ids},
     )
+
+
+def prevent_sync_event_circular_query(func):
```
maarcingebala, line 21: We could add some docstring that would explain briefly the problem we are solving here i.e. why it's not allowed to resolve it in sync events.

saleor/graphql/checkout/utils.py
```diff
@@ -13,3 +16,15 @@ def prepare_insufficient_stock_checkout_validation_error(exc):
...
+    def wrapper(*args, **kwargs):
+        info = next(arg for arg in args if isinstance(arg, ResolveInfo))
+        if hasattr(info.context, "sync_event") and info.context.sync_event:
```
maarcingebala, line 24: Could be shortened to `getattr(info.context, "sync_event", False)` I guess?

saleor/graphql/checkout/utils.py
```diff
@@ -13,3 +16,21 @@ def prepare_insufficient_stock_checkout_validation_error(exc):
         code=exc.code.value,
         params={"variants": variant_ids},
     )
+
+
+def prevent_sync_event_circular_query(func):
+    """Prevent from resolving field in a synchronous event.
+
+    Synchronous events are not allowed to request fields that are resolved using
+    synchronous events. This prevents circular events call.
```
maarcingebala, line 25: I would rephrase that to be more precise:

```
"""Prevent circular dependencies in synchronous events resolvers.

Synchronous events are not allowed to request fields that are resolved using other synchronous events, which would lead to circular calls of the webhook. Using this decorator prevents such circular events resolution.

:raises CircularSubscriptionSyncEvent: When a field being resolved from a synchronous webhook's payload uses another synchronous webhook internally.
"""
```

jakubkuc: fixed

maarcingebala (approved):

PR #9781, opened 2022-05-17
Fix creating dummy context for subscription webhooks.

SzymJ:
I want to merge this change because it fixes creating context used in subscription webhooks.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/webhook/subscription_payload.py
+++ b/saleor/graphql/webhook/subscription_payload.py
@@ -1,6 +1,7 @@
 from typing import Any, Dict, Optional
 
 from celery.utils.log import get_task_logger
+from django.conf import settings
 from django.core.handlers.base import BaseHandler
 from django.http import HttpRequest
 from django.test.client import RequestFactory
@@ -54,10 +55,11 @@ def initialize_context() -> HttpRequest:
     return: HttpRequest
     """
     handler = BaseHandler()
-    context = RequestFactory().request()
+    context = RequestFactory().request(**{"SERVER_NAME": settings.ALLOWED_HOSTS[-1]})
     handler.load_middleware()
     response = handler.get_response(context)
-    assert response.status_code == 200
+    if not response.status_code == 200:
+        raise Exception("Unable to initialize context for webhook.")
     return context
 
 
--- a/saleor/graphql/webhook/subscription_payload.py
+++ b/saleor/graphql/webhook/subscription_payload.py
@@ -1,7 +1,6 @@
 from typing import Any, Dict, Optional
 
 from celery.utils.log import get_task_logger
-from django.conf import settings
 from django.core.handlers.base import BaseHandler
 from django.http import HttpRequest
 from django.test.client import RequestFactory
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [SALEOR-6981 Webhook Create with custom query is not working](https://app.clickup.com/t/2549495/SALEOR-6981)

saleor/graphql/webhook/subscription_payload.py
```diff
@@ -54,10 +55,11 @@ def initialize_context() -> HttpRequest:
     return: HttpRequest
     """
     handler = BaseHandler()
-    context = RequestFactory().request()
+    context = RequestFactory().request(**{"SERVER_NAME": settings.ALLOWED_HOSTS[-1]})
```
maarcingebala, line 58: Why -1? See how we [set up this value for WSGI app](https://github.com/saleor/saleor/blob/ebc0e0e66168c358373d594c0c50865ef8e6507f/saleor/wsgi/__init__.py#L37), maybe we should reuse it here.

saleor/graphql/webhook/subscription_payload.py
```diff
@@ -54,10 +56,13 @@ def initialize_context() -> HttpRequest:
     return: HttpRequest
     """
     handler = BaseHandler()
-    context = RequestFactory().request()
+    context = RequestFactory().request(
+        **{"SERVER_NAME": SimpleLazyObject(lambda: settings.ALLOWED_HOSTS[0])}
```
korycins, line 60: Just to confirm. Are you sure, that this is OK to provide only the first value of allowed hosts?
Why do we need a SimpleLazyObject? Is it requirement for request?

patrys: Also why create a dict to immediately unpack it to keyword args?

SzymJ: I think that in this case there is no matter which element of `ALLOWED_HOSTS` we choose, @maarcingebala suggested in the same way as in[ WSGI app](https://github.com/saleor/saleor/blob/ebc0e0e66168c358373d594c0c50865ef8e6507f/saleor/wsgi/__init__.py#L37) so I did that.

maarcingebala: @patrys Didn't we want to stop relying on domains from site settings and deprecate/drop the API to set it?

maarcingebala: OK so let's use site_settings then.

SzymJ: Are domains from Site objects always added to ALLOWED_HOSTS?

maarcingebala (approved):

saleor/graphql/webhook/subscription_payload.py
patrys: They better be or Saleor won't work 😄

IKarbowiak (approved):

PR #9789, opened 2022-05-18
Optimize update_products_search_vector to work in batches

tomaszszymanski129:

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/product/search.py
+++ b/saleor/product/search.py
@@ -26,13 +26,23 @@ PRODUCT_FIELDS_TO_PREFETCH = [
 
 
 def update_products_search_vector(products: "QuerySet"):
-    products = products.prefetch_related(*PRODUCT_FIELDS_TO_PREFETCH)
-    for product in products:
-        product.search_vector = prepare_product_search_vector_value(
-            product, already_prefetched=True
+    batch_size = 50
+    last_id = 0
+    while True:
+        products_batch = list(
+            products.order_by("id").filter(id__gt=last_id)[:batch_size]
         )
+        if not products_batch:
+            break
+        last_id = products_batch[-1].id
 
-    Product.objects.bulk_update(products, ["search_vector", "updated_at"])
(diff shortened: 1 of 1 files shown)


saleor/core/search_tasks.py
```diff
@@ -15,7 +15,7 @@
 
 task_logger = get_task_logger(__name__)
 
-BATCH_SIZE = 1000
```
patrys, line 18: Why?

saleor/product/search.py
```diff
@@ -26,13 +26,23 @@
 
 
 def update_products_search_vector(products: "QuerySet"):
-    products = products.prefetch_related(*PRODUCT_FIELDS_TO_PREFETCH)
-    for product in products:
-        product.search_vector = prepare_product_search_vector_value(
-            product, already_prefetched=True
+    batch_size = 300
```
patrys, line 29: Why?

saleor/core/search_tasks.py
tomaszszymanski129: Reduced it by half because we found out it was timing out with the larger batch.

saleor/product/search.py
tomaszszymanski129: That results in memory usage up to ~350MB from what i've tested locally, seems ok, wdyt?

saleor/core/search_tasks.py
patrys: It's a magic number that someone will change next week. If it needs to be 500, you have to add a long comment explaining why it needs to be 500 and how to retest it in a year to ensure it's still the correct threshold. Otherwise, it'll either get randomly changed back to 1000 or stuck at 500 for ten years from the fear of breaking something.

saleor/product/search.py
patrys: Same as above, without comments, it's just an arbitrary magic number.

maarcingebala (approved):

IKarbowiak (approved):

PR #9796, opened 2022-05-19
Update sample products set

mirekm:
I want to merge this change because it provides a more consistent simple products set to play with.


mirekm:
Thanks for the help with this PR @korycins and @patrys! 🙏 
Kudos to @subgrafik for the new graphics! 🔥

It seems to be ready.

saleor/core/utils/random_data.py
```diff
@@ -200,23 +198,22 @@ def create_categories(categories_data, placeholder_dir):
         pk = category["pk"]
         defaults = category["fields"]
         parent = defaults["parent"]
-        image_name = (
-            CATEGORY_IMAGES[pk] if pk in CATEGORY_IMAGES else CATEGORY_IMAGES[parent]
-        )
-        background_image = get_image(placeholder_dir, image_name)
-        defaults["background_image"] = background_image
+        image_name = CATEGORY_IMAGES[pk] if pk in CATEGORY_IMAGES else None
```
korycins, line 201: CATEGORY_IMAGES is a dict. instead of doing:
`CATEGORY_IMAGES[pk] if pk in CATEGORY_IMAGES else None`
I think that you could simply call `CATEGORY_IMAGES.get(pk)`

saleor/core/utils/random_data.py
```diff
@@ -228,17 +225,18 @@ def create_collections(data, placeholder_dir):
     for collection in data:
         pk = collection["pk"]
         defaults = collection["fields"]
-        image_name = COLLECTION_IMAGES[pk]
-        background_image = get_image(placeholder_dir, image_name)
-        defaults["background_image"] = background_image
+        image_name = COLLECTION_IMAGES[pk] if pk in COLLECTION_IMAGES else None
```
korycins, line 228: Same as above 🤔

saleor/core/utils/random_data.py
```diff
@@ -767,7 +756,13 @@ def create_fulfillments(order):
 
 
 def create_fake_order(discounts, max_order_lines=5, create_preorder_lines=False):
-    channel = Channel.objects.all().order_by("?").first()
+    channel = (
+        Channel.objects.filter(
+            Q(slug=settings.DEFAULT_CHANNEL_SLUG) | Q(slug="channel-pln")
```
korycins, line 761: I think that you could simplify this query as you can handle it without Q object:
```
Channel.objects.filter(slug__in=[settings.DEFAULT_CHANNEL_SLUG, "channel-pln"]).order_by("?")
         .first()
```

saleor/core/utils/random_data.py
```diff
@@ -920,20 +915,23 @@ def _create_staff_user(staff_password, email=None, superuser=False):
...
-        default_billing_address=address,
-        default_shipping_address=address,
-        is_staff=True,
-        is_active=True,
-        is_superuser=superuser,
-        search_document=_prepare_search_document_value(
-            User(email=email, first_name=first_name, last_name=last_name), address
-        ),
-    )
+    staff_user = User.objects.filter(email=email).first()
```
korycins, line 918: Not sure if I understand. There is the same query at the beginning of this function. That call `return` in case when user already exists. Do we need that part in that case? In this moment `staff_user` should be always `None`. (or maybe I am missing something 😄 )

saleor/core/utils/random_data.py
```diff
@@ -1438,7 +1434,9 @@ def create_vouchers():
 
 
 def create_gift_cards(how_many=5):
-    product_pk = Product.objects.get(name="Gift card").pk
+    product_pk = Product.objects.get(name="Gift card 100").pk
```
korycins, line 1437: `get` will raise `DoesNotExist` so it will never reach `if not product_pk:` condition. `.filter(name="Gift...").first()` would be a better option

krzysztofwolski (changes requested):
Populatedb does not assign product attributes:
image

The reason is probably a change in how we keep attribute values. "The old style" - values array:

```
{
    "model": "attribute.attributeproduct",
    "pk": 14,
    "fields": { "sort_order": 0, "attribute": 40, "product_type": 23, "values": [123] }
  },
```
"The new style" - separate model:

```
  {
    "model": "attribute.assignedvariantattributevalue",
    "pk": 677,
    "fields": { "sort_order": 0, "value": 108, "assignment": 137 }
  },
  {
    "model": "attribute.assignedvariantattributevalue",
    "pk": 678,
    "fields": { "sort_order": 0, "value": 109, "assignment": 138 }
  },
 ```

examples are from the json dump file.

File: saleor/core/utils/random_data.py (hunk 4 above)
Comment (mirekm) [reply saleor/core/utils/random_data.py:918]: Yes, there's an `if` in between when email is not provided. Previously, it was breaking when there already was a user in DB for the newly generated email.

File: saleor/core/utils/random_data.py (hunk 5 above)
Comment (mirekm) [reply saleor/core/utils/random_data.py:1437]: Yeah, this code is not great whatsoever and likely requires more work (including, for instance, how attribute values are assigned), my aim was to do minimal changes to make it work with the new products.

Comment (mirekm) [reply saleor/core/utils/random_data.py:1437]: Fixed.

**Comment by mirekm** [comment]
> Populatedb does not assign product attributes

Correct, I saw it but I won't be able to fix that one @krzysztofwolski

File: saleor/core/utils/random_data.py (hunk 2 above)
Comment (mirekm) [reply saleor/core/utils/random_data.py:228]: Fixed.

File: saleor/core/utils/random_data.py (hunk 1 above)
Comment (mirekm) [reply saleor/core/utils/random_data.py:201]: Fixed.

**Comment by krzysztofwolski** [comment]
@mirekm I'll prepare a fix for attributes, I'll need such one as well

**Comment by mirekm** [comment]
> @mirekm I'll prepare a fix for attributes, I'll need such one as well

Perfect, thanks @krzysztofwolski

File: saleor/core/utils/random_data.py (hunk 4 above)
Comment (IKarbowiak) [reply saleor/core/utils/random_data.py:918]: So it seems like instead of looking for an existing user two times, we can just leave the second query and drop the query and check if the user exists from lines 909 to 911.

**Review by IKarbowiak — COMMENTED** [review]
Also, two tests are failing.

**Comment by krzysztofwolski** [comment]
@mirekm 
Product attrs
image

Variant attrs
image

File: saleor/core/utils/random_data.py [hunk 6]
```diff
@@ -7,22 +7,24 @@
...
+from django.db.models import F
 from django.utils import timezone
 from django.utils.text import slugify
 from faker import Factory
 from faker.providers import BaseProvider
 from measurement.measures import Weight
 from prices import Money, TaxedMoney
 
+from saleor.attribute.models.product import AssignedProductAttributeValue
+from saleor.attribute.models.product_variant import AssignedVariantAttributeValue
```
maarcingebala, line 26: Use relative imports 🚓  🚓

maarcingebala (approved):

korycins (approved):

eCzerniak (approved):

PR #9799, opened 2022-05-19
Fix inaccurate tax calculations

IKarbowiak:
Update `vatlayer` and `avatax` tax plugins - apply `discounts` on checkout and order lines before the taxes are calculated.

In the case of `avatax` the discount amount is attached to the request data, and the lines that should be included in discounts are marked with the use of the `discounted` flag.

In the case of `vatlayer` the discount is propagated proportionally to the ratio of the total line price to the total order price.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-6633 Inaccurate tax calculations when 100% voucher is applied](https://app.clickup.com/t/2549495/SALEOR-6633)

saleor/plugins/vatlayer/__init__.py
```diff
@@ -89,3 +98,131 @@ def get_taxed_shipping_price(shipping_price, taxes):
...
+    return quantize_price(
+        (line_total_price - Money(discount_amount, currency)) / order_line.quantity,
+        currency,
+    )
+
+
+def _calculate_discount_for_last_element_in_qs(
+    lines_total_prices, total_price, total_discount_amount, currency
+):
+    """Calculate the discount for last element of query set.
```
fowczarek, line 213: Could we fix names here? Using `qs` and `query set` could be confusing.

fowczarek (approved):

saleor/plugins/avatax/plugin.py
```diff
@@ -261,6 +253,7 @@ def _calculate_checkout_shipping(
         for line in lines:
             if line["itemCode"] == "Shipping":
                 shipping_net = Decimal(line["lineAmount"])
+                discount_amount = Decimal(line.get("discountAmount", 0.0))
```
korycins, line 256: Can you add here a comment, that `lineAmount` doesn't include an applied `discoutnAmount` but `tax` is calculated for:  net == `lineAmount` - `discountAmount`? I think this could be helpfull especially after a checking belowed lines, like:
```
shipping_net -= discount_amount
```

saleor/plugins/vatlayer/__init__.py
```diff
@@ -89,3 +98,131 @@ def get_taxed_shipping_price(shipping_price, taxes):
...
+    checkout_line_info: "CheckoutLineInfo",
+    discounts: Iterable["DiscountInfo"],
+    line_price: Money,
+):
+    """Calculate the checkout line price with discounts.
+
+    Include the entire order voucher discount.
+    """
+    voucher = checkout_info.voucher
+    if not voucher or voucher.type == VoucherType.SHIPPING:
```
korycins, line 115: Is it okay that we will take into account a voucher of type `SPECIFIC_PRODUCTS`? ☝️

saleor/plugins/vatlayer/__init__.py
```diff
@@ -89,3 +98,131 @@ def get_taxed_shipping_price(shipping_price, taxes):
...
+
+    line_quantity = checkout_line_info.line.quantity
+    total_discount_amount = checkout_info.checkout.discount_amount
+    line_total_price = line_price * line_quantity
+    currency = checkout_info.checkout.currency
+
+    # if the checkout has only one line, the hole discount amount will be applied
+    # to this line
+    if len(lines) == 1:
+        return (
```
korycins, line 126: Shouldn't we check here if returned price is not < 0 ?Like
```
max(line_total_price, zero)
```

saleor/plugins/vatlayer/__init__.py
```diff
@@ -89,3 +98,131 @@ def get_taxed_shipping_price(shipping_price, taxes):
...
+            line_total_price - Money(total_discount_amount, currency)
+        ) / line_quantity
+
+    # if the checkout has more lines we need to propagate the discount amount
+    # proportionally to total prices of items
+    lines_total_prices = [
```
korycins, line 132: Maybe I am missing something, but would be faster and easier to define the discount that should be applied for single value of given currency (like 1 USD, or 1PLN)?
Something like this:
  ```
  line_1 = Decimal(10)
  line_2 = Decimal(80)
  subtotal = line_1 + line_2
  discount_percent_value = 0.5 # 50%
  discount = subtotal * discount_percent_value
  discount_per_single_unit = discount/subtotal # 45/90 == 0.5 USD per 1USD
  discount_line_1 = line_1 * discount_per_single_unit
  discount_line_2 = line_2 * discount_per_single_unit
  ```
  I assume that maybe I am missing something but thought that it would be good to mention

saleor/plugins/vatlayer/__init__.py
```diff
@@ -89,3 +98,131 @@ def get_taxed_shipping_price(shipping_price, taxes):
...
+        not order.voucher_id
+        or order.voucher.type == VoucherType.SHIPPING  # type: ignore
+    ):
+        return order_line.base_unit_price
+
+    currency = order.currency
+    total_discount_amount = get_total_order_discount_excluding_shipping(order).amount
+    line_total_price = order_line.base_unit_price * order_line.quantity
+
+    # if the order has only one line, the hole discount amount will be applied
```
patrys, line 177:
```suggestion
    # if the order has a single line, the whole discount amount will be applied
```

saleor/plugins/vatlayer/__init__.py
```diff
@@ -89,3 +98,131 @@ def get_taxed_shipping_price(shipping_price, taxes):
...
+        return (
+            line_total_price - Money(total_discount_amount, currency)
+        ) / order_line.quantity
+
+    order_lines = order.lines.all()
+
+    # if the order has more lines we need to propagate the discount amount
+    # proportionally to total prices of items
+    lines_total_prices = [
+        line.base_unit_price.amount * line.quantity
```
patrys, line 189: I don't like the fact that we reimplement the logic of how line totals are calculated here

saleor/plugins/vatlayer/__init__.py
```diff
@@ -89,3 +98,131 @@ def get_taxed_shipping_price(shipping_price, taxes):
...
+        for line in order_lines
+        if line.id != order_line.id
+    ]
+    total_price = sum(lines_total_prices) + line_total_price.amount
+
+    last_element = order_lines.last().pk == order_line.pk  # type: ignore
```
patrys, line 195: Isn't `order_lines.last()` executing a query?

saleor/plugins/vatlayer/__init__.py
```diff
@@ -89,3 +98,131 @@ def get_taxed_shipping_price(shipping_price, taxes):
...
+    if last_element:
```
patrys, line 196: This tells me we're implementing this logic on the wrong layer. Logic executed for a single line should not have to consult any of the other lines, let alone fetch them all from the database. We need a function that operates on the entire order, that can loop over the lines and distribute whatever discount there is.

saleor/plugins/vatlayer/__init__.py
IKarbowiak: Yeah, I also was not sure if we should do it in this way. But it's not exactly reimplementing the logic of calculating order line totals in vatlayer, as to get the total line value we are multiplying the price with tax applied, not base price. In case of calculating discount we take into account base prices.

fowczarek (approved):

saleor/core/utils/random_data.py
```diff
@@ -632,6 +632,7 @@ def create_order_lines(order, discounts, how_many=10):
         )
         warehouse = next(warehouse_iter)
         increase_stock(line, warehouse, line.quantity, allocate=True)
+    manager.update_taxes_for_order_lines(order, lines)
```
korycins, line 635: Are you sure of this? If someone will run populateDB after `cleardb` this will generate a huge amount of requests to external API. I would consider having dummy tax rates here instead of risking that we will call external API

saleor/core/utils/random_data.py
```diff
@@ -692,6 +693,7 @@ def create_order_lines_with_preorder(order, discounts, how_many=1):
                 quantity=line.quantity,
             )
         )
+    manager.update_taxes_for_order_lines(order, lines)
```
korycins, line 696: Same as above

saleor/plugins/avatax/plugin.py
```diff
@@ -431,13 +428,19 @@ def _calculate_checkout_line_total_price(
...
 
+            # The lineAmount does not include the discountAmount, but tax is calculated
+            # for discounted net price, that take into account provided discount.
             tax = Decimal(line.get("tax", 0.0))
+            discount_amount = Decimal(line.get("discountAmount", 0.0))
             net = Decimal(line["lineAmount"])
 
             if currency == "JPY" and tax_included():
-                line_gross = base_value.price_with_sale.gross
+                line_gross = Money(
```
korycins, line 438: Please, make sure that the scenario with JPY currency and discount will also be tested during the QA. We should compare the results values from Avatax and Saleor.

saleor/core/utils/random_data.py
fowczarek: This function is implemented only in VatLayer. It does not generate any external API calls during the tax calculation process.

patrys (approved):

korycins (approved):

fowczarek (approved):

maarcingebala (approved):

SzymJ (approved):

korycins (approved):

maarcingebala (approved):

PR #9882, opened 2022-05-30
Observability background worker

przlada:
I want to merge this change because it optimizes the observability reporter. It adds per process background worker, which moves payload generation and Redis buffer uploads to a separate thread. Observability events are gathered and uploaded in batches to reduce the number of requests to the Redis buffer. As a result, the overhead introduced by turning on observability decreases significantly.

Observability overhead measured per Saleor request

|   |  Median |  90% percentile |
| --- | --- | --- |
| No background worker | 9,14 ms  | 21,023 ms |
| Background worker | 0,235 ms  | 0,833 ms |

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/webhook/observability/tests/test_worker.py
+++ b/saleor/webhook/observability/tests/test_worker.py
@@ -4,6 +4,7 @@ from functools import partial
 from unittest.mock import patch
 
 import pytest
+from redis.exceptions import RedisError
 
 from .. import worker
 from .conftest import KEY, MAX_SIZE
@@ -57,6 +58,17 @@ def test_background_worker(mock_buffer_put_multi_key_events, init_worker):
     mock_buffer_put_multi_key_events.assert_called_once_with(
         {buffer_name: [payload.format(i) for i in range(BATCH_SIZE)]}
     )
+
+
+@patch("saleor.webhook.observability.worker.buffer_put_multi_key_events")
+def test_background_worker_put_to_buffer_exception(
+    mock_buffer_put_multi_key_events, init_worker
+):
+    mock_buffer_put_multi_key_events.side_effect = [RedisError, True]
+    for i in range(BATCH_SIZE * 2):
+        worker.put_event("buffer_a", lambda: "payload")
+    time.sleep(safe_delay)
+    mock_buffer_put_multi_key_events.call_count == 2
 
 
 @patch("saleor.webhook.observability.worker.buffer_put_multi_key_events")
--- a/saleor/webhook/observability/tests/test_worker.py
+++ b/saleor/webhook/observability/tests/test_worker.py
@@ -1,4 +1,3 @@
-import time
 from datetime import timedelta
 from functools import partial
 from unittest.mock import call, patch
@@ -7,9 +6,8 @@ import pytest
 from redis.exceptions import RedisError
 
 from .. import worker
-from .conftest import KEY, MAX_SIZE
+from .conftest import BATCH_SIZE, KEY
 
-BATCH_SIZE = 10
 QUEUE_SIZE = worker.BackgroundWorker.queue_size(BATCH_SIZE)
 REPORT_PERIOD = timedelta(seconds=0.01)
 JOIN_TIMEOUT = REPORT_PERIOD.total_seconds() * 2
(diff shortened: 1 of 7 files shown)


saleor/webhook/observability/tests/test_worker.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+@patch("saleor.webhook.observability.worker.buffer_put_multi_key_events")
+def test_background_worker_put_to_buffer_exception(
+    mock_buffer_put_multi_key_events, init_worker
+):
+    mock_buffer_put_multi_key_events.side_effect = [RedisError, True]
+    for i in range(BATCH_SIZE * 2):
+        worker.put_event("buffer_a", lambda: "payload")
+    time.sleep(safe_delay)
```
korycins, line 70: IMHO I don't like using `time.sleep` in the logic/tests. There is always a risk, that the sleep will not be enough or something. I think that we need to consider better solution for testing it.

przlada: Agree, working on it 👍

korycins (approved):

IKarbowiak (approved):

saleor/webhook/observability/worker.py
```diff
@@ -0,0 +1,190 @@
...
+            logger.debug("[Observability] %d events pending on flush", pending)
+            if not self.queue_join(timeout - initial_timeout):
+                pending = self._queue.qsize()
+                logger.error(
+                    "[Observability] Flush timed out, dropped %s events", pending
+                )
+                return False
+        return True
+
+    def flush(self, timeout: float) -> bool:
```
NyanKiyoshi, line 179: That name is confusing, I expect the queue to be flushed but the worker thread to remain alive. I would suggest having `flush()` but the shutdown part to be `shutdown()` method

saleor/webhook/observability/worker.py
```diff
@@ -0,0 +1,190 @@
...
+                "[Observability] Buffer %s full, %s event(s) dropped",
+                buffer_name,
+                events_count,
+            )
+    return all_events_delivered
+
+
+class BackgroundWorker:
+    def __init__(self, batch_size: int, timeout: float):
+        self._batch_size = max(batch_size, 1)
```
NyanKiyoshi, line 70: Why not just `self._batch_size = batch_size`? Seems rather odd to silently change the value

saleor/webhook/observability/worker.py
```diff
@@ -0,0 +1,190 @@
...
+            batch_size=settings.OBSERVABILITY_BUFFER_BATCH_SIZE,
+            timeout=settings.OBSERVABILITY_REPORT_PERIOD.total_seconds(),
+        )
+        atexit.register(shutdown)
+
+
+def put_event(buffer_name: str, generate_payload: Callable[[], Any]) -> bool:
+    if _worker:
+        return _worker.submit_event(buffer_name, generate_payload)
+    logger.warning("[Observability] Worker not initialized, event dropped")
```
NyanKiyoshi, line 44: I would say putting `[Observability]` in each log is unnecessary, we can already see the module name in JSON logs. But it doesn't hurt to make it clear, so I have mixed feelings about it

NyanKiyoshi (approved):
Great job!

patrys:
What is the concurrency model proposed? How does it interact with ASGI, and how will it interact with `asyncio` where there is no process isolation between workers? What's the memory cost of this performance optimization?

saleor/webhook/observability/worker.py
patrys: Please don't prefix log messages. If you want those for local debugging, use a custom format for the console logger.

In production, Saleor outputs logs in JSON format. Maybe you want to add something to `extra`, like the type of event that was dropped?

przlada:
> What is the concurrency model proposed? How does it interact with ASGI, and how will it interact with `asyncio` where there is no process isolation between workers? What's the memory cost of this performance optimization?

There is one observability worker thread started per Django process. Django threads communicate with it using a shared thread-safe queue. The observability worker thread empties the queue and uploads them to the Redis buffer when the time interval has passed, or the events batch has been collected. 

It is working and is tested in ASGI and WSGI mode. Initially, I started implementing observability as asyncio task, but decided to begin with threading as it is needed for WSGI mode and can be used by both. I will be glad to try implementing the async version for ASGI as a separate task.

Memory cost should be low. It comes from retaining Django requests and responses for a brief moment in memory. They are stored in a queue before the observability worker processes them. But based on my measurements, even a super simple query to Saleor is processed a couple of times longer than the observability worker generates and uploads events, so the shared queue stays near empty even in case of multiple concurrent clients generating >80 requests per second to Saleor. I tested it locally with simple and more complex (returning bigger responses) Saleor queries and measured a maximum of ~9% increase in peak memory consumption. I am preparing a better test suite to measure it and will present more detailed data.

saleor/webhook/observability/worker.py
przlada: I agree. Fixing that 👍

saleor/webhook/observability/worker.py
przlada: I agree. Fixing that 👍

saleor/webhook/observability/worker.py
przlada: Sorry, I automatically copied the convention from webhooks plugin tasks `"[Webhook] Sending payload to %r for event %r."`. I will fix it and use `extra` while logging, thanks!

patrys (approved):

maarcingebala:
Closing for now, we will reopen or create a new PR when getting back to this topic.

PR #9904, opened 2022-06-02
Disable webhook when reponse is 4XX

tomaszszymanski129:
I want to merge this change because it immediately disables webhook if it's http response status code is 4XX.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/plugins/webhook/tasks.py
+++ b/saleor/plugins/webhook/tasks.py
@@ -351,6 +351,13 @@ def send_webhook_request_async(self, event_delivery_id):
                 delivery.event_type,
                 attempt.id,
             )
+            if status_code := getattr(response, "response_status_code", None):
+                if 400 <= status_code < 500:
+                    # endpoint returned 4XX code, we should disable the webhook.
+                    webhook.is_active = False
+                    webhook.save(update_fields=["is_active"])
+                    delivery_update(delivery, EventDeliveryStatus.FAILED)
+                    return
             try:
                 countdown = self.retry_backoff * (2**self.request.retries)
                 self.retry(countdown=countdown, **self.retry_kwargs)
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [SALEOR-6655 Disable webhook endpoint after N failed attempts to deliver it](https://app.clickup.com/t/2549495/SALEOR-6655)

maarcingebala (commented):
We should remember about updating the docs about this behavior. Also how would users learn about their webhook being disabled by us? I think it should come with some UI changes in the dashboard (or observability app?) or some notification (e.g. by email or visual in the dashboard).

saleor/plugins/webhook/tasks.py
```diff
@@ -351,6 +351,13 @@ def send_webhook_request_async(self, event_delivery_id):
                 delivery.event_type,
                 attempt.id,
             )
+            if status_code := getattr(response, "response_status_code", None):
+                if 400 <= status_code < 500:
+                    # endpoint returned 4XX code, we should disable the webhook.
+                    webhook.is_active = False
+                    webhook.save(update_fields=["is_active"])
```
patrys, line 358: Can we also somehow store why it was disabled? It's going to be a pain to debug something that automatically turns itself off without giving a reason.

tomaszszymanski129:
I agree we should both have this stored and user should get notified. I suppose we could:
1) send an email informing user that his webhook has been disabled and why.
2) save info in the logs when that happens, ex. `logging.info("Webhook ID X has been disabled due to {code} response.")` so we can filter it out from logs in case we need to.
3) update the docs

What do you think @maarcingebala @patrys ?

patrys:
@tomaszszymanski129 This needs to be available through the API and visible in the dashboard. Customers don't have access to logs and neither do app authors. An app should be able to inspect its own webhooks, see why they were disabled, and re-enable them if needed. We could even try to inform the app of webhooks getting disabled using the same URL we use for installation events.

korycins (commented):
What about changes in `saleor.graphql.webhook.mutations.EventDeliveryRetry`?
This mutation is used to retry trigger the webhook. If we change `is_active` to False,  we will not be able to call it without activating Webhook first. Is it expected?

tomaszszymanski129:
Ticket needs further specification and this cannot be merged as is, moving to backlog with link to this PR.

PR #10035, opened 2022-06-21
New events related to attribute values changes

SzymJ:
I want to merge this change because it adds new subscription base webhooks for attribute value create, update, delete events.

New webhook events:
- `ATTRIBUTE_VALUE_CREATED`
- `ATTRIBUTE_VALUE_UPDATED`
- `ATTRIBUTE_VALUE_DELETED`

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-7223 Add attribute values webhook events](https://app.clickup.com/t/2549495/SALEOR-7223)

saleor/plugins/webhook/tests/subscription_webhooks/payloads.py
```diff
@@ -34,6 +34,18 @@ def generate_attribute_payload(attribute):
     )
 
 
+def generate_attribute_value_payload(attribute_value):
+    return json.dumps(
+        {
+            "attributeValue": {
+                "name": attribute_value.name,
+                "slug": attribute_value.slug,
+                "value": attribute_value.value,
```
maarcingebala, line 43: What about other fields that hold values for other data types, e.g. rich/plain text dates etc? Maybe we don't need it because IIRC for some of the types we store their value in `slug` anyway, but I'm not 100% sure.

saleor/graphql/attribute/mutations.py
```diff
@@ -864,5 +867,9 @@ def perform_mutation(cls, _root, info, attribute_id, moves):
         with traced_atomic_transaction():
             perform_reordering(values_m2m, operations)
         attribute.refresh_from_db(fields=["values"])
+
+        for value in values_m2m.filter(id__in=operations.keys()):
```
maarcingebala, line 871: Do we need to query the DB here? Maybe filtering in Python would be enough. Not sure how much data we can have here.

SzymJ: `values_m2m` is a `RelatedManager` object so I cant use python `filter()` on it

saleor/plugins/webhook/tests/subscription_webhooks/payloads.py
SzymJ: its just some random fields that I'm using only in tests, if you want I can add more

IKarbowiak (approved):
Looks good for me. I'm just wonder if we shouldn't split `saleor/plugins/webhook/tests/subscription_webhooks/test_create_deliveries_for_subscription.py` file somehow. Right now it has 1625 lines and it's getting bigger and bigger.

SzymJ:
> Looks good for me. I'm just wonder if we shouldn't split `saleor/plugins/webhook/tests/subscription_webhooks/test_create_deliveries_for_subscription.py` file somehow. Right now it has 1625 lines and it's getting bigger and bigger.

yep, I'm planning to do that in separate PR when all events will be merged.

saleor/graphql/attribute/mutations.py
maarcingebala: I was thinking about sth similar to this:
```
for value in [v for v in values_m2m if v.id in operations.keys()]
```

saleor/plugins/webhook/tests/subscription_webhooks/payloads.py
maarcingebala: 👍

maarcingebala (approved):

PR #10083, opened 2022-07-01
http.user_id tag in opentracing spans

tomaszszymanski129:
I want to merge this change because it adds `http.user_id` tag to opentracing spans.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [X] The changes are tested

### DIFF
--- a/saleor/graphql/context.py
+++ b/saleor/graphql/context.py
@@ -27,6 +27,7 @@ class RequestWithUser(Protocol):
     _cached_user: UserType
     app: Optional[App]
     user: Union[UserType, SimpleLazyObject]
+    user_id: Optional[int]
 
 
 def get_app(raw_auth_token) -> Optional[App]:
@@ -58,6 +59,8 @@ def set_app_on_context(request):
 def get_user(request: RequestWithUser) -> Optional[UserType]:
     if not hasattr(request, "_cached_user"):
         request._cached_user = cast(UserType, authenticate(request=request))
+        if request._cached_user and not request._cached_user.is_anonymous:
+            request.user_id = cast(int, request._cached_user.id)
     return request._cached_user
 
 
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [SALEOR-7416 Add user_id tag to traces](https://app.clickup.com/t/2549495/SALEOR-7416)

saleor/graphql/context.py
```diff
@@ -58,6 +59,12 @@ def set_app_on_context(request):
 def get_user(request: RequestWithUser) -> Optional[UserType]:
     if not hasattr(request, "_cached_user"):
         request._cached_user = cast(UserType, authenticate(request=request))
+        try:
+            user_id = request._cached_user.id
+        except AttributeError:
+            pass
```
NyanKiyoshi, line 65: In which cases would this exception happen?

patrys:
Why? We explicitly do not trace any information that could be used to profile individual users.

tomaszszymanski129:
@patrys There was a case where we needed to perform an audit of who performed one particular action. There were no logs or events saved so we thought it would be helpful in the future to at least have the user id (like we also trace an app id).

saleor/graphql/context.py
NyanKiyoshi: Can't you just do `getattr`? But it does look like there `id` property available for use: https://github.com/django/django/blob/0dd29209091280ccf34e07c9468746c396b7778e/django/contrib/auth/models.py#L417-L418

tomaszszymanski129: strange, I will double check

patrys:
@tomaszszymanski129 We can't do that as that would track every step of every user, not just staff members. If you want to trace admin actions, use logging (not tracing tags, as those are sampled, and we only save about 1% of them) and only do it in mutations that require admin privileges.

tomaszszymanski129:
@patrys Sounds reasonable. Closing this one. I will prepare subsequent ticket to improve the logging and add the missing event (and others if I find any).

PR #10095, opened 2022-07-04
Multiple copies of a single variant in the same Checkout/Order

SzymJ:
This feature branch contains implementation for the split variant in the multiple checkout lines feature, this will allow dynamic prices to be used to their full extent.
Consider an example where a product can be engraved, and the price depends on the number of letters. A person can't order two such items without placing two separate orders (and paying for shipping twice).

## General assumptions
Introduce the following concepts:
- add `forceNewLine` flag to `CheckoutLineInput` and `OrderLineCreateInput` - this will allow force splitting the same variant into multiple lines by skipping the matching logic
- add `lineId` parameter to `CheckoutLinesUpdate` mutation
  - extend the `CheckoutLineUpdateInput` with an optional `lineId` (allowing a particular line to be targeted)
  - reject `checkoutLinesUpdate` if `variantId` is passed and it matches more than one line
  - deprecate `variantId` parameter

### Inputs

`CheckoutLineInput` and `OrderLineCreateInput`
- New `forceNewLine` flag added

```graphql
input CheckoutLineInput {
  forceNewLine: Boolean = false
  quantity: Int!
  variantId: ID!
  price: PositiveDecimal
}

input OrderLineCreateInput {
  forceNewLine: Boolean = false
  quantity: Int!
  variantId: ID!
}
```

`CheckoutLineUpdateInput`
 - New optional `lineId` parameter added
 - Make `variantId` as optional and depreacted
 
```graphql
input CheckoutLineUpdateInput {
  quantity: Int
  variantId: ID
  price: PositiveDecimal
  lineId: ID
}
```

### Mutations

`DraftOrderCreate` and `OrderLinesCreate`
 - create a new line for line input with `forceNewLine` set to `true`
 
`CheckoutCreate`
 - create a new line for line input with `forceNewLine` set to `true`
 - combine checkout rows when:
   - same variants without custom price passed
   - same variants with the same custom price passed

`CheckoutLinesAdd` 
  - create a new line when `forceNewLine` set to `true`
  - create a new line when new variant is added
  - create a new line when existing variant but with new custom price is added 
  - update line when existing variant with the same custom price is added
  - update line when existing variant without a custom price is added

`CheckoutLinesUpdate`
- use`lineId` to target particular line
- ~~allow to split line with quantity > 1 into multiple lines~~ not needed
- return error if `variantId` is passed and it matches more than one line

`CheckoutLineDelete` and `CheckoutLinesDelete` already support `line ID` as parameter so no changes needed.

`CheckoutLinesUpdate`, `CheckoutLineDelete`, `CheckoutLinesDelete` mutations can be called by `App` or `Customer`. At this moment we would like to block to make those actions via `customer API key` on lines when there is a custom price added or a single variant in multiple lines.

## Additional things to do
- adjust `checkoutLimits` to calculate if limit is exceeded when same variant occurs in multiple lines
- for each unique checkout line, create a separate order line - `orderCreateFromCheckout`

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/fetch.py
+++ b/saleor/checkout/fetch.py
@@ -211,7 +211,9 @@ def get_delivery_method_info(
 
 
 def fetch_checkout_lines(
-    checkout: "Checkout", prefetch_variant_attributes=False
+    checkout: "Checkout",
+    prefetch_variant_attributes=False,
+    skip_lines_with_unavailable_variants=True,
 ) -> Tuple[Iterable[CheckoutLineInfo], Iterable[int]]:
     """Fetch checkout lines as CheckoutLineInfo objects."""
     from .utils import get_voucher_for_checkout
@@ -250,6 +252,17 @@ def fetch_checkout_lines(
             checkout, product, variant_channel_listing, product_channel_listing_mapping
         ):
             unavailable_variant_pks.append(variant.pk)
+            if not skip_lines_with_unavailable_variants:
+                lines_info.append(
+                    CheckoutLineInfo(
+                        line=line,
+                        variant=variant,
+                        channel_listing=variant_channel_listing,
+                        product=product,
+                        product_type=product_type,
+                        collections=collections,
+                    )
+                )
             continue
 
         lines_info.append(
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -127,6 +127,7 @@ def add_variant_to_checkout(
     price_override: Optional["Decimal"] = None,
     replace: bool = False,
     check_quantity: bool = True,
+    force_new_line: bool = False,
 ):
     """Add a product variant to checkout.
 
@@ -153,16 +154,22 @@ def add_variant_to_checkout(
         check_quantity=check_quantity,
     )
 
+    if force_new_line:
+        checkout.lines.create(
+            variant=variant,
+            quantity=quantity,
+            price_override=price_override,
+        )
+        return checkout
+
     if line is None:
         line = checkout.lines.filter(variant=variant).first()
 
     if new_quantity == 0:
         if line is not None:
             line.delete()
-            line = None
     elif line is None:
-        line = checkout.lines.create(  # type: ignore
-            checkout=checkout,
+        checkout.lines.create(  # type: ignore
             variant=variant,
             quantity=new_quantity,
             currency=checkout.currency,
@@ -199,18 +206,20 @@ def add_variants_to_checkout(
     country_code = checkout.get_country()
 
     checkout_lines = checkout.lines.select_related("variant")
-    variant_ids_in_lines = {line.variant_id: line for line in checkout_lines}
+
+    lines_by_id = {str(line.pk): line for line in checkout_lines}
+    variants_map = {str(variant.pk): variant for variant in variants}
+
     to_create: List[CheckoutLine] = []
     to_update: List[CheckoutLine] = []
     to_delete: List[CheckoutLine] = []
-    for variant, line_data in zip(variants, checkout_lines_data):
-        _append_line_to_update(
-            to_update, to_delete, variant, line_data, replace, variant_ids_in_lines
-        )
-        _append_line_to_delete(to_delete, variant, line_data, variant_ids_in_lines)
-        _append_line_to_create(
-            to_create, checkout, variant, line_data, variant_ids_in_lines
-        )
+
+    for line_data in checkout_lines_data:
+        line = _get_line_if_exist(line_data, lines_by_id)
+        _append_line_to_update(to_update, to_delete, line_data, replace, line)
+        _append_line_to_delete(to_delete, line_data, line)
+        _append_line_to_create(to_create, checkout, variants_map, line_data, line)
+
     if to_delete:
         CheckoutLine.objects.filter(pk__in=[line.pk for line in to_delete]).delete()
     if to_update:
@@ -219,6 +228,7 @@ def add_variants_to_checkout(
         CheckoutLine.objects.bulk_create(to_create)
 
     to_reserve = to_create + to_update
+
     if reservation_length and to_reserve:
         updated_lines_ids = [line.pk for line in to_reserve + to_delete]
         for line in checkout_lines:
@@ -238,12 +248,15 @@ def add_variants_to_checkout(
     return checkout
 
 
-def _append_line_to_update(
-    to_update, to_delete, variant, line_data, replace, variant_ids_in_lines
-):
-    if variant.pk not in variant_ids_in_lines:
+def _get_line_if_exist(line_data, lines_by_ids):
+    if line_data.line_id and line_data.line_id in lines_by_ids:
+        return lines_by_ids[line_data.line_id]
+
+
+def _append_line_to_update(to_update, to_delete, line_data, replace, line):
+    if line is None:
         return
-    line = variant_ids_in_lines[variant.pk]
+
     if line_data.quantity_to_update:
         quantity = line_data.quantity
         if quantity > 0:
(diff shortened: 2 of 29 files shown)


maarcingebala:
The draft of the new API looks good to me. One question about this point:

> At this moment we would like to block to make those actions via customer API key on lines when there is a custom price added or a single variant in multiple lines.

I think we don't currently track which line uses custom prices and we can only guess that by comparing line and variant prices (correct me if I'm wrong). I'm wondering if it wouldn't be convenient to add a model field `hasCustomPrice` or similar to track this fact and make permission checks easier.

fowczarek:
@maarcingebala on checkout line we have `price_override` field.

korycins:
@SzymJ How:

```
input CheckoutLineUpdateInput {
  forceNewLine: Boolean = false
  quantity: Int
  variantId: ID
  price: PositiveDecimal
  lineId: ID
}
```
will work?

`forceNewLine` means that we will create a new, separate CheckoutLine. In that case, shouldn't we just call mutation with `CheckoutLineInput`?  Setting `forceNewLine` to True in `CheckoutLineUpdateInput`, means that we create a new line, right? In that case (maybe I am missing something) but we will have two mutations that do the same (checkout-create-line and checkout-update-line)

SzymJ:
> @SzymJ How:
> 
> ```
> input CheckoutLineUpdateInput {
>   forceNewLine: Boolean = false
>   quantity: Int
>   variantId: ID
>   price: PositiveDecimal
>   lineId: ID
> }
> ```
> 
> will work?
> 
> `forceNewLine` means that we will create a new, separate CheckoutLine. In that case, shouldn't we just call mutation with `CheckoutLineInput`? Setting `forceNewLine` to True in `CheckoutLineUpdateInput`, means that we create a new line, right? In that case (maybe I am missing something) but we will have two mutations that do the same (checkout-create-line and checkout-update-line)

I was thinking that it would be nice to allow split already created line with the quantity  > 1, but I'm not sure if there is a use case for that. I not we can remove its flag from update input.

fowczarek:
We shouldn't forget about validating LineUpdate and LineDelete mutation required that one of the params `lineId` and `variantId`.

fowczarek:
IMO we should describe OrderLienes CRUD in `Mutations` section.

saleor/checkout/fetch.py
```diff
@@ -250,6 +252,17 @@ def fetch_checkout_lines(
             checkout, product, variant_channel_listing, product_channel_listing_mapping
         ):
             unavailable_variant_pks.append(variant.pk)
+            if not skip_lines_with_unavailable_variants:
```
korycins, line 255: What is the reason for having this?
Question: Did you check your logic with the scenario when checkout has some lines that are unavailable?

saleor/checkout/utils.py
```diff
@@ -238,12 +248,15 @@ def add_variants_to_checkout(
     return checkout
 
 
-def _append_line_to_update(
-    to_update, to_delete, variant, line_data, replace, variant_ids_in_lines
-):
-    if variant.pk not in variant_ids_in_lines:
+def _get_line_if_exist(line_data, lines_by_ids):
+    if line_data.line_id and line_data.line_id in lines_by_ids:
```
korycins, line 252: Is it possible that `line_id` will be None?
Also, can't we replace this function by `lines_by_ids.get(line_data.line_id)`? 
or `lines_by_ids.get(line_data.line_id) if line_data.line_id else None`

saleor/checkout/utils.py
```diff
@@ -258,25 +271,23 @@ def _append_line_to_update(
...
         if quantity <= 0:
             to_delete.append(line)
 
 
-def _append_line_to_create(
-    to_create, checkout, variant, line_data, variant_ids_in_lines
-):
-    if variant.pk not in variant_ids_in_lines:
+def _append_line_to_create(to_create, checkout, variants_map, line_data, line):
+    if line is None:
```
korycins, line 285: It looks like you added `if line is None` to all helpers functions called in `add_variants_to_checkout(`
Wouldn't be good to move `if line is None` to the `for`:
```
    for line_data in checkout_lines_data:
         line = _get_line_if_exist(line_data, lines_by_id)
         if line is None:
             continue
```

saleor/checkout/utils.py
```diff
@@ -258,25 +271,23 @@ def _append_line_to_update(
...
     if line_data.quantity_to_update:
         if quantity <= 0:
             to_delete.append(line)
 
 
-def _append_line_to_create(
-    to_create, checkout, variant, line_data, variant_ids_in_lines
-):
-    if variant.pk not in variant_ids_in_lines:
+def _append_line_to_create(to_create, checkout, variants_map, line_data, line):
```
korycins, line 284: You use `variants_map` only in `variant=variants_map[line_data.variant_id],`. In that case, wouldn't be better to pass here `variant` instead of `variants_map`?

saleor/core/utils/random_data.py
```diff
@@ -1561,11 +1561,16 @@ def get_image(image_dir, image_name):
     return File(open(img_path, "rb"), name=image_name)
 
 
-def create_checkout_with_preorders():
+def prep_checkout():
```
korycins, line 1564: Is `prep` mean `prepare`? If yes, I would recommend to use full word instead of shortcut (the function name is short, so I think there is no need to shortcut words). 
Also, `prep_checkout` doesn't prepare a `Checkout` but `CheckoutInfo`. Lets add correct typing or rename a function

saleor/graphql/checkout/mutations/utils.py
```diff
@@ -273,24 +275,100 @@ def get_checkout(
...
-    lines: List[Dict[str, Any]]
+def group_lines_input_on_add(
+    lines: List[Dict[str, Any]], existing_lines_info=None
 ) -> List[CheckoutLineData]:
-    variant_checkout_line_data_map: Dict[str, CheckoutLineData] = defaultdict(
-        CheckoutLineData
-    )
+    """Return list od CheckoutLineData objects.
+
+    This function is used in CheckoutLinesAdd and CheckoutCreate mutation.
```
korycins, line 283: Not sure if we should explicitly said in the docstring, where the function is used. If I use it in different place, then to keep a consistency I will have to also update the description.
If I change the name of mutation (or something), the description of the function will be incorrect.

saleor/graphql/checkout/mutations/utils.py
```diff
@@ -303,3 +381,36 @@ def check_permissions_for_custom_prices(app, lines):
         not app or not app.has_perm(CheckoutPermissions.HANDLE_CHECKOUTS)
     ):
         raise PermissionDenied(permissions=[CheckoutPermissions.HANDLE_CHECKOUTS])
+
+
+def find_line_id_when_variant_parameter_used(variant_id, lines_info):
+    """Return line id by using provided variantId parameter.
```
korycins, line 387: IMHO, would be good if you  could add a note here, that you're expecting `DB ID` for variant ID (It is `graphql/` directory so this could be tricky for someone).

korycins (commented):
@SzymJ  looks good, but make sure that you checked your code with:
- voucher for entire order
- voucher for specific products
- AvataxPlugin

saleor/checkout/utils.py
SzymJ: yes, it possible, on `linesAdd` line id can be `None`, but you are right I will change this function to one line

saleor/checkout/fetch.py
SzymJ: This flag was added because in some places where `fetch_checkout_lines` is used returning `CheckoutLineInfo` for unavailable variant breaks things for example in stripe/aydien plugins

SzymJ:
> @SzymJ looks good, but make sure that you checked your code with:
> 
> * voucher for entire order
> * voucher for specific products
> * AvataxPlugin

I think @eCzerniak was testing those cases

maarcingebala (approved):

saleor/checkout/utils.py
korycins: What about this? Is there any reason for passing the whole variants_amp here instead of variant taken from the map?

saleor/core/utils/random_data.py
korycins: What about this? If you don't agree, would be good to place a comment as I don't know if you miss this comment acidentialy 😄

saleor/checkout/utils.py
SzymJ: done, sorry i didnt pushed last commit

saleor/core/utils/random_data.py
maarcingebala: cc @SzymJ

korycins (approved):
Looks good 👍 . Let's make sure that QA team is aware, that vouchers & Avatax should be double checked.

saleor/core/utils/random_data.py
SzymJ: yesterday I removed some change by accident sorry for that I was sure is done :/ But its now!

PR #10108, opened 2022-07-06
Fix failed event delivery request headers

jakubkuc:
I want to merge this change because it mainly allows request_headers to be saved to `WebhookResponse` in case event delivery fails for sync events. Additionaly it simplifies  `send_webhook_using_scheme_method` , every scheme method will now return `WebhookResponse` and handle exceptions inside itself. It will also allow us to to set more detailed info to be returned in `WebhookResponse` for every method in case there is an exception.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/plugins/webhook/tasks.py
```diff
@@ -219,60 +222,89 @@ def send_webhook_using_http(
...
-        "QueueUrl": queue_url,
-        "MessageAttributes": msg_attributes,
-        "MessageBody": message.decode("utf-8"),
-    }
-    if is_fifo:
-        message_kwargs["MessageGroupId"] = domain
-    with catch_duration_time() as duration:
-        response = client.send_message(**message_kwargs)
-        return WebhookResponse(content=response, duration=duration())
+    try:
```
maarcingebala, line 225: Can't we put the `try..except` around the code which we exactly know can throw a specific exception? It would be a bit prettier and easier to debug when the exception occurs I think.

saleor/plugins/webhook/tasks.py
```diff
@@ -219,60 +222,89 @@ def send_webhook_using_http(
...
+                signature=signature,
+            )
+            response_duration = duration()
+            response = future.result()
+            return WebhookResponse(content=response, duration=response_duration)
+    except (pubsub_v1.publisher.exceptions.MessageTooLargeError, RuntimeError) as e:
+        return WebhookResponse(content=str(e), status=EventDeliveryStatus.FAILED)
+
+
+def generate_request_headers(event_type, domain, signature):
```
maarcingebala, line 297: Do we need this function? It's used in one place and doesn't do anything except returning a dict.

jakubkuc: Right, I made this function before simplifying when there were more uses... Right now it is used only in one place and in tests. Maybe I'll move it to tests utils to be used and remove it from code

fowczarek (approved):

maarcingebala (approved):

PR #10109, opened 2022-07-06
Add custom SearchVector class

8r2y5:
I want to merge this change because this adds custom SearchVector class, that when used skips Django's SQL Compiler validation that leads into RecursionError while processing large number of expressions (for example, 250 product variants).

The validation that is being skipped by this class: https://github.com/django/django/blob/08e6073f878264a0c091da0d3db456820252ef6c/django/db/models/sql/compiler.py#L1502

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/product/postgres.py
+++ b/saleor/product/postgres.py
@@ -1,0 +1,37 @@
+from django.contrib.postgres.search import (
+    CombinedSearchVector,
+    SearchVector,
+    SearchVectorCombinable,
+)
+
+
+class NoValidationSearchVectorCombinable(SearchVectorCombinable):
+    def _combine(self, other, connector, reversed):
+        if not isinstance(other, NoValidationSearchVectorCombinable):
+            raise TypeError(
+                "SearchVector can only be combined with other SearchVector "
+                "instances, got %s." % type(other).__name__
+            )
+        if reversed:
+            return NoValidationCombinedSearchVector(other, connector, self, self.config)
+        return NoValidationCombinedSearchVector(self, connector, other, self.config)
+
+
+class NoValidationCombinedSearchVector(
+    NoValidationSearchVectorCombinable, CombinedSearchVector
+):
+    contains_aggregate = False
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [SALEOR-7489 productVariantBulkCreate errors with maximum recursion depth exceeded](https://app.clickup.com/t/2549495/SALEOR-7489)

maarcingebala (approved):
Looks ok, just wondering if this is only specific to products or there are chances we will use it for other modules as well? And related question - how do we know which search vector to use, the Django's one or this one?

saleor/product/postgres.py
```diff
@@ -0,0 +1,32 @@
...
+    contains_aggregate = False
+    contains_over_clause = False
+
+
+class NoValidationSearchVector(SearchVector, NoValidationSearchVectorCombinable):
+    """Purpose of this class is to omit django's SQL compiler's validation.
+
+    This validation can lead into RecursionError while processing large number of
+    expressions.
+    """
```
fowczarek, line 32: Could you add info on what could happen without this validation?

fowczarek (commented):
I'm not sure that turning off validation is a good solution. @patrys @maarcingebala 
Maybe it's a bug in external code?

8r2y5:
> Looks ok, just wondering if this is only specific to products or there are chances we will use it for other modules as well?

This is not specific to product, it could be used with any SearchVector field. I created it product directory as it currently only used there.

> And related question - how do we know which search vector to use, the Django's one or this one?

It's good question which SeachVector class should be used. This class skips validation of expressions, normally if any expression would have aggregation or over clause, that validation would raise exception.
I would say that if you know what you are doing (not using aggregation and over clause expressions) and do not evaluate any code from external source that could contain aggregation - it is safe to use it.

patrys (approved):

fowczarek (approved):

Issue #10067, opened 2022-06-27
Add scheduled publishing webhook


timuric:
### What I'm trying to achieve
Trigger a webhook when a product/sale gets published given a scheduled date.

CossackDex:
@timuric It seems like a helpful tool. I will ask team to put it on our roadmap.

timuric:
@CossackDex thanks! Send my regards to the team!

CossackDex:
I will @timuric! I will

timuric:
Internal task: [7520](https://app.clickup.com/t/2549495/SALEOR-7520)

PR #10110, opened 2022-07-06
New event for starting and ending sales

IKarbowiak:
- Add new webhooks: `sale_toggle` that is sent every time the sale is started or ended. The webhook payload contains `is_active` flag. 
- Add a new celery task that calls `sale_toggle` webhook. The task is called periodically by celery schedule at least every minute or more frequently if any sale is starting or ending in less than a minute.
- Extend the `Sale` model with `notification_sent_datetime` that keeps date time of the last notification. As default the value is `None`, which means that the notification hasn't been sent yet for this sale.
  - When the sale is created with an already passed date the notification is sent and the notification date is set
  - When the sale is updated with an already passed date and the notification hasn't been sent before - the notification is sent and the notification date is set.
  - When the sale is updated with an already passed end date and the was no end date before, the notification is sent.

Docs: https://github.com/saleor/saleor-docs/pull/462

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-7520 Add publishing webhook](https://app.clickup.com/t/2549495/SALEOR-7520)

saleor/core/schedules.py
```diff
@@ -0,0 +1,78 @@
...
+
+import pytz
+from celery.schedules import BaseSchedule
+from celery.utils.time import maybe_timedelta, remaining
+from django.db.models import Q
+
+schedstate = namedtuple("schedstate", ("is_due", "next"))
+
+
+class sale_webhook_schedule(BaseSchedule):
```
IKarbowiak, line 12: I used here lowercase with an underscore as all celery schedules are written this way. Also, the [PEP](https://peps.python.org/pep-0008/#class-names) is saying:

> The naming convention for functions may be used instead in cases where the interface is documented and used primarily as a callable.

saleor/core/schedules.py
```diff
@@ -0,0 +1,77 @@
...
```
korycins, line 12: Is it expected that you didn't use python format for classes?

saleor/graphql/discount/mutations/sale_create.py
```diff
@@ -90,12 +93,49 @@ def clean_instance(cls, info, instance):
...
+        or ending date already passed.
+        """
+        now = datetime.now(pytz.utc)
+        notification_field = f"{field}ed_notification_sent"
+        date = getattr(instance, f"{field}_date")
+        if date and date <= now:
+            if field == "start":
+                manager.sale_started(instance, catalogue)
+            else:
+                manager.sale_ended(instance, catalogue)
```
korycins, line 139: [Question] It is SaleCreate mutation. It only creates a new sale. In that case do we need to trigger `sale_ended` hook? Like is it expected that we will send webhook that sale is ended even if there was never a way to use it?

saleor/graphql/discount/mutations/sale_update.py
```diff
@@ -30,16 +33,76 @@ class Meta:
...
+        cls.clean_instance(info, instance)
+        cls.save(info, instance, cleaned_input)
+        cls._save_m2m(info, instance, cleaned_input)
+        cls.send_sale_notifications(info, instance, cleaned_input, previous_catalogue)
+        return cls.success_response(instance)
+
+    @classmethod
+    def clean_input(cls, info, instance, data, input_cls=None):
+        clean_input = super().clean_input(info, instance, data)
+        cls.update_notification_flag_if_needed(instance, clean_input, "start")
```
korycins, line 50: [Question] Wouldn't be better to have one single method for this?  You always call it twice for `start` and `end`, so maybe would be worth to have single method here. Especially that you relay on `field` field. Where `field` is a part of the name of DB field

saleor/discount/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+    for sale in sales_ended:
+        catalogues = catalogue_infos.get(sale.id)
+        manager.sale_ended(sale, catalogues)
+
+    sales_ended_ids = ", ".join([str(sale.id) for sale in sales_ended])
+    task_logger.info(
+        "The sale_ended webhook sent for sales with ids: %s", sales_ended_ids
+    )
+
+    sales_started.update(started_notification_sent=True)
```
korycins, line 61: To save one DB query for each task call, maybe we could do something like:
`Sale.objects.bulk_update(sales_stared + sales_ended, ["started_notification_sent", "ended_notification_sent"])`
(This should work if we want to have a single task here).
What do you think?

fowczarek (approved):
Do not forget to regenerate types on the dashboard.

saleor/core/schedules.py
fowczarek: Should we add this info to the docstring?

saleor/core/schedules.py
fowczarek: https://github.com/saleor/saleor/pull/10110#discussion_r914685430

saleor/graphql/discount/mutations/sale_create.py
IKarbowiak: Hm, but It's possible to create `Sale` that has started and ended date in the past. And in this case, we will send the notification for the start and not for the end, and products might be on sale because of that.

saleor/core/schedules.py
IKarbowiak: Added.

saleor/graphql/discount/mutations/sale_create.py
korycins: Yeah, but what is the reason for sending such notifications for past events? It is already ended from the beginning of its life and I am not sure if this gives you anything except a noise that you (as an app) have to filter out. (But maybe I don't know all requirments or usecases).

IKarbowiak: I introduced a new solution that handles this case and does not send the notification in such a situation.

fowczarek (approved):

saleor/graphql/webhook/subscription_types.py
```diff
@@ -663,6 +664,19 @@ class Meta:
...
+class SaleToggle(ObjectType, SaleBase):
+    sale = graphene.Field(
+        "saleor.graphql.discount.types.Sale",
+        channel=graphene.String(
+            description="Slug of a channel for which the data should be returned."
+        ),
+        description="The sale the event relates to." + ADDED_IN_35 + PREVIEW_FEATURE,
+    )
+
+    class Meta:
```
maarcingebala, line 676: Hmm, it seems we are missing the general description of subscription events? We should have some description here that describes when the event is fired.

IKarbowiak: I can add, but we are missing the description for all events types :D

maarcingebala: Let's add here for now and we'll open task to add for others separately.

maarcingebala (approved):

saleor/discount/migrations/0043_sale_notification_sent_datetime.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("discount", "0042_migrate_orderdiscount_id_to_uuid"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="sale",
+            name="notification_sent_datetime",
+            field=models.DateTimeField(null=True, blank=True),
```
korycins, line 16: Setting this to null here will fire up end hooks for all existing and ended Sales, am I right?
Assuming that I have 100 old sales, then this will trigger 100 webhooks that sale has ended, right?
Is it expected? Shouldn't we ignore the old sales here?

korycins (approved):

saleor/discount/migrations/0043_sale_notification_sent_datetime.py
IKarbowiak: Good point. So maybe as default we should set the current time?

maarcingebala: We can change to the current time with some explaining comments. I find this a bit confusing since `null` seems more appropriate value if there was no notification. But even if we set null and then Celery would trigger the tasks right after the deployment, we will end up having a similar date set in the database, but in the second case we actually fire events that no one listens to.

fowczarek (approved):

korycins (approved):

jakubkuc (approved):

timuric:
https://github.com/saleor/saleor/issues/10067

PR #10111, opened 2022-07-06
Display only shippable countries for a channel

korycins:
The problem that we want to solve: Public link: https://github.com/saleor/saleor-checkout/issues/101, Internal task: https://app.clickup.com/t/2549495/SALEOR-7167

- Currently, we don't have a possibility to get a list of countries that are supported by channel. We can only fetch a list of ALL countries provided by the external package `django_countries`.  We have a possibility to limit the countries that are attached to the shipping zones but it doesn't give us any information about relationships with the channel. 
- The next issue is that we don't have the possibility to provide storefront information about all shipping methods that can be grouped by country. This could be useful when we want to provide a list of shipping methods for each country, like:
  - Poland:
    - X1 Shipping method
    - X2 Shipping method
  - France:
    - F1 Shipping method
    - F2 Shipping method
  - US:
    - U1 Shipping method
    - U2 Shipping method

  We have `shop.availableShippingMethods` but it allows us to fetch ALL available shipping methods or the ones for a specific address, but there is no way to fetch shipping methods grouped by country.
~~The proposition is to extend `CountryDisplay` by adding a new field `availableShippingMethods` that will provide available internal shipping methods available for the country.~~

The proposed scheme changes:
```
type Channel implements Node {

  countries: [CountryDisplay!]
  availableShippingMethodsPerCountry(countries: [CountryCode!]): [ShippingMethodsPerCountry!]
}
type ShippingMethodsPerCountry {
   countryCode: CountryCodeEnum
   shippingMethods: [ShippingMethod!]
}

type Query {
    channel(
    """ID of the channel."""
    id: ID

    """Slug of the channel."""
    slug: String
  ): Channel
}
```
The `channel` query will get a new input parameter `slug`. This will allow fetching basic channel info without having an ID o channel.  **Channel needs to be active**
Channel type will receive two new fields. `countries` - a list of countries that can be shipped to them from the channel, and available shipping methods. Both of the fields will be available without permission. 
channel.availableShippingMethodsPerCountry will receive `countries` field that will allow returning a list of shipping methods filtered by country codes.

~~The `CountryDisplay` has the field `VAT` that is fulfilled only when we're making the query `shop.countries`, but the `CountryDisplay` is also used in `Address` type, where `VAT` is always empty.  It looks like a place where we could separate `CountryDisplay` into two separate types. 
First one should contain only `code` and `country`. Second one could contain additional fields: `vat` and `availableShippingMethods`.
`shop.countries.vat` field is fulfilled by using data from `VAT.objects.all()`
where `VAT` model comes from:
`from django_prices_vatlayer.models import VAT`~~

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-7167 Display only shippable countries for a channel](https://app.clickup.com/t/2549495/SALEOR-7167)

fowczarek:
LGTM :tada: IMO we should split `CountryDisplay` as you suggest.

maarcingebala:
Few thoughts:
- Do we still need the query `shop { countries }`? I think dashboard and any other client can use the `CountryCode` enum to get the list of all available countries. Question is - do we need in API to return country names? I think frontend should only rely on enum values and not use names from the backend.
- If we deprecated `shop { countries }`, I wouldn't extend it as Maciek suggested, but instead I would make the `channel(slug: "")` query public and have: `channel { countries }` instead. Obviously, we would make any other channel fields protected with permissions.
- Loose idea - shouldn't we group shipping methods by countries in the frontend, instead of providing this grouping in API? We could only make sure that frontend has enough info to build these groups.

What do you think @patrys @timuric?

patrys:
I like Marcin's proposal of making `Query.channel` partially public.

We should not name the tax field `vat` as VAT is only one type of tax.

maarcingebala:
@patrys Most likely we're going to deprecate `CountryDisplay.vat` field with simple taxes API.

korycins:
@maarcingebala `Channel` query uses `id`, not a `slug`. Storefront knows only a `slug` of channels. In the proposed way, we have to provide two optional `input` fields: `id`  (changed from required to optional) and `slug`.

maarcingebala:
@korycins Sounds good

saleor/graphql/channel/resolvers.py
```diff
@@ -1,8 +1,26 @@
...
+from .types import Channel
+
+
+def resolve_channel(info, id: Optional[str], slug: Optional[str]):
+    validate_one_of_args_is_in_query("id", id, "slug", slug)
+    if id:
+        _, db_id = from_global_id_or_error(id, Channel)
+        channel = models.Channel.objects.filter(id=db_id).first()
+    else:
+        channel = models.Channel.objects.filter(slug=slug).first()
```
jakubkuc, line 16: what if channel with given slug does not exists? shouldn't we raise `GraphQLError`? in this case it would return None, as well as if it gets to line: 20 `if is_staff_user(info.context) or is_app(info.context):`

patrys: Do we raise errors when trying to fetch a product that is not public or one that does not exist?

jakubkuc: We do not, my bad. Got confused a bit there.

saleor/graphql/channel/resolvers.py
```diff
@@ -1,8 +1,26 @@
...
 
+    if channel and channel.is_active:
+        return channel
+    if is_staff_user(info.context) or is_app(info.context):
+        return channel
```
fowczarek, line 21: Should we return inactive channels for staff and apps with channel permissions?

korycins: Previously we didn't validate it. This means any app/staff could get not active channels.

saleor/graphql/channel/schema.py
```diff
@@ -16,14 +16,17 @@
...
-        id=graphene.Argument(graphene.ID, description="ID of the channel."),
+        id=graphene.Argument(
+            graphene.ID, description="ID of the channel.", required=False
+        ),
+        slug=graphene.Argument(
+            graphene.String,
+            description="Slug of the channel." + ADDED_IN_36 + PREVIEW_FEATURE,
+            required=False,
+        ),
         description="Look up a channel by ID.",
```
maarcingebala, line 29: Need to update the description here, is no longer only by ID.

saleor/graphql/channel/types.py
```diff
@@ -119,11 +123,32 @@ class Meta:
 
 class Channel(ModelObjectType):
     id = graphene.GlobalID(required=True)
-    name = graphene.String(required=True)
-    is_active = graphene.Boolean(required=True)
-    slug = graphene.String(required=True)
-    currency_code = graphene.String(required=True)
     slug = graphene.String(required=True)
+
+    name = PermissionsField(
```
maarcingebala, line 128: We are missing descriptions for these fields, it's a good occasion to add them ;)

saleor/graphql/shipping/types.py
```diff
@@ -365,3 +366,8 @@ def resolve_minimum_order_weight(root: ShippingMethodData, _info):
 class ShippingZoneCountableConnection(CountableConnection):
     class Meta:
         node = ShippingZone
+
+
+class ShippingMethodsPerCountry(graphene.ObjectType):
+    country_code = graphene.Field(CountryCodeEnum, required=True)
+    shipping_methods = NonNullList(ShippingMethod)
```
maarcingebala, line 373: 🚓 Missing descriptions :p

maarcingebala (approved):

IKarbowiak (approved):

jakubkuc (approved):

PR #10142, opened 2022-07-12
Fix stock by ID dataloader

jakubkuc:
I want to merge this change because it fixes problem with dataloader that causes it to return stocks in wrong order. That in some situations (like in linked ticket) caused API to return data inconsistent with data from DB.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/warehouse/dataloaders.py
+++ b/saleor/graphql/warehouse/dataloaders.py
@@ -347,7 +347,8 @@ class StockByIdLoader(DataLoader):
     context_key = "stock_by_id"
 
     def batch_load(self, keys):
-        return Stock.objects.using(self.database_connection_name).in_bulk(keys).values()
+        stocks = Stock.objects.using(self.database_connection_name).in_bulk(keys)
+        return [stocks.get(key) for key in keys]
 
 
 class WarehousesByChannelIdLoader(DataLoader):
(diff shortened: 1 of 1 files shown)


patrys:
Task linked: [SALEOR-7078 OrderFulfulfill change previously created fulfillment.](https://app.clickup.com/t/2549495/SALEOR-7078)

saleor/graphql/warehouse/dataloaders.py
```diff
@@ -347,7 +347,8 @@ class StockByIdLoader(DataLoader):
     context_key = "stock_by_id"
 
     def batch_load(self, keys):
-        return Stock.objects.using(self.database_connection_name).in_bulk(keys).values()
+        stocks = Stock.objects.using(self.database_connection_name).in_bulk(keys)
+        return [stocks.get(key) for key in keys]
```
korycins, line 351: Shouldn't be `.get(key, [])`?

korycins (approved):

saleor/graphql/warehouse/dataloaders.py
fowczarek: IMO I think we shouldn't use default here. What code should do if receives `[]` instead of `Stock` object. ?

fowczarek: We could use `[]` as the default in dataloaders like `AvailableShippingMethodByCheckoutId` when we return list of objects.

IKarbowiak: I agree with @fowczarek

IKarbowiak (approved):

fowczarek (approved):

maarcingebala (approved):

PR #10151, opened 2022-07-14
Handle same variant in multiple lines in CheckoutLineAdd/Update

SzymJ:
I want to merge this change because it adds the possibility to force a new line in checkout and use `lineId` to update lines.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/fetch.py
+++ b/saleor/checkout/fetch.py
@@ -278,6 +278,7 @@ def fetch_checkout_lines(
             apply_voucher_to_checkout_line(
                 voucher_info, checkout, lines_info, discounts
             )
+
     return lines_info, unavailable_variant_pks
 
 
--- a/saleor/checkout/fetch.py
+++ b/saleor/checkout/fetch.py
@@ -211,7 +211,9 @@ def get_delivery_method_info(
 
 
 def fetch_checkout_lines(
-    checkout: "Checkout", prefetch_variant_attributes=False
+    checkout: "Checkout",
+    prefetch_variant_attributes=False,
+    skip_lines_with_unavailable_variants=True,
 ) -> Tuple[Iterable[CheckoutLineInfo], Iterable[int]]:
     """Fetch checkout lines as CheckoutLineInfo objects."""
     from .utils import get_voucher_for_checkout
@@ -250,6 +252,17 @@ def fetch_checkout_lines(
             checkout, product, variant_channel_listing, product_channel_listing_mapping
         ):
             unavailable_variant_pks.append(variant.pk)
+            if not skip_lines_with_unavailable_variants:
+                lines_info.append(
+                    CheckoutLineInfo(
+                        line=line,
+                        variant=variant,
+                        channel_listing=variant_channel_listing,
+                        product=product,
+                        product_type=product_type,
+                        collections=collections,
+                    )
+                )
             continue
 
         lines_info.append(
@@ -278,7 +291,6 @@ def fetch_checkout_lines(
             apply_voucher_to_checkout_line(
                 voucher_info, checkout, lines_info, discounts
             )
-
     return lines_info, unavailable_variant_pks
 
 
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -251,7 +251,6 @@ def add_variants_to_checkout(
 def _get_line_if_exist(line_data, lines_by_ids):
     if line_data.line_id and line_data.line_id in lines_by_ids:
         return lines_by_ids[line_data.line_id]
-    return
 
 
 def _append_line_to_update(to_update, to_delete, line_data, replace, line):
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -127,6 +127,7 @@ def add_variant_to_checkout(
     price_override: Optional["Decimal"] = None,
     replace: bool = False,
     check_quantity: bool = True,
+    force_new_line: bool = False,
 ):
     """Add a product variant to checkout.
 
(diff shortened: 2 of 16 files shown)


patrys:
Task linked: [SALEOR-7111 Allow multiple copies of a single variant in the same Checkout](https://app.clickup.com/t/2549495/SALEOR-7111)

saleor/checkout/utils.py
```diff
@@ -132,15 +133,23 @@ def add_variant_to_checkout(
         check_quantity=check_quantity,
     )
 
+    if force_new_line:
+        checkout.lines.create(
+            checkout=checkout,
```
IKarbowiak, line 138: I suppose you do not have to specify the `checkout` parameter in such case when you are creating the lines with the use of `checkout.lines.create`

saleor/graphql/checkout/mutations/utils.py
```diff
@@ -274,24 +276,49 @@ def get_checkout(
...
+        if variant_id:
+            _, variant_db_id = graphene.Node.from_global_id(variant_id)
+            variant_db_id = int(variant_db_id)
+
+        if force_new_line:
+            line_data = CheckoutLineData(variant_id=variant_db_id)
+            force_new_line_checkout_lines_data_map.append(line_data)
+        else:
+            if line_id:
+                line_data = variant_checkout_line_data_map[line_id]
```
IKarbowiak, line 304: I don't get why we need the `variant_checkout_line_data_map`? We are never assigning anything to it, and only getting the value from default dict which always will be `CheckoutLineData`, cannot we just define the `line_data` at the beginning of the for loop?

fowczarek: Also, the name of the variable suggests that the key in this map is `variant` but we use `line_id` and `variant_id` as keys.

fowczarek (commented):
Please do not forget to update populate_db. We should have at least one checkout and one order with the same variant in multiple lines. (Could be done in separate PR)

saleor/graphql/checkout/mutations/checkout_lines_update.py
```diff
@@ -113,6 +127,26 @@ def clean_input(
                     }
                 )
 
+        # # if same variant occur in multiple lines `lineId` parameter have to be used
+        variants_ids_in_existing_lines = [line.variant.pk for line in lines]
```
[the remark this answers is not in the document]
SzymJ: `lines` here are a list of `CheckoutLineInfo` objects so there is no `line.variant_id` but all variants are prefetched so there are no additional querys.

saleor/graphql/checkout/mutations/checkout_lines_update.py
```diff
@@ -113,6 +127,26 @@ def clean_input(
...
+        for line in checkout_lines_data:
+            if (
+                line.variant_id
+                and variants_ids_in_existing_lines.count(line.variant_id) > 1
+            ):
+                message = (
+                    f"Variant #{line.variant_id} occur in multiple lines. "
```
korycins, line 138: Just to confirm. Is `variant_id` graphql ID?

saleor/checkout/utils.py
```diff
@@ -175,18 +184,29 @@ def add_variants_to_checkout(
     country_code = checkout.get_country()
 
     checkout_lines = checkout.lines.select_related("variant")
-    variant_ids_in_lines = {line.variant_id: line for line in checkout_lines}
+
+    lines_by_variant_id = {line.variant_id: line for line in checkout_lines}
```
korycins, line 188: I see naming incosistency here. You have `lines_by_variant_id` here and `lines_by_ids` - `id` vs `ids`.

saleor/checkout/utils.py
```diff
@@ -175,18 +184,29 @@ def add_variants_to_checkout(
...
+
     to_create: List[CheckoutLine] = []
     to_update: List[CheckoutLine] = []
     to_delete: List[CheckoutLine] = []
-    for variant, line_data in zip(variants, checkout_lines_data):
+
+    for line_data in checkout_lines_data:
         _append_line_to_update(
-            to_update, to_delete, variant, line_data, replace, variant_ids_in_lines
+            to_update, to_delete, line_data, replace, lines_by_variant_id, lines_by_ids
```
korycins, line 198: Also, do we really need to pass two separate dicts that store the line details? 
Maybe I am missing something but can't we unify the functions that operate on lines to use only line IDs instead of trying to use lineID and then trying to use variantID?

saleor/graphql/checkout/mutations/checkout_lines_update.py
SzymJ: I had to move this error to another place, but now I used `variants` parameter as @fowczarek suggested and global IDs are returned.

saleor/checkout/fetch.py
```diff
@@ -276,6 +276,7 @@ def fetch_checkout_lines(
             apply_voucher_to_checkout_line(
                 voucher_info, checkout, lines_info, discounts
             )
+
```
fowczarek, line 279: We could leave this file unchanged.

saleor/checkout/utils.py
SzymJ: Right, I did that to handle cases where line has a variant that is unavailable - `fetch_checkout_lines` was not returning `CheckoutLineInfo` for those lines so I was not able to find line for provided variant. 
But now I added flag `skip_lines_with_unavailable_variants`  to `fetch_checkout_lines` function and removed `lines_by_variant_id`  to operate only on `line_id` as you suggested.

IKarbowiak (approved):

jakubkuc (approved):

eCzerniak (approved):

PR #10181, opened 2022-07-18
App dataloader

dekoza:
I want to merge this change because we need to get rid of Lazy objects.
Internal ticket: SALEOR-7590

This PR covers `context.app` instances. Other lazy objects will be covered in upcoming PRs.

# Impact

* Internal changes removing some lazy objects from request and context

# Pull Request Checklist

* [x] The changes are tested


patrys:
Task linked: [SALEOR-7590 Get rid of lazy objects from codebase](https://app.clickup.com/t/2549495/SALEOR-7590)

saleor/graphql/app/dataloaders.py
```diff
@@ -35,3 +38,49 @@ def batch_load(self, keys):
...
+        for auth_token, token_last_4, app_id in tokens:
+            for raw_token in last_4s_map[token_last_4]:
+                if check_password(raw_token, auth_token):
+                    authed_apps[raw_token] = app_id
+
+        apps = App.objects.using(self.database_connection_name).filter(
+            id__in=authed_apps.values(), is_active=True
+        )
+
+        return [apps.filter(id=authed_apps.get(key, None)).first() for key in keys]
```
IKarbowiak, line 80: `None` can be removed as `get` returned the `None` by default when the key doesn't exist.

saleor/graphql/app/dataloaders.py
```diff
@@ -35,3 +38,49 @@ def batch_load(self, keys):
...
+        return None
+    tokens = AppToken.objects.filter(token_last_4=raw_auth_token[-4:]).values_list(
+        "app_id", "auth_token"
+    )
+    app_ids = [
+        app_id
+        for app_id, auth_token in tokens
+        if check_password(raw_auth_token, auth_token)
+    ]
+    return App.objects.filter(id__in=app_ids, is_active=True).first()
```
IKarbowiak, line 54: This method shouldn't be in the data loader file, as it does not use any dataloader.

saleor/app/models.py
```diff
@@ -117,6 +117,10 @@ def create(self, app, name="", auth_token=None, **extra_fields):
         app_token.save()
         return app_token, auth_token
 
+    def create_with_token(self, *args, **kwargs):
+        """See https://app.clickup.com/t/2549495/SALEOR-7847 ."""
```
IKarbowiak, line 121: IMO, we shouldn't keep internal ticket URLs in code.

patrys, line 121: This is more of a comment than a docstring. Also, the URL is not public, so most people won't be able to understand what the docstring is trying to convey.

IKarbowiak: You can drop the ticket from the code and add it as a review comment. Also, AFAIU adding the `create_with_token` method without updating the `create` is useless.

saleor/graphql/app/dataloaders.py
```diff
@@ -35,3 +38,49 @@ def batch_load(self, keys):
...
+
+
+class AppByTokenLoader(DataLoader):
+    context_key = "app_by_token"
+
+    def batch_load(self, keys):
+        last_4s_map = defaultdict(list)
```
IKarbowiak, line 61: I will suggest creating more meaningful variable name, that will also explain to what this map is, like `last_4s_to_raw_token_map`

saleor/graphql/invoice/mutations.py
```diff
@@ -173,17 +174,19 @@ def perform_mutation(cls, _root, info, **data):
         invoice.order = order
         invoice.status = JobStatus.SUCCESS
         invoice.save()
+        app = load_app(info.context)
+        # app = AppByTokenLoader(info.context).load(info.context.auth_token)
```
IKarbowiak, line 178: To remove I suppose. 👮‍♀️

saleor/graphql/app/dataloaders.py
dekoza: Overcautious :sweat_smile:

saleor/graphql/app/dataloaders.py
dekoza: I needed to refactor it out of its original position to mitigate circular dependency. Can you propose a proper location for it?

saleor/app/models.py
dekoza: > adding the `create_with_token` method without updating the create is useless.

Not exactly. It lets me use the proper method while waiting for `create` to get fixed without the need to change it afterwards. I say it's convenience.

saleor/graphql/invoice/mutations.py
dekoza: Should be marked as `# TODO` or `# FIXME` as I'm working on this part for future PR.

saleor/graphql/app/dataloaders.py
korycins: IMO it is something more related to `saleor/app` than `saleor/graphql` module.

saleor/graphql/app/dataloaders.py
dekoza: Also, `explicit is better than implicit` :laughing:

saleor/graphql/invoice/mutations.py
dekoza: Nah, it uses `load_app` so the comment should be removed. Thanks!

saleor/graphql/app/dataloaders.py
dekoza: Done.

saleor/graphql/app/dataloaders.py
korycins: Maybe I am missing something, but it looks like we will make query for each key in keys.
```
res = []
for key in keys:
    res.append(apps.filter(id=authed_apps.get(key, None)).first())
return res
```

dekoza: :thinking: Good point

dekoza: Yeas, I should have used `in_bulk` and `.get()` the objects from the resulting dict. Much cleaner. Thank you!

saleor/graphql/core/dataloaders.py
```diff
@@ -33,7 +33,7 @@ def __new__(cls, context: HttpRequest):
     def __init__(self, context):
         if self.context != context:
             self.context = context
-            self.user = context.user
+            self.user = context.user if hasattr(context, "user") else None
```
maarcingebala, line 36: Cosmetics: `getattr(context, "user", None)`

saleor/graphql/tests/fixtures.py
```diff
@@ -161,6 +166,21 @@ def info(schema_context):
     return Mock(context=schema_context)
 
 
+@pytest.fixture
+def anon_user():
```
maarcingebala, line 170: I would make the names more explicit here: `anonymous_user` etc.

saleor/order/utils.py
```diff
@@ -944,8 +944,8 @@ def _update_order_total_charged(order: Order):
         sum(order.payments.values_list("captured_amount", flat=True))  # type: ignore
         or 0
     )
-    order.total_charged_amount += sum(
-        order.payment_transactions.values_list("charged_value", flat=True)  # type: ignore # noqa: E501
+    order.total_charged_amount += sum(  # type: ignore
```
maarcingebala, line 947: If I recall correctly this was failing when we upgraded mypy to the version that we currently have in repo, that's why it was moved to the line below.

saleor/graphql/checkout/mutations/checkout_lines_update.py
```diff
@@ -130,7 +130,9 @@ def clean_input(
         replace,
     ):
         # if the requestor is not app, the quantity is required for all lines
-        if not info.context.app:
+        if (
+            not info.context.app
+        ):  # TODO: how to use get_app without increasing query count
```
maarcingebala, line 135: Can't you use `load_app` here as well?

saleor/graphql/app/dataloaders.py
```diff
@@ -35,3 +38,49 @@ def batch_load(self, keys):
...
+            id__in=authed_apps.values(), is_active=True
+        )
+
+        return [apps.filter(id=authed_apps.get(key)).first() for key in keys]
+
+
+def load_app(context):
+    if not getattr(context, "auth_token", None):
+        return None
+    return AppByTokenLoader(context).load(context.auth_token).get()
```
maarcingebala, line 86: How does it works exactly? From the usages I see we don't have to "wait" or resolve a promise with "then", so does the `get` call resolve the app immediately? If so, is `load_app` ensuring the constant number of DB queries, regardless of how many times it was called during a single request?

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable, unreleased changes to this project will be documented in this file.
 
 # 3.7.0 [Unreleased]
 
+### Other changes
+
+- Started process of removing lazy objects from codebase. This is the first step on the road to async.
```
maarcingebala, line 9: I would be more explicit here - "Replaced `context.app` with dataloader" or something like that.

saleor/graphql/core/dataloaders.py
dekoza: Obviously :sweat_smile:

saleor/order/utils.py
dekoza: I'll test again but I had this failing in tests.

saleor/graphql/checkout/mutations/checkout_lines_update.py
dekoza: Yes but it triggers the test checking query count. The other tests (on github) are triggered anyway so I probably should use that here too and try to lower the count in next iteration :thinking:

CHANGELOG.md
patrys: Not sure this is even worth mentioning in the changelog, as `context.app` is likely not even useful for plugin authors.

saleor/graphql/app/dataloaders.py
dekoza: It works very alike to a LazyObject but is more eager - it fetches the App from DB asap and caches it for future. Yes, right now this increases the query count in some places. I'm in the process of teaching Django to understand Promises and after that the `.get()` could go someplace else to work more lazily.

CHANGELOG.md
dekoza: Transparency, I guess :thinking:

patrys: Transparency at scale is noise. 😄 We really need to make sure we only write about things that are relevant to our users as their attention is a finite resource.

saleor/graphql/checkout/mutations/checkout_lines_update.py
dekoza: Fixed.

saleor/order/utils.py
dekoza: Fixed.

saleor/graphql/app/dataloaders.py
dekoza: I can refactor it back to original place (`saleor/graphql/context.py`) but it is getting removed anyway in next iteration because it's only used for lazy object.

saleor/graphql/app/dataloaders.py
```diff
@@ -35,3 +38,52 @@ def batch_load(self, keys):
...
+                    authed_apps[raw_token] = str(app_id)
+
+        apps = (
+            App.objects.using(self.database_connection_name)
+            .filter(id__in=authed_apps.values(), is_active=True)
+            .in_bulk()
+        )
+        apps = {str(k): v for k, v in apps.items()}
+
+        return [apps.get(authed_apps.get(key)) for key in keys]
```
fowczarek, line 83:
```suggestion
                if check_password(raw_token, auth_token):
                    authed_apps[raw_token] = app_id

        apps = (
            App.objects.using(self.database_connection_name)
            .filter(id__in=authed_apps.values(), is_active=True)
            .in_bulk()
        )

        return [apps.get(authed_apps.get(key)) for key in keys]
```

saleor/graphql/tests/fixtures.py
```diff
@@ -161,6 +166,21 @@ def info(schema_context):
...
 
 
+@pytest.fixture
+def anonymous_user():
+    return AnonymousUser()
+
+
+@pytest.fixture
+def anonymous_app():
+    return None
```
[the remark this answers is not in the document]
dekoza: IMHO `anonymous_app` fixture says more than just passing `None` in the test code. This (along with two other fixtures) replaces the convoluted `info` fixture which had `"app": None` anyway.

saleor/graphql/tests/fixtures.py
```diff
@@ -161,6 +166,21 @@ def info(schema_context):
...
+
+
+@pytest.fixture
+def anonymous_plugins():
+    return get_plugins_manager()
```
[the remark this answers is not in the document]
dekoza: Yes, fixture cleanup would be most welcome. I'll check the ones you proposed instead of this one.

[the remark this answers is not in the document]
dekoza: `get_plugins_manager` returns PluginManager filled with plugins from settings. The managers you mentioned are bit different :wink: But I'm all for putting fixtures in one place.

saleor/graphql/tests/fixtures.py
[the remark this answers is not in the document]
fowczarek: Maybe we could pass `app=None` that should be clear enough?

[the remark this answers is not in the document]
maarcingebala: I agree with Owczar, don't see much value in having a fixture that always returns None. What Owczar suggests seems readable enough to me.

saleor/graphql/app/dataloaders.py
IKarbowiak: Maybe `saleor/app/utils.py`?

saleor/graphql/tests/fixtures.py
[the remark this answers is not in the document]
dekoza: Done.

saleor/graphql/app/dataloaders.py
dekoza: Done.

fowczarek (approved):

saleor/graphql/account/resolvers.py
```diff
@@ -214,7 +215,7 @@ def resolve_address(info, id):
 
 def resolve_addresses(info, ids):
     user = info.context.user
-    app = info.context.app
+    app = load_app(info.context)  # TODO: investigate
```
IKarbowiak, line 218: `TODO` to remove?

IKarbowiak (approved):

saleor/app/utils.py
```diff
@@ -0,0 +1,19 @@
+from typing import Union
+
+from django.contrib.auth.hashers import check_password
+
+from saleor.app.models import App, AppToken
+
+
+def get_app(raw_auth_token) -> Union[None, App]:
```
maarcingebala, line 8: Is this function used anymore? I'm struggling to find its usages or I'm blind ;) There is a bunch of warning that it is untested.

dekoza: Right now it might not be used anymore, I'll double-check it, thanks.

dekoza: Removed.

maarcingebala (approved):

saleor/app/utils.py
```diff
@@ -0,0 +1,19 @@
+from typing import Union
+
+from django.contrib.auth.hashers import check_password
+
+from saleor.app.models import App, AppToken
```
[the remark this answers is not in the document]
dekoza: Done.

PR #10183, opened 2022-07-18
Fix problem with updating draft order with active avalara

IKarbowiak:
Fix problem with updating draft order when order does not require shipping and avalara is active.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/plugins/avatax/__init__.py
```diff
@@ -159,11 +159,12 @@ def _validate_order(order: "Order") -> bool:
     )
     if not valid_address_details:
         return False
-    channel_listing = shipping_method.channel_listings.filter(  # type: ignore
-        channel_id=order.channel_id
-    ).first()
-    if not channel_listing:
-        return False
+    if shipping_required:
```
maarcingebala, line 162: Is it possible at this point that shipping is required, but shipping method is not set and can be null anyway?

IKarbowiak: No, in such a situation, the `_validate_adddress_details` will return `False` and this part of the code will be not reached, but there can be a situation, when shipping is not required, the `valid_address_details` returned `True` and we do not have shipping method set. So we need to ensure that the following part of the code is only reached for orders with shipping required.

maarcingebala: 👍

maarcingebala (approved):

PR #10185, opened 2022-07-18
Fix inconsistent beat scheduling and compatibility with db scheduler

NyanKiyoshi:
This fixes the following bugs:
- `tick()` could decide to never schedule anything else than `send-sale-toggle-notifications` if `send-sale-toggle-notifications` doesn't return `is_due = False` (stuck forever until beat restart or a `is_due = True`)
- `tick()` was sometimes scheduling other schedulers such as observability to be ran every 5m instead of every 20s
- `is_due()` from `send-sale-toggle-notifications` was being invoked every 5s on django-celery-beat instead of every 60s
- `send-sale-toggle-notifications` would crash on django-celery-beat with `Cannot convert schedule type <saleor.core.schedules.sale_webhook_schedule object at 0x7fabfdaacb20> to model`

This fixes both the `shelve` and the database backends. Users should be able to add compatibility to other Celery beat backends using the same flow.

Tickets are still pending creation to notify Celery about the issues.

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/schedulers/schedulers.py
```diff
@@ -0,0 +1,170 @@
...
+                defaults=cls._unpack_fields(**entry),
+            ),
+            app=app,
+        )
+
+
+class BaseScheduler(celery.beat.Scheduler):
+    """Define the base scheduler for Celery beat."""
+
+    def tick(
```
NyanKiyoshi, line 65: Original code is here: https://github.com/celery/celery/blob/aa9fd8a6c06e69c7eda2a59866c3d84622c85d20/celery/beat.py#L321-L359

Note that `is_numeric_value` is new at [v5.3.0a1](https://github.com/celery/celery/releases/tag/v5.3.0a1) thus can't be imported yet

saleor/core/schedules.py
```diff
@@ -2,14 +2,15 @@
 from datetime import datetime
 
 import pytz
-from celery.schedules import BaseSchedule
 from celery.utils.time import maybe_timedelta, remaining
 from django.db.models import F, Q
 
+from saleor.schedulers.customschedule import CustomSchedule
```
IKarbowiak, line 8: Should be used relative import.

IKarbowiak (approved):

8r2y5 (approved):

maarcingebala (approved):

PR #10195, opened 2022-07-19
RFC: Warehouse assignment improvement

IKarbowiak:
# Problem
Right now there is no way to decide for which warehouse the allocation is made.

## Requirements
We need a way to allow deciding which warehouse should be used for allocation in the checkout process.

## Proposed solution
- Allow explicitly picking a warehouse that should be used to allocate stock for a specific checkout line.
- Setting explicit warehouse will be allowed to apps with `HANDLE_CHECKOUTS` permission.

### API changes
- Add a new `warehouse: String / ID`  parameter to the `CheckoutLinesAdd` and `CheckoutLinesUpdate` - to discuss if the `warehouse` should require `ID`, `slug` or either of them. Restricted to apps with `HANDLE_CHECKOUTS` permission.
- Extend the `warehouse` query with the `slug` parameter that will allow query warehouse either by `slug` or `ID`.
- Adding lines with the same variant but different warehouses results in separate checkout lines.
- When the line for variant A and warehouse B is updated with warehouse A, and there is already a line for variant A and warehouse A - we will end up with two the same lines ~the lines will be merged - we will get one line instead of two with summed quantities~

```graphql
type Query {
  warehouse(
    id: ID
    slug: String
  ): Warehouse
}

input CheckoutLineInput {
  warehouse: ID/String
}

input CheckoutLineUpdateInput {
  warehouse: ID/String
}
```

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


timuric:
I have a concern about checkout level warehouse allocations. 

Scenario:

1 Add to bag item with warehouse A
2 During checkout get "No items" in stock
3 Go to PDP and see "Item is in stock"

In my opinion products, and queries should be consistent with checkout when it comes to stock, customers should not be removing and adding the same item to be able to purchase them.

I would suggest adding warehouse assignment API during the order creation.

PR #10240, opened 2022-07-26
Update method for setting filter descriptions.

dekoza:
I want to merge this change because there was no way to set description for complex filters - the ones that require descriptions the most.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


saleor/graphql/product/filters.py
```diff
@@ -585,6 +585,7 @@ class ProductFilter(MetadataFilterBase):
         input_class=PriceRangeInput,
         method="filter_minimal_price",
         field_name="minimal_price_amount",
+        help_text="Product's minimal price (after discounts) is in given range.",
```
fowczarek, line 588: Maybe
"Filter by lowest variant price after discounts"

saleor/graphql/core/types/converter.py
```diff
@@ -17,7 +17,10 @@
 
 
 def get_form_field_description(field):
-    return str(field.help_text) if field.help_text else None
+    if hasattr(field, "help_text"):
```
fowczarek, line 20: IMO we could use `description` as the name of this field.

dekoza: `help_text` is standard attribute for Django models and thus it's somewhat familiar. I also saw it used frequently in saleor's filter definitions. I think changing that notation is out of scope here.

IKarbowiak (approved):

saleor/graphql/product/filters.py
dekoza: Done

maarcingebala (approved):

PR #10244, opened 2022-07-27
Add product creation prerequisites

jakubkuc:
I want to merge this change because it adds migrations that creates default channel, warehouse, category and product_type objects so that user is able to create his first product right after setting up Saleor.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/channel/migrations/0004_create_default_channel.py
```diff
@@ -0,0 +1,20 @@
+from django.db import migrations
+
+
+def create_default_channel(apps, schema_editor):
+    Channel = apps.get_model("channel", "Channel")
+
+    Channel.objects.create(
```
maarcingebala, line 7: We already have a migration "channel/0001" which creates a default channel in some cases; it also happens to use "default-channel" as a slug. If such an instance is already in the database, this migration will fail as slug needs to be unique. We should use `get_or_create` in this migration to avoid it. Also since we have `DEFAULT_CHANNEL_SLUG` env, we could use it here as well.

saleor/product/migrations/0172_create_default_category_and_product_type.py
```diff
@@ -0,0 +1,39 @@
+from django.db import migrations
+
+from ...product import ProductTypeKind
+
+
+def create_default_category(apps, schema_editor):
+    Category = apps.get_model("product", "Category")
+    Category.objects.create(
```
maarcingebala, line 8: We should use `get_or_create` in all places.

saleor/product/migrations/0172_create_default_category_and_product_type.py
```diff
@@ -0,0 +1,39 @@
...
+        rght=0,
+        tree_id=0,
+        level=0,
+    )
+
+
+def create_default_product_type(apps, schema_editor):
+    ProductType = apps.get_model("product", "ProductType")
+    ProductType.objects.create(
+        name="Default Type",
```
maarcingebala, line 21:
```suggestion
        name="Default Product Type",
```

maarcingebala (commented):
@jakubkuc Make sure to test applying these migrations in two cases:
- on a already populated database
- empty DB

saleor/product/migrations/0172_create_default_category_and_product_type.py
```diff
@@ -0,0 +1,39 @@
...
+        slug="default-type",
+        kind=ProductTypeKind.NORMAL,
+        has_variants=False,
```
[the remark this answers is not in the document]
patrys: It does not matter, and there are no default attributes anyway. If someone wants variants, they can edit the type.

saleor/product/migrations/0172_create_default_category_and_product_type.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def create_default_product_type(apps, schema_editor):
+    ProductType = apps.get_model("product", "ProductType")
+    if not ProductType.objects.all().exists() and not settings.BLOCK_POPULATE_DEFAULTS:
+        ProductType.objects.create(
+            name="Default Type",
+            slug="default-type",
+            kind=ProductTypeKind.NORMAL,
+            has_variants=False,
```
korycins, line 27: Is it expected that we create Simple ProductType?

saleor/settings.py
```diff
@@ -598,6 +598,9 @@ def get_host():
 # Slug for channel precreated in Django migrations
 DEFAULT_CHANNEL_SLUG = os.environ.get("DEFAULT_CHANNEL_SLUG", "default-channel")
 
+# Set this to `True` if you want to skip creating default objects in database.
+BLOCK_POPULATE_DEFAULTS = False
```
korycins, line 602: Wouldn't be good to use os.environ here also?

saleor/settings.py
```diff
@@ -598,6 +598,9 @@ def get_host():
 # Slug for channel precreated in Django migrations
 DEFAULT_CHANNEL_SLUG = os.environ.get("DEFAULT_CHANNEL_SLUG", "default-channel")
 
+# Set this to `True` if you want to skip creating default objects in database.
```
korycins, line 601: Would be good to have more context here. Like what kind of the objects we will create etc. `default objects` is a little bit too generic.

saleor/settings.py
korycins: Maybe a name like `CREATE_BASE_STORE_DATA` would be more meaningful?
WDYT @maarcingebala @jakubkuc

maarcingebala: Yes, I would also revert the condition, it's more intuitive to me if things are enabled by default, I would change it to `POPULATE_DEFAULTS = True`. We can use os.environ as well.

saleor/settings.py
maarcingebala: Yep, what default objects?

saleor/product/migrations/0172_create_default_category_and_product_type.py
maarcingebala: It's been discussed already; see my comments above.

saleor/settings.py
```diff
@@ -598,6 +598,10 @@ def get_host():
 # Slug for channel precreated in Django migrations
 DEFAULT_CHANNEL_SLUG = os.environ.get("DEFAULT_CHANNEL_SLUG", "default-channel")
 
+# Set this to `True` if you want to create default channel, warehouse, product type and
+# category. It will enable user to create his first product.
```
maarcingebala, line 602:
```suggestion
# category during migrations. It makes it easier for the users to create their first product.
```

patrys:
Where does the new `POPULATE_DEFAULTS` setting come from?

korycins (approved):

SzymJ (approved):

maarcingebala (approved):

PR #10256, opened 2022-07-28
Fix prepare_thumbnail_file_name function

SzymJ:
I want to merge this change because it fixes `prepare_thumbnail_file_name` function when filename contains `.`

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/thumbnail/utils.py
+++ b/saleor/thumbnail/utils.py
@@ -50,7 +50,7 @@ def get_thumbnail_size(size: Union[str, int]) -> int:
 def prepare_thumbnail_file_name(
     file_name: str, size: int, format: Optional[str]
 ) -> str:
-    file_path, file_ext = file_name.rsplit(".")
+    file_path, file_ext = file_name.rsplit(".", 1)
     file_ext = format or file_ext
     return file_path + f"_thumbnail_{size}." + file_ext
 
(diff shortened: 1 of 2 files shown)


saleor/thumbnail/utils.py
```diff
@@ -50,7 +50,7 @@ def get_thumbnail_size(size: Union[str, int]) -> int:
 def prepare_thumbnail_file_name(
     file_name: str, size: int, format: Optional[str]
 ) -> str:
-    file_path, file_ext = file_name.rsplit(".")
+    file_path, file_ext = file_name.rsplit(".", 1)
```
NyanKiyoshi, line 53: Is it possible to pass a file without an extension?

SzymJ: This function uses already uploaded files, so I think those should always have extensions because we validate them in mutations.

NyanKiyoshi: We would need to check so we are actually sure of it

SzymJ: confirmed @NyanKiyoshi

fowczarek (approved):

NyanKiyoshi (approved):

maarcingebala (approved):

PR #10261, opened 2022-07-29
Fix ORM crash when generating hundreds of search vector in SQL

NyanKiyoshi:
This fixes a recursion error crash when generating hundreds of `SearchVector` for a single SQL update statement.

Known issue: PostgreSQL may reject the statement when thousands of `SearchVector` are being generated with the following error (fixed by #10279):
```
django.db.utils.OperationalError: stack depth limit exceeded
HINT:  Increase the configuration parameter "max_stack_depth" (currently 2048kB), after ensuring the platform's stack depth limit is adequate.
```

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/core/postgres.py
+++ b/saleor/core/postgres.py
@@ -40,11 +40,14 @@ class NoValidationSearchVector(SearchVector, NoValidationSearchVectorCombinable)
     """
 
 
-class FlatSearchVector(Expression):
-    """Generate a SQL statements for combined ``SearchVector`` expressions.
+class FlatConcat(Expression):
+    """Generate a SQL statements for expressions to be concatenated.
 
     This replaces the logic of Django ORM from recursive ``as_sql()`` and
-    ``resolve_expression()``. Allowing to set hundreds of ``SearchVector``.
+    ``resolve_expression()`` from some functions such as ``SearchVector``.
+
+    The function ``django.db.models.functions.text.Concat`` is recursive thus
+    will crash when setting hundreds of expressions to be concatenated.
     """
 
     function = None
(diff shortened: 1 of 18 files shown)


saleor/core/postgres.py
```diff
@@ -35,3 +38,69 @@ class NoValidationSearchVector(SearchVector, NoValidationSearchVectorCombinable)
     This class is only safe to use with expressions that do not contain aggregation
     and/or over clause.
     """
+
+
+class FlatSearchVector(Expression):
```
patrys, line 43: Is it a flat search vector or a flat concatenation? If I understand the code correctly, the resulting expression is something like:

```
(setweight(to_tsvector('foo'), 'A')||setweight(to_tsvector('bar'), 'B')||setweight(to_tsvector('baz'), 'C'))
```

NyanKiyoshi: Yes it is a flat concat, that's a good point I can rename it so it's more general

saleor/product/search.py
```diff
@@ -44,46 +38,41 @@ def update_products_search_vector(products: "QuerySet"):
...
-        search_vector += variants_vector
-
-    return search_vector
+    search_vectors = [
+        NoValidationSearchVector(Value(product.name), config="simple", weight="A"),
+        NoValidationSearchVector(
+            Value(product.description_plaintext), config="simple", weight="C"
+        ),
+        *generate_attributes_search_vector_value(product.attributes.all()),
+        *generate_attributes_search_vector_value(product.attributes.all()),
```
NyanKiyoshi, line 64:
```suggestion
        *generate_attributes_search_vector_value(product.attributes.all()),
```

maarcingebala (approved):

patrys (approved):

SzymJ (approved):

saleor/product/search.py
```diff
@@ -44,46 +38,40 @@ def update_products_search_vector(products: "QuerySet"):
...
+        *generate_variants_search_vector_value(product),
+    ]
+    return search_vectors
 
 
 def generate_variants_search_vector_value(
     product: "Product",
-) -> Optional[NoValidationSearchVector]:
+) -> List[NoValidationSearchVector]:
```
jakubkuc, line 71: If there is possibility that a list returned will be empty shouldn't we use `List[Optional[NoValidationSearchVector]]` ? The same goes to all other methods that generates list of `NoValidationSearchVector`

jakubkuc (approved):

saleor/product/search.py
NyanKiyoshi: There is no null values inside the list, the typing is valid for empty lists

PR #10288, opened 2022-08-01
Move product search vector indexing to scheduled task

SzymJ:
I want to merge this change because it adds a scheduler for updating products search vectors. A new field  `search_index_dirty` is added to `Product` model. After `AttributeValue` is updated flag `search_index_dirty` is set to `True` for all products with this attribute. The scheduler will run every 20 seconds to index vectors to products with `search_index_dirty=True`.

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-7775 Replace product search indexing task with a dirty flag and a periodic task](https://app.clickup.com/t/2549495/SALEOR-7775)

saleor/graphql/attribute/tests/mutations/test_attribute_value_delete.py
```diff
@@ -37,6 +37,29 @@ def test_delete_attribute_value(
...
+    value = product.attributes.all()[0].values.first()
+    query = ATTRIBUTE_VALUE_DELETE_MUTATION
+    node_id = graphene.Node.to_global_id("AttributeValue", value.id)
+    variables = {"id": node_id}
+
+    # when
+    staff_api_client.post_graphql(
+        query, variables, permissions=[permission_manage_product_types_and_attributes]
+    )
+    product.refresh_from_db()
```
NyanKiyoshi, line 55:
```suggestion
    product.refresh_from_db(fields=["search_index_dirty"])
```

saleor/graphql/attribute/tests/mutations/test_attribute_value_update.py
```diff
@@ -67,6 +67,28 @@ def test_update_attribute_value(
...
+    value = product.attributes.all()[0].values.first()
+    node_id = graphene.Node.to_global_id("AttributeValue", value.id)
+    name = "Crimson name"
+    variables = {"input": {"name": name}, "id": node_id}
+
+    # when
+    staff_api_client.post_graphql(
+        query, variables, permissions=[permission_manage_product_types_and_attributes]
+    )
+    product.refresh_from_db()
```
NyanKiyoshi, line 86:
```suggestion
    product.refresh_from_db(fields=["search_index_dirty"])
```

saleor/product/tests/test_tasks.py
```diff
@@ -114,3 +115,16 @@ def test_get_preorder_variants_to_clean(
...
+
+
+def test_update_products_search_vector_task(product):
+    # given
+    product.search_index_dirty = True
+    product.save()
+
+    # when
+    update_products_search_vector_task()
+    product.refresh_from_db()
```
NyanKiyoshi, line 127:
```suggestion
    product.refresh_from_db(fields=["search_index_dirty"])
```

saleor/product/tests/test_tasks.py
```diff
@@ -114,3 +115,16 @@ def test_get_preorder_variants_to_clean(
     variants_to_clean = _get_preorder_variants_to_clean()
     assert len(variants_to_clean) == 1
     assert variants_to_clean[0] == preorder_variant_after_end_date
+
+
+def test_update_products_search_vector_task(product):
+    # given
+    product.search_index_dirty = True
+    product.save()
```
NyanKiyoshi, line 123:
```suggestion
    product.save(update_fields=["search_index_dirty"])
```

saleor/product/tasks.py
```diff
@@ -108,3 +112,9 @@ def _get_preorder_variants_to_clean():
     return ProductVariant.objects.filter(
         is_preorder=True, preorder_end_date__lt=timezone.now()
     )
+
+
+@app.task
```
NyanKiyoshi, line 117: Could you use `queue=settings.UPDATE_INDEX_QUEUE_NAME` so users can use a separate queue for it? By default it should be `None`, I believe that's valid to tell Celery to use the main queue (do check if that's valid)

saleor/product/search.py
```diff
@@ -27,27 +27,34 @@
...
-        Product.objects.bulk_update(products_batch, ["search_vector", "updated_at"])
+def update_products_search_vector(products: "QuerySet", use_batches=True):
+    if use_batches:
+        last_id = 0
+        while True:
+            products_batch = list(products.filter(id__gt=last_id)[:PRODUCTS_BATCH_SIZE])
+            if not products_batch:
+                break
+            last_id = products_batch[-1].id
+            _prep_product_search_vector_index(products_batch)
```
NyanKiyoshi, line 55: Is the code path for `use_batches=True` still needed?

patrys (approved):

saleor/product/migrations/0171_product_search_index_dirty.py
```diff
@@ -0,0 +1,18 @@
...
+        ("product", "0170_rewrite_digitalcontenturl_orderline_relation"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="product",
+            name="search_index_dirty",
+            field=models.BooleanField(default=False),
+        ),
+    ]
```
NyanKiyoshi, line 18: Make sure the users can upgrade to 3.5 safely once the same migration will be added to 3.5/main

saleor/settings.py
```diff
@@ -555,6 +555,10 @@ def get_host():
         "task": "saleor.csv.tasks.delete_old_export_files",
         "schedule": crontab(hour=1, minute=0),
     },
+    "update-products-search-vectors": {
+        "task": "saleor.product.tasks.update_products_search_vector_task",
+        "schedule": timedelta(seconds=20),
+    },
```
NyanKiyoshi, line 561: What if it takes longer than 20s to execute the task? Won't it reindex multiple times the same products?

saleor/product/search.py
SzymJ: yes

NyanKiyoshi: Shouldn't all those reference be handled by a task? Especially the ones that update N products after updating attributes

SzymJ: This  PR change is only for AttributeValueUpdate/Delete mutation, but this function is used in other places where we are passing products without limitations. We have to handle those places in other PRs.

NyanKiyoshi: Can you create a task for that?

saleor/product/tasks.py
SzymJ: I dont see setting like this anywhere

saleor/settings.py
SzymJ: @patrys any thoughts about this?

maarcingebala: It might happen that two tasks picked up the same products for an update, but if it happens for some small set of products and it reindexes them multiple times, would that be a big problem? Maybe we should actually measure how much time it usually takes and adjust the time delta?

maarcingebala (approved):

saleor/settings.py
patrys: It will spawn a Celery task every 20 seconds. There is no guarantee that the task will even start executing within the 20-second window. I'm not aware of better alternatives as locking product rows is going to hurt business logic (it would block transactional code).

SzymJ: so should we stay with this for now, check how it works and adjust later if it will be needed?

saleor/settings.py
```diff
@@ -719,4 +723,6 @@ def SENTRY_INIT(dsn: str, sentry_opts: dict):
     seconds=parse(os.environ.get("JWT_TTL_REQUEST_EMAIL_CHANGE", "1 hour")),
 )
 
-# Support multiple interface notation in schema for Apollo tooling.
+UPDATE_SEARCH_VECTOR_INDEX_QUEUE_NAME = os.environ.get(
+    "UPDATE_SEARCH_VECTOR_INDEX_QUEUE_NAME", "update-products-search-vector"
+)
```
NyanKiyoshi, line 728: I would say to put it to `None` by default, otherwise users will need to setup a separate Celery worker by default when they may not actually need it

SzymJ: Done

PR #10320, opened 2022-08-04
Fix assigning an email address that does not belong to an existing user to draft order

SzymJ:
I want to merge this change because it fixes bug that does not allow assigning not existing email addresses to draft orders.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/mutations/draft_order_create.py
+++ b/saleor/graphql/order/mutations/draft_order_create.py
@@ -117,6 +117,9 @@ class DraftOrderCreate(ModelMutation, I18nMixin):
             object_id=data.pop("shipping_method", None), raise_error=False
         )
 
+        if data.get("user_email", None):
+            data["user"] = None
+
         cleaned_input = super().clean_input(info, instance, data)
 
         channel = cls.clean_channel_id(info, instance, cleaned_input, channel_id)
--- a/saleor/graphql/order/mutations/draft_order_create.py
+++ b/saleor/graphql/order/mutations/draft_order_create.py
@@ -6,6 +6,7 @@ from django.core.exceptions import ValidationError
 from django.db import transaction
 from graphene.types import InputObjectType
 
+from ....account.models import User
 from ....checkout import AddressType
 from ....core.permissions import OrderPermissions
 from ....core.taxes import TaxError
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [SALEOR-7832 Cannot unassign a user from an order](https://app.clickup.com/t/2549495/SALEOR-7832)

saleor/graphql/order/mutations/draft_order_create.py
```diff
@@ -117,6 +117,9 @@ def clean_input(cls, info, instance, data):
             object_id=data.pop("shipping_method", None), raise_error=False
         )
 
+        if data.get("user_email", None):
+            data["user"] = None
```
patrys, line 121: Does it still work if the email matches an existing user?

korycins: The order resolvers, first try to fetch email from `order.user.email`, then as a fallback it tries to fetch `order.user_email`.

patrys: My question is will it assign the correct user?

SzymJ: @patrys right, it works on dashboard, because if provided email exists FE will send `user` key not ` user_email` but if someone will use playground and send existing email in `user_email` it will not assign that user to order. I will fix it.

saleor/graphql/order/mutations/draft_order_create.py
```diff
@@ -117,6 +118,13 @@ def clean_input(cls, info, instance, data):
             object_id=data.pop("shipping_method", None), raise_error=False
         )
 
+        if email := data.get("user_email", None):
+            try:
+                user = User.objects.get(email=email, is_staff=False)
```
patrys, line 123: Is email matching the same in all places we do email matching? For example, is it case-sensitive everywhere? If not, could looking up a user by email be a helper function?

korycins: Is `is_staff=False` mandatory here? Staff user also could be a customer, right? I think we should check `is_active=True` also.

SzymJ: @patrys It looks like it's always case-sensitive, in all places we put `email` data to `get()` `filter()` objects methods directly from the input without any lowercasing, etc.

IKarbowiak (approved):

korycins (approved):

PR #10332, opened 2022-08-08
Allow tracking number to be passed to orderFulfill

jakubkuc:
I want to merge this change because it allows to add optional `tracking_number` to `orderFulfill` mutation. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/mutations/order_fulfill.py
+++ b/saleor/graphql/order/mutations/order_fulfill.py
@@ -10,6 +10,7 @@ from ....core.tracing import traced_atomic_transaction
 from ....order import models as order_models
 from ....order.actions import create_fulfillments
 from ....order.error_codes import OrderErrorCode
+from ...core.descriptions import ADDED_IN_36
 from ...core.mutations import BaseMutation
 from ...core.types import NonNullList, OrderError
 from ...core.utils import get_duplicated_values
@@ -55,7 +56,7 @@ class OrderFulfillInput(graphene.InputObjectType):
         default_value=False,
     )
     tracking_number = graphene.String(
-        description="Fulfillment tracking number.",
+        description="Fulfillment tracking number." + ADDED_IN_36,
         required=False,
     )
 
--- a/saleor/graphql/order/mutations/order_fulfill.py
+++ b/saleor/graphql/order/mutations/order_fulfill.py
@@ -54,6 +54,10 @@ class OrderFulfillInput(graphene.InputObjectType):
         description="If true, then allow proceed fulfillment when stock is exceeded.",
         default_value=False,
     )
+    tracking_number = graphene.String(
+        description="Fulfillment tracking number.",
+        required=False,
+    )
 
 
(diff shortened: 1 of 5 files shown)


patrys:
Task linked: [SALEOR-7831 Allow tracking number to be passed to `orderFulfill`](https://app.clickup.com/t/2549495/SALEOR-7831)

maarcingebala (approved):

SzymJ (approved):

saleor/graphql/order/mutations/order_fulfill.py
```diff
@@ -54,6 +54,10 @@ class OrderFulfillInput(graphene.InputObjectType):
         description="If true, then allow proceed fulfillment when stock is exceeded.",
         default_value=False,
     )
+    tracking_number = graphene.String(
+        description="Fulfillment tracking number.",
```
korycins, line 58: Shouldn't be ADDED_IN_3X?

saleor/order/actions.py
```diff
@@ -786,7 +788,9 @@ def create_fulfillments(
         else FulfillmentStatus.WAITING_FOR_APPROVAL
     )
     for warehouse_pk in fulfillment_lines_for_warehouses:
-        fulfillment = Fulfillment.objects.create(order=order, status=status)
+        fulfillment = Fulfillment.objects.create(
```
korycins, line 791: Setting `tracking number` previously we were also trigger hook: `tracking_number_updated`. We don't have any webhook for. such event but we have a payment plugin (`NPAtobaraiGatewayPlugin`)that listen to this event. AFAIR this payment plugin was using tracking number to process a payment. I think that to kepp an consistency, we should trigger hook here.

saleor/graphql/schema.graphql
```diff
@@ -17591,6 +17591,9 @@ input OrderFulfillInput {
 
   """If true, then allow proceed fulfillment when stock is exceeded."""
   allowStockToBeExceeded: Boolean = false
+
```
korycins, line 17594: Is it a really good place for this parameter?
`OrderFulfill` allows you to create multiple different fulfillments (based on warehousePK). If we add this input here, no matter how many fulfillments we will create, all of them will have the same tracking number. (And it could not be true, as each fulfillment comes from a different warehouse).
Example. 
In the input, we provide `line: {Juce , quantity:1, warehouseID:13}`, and the second `line: {Apple, quanity:2, warehouse:33}`. This will create two separate fulfillments, and both of them will get the same `trackingNumber`. (But they came from different warehouses). 
cc @maarcingebala @jakubkuc @patrys

jakubkuc: It can be changed to be passed inside lines ( in `OrderFulfillLineInput` ) so that every separate fulfillment from line has  its own tracking number.

patrys: @korycins That’s expected. There is no per-warehouse input and we can’t have a separate tracking number for each line.

PR #10346, opened 2022-08-10
Add bcrypt password hasher

pkucmus:
I'd like to add the bcrypt password hasher to saleor as that would help to move to Saleor from other platforms. 
Django handles other hashing algorithms than the one used for storing passwords in this case I'm seeking support for password entries that start with a `bcrypt$` prefix.

I've added an explicit list of `settings.PASSWORD_HASHERS` which till now defaulted to https://docs.djangoproject.com/en/3.2/ref/settings/#std-setting-PASSWORD_HASHERS. This along with the additional `bcrypt` package in the requirements allows me to use the `django.contrib.auth.hashers.BCryptPasswordHasher` hasher (not to mistake it with the `BCryptSHA256PasswordHasher` (notice the SHA256 part) that comes out of the box. 

With that change, I'm able to check passwords that were hashed by different e-commerce solutions:

```pycon
>>> from django.contrib.auth import get_user_model
>>> from django.contrib.auth.hashers import get_hasher, check_password
>>> User = get_user_model()
>>> user = User.objects.all()[1]
>>> get_hasher("bcrypt")
<django.contrib.auth.hashers.BCryptPasswordHasher object at 0x7fdc0013ed60>
>>> encoded = "$2a$10$RrHEKriysPojlt7jJWluQuuNs5i0agqFK1T4WATvngkejHbB3QxSy"  # incoming from a different solution
>>> check_password("lol12345", "bcrypt$" + encoded, None, "bcrypt")
True
```

while not impacting how new passwords are created:

```pycon
>>> user.set_password("admin")
>>> user.password
'pbkdf2_sha256$260000$SyeE2fdjs4PYdGQ3dX439g$3GW+ECMzi4EM0uAFiP8bYNqRJJKJM5e5MmD67nDBg+M='
>>> user.set_password("admin2")
>>> user.password
'pbkdf2_sha256$260000$163303SdHNS0vjP9isZ9BH$yrsKkEMnyVHenPQMI9tgll5K3YeuiyS9TASZNwCa/1s='
```

Django will also update the obsolete bcrypt hash of a successfully used password to the one specified in `settings.PASSWORD_HASHERS[0]` (ref: https://docs.djangoproject.com/en/3.2/topics/auth/passwords/#how-django-stores-passwords)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -11,6 +11,7 @@ import pkg_resources
 import sentry_sdk
 import sentry_sdk.utils
 from celery.schedules import crontab
+from django.conf import global_settings
 from django.core.exceptions import ImproperlyConfigured
 from django.core.management.utils import get_random_secret_key
 from graphql.execution import executor
@@ -175,11 +176,11 @@ TEMPLATES = [
 # Make this unique, and don't share it with anybody.
 SECRET_KEY = os.environ.get("SECRET_KEY")
 
+# Add additional password algorithms that can be used
+# The password hashes not using the preferred algorithm (first in the list)
+# will be automatically converted upon user login
 PASSWORD_HASHERS = [
-    "django.contrib.auth.hashers.PBKDF2PasswordHasher",
-    "django.contrib.auth.hashers.PBKDF2SHA1PasswordHasher",
-    "django.contrib.auth.hashers.Argon2PasswordHasher",
-    "django.contrib.auth.hashers.BCryptSHA256PasswordHasher",
+    *global_settings.PASSWORD_HASHERS,
     "django.contrib.auth.hashers.BCryptPasswordHasher",
 ]
 
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -176,9 +176,9 @@ TEMPLATES = [
 # Make this unique, and don't share it with anybody.
 SECRET_KEY = os.environ.get("SECRET_KEY")
 
-# Add additional password algorithms that can be used
-# The password hashes not using the preferred algorithm (first in the list)
-# will be automatically converted upon user login
+# Additional password algorithms that can be used by Saleor.
+# The first algorithm defined by Django is the preferred one; users not using the
+# first algorithm will automatically be upgraded to it upon login
 PASSWORD_HASHERS = [
     *global_settings.PASSWORD_HASHERS,
     "django.contrib.auth.hashers.BCryptPasswordHasher",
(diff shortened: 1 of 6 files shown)


saleor/settings.py
```diff
@@ -175,6 +175,14 @@ def get_bool_from_env(name, default_value):
 # Make this unique, and don't share it with anybody.
 SECRET_KEY = os.environ.get("SECRET_KEY")
 
+PASSWORD_HASHERS = [
+    "django.contrib.auth.hashers.PBKDF2PasswordHasher",
+    "django.contrib.auth.hashers.PBKDF2SHA1PasswordHasher",
+    "django.contrib.auth.hashers.Argon2PasswordHasher",
+    "django.contrib.auth.hashers.BCryptSHA256PasswordHasher",
+    "django.contrib.auth.hashers.BCryptPasswordHasher",
+]
```
NyanKiyoshi, line 184: Maybe we could import the default hasher list from django and add the ones we want instead so we always have the preferred hashers of Django.
Someday `PBKDF2PasswordHasher` may no longer be the default/preferred.

So in the case of this PR it would only add `BCryptPasswordHasher`.



https://github.com/django/django/blob/0dd29209091280ccf34e07c9468746c396b7778e/django/conf/global_settings.py#L535-L541

maarcingebala (approved):

Dieg0x17:
Could be useful to add custom hashers to the PASSWORD_HASHERS list.

saleor/settings.py
patrys: @NyanKiyoshi can you change it if you think it's blocking the merge?

NyanKiyoshi (approved):

saleor/settings.py
```diff
@@ -175,6 +176,14 @@ def get_bool_from_env(name, default_value):
 # Make this unique, and don't share it with anybody.
 SECRET_KEY = os.environ.get("SECRET_KEY")
 
+# Add additional password algorithms that can be used
+# The password hashes not using the preferred algorithm (first in the list)
+# will be automatically converted upon user login
```
IKarbowiak, line 181: To be honest, I have a problem with understanding this description. Could you improve that? There is no dot between the first and second lines. I also do not understand what will be converted and into what.

NyanKiyoshi: Does that make more sense now?
```suggestion
# Additional password algorithms that can be used by Saleor.
# The first algorithm defined by Django is the preferred one; users not using the
# first algorithm will automatically be upgraded to it upon login

PR #10350, opened 2022-08-11
[RFC] Support for payment transaction actions

korycins:
The changes are related to the `transactionItem` that we introduced in previous releases. The proposed changes improve and simplify the usage of the interface that we added previously. It also adds an interface that can be used to handle the actions like `refund`, `charge`, or `cancel`. The main assumption it that we use webhook to notify the app, that user requested the action.

# Current Status
The status of work related to implement the feature: 
- [x] added granted refund types
- [x] added granted refund mutations
- [x] added new fields to order related to granted refunds
- [x] added support for manual transactions
- [ ] added `Pending` `Amount`s to `transactionItem` case:
	- `pendingRefundAmount`
	- `pengingChargeAmount` 
	- `pendingAuthorizeAmount`
- [ ] Request Refund
    - [ ] added mutation `transactionEventReport`
    - [x] added new webhooks `TransactionChargeRequested`, `TransactionRefundRequested`, `TransactionCancelationRequested` (hardcoded+subscription payload) - #11263
    - [x] extended TransactionActionRequest to create `REQUEST` event
- [ ] added logic for re-calculation the amounts based on the transaction's events
- [ ] added changes to transactionItem required for new flow
- [ ] transactionCreate, transactionUpdate
   - de-duplication of the transactions
   - pspReference & reference
- [ ] `parse_transaction_action_data` should take into acount the type of the request. If we're `CHARGE_REQUEST`, the response should contain `CHARGE_SUCCESS` or `CHARGE_FAILURE`
- [ ] deprecate `transactionEvent.status` from `transactionCreate` and `transactionUpdate`
- [ ] added `message` field to `transactionEvent` - #11476
- [ ] added `user`&`app` to `TransactionEvent`
- [x] update logic for return&replace and refund for old payments - #11378
- [ ] update the docs
- [ ] `OrderEvents`, rename transaction events `VOID`->`CANCEL`, `CAPTURE`->`CHARGE`
- [ ] `TransactionItem` inputs and field for `void` should be renamed to `CANCEL`
- [ ] Drop from webhooks response, `response.event.pspReference` field, use field `response.pspRefernce` - #11476
- [ ] Drop changes added to types like `TransactionEventInput` that are not needed. - #11476
- [ ] add migration guide (placed at the end of the description) [need to be double checked and compared with all feature breaking changes that we introduced]
- [ ] collapse all introduced migrations (if possible) - #11476
- [ ] proceed migration test from 3.7 with existing transactionItem and transactionEvent. Make sure that all fields that become unique now are handled correctly
- [ ] Connect transaction with app by identifier: The transaction will store App identifier instead of direct relation between object. This will allow us to better match transactions with the owner app.
- [ ] Update feature branch to contain correct `ADDED_IN_3X` labels
- [ ] Make sure that changelog is up-to-date

# 1. General Assumptions
- requesting an action on `transactionItem` is an async operation from Staff user perspective
- `refund` is not strictly related with `return`
- each action will have own webhook event
- new object, `GrantedRefund` will determine the amount of the refund that should be returned to user.
- Staff user can use actions like `charge`, `refund` and `cancel`.

# 2. API Changes
#### 2.1.1.1 GrantedRefund type
```graphql
type GrantedRefund {
  amount: Money
  requester: App | User
  message: String
  created: DateTime
}
```
> **Note**  `requester: App | User` will work only if App /User exists.

#### 2.1.1.2 orderGrantRefundCreate
```graphql
type OrderGrantRefundCreate {
  """Order which has assigned new grant refund."""
  order: Order

  """Created granted refund."""
  grantedRefund: OrderGrantedRefund
  errors: [OrderGrantRefundCreateError!]!
}
type Mutation{
  """Requires MANAGE_ORDERS permission"""
  orderGrantRefundCreate(
    id: ID!,
    input: {
      amount: Money
      message: String
    }
  ): OrderGrantRefundCreate
}
```
#### 2.1.1.3 orderGrantRefundUpdate
```graphql
type OrderGrantRefundUpdate {
  """Order which has assigned new grant refund."""
  order: Order

  """Created granted refund."""
  grantedRefund: OrderGrantedRefund
  errors: [OrderGrantRefundCreateError!]!
}
type Mutation{
  """Requires MANAGE_ORDERS permission."""
  orderGrantRefundUpdate(
    id: ID!,
    input: {
      amount: Money
      message: String
    }
  ): OrderGrantRefundUpdate
}
```
#### 2.1.2 TransactionEventTypeEnum
```graphql
enum TransactionEventTypeEnum {
  AUTHORIZATION_SUCCESS
  AUTHORIZATION_FAILURE
  AUTHORIZATION_ADJUSTMENT
  AUTHORIZATION_REQUEST
  CHARGE_SUCCESS
  CHARGE_FAILURE
  CHARGE_BACK
  CHARGE_REQUEST
  REFUND_SUCCESS
  REFUND_FAILURE
  REFUND_REVERSE
  REFUND_REQUEST
  CANCEL_SUCCESS
  CANCEL_FAILURE
  CANCEL_REQUEST
}
```
#### 2.1.3 transactionEventReport
```graphql
type Mutation{
  transactionEventReport(
    transactionId: ID!
    pspReference: String!
    type: TransactionEventType!
    amount: PositiveDecimal
    time: DateTime
    externalUrl: String
    message: String,
    availableActions: [TransactionActionEnum!]
  ): TransactionEventReport
}

type TransactionEventReport{
  transaction: TransactionItem,
  event: TransactionEvent
  errors: [TransactionEventReportError!]!
}

type TransactionEventReportError {
  """
  Name of a field that caused the error. A value of `null` indicates that the error isn't associated with a particular field.
  """
  field: String

  """The error message."""
  message: String

  """The error code."""
  code: TransactionRequestCompleteErrorCode!
}

"""An enumeration."""
enum TransactionRequestCompleteErrorCode {
  INVALID
  GRAPHQL_ERROR
  NOT_FOUND
  INCORRECT_DETAILS
}

```

### 2.2.1 Order
```graphql
"""Represents an order in the shop."""
type Order implements Node & ObjectWithMetadata {
  id: ID!

  ...
  """List of all granted refunds"""
  grantedRefunds: [
    GrantedRefund!
  ]!
  """ Total amount of granted refund """
  totalGrantedRefund: Money!

  """Sum of order.transactions.all().refundedAmount"""
  totalRefunded: Money!

  """Sum of order.transactions.all().pendingRefundAmount"""
  totalPendingRefund: Money!

  """Sum of order.transactions.all().pendingChargeAmount"""
  totalPendingCharge: Money!

  """Amount that is not yet refunded: grantedRefund - totalRefunded - totalPendingRefund """
  totalRemainingGrant: Money!
}
```
### 2.2.2 TransactionItem
```graphql

type TransactionItem {
   ...
   createdBy: App | User
   externalUrl: String!
   pspReference: String!
   reference: String! @deprecated(reason: "This field will be removed in Saleor 3.10 (Feature Preview). Use `pspReference` instead.")
   """Status of transaction."""
   status: String! @deprecated(reason: "This field will be removed in Saleor 3.10 (Feature Preview). The status is not needed. The amounts can be used to define the current status of transactions.")

   pendingRefundAmount: Money!
   pengingChargeAmount: Money!
   pendingAuthorizeAmount: Money!
}
```
> **Note**  `createdBy: App | User` will work only if App /User exists. I think this is enough for now. In the future we could consider extending this part, to provide some denormalized data related to user/app.
### 2.2.3 TransactionEvent
```graphql

"""Represents transaction's event."""
type TransactionEvent implements Node {
  ...
  status: TransactionStatus! @deprecated(Use TransactionEventTypeEnum instead.)
  reference: String! @deprecated(reason: "This field will be removed in Saleor 3.10 (Feature Preview). Use `pspReference` instead.")
  pspReference: String!

  name: String @deprecated(Use message. message == message+name)

  message: String!

  externalUrl: String!

  amount: Money!

  type: TransactionEventTypeEnum! (2.1.2)
  createdBy: App | User
}
```
### 2.2.4 TransactionCreateInput
```graphql
input TransactionCreateInput {
  ...
  """[DEPRECEATED] Reference of the transaction."""
  reference: String

  pspReference: String
}
```

### 2.2.5 TransactionUpdateInput
```graphql
input TransactionUpdateInput {
  ...
  """[DEPRECEATED] Reference of the transaction."""
  reference: String

  pspReference: String
}
```

### 2.2.6 OrderReturnProductsInput
```graphql
input OrderReturnProductsInput {
  ...
  """DEPRECEATED. The fields related to refunds will be deprecated. It will not work when the order has attached transactions. It will still allow the processing of old payments. """
  amountToRefund: PositiveDecimal

  """DEPRECEATED. The fields related to refunds will be deprecated. It will not work when the order has attached transactions. It will still allow the processing of old payments. """
  includeShippingCosts: Boolean = false

  """DEPRECEATED. The fields related to refunds will be deprecated. It will not work when the order has attached transactions. It will still allow the processing of old payments. """
  refund: Boolean = false
}
```

### 2.2.7 OrderRefundProductsInput
```graphql
input OrderRefundProductsInput {
  ...

  """DEPRECEATED. The fields related to refunds will be deprecated. It will not work when the order has attached transactions. It will still allow the processing of old payments. """
  amountToRefund: PositiveDecimal

  """DEPRECEATED. The fields related to refunds will be deprecated. It will not work when the order has attached transactions. It will still allow the processing of old payments. """
  includeShippingCosts: Boolean = false
}
```

# 3. Granted refunds
The idea is that we use `grantRefund` mutations to declare what amount should be refunded by someone with handle payment permissions. It doesn't have any impact on order values as this is only the "declaration" of future refund.
The list of granted refunds attached to the orders can be fetched by querying `order.grantedRefunds` (2.2.1).
`order.totalGrantedRefund` (2.2.1) stores the total amount of the granted refund.
The mutation `orderGrantRefundCreate` (2.1.1.2) can be used to create a new granted refund, that will be attached to the order. It determines the amount of the refund granted for the order. The next step is to trigger a refund.
The mutation `orderGrantRefundUpdate`(2.1.1.3) allows updating message and amount of the given granted refund.

# 4. Manual transactions
We will allow the creation of new transactions by staff users who have assigned permission handle_payments. Previously all transaction mutations  (`transactionCreate`, `transactionUpdate`) were requiring an app token.

The idea of this change is to give a staff user the possibility to create a manual transaction in a case when payment has been processed in a different channel than we support (for example: refund has been made in cash etc). 

- The transaction object (2.2.2) will get a field `createdBy` that will store User or App object. Depends on who is the creator of the transaction.
- The transaction will store App identifier instead of direct relation between object. This will allow us to better match transaction with owner app.
# 5. Webhooks
## 5.1 `TransactionChargeRequested`
SYNC webhook called inside celery task. Called when staff user requests the charge action. It expects in the response at least `pspReference` (5.1.2). The `pspReference` will be attached to `REQUEST` `TransactionEvent` object. Attached `pspReference` to `REQUEST` object means that the webhook has been successfully processed by App. Optionally the `event` (5.1.2) data can be provided. The `event` data is used to provide the final status of the requested action, and it will be used to create a new `TransactionEvent` object. The new `TransactionEvent` object will have attached provided `pspReference`.
### 5.1.1 Request
The request is the same as for `TransactionActionRequest` webhook. The subscription query is also supported.
```graphql
type TransactionChargeRequested implements Event {
  """Time of the event."""
  issuedAt: DateTime

  """Saleor version that triggered the event."""
  version: String

  """The user or application that triggered the event."""
  issuingPrincipal: IssuingPrincipal

  """The application receiving the webhook."""
  recipient: App

  """Look up a transaction."""
  transaction: TransactionItem

  """Requested action data."""
  action: TransactionAction!
}
```
### 5.1.2 Response
The structure of the response:
```
{
  pspReference: String,
  event: {
    type: TransactionEventType!
    amount: PositiveDecimal!
    time: DateTime
    externalUrl: String 
    message: String
  }
}
```

## 5.2 `TransactionRefundRequested`
SYNC webhook called inside celery task. Called when staff user requests the refund action. It expects in the response at least `pspReference` (5.2.2). The `pspReference` will be attached to `REQUEST` `TransactionEvent` object. Attached `pspReference` to `REQUEST` object means that the webhook has been successfully processed by App. Optionally the `event` (5.2.2) data can be provided. The `event` data is used to provide the final status of the requested action, and it will be used to create a new `TransactionEvent` object. The new `TransactionEvent` object will have attached provided `pspReference`.
### 5.2.1 Request
The request is the same as for `TransactionActionRequest` webhook. The subscription query is also supported.
```graphql
type TransactionRefundRequested implements Event {
  """Time of the event."""
  issuedAt: DateTime

  """Saleor version that triggered the event."""
  version: String

  """The user or application that triggered the event."""
  issuingPrincipal: IssuingPrincipal

  """The application receiving the webhook."""
  recipient: App

  """Look up a transaction."""
  transaction: TransactionItem

  """Requested action data."""
  action: TransactionAction!
}
```
### 5.2.2 Response
The structure of the response:
```
{
  pspReference: String,
  event: {
    type: TransactionEventType!
    amount: PositiveDecimal!
    time: DateTime
    externalUrl: String 
    message: String
  }
}
```
## 5.3 `TransactionCancelationRequested`
SYNC webhook called inside celery task. Called when staff user requests the cancel action. It expects in the response at least `pspReference` (5.3.2). The `pspReference` will be attached to `REQUEST` `TransactionEvent` object. Attached `pspReference` to `REQUEST` object means that the webhook has been successfully processed by App. Optionally the `event` (5.3.2) data can be provided. The `event` data is used to provide the final status of the requested action, and it will be used to create a new `TransactionEvent` object. The new `TransactionEvent` object will have attached provided `pspReference`.
### 5.3.1 Request
The request is the same as for `TransactionActionRequest` webhook. The subscription query is also supported.
```graphql
type TransactionCancelationRequested implements Event {
  """Time of the event."""
  issuedAt: DateTime

  """Saleor version that triggered the event."""
  version: String

  """The user or application that triggered the event."""
  issuingPrincipal: IssuingPrincipal

  """The application receiving the webhook."""
  recipient: App

  """Look up a transaction."""
  transaction: TransactionItem

  """Requested action data."""
  action: TransactionAction!
}
```
### 5.3.2 Response
The structure of the response:
```
{
  pspReference: String,
  event: {
    type: TransactionEventType!
    amount: PositiveDecimal
    time: DateTime
    externalUrl: String 
    message: String
  }
}
```
# 6. Action flow
The idea for requesting an action, is to give a staff user (with handle payments permission) the possibility to trigger a request for an action. The granted refund amount created in the previous step should be helpful in creating a new request refund. 
Calling mutation `TransactionActionRequest` will also create a new `transactionEvent` with status `REQUEST`, amount and the requester (user/app). Saleor will send two webhooks: `transactionActionRequest` (which will be deprecated in future releases) and webhook dedicated to the action (`TransactionChargeRequested`, `TransactionRefundRequested`, `TransactionCancelationRequested`). The new webhooks will be called in celery task but it will be sync request: it will require a response from the app. The response should contain at least `pspReference` of the action. The `pspReference` will be placed in previously created event of `REQUEST` type. Optionally the response can contain the details of the completed action (5.). 

Only the app that is owner of `transactionItem`  should recieve an action webhook (`TransactionChargeRequested` 5.1, `TransactionRefundRequested` 5.2, `TransactionCancelationRequested` 5.3). If `transactionItem` doesn't have any relation we should call the event for all related webhooks.

## 6.1 ASYNC payment processing
 In case when refund is processing asynchronously on payment provider side, app should call `transactionEventReport` mutation (2.1.3) when it will recieve a webhook notification from payment provider.

 ```mermaid
sequenceDiagram
    Staff ->>+ Saleor: Create return <br>fulfillments for 2 <br> products (total : 50 USD)
    Saleor -->>- Staff: Created
    Staff ->>+ Saleor: Create granted <br>refund of 50 USD
    Saleor -->>- Staff: Granted refund created
    Staff ->>+ Saleor: Request refund<br>[transactionActionRequest]
    Saleor ->> Saleor: Create REQUEST event for transaction
    Saleor ->> Saleor: Schedule transaction action webhook
    Saleor -->>- Staff: Requested
    Note over Saleor,App: Celery task
    Saleor -->>+ App: Send transaction-refund-requested <br>webhook
    App ->>+ PaymentProvider: Request refund
    PaymentProvider -->>- App: Processing refund
    App -->>- Saleor: PSP reference
    Note over Saleor,PaymentProvider: Async notification
    PaymentProvider -)+ App: Refund processed
    App ->>+ Saleor: Call transactionEventReport <br>mutation with all details<br> of completed action
    Saleor ->> Saleor: Update transaction amounts
    Saleor -->>- App: Updated
    App --)- PaymentProvider: OK
```

## 6.2 SYNC payment processing
 In case when refund is processing synchronously on payment provider side. The app receives the completed status of the action as a response to the request of the action, then the App can pass the details of the action in the response of the webhook (`TransactionChargeRequested` 5.1, `TransactionRefundRequested` 5.2, `TransactionCancelationRequested` 5.3).

 ```mermaid
sequenceDiagram
    Staff ->>+ Saleor: Create return <br>fulfillments for 2 <br> products (total : 50 USD)
    Saleor -->>- Staff: Created
    Staff ->>+ Saleor: Create granted <br>refund of 50 USD
    Saleor -->>- Staff: Granted refund created
    Staff ->>+ Saleor: Request refund<br>[transactionActionRequest]
    Saleor ->> Saleor: Create REQUEST event for transaction
    Saleor ->> Saleor: Schedule transaction action webhook
    Saleor -->>- Staff: Requested
    Note over Saleor,App: Celery task
    Saleor ->>+ App: Send transaction-refund-requested <br>webhook
    App ->>+ PaymentProvider: Request refund
    PaymentProvider -->>- App: Refunded
    App -->>- Saleor: Return PSP<br> Reference and details<br> about completed action
    Saleor ->> Saleor: Update transaction amounts
```

## 6.3 transactionEventReport
`transactionEventReport` should be used to report the new event for the transaction. The newly created event will be used to re-calculate (6.4) the transactionItem's amounts.
The mutation should be used in case of asynchronously processing the requested action. It should be also used to report any changes that happened on payment provider side.
## 6.4 Re-calculations of transactionItem's amounts
Based on the value of the enum `TransactionEventTypeEnum` (2.1.2) used in `transactionEventReport` (6.3) and optionally in webhook response (5.), the re-calculation will be different:

### 6.4.1 AUTHORIZATION_SUCCESS
Assigns provided `amount` to `transaction.authorizedAmount`. Can’t have two success authorization for single transaction. The next requests after the first one will raise an exception. `AUTHORIZATION_ADJUSTMENT` should be used to change the `transaction.authorizedAmount`

> **Warning**
> In the case of finding `AUTHORISATION_SUCCESS` and `AUTHORISATION_FAILURE` with the same `pspReference`, the re-calculation for  `transaction.authorizedAmount` will not take into account the amount from `AUTHORISATION_SUCCESS`.

### 6.4.2 AUTHORIZATION_FAILURE
The provided amount will be used to provide the history of transactionItem's actions.

> **Warning**
> In the case of finding `AUTHORISATION_SUCCESS` and `AUTHORISATION_FAILURE` with the same `pspReference`, the re-calculation for  `transaction.authorizedAmount` will not take into account the amount from `AUTHORISATION_SUCCESS`.

### 6.4.3 AUTHORIZATION_ADJUSTMENT
Assigns provided `amount` to `transaction.authorizedAmount`.

### 6.4.4 AUTHORIZATION_REQUEST
The event created by Saleor when staff user requests the authorization action.

### 6.4.5 CHARGE_SUCCESS
Add provided `amount` to `transaction.chargedAmount`. Any next call will add the `amount` to existing `transaction.chargedAmount` (`transaction.chargedAmount += amount`)

> **Warning**
> In the case of finding `CHARGE_SUCCESS` and `CHARGE_FAILURE` with the same `pspReference`, the re-calculation for  `transaction.chargedAmount` will not take into account the amount from `CHARGE_SUCCESS`.
  
### 6.4.6 CHARGE_FAILURE
The provided amount will be used to provide the history of transactionItem's actions.

> **Warning**
> In the case of finding `CHARGE_SUCCESS` and `CHARGE_FAILURE` with the same `pspReference`, the re-calculation for  `transaction.chargedAmount` will not take into account the amount from `CHARGE_SUCCESS`.

### 6.4.7 CHARGE_BACK
The provided `amount` will be used to reduce `transactionItem.chargedAmount` (`transactionItem.chargedAmount -= amount`).

### 6.4.8 CHARGE_REQUEST
The event created by Saleor when staff user requests the charge action.

### 6.4.9 REFUND_SUCCESS
Add provided `amount` to `transaction.refundedAmount`. Any next call will add the `amount` to existing `transaction.refundedAmount` (`transaction.refundedAmount += amount`)

> **Warning**
> In the case of finding `REFUND_SUCCESS` and `REFUND_FAILURE` with the same `pspReference`, the re-calculation for  `transaction.refundedAmount` will not take into account the amount from `REFUND_SUCCESS`.

### 6.4.10 REFUND_FAILURE
The provided amount will be used to provide the history of transactionItem's actions.

> **Warning**
> In the case of finding `REFUND_SUCCESS` and `REFUND_FAILURE` with the same `pspReference`, the re-calculation for  `transaction.refundedAmount` will not take into account the amount from `REFUND_SUCCESS`.

### 6.4.11 REFUND_REVERSE
The provided `amount` will be used to reduce `transactionItem.refundedAmount` (`transactionItem.refundedAmount -= amount`).

### 6.4.12 REFUND_REQUEST
The event created by Saleor when staff user requests the refund action.

### 6.4.13 CANCEL_SUCCESS
In case of providing `amount`, the `amount` will be used to reduce `transaction.authorizedAmount` (`transaction.authorizedAmount -= amount`).
In case of not providing `amount`, the `transaction.authorizedAmount` will be set to `0`.

> **Warning**
> In the case of finding `CANCEL_SUCCESS` and `CANCEL_FAILURE` with the same `pspReference`, the re-calculation for  `transaction.authorizedAmount` will not take into account the amount from `CANCEL_SUCCESS`.

### 6.4.14 CANCEL_FAILURE
The provided amount will be used to provide the history of transactionItem's actions.

> **Warning**
> In the case of finding `CANCEL_SUCCESS` and `CANCEL_FAILURE` with the same `pspReference`, the re-calculation for  `transaction.authorizedAmount` will not take into account the amount from `CANCEL_SUCCESS`.

### 6.4.15 CANCEL_REQUEST
The event created by Saleor when staff user requests the refund action.

# 7. TransactionCreate
The input (2.2.4) will get a new input `pspReference`. The `reference` field will be depreceated. The value of both fields is the same. We change the name of the field as we want to be more precise what kind of the reference should be placed here.

For each amount provided in mutation, the proper event will be created by Saleor

> **Warning**
>We will require providing pspReference or reference. When we will drop `reference` field, `pspReference` will become required as an input field.

# 8. TransactionUpdate
- The input (2.2.5) will get a new input `pspReference`. The `reference` field will be depreceated. The value of both fields is the same. We change the name of the field as we want to be more precise what kind of the reference should be placed here.

> **Warning**
>We will require providing pspReference or reference. When we will drop `reference` field, `pspReference` will become required as an input field.

> **Warning**
> We want to change the way how the App needs to handle the amounts. Previously the App had to reduce the previous value, like: calling Refund action, the update mutation should also reduce the amount of charge.
https://docs.saleor.io/docs/3.x/developer/checkout#updating-transaction
>
>With the current approach, there is no need to reduce a previous value. The app in case of action like refund should only update the refundedAmount. The chargedAmount will be the same as previously.

# 9. Return & Replace
The transaction actions are not related to return & replace feature. The one thing is that we need to drop the current refund logic embedded in return views. The plugins will work as previously, but the return/refund & replace (2.2.6, 2.2.7) views will not trigger any transaction action webhooks.

# 10. Migration Guide
Steps that need to be done when your implementation is using `transactionUpdate` mutation and/or `transaction-action-request` webhook.

## 10.1 transactionUpdate
`transactionEventReport` simplify the flow of reporting new changes related to transaction. The new mutation allows reporting any changes that have been done for the transaction. Based on the input Saleor will re-calculate the amounts. All your `transactionUpdate` calls should be replaced with `transactionEventReport`.  The below examples show calls of both mutations where the result will be the same

For example, the query:
```graphql
mutation {
  transactionUpdate(
    id: "VHJhbnNhY3Rpb25JdGVtOjE="
    transaction: {
      status: "Charged"
      availableActions: [REFUND]
      reference: "PSP-ref123.charge"
      amountAuthorized: { currency: "USD", amount: 0 }
      amountCharged: { currency: "USD", amount: 99 }
    }
    transactionEvent: {
      status: SUCCESS
      name: "Charged credit card"
      reference: "PSP-ref123.charge"
    }
  ) {
    transaction {
      id
    }
  }
}
```
should be converted to:
``` graphql
mutation {
  transactionEventReport(
    id: "VHJhbnNhY3Rpb25JdGVtOjE="
    pspReference: "PSP-ref123.charge"
    type: CHARGE_SUCCESS
    amount: 99
    message: "Charged credit card"
    availableActions: [REFUND]

  ){
    transaction {
      id
    }
  }
}
```

- `transactionUpdate.id` is the same as `transactionEventReport.id`
- `transactionUpdate.transactionEvent.reference` is `transactionEventReport.pspReference`
- `transactionUpdate.transaction.reference` is `transactionEventReport.pspReference`
- `transactionUpdate.transaction.amountAuthorized` App doesnt needed to make reduce calculation on its side, in `transactionEventReport`.
- `transactionUpdate.transaction.amountCharged` is `transactionEventReport.amount`
- `trasactionUpdate.transaction.availableActions` doesn't exist in `transactionEventReport`. Possible actions will be callculated by Saleor.
- `transactionUpdate.transactionEvent.status` is not needed. `transactionEventReport.type` covers all statuses
- `transactionUpdate.transactionEvent.name` is `transactionEventReport.message`

## 10.2 transaction-action-request
`TransactionChargeRequested` (5.1), `TransactionRefundRequested` (5.2), `TransactionCancelationRequested` (5.3) will have the same payload as `transaction-action-request`. The webhooks will require providing a `pspReference` in the returned payload (5.1.2, 5.2.2, 5.3.2).

The app implementation needs to be modify to provide at least a PSP reference in the response. The app on Saleor side needs to be subscribed to the webhook with events `TransactionChargeRequested` (5.1), `TransactionRefundRequested` (5.2), `TransactionCancelationRequested` (5.3), webhook subscribtion for the event: `transaction-action-request` should be removed from the App.

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


timuric:
`TransactionCreateInput` only accepts amountPendingRefund
```graphql
"""The input used in transactionCreate mutation"""
input TransactionCreateInput {
  ...
  """Amount that defines pending refund"""
  amountPendingRefund: MoneyInput

  """The url that will allow to redirect user to payment provider page with transaction details"""
  referenceURL: String
}
```

witoszekdev:
> The one thing that I am worried about with this approach is: that we allow to modify the transaction (transactionUpdate mutation), which means that someone different could modify the transaction's data. In that case I am not sure if we should track the last editor or only a creator of the transaction. Or maybe we should not allow to modify user's transactions?

I think we should allow editing user transactions. It's easy to image a situation where I messed up the transaction amount by one `0` and now I want to fix it. I think it's fine as long, as we store a log of transaction changes.

korycins:
> `TransactionCreateInput` only accepts amountPendingRefund?
> 
> ```graphql
> """The input used in transactionCreate mutation"""
> input TransactionCreateInput {
>   ...
>   """Amount that defines pending refund"""
>   amountPendingRefund: MoneyInput
> 
>   """The url that will allow to redirect user to payment provider page with transaction details"""
>   referenceURL: String
> }
> ```

@timuric  There are `...`.It just shows the new added fileds. I didn't add whole input to better show a new ones

korycins:
> > The one thing that I am worried about with this approach is: that we allow to modify the transaction (transactionUpdate mutation), which means that someone different could modify the transaction's data. In that case I am not sure if we should track the last editor or only a creator of the transaction. Or maybe we should not allow to modify user's transactions?
> 
> I think we should allow editing user transactions. It's easy to image a situation where I messed up the transaction amount by one `0` and now I want to fix it. I think it's fine as long, as we store a log of transaction changes.

@witoszekdev Yeah, that is clear, but the thing that I am worried about is: what should happen with the user attached to the transaction, when someone edit the transaction's data.
Example:
You create a manual capture transaction for 50USD, Saleor attach you to created transaction. Then I am taking transaction created by you, and modify the data like 50USD -> 5USD. Which staff user should be stored in transaction data? You as you are the creator of the transaction OR me as a last editor of the transaction?

fowczarek:
Maybe `amount` and `created` should be required? 
```graphql
type GrantedRefund {
  amount: Money!
  requester: App | User
  message: String
  created: DateTime!
}
```

fowczarek (approved):

patrys:
I suggest that we always use the word `total` for denormalized fields that are sums of other fields (like `grantedRefund`).

How would `settledRefund` work? What if the order is $100, I charge $60 and then grant a refund for $40? Do I need to still charge the $40 and immediately refund it for numbers to align?

witoszekdev:
> ```graphql
> """Represents an order in the shop."""
> type Order implements Node & ObjectWithMetadata {
> ...
>   """Amount that is not yet refunded: grantedRefund - totalRefunded - totalPendingRefund """
>   totalNonIssueRefund: Money!
> }
> ```

If we have amount of _non issued_ refunds we should also add _non granted_ refunds so that we know if for example a refund was requested inside Adyen / Mollie dashboard but not granted in Saleor. We'll need that on the frontend to display appropriate notification

witoszekdev:
> transaction.actions will be calculated by Saleor based on the amounts that are attached to the transaction

I'm to sure we should do that - some payment methods might not allow transactions to be voided or refunded. To give you and example take a look at [Adyen's payment methods list](https://docs.adyen.com/payment-methods#cash-and-atm-payment-methods) - none of "Cash & ATM payment methods" support refunds.

Inside `saleor-app-checkout` we set the actions based on response we get from payment provider. I guess we could have a fallback when app doesn't provide such information.

witoszekdev:
@korycins 

> You will always know if refund has been requested on payment provider side. New webhook will require pspReference as a response from the App. But we're still discussing about better name for totalNonIssueRefund field

I was talking here from Dashboard perspective, not app developer. We could theoretically calculate this data in dashboard, but since we won't be doing that for non issued refund I think it's a good idea to also add a field for not granted refunds.

korycins:
> I'm not sure if we should do that - some payment methods might not allow transactions to be voided or refunded. To give you an example take a look at [Adyen's payment methods list](https://docs.adyen.com/payment-methods#cash-and-atm-payment-methods) - none of "Cash & ATM payment methods" support refunds.

Hhm, right... I didn't think about this. Do you suggest that transactionEventReport should accept `availableActions` in that case?

Inside saleor-app-checkout we set the actions based on response we get from payment provider. I guess we could have a fallback when app doesn't provide such information.

witoszekdev:
@korycins yes, `transactionEventReport` should accept `availableActions`, but maybe it should be optional - if it's not used we can have a fallback on Saleor side.

korycins:
> I was talking here from Dashboard perspective, not app developer. We could theoretically calculate this data in dashboard, but since we won't be doing that for non issued refund I think it's a good idea to also add a field for not granted refunds.

I want to display this in dashboard when a refund wasn't granted in Saleor, but it was sent on payment provider side:

What about using `totalNonIssueRefund` value for that?
This value will be calculated based on:
`totalNonIssueRefund  = grantedRefund - totalRefunded - totalPendingRefund`
If the value is < 0, it means that you didn't created granted refund, right?

witoszekdev:
> `totalNonIssueRefund = grantedRefund - totalRefunded - totalPendingRefund`

@korycins right, I forgot that the values can be negative 😅

witoszekdev:
> Saleor will silently accept a call with the same details but no values will be updated for each next call. 

Can we add some clue that the call was ignored in the response from `transactionEventReport` mutation? I can imagine that it might be confusing to app developers why the mutation didn't update anything, because they just forgot they did that same mutation moments ago... _app developers = me 🙈_

witoszekdev:
> pspReference is required. In case of missing psp reference, the webhook will be treat as undelivered.

Does it mean that Saleor can make another request to app to retry the refund? Or will we create another transaction event that it request to do transaction action failed, and we won't do any retries?

witoszekdev:
Will this change affect `transactionCreate` webhook? Will I still be able to create a transaction + an event with it?

Another one: in the new dashboard views we will be able to create a "Manual refund" - that is a transaction with just `amountRefunded` set by staff member. Will it require a PSP reference to be set?

Example:
CleanShot 2022-09-08 at 16 15 59@2x

korycins:
@witoszekdev 
> >Saleor will silently accept a call with the same details but no values will be updated for each next call.

>Can we add some clue that the call was ignored in the response from transactionEventReport mutation? I can imagine that it might be confusing to app developers why the mutation didn't update anything, because they just forgot they did that same mutation moments ago... app developers = me 

Like:
```
type TransactionEventReport {
  transaction: TransactionItem
  duplicated: Boolean!
  errors: [TransactionRequestCompleteError!]!
}
```
?

korycins:
> > pspReference is required. In case of missing psp reference, the webhook will be treat as undelivered.
> 
> Does it mean that Saleor can make another request to app to retry the refund? Or will we create another transaction event that it request to do transaction action failed, and we won't do any retries?

Webhook by default will be retried a couple of times in case of receiving HTTP  error . 
By saying ` the webhook will be treat as undelivered.` I meant, that we will not update PSP reference for the event, so it will be missing for REQUESTED event.

But now after thinking more about this, I am not sure what should be our behaviour when we don't receive a PSP reference. Like:
we can retry the webhook, but we don't know what happen on App side, we don't know if the App already processed the refund and hit the timeout, or if the implementation is missing returning the PSP reference. In that case, App needs to know how to de-duplicate the webhook: we don't want to make the same refund twice. Maybe the app should store in the metadata the list of requested event that has been already processed?

korycins:
> Will this change affect `transactionCreate` webhook? Will I still be able to create a transaction + an event with it?
> 
> Another one: in the new dashboard views we will be able to create a "Manual refund" - that is a transaction with just `amountRefunded` set by staff member. Will it require a PSP reference to be set?
> 
> Example: CleanShot 2022-09-08 at 16 15 59@2x

You meant `transactionCreate` **mutation** right? For `transactionCreate` the one change that we will add is a giving staff user to call this mutation. No PSP reference will be required here, but availableActions will be set to `[]`

witoszekdev:
Please change type on `transactionEvent.amount` to `PositiveDecimal` like we have on the rest of the transaction ;)

korycins:
> Please change type on `transactionEvent.amount` to `PositiveDecimal` like we have on the rest of the transaction ;)

PositiveDecimal is a something that we need to discuss. We need to allow app to report Saleor any new event that happen. For example, If the refund amount that Saleor received is bigger than the total capture amount, we want to show capture as a < 0 value. We can't round it to 0, as we lose precision in that case. So probably all Postivie Decimal fields will need to be converted to standard Decimals.

korycins:
@witoszekdev  sorry, I forgot to mention you ☝️

witoszekdev:
@korycins 

## Missing total amounts

2.2.1 - I don't see the other pending amounts for transactions statuses (from `TransactionEventTypeEnum`):
- `totalPendingRefund` ✅ 
- `totalPendingCharge` ✅
- `totalPendingAuthorization` ❌  
- `totalPendingCancel` ❌ 
- `totalPendingChargeback` ❌ 

Same for `TransactionItem` (2.2.2):
- `pendingRefundAmount` ✅ 
- `pengingChargeAmount` ✅ 
- `pendingAuthorizeAmount` ✅ 
- `pendingCancelAmount` ❌ 
- `pendingChargebackAmount` ❌ 

## Manual transactions

I'm wondering what should we pass as `pspReference` when we're creating manual transactions or manual refunds (in Dashboard, by staff members).

For example this is how we create "Manual refunds" (a transaction with just the `amountRefunded` set by user):

```graphql
  mutation CreateManualTransactionRefund(
    $orderId: ID!
    $amount: PositiveDecimal!
    $currency: String!
    $description: String
  ) {
    transactionCreate(
      id: $orderId
      transaction: {
        type: "Manual refund"
        status: "Success"
        pspReference: $description
        amountRefunded: { amount: $amount, currency: $currency }
      }
      transactionEvent: { status: SUCCESS, type: REFUND, name: $description }
    ) {
      transaction {
        ...TransactionItem
      }
      errors {
        ...TransactionCreateError
      }
    }
  }
```

What should we pass as `pspReference`? Presumably this value is used for deduplication, but what if users use the same `pspReference` to describe a manual refund? (e.g. "cash"). Maybe this value should be optional for transactions created by users, and we would only use `message` field for describing the manual transactions? 

## State machine / invalid states

There are a few edge-cases that our proposed re-calculation (6.4) might not support:

### `CHARGE_FAILURE` → `CHARGE_SUCCESS` (retried capture)

There can be a situation when a charge initially failed, but after resolving an issue in Adyen can be retried again and can succeed

> E.g. Adyen sends [`CAPTURE` notification](https://docs.adyen.com/online-payments/capture#capture-notification) with `success: false` and reason. One of them can be [`Internal error`](https://docs.adyen.com/online-payments/capture#failed-capture-request) - an issue on Adyen side. On such case a capture can be retried again after an issue is resovled.
> 
> Another example could be [`CAPTURE_FAILED` webhook](https://docs.adyen.com/online-payments/capture/failure-reasons#page-introduction) it can also be send due to [technical issue](https://docs.adyen.com/online-payments/capture/failure-reasons#technical-issue). In such a case Adyen will try to fix the issue and re-capture the tnrasaction automatically. If it succeeds then [another `CAPTURE` webhook](https://docs.adyen.com/online-payments/capture/failure-reasons#what-you-need-to-do) with `success: true` is sent. This means in history we'll have 3 events:
> - `CAPTURE`
> - `CAPTURE_FAILED`
> - `CAPTURE`
> 
> The end result is that the transaction was captured successfully.

I think a capture should be allowed to be successful again, after a failed attempt (i.e. we shouldn't set the amount to 0)

### Cancelling refunds

We have now a `CENCEL_*` events - but this event IMHO is too generic. Looking at it, you can't tell what was cancelled (a refund? a capture? transaction was voided?).

I think we maybe should have `CANCEL` for each event type, for example:

For charge:
- `CHARGE_CANCEL_REQUEST`
- `CHARGE_CANCEL_SUCCESS`
- `CHARGE_CANCEL_FAILURE`

For refunds:
- `REFUND_CANCEL_REQUEST`
- `REFUND_CANCEL_SUCCESS`
- `REFUND_CANCEL_FAILURE`

> Here are notifications from Adyen with cancellations:
> - [`VOID_PENDING_REFUND`](https://docs.adyen.com/api-explorer/#/Webhooks/latest/post/VOID_PENDING_REFUND) - send after [a refund is cancelled](https://docs.adyen.com/point-of-sale/refund-payment/cancel-unreferenced)
> - [`CANCEL_OR_REFUND`](https://docs.adyen.com/api-explorer/#/Webhooks/latest/post/CANCEL_OR_REFUND)
> - [`CANCEL`](https://docs.adyen.com/api-explorer/#/Webhooks/latest/post/CANCEL)
> - [`TECHNICAL_CANCEL`](https://docs.adyen.com/online-payments/cancel#technical-cancel-webhook)

This would also mean that we should have more differentiation inside `TransactionActionEnum`:

- `CANCEL_CHARGE`
- `CANCEL_REFUND`
- `CANCEL_AUTHORIZATION` (aka void)

### Chargeback reversal

A chargeback can be reversed in Adyen ([`CHARGEBACK_REVERSED` notification](https://docs.adyen.com/api-explorer/#/Webhooks/latest/post/CHARGEBACK_REVERSED)). We should also handle this type of situation in `TransactionEventTypeEnum` by adding `CHARGE_BACK_REVERSED`.

A reversed chargeback can also be re-reversed ([`SECOND_CHARGEBACK` notification](https://docs.adyen.com/api-explorer/#/Webhooks/latest/post/SECOND_CHARGEBACK)) which means we should allow a successful chargeback after it was reversed (just like with refunds and charges)

> This is pure madness 😈

saleor/graphql/channel/mutations/channel_create.py
```diff
@@ -43,6 +43,17 @@ class OrderSettingsInput(graphene.InputObjectType):
...
     )
+    mark_as_paid_strategy = MarkAsPaidStrategyEnum(
+        required=False,
+        description=(
+            "Determine what strategy will be used to mark the order as paid. "
+            "Based on the choosen option the proper object will be created "
+            "and attached to the order, when order is manualy marked as paid."
+            "\n`PAYMENT_FLOW` - [default option] creates the `Payment` object."
+            "\n`TRANSACTION_FLOW` - creates the `TransactionItem` object."
+            + PREVIEW_FEATURE
```
maarcingebala, line 54:
```suggestion
            "Determine what strategy will be used to mark the order as paid. "
            "Based on the chosen option, the proper object will be created "
            "and attached to the order when it's manually marked as paid."
            "\n`PAYMENT_FLOW` - [default option] creates the `Payment` object."
            "\n`TRANSACTION_FLOW` - creates the `TransactionItem` object."
            + PREVIEW_FEATURE
```

saleor/graphql/channel/types.py
```diff
@@ -179,6 +179,18 @@ class OrderSettings(ObjectType):
         ),
     )
 
+    mark_as_paid_strategy = MarkAsPaidStrategyEnum(
+        required=True,
+        description=(
+            "Determine what strategy will be used to mark the order as paid. "
+            "Based on the choosen option the proper object will be created "
```
maarcingebala, line 186: Update this description accordingly as the first one (there were some typos).

saleor/plugins/webhook/tasks.py
```diff
@@ -481,25 +490,53 @@ def send_webhook_using_scheme_method(
     raise ValueError("Unknown webhook scheme: %r" % (parts.scheme,))
 
 
+def handle_webhook_retry(
```
maarcingebala, line 493: This could go to webhook/utils as well, no?

maarcingebala (approved):

saleor/graphql/channel/mutations/channel_create.py
korycins: Also missing ADDED_IN_ label

saleor/channel/__init__.py
```diff
@@ -14,3 +14,23 @@ class AllocationStrategy:
...
+    `Payment` object, that will cover the `order.total`.
+
+    """
+
+    TRANSACTION_FLOW = "transaction_flow"
+    PAYMENT_FLOW = "payment_flow"
+
+    CHOICES = [
+        (TRANSACTION_FLOW, "Use transaction"),
+        (PAYMENT_FLOW, "Use payment"),
```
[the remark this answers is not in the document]
korycins: @maarcingebala  Sorry for late response.
Payment flow is a flow when Saleor creates a Payment object. So it is the old one, but it is still not deprecated. The transaction API is still in the preview feature.

szczecha (approved):

saleor/plugins/webhook/tasks.py
korycins: @maarcingebala I didn't want to place this inside the utils, as inside `tasks.py` file we store everything related to sending the webhook and everything related to celery tasks. This function as an input accepts the celery task that should be re-tried. So personally I would prefer to keep it in `tasks.py` file. 
Let me know if you agree with this.

maarcingebala: Sure it's fine 👍

PR #10400, opened 2022-08-16
Simplify a registration of new subscription events

korycins:
I want to merge this change because we had multiple structure data, where we had to add a new event to provide a possibility for subscribing it.  Now, we have only one const, where we need to declare a pair: `event:type`. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/plugins/webhook/tasks.py
```diff
@@ -81,7 +78,7 @@ def create_deliveries_for_subscriptions(
     :param requestor: used in subscription webhooks to generate meta data for payload.
     :return: List of event deliveries to send via webhook tasks.
     """
-    if event_type not in SUBSCRIBABLE_EVENTS:
+    if event_type not in list(WEBHOOK_TYPES_MAP.keys()):
```
IKarbowiak, line 81: You can simplify this:
```suggestion
    if event_type not in WEBHOOK_TYPES_MAP:
```

saleor/plugins/webhook/tasks.py
```diff
@@ -134,7 +131,7 @@ def create_delivery_for_subscription_sync_event(
...
```
IKarbowiak, line 134: Same here

IKarbowiak (approved):

saleor/plugins/webhook/tasks.py
korycins: Right 🤦

fowczarek (approved):

maarcingebala (approved):

PR #10445, opened 2022-08-21
Add ability to set tax config in Avalara plugin

mociepka:
I want to merge this change because I cannot set the "All products prices are entered with tax included." tax setting per channel.

This change extends Avalara plugin configuration with the ability to override the "All products prices are entered with tax included." per channel.

image

# Impact

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fix situation when Payment Gateway try to save to long error message - #10402 by @fowczarek
 - Replaced `context.app` lazy object with a dataloader.
 - Add support for bcrypt password hashes - #10346 by @pkucmus
-
+- Add ability to set taxes configuration per channel in the Avatax plugin - #10445 by @mociepka
 
 # 3.6.0
 
--- a/saleor/plugins/avatax/plugin.py
+++ b/saleor/plugins/avatax/plugin.py
@@ -71,6 +71,8 @@ class AvataxPlugin(BasePlugin):
         {"name": "from_country_area", "value": None},
         {"name": "from_postal_code", "value": None},
         {"name": "shipping_tax_code", "value": "FR000000"},
+        {"name": "override_global_tax", "value": False},
+        {"name": "include_taxes_in_prices", "value": True},
     ]
     CONFIG_STRUCTURE = {
         "Username or account": {
@@ -136,6 +138,16 @@ class AvataxPlugin(BasePlugin):
             ),
             "label": "Shipping tax code",
         },
+        "override_global_tax": {
+            "type": ConfigurationTypeField.BOOLEAN,
+            "help_text": "Used when setting per channel is needed.",
+            "label": "Override global tax settings.",
+        },
+        "include_taxes_in_prices": {
+            "type": ConfigurationTypeField.BOOLEAN,
+            "help_text": 'Applied only if "Override global tax settings" is on.',
+            "label": "All products prices are entered with tax included.",
+        },
     }
 
     def __init__(self, *args, **kwargs):
@@ -198,10 +210,6 @@ class AvataxPlugin(BasePlugin):
         if not response or "error" in response:
             return checkout_total
 
-        tax_included = (
-            lambda: Site.objects.get_current().settings.include_taxes_in_prices
-        )
-
         currency = checkout_info.checkout.currency
         taxed_total = zero_taxed_money(currency)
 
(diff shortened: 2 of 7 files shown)


saleor/plugins/avatax/plugin.py
```diff
@@ -136,6 +138,16 @@ class AvataxPlugin(BasePlugin):
             ),
             "label": "Shipping tax code",
         },
+        "override_global_tax": {
+            "type": ConfigurationTypeField.BOOLEAN,
+            "help_text": "Used when setting per channel is needed.",
```
korycins, line 143: Maybe something that will tell us that it overwrites global settings for include taxes in prices. `Overwride global tax` can be a little bit too generic.

saleor/plugins/avatax/plugin.py
```diff
@@ -573,6 +572,17 @@ def calculate_order_line_unit(
             / quantity,
         )
 
+    @property
+    def _tax_included(self) -> Callable[[], bool]:
+        configuration = {item["name"]: item["value"] for item in self.configuration}
```
korycins, line 577: Do you need this? Can't you set this value in `__init__` method? This is something that will be used in all calculations method.

maarcingebala (approved):
Looks good but it needs to things:
- changelog
- tests are currently failing

saleor/plugins/avatax/plugin.py
maarcingebala: Yep I had the same feeling. At least we could give an example which options would be overridden.

patrys: Can we make it a radio group or a drop-down with three options:

```
Prices:
   (o) Respect Saleor's configuration (default)
   ( ) Treat all prices as net amounts (excl. taxes)
   ( ) Treat all prices as gross amounts (inc. taxes)
```

mociepka:
@maarcingebala I missed one file in commit. I just pushed it. @maarcingebala, what and where do I need to do with the changelog?

saleor/plugins/avatax/plugin.py
mociepka: @patrys sure, you described `label` what I should put in the `help_text`?

saleor/plugins/avatax/plugin.py
mociepka: I took this approach from the [__init__](https://github.com/saleor/saleor/blob/main/saleor/plugins/avatax/plugin.py#L144). The `self.configuration` already exists and has a different structure.

What do you expect @korycins ? Another attribute like `self._configuration`? I rather prefer a temp structure.

maarcingebala:
@patrys Are we going to release these changes as 3.7? I would say so as this is new feature. If yes, then the changelog should be added to [the 3.7.0 [Unreleased] section ](https://github.com/saleor/saleor/blob/main/CHANGELOG.md#370-unreleased).

saleor/plugins/avatax/plugin.py
korycins: @mociepka Personally, I would consider adding new fields to `self.configuration` (can be with `_` also). 
And maybe adding a function to AvataxConfiguration dataclass `tax_included`, that will return you info, if the tax should be included in prices or not - especially since now, Avatax plugin has its own configuration to change this approach. 
Then you don't need to have the parameter `tax_included` in functions/methods.

mociepka: @korycins pushed new changes.

patrys:
@maarcingebala This can't wait for taxes so either 3.6 or we release 3.7 now and move taxes to 3.8.

mociepka:
@maarcingebala @patrys I extended the 3.6 change log with information about this change.

saleor/plugins/avatax/plugin.py
mociepka: I started to set this up, and I recall there is no way of getting choices input (radio or select) in the plugin settings. I discussed this with @korycins, and we agreed that two checkboxes solution is good enough.

@patrys, do you still want to change labels with this additional context? If yes, please provide me with labels that suites you.

korycins (approved):

timuric:
@mociepka could you explain what "Override global tax settings" means. 
Does it override all tax settings?
image

CHANGELOG.md
```diff
@@ -51,6 +52,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fix triggering `ORDER_CANCELED` event at the end of transaction (#10242) (d9eecb2ca)
 - Fix post-migrate called for each app module (#10252) (60205eb56)
 - Only handle known URLs (disable appending slash to URLs automatically) - #10290 by @patrys
+- Introduce tax configuration base on channel in the Avatax plugin - #10445 by mociepka
```
maarcingebala, line 55: @mociepka Please just move this entry to the "3.7.0 Unreleased" section, as we're going to release it as 3.7.

mociepka:
@timuric override setting listed below. Only one is listed:
image

image

I can change wording no problem with that, but please provide labels and tool tips.

CHANGELOG.md
mociepka: Done.

PR #10469, opened 2022-08-24
Update JSONWebTokenBackend (auth) to use dataloader

dekoza:
This is the next step on the way to get rid of lazy objects in order to introduce async to saleor. This time I started working on removing `request.user` and `context.requestor`. It turned out the authentication flow requires much bigger refactor than anticipated so I decided to do it in small steps. This PR makes the authentication backend `JSONWebTokenBackend` use dataloader. There will be duplicated code at the moment (`load_user_from_request` function) but it will get cleaned gradually in upcoming steps.

You will see some methods get an extra `request=None` param - this is a teaser for upcoming refactoring :wink: 

# Impact

* Internal change, no real effect other than using dataloader instead of lazy object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add support for bcrypt password hashes - #10346 by @pkucmus
 - Add ability to set taxes configuration per channel in the Avatax plugin - #10445 by @mociepka
 
+### GraphQL API
+ - Add `name` parameter to `ProductVariantInput` - #10456 by @SzymJ
 
 # 3.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add support for bcrypt password hashes - #10346 by @pkucmus
 - Add ability to set taxes configuration per channel in the Avatax plugin - #10445 by @mociepka
 
+
 # 3.6.0
 
 ### Breaking changes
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add support for bcrypt password hashes - #10346 by @pkucmus
 - Add ability to set taxes configuration per channel in the Avatax plugin - #10445 by @mociepka
 
+
```
IKarbowiak, line 14: 👀

IKarbowiak (approved):

CHANGELOG.md
dekoza: Yeah, don't know where it came from :wink: Anyway I was told not to put "invisible" internal changes to changelog.

saleor/graphql/account/dataloaders.py
```diff
@@ -49,3 +64,64 @@ def batch_load(self, keys):
...
+    def batch_load(self, keys):
+        permission_map = (
+            Permission.objects.filter(codename__in=keys)
+            .prefetch_related("content_type")
+            .in_bulk(field_name="codename")
+        )
+        return [permission_map.get(codename) for codename in keys]
+
+
+def load_user_from_request(request):
```
korycins, line 93: Not sure if this is proper place.  It looks like something more related to auth layer, and processing a JWT token than `dataloaders`. I would say that a better place for this is `saleor/core/auth_backend.py` especially, that it is used only there.

dekoza: Good point, thank you. Fixed.

korycins (approved):

szczecha (approved):

PR #10481, opened 2022-08-25
Allow to pass metadata directly in checkoutComplete and OrderCreateFromCheckout mutations

SzymJ:
I want to merge this change because it adds possibility to update `metadata` directly in `checkoutComplete` and `OrderCreateFromCheckout` mutations.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/complete_checkout.py
+++ b/saleor/checkout/complete_checkout.py
@@ -469,9 +469,9 @@ def _create_order(
     user: User,
     app: Optional["App"],
     manager: "PluginsManager",
-    site_settings=None,
-    metadata_list=None,
-    private_metadata_list=None
+    site_settings: Optional["SiteSettings"] = None,
+    metadata_list: Optional[List] = None,
+    private_metadata_list: Optional[List] = None
 ) -> Order:
     """Create an order from the checkout.
 
@@ -746,8 +746,8 @@ def complete_checkout(
     site_settings=None,
     tracking_code=None,
     redirect_url=None,
-    metadata_list=None,
-    private_metadata_list=None,
+    metadata_list: Optional[List] = None,
+    private_metadata_list: Optional[List] = None,
 ) -> Tuple[Optional[Order], bool, dict]:
     """Logic required to finalize the checkout and convert it to order.
 
--- a/saleor/checkout/complete_checkout.py
+++ b/saleor/checkout/complete_checkout.py
@@ -469,7 +469,9 @@ def _create_order(
     user: User,
     app: Optional["App"],
     manager: "PluginsManager",
-    site_settings=None
+    site_settings=None,
+    metadata_list=None,
+    private_metadata_list=None
 ) -> Order:
     """Create an order from the checkout.
 
@@ -565,8 +567,17 @@ def _create_order(
 
     # copy metadata from the checkout into the new order
     order.metadata = checkout.metadata
+    if metadata_list:
+        order.store_value_in_metadata({data.key: data.value for data in metadata_list})
+
     order.redirect_url = checkout.redirect_url
+
     order.private_metadata = checkout.private_metadata
+    if private_metadata_list:
+        order.store_value_in_private_metadata(
+            {data.key: data.value for data in private_metadata_list}
+        )
+
     update_order_charge_data(order, with_save=False)
     update_order_authorize_data(order, with_save=False)
     order.search_vector = FlatConcatSearchVector(
@@ -735,6 +746,8 @@ def complete_checkout(
     site_settings=None,
     tracking_code=None,
     redirect_url=None,
+    metadata_list=None,
+    private_metadata_list=None,
 ) -> Tuple[Optional[Order], bool, dict]:
     """Logic required to finalize the checkout and convert it to order.
 
@@ -807,6 +820,8 @@ def complete_checkout(
                 app=app,
                 manager=manager,
                 site_settings=site_settings,
+                metadata_list=metadata_list,
+                private_metadata_list=private_metadata_list,
             )
             # remove checkout after order is successfully created
             checkout.delete()
@@ -1003,6 +1018,8 @@ def _create_order_from_checkout(
     user: User,
     app: Optional["App"],
     tracking_code: Optional[str] = None,
+    metadata_list: Optional[List] = None,
+    private_metadata_list: Optional[List] = None,
 ):
     from ..order.utils import add_gift_cards_to_order
 
@@ -1040,6 +1057,16 @@ def _create_order_from_checkout(
         if site_settings.automatically_confirm_all_new_orders
         else OrderStatus.UNCONFIRMED
     )
+
+    # update metadata
+    if metadata_list:
+        checkout_info.checkout.store_value_in_metadata(
+            {data.key: data.value for data in metadata_list}
+        )
+    if private_metadata_list:
+        checkout_info.checkout.store_value_in_private_metadata(
+            {data.key: data.value for data in private_metadata_list}
+        )
 
     # order
     order = Order.objects.create(
@@ -1128,6 +1155,8 @@ def create_order_from_checkout(
     app: Optional["App"],
     tracking_code: str,
     delete_checkout: bool = True,
+    metadata_list: Optional[List] = None,
+    private_metadata_list: Optional[List] = None,
 ) -> Order:
     """Crate order from checkout.
 
@@ -1161,6 +1190,8 @@ def create_order_from_checkout(
                 user=user,
                 app=app,
                 tracking_code=tracking_code,
+                metadata_list=metadata_list,
+                private_metadata_list=private_metadata_list,
             )
             if delete_checkout:
                 checkout_info.checkout.delete()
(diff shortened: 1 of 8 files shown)


patrys:
Task linked: [SALEOR-7950 Allow metadata to be passed to checkoutComplete and orderFromCheckoutId](https://app.clickup.com/t/2549495/SALEOR-7950)

saleor/graphql/checkout/tests/mutations/test_checkout_complete.py
```diff
@@ -289,6 +299,128 @@ def test_checkout_complete(
     order_confirmed_mock.assert_called_once_with(order)
```
fowczarek, line 301: Could you add a test for situations when we have metadata stored in checkout and we perform `CheckoutComplete` with additional metadata(But updating any key stored in the original checkout object)? 

example date:
Checkout_meta - {"a": 1, "b":2}
Meta passed to mutation -  {"a":4, "c":3}
Result: {"a":4, "b":2, "c":3}

fowczarek:
`checkoutComplete` shouldn't be able to update private metadata. This mutation is used to finish the own checkout process. Customers are not able to see checkout private metadata.

saleor/graphql/checkout/mutations/checkout_complete.py
```diff
@@ -87,6 +87,11 @@ class Arguments:
                 "Client-side generated data required to finalize the payment."
             ),
         )
+        metadata = NonNullList(
+            MetadataInput,
+            description="Fields required to update the object's metadata.",
```
korycins, line 92: Missing `ADDED_IN_X`. What about `PREVIEW_FEATURE`? Do we want to mark this field with `PREVIEW_FEATURE` also?

saleor/graphql/schema.graphql
```diff
@@ -13578,6 +13581,12 @@ type Mutation {
     """ID of a checkout that will be converted to an order."""
     id: ID!
 
+    """Fields required to update the object's metadata."""
+    metadata: [MetadataInput!]
+
+    """Fields required to update the object's private metadata."""
+    privateMetadata: [MetadataInput!]
```
korycins, line 13588: Same as for `checkoutCoplete`. Missing `ADDED_IN_X`. `PREVIEW_FEATURE` also?

saleor/checkout/complete_checkout.py
```diff
@@ -465,7 +465,9 @@ def _create_order(
     user: User,
     app: Optional["App"],
     manager: "PluginsManager",
-    site_settings=None
+    site_settings=None,
+    metadata_list=None,
+    private_metadata_list=None
```
maarcingebala, line 470: We could add typings for these.

saleor/graphql/checkout/mutations/checkout_complete.py
maarcingebala: I wouldn't say so, it's not entirely a new feature, but rather an improvement to the existing one, what do you think @timuric ?

saleor/graphql/meta/mutations.py
```diff
@@ -400,3 +402,98 @@ def perform_mutation(cls, _root, info, **data):
                 instance.delete_value_from_private_metadata(key)
             _save_instance(instance, "private_metadata")
         return cls.success_response(instance)
+
+
+class BaseMutationWithMetadata(BaseMutation):
```
maarcingebala, line 407: `BaseMetadataMutation` vs `BaseMutationWithMetadata` - sounds very similar, will be hard to distinguish. Don't have a better name yet but we could discuss.

fowczarek (approved):

saleor/graphql/checkout/mutations/checkout_complete.py
timuric: @maarcingebala @korycins _preview_  is used when we expect API to change in minor versions, unless you anticipate this to happen it is not necessary

korycins: Okay, so only ADDED_IN is missing here. cc @SzymJ Let's add the missing labels and move forward with this PR.

fowczarek (approved):

maarcingebala (approved):

saleor/graphql/meta/mutations.py
```diff
@@ -399,3 +402,102 @@ def perform_mutation(cls, _root, info, **data):
...
+                f"Make sure that type exists inside {map_name}."
+            )
+
+        if not cls.check_permissions(info.context, meta_permission(info, object_id)):
+            metadata_type = "private metadata" if private else "metadata"
+            message = f"You do not have permission to update {metadata_type}."
+            raise PermissionDenied(message)
+
+    @classmethod
+    def mutate(cls, root, info, **data):
```
korycins, line 473: I would recommend keeping this mutation as tight as possible. We should place here only a custom logic related to metadata permission, but we also should run `result = super().mutate(root, info, **data)`. Thanks to this, any updates that we will add to baseMutation, will apply to this mutation also. Without this, we need to put additional attention to maintaining this part of the code and keeping it up to date.

saleor/graphql/checkout/mutations/order_create_from_checkout.py
```diff
@@ -62,6 +62,7 @@ class Meta:
         object_type = Order
         permissions = (CheckoutPermissions.HANDLE_CHECKOUTS,)
         error_type_class = OrderCreateFromCheckoutError
+        metadata_permissions_map_key = "Checkout"
```
korycins, line 65: Do you need this? Can't we determine a type name in different way? Like using `id` field that is `global` and contain a `typename`?

saleor/graphql/meta/mutations.py
```diff
@@ -399,3 +402,102 @@ def perform_mutation(cls, _root, info, **data):
...
+    def check_metadata_permissions(cls, type_name, info, object_id, private=False):
+        if private:
+            meta_permission = PRIVATE_META_PERMISSION_MAP.get(type_name)
+            map_name = "PRIVATE_META_PERMISSION_MAP"
+        else:
+            meta_permission = PUBLIC_META_PERMISSION_MAP.get(type_name)
+            map_name = "PUBLIC_META_PERMISSION_MAP"
+
+        if not meta_permission:
+            raise NotImplementedError(
```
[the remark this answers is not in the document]
SzymJ: @korycins  I copied this logic from meta mutations, but I will change it

saleor/graphql/meta/mutations.py
```diff
@@ -399,3 +402,102 @@ def perform_mutation(cls, _root, info, **data):
...
             _save_instance(instance, "private_metadata")
         return cls.success_response(instance)
+
+
+class BaseMutationWithMetadata(BaseMutation):
+    class Arguments:
+        private_metadata = NonNullList(
+            MetadataInput,
+            description=(
+                "Fields required to update the object's private metadata." + ADDED_IN_38
```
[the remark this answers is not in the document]
SzymJ: @korycins so what do you think? should I remove those `Arguments` from `BaseMutationWithMetadata` and then we have to add these arguments to every mutation with labels?

saleor/graphql/meta/mutations.py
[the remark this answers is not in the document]
korycins: Yeah, I saw that the same issue exist there, but I think that we could fix it here at least.

saleor/graphql/meta/mutations.py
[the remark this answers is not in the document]
korycins: I would say, that adding them explicitly would be "safer" from API perspective and proper field labels. We already had(or still have) the issue with subscription types, where base type had label "ADDED_IN_3X" and all child types had generated schema with this info (which wasn't a true in all cases).
@maarcingebala  What do you think?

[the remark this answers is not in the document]
maarcingebala: Maciek is right, we cannot add it to a generic class, as this information wouldn't be true for any mutations changed in other versions.

saleor/graphql/checkout/mutations/checkout_complete.py
```diff
@@ -88,6 +88,13 @@ class Arguments:
                 "Client-side generated data required to finalize the payment."
             ),
         )
+        metadata = NonNullList(
+            MetadataInput,
+            description=(
+                "Fields required to update the object's metadata." + ADDED_IN_38
```
korycins, line 94: `object` -> `checkout`?

saleor/graphql/checkout/mutations/order_create_from_checkout.py
```diff
@@ -50,6 +50,20 @@ class Arguments:
             ),
             default_value=True,
         )
+        private_metadata = NonNullList(
+            MetadataInput,
+            description=(
+                "Fields required to update the object's private metadata." + ADDED_IN_38
```
korycins, line 56: `object`->`checkout`?

saleor/graphql/checkout/mutations/order_create_from_checkout.py
```diff
@@ -50,6 +50,20 @@ class Arguments:
...
+            ),
+            required=False,
+        )
+        metadata = NonNullList(
+            MetadataInput,
+            description=(
+                "Fields required to update the object's metadata." + ADDED_IN_38
```
korycins, line 63: `object`->`checkout`?

korycins (approved):

PR #10493, opened 2022-08-26
Dataloader for Site

dekoza:
I want to merge this change because we are removing all lazy objects from code.

# Impact
Changes only how internals work, no changes to any public interfaces.


saleor/site/models.py
```diff
@@ -159,3 +164,56 @@ def get_translated_keys(self):
...
+            "set the SITE_ID setting. Create a site in your database and "
+            "set the SITE_ID setting."
+        )
+    else:
+        # Populate the other loader for free
+        SiteByIdLoader(request).prime(site.id, site)
+    return site
+
+
+def load_site(request, site_id=None):
```
korycins, line 216: Do we need `site_id` here? AFAIR we always fetch site model by default `env.SITE_ID`. if not needed then we could simplify the loader, to only use `load_current_site`.

saleor/site/models.py
```diff
@@ -159,3 +164,56 @@ def get_translated_keys(self):
             "header_text": self.header_text,
             "description": self.description,
         }
+
+
+class SiteByIdLoader(DataLoader):
```
korycins, line 169: Wouldn't be better to migrate loaders to `saleor/site/dataloaders.py`? This will keep consistency with the way how we store current data loaders.
Helper methods could be placed in a separate file or (IMHO) it could be also placed in the file with dataloaders.

saleor/graphql/order/mutations/draft_order_create.py
```diff
@@ -279,13 +281,14 @@ def _save_addresses(info, instance: models.Order, cleaned_input):
     @staticmethod
     def _save_lines(info, instance, lines_data):
         if lines_data:
+            site = load_site(info.context)
```
IKarbowiak, line 284: We are fetching the sites two times in this mutation, IMO, we should do it once in `perform_mutation` and pass to all methods that require `site` object.

saleor/graphql/order/mutations/order_fulfill.py
```diff
@@ -177,10 +178,10 @@ def check_total_quantity_of_items(cls, quantities_for_lines):
 
     @classmethod
     def clean_input(cls, info, order, data):
-        site_settings = info.context.site.settings
+        site = load_site(info.context)
```
IKarbowiak, line 181: Same here. We should pass the `site` as an argument, we already have the `site` object in perform_mutation.

saleor/site/models.py
maarcingebala: Yep, data loaders are GraphQL specific, so we should put them somewhere under `saleor/graphql`, model files are not the correct layer.

saleor/warehouse/reservations.py
```diff
@@ -381,9 +382,10 @@ def is_reservation_enabled(settings) -> bool:
 
 
 def get_reservation_length(request) -> Optional[int]:
+    site = load_site(request)
```
maarcingebala, line 385: Similar to my previous comment, I think we should instead pass the site and whatever else is needed here as parameters instead of passing the entire request. Then we could keep `load_*` functions in the GraphQL layer.

saleor/graphql/order/mutations/draft_order_create.py
maarcingebala: I think it calls the DB only once, as there is a data loader underneath. Then we should be able to call `load_*` as many times as we want within a single request without worrying about DB. This would look a bit better than passing the site to other mutation methods. But if the queries are constant, that would be the preferred solution.

dekoza: Both of you are right :smile: Other methods should get `site` as param as this is more testable. And dataloader ensures that the DB is only hit once - the answer is cached for future use.

saleor/graphql/site/dataloaders.py
```diff
@@ -0,0 +1,62 @@
+from functools import reduce
+
+from django.contrib.sites.models import Site
+from django.contrib.sites.requests import RequestSite
+from django.core.exceptions import ImproperlyConfigured
+from django.db.models import Q
+from django.http.request import split_domain_port
+
+from saleor.graphql.core.dataloaders import DataLoader
```
korycins, line 9: Should be relative import

saleor/graphql/site/dataloaders.py
```diff
@@ -0,0 +1,62 @@
...
+        # simulate nonexisting `domain__iexact__in`
+        q_list = map(lambda k: Q(domain__iexact=k), keys)
+        q_list = reduce(lambda a, b: a | b, q_list)
+        sites = Site.objects.using(self.database_connection_name).filter(q_list)
+        sites_mapped = {s.domain.lower(): s for s in sites}
+        return [sites_mapped.get(host.lower()) for host in keys]
+
+
+def load_current_site(request):
+    from django.conf import settings
```
korycins, line 34: Why is it required to import this locally?

saleor/graphql/site/dataloaders.py
```diff
@@ -0,0 +1,62 @@
...
+        domain, port = split_domain_port(host)
+        site = SiteByHostLoader(request).load(domain).get()
+
+    if site is None:
+        raise ImproperlyConfigured(
+            'You\'re using the Django "sites framework" without having '
+            "set the SITE_ID setting. Create a site in your database and "
+            "set the SITE_ID setting."
+        )
+    else:
```
korycins, line 53: `else` is not needed. You raise an exception in main `if` block

saleor/graphql/order/mutations/order_fulfill.py
dekoza: Does not matter. Loader makes only one db hit per request and uses cache for consecutive loads.

saleor/site/models.py
korycins: @dekoza What about this comment? I don't see any usage of `load_site(info.context, site_id=<something>`. In that case, why do we need to have this?

saleor/graphql/order/mutations/order_fulfill.py
IKarbowiak: Yep, but we still need to get this data from the cache. I agree that when the method has plenty of attributes already or we fetched the object like 4 steps above, it's not worth it. But if we have the object in the method that is calling the `clean_input` IMO, it's worth providing it as a method argument.

saleor/site/models.py
dekoza: I removed it

saleor/graphql/order/mutations/order_fulfill.py
dekoza: Fixed

korycins (approved):

IKarbowiak (approved):

maarcingebala (approved):

maarcingebala:
@dekoza Please resolve conflicts and this PR will be ready to merge.

PR #10512, opened 2022-08-29
Use dataloader for discounts instead of lazy object

dekoza:
Internal change removing use of lazy object in favor of dataloader - this time for discounts.

One extra effect - in two tests the query count dropped by 1 :smile:

### DIFF
--- a/saleor/graphql/discount/dataloaders.py
+++ b/saleor/graphql/discount/dataloaders.py
@@ -229,5 +229,5 @@ class OrderDiscountsByOrderIDLoader(DataLoader):
         return [discount_map.get(order_id, []) for order_id in keys]
 
 
-def load_request_discounts(request):
+def load_discounts(request):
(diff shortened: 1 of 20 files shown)


korycins (approved):

saleor/graphql/discount/dataloaders.py
```diff
@@ -227,3 +227,7 @@ def batch_load(self, keys):
         for discount in discounts:
             discount_map[discount.order_id].append(discount)
         return [discount_map.get(order_id, []) for order_id in keys]
+
+
+def load_request_discounts(request):
```
maarcingebala, line 232: Could be just `load_discounts` I guess? Seems consistent with other "load_*" functions.

IKarbowiak (approved):

fowczarek (approved):

maarcingebala (approved):

PR #10513, opened 2022-08-29
Add custom app installation error message

przlada:
I want to merge this change because it adds custom app installation error message to `AppInstallation`

Zrzut ekranu 2022-08-29 o 21 41 14

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/app/installation_utils.py
+++ b/saleor/app/installation_utils.py
@@ -1,6 +1,7 @@
 import requests
 from django.conf import settings
 from django.contrib.sites.models import Site
+from requests import HTTPError, Response
 
 from ..core.permissions import get_permission_names
 from ..plugins.manager import PluginsManager
@@ -10,6 +11,21 @@ from .models import App, AppExtension, AppInstallation
 from .types import AppExtensionTarget, AppType
 
 REQUEST_TIMEOUT = 25
+
+
+class AppInstallationError(HTTPError):
+    pass
+
+
+def validate_app_install_response(response: Response):
+    try:
+        response.raise_for_status()
+    except HTTPError as err:
+        try:
+            error_msg = str(response.json()["error"]["message"])
+        except Exception:
+            raise err
+        raise AppInstallationError(error_msg, response=response)
 
 
 def send_app_token(target_url: str, token: str):
@@ -24,7 +40,7 @@ def send_app_token(target_url: str, token: str):
     response = requests.post(
         target_url, json=json_data, headers=headers, timeout=REQUEST_TIMEOUT
     )
-    response.raise_for_status()
+    validate_app_install_response(response)
 
 
 def install_app(app_installation: AppInstallation, activate: bool = False):
--- a/saleor/app/tasks.py
+++ b/saleor/app/tasks.py
@@ -5,7 +5,7 @@ from requests import HTTPError, RequestException
 
 from .. import celeryconf
 from ..core import JobStatus
-from .installation_utils import install_app
+from .installation_utils import AppInstallationError, install_app
 from .models import AppInstallation
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


saleor/app/installation_utils.py
```diff
@@ -12,6 +13,21 @@
...
+class AppInstallationError(HTTPError):
+    pass
+
+
+def validate_app_install_response(response: Response):
+    try:
+        response.raise_for_status()
+    except HTTPError as err:
+        try:
+            error_msg = str(response.json()["error"]["message"])
```
krzysztofwolski, line 25: If the server did not return the message, maybe we can fall back to a generic one.

For example `Server responded with status ${response.status_code}`?

But that's just a suggestion, besides that LGTM.

krzysztofwolski (approved):

saleor/app/installation_utils.py
przlada: Standard `HTTPError` will be raised if the server will not return an error message. Then [here](https://github.com/saleor/saleor/blob/1e9499fef5cd3077c2dfdcb7796a6525c280e493/saleor/app/tasks.py#L27-L31), this error will be handled, and default message will be set:

> Failed to connect to app. Try later or contact with app support.

When we introduce error context, we could add there response `status_code`.

korycins, line 25: Shouldn't you use `.get` here? This can fail if these field will not be returned.

korycins: you expect that the error response from the app will be in defined shape: json: `{"error": {"message": "err_msg"}}`. This is something that should be added to the docs as the app owner need to know the proper response shape.

przlada: > Shouldn't you use .get here? This can fail if these field will not be returned.

There is the intention for this line to fail if the error response has improper structure or does not contain an error message at all. In such a case, it will fall back to default HTTPError and then default error message. Any idea for a cleaner way to do that?

> you expect that the error response from the app will be in defined shape: json: {"error": {"message": "err_msg"}}. This is something that should be added to the docs as the app owner need to know the proper response shape.

Yes, I will add it to the documentation when it is accepted. Extending error information with context like error code, source, response status_code, etc., is also being discussed. I will probably document it in bulk.

korycins: > There is the intention for this line to fail if the error response has improper structure or does not contain an error message at all. In such a case, it will fall back to default HTTPError and then default error message. 

Right, I missed `try/except`, shouldn't do review before first coffee 😅 

> Any idea for a cleaner way to do that?

When building an app with a local Saleor instance, I would be happy to get here warning info like "response structure is incorrect: missing error field", something that would give me a more context of what is wrong or what is missing.

saleor/app/tasks.py
```diff
@@ -21,13 +21,25 @@ def install_app_task(job_id, activate=False):
...
+        )
+    except RequestException as e:
+        logger.warning("Failed to install app. Error: %s", e)
         app_installation.message = (
             "Failed to connect to app. Try later or contact with app support."
         )
     except Exception as e:
-        logger.warning("Failed to install app. error %s", e)
+        logger.warning("Failed to install app. Error: %s", e)
         app_installation.message = "Unknown error. Contact with app support."
```
przlada, line 43: > When building an app with a local Saleor instance, I would be happy to get here warning info like "response structure is incorrect: missing error field", something that would give me a more context of what is wrong or what is missing.

@korycins I refactored how installation errors are logged and added the "missing error field" hint.

krzysztofwolski: IMO *depending* on the error message in the response is not the way to go. It's not part of the HTTP spec ;). 
If installation failed because of some infrastructure _before_ the app (tunnel, proxy, etc), the message is not controllable by the app developer. In case of critical failures (500), handling them can also be out of dev control.

My suggestion is to:
- if the message in the response is present, use it
- if not, add a message with a response code

Later we can refactor it to put the response code in the context, but this solution will make debugging much more straightforward.

przlada: I agree. I can change the error message to something like this:
```
App internal error (500). Try later or contact with app support.
```
Later we could simplify these messages and move error details to a dedicated place. Is that ok Krzych?

krzysztofwolski: Yup! I think that will make debugging easier until we settle down with the more complex `context`.

korycins (approved):

IKarbowiak (approved):

PR #10524, opened 2022-08-31
Add database router to fix false positives on relation mismatch.

### LINKED ISSUE #10467: giftCardCreate mutation tries to create GiftCard and GiftCardTag on different databases
[Error from Sentry](https://sentry.io/organizations/saleor/issues/3526888361/):
Cannot add "<GiftCardTag: GiftCardTag object (290)>": instance is on database "replica", value is on database "default"

This happened on staging when executing giftCardCreate mutation. Apparently GiftCard instance was created on `replica` database and the Tag is created on `default` db.

dekoza:

Fixes #10467 

# Pull Request Checklist

* [x] The changes are tested

Internal change, no need to update changelog.

### DIFF
--- a/saleor/core/db_routers.py
+++ b/saleor/core/db_routers.py
@@ -1,4 +1,11 @@
+from django.conf import settings
+
+
 class PrimaryReplicaRouter:
+    def db_for_write(self, model, **hints):
+        """Write only to primary."""
+        return settings.DATABASE_CONNECTION_DEFAULT_NAME
+
     def allow_relation(self, obj1, obj2, **hints):
         """All relations are allowed as we don't have pool separation."""
         return True
--- a/saleor/core/db_routers.py
+++ b/saleor/core/db_routers.py
@@ -1,0 +1,18 @@
+from django.conf import settings
+
+
+class PrimaryReplicaRouter:
+    def db_for_read(self, model, **hints):
+        """Read from replica."""
+        return settings.DATABASE_CONNECTION_REPLICA_NAME
+
+    def db_for_write(self, model, **hints):
+        """Write only to primary."""
+        return settings.DATABASE_CONNECTION_DEFAULT_NAME
+
+    def allow_relation(self, obj1, obj2, **hints):
+        """All relations are allowed as we don't have pool separation."""
+        return True
+
+    def allow_migrate(self, db, app_label, model_name=None, **hints):
+        return True
--- a/saleor/core/db_routers.py
+++ b/saleor/core/db_routers.py
@@ -1,18 +1,4 @@
-from django.conf import settings
-
-
 class PrimaryReplicaRouter:
-    def db_for_read(self, model, **hints):
-        """Read from replica."""
-        return settings.DATABASE_CONNECTION_REPLICA_NAME
-
-    def db_for_write(self, model, **hints):
-        """Write only to primary."""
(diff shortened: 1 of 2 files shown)


saleor/core/db_routers.py
```diff
@@ -0,0 +1,18 @@
+from django.conf import settings
+
+
+class PrimaryReplicaRouter:
+    def db_for_read(self, model, **hints):
+        """Read from replica."""
+        return settings.DATABASE_CONNECTION_REPLICA_NAME
```
fowczarek, line 7: Won't this switch our whole read traffic to replicas? 
In my opinion, we shouldn't switch all read traffic. If we want to read from a replica we should do it explicitly. 
When we switch to always read from replica we could cause that situation when we respond to mutation from replica before this database receives propagated data from primary database.

patrys, line 7: We can't implement this method. We don't want to read from a replica during mutations, especially not inside transactions when we do `SELECT ... FOR UPDATE`.

saleor/core/db_routers.py
```diff
@@ -0,0 +1,18 @@
...
+    def db_for_write(self, model, **hints):
+        """Write only to primary."""
+        return settings.DATABASE_CONNECTION_DEFAULT_NAME
+
+    def allow_relation(self, obj1, obj2, **hints):
+        """All relations are allowed as we don't have pool separation."""
+        return True
+
+    def allow_migrate(self, db, app_label, model_name=None, **hints):
+        return True
```
patrys, line 18: Why implement this?

dekoza:
OK, so basically I should only leave `allow_relation`, right?

saleor/core/db_routers.py
dekoza: Removed.

saleor/core/db_routers.py
dekoza: Removed.

dekoza: Removed.

saleor/core/db_routers.py
```diff
@@ -0,0 +1,4 @@
+class PrimaryReplicaRouter:
+    def allow_relation(self, obj1, obj2, **hints):
```
fowczarek, line 2:
```suggestion
class PrimaryReplicaRouter:

    def db_for_write(self, model, **hints):
        """Write only to primary."""
        return settings.DATABASE_CONNECTION_DEFAULT_NAME

    def allow_relation(self, obj1, obj2, **hints):
```

fowczarek (approved):

saleor/settings.py
```diff
@@ -89,6 +89,8 @@ def get_bool_from_env(name, default_value):
     # ),
 }
 
+DATABASE_ROUTERS = ["saleor.core.db_routers.PrimaryReplicaRouter"]
```
cmiacz, line 92: We'll have to override this class in cloud builds, however methods used here does not conflict with the DB router for cloud.

cmiacz: What Saleor versions will have this change?

cmiacz (approved):

saleor/settings.py
dekoza: Right now this goes only to main, so 3.8 onwards.

IKarbowiak (approved):

PR #10532, opened 2022-09-01
Add taxCalculationStrategy field

maarcingebala:
This PR changes the schema to reflect the app integration options. It introduces new `TaxCalculationStrategy` enum to explicitly select whether Saleor should use a tax app or flat rates for calculations.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-6391 Simple taxes](https://app.clickup.com/t/2549495/SALEOR-6391)

patrys:
Task linked: [SALEOR-6798 App integration](https://app.clickup.com/t/2549495/SALEOR-6798)

saleor/tax/models.py
```diff
@@ -80,6 +81,9 @@ class TaxConfiguration(ModelWithMetadata):
         Channel, related_name="tax_configuration", on_delete=models.CASCADE
     )
     charge_taxes = models.BooleanField(default=True)
+    tax_calculation_strategy = models.CharField(
```
korycins, line 84: Do you expect that in the future we will have more options here? If not, wouldn't be simpler to make just boolean field like `use_flat_rate`?

korycins: Also I think, that it would be usefull to have here a default value (if we are able to determine a proper value in migrations)

korycins (commented):
If I set `FLAT_RATES`, it means that we shouldn't send a tax webhook for this channel. Am I right?  If we don't do that, then it is part of the app implementation to handle this but Saleor can't guarantee that the taxes for this channel will not be calculated by App.

saleor/tax/models.py
maarcingebala: Null is the default value- it is allowed that taxes are not calculated, so no method is selected. I would keep the enum, is more readable to me.

patrys: `use_flat_rate` does not explain what happens when the option is `false`. Disabling taxes means not charging tax, it's not the same as not calculating tax. If you provide gross prices, we still need to determine a net price, and we need to know how to do it (app or flat rate).

maarcingebala: I solved the NON_TAXES value by making the enum optional in API.

saleor/graphql/tax/mutations/tax_configuration_update.py
```diff
@@ -119,12 +139,14 @@ def update_countries_configuration(cls, instance, countries_configuration):
             data = input_data_by_country[obj.country]
             obj.charge_taxes = data["charge_taxes"]
             obj.display_gross_prices = data["display_gross_prices"]
+            obj.tax_calculation_strategy = data["tax_calculation_strategy"]
```
IKarbowiak, line 142: As the `tax_calculation_strategy` is not required it might raise an error when there will be not `taxCalculationStrategy` in the input, am I right? IMO, we should use `get` here.

saleor/graphql/tax/mutations/tax_configuration_update.py
```diff
@@ -134,6 +156,7 @@ def update_countries_configuration(cls, instance, countries_configuration):
                 tax_configuration=instance,
                 country=item["country_code"],
                 charge_taxes=item["charge_taxes"],
+                tax_calculation_strategy=item["tax_calculation_strategy"],
```
IKarbowiak, line 159: Same here.

saleor/graphql/tax/mutations/tax_configuration_update.py
IKarbowiak: So it seems like there is also a lack of test for this situation.

maarcingebala: Good catch 👍

korycins (approved):

fowczarek (approved):

saleor/graphql/tax/mutations/tax_configuration_update.py
maarcingebala: Fixed

saleor/graphql/tax/mutations/tax_configuration_update.py
maarcingebala: Fixed

saleor/graphql/tax/mutations/tax_configuration_update.py
```diff
@@ -119,12 +139,14 @@ def update_countries_configuration(cls, instance, countries_configuration):
             data = input_data_by_country[obj.country]
             obj.charge_taxes = data["charge_taxes"]
             obj.display_gross_prices = data["display_gross_prices"]
+            obj.tax_calculation_strategy = data.get("tax_calculation_strategy", None)
```
IKarbowiak, line 142: `data.get("tax_calculation_strategy", None)`, `None` can pe dropped. 🚔

saleor/graphql/tax/mutations/tax_configuration_update.py
```diff
@@ -134,6 +156,7 @@ def update_countries_configuration(cls, instance, countries_configuration):
                 tax_configuration=instance,
                 country=item["country_code"],
                 charge_taxes=item["charge_taxes"],
+                tax_calculation_strategy=item.get("tax_calculation_strategy", None),
```
IKarbowiak, line 159: Same here 👮‍♀️

IKarbowiak (approved):

saleor/graphql/tax/mutations/tax_configuration_update.py
maarcingebala: 👍

saleor/graphql/tax/mutations/tax_configuration_update.py
maarcingebala: 👍

korycins (approved):

IKarbowiak (approved):

Issue #9452, opened 2022-04-01
Add optional filter parameter to properties


timuric:
### What I'm trying to achieve
I would like to be able to filter some of the array properties to avoid implementing filtering client-side as well as not inflate payload. 
Pages as PLP do not require all attributes or images but only a few to render thumbnails.

Use cases:
• Query only the first image of the product for thumbnails
• Query a specific attribute such as color

### Describe a proposed solution

Add filter parameters to the following properties
```media(filter: { sortOrder: Int! } )```
```attribute( slug: String! )```
```variant( filter: { sku: String, name: String } )```

PR #10537, opened 2022-09-02
New properties for Product type for improving client-side filters

korycins:
The RFC prepared based on the description from the issue https://github.com/saleor/saleor/issues/9452.

The assumption is that we add similar interface to `variant`, `attributes` as we have for `metafield`. 

```graphql

enum MediaChoicesSortField {
  ID
}

input MediaSortingInput {
  """Specifies the direction in which to sort products."""
  direction: OrderDirection!

  """Sort attribute choices by the selected field."""
  field: MediaChoicesSortField!
}

"""Represents an individual item for sale in the storefront."""
type Product implements Node & ObjectWithMetadata {
  id: ID!

  metafield(key: String!): String
  attributes: [SelectedAttribute!]!
  variants: [ProductVariant!]

 (...)
  media(sortBy: MediaSortingInput): [ProductMedia!]
  attribute(slug:String!): SelectedAttribute
  variant(sku:String, id: ID): ProductVariant
}
```

Fields to update:
 - Give a possibility to sort media by provided order. If not provided, the default order will be used.
```graphql
media(sortBy: MediaSortingInput): [ProductMedia!]
```
Fields to add:
 - Fetch single attribute attached to the product based on attribute slug
```graphql
attribute(slug:String!): SelectedAttribute
```

- Fetch single variant based on SKU or ID.
```graphql
variant(sku:String, id: ID): ProductVariant
```

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


fowczarek:
LGTM :tada:

IKarbowiak:
Maybe instead of adding a new field to `Product`, where we already have `variants` and `default_variant` fields, we should add the filter option to `Product.variants` field?

timuric:
> Maybe instead of adding a new field to `Product`, where we already have `variants` and `default_variant` fields, we should add the filter option to `Product.variants` field?
> Same for `attributes`.

Difference is that singular field would not add pagination, which makes it much more pleasant to work with on the front end

afewyards:
Lgtm!

saleor/graphql/product/types/products.py
```diff
@@ -1090,37 +1148,54 @@ def resolve_image_by_id(root: ChannelContext[models.Product], _info, *, id):
...
+            keys["id"] = int(pk)
+
+        if sku is not None:
+            keys["sku"] = sku
+
+        def get_product_variant_by_id_and_sku(
+            product_variants, keys
+        ) -> Optional[ProductVariant]:
+            def condition(variant):
+                con = True
```
IKarbowiak, line 1180: IMO, we should use better variable name than `con`

saleor/graphql/product/types/products.py
```diff
@@ -1090,37 +1148,54 @@ def resolve_image_by_id(root: ChannelContext[models.Product], _info, *, id):
...
+                for key, val in keys.items():
```
IKarbowiak, line 1181: What are `key` and `value`? IMO, we should name it with something meaningful.  For example, instead of `key`, we could have `field_name` and `val` could be `value`.

saleor/graphql/product/types/products.py
```diff
@@ -1090,37 +1148,54 @@ def resolve_image_by_id(root: ChannelContext[models.Product], _info, *, id):
...
+                    if val is None:
+                        con &= getattr(variant, key) is None
+                    else:
+                        con &= getattr(variant, key) == val
```
IKarbowiak, line 1185: Do we need this `if - else` part, for me it looks like it could be just:
`con &= getattr(variant, key) == val`.

saleor/graphql/core/dataloaders.py
```diff
@@ -50,3 +50,17 @@ def batch_load_fn(  # pylint: disable=method-hidden
 
     def batch_load(self, keys: Iterable[K]) -> Union[Promise[List[R]], List[R]]:
         raise NotImplementedError()
+
+
+class DataLoaderWithSort(DataLoader):
```
korycins, line 55: I don't think that we need this. 
You can have simple `class MediaByProductIdLoaderSort(DataLoader)` and then in `resolve_media`
```
    @staticmethod
     def resolve_media(root: ChannelContext[models.Product], info, sort_by=None):
         if sort_by is None:
             sort_by = {
                 "field": MediaChoicesSortField.ID.value,  # type: ignore
                 "direction": "",
             }
         def sort_media(media):
             ...
             return sorted_media
         return MediaByProductIdLoader(info.context, sort_by).load(root.node.id).then(sort_media)
```

With current approach, If someone use normal `Media` dataloader and `MediaWithSort` in the same query, then the loader will fetch the data twice. This is why we should use the same dataloader.

saleor/graphql/product/tests/test_product.py
```diff
@@ -4061,6 +4061,40 @@ def test_query_product_image_by_invalid_id(
     assert content["data"]["product"]["imageById"] is None
 
 
+@pytest.mark.parametrize(
+    "sort_by",
+    ({"field": "ID", "direction": "ASC"}, {"field": "ID", "direction": "DESC"}),
```
korycins, line 4066: You could also add a test that if you don't provide any sort input, then the default sorter is used (media has field order)

saleor/graphql/product/types/products.py
```diff
@@ -1090,37 +1148,54 @@ def resolve_image_by_id(root: ChannelContext[models.Product], _info, *, id):
...
-                ChannelContext(node=variant, channel_slug=root.channel_slug)
-                for variant in variants
-            ]
+        variants = resolve_variants(root, info)
+        return variants.then(lambda x: get_product_variant_by_id_and_sku(x, keys))
 
-        return variants.then(map_channel_context)
+    @staticmethod
+    def resolve_variants(root: ChannelContext[models.Product], info):
+        return resolve_variants(root, info)
```
korycins, line 1198: Not sure about extracting resolver outside the type. What do we get by extracting it outside the type? I see that you use it also in `resolve_variant` but you could also call `Product.resolve_variants(root, info)`, right?

saleor/graphql/product/types/products.py
```diff
@@ -1090,37 +1148,54 @@ def resolve_image_by_id(root: ChannelContext[models.Product], _info, *, id):
...
-        if has_required_permissions and not root.channel_slug:
-            variants = ProductVariantsByProductIdLoader(info.context).load(root.node.id)
-        elif has_required_permissions and root.channel_slug:
-            variants = ProductVariantsByProductIdAndChannel(info.context).load(
-                (root.node.id, root.channel_slug)
-            )
-        else:
-            variants = AvailableProductVariantsByProductIdAndChannel(info.context).load(
-                (root.node.id, root.channel_slug)
+    def resolve_variant(root: ChannelContext[models.Product], info, id=None, sku=None):
```
korycins, line 1164: I think that you can make this simpler. Something like this one:
```python
    @staticmethod
    def resolve_variant(root: ChannelContext[models.Product], info, id=None, sku=None):
        if id is None and sku is None:
            return None

        def get_product_variant(variants):
            if id:
                id_type, variant_id = graphene.Node.from_global_id(id,)
                if id_type != "ProductVariant":
                    return None
                variant = [variant for variant in variants if variant.pk == variant_id ]
                return variant[0] if variant else None
            if sku:
                variant = [variant for variant in variants if variant.sku == sku ]
                return variant[0] if variant else None
            return None

        variants = resolve_variants(root, info)
        return variants.then(get_product_variant)
```
No need to pass `keys` to dataloader. The function used in `then` already has an access to `resolve_variant`'s context. 
No need to check both of these fields. Personally I would say that we should extend description that if both fields are provided then we return null.

saleor/graphql/schema.graphql
```diff
@@ -4472,6 +4472,12 @@ type Product implements Node & ObjectWithMetadata {
   """A type of tax. Assigned by enabled tax gateway"""
   taxType: TaxType
 
+  """Get a single attribute attached to product by attribute slug"""
```
korycins, line 4475: New fields should get `ADDED_IN_38` label

saleor/graphql/product/types/products.py
```diff
@@ -743,6 +744,31 @@ class Meta:
         node = ProductVariant
 
 
+def resolve_variants(root: ChannelContext[models.Product], info):
```
fowczarek, line 747: IMO this method should be in `saleor/graphql/product/resolvers.py`

kadewu:
Two failing tests are not related to the changes from that PR.

saleor/graphql/product/tests/test_product.py
```diff
@@ -4061,6 +4061,44 @@ def test_query_product_image_by_invalid_id(
...
+    variables = {
+        "id": graphene.Node.to_global_id("Product", product_with_image_list.pk),
+        "channel": channel_USD.slug,
+        "sort_by": sort_by,
+    }
+
+    response = staff_api_client.post_graphql(query, variables)
+    content = get_graphql_content(response)
+    media = content["data"]["product"]["media"]
+    media1 = graphene.Node.from_global_id(media[0]["id"])
```
korycins, line 4094: AFAIK, `from_global_id` returns a tuple (`Type`, `Id`). Is it expected here that you take into account both values?
Shouldn't be, `_, media1 = graphene.Node....`

IKarbowiak:
@kadewu The `pytest` CI is mandatory to pass. Maybe you need to rebase your PR.

saleor/graphql/product/types/products.py
```diff
@@ -1094,13 +1132,63 @@ def resolve_image_by_id(root: ChannelContext[models.Product], _info, *, id):
...
+        return MediaByProductIdLoader(info.context).load(root.node.id).then(sort_media)
 
     @staticmethod
     def resolve_images(root: ChannelContext[models.Product], info):
         return ImagesByProductIdLoader(info.context).load(root.node.id)
 
+    @staticmethod
+    def resolve_variant(root: ChannelContext[models.Product], info, id=None, sku=None):
+        if id is not None and sku is not None:
+            return None
```
fowczarek, line 1160: IMO we could use `validate_one_of_args_is_in_query` from [saleor/saleor/graphql/core/validators/__init__.py](https://github.com/mirumee/saleor/blob/a5c3dd97b5c81bb8121617ed019971fb96ab8172/saleor/graphql/core/validators/__init__.py#L21)
```suggestion
        validate_one_of_args_is_in_query("id", id, "slug", slug)
```

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add ability to filter by slug. #10578 by @kadewu
   - Affected types: Attribute, Category, Collection, Menu, Page, Product, ProductType, Warehouse
   - Deprecated `slug` in filter for `menus`. Use `slugs` instead
+- Add new properties for `Product` by @kadewu
+  - `attribute` and `variant`
+  - `sortBy` to `media`
```
maarcingebala, line 13:
```suggestion
- Add new properties in the `Product` type - #10537 by @kadewu
  - Add new fields: `Product.attribute`, `Product.variant`
  - Add `sortBy` argument to `Product.media`
```

saleor/graphql/product/sorters.py
```diff
@@ -299,3 +299,22 @@ class ProductTypeSortingInput(SortInputObjectType):
...
         type_name = "product types"
+
+
+class MediaChoicesSortField(graphene.Enum):
+    ID = ["id"]
+
+    @property
+    def description(self):
+        descriptions = {
+            MediaChoicesSortField.ID.name: "Sort media by id.",
```
maarcingebala, line 310:
```suggestion
            MediaChoicesSortField.ID.name: "Sort media by ID.",
```

maarcingebala (approved):

korycins:
Looks good ✅  (can't approve own PR 😅 )

saleor/graphql/product/tests/queries/test_product_query.py
```diff
@@ -1970,3 +1970,313 @@ def test_query_product_media_for_federation(
...
+    # given
+    query = """
+    query Product($id: ID!, $channel: String, $variant_id: ID, $sku: String){
+        product(id: $id, channel: $channel){
+           variant(id: $variant_id, sku: $sku){
+            id
+            sku
+           }
+        }
+    }
```
IKarbowiak, line 2032: You could create the const from that as it is done in other tests, not copy it in multiple tests.

fowczarek (approved):

IKarbowiak (approved):

PR #10575, opened 2022-09-07
Drop `AnonymouUser` from the context, and assign None instead

### LINKED ISSUE #10574: Bug: Deleting product added to draft order crashes

Reported by @NyanKiyoshi | [Sentry report](https://sentry.io/organizations/saleor/issues/3565494521/?referrer=slack) 

1. Create a draft order with a product you want to delete (don't finalize it)
2. Create an app
3. Delete the product that is assigned to draft order using the app's token
4. Should crash
```
mutation DeleteProduct($id: ID!) {
  productDelete(id: $id) {
    product {
      id
    }
  }
}
```

Should pass

I prepared a test that replicates this:

```python
def test_draft_order_delete_product(
        app_api_client,
        permission_manage_products,
        draft_order
):
    query = """
        mutation DeleteProduct($id: ID!) {
          productDelete(id: $id) {
            product {
              id
            }
          }
        }
    """
    order = draft_order
    line = order.lines.first()
    product = line.variant.product
    product_id = graphene.Node.to_global_id("Product", product.id)
    variables = {"id": product_id}
    response = app_api_client.post_graphql(
        query, variables, permissions=[permission_manage_products]
    )
    # Right now this ^ goes boom.
    # Insert appropriate assertions here.
```

### Logs

_No response_

### Environment

Saleor version: 3.7
OS and version: staging server

dekoza:
Fixes #10574

Drop `AnonymouUser` from the context, and assign `None` instead.

# Pull Request Checklist

* [x] The changes are tested

### DIFF
--- a/saleor/account/events.py
+++ b/saleor/account/events.py
@@ -1,7 +1,6 @@
 from typing import Optional
 
 from ..app.models import App
-from ..core.utils.validators import user_is_valid
 from ..order.models import Order, OrderLine
 from . import CustomerEvents
 from .models import CustomerEvent, User
@@ -17,8 +16,6 @@ def customer_account_created_event(*, user: User) -> Optional[CustomerEvent]:
 def customer_account_activated_event(
     *, staff_user: UserType, app: AppType, account_id: int
 ) -> Optional[CustomerEvent]:
-    if not user_is_valid(staff_user):
-        staff_user = None
     return CustomerEvent.objects.create(
         user=staff_user,
         app=app,
@@ -30,8 +27,6 @@ def customer_account_activated_event(
 def customer_account_deactivated_event(
     *, staff_user: UserType, app: AppType, account_id: int
 ) -> Optional[CustomerEvent]:
-    if not user_is_valid(staff_user):
-        staff_user = None
     return CustomerEvent.objects.create(
         user=staff_user,
         app=app,
@@ -82,8 +77,6 @@ def customer_placed_order_event(*, user: User, order: Order) -> Optional[Custome
 def customer_added_to_note_order_event(
     *, user: UserType, order: Order, message: str
 ) -> CustomerEvent:
-    if not user_is_valid(user):
-        user = None
     return CustomerEvent.objects.create(
         user=user,
         order=order,
@@ -106,8 +99,6 @@ def customer_downloaded_a_digital_link_event(
 def customer_deleted_event(
     *, staff_user: UserType, app: AppType, deleted_count: int = 1
 ) -> CustomerEvent:
-    if not user_is_valid(staff_user):
-        staff_user = None
     return CustomerEvent.objects.create(
         user=staff_user,
         app=app,
@@ -120,8 +111,6 @@ def customer_deleted_event(
 def assigned_email_to_a_customer_event(
     *, staff_user: UserType, app: AppType, new_email: str
 ) -> CustomerEvent:
-    if not user_is_valid(staff_user):
-        staff_user = None
     return CustomerEvent.objects.create(
         user=staff_user,
         app=app,
@@ -134,8 +123,6 @@ def assigned_email_to_a_customer_event(
 def assigned_name_to_a_customer_event(
     *, staff_user: UserType, app: AppType, new_name: str
 ) -> CustomerEvent:
-    if not user_is_valid(staff_user):
-        staff_user = None
     return CustomerEvent.objects.create(
         user=staff_user,
         app=app,
--- a/saleor/account/events.py
+++ b/saleor/account/events.py
@@ -71,7 +71,7 @@ def customer_email_changed_event(
 
 
 def customer_placed_order_event(*, user: User, order: Order) -> Optional[CustomerEvent]:
-    if user.is_anonymous:
+    if not user:
         return None
 
     return CustomerEvent.objects.create(
--- a/saleor/account/events.py
+++ b/saleor/account/events.py
@@ -66,9 +66,6 @@ def customer_email_changed_event(
 
 
 def customer_placed_order_event(*, user: User, order: Order) -> Optional[CustomerEvent]:
-    if not user:
-        return None
-
     return CustomerEvent.objects.create(
         user=user, order=order, type=CustomerEvents.PLACED_ORDER
     )
--- a/saleor/checkout/tests/test_checkout_complete.py
+++ b/saleor/checkout/tests/test_checkout_complete.py
@@ -222,7 +222,7 @@ def test_create_order_captured_payment_creates_expected_events_anonymous_user(
             discounts=None,
             taxes_included_in_prices=True,
         ),
-        user=(),
+        user=None,
         app=None,
         manager=manager,
     )
@@ -480,7 +480,7 @@ def test_create_order_preauth_payment_creates_expected_events_anonymous_user(
             discounts=[],
             taxes_included_in_prices=True,
         ),
-        user=(),
+        user=None,
         app=None,
         manager=manager,
     )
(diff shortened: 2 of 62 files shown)


saleor/order/events.py
```diff
@@ -904,6 +904,8 @@ def order_line_product_removed_event(
     app: AppType,
     order_lines: List[Tuple[int, OrderLine]],
 ):
+    if not user_is_valid(user):
+        user = None
```
patrys, line 908: How did we make it this far in code without a user? It looks like you're fixing bugs on the wrong layer.

dekoza: That's a good point :thinking:

dekoza: If `requestor` is an `App` then the `user` is `AnonymousUser`. Then if we try to pass `AnonymousUser` to `OrderEvent.user` field, it crashes and rightfully so.

maarcingebala: This check is present in some event functions, but not all of them. Shouldn't we always be checking this?

IKarbowiak: We should, as we can get `AnonymousUser` there. But the better option will be not providing the `AnonymousUser` to the events, IMO we should fix the problem on the different layer as @patrys proposed.

patrys: We should not be passing AnonymousUser that deep, it should be None when we don't have a user instance.

dekoza: @patrys You're right but this fixes the issue. Rejecting user earlier in the flow requires more work with refactoring and adjusting tests. I think we should use a band-aid (merge this PR) and then work on proper solution (which is what I'm currently doing anyway).

saleor/order/events.py
```diff
@@ -762,30 +689,6 @@ def order_discount_event(
...
-):
-    for previous_order_discount, current_order_discount in changed_order_discounts:
-        order_discount_automatically_updated_event(
-            order=order,
-            order_discount=current_order_discount,
-            old_order_discount=previous_order_discount,
-        )
-
-
-def order_discount_automatically_updated_event(
```
IKarbowiak, line 776: Those events are unused.

saleor/checkout/tests/test_checkout_complete.py
```diff
@@ -221,7 +220,7 @@ def test_create_order_captured_payment_creates_expected_events_anonymous_user(
             discounts=None,
             taxes_included_in_prices=True,
         ),
-        user=AnonymousUser(),
+        user=(),
```
dekoza, line 223: Why empty tuple instead of `None`?

saleor/graphql/account/schema.py
```diff
@@ -194,7 +194,7 @@ def resolve_permission_group(_root, _info, *, id):
     @staticmethod
     def resolve_me(_root, info):
         user = info.context.user
-        return user if user.is_authenticated else None
+        return user if bool(user) else None
```
dekoza, line 197: You don't need to cast it on `bool`.

IKarbowiak: I need, as this is SimpleLazyObject, and otherwise, I will return a lazy object instead of a user instance.

dekoza: I see. Thanks for clarification.

saleor/graphql/account/mutations/authentication.py
```diff
@@ -142,6 +142,7 @@ def perform_mutation(cls, _root, info, **data):
         csrf_token = _get_new_csrf_token()
         refresh_token = create_refresh_token(user, {"csrfToken": csrf_token})
         info.context.refresh_token = refresh_token
+        info.context.user = user
```
dekoza, line 145: I assume this is necessary until I finish my work on requestor loader?

saleor/graphql/context.py
```diff
@@ -45,18 +44,18 @@ def set_app_on_context(request: SaleorContext):
...
+def get_user(request: SaleorContext) -> UserType:
     if not hasattr(request, "_cached_user"):
         request._cached_user = cast(UserType, authenticate(request=request))
     return request._cached_user
 
 
 def set_auth_on_context(request: SaleorContext):
     if hasattr(request, "app") and request.app:
-        request.user = AnonymousUser()
+        request.user = SimpleLazyObject(lambda: None)  # type: ignore
```
dekoza, line 55: I see what you did there :smirk:

saleor/graphql/core/tests/test_graphql.py
```diff
@@ -52,7 +51,7 @@ def test_jwt_middleware(client, admin_user):
     response = api_client_post(data={"query": user_details_query})
     repl_data = response.json()
     assert response.status_code == 200
-    assert isinstance(response.wsgi_request.user, AnonymousUser)
+    assert response.wsgi_request.user == None  # noqa: E711
```
dekoza, line 54: why not `is None`?

dekoza (commented):
I love the amount of code removed by @IKarbowiak :heart_eyes:

saleor/graphql/context.py
patrys: Why can't we assign `None` here directly?

saleor/graphql/core/context.py
```diff
@@ -51,3 +52,9 @@ def get_database_connection_name(context: SaleorContext) -> str:
     if not is_mutation:
         return settings.DATABASE_CONNECTION_REPLICA_NAME
     return settings.DATABASE_CONNECTION_DEFAULT_NAME
+
+
+def setup_context_user(context: SaleorContext) -> None:
+    if getattr(context.user, "_wrapped", None) is empty:
+        context.user._setup()  # type: ignore
+        context.user = context.user._wrapped  # type: ignore
```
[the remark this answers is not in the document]
IKarbowiak: But I will need to add this part to every event function. Before we had `user_is_valid`, and we will end up with something similar.

saleor/graphql/context.py
IKarbowiak: As, once I get `SimpleLazyObject` and once I get the `None` instance, it was hard to handle, I cannot use the methods for lazy objects without checking the type. IMO, we should always return the object of the same type in the context.

saleor/graphql/checkout/tests/mutations/test_checkout_complete.py
```diff
@@ -564,7 +563,7 @@ def test_checkout_complete_by_app_with_missing_permission(
         payment_data=ANY,
         store_source=ANY,
         discounts=ANY,
-        user=AnonymousUser(),
+        user=ANY,
```
fowczarek, line 566: Shouldn't we use `None` instead of `ANY`?

fowczarek (approved):

saleor/graphql/core/tests/test_graphql.py
IKarbowiak: Because `response.wsgi_request.user` is lazy. So if I compare it to `is None` it will be False.

saleor/graphql/context.py
dekoza: If this works it's fine for now. I'm changing this to dataloader anyway :wink:

saleor/account/events.py
```diff
@@ -71,7 +66,7 @@ def customer_email_changed_event(
 
 
 def customer_placed_order_event(*, user: User, order: Order) -> Optional[CustomerEvent]:
-    if user.is_anonymous:
+    if not user:
```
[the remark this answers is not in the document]
IKarbowiak: This is the event for customers, so I suppose the assumption was that we don't want to create this event if it wasn't performed by the logged user.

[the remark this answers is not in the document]
maarcingebala: Do we have similar events when an app placed an order? I think it would be more consistent to keep these functions here similar, and maybe move this check outside to where this event is called?

maarcingebala (approved):

eCzerniak (approved):

PR #10581, opened 2022-09-08
Use dataloader for plugin manager

dekoza:
Another step to async

# Impact

* Internal change

# Pull Request Checklist

* internal change, no entry in changelog


fowczarek (commented):
How about introducing `BaseMutationWithManager` which load `manager` on start and pass it as an argument to every mutation function?
Or maybe we should add it to `BaseMutation`?  
It's could be done in separate PR.

saleor/core/tests/test_middleware.py
```diff
@@ -80,6 +82,7 @@ def test_jwt_refresh_token_middleware_samesite_none(rf, customer_user, settings)
     assert cookie["samesite"] == "None"
 
 
+@pytest.mark.skip("plugins middleware removed")  # TODO: change to testing loader
```
korycins, line 85: Is it expected?

saleor/graphql/webhook/subscription_payload.py
```diff
@@ -98,7 +98,6 @@ def _get_plugins(requestor_getter):
     request.requestor = requestor  # type: ignore
     request.request_time = request_time  # type: ignore
     request.site = SimpleLazyObject(lambda: Site.objects.get_current())  # type: ignore
-    request.plugins = SimpleLazyObject(lambda: _get_plugins(requestor))  # type: ignore
```
korycins, line 99: What about this? What is a reason for removing this? Did you check this with webhook that has subscription query defined?

saleor/core/auth_backend.py
```diff
@@ -55,7 +57,8 @@ class PluginBackend(JSONWebTokenBackend):
     def authenticate(self, request=None, **kwargs):
         if request is None:
             return None
-        return request.plugins.authenticate_user(request)
+        manager = PluginsManager(settings.PLUGINS)
```
korycins, line 60: I am not conviced about this. The PluginBackend is called after we tried to fetch user by `JSONWebTokenBackend`, if it can't find a userl, then `PluginBackend` will be called. It means that for each non-authorized request we will fetch all manager's details. It means -/+ 7 db queries for each request, and we don't use dataloader here, so it means that call dataloader for plugin manager will give us 7 additionally DB queries.

saleor/graphql/plugins/dataloaders.py
```diff
@@ -19,3 +21,17 @@ def batch_load(self, keys):
             config_to_template[et.plugin_configuration_id].append(et)
 
         return [config_to_template[key] for key in keys]
+
+
+class PluginManagerByRequestorDataloader(DataLoader):
+    context_key = "plugin_by_requestor"
+
+    def batch_load(self, keys):
+        return [get_plugins_manager(lambda: key) for key in keys]
```
korycins, line 30: I don't think that you need to add `lambda` here. The plugins manager pass this value to the Plugins. `BasePlugin` has typing for this field: `Optional[Callable[[], "Requestor"]]`

saleor/graphql/app/dataloaders.py
```diff
@@ -78,5 +78,7 @@ def promise_app(context):
 
 
 def load_app(context):
+    if hasattr(context, "app"):
+        return context.app
```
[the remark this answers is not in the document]
dekoza: Yeah, I should probably fix all those test fixtures that put `app` into `context` :thinking:  Do you think this should be done here or in separate PR?

saleor/graphql/checkout/tests/benchmark/test_checkout_mutations.py
```diff
@@ -389,7 +389,7 @@ def test_create_checkout_with_reservations(
         }
     }
 
-    with django_assert_num_queries(55):
+    with django_assert_num_queries(57):  # TODO: bring back to 55
```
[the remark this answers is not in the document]
dekoza: Usually that was the case but not this time. I should probably remove the TODOs - once they are commited they turn to DON'Ts :wink:

saleor/graphql/core/tests/test_graphql.py
```diff
@@ -59,7 +59,7 @@ def test_jwt_middleware(client, admin_user):
     response = api_client_post(data={"query": create_token_query})
     repl_data = response.json()
     assert response.status_code == 200
-    assert response.wsgi_request.user == admin_user
+    assert isinstance(response.wsgi_request.user, AnonymousUser)
```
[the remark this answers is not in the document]
dekoza: I think the original logic is slightly backward - we send anonymous request with admin credentials to create new token for admin and then we check the response if its request's user is the admin we created the token for. In my opinion the user should still be anonymous or should not matter. The user was there because we had old login mechanics which this PR modifies and/or removes. In next lines we use the generated token to succesfully authenticate the admin so this assertion shouldn't even be here because it does not matter. If I'm wrong and this users has to be admin, please provide a better usecase than this test.

saleor/graphql/webhook/subscription_payload.py
dekoza: Well, all tests pass (not counting the ones that tested middleware). The whole point of my work is to get rid of each LazyObject attached to request/context. Each mention of `request.plugins` or `context.plugins` has been replaced with `load_plugins(...)` call.

saleor/core/auth_backend.py
dekoza: How exactly does a dataloader work in a request life cycle? If authentication is done once per request, that means so many times would this particular PluginsManager be called here. Please note that particular PluginManager does not use requestor because it needs to provide authentication if JWT bails out. But that makes it unusable in other cases where we need to have a PluginManager that uses requestor. So there's no use for dataloader in this particular place.

saleor/graphql/plugins/dataloaders.py
dekoza: Please note that `key` has type `Requestor` (implicit, I know, sorry) and thus `lambda: key` has exactly the type `Callable[[], Requestor]`

saleor/core/tests/test_middleware.py
dekoza: I need to redo this test, thanks.

saleor/graphql/account/tests/test_account.py
```diff
@@ -4978,11 +4978,12 @@ def test_account_reset_password_user_is_inactive(
...
     }
     response = user_api_client.post_graphql(REQUEST_PASSWORD_RESET_MUTATION, variables)
-    content = get_graphql_content(response)
-    data = content["data"]["requestPasswordReset"]
-    assert data["errors"] == [
-        {"field": "email", "message": "User with this email is inactive"}
-    ]
+    results = response.json()
+    assert "errors" in results
+    assert (
```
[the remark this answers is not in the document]
dekoza: Yes but that is not the case here as the error is one level higher. Please note that previously there was a `content = get_graphql_content(response)` call. It's not there anymore as it would crash due to assert within that function. That's why I changed this assert here.

saleor/graphql/webhook/subscription_payload.py
maarcingebala: Are we sure the tests are actually covering this? I would double-check manually as Maciej suggests.

saleor/graphql/plugins/dataloaders.py
```diff
@@ -19,3 +21,17 @@ def batch_load(self, keys):
...
+
+
+def load_plugins(request):
```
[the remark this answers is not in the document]
dekoza: Good point, thanks!

saleor/graphql/webhook/subscription_payload.py
dekoza: This is a good case to push for 100% coverage and mutation tests :wink:

saleor/core/auth_backend.py
```diff
@@ -75,10 +78,14 @@ def load_user_from_request(request):
...
     user = UserByEmailLoader(request).load(payload["email"]).get()
     user_jwt_token = payload.get("token")
-    if not user_jwt_token or not user:
+    if not user_jwt_token:
         raise jwt.InvalidTokenError(
             "Invalid token. Create new one by using tokenCreate mutation."
         )
+    elif not user:
+        raise jwt.InvalidTokenError(
+            "Invalid token. User does not exist or is inactive."
```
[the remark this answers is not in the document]
dekoza: I think that errors should be informative on what went wrong. If we don't provide this information here, the mutation will have to do this but this requires another query.

saleor/core/auth_backend.py
korycins: I made mistake in DB query calculations. It is not 7 but 3. 
With current implementation each anonymous request or with external log-in will add 3 DB queries. It is not a big number from perspective of single request, but if we will take into account  hundreds request per minute, the number is big.
The ideal approach would be change the behaviour of manager to assign requestor in different way than it is currently implemented or something that would allow us to not initialize Manager twice (as it is done now). 
I think that we should cover here at least anonymous request. Instead of initializing PluginsManager manualy, we can call `manager = AnonymousPluginManagerLoader(request).load("Anonymous").get()`
```
class PluginBackend(JSONWebTokenBackend):
    def authenticate(self, request=None, **kwargs):
        if request is None:
            return None
        
        manager = AnonymousPluginManagerLoader(request).load("Anonymous").get()
        return manager.authenticate_user(request)
```
(Anonymous request will use the same dataloader as is used here)
IMO External user could be solved in separate PR (or maybe you have an idea, how to cover this part also?)

saleor/core/tests/test_middleware.py
```diff
@@ -92,12 +95,14 @@ def test_plugins_middleware_loads_requestor_in_plugin(rf, customer_user, setting
...
     handler.get_response(request)
-    plugin = request.plugins.all_plugins.pop()
+    manager = load_plugins(request)
+    plugin = manager.all_plugins.pop()
 
     assert isinstance(plugin.requestor, type(customer_user))
     assert plugin.requestor.id == customer_user.id
 
 
+@pytest.mark.skip("plugins middleware removed")  # TODO: change to testing loader
```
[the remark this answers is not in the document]
dekoza: Fixed.

saleor/core/auth_backend.py
dekoza: Yeah, I came up with the same anonymous loader :wink: And after this PR I started working on requestor dataloader that should address what you just mentioned :smile:

saleor/core/tests/test_middleware.py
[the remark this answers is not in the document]
dekoza: Test has been changed, thanks.

saleor/core/tests/test_middleware.py
dekoza: Test restructured, thanks.

saleor/graphql/plugins/dataloaders.py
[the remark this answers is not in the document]
dekoza: Renamed.

saleor/graphql/app/dataloaders.py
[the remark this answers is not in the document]
fowczarek: I don't have an opinion on this topic. If you want to do it in a separate PR, you should create a task for this.

saleor/core/tests/test_dataloaders.py
```diff
@@ -0,0 +1,36 @@
+from django.core.handlers.base import BaseHandler
+
+from saleor.graphql.plugins.dataloaders import load_plugin_manager
```
korycins, line 3: Should be relative import.

korycins (approved):

fowczarek (approved):

PR #10585, opened 2022-09-08
Migrate Product.charge_taxes to new tax configuration

maarcingebala:
This PR migrates the `Product.charge_taxes` flag to the new tax configuration approach. From now on we use `charge_taxes` from global (channel-specific) tax configuration object or from country-specific configuration. To control whether taxes are charged on a product level, user should assign appropriate tax class:
- For flat rates, a tax class that has all country rates set to 0%
- For Avalara, a tax class with proper tax code that disabled tax calculation in Avalara. For this reason, the Avalara plugin has be changed to use metadata from TaxClass object and not products and product types.

Breaking changes:
- plugin manager's `assign_tax_code_to_object_meta` signature changed
- `charge_taxes` field in CSV export will return null values; it will be removed in Saleor 4.0

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-6804 Deprecate old tax API](https://app.clickup.com/t/2549495/SALEOR-6804)

saleor/tax/migrations/0006_auto_20220830_1555.py
```diff
@@ -1,33 +0,0 @@
-# Generated by Django 3.2.14 on 2022-08-30 15:55
```
maarcingebala, line 1: This migration should have been removed in earlier PR but I missed it. These changes were moved to tax/0001_initial migration.

saleor/webhook/serializers.py
```diff
@@ -78,10 +78,11 @@ def serialize_checkout_lines_for_tax_calculation(
     discounts: Optional[Iterable[DiscountInfo]] = None,
 ) -> List[dict]:
     channel = checkout_info.channel
-
+    charge_taxes = get_charge_taxes_for_checkout(checkout_info, lines)
     return [
         {
             **_get_checkout_line_payload_data(line_info),
+            "charge_taxes": charge_taxes,
```
fowczarek, line 85: We lose the whole feature here. It could be a breaking change if someone creates a tax app based on options to turn off taxes for some products. After upgrading to the new version we lose options to turn off taxes on one product without implementing tax_class in the tax app.

fowczarek (approved):

saleor/webhook/serializers.py
maarcingebala: In the new approach apps should use the subscription query to fetch the tax class assigned to a product (or product type) and then decide how to handle tax calculation based on it. I agree this might be breaking if anyone had an app that in 3.7.

saleor/graphql/tax/dataloaders.py
```diff
@@ -57,3 +60,18 @@ def batch_load(self, keys):
...
+    def batch_load(self, keys):
+        non_zero_rates = TaxClassCountryRate.objects.filter(
+            tax_class=OuterRef("pk")
+        ).exclude(rate=Decimal(0))
+        tax_class_map = (
+            TaxClass.objects.filter(pk__in=keys)
+            .annotate(charge_taxes=Exists(non_zero_rates))
+            .in_bulk()
+        )
+        return [tax_class_map[tax_class_id].charge_taxes for tax_class_id in keys]
```
IKarbowiak, line 77: We will get an error when the TaxClass with the given `id` won't exist.

maarcingebala: Fixed by adding `in_bulk(keys)` and condition that checks if the tax class exists.

saleor/webhook/serializers.py
patrys: @fowczarek Can you describe a scenario that is impossible to solve using the new logic? Under the old logic, turning off "charge taxes" for a product was functionally the same as making the tax 0%.

saleor/tax/migrations/0004_migrate_tax_classes.py
```diff
@@ -107,6 +111,35 @@ def migrate_product_tax_codes(apps, _schema_editor):
             Product.objects.filter(id__in=ids).update(tax_class=tax_class)
 
 
+def migrate_products_with_disabled_taxes(apps, _schema_editor):
+    Product = apps.get_model("product", "Product")
+    TaxClass = apps.get_model("tax", "TaxClass")
+    TaxClassCountryRate = apps.get_model("tax", "TaxClassCountryRate")
+
+    qs = Product.objects.filter(charge_taxes=False).order_by("pk")
+    zero_rate_tax_class, _ = TaxClass.objects.get_or_create(
```
korycins, line 120: Is it expected that you will always create the zero rate tax class? Shouldn't we create it only when we have at least one product with charge_taxes==False?

saleor/graphql/core/types/taxes.py
```diff
@@ -139,19 +145,61 @@ def resolve_source_line(root: Union[CheckoutLine, OrderLine], info):
...
+                    load_tax_configuration, country_code=country_code
+                )
+                return (
+                    ChannelByIdLoader(info.context)
+                    .load(checkout.channel_id)
+                    .then(load_tax_config_with_country)
+                )
+
+            return checkout.then(load_channel)
+        else:
```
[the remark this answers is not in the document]
maarcingebala: Yes it's not necessary, I just find it more readable :)

saleor/graphql/product/mutations/product_type_create.py
```diff
@@ -106,10 +112,6 @@ def clean_input(cls, info, instance, data):
                 is_default=True
             ).first()
 
-        tax_code = cleaned_input.pop("tax_code", "")
```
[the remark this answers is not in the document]
maarcingebala: Yes, I'll make it raise error when you use it it future PRs.

saleor/graphql/tax/dataloaders.py
```diff
@@ -57,3 +60,22 @@ def batch_load(self, keys):
...
+        tax_class_map = (
+            TaxClass.objects.filter(pk__in=keys)
+            .annotate(charge_taxes=Exists(non_zero_rates))
+            .in_bulk(keys)
+        )
+        return [
+            tax_class_map[tax_class_id].charge_taxes
+            for tax_class_id in keys
+            if tax_class_map.get(tax_class_id)
+        ]
```
IKarbowiak, line 81: It can raise an error too, as AFAIK the output list from dataloader must be the same length as the input list.

saleor/graphql/csv/enums.py
```diff
@@ -32,7 +32,7 @@ class ProductFieldEnum(graphene.Enum):
     CATEGORY = "category"
     PRODUCT_WEIGHT = "product weight"
     COLLECTIONS = "collections"
-    CHARGE_TAXES = "charge taxes"
+    CHARGE_TAXES = "charge taxes"  # deprecated; remove in Saleor 4.0
```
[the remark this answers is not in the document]
maarcingebala: This column would then return data that is not accurate or invalid, that's why we don't want to return anything here.

saleor/plugins/avatax/__init__.py
```diff
@@ -263,18 +268,36 @@ def get_checkout_lines_data(
...
+        product_type = line_info.product_type
+
+        tax_code = (
+            retrieve_tax_code_from_meta(product.tax_class, default=None)
+            if product.tax_class
+            else None
+        )
+        tax_code = (
+            tax_code or retrieve_tax_code_from_meta(product_type.tax_class)
+            if product_type.tax_class
```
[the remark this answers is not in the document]
maarcingebala: Good point, it would return the default one.

saleor/plugins/avatax/__init__.py
```diff
@@ -263,18 +268,36 @@ def get_checkout_lines_data(
...
         else False
     )
+
     for line_info in lines_info:
+        tax_code = None
+        product = line_info.product
+        product_type = line_info.product_type
+
+        tax_code = (
+            retrieve_tax_code_from_meta(product.tax_class, default=None)
```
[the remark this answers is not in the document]
maarcingebala: Not yet, I want to do all optimizations at the end.

saleor/plugins/base_plugin.py
```diff
@@ -195,7 +196,7 @@ def __str__(self):
 
     #  Assign tax code dedicated to plugin.
     assign_tax_code_to_object_meta: Callable[
-        [Union["Product", "ProductType"], Union[str, NoneType], Any], Any
+        ["TaxClass", Union[str, NoneType], Any], Any
```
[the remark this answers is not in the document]
maarcingebala: Fixed

saleor/tax/migrations/0004_migrate_tax_classes.py
maarcingebala: Fixed

saleor/graphql/tax/dataloaders.py
maarcingebala: Hmm don't we have this problem in other dataloaders as well? I'm looking at usages of `in_bulk` in dataloaders and many seem to use similar pattern.

patrys: The problem is in using `if` without an `else`. The data loader is a `map` function, and it needs to return the same number of results as there are keys and in the same order.

patrys: Most other loaders use a `defaultdict` for the map.

maarcingebala: Thank you, fixed!

saleor/graphql/tax/dataloaders.py
```diff
@@ -57,3 +60,24 @@ def batch_load(self, keys):
...
+        tax_class_map = (
+            TaxClass.objects.using(self.database_connection_name)
+            .filter(pk__in=keys)
+            .annotate(charge_taxes=Exists(non_zero_rates))
+            .in_bulk(keys)
+        )
+        return [
+            tax_class_map[tax_class_id].charge_taxes
+            if tax_class_map.get(tax_class_id)
+            else False
```
fowczarek, line 81: Shouldn't we return "None" in this case? When I pass some fake/invalid key to this dataloader I receive `False` as a response because such `TaxClass`   or `TaxClassCountryRate` don't exist.

maarcingebala: This behavior is added only for deprecated fields `Product.chargeTaxes` and `ProductType.chargeTaxes` and only relies on existing tax classes to determine if there are any non-zero rates set. If there is no tax class, there is no chance that there is any tax class, so I think returning `False` is reasonable value. I added a note about this deprecated behavior to the dataloader, so that it isn't used in any other cases.

korycins (approved):

saleor/webhook/serializers.py
fowczarek: I don't know any scenario as an example. I think about it as a theoretical edge case.  We could change this behavior and update [docs here](https://docs.saleor.io/docs/3.x/developer/extending/apps/synchronous-webhooks/tax-webhooks#fields). We suggest creating a tax app with a usage subscription base webhooks. In the subscription base webhook, the developer is able to fetch any required data to calculate taxes including tax classes that are introduced in this feature.

PR #10601, opened 2022-09-09
Prevent sending request without lines data to avatax

IKarbowiak:
- Do not append the shipping data when the price is zero
- Do not send the avatax request when there are no lines

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/plugins/avatax/plugin.py
+++ b/saleor/plugins/avatax/plugin.py
@@ -203,17 +203,11 @@ class AvataxPlugin(BasePlugin):
         discounts: Iterable[DiscountInfo],
         previous_value: TaxedMoney,
     ) -> TaxedMoney:
-        if self._skip_plugin(previous_value):
-            return previous_value
-        checkout_total = previous_value
-
-        if not _validate_checkout(checkout_info, lines):
-            return checkout_total
-        response = get_checkout_tax_data(
-            checkout_info, lines, self.config.tax_included, discounts, self.config
+        response = self._get_checkout_tax_data(
+            checkout_info, lines, discounts, previous_value
         )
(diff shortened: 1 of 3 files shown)


patrys:
Task linked: [SALEOR-6539 Attempting tax calculation with no lines i](https://app.clickup.com/t/2549495/SALEOR-6539)

saleor/plugins/avatax/plugin.py
```diff
@@ -203,17 +203,11 @@ def calculate_checkout_total(
         discounts: Iterable[DiscountInfo],
         previous_value: TaxedMoney,
     ) -> TaxedMoney:
-        if self._skip_plugin(previous_value):
-            return previous_value
-        checkout_total = previous_value
-
-        if not _validate_checkout(checkout_info, lines):
```
korycins, line 210: Does it mean that we drop `_validate_checkout` from all plugin's methods?
If yes, what about checking if user provided shipping address? User address is mandatory to call Avatax request. Without it, I don't think that there is a sense to call Avatax - it will return an error (AFAIR).

IKarbowiak: No, this is checked in `_get_checkout_tax_data` method.

IKarbowiak: I do not see a reason to check it twice 😄 . The `_get_checkout_tax_data` is the helper that check both if the plugin is active and validates the checkout.

fowczarek (approved):
Should we also check it in Avalara Exciles?

maarcingebala (approved):

PR #10606, opened 2022-09-09
[Change] Change the way transactions are handled in mutations

jakubkuc:
The main reason to do this change is to unify how we handle transactions and webhook triggering in our mutations. 
I added `call_event` method to `BaseMutation` class which will automatically detect if it is called in atomic transaction and if it is - it will trigger the webhook `on_commit`. With this, we will prevent calling webhooks with not-yet-existing objects and make webhook triggering more unified and readable.

Additionally, this PR will change `tansaction_atomic` decorator usage in mutations into more explicit and readable context manager usage. This way we also get rid of unneeded transactions and code in transactions. 

⚠️ 
Note that lambdas are called with bound variables - which is caused by a case, where multiple calls are made with the same variable with different values (e.g. in `for` loop) and got called `on_commit` - this would cause to call multiple lambdas with the same value instead of few different values. 
An example of this behavior can be seen in the picture:

image

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-7795 Change way we handle transactions and event calls in mutations](https://app.clickup.com/t/2549495/SALEOR-7795)

saleor/graphql/discount/mutations/sale_channel_listing_update.py
```diff
@@ -118,11 +118,11 @@ def remove_channels(cls, sale: "SaleModel", remove_channels: List[int]):
         sale.channel_listings.filter(channel_id__in=remove_channels).delete()
 
     @classmethod
-    @traced_atomic_transaction()
     def save(cls, info, sale: "SaleModel", cleaned_input: Dict):
         cls.add_channels(sale, cleaned_input.get("add_channels", []))
         cls.remove_channels(sale, cleaned_input.get("remove_channels", []))
-        update_products_discounted_prices_of_discount_task.delay(sale.pk)
+        with traced_atomic_transaction():
```
fowczarek, line 124: Shouldn't be a transaction be started in line 122?

saleor/graphql/order/mutations/order_cancel.py
```diff
@@ -37,7 +36,6 @@ class Meta:
         error_type_field = "order_errors"
 
     @classmethod
-    @traced_atomic_transaction()
```
fowczarek, line 40: We still need transactions from lines 46 to 52.

saleor/graphql/order/mutations/order_discount_add.py
```diff
@@ -60,21 +59,21 @@ def perform_mutation(cls, _root, info, **data):
         reason = input.get("reason")
         value_type = input.get("value_type")
         value = input.get("value")
+        with traced_atomic_transaction():
+            order_discount = create_order_discount_for_order(
+                order, reason, value_type, value
+            )
+            app = load_app(info.context)
```
fowczarek, line 66: We could load the app before the transaction.

saleor/graphql/order/mutations/order_line_discount_remove.py
```diff
@@ -34,25 +34,27 @@ def validate(cls, info, order):
...
-
-        invalidate_order_prices(order, save=True)
+        with traced_atomic_transaction():
+            remove_discount_from_order_line(
+                order_line,
+                order,
+                manager=info.context.plugins,
+                tax_included=tax_included,
+            )
+            app = load_app(info.context)
```
fowczarek, line 51: We could load the app before the transaction.

saleor/graphql/order/mutations/order_line_update.py
```diff
@@ -57,42 +56,42 @@ def clean_input(cls, info, instance, data):
...
+            line_info = OrderLineInfo(
+                line=instance,
+                quantity=instance.quantity,
+                variant=instance.variant,
+                warehouse_pk=warehouse_pk,
             )
-        invalidate_order_prices(instance.order)
-        recalculate_order_weight(instance.order)
-        instance.order.save(update_fields=["should_refresh_prices", "weight"])
+            app = load_app(info.context)
```
fowczarek, line 73: We could load the app before the transaction.

saleor/graphql/core/mutations.py
```diff
@@ -398,6 +399,14 @@ def handle_typed_errors(cls, errors: list, **extra):
             extra.update({cls._meta.error_type_field: errors})
         return cls(errors=errors, **extra)
 
+    @staticmethod
+    def call_event(func_obj):
```
maarcingebala, line 403: What do you think about this function throwing an error when it's not called in a transaction? Then we could also rename it to "on_transaction_commit" and use it as a general-purpose function that calls different things (not only events) and ensures that the call always happen after the transaction is done? CC @patrys

saleor/graphql/account/mutations/account.py
```diff
@@ -133,27 +132,29 @@ def clean_input(cls, info, instance, data, input_cls=None):
...
-        )
-        if settings.ENABLE_ACCOUNT_CONFIRMATION_BY_EMAIL:
-            user.is_active = False
-            user.save()
-            notifications.send_account_confirmation(
-                user,
-                cleaned_input["redirect_url"],
-                info.context.plugins,
-                channel_slug=cleaned_input["channel"],
+        with traced_atomic_transaction():
```
maarcingebala, line 136: Would that work if we wrapped entire `perform_mutation` in the base mutation class, so that all functions actually happens in a transaction and we don't have to wrap these pieces individually?

korycins: Keep in mind that `traced_atomic_transaction` creates `opentracing` block, that sends a span. Setting up this for each mutation will cause that we will gather there a huge amount of data.
 (Also not sure what about mutation like `tokenCreate`, that has user password as an input)

saleor/graphql/core/mutations.py
korycins: Wouldn't be more readable if instead of:
```
       cls.call_event(
             lambda: info.context.plugins.sale_created(
                 instance,
                 current_catalogue,
            )
        )
```
We would have:
```
cls.call_event(info.context.plugins.sale_created, instance, current_catalogue)
```
And inside `call_event`:
```
def call_event(func_obj, *fun_args):
    ...
     if connection.in_atomic_block:
         transaction.on_commit(func_obj, *fun_args)
     else:
         func_obj(*fun_args)
```
Thanks to this, we don't need to remember about adding `lambda`, and I think the rest of the code will be more readable.

saleor/graphql/account/mutations/account.py
fowczarek: I think that we should use transactions only when required. Having big transactions could generate problems, like that we miss some data when a transaction fails. A few months ago we speak about it with @patrys and decide it wasnt solve any of our problems.

saleor/graphql/core/mutations.py
jakubkuc: @maarcingebala As I remember we didn't want all the webhook events to be called in transaction. But we wanted to make sure that if they  were - they are called `on_commit`.

@korycins I like this idea

saleor/graphql/order/mutations/order_cancel.py
jakubkuc: true. Fixed it.

saleor/graphql/account/mutations/account.py
jakubkuc: Just like @fowczarek mentioned - the whole purpose of this PR was to explicitly use transactions only when required.

fowczarek (approved):

saleor/graphql/account/mutations/permission_group.py
```diff
@@ -215,20 +215,20 @@ class Meta:
...
-            instance.user_set.remove(*remove_users)
-        remove_permissions = cleaned_data.get("remove_permissions")
-        if remove_permissions:
-            instance.permissions.remove(*remove_permissions)
+        with traced_atomic_transaction():
+            if remove_users:
+                instance.user_set.remove(*remove_users)
+            remove_permissions = cleaned_data.get("remove_permissions")
+            if remove_permissions:
+                instance.permissions.remove(*remove_permissions)
```
IKarbowiak, line 226: Shouldn't we have whole this method in the transaction? This method is also adding the users, so if something goes wrong during user removal we will end up with saved added users and unsaved removed users.

saleor/graphql/account/mutations/staff.py
```diff
@@ -462,15 +461,15 @@ def check_if_deactivating_left_not_manageable_permissions(
             errors["is_active"].append(error)
 
     @classmethod
-    @traced_atomic_transaction()
     def _save_m2m(cls, info, instance, cleaned_data):
         super()._save_m2m(info, instance, cleaned_data)
```
IKarbowiak, line 465: Shouldn't `super()._save_m2m(...)` be in the transaction too?

saleor/graphql/order/mutations/order_line_discount_update.py
```diff
@@ -64,28 +63,28 @@ def perform_mutation(cls, _root, info, **data):
...
+                manager=manager,
+                tax_included=tax_included,
             )
-            invalidate_order_prices(order, save=True)
+            if (
+                order_line_before_update.unit_discount_value != value
+                or order_line_before_update.unit_discount_type != value_type
+            ):
+                # Create event only when we change type or value of the discount
+                app = load_app(info.context)
```
IKarbowiak, line 81: The same here :) Load app before transaction

saleor/graphql/discount/mutations/sale_create.py
```diff
@@ -90,9 +89,9 @@ def clean_instance(cls, info, instance):
             raise ValidationError({"end_date": error})
 
     @classmethod
-    @traced_atomic_transaction()
     def perform_mutation(cls, _root, info, **data):
-        response = super().perform_mutation(_root, info, **data)
+        with traced_atomic_transaction():
```
korycins, line 93: Is it expected that we reduced size of transaction here? Like previously we were sending `sale_created` with `transaction.on_commit` but now I see, that only perform_mutation is wrapped by transaction block

saleor/order/actions.py
```diff
@@ -576,15 +582,17 @@ def _increase_order_line_quantity(order_lines_info):
...
-    _decrease_stocks(order_lines_info, manager, allow_stock_to_be_exceeded)
-    _increase_order_line_quantity(order_lines_info)
+    # transaction ensures that there is a consistency between quantities in order line
+    # and stocks
+    with traced_atomic_transaction():
+        _decrease_stocks(order_lines_info, manager, allow_stock_to_be_exceeded)
+        _increase_order_line_quantity(order_lines_info)
 
 
 @traced_atomic_transaction()
```
korycins, line 598: Is it expected? You added new transaction below

saleor/order/actions.py
```diff
@@ -417,83 +418,86 @@ def approve_fulfillment(
...
-                line=order_line,
-                quantity=fulfillment_line.quantity,
-                variant=variant,
-                warehouse_pk=str(warehouse_pk) if warehouse_pk else None,
+    with traced_atomic_transaction():
+        fulfillment.status = FulfillmentStatus.FULFILLED
+        fulfillment.save()
+        order = fulfillment.order
+        if notify_customer:
+            send_fulfillment_confirmation_to_customer(
```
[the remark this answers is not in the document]
jakubkuc: I did not change it because in tests we are checking if its called. so I didn't want to change any logic here.

saleor/order/actions.py
jakubkuc: thanks, good catch. It seems I've missed it somehow

saleor/order/actions.py
```diff
@@ -14,6 +14,7 @@
 from ..core.tracing import traced_atomic_transaction
 from ..core.transactions import transaction_with_commit_on_errors
 from ..giftcard import GiftCardLineData
+from ..graphql.core.mutations import BaseMutation
```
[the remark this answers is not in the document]
jakubkuc: I like this idea. Fixed it.

saleor/graphql/core/mutations.py
```diff
@@ -403,6 +404,15 @@ def handle_typed_errors(cls, errors: list, **extra):
             extra.update({cls._meta.error_type_field: errors})
         return cls(errors=errors, **extra)
 
+    @staticmethod
+    def call_event(func_obj, *func_args):
+        """"""
```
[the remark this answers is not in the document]
jakubkuc: Good catch... I  left it empty so I wouldn't forget to fill it up 😞 thanks

saleor/order/actions.py
```diff
@@ -177,18 +179,19 @@ def cancel_order(
...
-
-    transaction.on_commit(lambda: manager.order_cancelled(order))
-    transaction.on_commit(lambda: manager.order_updated(order))
+        call_event(manager.order_cancelled, order)
+        call_event(manager.order_updated, order)
 
-    transaction.on_commit(
-        lambda: send_order_canceled_confirmation(order, user, app, manager)
-    )
+        transaction.on_commit(
```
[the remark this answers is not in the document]
jakubkuc: Yes, after latest changes `call_event` takes function object and *args to either run the function with args normally or `on_commit ` . So it can be used with any function that needs to be wrapped in `transaction.on_commit`

korycins (approved):

saleor/order/actions.py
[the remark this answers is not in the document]
jakubkuc: This will be a part of separate PR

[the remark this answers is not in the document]
maarcingebala: Also it could have more accurate name then e.g. `on_transaction_commit`, as it's not only about webhook events.

PR #10642, opened 2022-09-13
Add select_for_update to prepare checkout

jakubkuc:
I want to merge this change because it prevents the issue where checkout object is deleted before it is updated. 
The problem appears when somehow `CheckoutComplete` is called more than once on the same checkout. 
After the first thread is finished checkout is deleted, in the same time the other thread tries to operate on checkout that was deleted. `select_for_update` will prevent this from happening.

This PR also adds a new model - `CheckoutMetadata`. This model will hold all the checkout metadata so when checkout is locked by `select_for_update` the metadata can be changed by apps in sync webhooks. 

This graph shows how I want this to work now: 
[](https://mermaid.live/edit#pako:eNqdUsFqwzAM_RXj08aafYAPvXTXbdDtGDCe_dKExFamOINS-u9zmjhllMGYTsLvPelZ0klacpBKDvgcESyeGnNg48sgUrzXDOOK7dbWsC2NUVvyfYcIJXZLtlugu_tZc0NN8odMUqIj24qPoxjQwUZdEeuxdyZiVmdiEhU2NY_QxA6sKyav7VomZ48OU4-1-a-SVHCPoacwJO-vEy4mfC30J_MHRP1TkGsWiXjlMTx94fLX_w3ydhhX8y-02hQTM7NzyI30YG8al9Z6mrBSxhoepVQpdYbbUpbhnHhmjPR2DFaqyCM2ct7EcgJSVaYb0itcE4mf5zu5nMv5G6Tfxcc)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-8021 Save with update_fields did not affect any rows.](https://app.clickup.com/t/2549495/SALEOR-8021)

IKarbowiak (approved):

saleor/checkout/complete_checkout.py
```diff
@@ -607,44 +607,48 @@ def _prepare_checkout(
...
-        raise ValidationError(
-            {
-                "channel": ValidationError(
-                    "Cannot complete checkout with inactive channel.",
-                    code=CheckoutErrorCode.CHANNEL_INACTIVE.value,
-                )
-            }
+    with traced_atomic_transaction():
+        # select_for_update locks objects rows till end of transaction block
+        checkout = Checkout.objects.select_for_update().get(pk=checkout.pk)
```
korycins, line 612: In case of checkout that is already deleted, second call will raise unhandled exception here (`.get()`)

korycins (commented):
I am not sure of these changes. It protects us from raising an exception as it was raised previously but with the proposed solution, the second thread will go to the next steps related to converting checkout to order. Like processing a payment, creating an order etc. I think, that we should consider the different place for placing a `select_for_update`, or placing them before each mandatory steps.
For example, assuming that two threads are inside `_create_order`. If both of them pass the condition:
```
    order = Order.objects.filter(checkout_token=checkout.token).first()
    if order is not None:
        return order
```
Then we can have two the same orders created from the same checkout.
(I might be wrong as I just quickly checked the code, and didn't test my assumptions).

jakubkuc:
Moved to #10994

PR #10689, opened 2022-09-20
Allow to pass metadata directly in create and update mutations for product app models

SzymJ:
I want to merge this change because it adds possibility to pass `metadata` and `privateMetadata` directly in `create/update` mutations for product app models.

 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-8010 Allow metadata to be passed to Category, Collection, Product, ProductVariant, DigitalContent create mutations](https://app.clickup.com/t/2549495/SALEOR-8010)

saleor/graphql/core/mutations.py
```diff
@@ -575,13 +575,30 @@ def perform_mutation(cls, _root, info, **data):
         instance = cls.get_instance(info, **data)
         data = data.get("input")
         cleaned_input = cls.clean_input(info, instance, data)
+        metadata_list = cleaned_input.pop("metadata", None)
```
korycins, line 578: Maybe it is already covered but shouldn't we validate the permission of writing to metadata here?

saleor/graphql/product/mutations/products.py
```diff
@@ -78,9 +80,21 @@ class CategoryInput(graphene.InputObjectType):
...
+        MetadataInput,
+        description=(
+            "Fields required to update the category private metadata." + ADDED_IN_38
+        ),
+        required=False,
+    )
 
 
-class CategoryCreate(ModelMutation):
+class CategoryCreate(MutationWithMetadataMixin, ModelMutation):
```
korycins, line 97: I am not convinced about the current approach. Some metadata logic is placed in `MutationWithMetadataMixin` and some logic is placed `ModelMutation`.  It is hard to get full context jumping between mixin and model mutation. 
Can't we extend `ModelMutation` to have full metadata support?
The support should cover, validating input, validating permission for given instance, saving in the instance etc. 
We could control it by setting the flag in Meta input. Something like:
```
class CategoryCreate(ModelMutation):
    class Arguments:
         ...
         <some input with metadata>
    class Meta:
        ...
        support_meta_field=True,
        support_private_meta_field=True
```
Based on provided flags (i didn't think a lot about the naming of the fields, just use the first one to show the proposed solution), ModelMutation can control if should handle metadata fields from the input or not. Thanks to this, as a developer, one thing that you need to remember during adding new meta fields is to provide correct input fields and setting the flags to correct value.
I think that we could provide similar flow to BaseMutation but more generic (Probably we could add it only to BaseMutation, and use it in ModelMutation)

@SzymJ @maarcingebala  What do you think?

IKarbowiak: cc. @fowczarek as he's the author of the current metadata solution :)

fowczarek: IMO solution proposed by @korycins is better.

SzymJ: Done, I moved all methods to `BaseMutation`

IKarbowiak (approved):

korycins (approved):

maarcingebala (approved):

PR #10696, opened 2022-09-21
Fix deadlocks on attributeValueUpdate

tomaszszymanski129:
I want to merge this change because it fixes deadlocks on `attributeValueUpdate` mutation.

Testing method: script that executes 100 subprocesses at the same time that sends `AttributeValueUpdate` of a same attribute. Always ended up with ~10 - 20 deadlocks.

Postgres update works in a way that it creates the transaction and locks the relevant rows one by one as it goes, **in arbitrary order**. Locks are freed when the transaction either commits or rolls back. Because of that arbitrary locking order, deadlocks are created (by mutual dependency - thread1 has a lock on row A, but wants B. At the same time thread2 has a lock on B but wants on A).

ℹ️ The solution is about locking the rows for the update upfront - while ensuring the locking is performed always in the same order. This can be achieved in a subquery that does a SELECT but with ORDER BY and FOR UPDATE - then the outer update statement gets a list of IDS of already correctly locked rows.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [x] The changes are tested


patrys:
Task linked: [CLOUD-2039 Fix attributeValueUpdate causing deadlocks on RT](https://app.clickup.com/t/2549495/CLOUD-2039)

saleor/graphql/attribute/mutations.py
```diff
@@ -747,14 +748,16 @@ def success_response(cls, instance):
...
-        ).update(search_index_dirty=True)
+        with transaction.atomic():
+            variants = product_models.ProductVariant.objects.filter(
+                Exists(instance.variantassignments.filter(variant_id=OuterRef("id")))
+            )
+            qs = product_models.Product.objects.select_for_update(of=("self",)).filter(
+                Q(Exists(instance.productassignments.filter(product_id=OuterRef("id"))))
+                | Q(Exists(variants.filter(product_id=OuterRef("id"))))
+            )
+            list(qs)  # evaluate qs to enforce acquiring the lock.
```
patrys, line 759: What is the benefit of locking rows for an update immediately before the update? Update and select-for-update both use the same type of lock. This line generates an extra query but does not change how the code works.

tomaszszymanski129: without this line, no SELECT FOR UPDATE statement is being executed (only the UPDATE WHERE is executed) Therefore no lock seems to get acquired and in high-concurrency situation it deadlocks.

patrys: `SELECT … FOR UPDATE` and `UPDATE` both create the same type of lock.

saleor/graphql/attribute/mutations.py
```diff
@@ -747,14 +748,31 @@ def success_response(cls, instance):
...
-        product_models.Product.objects.filter(
-            Q(Exists(instance.productassignments.filter(product_id=OuterRef("id"))))
-            | Q(Exists(variants.filter(product_id=OuterRef("id"))))
-        ).update(search_index_dirty=True)
+        with transaction.atomic():
+            variants = product_models.ProductVariant.objects.filter(
+                Exists(instance.variantassignments.filter(variant_id=OuterRef("id")))
+            )
+            # select_for_update needs to lock objects in consistent way to avoid
+            # deadlocks notice .order_by("pk") which ensures that.
```
patrys, line 756:
```suggestion
            # SELECT … FOR UPDATE needs to lock rows in a consistent order
            # to avoid deadlocks between updates touching the same rows.
```

patrys (approved):
LGTM! :shipit:

maarcingebala (approved):

saleor/graphql/attribute/mutations.py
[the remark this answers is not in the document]
tomaszszymanski129: reworked this to use a subquery so that we don't fetch the rows at all since it's not needed (we just need proper locking to take place upfront of actual update)

NyanKiyoshi (approved):

IKarbowiak (approved):

PR #10716, opened 2022-09-23
Fix deadlocks on attributeValueUpdate

tomaszszymanski129:
ℹ️  This is a 3.7 port of https://github.com/saleor/saleor/pull/10696

I want to merge this change because it fixes deadlocks on attributeValueUpdate mutation.

Testing method: script that executes 100 subprocesses at the same time that sends AttributeValueUpdate of a same attribute. Always ended up with ~10 - 20 deadlocks.

Postgres update works in a way that it creates the transaction and locks the relevant rows one by one as it goes, in arbitrary order. Locks are freed when the transaction either commits or rolls back. Because of that arbitrary locking order, deadlocks are created (by mutual dependency - thread1 has a lock on row A, but wants B. At the same time thread2 has a lock on B but wants on A).

information_source The solution is about locking the rows for the update upfront - while ensuring the locking is performed always in the same order. This can be achieved in a subquery that does a SELECT but with ORDER BY and FOR UPDATE - then the outer update statement gets a list of IDS of already correctly locked rows.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/attribute/mutations.py
+++ b/saleor/graphql/attribute/mutations.py
@@ -2,7 +2,8 @@ from typing import TYPE_CHECKING
 
 import graphene
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
-from django.db.models import Exists, OuterRef, Q
+from django.db import transaction
+from django.db.models import Exists, OuterRef, Q, Subquery
 from django.utils.text import slugify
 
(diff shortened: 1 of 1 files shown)


patrys:
Task linked: [CLOUD-2039 Fix attributeValueUpdate causing deadlocks on RT](https://app.clickup.com/t/2549495/CLOUD-2039)

fowczarek (approved):

PR #10717, opened 2022-09-23
Extend trasnaction types for new request action flow

korycins:
Changes introduces by this PR:
- Add new fields `pspReference`, `type`, `amount`, `referenceUrl` to `TransactionEventInput`, `TransactionEvent`.
- Depreceate `reference` field in `TransactionEventInput`, `TransactionEvent`. Use `pspReference` instead.
- [FEATURE PREVIEW BREAKING CHANGE] - `pspReference` and deprecated `reference` input can accept only unique values across `TransactionEvent` objects. Saleor will raise an exception if the `TransactionEvent` with provided `pspReference` already exists(error code: `UNIQUE`). Saleor can also accept `pspReference`/`reference` as Null value.
- Add new fields `pspReference`, `referenceUrl` to `TransactionCreateInput`, `TransactionUpdateInput`, and `TransactionItem`.
- Depreceate `reference` field in `TransactionCreateInput`, `TransactionUpdateInput`, and `TransactionItem`. Use `pspReference` instead. The field will be removed in Saleor 3.9, as the API is a preview feature.
- [FEATURE PREVIEW BREAKING CHANGE] - `pspReference` and deprecated `reference` can accept only unique values across `TransactionItem` objects. Saleor will raise an exception if the transaction with provided `pspReference` already exists(error code: `UNIQUE`). Saleor can also accept `pspReference`/`reference` as Null value.
- Rename `TransactionStatus` to `TransactionEventStatus`.
- Add `REQUEST` to `TransactionEventStatus`.
- Deprecate `TransactionEventStatus.PENDING` - Will be removed in Saleor3.9, as the API is the preview feature.
- Add `CANCEL` to `TransactionActionEnum`.
- Deprecate `TransactionActionEnum.VOID` - Will be removed in Saleor3.9, as the API is the preview feature Use `CANCEL` instead.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/payment/mutations.py
```diff
@@ -603,7 +611,22 @@ class TransactionUpdateInput(graphene.InputObjectType):
     type = graphene.String(
         description="Payment type used for this transaction.",
     )
-    reference = graphene.String(description="Reference of the transaction.")
+    reference = graphene.String(
+        description=(
+            "Reference of the transaction. "
+            "The reference and PSP reference must be unique across all "
+            "`transactionItem` objects. "
+            "DEPRECATED: this field will be removed in Saleor 3.9 (Feature Preview). "
```
IKarbowiak, line 619: IMO, you should use `DEPRECATED_IN_3X_INPUT` label to make it consistent.

saleor/graphql/payment/mutations.py
```diff
@@ -634,12 +663,43 @@ class TransactionCreateInput(TransactionUpdateInput):
...
         required=True,
         description="Current status of the payment transaction.",
     )
-    reference = graphene.String(description="Reference of the transaction.")
+    reference = graphene.String(
+        description=(
+            "Reference of the transaction. "
+            "The reference and PSP reference must be unique across all "
+            "`transactionEvent` objects. "
+            "DEPRECATED: this field will be removed in Saleor 3.9 (Feature Preview). "
```
IKarbowiak, line 675: Same here.

saleor/graphql/payment/mutations.py
```diff
@@ -675,12 +735,26 @@ def validate_metadata_keys(cls, metadata_list: List[dict], field_name, error_cod
...
+                        code=error_code,
+                    )
+                }
+            )
+
+    @classmethod
+    def validate_external_url(
+        cls, external_url: Optional[str], parent_field_name: str, error_code: str
+    ):
+        if external_url is None:
```
IKarbowiak, line 748: If the `external_url` will be an empty string we will raise an error, is it expected?

saleor/graphql/payment/types.py
```diff
@@ -260,20 +260,48 @@ class Meta:
...
+        TransactionEventStatusEnum,
         description="Status of transaction's event.",
         required=True,
     )
     reference = graphene.String(
-        description="Reference of transaction's event.", required=True
+        description="Reference of transaction's event.",
+        required=True,
+        deprecation_reason=(
+            "This field will be removed in Saleor 3.9 (Feature Preview). "
```
IKarbowiak, line 271: Use `DEPRECATED_IN_3X_FIELD`

saleor/graphql/payment/types.py
```diff
@@ -299,7 +327,17 @@ class TransactionItem(ModelObjectType):
     )
     status = graphene.String(description="Status of transaction.", required=True)
     type = graphene.String(description="Type of transaction.", required=True)
-    reference = graphene.String(description="Reference of transaction.", required=True)
+    reference = graphene.String(
+        description="Reference of transaction.",
+        required=True,
+        deprecation_reason=(
+            "This field will be removed in Saleor 3.9 (Feature Preview). "
```
IKarbowiak, line 334: Same here 👮‍♀️

IKarbowiak (approved):

saleor/graphql/payment/mutations.py
korycins: But `DEPRECATED_IN_3X_INPUT` says it will be removed in 4.0, but this is a preview feature, so we don't need to wait till 4.0. We want to give users enough time to switch to new field and drop it in future releases of 3.X

saleor/graphql/payment/mutations.py
korycins: I would say that this is expected. Like, when `externalUrl` is None, it means that input doesn't have this value or it was provided as a `null`, but empty string is a just incorrect value provided by user. 
Do you agree?

korycins: We agreed that from frontend perspective it is better to recieve, `null` or fully valid `url`.

PR #10745, opened 2022-09-27
Fix sending sync webhooks

IKarbowiak:
The synchronous events (apart from payment action events) should be sent for all existing webhooks, not only for the first webhook with this event for a given app.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/plugins/webhook/plugin.py
+++ b/saleor/plugins/webhook/plugin.py
@@ -470,6 +470,11 @@ class WebhookPlugin(BasePlugin):
         previous_value,
         **kwargs
     ) -> "GatewayResponse":
+        """Trigger payment webhook event.
+
+        Only one app should have defined the webhook for payment event.
+        If more than one app has, the webhook is sent only for the first one.
+        """
         if not self.active:
             return previous_value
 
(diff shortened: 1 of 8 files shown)


patrys:
Task linked: [SALEOR-8062 ORDER_FILTER_SHIPPING_METHODS doesn't exclude from available shipping methods](https://app.clickup.com/t/2549495/SALEOR-8062)

fowczarek (approved):

korycins (approved):

saleor/plugins/webhook/plugin.py
```diff
@@ -470,6 +470,11 @@ def __run_payment_webhook(
         previous_value,
         **kwargs
     ) -> "GatewayResponse":
+        """Trigger payment webhook event.
+
+        Only one app should have defined the webhook for payment event.
+        If more than one app has, the webhook is sent only for the first one.
```
tomaszszymanski129, line 476: I think we should consider blocking this behavior, or at least give a warning at the point of creation, of course not in the scope of this PR. WDYT?

tomaszszymanski129 (approved):

saleor/plugins/webhook/plugin.py
IKarbowiak: This is current behavior, I just added the docstring to clarify what happened. We can consider smth like that but it should be a separate ticket

PR #10833, opened 2022-10-06
Fix attribute value matching during attribute update

jakubkuc:
I want to merge this change because if we use `productVariantUpdate` mutation and update attribute values our changed values would not change if slug for the value was unchanged (e.g. slug for both "12L" and "12.L" values is "12L").
 It is because `get_or_create` was used and if id and slug matched it returned the instance instead of updating it with the new value.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/attribute/utils.py
+++ b/saleor/graphql/attribute/utils.py
@@ -143,14 +143,14 @@ class AttributeAssignmentMixin:
     def _pre_save_values(
         cls, attribute: attribute_models.Attribute, attr_values: AttrValuesInput
     ):
-        """Lazy-retrieve or create the database objects from the supplied raw values."""
-        get_or_create = attribute.values.get_or_create
+        """Lazy-update or create the database objects from the supplied raw values."""
+        update_or_create = attribute.values.update_or_create
 
         if not attr_values.values:
             return tuple()
 
         return tuple(
-            get_or_create(
+            update_or_create(
                 attribute=attribute,
                 slug=slugify(value, allow_unicode=True),
                 defaults={"name": value},


saleor/graphql/attribute/utils.py
```diff
@@ -143,14 +143,14 @@ def _resolve_attribute_global_id(cls, error_class, global_id: str) -> int:
...
-        get_or_create = attribute.values.get_or_create
+        """Lazy-update or create the database objects from the supplied raw values."""
+        update_or_create = attribute.values.update_or_create
 
         if not attr_values.values:
             return tuple()
 
         return tuple(
-            get_or_create(
+            update_or_create(
```
maarcingebala, line 153: Shouldn't we use `_update_or_create_value` here? I see it has some differences regarding the slug generation, compared to `_pre_save_values`, cc @IKarbowiak?

IKarbowiak: We cannot update the attribute value for attributes with choices, so any of these solutions is incorrect.

IKarbowiak (approved):

saleor/graphql/attribute/utils.py
```diff
@@ -144,19 +144,22 @@ def _pre_save_values(
...
-        return tuple(
-            get_or_create(
-                attribute=attribute,
-                slug=slugify(value, allow_unicode=True),
-                defaults={"name": value},
-            )[0]
-            for value in attr_values.values
-        )
+        for value in attr_values.values:
+            value_obj = attribute.values.filter(name=value).first()
```
maarcingebala, line 151: Here we run a DB query in each iteration. Can we optimize this? E.g. build some dict with all the needed data before entering the for-loop, with one DB query?

jakubkuc: If there would be more values with matching slugs, the filtering and creating unique slug would not work correctly... Or am I wrong? @IKarbowiak

maarcingebala:
@jakubkuc Please remember about updating the changelog.

saleor/graphql/attribute/utils.py
IKarbowiak: Yep. If somebody will provide `red` and `RED` we will get the same slug from `generate_unique_slug` and we will get the error when saving. We can optimize that but in such cases, we need to handle that provided values may end up with the same slug

szczecha (approved):

maarcingebala (approved):

Issue #10823, opened 2022-10-05
Bug: Email is required to apply vouchers


mmiszy:
### What are you trying to achieve?

Applying vouchers requires being logged in. It shouldn't.

### Steps to reproduce the problem

1. Go to storefront
2. Make sure you're not logged in
3. Apply a voucher
4. Observe an error

![image-6](https://user-images.githubusercontent.com/1338731/194021698-90f5b5cc-24f8-417e-9d9e-7c4e66b98df8.png)

### What did you expect to happen?

There is no reason to require an email address for vouchers.

### Environment

.

PR #10853, opened 2022-10-07
Do not require email when adding voucher to checkout

kadewu:
Fix https://github.com/saleor/saleor/issues/10823

This pull request removes email validation from adding the promo code to checkout.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/giftcard/utils.py
+++ b/saleor/giftcard/utils.py
@@ -34,8 +34,13 @@ def add_gift_card_code_to_checkout(
 ):
     """Add gift card data to checkout by code.
 
+    Raise ValidationError if email is not provided.
     Raise InvalidPromoCode if gift card cannot be applied.
     """
+    from ..checkout.checkout_cleaner import validate_checkout_email
+
+    validate_checkout_email(checkout)
+
     try:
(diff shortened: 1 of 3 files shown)


maarcingebala (approved):

IKarbowiak (changes requested):
This validation was added as this mutation is also used for adding the gift cards, and gift card validation required the email address. If we remove this condition, we will break the gift card assumptions.

IKarbowiak:
I think we may validate the checkout email only when gift card code is given.

maarcingebala:
Then we should also add some comments in code around this email validation, as it may not be obvious that it's required for gift card assumptions.

IKarbowiak:
It seems like we can move this validation to `add_gift_card_code_to_checkout`and then it will be obvious

maarcingebala:
Yep that's a good idea 👍

eCzerniak:
Please add QA when it will be ready to test.

maarcingebala (approved):

fowczarek (approved):

szczecha (approved):

PR #10892, opened 2022-10-10
Fix handling file url of file attribute values

### LINKED ISSUE #10831: There's inconsistency in a way we treat storage-based assets
## It looks that
 * `MEDIA_ROOT` and `MEDIA_URL` are settings, that Django’s core `FileSystemStorage` relies on (see: https://github.com/django/django/blob/main/django/core/files/storage.py#L212 )
 * in the same way exist `STATIC_ROOT` and `STATIC_URL` and Django’s core `StaticFilesStorage` ( see: https://github.com/django/django/blob/main/django/contrib/staticfiles/storage.py#L16)
 * neither `S3Boto3Storage` (see: https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py#L89), nor `GoogleCloudStorage` (see: https://github.com/jschneier/django-storages/blob/master/storages/backends/gcloud.py#L104) relies on settings mentioned above, instead of that they introduced their own set of settings
 * if we want to use this `django-storages` ’s backends directly we cannot relies on at least `STATIC_URL` as even Django template tag doesn’t relies on it, it uses `staticfiles_storage.url(path)` instead (see: https://github.com/django/django/blob/main/django/templatetags/static.py#L129), and that required that second pull request to Saleor@3.5

## My suggestion
* treat `MEDIA_URL`, `MEDIA_ROOT`, `STATIC_URL` and `STATIC_ROOT` as not a way to go in Django, regardless of that storage you’re using, and rather as `FileSystemStorage`  and `StaticFilesStorage`  -specific settings
 * use `django-storages` backend-specific settings, when these storages in use
 * relies on
```python
default_storage.url(path)
```
and
```python
from django.contrib.staticfiles.storage import staticfiles_storage
staticfiles_storage.url(path)
```
when URL is required.
instead of naive
```python
urljoin(settings.MEDIA_URL, path)
```

## Non-functional requirements
- [ ]  clean and document storage-oriented settings in `saleor/settings.py`

IKarbowiak:
Update `file_url` handling in values of file attributes
- reject URLs that do not start from the URLs root
- URL root is taken from the current storage
- use storage when resolving the URL

Resolves #10831

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/attribute/utils.py
+++ b/saleor/graphql/attribute/utils.py
@@ -238,10 +238,8 @@ class AttributeAssignmentMixin:
     @staticmethod
     def _clean_file_url(file_url: Optional[str], error_class):
         # extract storage path from file URL
-        storage_root_url = build_absolute_uri(default_storage.base_url)
-        if file_url is not None and not file_url.startswith(
-            storage_root_url  # type: ignore
-        ):
+        storage_root_url = build_absolute_uri(default_storage.url("/"))
+        if file_url and not file_url.startswith(storage_root_url):  # type: ignore
             raise ValidationError(
                 "The file_url must be the path to the default storage.",
                 code=error_class.INVALID.value,
--- a/saleor/graphql/attribute/utils.py
+++ b/saleor/graphql/attribute/utils.py
@@ -6,8 +6,8 @@ from typing import TYPE_CHECKING, Dict, Iterable, List, Optional, Tuple, Union
 from urllib.parse import urlparse
 
 import graphene
-from django.conf import settings
 from django.core.exceptions import ValidationError
+from django.core.files.storage import default_storage
 from django.db.models import Q
 from django.template.defaultfilters import truncatechars
 from django.utils import timezone
@@ -18,7 +18,7 @@ from text_unidecode import unidecode
 from ...attribute import AttributeEntityType, AttributeInputType, AttributeType
 from ...attribute import models as attribute_models
 from ...attribute.utils import associate_attribute_values_to_instance
-from ...core.utils import generate_unique_slug
+from ...core.utils import build_absolute_uri, generate_unique_slug
 from ...core.utils.editorjs import clean_editor_js
 from ...page import models as page_models
 from ...page.error_codes import PageErrorCode
@@ -190,7 +190,9 @@ class AttributeAssignmentMixin:
             values = AttrValuesInput(
                 global_id=global_id,
                 values=attribute_input.pop("values", []),
-                file_url=cls._clean_file_url(attribute_input.pop("file", None)),
+                file_url=cls._clean_file_url(
+                    attribute_input.pop("file", None), error_class
+                ),
                 **attribute_input,
             )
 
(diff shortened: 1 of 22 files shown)


saleor/graphql/core/types/common.py
```diff
@@ -435,7 +437,10 @@ class File(graphene.ObjectType):
 
     @staticmethod
     def resolve_url(root, info):
-        return info.context.build_absolute_uri(urljoin(settings.MEDIA_URL, root.url))
+        # check if URL is absolute:
+        if urlparse(root.url).netloc:
```
magul, line 441: Do we anticipate a situation, that `File` or `Image` object will have absolute URL? (I recall discussion with @patrys that we will have `UrlField` at some point, but that will not touch `File` and `Image` at all?)

saleor/graphql/attribute/utils.py
```diff
@@ -234,10 +236,18 @@ def clean_input(
         return cleaned_input
 
     @staticmethod
-    def _clean_file_url(file_url: Optional[str]):
+    def _clean_file_url(file_url: Optional[str], error_class):
         # extract storage path from file URL
+        storage_root_url = build_absolute_uri(default_storage.base_url)
```
magul, line 241: there's not `base_url` attribute in all storage classes that we use.

saleor/graphql/attribute/utils.py
```diff
@@ -234,10 +236,18 @@ def clean_input(
...
+        if file_url is not None and not file_url.startswith(
+            storage_root_url  # type: ignore
+        ):
+            raise ValidationError(
+                "The file_url must be the path to the default storage.",
+                code=error_class.INVALID.value,
+            )
         return (
-            re.sub(f"^{settings.MEDIA_URL}", "", urlparse(file_url).path)
+            re.sub(f"^{default_storage.base_url}", "", urlparse(file_url).path)
```
magul, line 250: two issues here:
 * `base_url` doesn't exist in all storage classes
 * in `FileSystemStorage` it tend to return `settings.MEDIA_URL`, which is just a path prefix (so will not catch protocol and host part in absolute URL)

saleor/graphql/attribute/utils.py
```diff
@@ -234,10 +236,18 @@ def clean_input(
...
             if file_url is not None
```
magul, line 251: this `if` doesn't make sense, as few lines above you `raise ValidationError` in such situation.

magul (changes requested):
`base_url` is definitely something that need be fixed. Also having unit tests, that actually test this logic using storage classes beyond `FileSystemStorage` would ensure, that it's going to work on AWS, GCP or Azure.

saleor/graphql/attribute/utils.py
IKarbowiak: No, I'm not raising an error for such a situation.

saleor/graphql/core/types/common.py
IKarbowiak: For swatch attributes, we'll have. And we don't want to change anything for swatches.

saleor/graphql/attribute/utils.py
IKarbowiak: Yeah, but here I replacing the whole part of the protocol and host.

IKarbowiak:
New PR: https://github.com/saleor/saleor/pull/10958

PR #10906, opened 2022-10-12
Fix handling file url of file attribute values

### LINKED ISSUE #10831: There's inconsistency in a way we treat storage-based assets

IKarbowiak:
Update `file_url` handling in values of file attributes
- reject URLs that do not start from the URLs root
- URL root is taken from the current storage
- use storage when resolving the URL

Resolves #10831

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/attribute/migrations/0023_lstrip_slash_from_value_file_url.py
```diff
@@ -0,0 +1,48 @@
+# Generated by Django 3.2.16 on 2022-10-12 08:35
+
+from django.db import migrations
+
+BATCH_SIZE = 10000
+
+
+def update_file_urls(apps, schema_editor):
+    AttributeValue = apps.get_model("attribute", "AttributeValue")
+    queryset = AttributeValue.objects.filter(file_url__startswith="/")
```
fowczarek, line 10: We should add an order by "pk" here. (Default ordering is `ordering = ("sort_order", "pk")` but `queryset_in_batches` required ordering by `pk`)

saleor/attribute/migrations/0023_lstrip_slash_from_value_file_url.py
```diff
@@ -0,0 +1,48 @@
...
+    for batch_pks in queryset_in_batches(queryset):
+        values = AttributeValue.objects.filter(
+            pk__in=batch_pks, file_url__startswith="/"
```
tomaszszymanski129, line 13: unnecessary extra filter case of `file_url__startswith`, you already filter the QS by that when batching.

saleor/attribute/migrations/0023_lstrip_slash_from_value_file_url.py
```diff
@@ -0,0 +1,46 @@
...
+from django.db import migrations
+
+BATCH_SIZE = 10000
+
+
+def update_file_urls(apps, schema_editor):
+    AttributeValue = apps.get_model("attribute", "AttributeValue")
+    queryset = AttributeValue.objects.filter(file_url__startswith="/").order_by("pk")
+    for batch_pks in queryset_in_batches(queryset):
+        values = AttributeValue.objects.filter(pk__in=batch_pks)
```
NyanKiyoshi, line 12: `batch_pks` should be the list of values rather than doing `SELECT ... FROM my_table WHERE id IN (<up to 10000 values here>)`

fowczarek (approved):

saleor/attribute/migrations/0023_lstrip_slash_from_value_file_url.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def update_file_urls(apps, schema_editor):
+    AttributeValue = apps.get_model("attribute", "AttributeValue")
+    queryset = AttributeValue.objects.filter(file_url__startswith="/").order_by("pk")
+    for values in queryset_in_batches(queryset):
+        for value in values:
+            value.file_url = value.file_url.lstrip("/")
+
+        AttributeValue.objects.bulk_update(values, ["file_url"])
```
tomaszszymanski129, line 15: On my local dataset (145k values) this takes 125 seconds
Instead we could go with
```
AttributeValue.objects.filter(values__in=pks).update(file_url=Func(F('file_url'), Value('/'), function='ltrim'))
```
This takes below 60 seconds to complete (`queryset_in_batches` would have to yield `pks` instead of `qs`) and from what I've checked, it works like it should. From postgres docs on LTRIM: 

_Removes the longest string containing only characters in characters (a space by default) from the start of string._

saleor/attribute/migrations/0023_lstrip_slash_from_value_file_url.py
```diff
@@ -0,0 +1,45 @@
...
+    while True:
+        qs = queryset.filter(pk__gt=start_pk)[:BATCH_SIZE]
+        pks = list(qs.values_list("pk", flat=True))
+
+        if not pks:
+            break
+
+        yield qs
+
+        start_pk = pks[-1]
```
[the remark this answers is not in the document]
IKarbowiak: It's invalid right now as I introduced the @tomaszszymanski129 suggestion.

tomaszszymanski129 (approved):

maarcingebala (approved):

Issue #10873, opened 2022-10-07
Optimize pre saving values in AttributeAssignmentMixin


jakubkuc:
In `AttributeAssignmentMixin` method `_pre_save_values` can be optimized to use bulk create. During optimization keep in mind that if multiple values, that would have the same slugs, are passed (e.g. `red`, `Red`, `R.ed`) - unique slugs need to be generated for them.

PR #10923, opened 2022-10-14
Fix duplicated attribute values

### LINKED ISSUE #10914: Bug: Duplicate attribute values

When assigning attributes for a product, by using a slug, it returns duplicated values. Seems like the API sometimes create a new value instead of using an existing one.

1. Create a list of attribute values that are similar to each other eg test1, test2, Test3, test-2
2. Fetch the product by the given ID, including the field of attributes, example query:
```graphql
query getProduct($id: ID!) {
  product(id: $id, channel: "") {
    attributes {
      attribute {
        id
        name
        choices(first: 10) {
          edges {
            node {
              name
              value
              slug
            }
          }
        }
      }
      values {
        slug
        name
      }
    }
  }
}
```
vars:
```json
{
  "id": "xxxx",
}
```
3. Grab some of the attribute slugs from the choices list
4. Use it to assign the attribute, example query:
```graphql
 mutation updateProduct($id: ID!, $input: ProductInput!) {
   productUpdate(id: $id, input: $input) {
     product {
       attributes {
         values {
           slug
           name
         }
       }
     }
   }
}
```
vars:
```json
{
  "id": "xxx",
  "input": {
    "attributes": [
      {
        "id": "xxx",
        "values": [
          "test1-2"
        ]
      }
    ],
    "category": "aaaaa",
    "chargeTaxes": true,
    "collections": [],
    "description": null,
    "name": "aaaa",
    "rating": null,
    "slug": "aaaa",
    "taxCode": null
  }
}
```
5. You end up with the wrong attribute assigned to the product

Atm seems like API works properly when we send names/values of an attribute but breaks when I send the slug. I think the algorithm could be like this:

When the user sends the value:
1. use it as a slug, if an attribute found, return its value
2. use it as value, if an attribute is found, return its value
3. if not found, create a new value, return it

It is still not perfect, but that can be a hotfix

### Logs

_No response_

### Environment

Saleor version: …
OS and version: …

IKarbowiak:
Fix handling the attribute value creation when providing values in product/variant/page updates or creation.
Right now there will be case-sensitive provided attribute value matching for name or slug:
- if a match exists for the slug, use that value
- if a match exists for the name, use that value
- otherwise, create a new attribute like it does in the new code (no get-or-create based on slugified name like in the old code)

Fixes #10914, #10873

Explain for fetching the attribute values slugs for an attribute with 400K attribute values:
https://explain.dalibo.com/plan/c59708e3g018954e

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/attribute/utils.py
+++ b/saleor/graphql/attribute/utils.py
@@ -23,7 +23,7 @@ from ...page import models as page_models
 from ...page.error_codes import PageErrorCode
 from ...product import models as product_models
 from ...product.error_codes import ProductErrorCode
-from ..core.utils import from_global_id_or_error
+from ..core.utils import from_global_id_or_error, get_duplicated_values
 from ..utils import get_nodes
 
 if TYPE_CHECKING:
@@ -581,6 +581,10 @@ class AttributeInputErrors:
         "Attribute values cannot be blank.",
         "REQUIRED",
     )
+    ERROR_DUPLICATED_VALUES = (
+        "Duplicated attribute values are provided.",
+        "DUPLICATED_INPUT_ITEM",
+    )
 
     # file errors
     ERROR_NO_FILE_GIVEN = (
@@ -774,6 +778,10 @@ def validate_values(
 ):
     name_field = attribute.values.model.name.field  # type: ignore
     is_numeric = attribute.input_type == AttributeInputType.NUMERIC
+    if get_duplicated_values(values):
+        attribute_errors[AttributeInputErrors.ERROR_DUPLICATED_VALUES].append(
+            attribute_id
+        )
     for value in values:
         if value is None or (not is_numeric and not value.strip()):
             attribute_errors[AttributeInputErrors.ERROR_BLANK_VALUE].append(
--- a/saleor/graphql/attribute/utils.py
+++ b/saleor/graphql/attribute/utils.py
@@ -527,16 +527,18 @@ def get_variant_selection_attributes(qs: "QuerySet") -> "QuerySet":
 def prepare_attribute_values(
     attribute: attribute_models.Attribute, attr_values: AttrValuesInput
 ):
+    values = attr_values.values
     slug_to_value_map = {}
     name_to_value_map = {}
-    for val in attribute.values.all():
+    for val in attribute.values.filter(Q(name__in=values) | Q(slug__in=values)):
         slug_to_value_map[val.slug] = val
         name_to_value_map[val.name] = val
 
-    existing_slugs = set(slug_to_value_map.keys())
+    existing_slugs = get_existing_slugs(attribute, values)
+
     result = []
     values_to_create = []
-    for value in attr_values.values:
+    for value in values:
         # match the value firstly by slug then by name
         value_obj = slug_to_value_map.get(value) or name_to_value_map.get(value)
         if value_obj:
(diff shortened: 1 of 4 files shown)


maarcingebala (approved):

fowczarek (approved):

saleor/graphql/attribute/utils.py
```diff
@@ -148,7 +148,7 @@ def _pre_save_values(
         if not attr_values.values:
             return tuple()
         for value in attr_values.values:
-            value_obj = attribute.values.filter(name=value).first()
+            value_obj = attribute.values.filter(Q(slug=value) | Q(name=value)).first()
```
patrys, line 151: I think we should fetch all values and prefer the one where the slug matches, consider two values:
* `name: "red", slug: "red-2"`
* `name: "RED", slug: "red"`

It will randomly pick one of them every time I save.

saleor/graphql/attribute/utils.py
```diff
@@ -534,6 +524,45 @@ def get_variant_selection_attributes(qs: "QuerySet") -> "QuerySet":
...
+    result = []
+    values_to_create = []
+    for value in attr_values.values:
+        # match the value firstly by slug then by name
+        value_obj = slug_to_value_map.get(value) or name_to_value_map.get(value)
+        if value_obj:
+            result.append(value_obj)
+        else:
+            slug = prepare_unique_slug(
+                slugify(value, allow_unicode=True), existing_slugs
```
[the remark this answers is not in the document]
IKarbowiak: This PR was from 3.5+. So I will update it from 3.5

fowczarek (approved):

SzymJ (approved):

maarcingebala (approved):

saleor/graphql/attribute/utils.py
```diff
@@ -534,6 +524,45 @@ def get_variant_selection_attributes(qs: "QuerySet") -> "QuerySet":
     )
 
 
+def prepare_attribute_values(
+    attribute: attribute_models.Attribute, attr_values: AttrValuesInput
+):
+    slug_to_value_map = {}
+    name_to_value_map = {}
+    for val in attribute.values.all():
```
patrys, line 532: What if there are 1M attribute values? Think "artist" in a music store or "author" in a book store.

tomaszszymanski129 (approved):

SzymJ (approved):

fowczarek (approved):

saleor/graphql/attribute/utils.py
```diff
@@ -534,6 +524,56 @@ def get_variant_selection_attributes(qs: "QuerySet") -> "QuerySet":
...
+            name_to_value_map[instance.name] = instance
+
+    attribute_models.AttributeValue.objects.bulk_create(values_to_create)
+    return result
+
+
+def get_existing_slugs(attribute: attribute_models.Attribute, values: List[str]):
+    lookup = Q()
+    for value in values:
+        lookup |= Q(slug__startswith=slugify(value, allow_unicode=True))
```
patrys, line 571: @IKarbowiak did you confirm if this query uses a DB index?

fowczarek: Yes, It uses Index Scan. Explain is available in PR description. 
https://explain.dalibo.com/plan/c59708e3g018954e

patrys (approved):

PR #10994, opened 2022-10-19
Move checkout metadata to separate model and lock checkout on checkout update

jakubkuc:
This PR was initially reviewed for 3.7 #10642 but will be introduced in 3.8

I want to merge this change because it prevents the issue where checkout object is deleted before it is updated. 
The problem appears when somehow `CheckoutComplete` is called more than once on the same checkout. 
After the first thread is finished checkout is deleted, in the same time the other thread tries to operate on checkout that was deleted. `select_for_update` will prevent this from happening.

This PR also adds a new model - `CheckoutMetadata`. This model will hold all the checkout metadata so when checkout is locked by `select_for_update` the metadata can be changed by apps in sync webhooks. 

This graph shows how I want this to work now: 
[](https://mermaid.live/edit#pako:eNqdUsFqwzAM_RXj08aafYAPvXTXbdDtGDCe_dKExFamOINS-u9zmjhllMGYTsLvPelZ0klacpBKDvgcESyeGnNg48sgUrzXDOOK7dbWsC2NUVvyfYcIJXZLtlugu_tZc0NN8odMUqIj24qPoxjQwUZdEeuxdyZiVmdiEhU2NY_QxA6sKyav7VomZ48OU4-1-a-SVHCPoacwJO-vEy4mfC30J_MHRP1TkGsWiXjlMTx94fLX_w3ydhhX8y-02hQTM7NzyI30YG8al9Z6mrBSxhoepVQpdYbbUpbhnHhmjPR2DFaqyCM2ct7EcgJSVaYb0itcE4mf5zu5nMv5G6Tfxcc)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-8021 Save with update_fields did not affect any rows.](https://app.clickup.com/t/2549495/SALEOR-8021)

IKarbowiak (approved):

saleor/checkout/migrations/0055_create_checkout_metadata_model_and_move_checkout_metadata.py
```diff
@@ -0,0 +1,134 @@
+import django.contrib.postgres.indexes
```
fowczarek, line 1: We should split this migration.

saleor/graphql/meta/resolvers.py
```diff
@@ -106,5 +110,8 @@ def check_private_metadata_privilege(root: ModelWithMetadata, info):
 
 
 def resolve_private_metadata(root: ModelWithMetadata, info):
+    # Metadata for checkout uses separate model to store metadata
+    if isinstance(root, checkout_models.Checkout):
+        root = root.metadata
```
fowczarek, line 115: Shouldn't we use dataloader here? We should also add a benchmark for checkout metadata or update the existing one.

saleor/graphql/meta/types.py
```diff
@@ -90,34 +91,48 @@ class ObjectWithMetadata(graphene.Interface):
...
             instance = instance.node
+        else:
+            instance = get_valid_metadata_instance(instance)
         item_type, _ = resolve_object_with_metadata_type(instance)
         return item_type
+
+
+def get_valid_metadata_instance(instance):
+    if isinstance(instance, Checkout):
+        instance = instance.metadata_storage
```
fowczarek, line 137: Shouldn't we use dataloader here?

fowczarek: Maybe instead of adding logic to all base metadata resolvers, we should override it on `Checkout` type(with dataloader).

saleor/checkout/models.py
```diff
@@ -301,3 +300,9 @@ def __setstate__(self, data):
     def is_shipping_required(self) -> bool:
         """Return `True` if the related product variant requires shipping."""
         return self.variant.is_shipping_required()
+
+
+class CheckoutMetadata(ModelWithMetadata):
```
maarcingebala, line 305: A comment would be useful to explain why in this case metadata is moved to a separate model.

saleor/graphql/meta/types.py
maarcingebala: Good idea, it should be enough if we just load proper model in the Checkout type, while this could be kept unchanged.

saleor/checkout/migrations/0055_create_checkout_metadata_model_and_move_checkout_metadata.py
```diff
@@ -0,0 +1,134 @@
...
+            name="checkout_meta_idx",
+        ),
+        migrations.RemoveField(
+            model_name="checkout",
+            name="metadata",
+        ),
+        migrations.RemoveField(
+            model_name="checkout",
+            name="private_metadata",
+        ),
```
maarcingebala, line 121: Maybe I'm too precautious about this, but what do you think about not removing these fields yet, but instead treating them just as deprecated, and removing them after we release and everyone migrates to the new solution? I'm just thinking about having a backup option in case anything went wrong.

saleor/checkout/complete_checkout.py
```diff
@@ -764,96 +765,110 @@ def complete_checkout(
...
-        if txn.customer_id and user:
-            store_customer_id(user, payment.gateway, txn.customer_id)  # type: ignore
-
-        action_required = txn.action_required
-        if action_required:
-            action_data = txn.action_required_data
-            release_voucher_usage(
-                order_data.get("voucher"), order_data.get("user_email")
+    with transaction_with_commit_on_errors():
+        if checkout_for_update:
```
maarcingebala, line 774: If you have that many lines inside an if-condition or a with-block, I find it more readable if you inverted the condition and returned from the function early, then you wouldn't have to keep the deeper indent for so many lines. Example:

```python
checkout_for_update = Checkout.something.first()
if not checkout_for_update:
   return  # some meaningful return from the function here

with transaction():
    # the actual logic
```

saleor/checkout/migrations/0055_create_checkout_metadata_model_and_move_checkout_metadata.py
maarcingebala: @fowczarek What do you think?

jakubkuc: Migrations will be changed to use `SeparateDatabaseAndState`. So the fields will be removed from state only and within the next release after this one - removed from DB as well in another migration. Is it acceptable?

saleor/checkout/complete_checkout.py
```diff
@@ -764,15 +763,29 @@ def complete_checkout(
...
-    try:
+    with transaction_with_commit_on_errors():
+        checkout = checkout_info.checkout
+        if not checkout_for_update:
+            order = Order.objects.filter(checkout_token=checkout.token).first()
+            if order:
+                return order, False, {}
+
+        channel_slug = checkout_info.channel.slug
+        payment = checkout.get_last_active_payment()
```
fowczarek, line 788: Shouldn't we use `checkout_for_update` instead of `checkout = checkout_info.checkout` inside this transaction? 
```suggestion
        payment = checkout_for_update.get_last_active_payment()
```

saleor/checkout/migrations/0056_move_checkout_metadata_to_separate_model.py
```diff
@@ -0,0 +1,64 @@
...
+    checkouts_with_meta = (
+        Checkout.objects.filter(
+            Q(metadata__isnull=False) | Q(private_metadata__isnull=False)
+        )
+        .order_by("pk")
+        .distinct("token")
+    )
+    for batch_pks in queryset_in_batches(checkouts_with_meta):
+        checkouts = Checkout.objects.filter(pk__in=batch_pks)
+        CheckoutMetadata.objects.bulk_create(
```
fowczarek, line 44: @saleor/sre 
Could we use `bulk_create` without `queryset_in_batches`?

saleor/checkout/migrations/0057_remove_metadata_from_checkout_model_state.py
```diff
@@ -0,0 +1,47 @@
...
+            ],
+        ),
+        migrations.RemoveIndex(
+            model_name="checkout",
+            name="checkout_p_meta_idx",
+        ),
+        migrations.RemoveIndex(
+            model_name="checkout",
+            name="checkout_meta_idx",
+        ),
```
fowczarek, line 34: IMO we shouldn't remove this index before we drop those fields from the database. Some queries could still use it.

saleor/graphql/checkout/types.py
```diff
@@ -1,6 +1,10 @@
 import graphene
 from promise import Promise
 
+from saleor.graphql.meta import resolvers as MetaResolvers
+
+# from ..meta.resolvers import resolve_object_with_metadata_type, \
+#     check_private_metadata_privilege, resolve_private_metadata, resolve_metadata
```
fowczarek, line 7: 🚓

saleor/checkout/migrations/0056_move_checkout_metadata_to_separate_model.py
```diff
@@ -0,0 +1,64 @@
...
+
+def move_all_checkout_metadata(apps, schema_editor):
+    Checkout = apps.get_model("checkout", "Checkout")
+    CheckoutMetadata = apps.get_model("checkout", "CheckoutMetadata")
+    checkouts_with_meta = (
+        Checkout.objects.filter(
+            Q(metadata__isnull=False) | Q(private_metadata__isnull=False)
+        )
+        .order_by("pk")
+        .distinct("token")
```
tomaszszymanski129, line 40: The checkout token is a PK for checkout, do we need the distinct?

saleor/checkout/migrations/0056_move_checkout_metadata_to_separate_model.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+    for batch_pks in queryset_in_batches(checkouts_with_meta):
+        checkouts = Checkout.objects.filter(pk__in=batch_pks)
+        CheckoutMetadata.objects.bulk_create(
+            CheckoutMetadata(
+                checkout=checkout,
+                metadata=checkout.metadata if hasattr(checkout, "metadata") else {},
```
tomaszszymanski129, line 47: is hasattr necessary? we fetch the thing from ORM so it will always have a prop "metadata", can't we simplify by 
```suggestion
                metadata=checkout.metadata or {},
```
?

saleor/checkout/migrations/0056_move_checkout_metadata_to_separate_model.py
tomaszszymanski129: I can't seem to find a way since we need a checkout PK. I thought maybe we could loop it as a subquery to return the PKs to outer func but that would nohow work with bulk_create later on. I think it should be fine how it is, there is just a memory thing to improve but I will post it as a separate comment.

saleor/checkout/migrations/0056_move_checkout_metadata_to_separate_model.py
tomaszszymanski129: also, we should use `only` method on a QS to limit the data fetched and mem usage (since we don't care at all about most of the fields)
`.only("pk", "metadata", "private_metadata")`

saleor/checkout/migrations/0057_remove_metadata_from_checkout_model_state.py
```diff
@@ -0,0 +1,47 @@
...
+        migrations.AddIndex(
+            model_name="checkoutmetadata",
+            index=django.contrib.postgres.indexes.GinIndex(
+                fields=["private_metadata"], name="checkoutmetadata_p_meta_idx"
```
tomaszszymanski129, line 38: What do we think this index would be for? Is it just for filtering the checkouts? If so maybe we could add a combined one instead of two separate ones (unless there is a case of filtering by ONLY meta OR only private meta which I'm unaware). If there is another case for more than filtering or separate filtering, ignore this comment.

patrys: We should not mix the two. They are separate namespaces, they can share the same keys and values, and you should filter them separately.

NyanKiyoshi: Adding to the question of @tomaszszymanski129: what is it used for? Isn't `private_metadata` JSONB? Is the indexing brought by `JSONB` not enough?

saleor/checkout/migrations/0056_move_checkout_metadata_to_separate_model.py
jakubkuc: distinct is used so we don't get duplicates for those Checkout object that has both metadata and private_metadata. 
Great catch with `only`, I'll fix that

patrys: `.distinct` means there is a missing `EXISTS` subquery, and we're merging an M2M table instead.

jakubkuc: @patrys fields I'm filtering with here (`metadata` and `private_metadata`) are both fields of Checkout model. I'm not sure about which M2M table merging you're mentioning. This query (with added `only` mentioned by Tomasz) will look like in the screenshot: 
image

saleor/checkout/migrations/0057_remove_metadata_from_checkout_model_state.py
jakubkuc: I created indexes to stay consistent since they are present in `ModelWithMetadata` for `metadata` and `private_metadata` fields. I think that even though Checkout metadata is in separate model the functionality should stay the same between  `ModelWithMetadata` and `CheckoutMetadata`

saleor/checkout/migrations/0057_remove_metadata_from_checkout_model_state.py
jakubkuc: Since the field is being removed from state I moved the index removal to state as well

saleor/checkout/migrations/0056_move_checkout_metadata_to_separate_model.py
patrys: How can you get duplicate results when you're querying a single table?

PR #11069, opened 2022-10-27
Remove stock rows from lock during `checkoutComplete`

zedzior:
I want to merge this change, because currently, when two users complete their checkouts, which contain the same product variant, the code can brake. It is result of `select … for update` statement, which apply lock for stock rows.

This PR solves the problem, by:
1. splitting transaction to have a potential long lasting payment call out of transaction (unlock stock rows)
2. add additional reservation for the short time to mimic the lock

Previous flow:
```mermaid
graph TD
    subgraph transaction 1
    A[start mutation] --> B[validate checkout]
    B --> C[check stocks availability]
    C --> D[process payment]
    D --> E[create order]
    end
```

New flow:
```mermaid
graph TD
    subgraph transaction 1
    A[start mutation] --> B[validate checkout]
    B --> C[check stocks availability]
    C --> D[create additional reservartion]
    end
    D --> E[process payment]
    E --> F[validate checkout again]
    subgraph transaction 2
    F --> G[compare pre-payment and post-payment checkout]
    G --> H[create order]
    end
```

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/checkout/mutations/checkout_complete.py
```diff
@@ -164,6 +167,43 @@ def validate_checkout_addresses(
...
+
+        variants = [line.variant for line in lines]
+        stocks = Stock.objects.get_variants_stocks_for_country(
+            country_code=checkout_info.get_country(),
+            channel_slug=checkout_info.channel,
+            products_variants=variants,
+        )
+        variants_stocks_map = {stock.product_variant_id: stock for stock in stocks}
+
+        dummy_reservations = []
```
patrys, line 188: It's not a dummy, we're creating a proper reservation.

saleor/graphql/checkout/mutations/checkout_complete.py
```diff
@@ -164,6 +167,43 @@ def validate_checkout_addresses(
...
+                dummy_reservations.append(
+                    Reservation(
+                        quantity_reserved=line.line.quantity,
+                        reserved_until=timezone.now()
+                        + timedelta(seconds=RESERVE_DURATION),
+                        stock=variants_stocks_map[line.variant.id],
+                        checkout_line=line.line,
+                    )
+                )
+        Reservation.objects.bulk_create(dummy_reservations)
```
patrys, line 200: I don't like that we now have multiple implementations of reservation creation. There should be a common set of helpers like the ones for increasing/decreasing stock shared between this and reservations created for checkout.

saleor/graphql/checkout/mutations/checkout_complete.py
```diff
@@ -164,6 +167,43 @@ def validate_checkout_addresses(
...
+
+        if not site_settings.reserve_stock_duration_anonymous_user:
+            site_settings.reserve_stock_duration_anonymous_user = RESERVE_DURATION
```
patrys, line 203: This modifies the site settings object.

saleor/graphql/checkout/mutations/checkout_complete.py
zedzior: The thing is, existing `def reserve_stocks()` function, include stock availability check. But in some cases, with the additional reservation approach, we we will exceed number of stocks available and `reserve_stock` function will return an `InsufficientStock` error.

saleor/checkout/complete_checkout.py
```diff
@@ -559,7 +559,9 @@ def _create_order(
         additional_warehouse_lookup,
         check_reservations=is_reservation_enabled(site_settings),
         checkout_lines=[line.line for line in checkout_lines],
+        lock=False,
```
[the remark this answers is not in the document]
zedzior: I implemented suggested approach, with some exceptions:

1. Update of payment object is done out of transaction, within `_process_payment` function, but further in the code, in the second transaction, in case of any error we trigger `gateway.payment_refund_or_void` function, which updates the payments object back.
2. There is `validate_checkout_address` call out of transaction, which can update the address when is valid. Is it necessary to keep it within transaction?

saleor/graphql/checkout/mutations/checkout_complete.py
patrys: We want to validate stock in the first transaction. So the atomic operation is two-step: validate stock (locking it, so we know it's valid) and create reservation (while the stock is locked to avoid two processes reserving the same stock), then unlock both.

saleor/checkout/complete_checkout.py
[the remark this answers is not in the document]
patrys: Checks are tricky, we may need to run some of them twice, before and after the payment call, each time with the checkout object locked, to make sure no one else is modifying the checkout while we turn it into an order. The stock check does not have to be repeated as long as we have a valid reservation.

saleor/graphql/checkout/mutations/checkout_complete.py
zedzior: Validation is done in the first transaction, here: `_get_order_data` -> `_prepare_order_data` -> `_create_lines_for_order`-> `check_stock_and_preorder_quantity_bulk(..., check_reservations=True)`

SzymJ (approved):

saleor/checkout/complete_checkout.py
```diff
@@ -897,7 +947,6 @@ def _create_order_lines_from_checkout_lines(
         checkout_info,
         lines,
         discounts,
-        reservation_enabled,
```
maarcingebala, line 900: `reservation_enabled` is passed as an argument but is not used, can we drop it?

IKarbowiak (approved):

saleor/checkout/complete_checkout.py
```diff
@@ -353,7 +352,7 @@ def _create_lines_for_order(
         additional_filter_lookup=additional_warehouse_lookup,
         existing_lines=lines,
         replace=True,
-        check_reservations=check_reservations,
+        check_reservations=True,
```
korycins, line 355: What about the second flow for creating an order. It looks like this change has impact also on `_create_order_from_checkout` -> `_create_order_lines_from_checkout_lines` -> `_create_lines_for_order`.  Is it expected that we check reservation for it? The new flow is one step checkout (we don't process a payment like in checkout_complete), so I am not sure if we need to force checking reservation.

saleor/graphql/checkout/mutations/checkout_complete.py
```diff
@@ -164,116 +174,260 @@ def validate_checkout_addresses(
...
+        result["site"] = load_site(info.context)
+        result["tracking_code"] = analytics.get_client_id(info.context)
+
+        return result
+
+    @classmethod
+    def perform_mutation(
+        cls, _root, info, store_source, checkout_id=None, token=None, id=None, **data
+    ):
+        with transaction_with_commit_on_errors():
```
korycins, line 333: I am not convinced that the main body of complete_checkout logic should placed in perform_mutation body. 
The main task mutation should be validation of the data, and calling bussiness logic (previously `complete_checkout`). 
I think that `complete_checkout` function should do the logic that is placed currently in the `perform_mutation`.

saleor/checkout/complete_checkout.py
patrys: @korycins I don't understand the above. What do you mean by "force checking reservation?" Every bit of code that uses stock needs to respect reservations.

saleor/checkout/complete_checkout.py
```diff
@@ -1213,3 +1222,140 @@ def create_order_from_checkout(
...
+                txn=txn,
+                order_data=order_data,
+                user=user,
+                app=app,
+                site_settings=site_settings,
+                metadata_list=metadata_list,
+                private_metadata_list=private_metadata_list,
+            )
+        else:
+            release_voucher_usage(
```
korycins, line 1305: Question: If we are releasing a voucher and payment, in case of mismatch between order_data. Shouldn't we also drop the reservation for the checkout?

patrys: We should drop it if reservations are disabled.

korycins (commented):
Wouldn't be good to have some tests, that will confirm that checkout complete creates the reservation?
Also, wouldn't be good to have a test for checking the number of DB queries? We add some logic to the checkout, I think it would be worth confirming that the DB queries didn't increase too much.

saleor/checkout/complete_checkout.py
zedzior: Reservation is done for a short time, currently set on 45 seconds. As I recall, it is a max time for processing payments.

zedzior:
> Wouldn't be good to have some tests, that will confirm that checkout complete creates the reservation?
@korycins 
How can we test the state in the middle of mutation?

fowczarek (approved):

IKarbowiak (approved):

eCzerniak (approved):

saleor/graphql/checkout/tests/mutations/test_checkout_complete.py
```diff
@@ -3198,3 +3197,72 @@ def test_checkout_complete_with_not_normalized_billing_address(
...
+    assert reservations[0].reserved_until <= timezone.now() + timedelta(
+        seconds=settings.RESERVE_DURATION
+    )
+    assert Order.objects.count() == orders_count
+
+    gateway_response = action_required_gateway_response
+    gateway_response.action_required = False
+    mocked_process_payment.return_value = action_required_gateway_response
+
+    response_2 = user_api_client.post_graphql(MUTATION_CHECKOUT_COMPLETE, variables)
```
korycins, line 3261: IMHO. It should be two separate tests. If this one will fail, we will not know which part of the reservation is wrong.
I would suggest to have single test for creating a reservation during the first call of checkoutComplete. And the second one for handling the checkout that already had some actions.

PR #11128, opened 2022-11-02
Various small fixes

patrys:
* Enabled type checking in VSCode
* Various small fixes in type hints
* Fixed a flaky address generator

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/core/tests/test_core.py
```diff
@@ -248,14 +248,13 @@ def test_validate_image_file():
...
-    img = SimpleUploadedFile("product.jpg", img_data.getvalue(), "image/jpeg")
     field = "image"
 
     # when
-    result = validate_image_file(img, field, ProductErrorCode)
+    img = SimpleUploadedFile("product.jpg", img_data.getvalue(), "image/jpeg")
 
     # then
-    assert not result
+    validate_image_file(img, field, ProductErrorCode)
```
IKarbowiak, line 257: `validate_image_file` is a method that we are testing not `SimpleUploadedFile`, so IMO it should go to `# when` part.

fowczarek (approved):

maarcingebala (commented):
Maybe we can also fix this: https://github.com/saleor/saleor/discussions/11044

saleor/graphql/core/tests/test_core.py
patrys: The assertion is that calling this method does not raise an exception.

>When a valid JPEG file is uploaded, then `validate_image_file` does not raise.

IKarbowiak: So from my perspective, we are checking if `validate_image_file` is not raising an error, so it's the method that we are testing. 🤔

patrys: Ok, what is the test then? "When `validate_image_file` is called, then… `pass`?"

The old code checked if a function that never returns returned nothing.

IKarbowiak: Ok, I get your point of view. We sometimes use `# when` and `# then` together, but it looks more clear.

IKarbowiak (approved):

patrys:
@maarcingebala Removed that part of `.editorconfig` as it was never standardized and is not widely supported.

maarcingebala (approved):

PR #11129, opened 2022-11-02
REQUEST transactionEvent: transactionRequestAction

kadewu:
Add REQUEST event transaction when request refund

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/payment/tests/mutations/test_transaction_request_action.py
+++ b/saleor/graphql/payment/tests/mutations/test_transaction_request_action.py
@@ -122,6 +122,13 @@ def test_transaction_request_charge_action_for_order(
     assert Decimal(event.parameters["amount"]) == expected_called_charge_amount
     assert event.parameters["reference"] == transaction.psp_reference
 
+    assert TransactionEvent.objects.get(
+        transaction=transaction,
+        status=TransactionEventStatus.REQUEST,
+        type=TransactionAction.CHARGE,
+        amount_value=expected_called_charge_amount,
+    )
+
 
 @pytest.mark.parametrize(
     "refund_amount, expected_called_refund_amount",
@@ -189,6 +196,8 @@ def test_transaction_request_refund_action_for_order(
     assert TransactionEvent.objects.get(
         transaction=transaction,
         status=TransactionEventStatus.REQUEST,
+        type=TransactionAction.REFUND,
+        amount_value=expected_called_refund_amount,
     )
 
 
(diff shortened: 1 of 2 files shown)


saleor/graphql/payment/tests/mutations/test_transaction_request_action.py
```diff
@@ -573,3 +583,59 @@ def test_transaction_request_action_missing_event(
     )
 
     assert mocked_is_active.called
+
+
+@patch("saleor.plugins.manager.PluginsManager.is_event_active_for_any_plugin")
+def test_transaction_request_action_missing_event_refund_requested(
```
korycins, line 589: I think that you could extend above test (test_transaction_request_action_missing_event) to confirm that we didn't create an REQUEST event.

korycins (approved):

fowczarek (approved):

IKarbowiak:
@kadewu tests are failing.

saleor/graphql/payment/tests/mutations/test_transaction_request_action.py
```diff
@@ -122,6 +122,13 @@ def test_transaction_request_charge_action_for_order(
     assert Decimal(event.parameters["amount"]) == expected_called_charge_amount
     assert event.parameters["reference"] == transaction.psp_reference
 
+    assert TransactionEvent.objects.get(
```
IKarbowiak, line 125: IMO we should use `exists` instead of `get` in tests as if it the objects doesn't exist we will get the error from the `get` method instead of the assertion error.

saleor/graphql/payment/tests/mutations/test_transaction_request_action.py
```diff
@@ -573,3 +615,13 @@ def test_transaction_request_action_missing_event(
...
+
+    assert (
+        len(
+            TransactionEvent.objects.filter(
+                transaction=transaction,
+                status=TransactionEventStatus.REQUEST,
+            )
+        )
+        == 0
+    )
```
IKarbowiak, line 627: You could use `exists` here:
```suggestion
assert not TransactionEvent.objects.filter(
      transaction=transaction,
      status=TransactionEventStatus.REQUEST,
  ).exists()
```

kadewu:
> @kadewu tests are failing.

Yes, we have failing tests on the target branch.

saleor/graphql/payment/tests/mutations/test_transaction_request_action.py
kadewu: The difference is, that I'm checking if there is only one record with get()

IKarbowiak (approved):

PR #11172, opened 2022-11-09
Denormalize tax class for order lines and orders

maarcingebala:
TaxClass instance needs to be denormalized for order lines to keep the tax rate information even when tax rate is removed from the system.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-6391 Simple taxes](https://app.clickup.com/t/2549495/SALEOR-6391)

IKarbowiak (approved):

saleor/tax/utils.py
```diff
@@ -179,6 +179,11 @@ def normalize_tax_rate_for_db(tax_rate: Decimal) -> Decimal:
     return tax_rate / 100
 
 
+def denormalize_tax_rate_from_db(tax_rate: Decimal) -> Decimal:
+    # Revert results of `normalize_tax_rate_for_db`.
+    return tax_rate * 100
```
korycins, line 184: Question: shouldn't we apply a `decimal.quantize` here?
```
In [2]: Decimal(0.22)*100
Out[2]: Decimal('22.00000000000000011102230246')
```

saleor/tax/utils.py
```diff
@@ -179,6 +179,11 @@ def normalize_tax_rate_for_db(tax_rate: Decimal) -> Decimal:
     return tax_rate / 100
```
korycins, line 179: I know, that this is not a scope of this PR, but shouldn't we have a protection here for divide by zero?
something like:
```
if tax_rate == 0:
    return 0
```

korycins (approved):

korycins:
Also @maarcingebala  question. Shouldn't we add labels to the new fields? ADDED_IN_X, PREVIEW_FEATURE?

SzymJ (approved):

saleor/tax/utils.py
maarcingebala: Hmm I think this is a good idea.

maarcingebala: I'm checking this and it's not `Decimal(0.22)` but `Decimal("0.22")`. Any multiplication on that seems to give correct values:
```
ipdb> order.shipping_tax_rate
Decimal('0.2300')
ipdb> order.shipping_tax_rate * 100
Decimal('23.0000')
```

patrys: We don't convert floats to decimals, there is no need to do any quantization or rounding here.

saleor/tax/utils.py
maarcingebala: We're dividing by 100 always here, `0/100` is allowed ;)

eCzerniak (approved):

PR #11264, opened 2022-11-21
Move checkout metadata to separate model

jakubkuc:
I want to merge this change because we want to move checkout metadata to a separate model to be able to manipulate its metadata while it is locked during `checkout_complete`.

This PR originally was part of #10994 but due to changes to code in `main`, it was decided to split it into two parts. 
# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/checkout/models.py
```diff
@@ -299,3 +299,9 @@ def __setstate__(self, data):
     def is_shipping_required(self) -> bool:
         """Return `True` if the related product variant requires shipping."""
         return self.variant.is_shipping_required()
+
+
+class CheckoutMetadata(ModelWithMetadata):
+    checkout = models.OneToOneField(
+        Checkout, related_name="metadata_storage", on_delete=models.CASCADE
+    )
```
korycins, line 307: I think that you need to add `class Meta(ModelWithMetadata.Meta):` here. Without this, the indexes declared for metadata will not be created.

saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
```diff
@@ -231,12 +231,16 @@ def _update_delivery_method(
         )
         checkout.save(
             update_fields=[
-                "private_metadata",
                 "shipping_method",
                 "collection_point",
             ]
             + invalidate_prices_updated_fields
         )
+        checkout.metadata_storage.save(
```
korycins, line 239: Will it work in case of missing metadata_storage for a given checkout? (I am asking because I saw a helper method that creates the metadata object in the fly)

saleor/graphql/meta/types.py
```diff
@@ -115,9 +120,19 @@ def resolve_private_metafields(root: ModelWithMetadata, info, *, keys=None):
...
+            from ..checkout.types import Checkout as CheckoutType
+
+            return CheckoutType.resolve_type(instance, info)
         item_type, _ = resolve_object_with_metadata_type(instance)
         return item_type
+
+
+def get_valid_metadata_instance(instance):
+    if isinstance(instance, Checkout):
+        instance = instance.metadata_storage
```
korycins, line 137: Wont fail in case of missing metadata_storage relation?

korycins (commented):
In many places you're calling `checkout.metadata_storage.(private_)metadata`. Some functions have check `if hasattr(checkout, "metadata_storage")` ([example](https://github.com/saleor/saleor/pull/11264/files#diff-ec3617c5fe3b16ee657128e0f46d1a59d970e27d1435b1462d660b493d7bb478R1319)) but the rest of them will fail with error `RelatedObjectDoesNotExist` ([example](https://github.com/saleor/saleor/pull/11264/files#diff-933aa67b3e21a369ce178ab7c3e42cb1cf9b6a4f358fc87018dd4f9ca99f772dR355)). (Please correct me if I missed something).

saleor/graphql/checkout/types.py
```diff
@@ -1,6 +1,8 @@
 import graphene
 from promise import Promise
 
+from saleor.graphql.meta import resolvers as MetaResolvers
```
fowczarek, line 4: Absolute import 🚓

saleor/payment/gateways/adyen/utils/common.py
```diff
@@ -352,7 +352,7 @@ def request_data_for_gateway_config(
         country_code = country.code
     else:
         country_code = Country(settings.DEFAULT_COUNTRY).code
-    channel = checkout.get_value_from_metadata("channel", "web")
+    channel = checkout.metadata_storage.get_value_from_metadata("channel", "web")
```
fowczarek, line 355: Should we handle the case that we have `checkout` without `metadata_storage`?

saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
jakubkuc: Yes, because to update or delete metadata it uses helper methods `set_external_shipping_id` and `delete_external_shipping_id` that will create `checkout.metadata_storage` if it is missing.

saleor/checkout/models.py
jakubkuc: Will they? with just that `makemigrations` is trying to generate those indexes properly, filtering also works fine. Should anything be checked to make sure the indexes are properly created?

saleor/graphql/meta/types.py
jakubkuc: It seems that it would. Good catch!

jakubkuc:
> In many places you're calling `checkout.metadata_storage.(private_)metadata`. Some functions have check `if hasattr(checkout, "metadata_storage")` ([example](https://github.com/saleor/saleor/pull/11264/files#diff-ec3617c5fe3b16ee657128e0f46d1a59d970e27d1435b1462d660b493d7bb478R1319)) but the rest of them will fail with error `RelatedObjectDoesNotExist` ([example](https://github.com/saleor/saleor/pull/11264/files#diff-933aa67b3e21a369ce178ab7c3e42cb1cf9b6a4f358fc87018dd4f9ca99f772dR355)). (Please correct me if I missed something).
I went through the changes and all the usages are after a helper method `get_or_create_checkout_metadata` is being called. And it will create this relation in case it doesn't exist. The only case that is missed is what @fowczarek has found in adyen plugin.

saleor/checkout/migrations/0057_remove_metadata_from_checkout_model_state.py
```diff
@@ -0,0 +1,33 @@
...
+            ],
+        ),
+        migrations.RemoveIndex(
+            model_name="checkout",
+            name="checkout_p_meta_idx",
+        ),
+        migrations.RemoveIndex(
+            model_name="checkout",
+            name="checkout_meta_idx",
+        ),
```
fowczarek, line 32: IMO we shouldn't delete indexes before we drop fields from the database. The database should be backward compatible with the previous Saleor version.

saleor/graphql/checkout/types.py
```diff
@@ -748,7 +750,107 @@ def calculate_display_gross_prices(tax_configs_per_country):
...
+    @classmethod
+    def resolve_type(cls, root: models.Checkout, _info):
+        item_type, _ = MetaResolvers.resolve_object_with_metadata_type(
+            root.metadata_storage
+        )
+        return item_type
+
 
+# Checkout metadata is moved to separate model so it can be used when checkout model is
+# locked by select_for_update during complete_checkout.
```
fowczarek, line 853: I'm not sure that it's a good place for this comment. It suggests that the comment describes something related to the below class.

jakubkuc: good catch...I have no idea how it ended up here...

saleor/checkout/migrations/0057_remove_metadata_from_checkout_model_state.py
jakubkuc: After fix they are removed only from state

fowczarek (approved):

saleor/checkout/models.py
korycins: Hhm, I was sure, that I didn't find the indexes in the migrations, but it looks like it placed there.

saleor/payment/utils.py
```diff
@@ -176,7 +176,7 @@ def create_payment_information(
         email = cast(str, checkout.get_customer_email())
         user_id = checkout.user_id
         checkout_token = str(checkout.token)
-        checkout_metadata = checkout.metadata
+        checkout_metadata = checkout.metadata_storage.metadata
```
maarcingebala, line 179: No get_or_create here?

saleor/core/utils/anonymization.py
```diff
@@ -80,6 +80,6 @@ def anonymize_checkout(checkout: "Checkout") -> "Checkout":
     anonymized_checkout.shipping_address = generate_fake_address()
     anonymized_checkout.billing_address = generate_fake_address()
     anonymized_checkout.note = fake.paragraph()
-    anonymized_checkout.metadata = generate_fake_metadata()
-    anonymized_checkout.private_metadata = generate_fake_metadata()
+    anonymized_checkout.metadata_storage.metadata = generate_fake_metadata()
+    anonymized_checkout.metadata_storage.private_metadata = generate_fake_metadata()
```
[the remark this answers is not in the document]
jakubkuc: It doesn't use metadata_storage because its Order model not Checkout - only Checkout metadata is moved to separate model.

saleor/checkout/migrations/0056_move_checkout_metadata_to_separate_model.py
```diff
@@ -0,0 +1,60 @@
...
+
+        start_pk = pks[-1]
+
+
+def move_all_checkout_metadata(apps, schema_editor):
+    Checkout = apps.get_model("checkout", "Checkout")
+    CheckoutMetadata = apps.get_model("checkout", "CheckoutMetadata")
+    checkouts_with_meta = (
+        Checkout.objects.filter(
+            Q(metadata__isnull=False) | Q(private_metadata__isnull=False)
```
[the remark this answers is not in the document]
jakubkuc: Good catch. You're right it needs to be changed to search for non-empty dicts

saleor/core/utils/anonymization.py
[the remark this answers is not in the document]
maarcingebala: Yes, right 👍

szczecha (approved):

PR #11282, opened 2022-11-23
Fix observability reporter - add URLs obfuscation

przlada:
I want to merge this change because it adds URL obfuscation to the observability reporter, as webhook target URLs can contain sensitive information.
For example:

- `awssqs://key:secret@sqs.us-east-2.amazonaws.com/xxxx/myqueue.fifo` -> `awssqs://key:***@sqs.us-east-2.amazonaws.com/xxxx/myqueue.fifo`

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


kadewu (approved):

saleor/webhook/observability/obfuscation.py
```diff
@@ -47,6 +48,16 @@ def hide_sensitive_headers(
     }
 
 
+def obfuscate_url(url: str) -> str:
+    parts = urlparse(url)
+    if parts.username is None:
+        return url
```
NyanKiyoshi, line 54: RFC 1738 says empty username is valid (it's not no user, but empty username)
```
>>> urlparse('https://:passwd@tst.com/foo').password
passwd
```

NyanKiyoshi (changes requested):

saleor/webhook/observability/obfuscation.py
przlada: @NyanKiyoshi that is a great point, but this case is handled. If the user in URL is empty 
`urlparse("https://:passwd@tst.com/foo").username` returns an empty string, not `None`. So if username is `None`, that means there are no credentials in the URL. It is checked in the test case: 
```
"https://:password@example.com/test" -> "https://:{MASK}@example.com/test"
```

NyanKiyoshi: Indeed, I missed this. Maybe you could add a comment above the line to make it more explicit?

NyanKiyoshi (approved):

saleor/webhook/observability/obfuscation.py
przlada: done 😄

IKarbowiak (approved):

Issue #11277, opened 2022-11-23
Make email case insensitive 


timuric:
### What are you trying to achieve?

Make emails case insensitive to prevent login issues.

### Steps to reproduce the problem

Create account Test@test.com
Login with test@test.com

### What did you expect to happen?

It should be possible to login under the same account using test@test.com and Test@test.com. Essentially any provided email should be read in lower case. With exception if the system has two accounts under test@test.com and Test@test.com then login should happen to the corresponding account provided. 

### Environment

Saleor version: 3.9

korycins:
@timuric  Should we also take into account the case where we have email provided with upper case? Like [test@test.com](mailto:test@test.com), [Test@test.com](mailto:Test@test.com) and TEST@TEST.COM?

PR #11284, opened 2022-11-23
Make email authentication case insensitive

zedzior:
https://github.com/saleor/saleor/issues/11277

I want to merge this change, because I want to make email authentication case insensitive. It should be possible to login under the same account using [test@test.com](mailto:test@test.com) and [Test@test.com](mailto:Test@test.com).

This PR also force emails to be lower case only.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/account/tests/mutations/test_token_create.py
```diff
@@ -221,3 +224,34 @@ def test_create_token_active_user_logged_before(api_client, customer_user, setti
...
+        User.objects.create_user("andrew2@example.com", "password"),
+    ]
+    return users
+
+
+@freeze_time("2020-03-18 12:00:00")
+@pytest.mark.parametrize(
+    "email,expected_user",
+    [
+        ("andrew@example.com", 0),
```
korycins, line 243: Maybe I am wrong but I think, that the implemented logic works as previously:
You provide `Andrew@example.com` and expect that `Andrew@example.com` will be fetched from DB.
As fair as I understand the concept, you should provide `Andrew@example.com` and we should try to find a match, but if we can't then we should try to find `andrew@example.com`
Also, I think that, the registration mutation should be changed to also check&save emails in lower letters.
@timuric Can you confirm that I am talking with sense? 😅

timuric: Yep, registration should be in lower case, and we need to provide also logic for password reset

zedzior: So it means, that we don't need to handle exception, mentioned in issue, right?

> With exception if the system has two accounts under [test@test.com](mailto:test@test.com) and [Test@test.com](mailto:Test@test.com) then login should happen to the corresponding account provided.

saleor/graphql/account/utils.py
```diff
@@ -510,3 +510,21 @@ def check_is_owner_or_has_one_of_perms(
     """
     if not is_owner_or_has_one_of_perms(requestor, owner, *perms):
         raise PermissionDenied(permissions=list(perms) + [AuthorizationFilters.OWNER])
+
+
+def retrieve_user_by_email(email):
```
fowczarek, line 515: IMO it's shouldn't be in `graphql` path. Maybe it's should be in `/saleor/account/utils.py`. 
We should add some unit tests for this function.

zedzior: agree, fits better, thanks

saleor/account/tests/test_utils.py
```diff
@@ -189,3 +191,41 @@ def test_remove_the_oldest_user_address_if_address_limit_is_reached_limit_reache
...
+
+@pytest.fixture
+def users_with_similar_emails():
+    users = [
+        User.objects.create_user("andrew@example.com", "password"),
+        User.objects.create_user("Andrew@example.com", "password"),
+        User.objects.create_user("john@example.com", "password"),
+        User.objects.create_user("Susan@example.com", "password"),
+        User.objects.create_user("Cindy@example.com", "password"),
+        User.objects.create_user("CINDY@example.com", "password"),
```
IKarbowiak, line 204: I'm wondering if we need to have passwords set for those users. If not, we can create the user instances in bulk instead of creating each user separately. 🤔

zedzior: You are right, we don't need password anymore, thanks!

IKarbowiak (approved):

szczecha:
In therms of This PR also force emails to be the lower case only.
Should staffCreate customerCreate mutations return email to lowercase?

Screenshot 2022-12-02 at 09 37 58
Screenshot 2022-12-02 at 09 33 24
Screenshot 2022-12-02 at 09 31 13

szczecha (approved):

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add new properties in the `Product` type - #10537 by @kadewu
   - Add new fields: `Product.attribute`, `Product.variant`
   - Add `sortBy` argument to `Product.media`
+- Allow assigning attribute value using its ID. Add to `AttributeValueInput` dedicated field for each input type. #11206 by @zedzior
```
maarcingebala, line 31: Let's remove this duplicated line.

maarcingebala (approved):

PR #11362, opened 2022-12-01
Fix failing save on stripe plugin

korycins:
I want to merge this change because it fixes an issue related to turn off & turn on Stripe plugin.
We didn't save the `secret_webhook_key`, which causes a problem that we were not able to process the notification recieved from Stripe.
GH issue: https://github.com/saleor/saleor/issues/11373

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/payment/gateways/stripe/plugin.py
+++ b/saleor/payment/gateways/stripe/plugin.py
@@ -464,12 +464,11 @@ class StripeGatewayPlugin(BasePlugin):
     @classmethod
     def pre_save_plugin_configuration(cls, plugin_configuration: "PluginConfiguration"):
         configuration = plugin_configuration.configuration
-        flat_configuration = {item["name"]: item["value"] for item in configuration}
+        flat_configuration = {item["name"]: item for item in configuration}
 
-        api_key = flat_configuration["secret_api_key"]
-        webhook_id = flat_configuration.get("webhook_endpoint_id")
-        webhook_secret = flat_configuration.get("webhook_secret_key")
-
+        api_key = flat_configuration["secret_api_key"]["value"]
+        webhook_id = flat_configuration.get("webhook_endpoint_id", {}).get("value")
+        webhook_secret_data = flat_configuration.get("webhook_secret_key", {})
         if not plugin_configuration.active:
             if webhook_id:
                 # delete all webhook details when we disable a stripe integration.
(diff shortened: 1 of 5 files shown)


saleor/payment/gateways/stripe/plugin.py
```diff
@@ -464,12 +464,11 @@ def list_payment_sources(
...
     def pre_save_plugin_configuration(cls, plugin_configuration: "PluginConfiguration"):
         configuration = plugin_configuration.configuration
-        flat_configuration = {item["name"]: item["value"] for item in configuration}
-
-        api_key = flat_configuration["secret_api_key"]
-        webhook_id = flat_configuration.get("webhook_endpoint_id")
-        webhook_secret = flat_configuration.get("webhook_secret_key")
+        flat_configuration = {item["name"]: item for item in configuration}
 
+        api_key = flat_configuration["secret_api_key"]["value"]
```
patrys, line 469: Is it safe to assume this path always exists?

korycins: The `validate_plugin_configuration` that is called before, is checking if the value is provided.

kadewu (approved):

jakubkuc (approved):

saleor/plugins/base_plugin.py
```diff
@@ -1098,12 +1099,12 @@ def save_plugin_configuration(
 
         cls.validate_plugin_configuration(plugin_configuration)
         cls.pre_save_plugin_configuration(plugin_configuration)
+        plugin_configuration.save()
```
IKarbowiak, line 1102: Why you moved it here? I moved it to line 1106 as `_append_config_structure` also updating the config structure, and if the `save` method is called here, the changes made in `_append_config_structure` are not saved.

IKarbowiak (changes requested):

saleor/plugins/base_plugin.py
korycins: The config_structure is not a data that we should store in DB. It is a static data that is placed in `Plugin.CONFIG_STRUCTRUE`. 
IMHO `_append_config_structure` is a something that shouldn't be call here at all, as a main task of this method is attaching the config structure required for response send via API.  I've checked the code and I don't see any dependencies that would relay on the response from this method. We can drop it, but not in this PR.

IKarbowiak (approved):

szczecha (approved):

Issue #10820, opened 2022-10-04
Add external keys to objects


timuric:
### Problem
1. Saleor integrations are often, if not usually done on an existing infrastructure - that infrastructure relies on Business IDentifiers (BIDs) - which are understood across the whole platform. Introducing Saleor which can only operate on (mutate data in particular) on its own IDs (Saleor IDs). This requires additional, external mappings or API round trips for one to be able to plug Saleor to an existing platform. ERP systems are often the root of most issues of this kind. The ERP system can emit a domain event that a product was sold in a brick and mortar store, stocks for that product now need to be updated across the platform - event like that usually contains the BID of the product, the BID of the warehouse and the new stock value, the BIDs now need to be translated to Saleor IDs for Saleor to consume.
2. Another thing is batch updates where there is a lot of data to process, stock synchronization coming from a ERP system is one example. To reduce traffic other systems want to send updates in batches (one current example is a batch of 120k stock updates split to 50, 50 and 20k). Not only one needs to translate the BIDs for each, there’s also a need to update all these one-by-one.

### Functional requirements
1. There should be a way to operate (mutate) the following entities by BIDs:
a. Orders
b. Products
c. Product Variants (here SKU sometimes acts as a Product BID but it’s not a given)
d. Attributes
e. Attribute Values
f. Warehouses (only for the purpose of Stock update)
g. Customers
2. There should be a way to update the following entities in bulks (or batches): a. Stocks
b. Attributes and their Values (although it’s not MVP)

### Non-functional requirements
1. The solution should define the canonical naming for the BID fields.
2. API Call costs rules should be defined and well documented.
3. The solution should not impact the responsiveness of Saleor.
4. Batch updates might be done in an asynchronous matter, in that case, a promise should be returned which would allow checking if a batch process was completed.
5. Proposed API changes need to be backward compatible.

### Solution
The solution to the first problem would be to agree on the naming convention, here are some ideas:
• Bid - Business Identifier - an ID for an entity that is understood across the whole organization, it uniquely defines an Entity regardless of the queried system - this Foreign Key / Batch update in Saleor 2 is nice and descriptive but might be too constricting as it sounds like it has a hidden purpose to it.
• ExternalId or ForeignID - I don’t like the latter one as it steps on a “foreign” key relation naming in a database. ExternalId seems to fit the purpose but in turn, feels too loose to rely on.

The decision around this should be made by someone with a broad understanding of Saleor and where Saleor is heading.
[...]

PR #11410, opened 2022-12-07
Allow external references to be used instead of Saleor-assigned IDs

zedzior:
https://github.com/saleor/saleor/issues/10820

I want to merge this change because I want to select, update and delete records by external IDs. The new filed is called `external_reference` and will be applied to following models:
a. Orders
b. Products
c. Product Variants
d. Attributes
e. Attribute Values
f. Warehouses (only for the purpose of Stock update)
g. Customers

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/product/mutations/product/product_create.py
```diff
@@ -165,24 +165,6 @@ def clean_input(cls, info, instance, data):
         clean_seo_fields(cleaned_input)
         return cleaned_input
 
-    @classmethod
-    def get_instance(cls, info, **data):
```
korycins, line 169: Can you provide more details why we drop this part?

korycins (commented):
Looks good, but how we want to update the `externalReference`? As far as I see, we don't have any fields to save the `externalReference`. Shouldn't we have such possibility in case of `create` & `update` objects?

saleor/graphql/product/mutations/product/product_create.py
zedzior: Since it is "create" mutation, it doesn't have any `id` to serach for, so it always returns `super().get_instance(info, **data)`, therefore redundant. `ProductUpdate` is the only class, which inherits from `ProductCreate`, but it has own `get_instance` implementation.

saleor/graphql/product/resolvers.py
```diff
@@ -60,24 +60,20 @@ def resolve_digital_contents(_info):
     return models.DigitalContent.objects.all()
 
 
-def resolve_product_by_id(info, id, channel_slug, requestor):
+def resolve_product_by_id_slug_or_ext_ref(
```
maarcingebala, line 63: Maybe just `resolve_product`? I would also change `ext_ref` to `external_reference` for readability.

saleor/graphql/product/resolvers.py
```diff
@@ -122,17 +105,22 @@ def resolve_product_types(_info):
 
 
 @traced_resolver
-def resolve_product_variant_by_sku(
-    _info, sku, channel_slug, requestor, requestor_has_access_to_all
+def resolve_variant_by_id_sku_or_ext_ref(
```
maarcingebala, line 108: And here `resolve_variant`?

saleor/graphql/core/utils/resolvers.py
```diff
@@ -0,0 +1,24 @@
...
+def resolve_by_global_id_or_ext_ref(model, id, ext_ref):
+    validate_one_of_args_is_in_query("id", id, "external_reference", ext_ref)
+    if id:
+        _, id = from_global_id_or_error(id, model.__name__)
+        return model.objects.filter(id=id).first()
+    else:
+        return model.objects.filter(external_reference=ext_ref).first()
+
+
+def resolve_by_global_id_slug_or_ext_ref(model, id, slug, ext_ref):
```
maarcingebala, line 14: `ext_ref` -> `external_reference`

maarcingebala:
@korycins Do you mean allowing to pass external reference for a future order from the checkout flow? Make sense to me. Also, @timuric what do you think about renaming the field to `externalId`? It's shorter and looks more pretty in the schema. Then you would have `id: ID` and `externalId: String` and it looks to me more easy to guess about the purpose. But I'm fine with `externalReference` too if that's what was agreed earlier :)

timuric:
@korycins do you have any use case in mind for having external keys in checkout?

maarcingebala:
@timuric It's more about external reference of orders, not checkouts. As I understand it, if you knew your external order ID upfront and passed it to Saleor during checkout, it would be assigned to the newly created order.

korycins:
@timuric  I just thought that you could already have an `externalReference`, at the moment of creating the order on Saleor side. 🤔

maarcingebala:
@korycins We discussed with Timur that for now we are good with the mutations that this PR adds and we can extend it later, if there is a clear use-case for that.

korycins (commented):
@zedzior  `external_reference` is set as a unique in the abstract model. Can you confirm that we also handle the case when `external_reference` is already attached to different object? Maybe I am missing something, but I think that we will raise unhandled exception: `IntegrityError`

zedzior:
> Can you confirm that we also handle the case when `external_reference` is already attached to different object? Maybe I am missing something, but I think that we will raise unhandled exception: `IntegrityError`

@korycins You are right, the case is not handled, let me correct this.

maarcingebala (approved):

korycins (approved):

szczecha (approved):

PR #11415, opened 2022-12-07
Handle CORS at ASGI level

patrys:
This is faster than doing it inside the Django view but, like the health check, it won't work with `manage.py runserver`.

⚠️ Django's `runserver` command does not support ASGI yet, so if you want CORS to work on localhost, you'll need to run the ASGI server instead:

```
python -m saleor
```

or

```
uvicorn saleor.asgi:application --reload
```

Please note that this change will be needed anyway once we start using `asyncio` more.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
I don't know what to do with tests as they currently don't use ASGI either.

maarcingebala:
- We should add changelog entry.
- Shouldn't this be considered as a breaking change for OSS users? Is there a change someone is using a different server to run Saleor that doesn't support ASGI?
- As for the tests, I see Django has an [async client](https://docs.djangoproject.com/en/4.1/topics/testing/tools/#django.test.AsyncClient) can't we use it? Or is it a more general problem with Pytest not supporting async?

patrys:
Saleor can’t support both WSGI and ASGI if we want to take advantage of asyncio. I’ll see if I can get Pytest to play nice with ASGI, otherwise I can just test the ASGI middleware directly.

patrys:
@maarcingebala Switched tests to test the ASGI middleware explicitly.

fowczarek (approved):

fowczarek:
@patrys Should we change the command in [platform](https://github.com/saleor/saleor-platform/blob/main/docker-compose.yml#L27-L28)?

patrys:
@fowczarek I will if we decide we want to merge this.

patrys:
Here's a PR to use ASGI in `saleor-platform`: https://github.com/saleor/saleor-platform/pull/225

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ All notable, unreleased changes to this project will be documented in this file.
 - Remove `CUSTOMER_UPDATED` webhook trigger from address mutations - #11395 by @jakubkuc
 - Drop `Django.Auth` - #11305 by @fowczarek
 - Propagate voucher discount between checkout lines when charge_taxes is disabled - #11632 by @maarcingebala
+- CORS is now handled in the ASGI layer - #11415 by @patrys
```
maarcingebala, line 33: Need to be moved to 3.11 section.

maarcingebala (approved):

patrys:
Rebased on the current main branch

maarcingebala:
@patrys It needs rebase again 🙏

omars-code:
This solves the OPTIONS but not the POST

omars-code:
just an FYI there is this fork, but the package seems abandoned 
https://github.com/omars-code/asgi-cors

patrys:
@omars-code Good catch, see https://github.com/saleor/saleor/pull/12000

Issue #11337, opened 2022-11-29
Move OrderSettings to Channel type


korycins:
### What I'm trying to achieve
I would like to move `OrderSettings` from `Query` and place them inside a `Channel` type. Same way as we have `stockSettings` inside `Channel`. The proposed approach will allow us to configure order settings per channel.

#### Channel
Channel type will have a new field:
```graphql
"""Represents channel."""
type Channel implements Node {
  id: ID!
  ...
 """
  Define the stock setting for this channel.
  
  Added in Saleor 3.7.
  
  Note: this API is currently in Feature Preview and can be subject to changes at later point.
  
  Requires one of the following permissions: AUTHENTICATED_APP, AUTHENTICATED_STAFF_USER.
  """
  stockSettings: StockSettings!
  """MANAGE_ORDERS or MANAGE_CHANNELS"""
  orderSettings: OrderSettings!
}
```
Permission `MANAGE_ORDERS` or `MANAGE_CHANNELS` will be required to fetch `orderSettings`.

The `channelCreate` and `channelUpdate` mutations will get new input fields:
```graphql
input OrderSettingsInput {
  """
  When disabled, all new orders from checkout will be marked as unconfirmed. When enabled orders from checkout will become unfulfilled immediately.
  """
  automaticallyConfirmAllNewOrders: Boolean

  """
  When enabled, all non-shippable gift card orders will be fulfilled automatically.
  """
  automaticallyFulfillNonShippableGiftCard: Boolean
}
```
```graphql
input ChannelUpdateInput {
  """
  The channel stock settings.
  
  Added in Saleor 3.7.
  
  Note: this API is currently in Feature Preview and can be subject to changes at later point.
  """
  stockSettings: StockSettingsInput
  ...
  
  orderSettings: OrderSettingsInput
```
Permission `MANAGE_ORDERS` or `MANAGE_CHANNELS` will be required to update `orderSettings`.
`MANAGE_ORDERS` can be used only when `channelUpdate` mutation will get `OrderSettingsInput` and not any other fields will be updated. Otherwise needed permission will be `MANAGE_CHANNELS`

```graphql
input ChannelCreateInput {
  """
  The channel stock settings.
  
  Added in Saleor 3.7.
  
  Note: this API is currently in Feature Preview and can be subject to changes at later point.
  """
  stockSettings: StockSettingsInput
  ...
  orderSettings: OrderSettingsInput
```
#### OrderSettings
The `OrderSettings` type already exists:
```graphql
[...]

PR #11417, opened 2022-12-08
Move orderSetting to Channel type

kadewu:
Fix https://github.com/saleor/saleor/issues/11337

After successful merging, and 3.12 release, another PR for 3.13 will be prepared to drop fields from database.

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ All notable, unreleased changes to this project will be documented in this file.
 ### Breaking changes
 
 ### GraphQL API
+- Move `orderSettings` query to `Channel` type - #11417 by @kadewu:
+  - Mutation `channelCreate` and `channelUpdate` have new `orderSettings` input.
+  - Query `orderSettings` will be marked as depricated.
```
korycins, line 12:
```
Deprecate `orderSettings` query. Use `channel` query.
Deprecate `orderSettingsUpdate` mutation. Use `channelUpdate`.
```
It is already marked as a deprecated

saleor/channel/migrations/0006_order_settings_per_channel.py
```diff
@@ -0,0 +1,41 @@
...
+            model_name="channel",
+            name="automatically_confirm_all_new_orders",
+            field=models.BooleanField(default=True),
+        ),
+        migrations.AddField(
+            model_name="channel",
+            name="automatically_fulfill_non_shippable_gift_card",
+            field=models.BooleanField(default=True),
+        ),
+        migrations.RunPython(set_order_settings, migrations.RunPython.noop),
```
korycins, line 40: Running python script on just added fields should be in separate migration.

saleor/channel/migrations/0006_order_settings_per_channel.py
```diff
@@ -0,0 +1,41 @@
...
+from django.db import migrations, models
+
+
+def set_order_settings(apps, schema_editor):
+    SiteSettings = apps.get_model("site", "SiteSettings")
+    Channel = apps.get_model("channel", "Channel")
+
+    site_settings = SiteSettings.objects.first()
+
+    channels = Channel.objects.all()
```
korycins, line 12: No need to fetch channels and iterate over them
```
Channel.objects.update(
  automatically_confirm_all_new_orders=site_settings.automatically_confirm_all_new_orders,
  automatically_fulfill_non_shippable_gift_card=automatically_fulfill_non_shippable_gift_card
)
```

saleor/graphql/channel/mutations/channel_create.py
```diff
@@ -43,6 +63,11 @@ class ChannelInput(graphene.InputObjectType):
         + PREVIEW_FEATURE,
         required=False,
     )
+    order_settings = graphene.Field(
+        OrderSettingsInput,
+        description="The channel order settings" + ADDED_IN_39,
```
korycins, line 68: `ADDED_IN_310`. We already released 3.9

saleor/graphql/shop/schema.py
```diff
@@ -27,7 +29,10 @@ class ShopQueries(graphene.ObjectType):
     )
     order_settings = PermissionsField(
         OrderSettings,
-        description="Order related settings from site settings.",
+        description=(
+            "Order related settings from site settings."
+            f"{DEPRECATED_IN_3X_INPUT} Use `channelUpdate` instead."
```
korycins, line 34: The type fields have own field for marking it as a depreceated: `deprecation_reason`.

korycins: We should also add there the info, that returned `orderSettings` will be for a first channel in alphabetical order.

saleor/graphql/shop/schema.py
```diff
@@ -40,9 +45,16 @@ class ShopQueries(graphene.ObjectType):
     def resolve_shop(self, _info):
         return Shop()
 
-    @load_site_callback
-    def resolve_order_settings(self, _info, site):
-        return site.settings
+    def resolve_order_settings(self, _info):
+        channel = channel_models.Channel.objects.order_by("name").first()
```
korycins, line 49: I think `slug` instead of `name` would be better. It is unique and `name` can be a just empty string or something.

saleor/graphql/channel/mutations/channel_create.py
```diff
@@ -89,6 +114,13 @@ def clean_input(cls, info, instance, data, input_cls=None):
             cleaned_input["slug"] = slugify(slug)
         if stock_settings := cleaned_input.get("stock_settings"):
             cleaned_input["allocation_strategy"] = stock_settings["allocation_strategy"]
+        if order_settings := cleaned_input.get("order_settings"):
```
korycins, line 117: Calling update only for this field should be allowed with `MANAGE_ORDERS` permissions.

saleor/graphql/channel/mutations/channel_create.py
```diff
@@ -24,6 +30,20 @@ class StockSettingsInput(graphene.InputObjectType):
     )
 
 
+class OrderSettingsInput(graphene.InputObjectType):
```
korycins, line 33: I don't think, that we should add `required=True` to all fields here. You can want to modify only the single field.

saleor/graphql/shop/mutations.py
```diff
@@ -386,7 +388,10 @@ class Arguments:
         )
 
     class Meta:
-        description = "Update shop order settings."
+        description = (
+            "Update shop order settings."
+            f"{DEPRECATED_IN_3X_INPUT} Use `channelUpdate` instead."
```
korycins, line 393: We have specific label for mutation. `DEPRECATED_IN_3X_MUTATION`

korycins: We should also mention, that calling this mutation will update all values across all channels

saleor/graphql/shop/mutations.py
```diff
@@ -397,18 +402,23 @@ def perform_mutation(cls, _root, info, **data):
...
 
         if update_fields:
-            instance.save(update_fields=update_fields)
-        return OrderSettingsUpdate(order_settings=instance)
+            channel_models.Channel.objects.bulk_update(
+                channels,
+                list(update_fields.keys()),
+            )
+        for field in FIELDS:
+            update_fields.setdefault(field, True)
```
korycins, line 420: I don't think that this will be truth for all cases. If you update single value, the second one could be set to `False`.
I would propose rebuild `perform_mutation` to something like this one:
```    @classmethod
        fields_to_update = {}
        if "automatically_confirm_all_new_orders" in data:
            fields_to_update["automatically_confirm_all_new_orders"] = data["automatically_confirm_all_new_orders"]
        if "automatically_fulfill_non_shippable_gift_card" in data:
            fields_to_update["automatically_fulfill_non_shippable_gift_card"] = data["automatically_fulfill_non_shippable_gift_card"]
        Channel.objects.update(**fields_to_update)
        first_channel = Channel.objects.order_by("slug").first()

        return OrderSettingsUpdate(order_settings=<<values-from-first-channel>>)
```
Also, would be worth to mention in the `order_settings = graphene.Field(OrderSettings, description="Order settings.")` (line 383), that the field will return the order settings for the first channel in alpahbetical order for slug (same approach as we have for `orderSettings` query)

korycins (commented):
I think, the migration for dropping order settings from SiteSettings is missing.

saleor/graphql/shop/schema.py
```diff
@@ -27,7 +29,14 @@ class ShopQueries(graphene.ObjectType):
...
     order_settings = PermissionsField(
         OrderSettings,
-        description="Order related settings from site settings.",
+        description=(
+            "Order related settings from site settings."
+            "\n Returned `orderSettings` will be first `channel` in "
+            "alphabetical order."
+        ),
+        deprecation_reason=(
+            f"{DEPRECATED_IN_3X_FIELD} " "Use the `channel` field instead."
```
korycins, line 38: f"{DEPRECATED_IN_3X_FIELD} Use the `channel` field instead."
No need two `"` inside the string.
Also it should be something like. 
Use the `channel` query to fetch the `orderSettings` field instead.

saleor/graphql/channel/mutations/channel_create.py
```diff
@@ -24,6 +30,20 @@ class StockSettingsInput(graphene.InputObjectType):
...
+    automatically_confirm_all_new_orders = graphene.Boolean(
+        required=False,
+        description="When disabled, all new orders from checkout "
+        "will be marked as unconfirmed. When enabled orders from checkout will "
+        "become unfulfilled immediately.",
```
korycins, line 38: Would be worth to mention that by default it will be set to `True`/`False`.
Same for `automatically_fulfill_non_shippable_gift_card`

saleor/graphql/channel/mutations/channel_update.py
```diff
@@ -51,10 +51,16 @@ class Arguments:
         )
 
     class Meta:
-        description = "Update a channel."
+        description = (
+            "Update a channel.\n\n"
+            "Requires one of the following permissions: MANAGE_CHANNELS.\n"
```
korycins, line 56: @maarcingebala Any suggestion to make the description more readable?

saleor/graphql/channel/tests/mutations/test_channel_update.py
```diff
@@ -701,3 +714,39 @@ def test_channel_update_mutation_duplicated_warehouses(
     assert errors[0]["field"] == "warehouses"
     assert errors[0]["code"] == ChannelErrorCode.DUPLICATED_INPUT_ITEM.name
     assert errors[0]["warehouses"] == [add_warehouse]
+
+
+def test_channel_update_order_settings_manage_orders(
```
korycins, line 719: Can you also add a test with manage_orders and input that contains more that single field to update? (PermissionDenied)

korycins: Also, separate test for calling the mutation as an App

saleor/graphql/channel/tests/queries/test_channel.py
```diff
@@ -19,18 +19,28 @@
...
+                automaticallyFulfillNonShippableGiftCard
+            }
         }
     }
 """
 
 
-def test_query_channel_as_staff_user(staff_api_client, channel_USD):
+def test_query_channel_as_staff_user(
+    permission_manage_channels, staff_api_client, channel_USD
```
korycins, line 32: IMO, we shouldn't extend this test but create the new one. Previously this test was used that staff user without permissions has access to basic channel data. Now, we're attaching `MANAGE_ORDERS`.

saleor/graphql/channel/tests/queries/test_channel.py
```diff
@@ -44,15 +54,31 @@ def test_query_channel_as_staff_user(staff_api_client, channel_USD):
...
+        == channel_USD.automatically_confirm_all_new_orders
+    )
+    assert (
+        channel_data["orderSettings"]["automaticallyFulfillNonShippableGiftCard"]
+        == channel_USD.automatically_fulfill_non_shippable_gift_card
+    )
 
 
-def test_query_channel_as_app(app_api_client, channel_USD):
+def test_query_channel_as_app(
```
korycins, line 67: Same as above

saleor/graphql/channel/tests/queries/test_channel.py
korycins: Also, separate tests for staff/app that will confirm that without `MANAGE_ORDERS` we can't fetch `orderSettings` would be worth to have

saleor/graphql/channel/types.py
```diff
@@ -235,6 +248,15 @@ class Channel(ModelObjectType):
             AuthorizationFilters.AUTHENTICATED_STAFF_USER,
         ],
     )
+    order_settings = PermissionsField(
+        OrderSettings,
+        description="Channel order settings." + ADDED_IN_39,
```
korycins, line 253: ADDED_IN_310

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ All notable, unreleased changes to this project will be documented in this file.
 ### Breaking changes
 
 ### GraphQL API
+- Move `orderSettings` query to `Channel` type - #11417 by @kadewu:
+  - Mutation `channelCreate` and `channelUpdate` have new `orderSettings` input.
+  - Deprecate `orderSettings` query. Use `channel` query instead.
```
korycins, line 12: Use `channel.orderSettings` ...

saleor/graphql/channel/types.py
```diff
@@ -144,6 +149,14 @@ class Meta:
         )
 
 
+class OrderSettings(ObjectType):
+    automatically_confirm_all_new_orders = graphene.Boolean(required=True)
+    automatically_fulfill_non_shippable_gift_card = graphene.Boolean(required=True)
+
+    class Meta:
+        description = "Order related settings from site settings."
```
korycins, line 157: Maybe something similar as we have for stock allocation?
```
    class Meta:
        description = (
            "Represents the channel stock settings."
        )
```

saleor/graphql/shop/mutations.py
```diff
@@ -386,7 +393,12 @@ class Arguments:
         )
 
     class Meta:
-        description = "Update shop order settings."
+        description = (
+            "Update shop order settings across all channels"
+            "\nReturned `orderSettings` will be for first `channel` "
```
korycins, line 398: Maybe something the same as for `shop.orderSettings` query:
```
Returns orderSettings for the first channel in alphabetical order.
```

saleor/channel/migrations/0006__auto_20221208_0900.py
```diff
@@ -0,0 +1,39 @@
...
+        ("channel", "0005_channel_allocation_strategy"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="channel",
+            name="automatically_confirm_all_new_orders",
+            field=models.BooleanField(default=True),
+        ),
+        migrations.RunSQL(
```
fowczarek, line 18: We should add comments with a description that we are adding the database default value to not break backward database compatibility.

saleor/graphql/channel/mutations/channel_update.py
```diff
@@ -82,9 +88,27 @@ def clean_input(cls, info, instance, data, input_cls=None):
...
 
         return cleaned_input
 
+    @classmethod
+    def check_permissions(cls, context, permissions=None, **data):
+        permissions = [ChannelPermissions.MANAGE_CHANNELS]
+
+        # when updating only orderSettings allow MANAGE_ORDERS permission
+        input = data["data"]["input"]
+        if "order_settings" in input and len(input) == 1:
```
fowczarek, line 107: Why `and len(input) == 1`? 
If I have dict like below that mean Saleor shouldn't require `MANAGE_ORDERS`? 
```
input = {
    "stock_settings": ...,
    "order_settings":...,
}
````

Comment (fowczarek) [reply saleor/graphql/channel/mutations/channel_update.py:107]: If I'm right you should add a unit test for this situation.

File: saleor/graphql/shop/mutations.py [hunk 23]
```diff
@@ -397,18 +409,23 @@ def perform_mutation(cls, _root, info, **data):
...
-                setattr(instance, field, value)
-                update_fields.append(field)
+            if field in data["input"]:
+                update_fields[field] = data["input"][field]
 
         if update_fields:
-            instance.save(update_fields=update_fields)
-        return OrderSettingsUpdate(order_settings=instance)
+            channel_models.Channel.objects.update(**update_fields)
+        channel = channel_models.Channel.objects.order_by("slug").first()
```
fowczarek, line 419: I'm not sure we should update only one channel in this situation. We shouldn't guess in which channel those changes should be applied. 
IMO if we want to leave this behavior as deprecated then these mutations should change those values in all channels or require a channel to update.

saleor/graphql/shop/schema.py
```diff
@@ -40,9 +50,16 @@ class ShopQueries(graphene.ObjectType):
     def resolve_shop(self, _info):
         return Shop()
 
-    @load_site_callback
-    def resolve_order_settings(self, _info, site):
-        return site.settings
+    def resolve_order_settings(self, _info):
+        channel = channel_models.Channel.objects.order_by("slug").first()
```
fowczarek, line 54: Also not sure that we should return this value from the first channel ordered by slug. 
Maybe better to fetch the first active channel ordered by pk?
At lest we should filter out deactivated channels.

saleor/site/migrations/0035_sitesettings_separate_state_remove_order_settings.py
```diff
@@ -0,0 +1,65 @@
...
+# Database fields in this migration are changed to null to allow adding new rows to table.
+# To reverse that operation, we will set all values for `SiteSettings` regarding the
+# first channel in alphabetical order.
+
+
+def update_order_settings_values_to_match_first_channel(apps, schema_editor):
+    SiteSettings = apps.get_model("site", "SiteSettings")
+    Channel = apps.get_model("channel", "Channel")
+
+    channel = Channel.objects.order_by("slug").first()
```
fowczarek, line 17: Still not sure that we should use this channel as default.
Maybe better to fetch the first active channel ordered by `pk`?

fowczarek (commented):
Missing permission check on `ChannelCreate` mutation we should use the same permission check as in `ChannelUpdate`. I think it isn't ok to be able to set these values in `ChannelCreate` and not able to update if the requestor has only `MANAGE_CHANNELS` permission.

saleor/site/migrations/0035_sitesettings_separate_state_remove_order_settings.py
korycins: @fowczarek Yeah, agree with `isActive`. We should add them to all deprecated resolvers also (we use there a first channel also). 
It is filtered by `slug` to keep the same way of picking the channel as we do for deprecated SiteSettings resolvers.

saleor/graphql/channel/mutations/channel_update.py
kadewu: Yes, `MANAGE_ORDERS` permission can be used `only` when we are updating `order_settings`.
And I already added tests for this case.

saleor/graphql/shop/mutations.py
kadewu: But, we do update all channels here. Or I'm missing something?
I mean, the line: `channel_models.Channel.objects.update(**update_fields)` is updating every channel.

kadewu: Getting first channel is only to simplify logic in getting the value to return in mutation.
Note: The logic of getting first channel to show the `orderSettings` will be updated the same way as you are sugesting (that it should be first active channel)

kadewu:
> Missing permission check on `ChannelCreate` mutation we should use the same permission check as in `ChannelUpdate`. I think it isn't ok to be able to set these values in `ChannelCreate` and not able to update if the requestor has only `MANAGE_CHANNELS` permission.

That's some argument. But we cannot allow user to create `channel` with only `MANAGE_ORDER` permission.
@korycins What're your thoughts on this?

saleor/graphql/shop/mutations.py
fowczarek: My mistake. The code looks good in this place.

saleor/site/migrations/0035_sitesettings_separate_state_remove_order_settings.py
fowczarek: I think we shouldn't base it on a dynamic field because the first query could return `Channel(slug="ab")` and after a few seconds, the first channel returned by this query could be `Channel(slug="aa")`. I think we should always base on pk in this kind of situation.

fowczarek:
> That's some argument. But we cannot allow user to create channel with only MANAGE_ORDER permission.
@korycins What're your thoughts on this?

I think we should require 
- Manage_channels when OrderSettings is not provided to create channel mutation.
- Manage_channels and Manage_orders when OrderSettings in mutation input.

saleor/graphql/shop/schema.py
korycins: We want to have a slug here as we add this info in the field description. Finding first channel for user sorted by slug is easier than by PK.

korycins:
> > That's some argument. But we cannot allow user to create channel with only MANAGE_ORDER permission.
> > @korycins What're your thoughts on this?
> 
> I think we should require
> 
> * Manage_channels when OrderSettings is not provided to create channel mutation.
> * Manage_channels and Manage_orders when OrderSettings in mutation input.

@fowczarek   I assume that the proposed changes are related only to the `create` mutation.
Personally, I am not fully sure of this. Even if you don't provide the `orderSettings` in the mutation input,
you will still get the default value, attached by Saleor. So you will have there some values assigned, even if you don't have permission to provide it directly from the API. Personally, for me, it is still handling the channel settings (as orderSettings are a part of the channel settings), and creating/updating the channel should be allowed with only `manage_channels`

@maarcingebala @patrys  What is your opinion here?

saleor/site/migrations/0035_sitesettings_separate_state_remove_order_settings.py
```diff
@@ -0,0 +1,65 @@
...
+    channel = Channel.objects.filter(is_active=True).order_by("slug").first()
+
+    SiteSettings.objects.update(
+        automatically_confirm_all_new_orders=(
+            channel.automatically_confirm_all_new_orders
+        ),
+        automatically_fulfill_non_shippable_gift_card=(
+            channel.automatically_fulfill_non_shippable_gift_card
+        ),
+    )
```
IKarbowiak, line 26: I don't get why we need that as those fields (`automatically_confirm_all_new_orders` and `automatically_fulfill_non_shippable_gift_card`) are removed in this migration.

saleor/graphql/channel/mutations/channel_update.py
```diff
@@ -84,9 +90,27 @@ def clean_input(cls, info: ResolveInfo, instance, data, **kwargs):
...
+                "automatically_fulfill_non_shippable_gift_card"
+            ] = order_settings["automatically_fulfill_non_shippable_gift_card"]
 
         return cleaned_input
 
+    @classmethod
+    def check_permissions(cls, context, permissions=None, **data):
+        permissions = [ChannelPermissions.MANAGE_CHANNELS]
+
+        # when updating only orderSettings allow MANAGE_ORDERS permission
```
IKarbowiak, line 107: This comment is a little bit incomprehensible. Could you try to rewrite it?

saleor/site/migrations/0035_sitesettings_separate_state_remove_order_settings.py
kadewu: That function is provided to enable unapplying the migration (it's used as reverse code in line 51)

saleor/graphql/channel/mutations/channel_update.py
kadewu: Sure thing. What do you think about:
`Permission MANAGE_ORDERS or MANAGE_CHANNELS is required to update orderSettings.
MANAGE_ORDERS can be used only when channelUpdate mutation will get OrderSettingsInput and not any other fields. Otherwise needed permission is MANAGE_CHANNELS`

saleor/site/migrations/0035_sitesettings_separate_state_remove_order_settings.py
IKarbowiak: Thanks, I missed it!

saleor/graphql/channel/mutations/channel_update.py
IKarbowiak: Sounds good to me 👍

korycins:
> another PR for 3.11 will be prepared to drop fields from database.

@kadewu, lets create a gh issue for this as we don't want to miss it

korycins:
@kadewu, there is a PR https://github.com/saleor/saleor/pull/11645 that points to 3.9. I think that the most cleanable way would be creating a port of this changes that should be merge to the main (and extracting it from this PR as it is duplication of the code that need to be checked in the review). So we will have 3 PRs:
3.9 <- move-order-settings-to-channel-3.9
main <- move-order-settings-to-channel <- move-order-settings-to-channel
(where arrow shows the base direction of mergin [base-branch & feature-branch])

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ All notable, unreleased changes to this project will be documented in this file.
 ### Breaking changes
 
 ### GraphQL API
+- Move `orderSettings` query to `Channel` type - #11417 by @kadewu:
+  - Mutation `channelCreate` and `channelUpdate` have new `orderSettings` input.
+  - Deprecate `orderSettings` query. Use `channel.orderSetting` query instead.
```
maarcingebala, line 12:
```suggestion
  - Deprecate `orderSettings` query. Use `Channel.orderSettings` query instead.
```

saleor/graphql/channel/types.py
```diff
@@ -251,6 +264,15 @@ class Channel(ModelObjectType[models.Channel]):
             AuthorizationFilters.AUTHENTICATED_STAFF_USER,
         ],
     )
+    order_settings = PermissionsField(
+        OrderSettings,
+        description="Channel order settings." + ADDED_IN_310,
```
maarcingebala, line 269:
```suggestion
        description="Channel-specific order settings." + ADDED_IN_310,
```

saleor/graphql/channel/types.py
```diff
@@ -160,7 +165,15 @@ class Meta:
         )
 
 
-class Channel(ModelObjectType[models.Channel]):
+class OrderSettings(ObjectType):
+    automatically_confirm_all_new_orders = graphene.Boolean(required=True)
+    automatically_fulfill_non_shippable_gift_card = graphene.Boolean(required=True)
+
+    class Meta:
+        description = "Represents the channel order settings."
```
maarcingebala, line 173:
```suggestion
        description = "Represents the channel-specific order settings."
```

saleor/graphql/channel/types.py
```diff
@@ -160,7 +165,15 @@ class Meta:
         )
 
 
-class Channel(ModelObjectType[models.Channel]):
+class OrderSettings(ObjectType):
+    automatically_confirm_all_new_orders = graphene.Boolean(required=True)
+    automatically_fulfill_non_shippable_gift_card = graphene.Boolean(required=True)
```
maarcingebala, line 170: We should also provide descriptions here, it can be the same if we have it somewhere else (I think we have in inputs).

saleor/graphql/channel/mutations/channel_update.py
maarcingebala: I think it's ok.

saleor/graphql/channel/mutations/channel_create.py
```diff
@@ -90,6 +115,13 @@ def clean_input(cls, info: ResolveInfo, instance, data, **kwargs):
             cleaned_input["slug"] = slugify(slug)
         if stock_settings := cleaned_input.get("stock_settings"):
             cleaned_input["allocation_strategy"] = stock_settings["allocation_strategy"]
+        if order_settings := cleaned_input.get("order_settings"):
+            cleaned_input["automatically_confirm_all_new_orders"] = order_settings[
+                "automatically_confirm_all_new_orders"
+            ]
```
maarcingebala, line 121: Don't you need to use `.get` on the dict here? These two input fields are not required, so I think the user might provide only one, and then one of these assignments would break? Same in the update mutation.

saleor/graphql/shop/mutations.py
```diff
@@ -390,7 +397,12 @@ class Arguments:
         )
 
     class Meta:
-        description = "Update shop order settings."
+        description = (
+            "Update shop order settings across all channels"
+            "\nReturns `orderSettings` for the first `channel` "
+            f"in alphabetical order.{DEPRECATED_IN_3X_MUTATION}"
```
maarcingebala, line 403: See other usages of `DEPRECATED_IN_3X_MUTATION`. We should pass it as `deprecation_reason` in the field definition like this:
```
order_settings_update = OrderSettingsUpdate.Field(
    deprecation_reason=DEPRECATED_IN_3X_MUTATION
)
```

saleor/graphql/shop/mutations.py
```diff
@@ -401,18 +413,32 @@ def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
...
-            if value is not None:
-                setattr(instance, field, value)
-                update_fields.append(field)
+            if field in data["input"]:
+                update_fields[field] = data["input"][field]
 
         if update_fields:
-            instance.save(update_fields=update_fields)
-        return OrderSettingsUpdate(order_settings=instance)
+            channel_models.Channel.objects.update(**update_fields)
```
maarcingebala, line 422:
```suggestion
            channel_models.Channel.objects.update(**update_fields)
            
```

maarcingebala: Extra line for readability.

saleor/graphql/shop/mutations.py
```diff
@@ -390,7 +397,12 @@ class Arguments:
         )
 
     class Meta:
-        description = "Update shop order settings."
+        description = (
+            "Update shop order settings across all channels"
+            "\nReturns `orderSettings` for the first `channel` "
```
maarcingebala, line 402: We don't need new-line characters in the descriptions, they usually go into one line unless you want to write some really specific and longer descriptions.

saleor/graphql/shop/mutations.py
```diff
@@ -401,18 +413,32 @@ def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
...
-            instance.save(update_fields=update_fields)
-        return OrderSettingsUpdate(order_settings=instance)
+            channel_models.Channel.objects.update(**update_fields)
+        channel = (
+            channel_models.Channel.objects.filter(is_active=True)
+            .order_by("slug")
+            .first()
+        )
+        if channel is None:
+            raise ValidationError(
```
maarcingebala, line 429: Shouldn't we put this block at the start of the mutation? If there is no channel then there is no point in doing all these actions that happen before, right?

saleor/graphql/shop/schema.py
```diff
@@ -40,9 +50,22 @@ class ShopQueries(graphene.ObjectType):
     def resolve_shop(self, _info):
         return Shop()
 
-    @load_site_callback
-    def resolve_order_settings(self, _info, site):
-        return site.settings
+    def resolve_order_settings(self, _info):
+        channel = (
+            channel_models.Channel.objects.filter(is_active=True)
```
maarcingebala, line 55: Shouldn't it use dataloader?

saleor/site/migrations/0035_sitesettings_separate_state_remove_order_settings.py
```diff
@@ -0,0 +1,65 @@
...
+
+def update_order_settings_values_to_match_first_channel(apps, schema_editor):
+    SiteSettings = apps.get_model("site", "SiteSettings")
+    Channel = apps.get_model("channel", "Channel")
+
+    channel = Channel.objects.filter(is_active=True).order_by("slug").first()
+
+    SiteSettings.objects.update(
+        automatically_confirm_all_new_orders=(
+            channel.automatically_confirm_all_new_orders
```
maarcingebala, line 21: What if there is no channel?

maarcingebala:
> 

I think it makes sense what you're suggesting @korycins and we could actually only rely on MANAGE_CHANNELS permission.

saleor/graphql/shop/schema.py
kadewu: I'm not sure about using dataloader here. As I need to find proper slug, and maybe no need to load all channels.
Or do you mean to hit database twice, first to find `slug`, second time via dataloader load the channel data?

korycins: Right, we are fetching the first channel ordered by `slug`. I don't think that we are able to use any existing channel dataloader here.
@fowczarek WDYT?

fowczarek: IMO we could add `ChannelBySlug`  dataloader. In `then` we could checkout if it's active or not.

kadewu: I'm not sure if I understand the proposed solution. Cause we still need to know the `slug` to call `ChannelBySlug` dataloader. So you suggest to hit database for that slug and then reuse this dataloader?

fowczarek: No, you're right. This is the deprecated field, we could use db query here.
I didn't see any reason for multiple times fetching this field in one query.

fowczarek (approved):

saleor/graphql/shop/schema.py
fowczarek: I am still not sure about sorting by `slug` not `pk`. We should always display results for the same objects. When someone creates a channel with slug `aaa` then we will return these fields for this channel, not the previous ones.

korycins: @fowczarek @kadewu  Yeah, and this is an expected way, as we want to give you a possibility to create such channel, only to have an impact on these settings. Also I don't think that we should worry about this as described case can happen only when you're mix-in old & new fields.

korycins: So we're marking the conversation as resolved, @fowczarek @kadewu  agree?

saleor/graphql/shop/mutations.py
```diff
@@ -390,7 +392,11 @@ class Arguments:
         )
 
     class Meta:
-        description = "Update shop order settings."
+        description = (
+            "Update shop order settings across all channels "
+            "Returns `orderSettings` for the first `channel` in alphabetical order. "
```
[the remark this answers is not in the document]
kadewu: It is marked in `graphql/shop/schema.py` Line 88:
`order_settings_update = OrderSettingsUpdate.Field(
     deprecation_reason=DEPRECATED_IN_3X_MUTATION
)`

szczecha (approved):

saleor/graphql/shop/mutations.py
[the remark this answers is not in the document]
korycins: @kadewu  Yeah, but input mutation vs type field it is not something always related. Without marking this as a deprecated, we can't drop it in 4.0. 
If we're deprecating any interface we should always deprecate type and specific input fields/mutation related to it.

korycins (approved):

Issue #11227, opened 2022-11-17
Fix IntegrityError bug caused by task with payload deletion


jakubkuc:
There is a corner case that causes IntegrityError when a user is trying to delete Webhook. 
#### UPDATE:
Through more tests, we were able to reproduce this error. It happens when Webhook delivery fails and is queueing retries and at the same time we try to delete the webhook. 
To prevent that, the webhook should be deactivated in the`WebhookDelete` mutation before deletion.

PR #11475, opened 2022-12-13
Deactivate webhook before deletion

jakubkuc:
Discussion and reviews for this cherry-picks are in #11239 

I want to merge this change because it prevents issue where using WebhookDelete mutation while tasks that create EventDeliveryAttempt are queued resulted in IntegrityError.

The most simple resolution was to deactivate Webhook before deleting it in mutation. send_webhook_request_async task is checking if webhook is_active so this should prevent this situation.

Tasks:
I want to add tests after a discussion regarding edge-case.

Things to discuss:
⚠️ There is a small window for this solution to fail - when Webhook deactivation will happen between the is_active check and attempt creating/updating in send_webhook_request_async task. I'm not quite sure how to handle this case. Should I add more webhook activity checks inside send_webhook_request_async task or maybe catch this edge-case exception to fail gracefully? Any ideas?

Second thing: Do we want to do the sam analogically for deleting App? (deactivate webhook in app deletion)

https://github.com/saleor/saleor/issues/11227

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/webhook/mutations.py
+++ b/saleor/graphql/webhook/mutations.py
@@ -233,7 +233,8 @@ class WebhookDelete(ModelDeleteMutation):
         id = graphene.ID(required=True, description="ID of a webhook to delete.")
 
     class Meta:
-        description = "Deletes a webhook subscription."
+        description = "Deletes a webhook subscription. Deletion might fail due to " \
+                      "webhook delivery in progress. In this case deactivates webhook"
         model = models.Webhook
         object_type = Webhook
         permissions = (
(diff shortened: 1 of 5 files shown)


saleor/graphql/webhook/mutations.py
```diff
@@ -232,7 +233,10 @@ class Arguments:
         id = graphene.ID(required=True, description="ID of a webhook to delete.")
 
     class Meta:
-        description = "Deletes a webhook subscription."
+        description = (
+            "Deletes a webhook subscription. Deletion might fail due to "
+            "webhook delivery in progress. In this case deactivates webhook"
```
maarcingebala, line 238: Redacted it to be more descriptive. This is our docs so needs to be as clear as possible to the users. 

```suggestion
            "Delete a webhook. Before the deletion, the webhook is deactivated to "
            "pause any deliveries that are already scheduled. The deletion might fail "
            "if delivery is in progress. In such a case, the webhook is not deleted "
            "but remains deactivated."
```

maarcingebala: Will need schema regeneration after it's updated.

korycins (approved):

Issue #11456, opened 2022-12-09
[RFC] Webhooks dry run


maarcingebala:
We want to improve the experience of building and using webhooks, by giving the ability to:
- dry run webhook events for testing purposes
- run webhook events manually with an actual delivery

Below is the proposed scope of changes.

## 1. Dry run webhook events

Add a mutation that performs a dry run of a webhook event. Mutation should accept subscription query and an object ID to serialize. As a result it should return a JSON response with serialized data, based on given subscription query. Reuse functions that are already used by subscription webhooks to generate payload.

- Add validation to check if given object ID matches the event, e.g. `OrderCreated` event requires an order ID and should fail if other ID is provided.
- Perform the dry-run in the context of currently logged in staff user.

### Schema

```graphql
type Mutation {
  """Performs a dry run of a webhook event."""
  webhookDryRun(query: String!, objectId: ID!): WebhookDryRun
}

type WebhookDryRun {
  errors: [WebhookDryRunError!]!
  """JSON payload that would be sent out to webhook's target URL."""
  payload: JSONString
}

type WebhookDryRunError {
  field: String
  code: WebhookDryRunErrorCode!
}

enum WebhookDryRunErrorCode {
  """Triggered when object ID doesn't match with the event type."""
  INVALID_ID
}
```

Example:

```graphql
mutation {
  webhookDryRun(
    objectId: "UHJvZHVjdDoxNTI=",
    query: "subscription { event { ... on ProductUpdated { product { id name description } } } }"
    ) {
      errors {
        field
        code
        message
      }
      payload
    }
}
```

Note: Dry run should be support only for subscription webhooks.

## 2. Trigger webhook events manually

Add a mutation that allows to trigger and deliver a webhook for given object ID. 

### Schema

```graphql
type Mutation {
  """Trigger a webhook event."""
  webhookTrigger(webhookId: ID!, objectId: ID!): WebhookTrigger
}

type WebhookTrigger {
  errors: [WebhookTriggerError!]!
  delivery: EventDelivery
}

type WebhookTriggerError {
  field: String
  message: String
  code: WebhookTriggerErrorCode!
}

enum WebhookTriggerErrorCode {
  """Triggered when object ID doesn't match with the event type."""
  INVALID_ID
}
```

Example:

```graphql
mutation {
  webhookTrigger(webhookId: "V2ViaG9vazox", objectId: "UHJvZHVjdDoxNTI=") {
    errors {
      field
      code
      message
    }
    delivery {
      id
[...]

PR #11548, opened 2022-12-20
Dry run webhook events.

zedzior:
https://github.com/saleor/saleor/issues/11456

I want to merge this change, because there is a need to dry run webhook events for testing purposes and to increase developer experience.

This PR adds `WebhookDryRun` mutation, that performs a dry run of a webhook event. Mutation accepts subscription query and an object ID to serialize. As a result it returns a JSON response with serialized data, based on given subscription query.

**Schema changes**:
```graphql
type Mutation {
  """Performs a dry run of a webhook event."""
  webhookDryRun(query: String!, objectId: ID!): WebhookDryRun
}

type WebhookDryRun {
  errors: [WebhookDryRunError!]!
  """JSON payload that would be sent out to webhook's target URL."""
  payload: JSONString
}

type WebhookDryRunError {
  field: String
  code: WebhookDryRunErrorCode!
}

enum WebhookDryRunErrorCode {
  GRAPHQL_ERROR
  NOT_FOUND
  INVALID_ID
  MISSING_PERMISSION
  UNABLE_TO_PARSE
  TYPE_NOT_SUPPORTED
}
```
**Example**:
```graphql
mutation webhookDryRun {
  webhookDryRun(
    objectId: "UHJvZHVjdDoxNTI=",
    query: "subscription { event { ... on ProductUpdated { product { id name description } } } }"
    ) {
      errors {
        field
        code
        message
      }
      payload
    }
}
```

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/webhook/utils.py
```diff
@@ -0,0 +1,21 @@
+from graphene.utils.str_converters import to_snake_case
+from graphql import parse, print_ast
+
+
+def get_event_type_from_subscription(query: str):
```
zedzior, line 5: Do you think, the function is robust enough?

saleor/graphql/webhook/mutations.py
```diff
@@ -301,3 +308,69 @@ def perform_mutation(cls, _root, info, **data):
...
+class WebhookDryRun(BaseMutation):
+    payload = JSONString(
+        description="JSON payload, that would be sent out to webhook's target URL."
+    )
+
+    class Arguments:
+        input = WebhookDryRunInput(
+            description="Fields required to perform dry run of a webhook event.",
+            required=True,
+        )
```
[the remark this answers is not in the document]
maarcingebala: Hmm maybe actually we don't. It will be needed for the mutation `webhookTrigger` for sure to know for which webhook to send the request (target URL etc.). I would only check what happens when running dry-run using a query with fields you that don't have permission to fetch. E.g. the app might have only MANAGE_ORDERS permission, but in the query you might ask for some restricted product-level fields that require MANAGE_PRODUCT. I would check if the payload generation function handles that and returns the information that you cannot query those fields.

saleor/graphql/webhook/mutations.py
```diff
@@ -301,3 +308,69 @@ def perform_mutation(cls, _root, info, **data):
...
+                code=WebhookDryRunErrorCode.GRAPHQL_ERROR,
+            )
+
+        # TODO check if object_id match event_type
+        # if ... :
+        #     raise_validation_error(
+        #         field="objectId",
+        #         message="ObjectId doesn't match event type.",
+        #         code=WebhookDryRunErrorCode.INVALID_ID,
+        #     )
```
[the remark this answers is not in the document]
maarcingebala: I don't recall, @korycins is there something like that? We need to validate that if you passed an ID and event type ORDER_CREATED, only Order IDs are allowed. Maybe we need to add some map.

saleor/graphql/webhook/utils.py
maarcingebala: We need to check if it works well with fragments (I think it does, but worth checking) e.g:

```
fragment OrderFragment on Order {
  id
	number
  lines {
    id
  }
}

subscription {
  event {
    ...on OrderCreated {
      order {
        ...OrderFragment
      }
    }
  }
}
```

saleor/graphql/webhook/mutations.py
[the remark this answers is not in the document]
korycins: @zedzior  @maarcingebala  Hhm... I don't think that we have such logic. 
I think that map is a good idea but we need to do it cleverly to not have a situation, where someone who adds a new webhook subscription will forget about extending the map.
Second thing: I am not sure if providing a single ID will be enough. Some events have more than a single object. For example `ProductVariantOutOfStock`  has fields for `variant` and `warehouse`. Even if we don't want to cover this event now, we need to have an interface that will handle it.

saleor/graphql/webhook/utils.py
korycins: We could also cover the case like:
```graphql
fragment NotUsedEvents on Event {
  ... on OrderCreated {
    order {
      id
    }
  }
}

subscription {
  event {
    ... on OrderUpdated {
      order {
        id
      }
    }
  }
}
```
 or the case when `event` is a fragment
 
 ```graphql
 fragment EventFragment on Event{
  ...
}
subscription {
  event{
   ...EventFragment    
  }
}

```

File: saleor/graphql/webhook/mutations.py [hunk 4]
```diff
@@ -301,3 +311,95 @@ def perform_mutation(cls, _root, info, **data):
...
+        return event_type, object, app, query
+
+    @classmethod
+    def perform_mutation(cls, _root, info, **data):
+        event_type, object, app, query = cls.validate_input(info, **data)
+        payload = None
+        if all([event_type, object, app, query]):
+            request = initialize_request()
+            payload = generate_payload_from_subscription(
+                event_type, object, query, request, app
```
maarcingebala, line 402: As we discussed on Slack with @patrys we want to run dry-run as the staff user and not as the app. Then we would have to remove the `app` from this call I think.

saleor/graphql/webhook/mutations.py
```diff
@@ -301,3 +311,95 @@ def perform_mutation(cls, _root, info, **data):
...
+                field="query",
+                message=f"Event type: {event_name} is not defined in graphql schema.",
+                code=WebhookDryRunErrorCode.GRAPHQL_ERROR,
+            )
+
+        related_models = [
+            field.type.get_model().__name__
+            for field in event._meta.fields.values()
+            if hasattr(field, "type") and hasattr(field.type, "get_model")
+        ]
```
maarcingebala, line 371: We need to rethink this, for example event `ProductVariantOutOfStock` requires the `stock` object (see how `subscribable_object` is passed for webhooks).

saleor/graphql/webhook/mutations.py
```diff
@@ -301,3 +316,166 @@ def perform_mutation(cls, _root, info, **data):
...
+        payload = None
+        if all([event_type, object, query]):
+            request = initialize_request()
+            payload = generate_payload_from_subscription(
+                event_type, object, query, request
+            )
+        return WebhookDryRun(payload=payload)
+
+
+class WebhookTrigger(BaseMutation):
```
maarcingebala, line 398: Are we going to release it as part of this PR or a separate one? It would be better to move it to some other branch.

saleor/graphql/webhook/mutations.py
```diff
@@ -301,3 +316,166 @@ def perform_mutation(cls, _root, info, **data):
...
+            description="The ID of an object to serialize.", required=True
+        )
+
+    class Meta:
+        description = "Trigger a webhook event."
+        permissions = (AppPermission.MANAGE_APPS,)
+        error_type_class = WebhookTriggerError
+
+    @classmethod
+    def validate_input(cls, info, **data):
```
maarcingebala, line 413: A lot of common logic between this and dry-run; maybe we could extract some of it in a base class.

zedzior: Will fix it in webhook-manual-trigger branch (https://github.com/saleor/saleor/pull/11687)

CHANGELOG.md
```diff
@@ -22,6 +22,15 @@ All notable, unreleased changes to this project will be documented in this file.
 - Extend invoice object types with `Order` references - #11505 by @przlada
   - Add `Invoice.order` field
   - Add `InvoiceRequested.order`, `InvoiceDeleted.order` and `InvoiceSent.order` fields
+- Allow to mutate objects, by newly added `externalReference` field, instead of Saleor-assigned ID. Apply to following models: #11410 by @zedzior
```
korycins, line 25: Is it related to this change?

saleor/graphql/webhook/mutations.py
```diff
@@ -306,3 +316,80 @@ def perform_mutation(cls, _root, info: ResolveInfo, **data):
...
+            description="The subscription query that defines the webhook event and its "
+            "payload.",
+            required=True,
+        )
+        object_id = graphene.ID(
+            description="The ID of an object to serialize.", required=True
+        )
+
+    class Meta:
+        description = "Performs a dry run of a webhook event."
```
korycins, line 337: Shouldn't we add here FEATURE_PREVIEW and ADDED_IN_3X labels?

saleor/graphql/webhook/mutations.py
```diff
@@ -306,3 +316,80 @@ def perform_mutation(cls, _root, info: ResolveInfo, **data):
...
+    def validate_input(cls, info, **data):
+        query = data.get("query")
+        object_id = data.get("object_id")
+
+        event_type = get_event_type_from_subscription(query)
+        if not event_type:
+            raise_validation_error(
+                field="query",
+                message="Can't parse an event type from query.",
+                code=WebhookDryRunErrorCode.GRAPHQL_ERROR,
```
korycins, line 351: Can we add specific errors for each type of failing validation? 
Returning a generic ones, doesn't allow you to build better error handling on the frontend part (as all or most of the errors returns the same error code). 
So here we should have something like: `WebhookDryRunErrorCode.UNABLE_TO_PARSE`.

saleor/graphql/webhook/mutations.py
```diff
@@ -306,3 +316,80 @@ def perform_mutation(cls, _root, info: ResolveInfo, **data):
...
+                field="objectId",
+                message="ObjectId doesn't match event type.",
+                code=WebhookDryRunErrorCode.INVALID_ID,
+            )
+
+        object = cls.get_node_or_error(info, object_id, field="objectId")
+
+        if permission := WebhookEventAsyncType.PERMISSIONS.get(event_type):
+            codename = permission.value.split(".")[1]
+            user_permissions = [
```
korycins, line 375: this is related to above comment related to the permissions. This `if` condition takes into account only the user `permissions`

saleor/graphql/webhook/mutations.py
```diff
@@ -306,3 +316,80 @@ def perform_mutation(cls, _root, info: ResolveInfo, **data):
...
+            "payload.",
+            required=True,
+        )
+        object_id = graphene.ID(
+            description="The ID of an object to serialize.", required=True
+        )
+
+    class Meta:
+        description = "Performs a dry run of a webhook event."
+        permissions = (AppPermission.MANAGE_APPS,)
```
korycins, line 338: Shouldn't be authenticated_staff_user, instead of `MANAGE_APPS`? To generate dry payload for example for Order, you need to have manage_orders and manage_apps. I think that it would be enough to require authenticated staff user, as the resources are protected by proper permissions on the resovlers

saleor/graphql/webhook/mutations.py
```diff
@@ -306,3 +316,80 @@ def perform_mutation(cls, _root, info: ResolveInfo, **data):
...
+        error_type_class = WebhookDryRunError
+
+    @classmethod
+    def validate_input(cls, info, **data):
+        query = data.get("query")
+        object_id = data.get("object_id")
+
+        event_type = get_event_type_from_subscription(query)
```
korycins, line 346: This function returns the first event_type that is in the query, right? Maybe we should mention this in the description of the mutation, as user can provide a multiple events like `orderCreated`, `orderFullyPaid` etc in single query

korycins: Or maybe we should raise an exception when we find multiple events? 🤔

saleor/webhook/error_codes.py
```diff
@@ -8,3 +8,10 @@ class WebhookErrorCode(Enum):
     REQUIRED = "required"
     UNIQUE = "unique"
     DELETE_FAILED = "delete_failed"
+
+
+class WebhookDryRunErrorCode(Enum):
+    GRAPHQL_ERROR = "graphql_error"
+    NOT_FOUND = "not_found"
+    INVALID_ID = "invalid_id"
+    MISSING_PERMISSION = "missing_app_permission"
```
korycins, line 17: `missing_app_permission`? Is it correct? I thought that the dry run is called by the staff user, without using any existing app.

saleor/graphql/webhook/mutations.py
```diff
@@ -306,3 +316,80 @@ def perform_mutation(cls, _root, info: ResolveInfo, **data):
...
+        if not event:
+            event_name = event_type[0].upper() + to_camel_case(event_type)[1:]
+            raise_validation_error(
+                field="query",
+                message=f"Event type: {event_name} is not defined in graphql schema.",
+                code=WebhookDryRunErrorCode.GRAPHQL_ERROR,
+            )
+
+        model, _ = graphene.Node.from_global_id(object_id)
+        if model != WebhookEventAsyncType.ROOT_TYPE.get(event_type):
```
korycins, line 364: I think that we should first check if `event_type` exists in ROOT_TYPE. If not, raise an exception that the given type is not supported. We don't explicitly say which types are supported, so I think that could be usefull.

saleor/webhook/event_types.py
```diff
@@ -496,6 +496,115 @@ class WebhookEventAsyncType:
         OBSERVABILITY: AppPermission.MANAGE_OBSERVABILITY,
     }
 
+    # ROOT_TYPE mapping is needed by dry-run and manual trigger
+    # of the subscription webhooks events.
+    ROOT_TYPE = {
```
korycins, line 501: Is `ROOT_TYPE` correct name? It is a mapper for supported dry run webhooks, don't know how it is related with `ROOT`

saleor/graphql/webhook/mutations.py
zedzior: Can you give me an example of the query with multiple events you think about?

saleor/webhook/error_codes.py
zedzior: Good spot! Thanks

CHANGELOG.md
zedzior: Just found the comment has wrong indent and looks like sub-comment, so I just took it out. However, looks like rebase restored it. Thanks!

saleor/graphql/webhook/mutations.py
zedzior: You are right, thanks

saleor/graphql/webhook/subscription_types.py
```diff
@@ -1351,8 +1443,9 @@ def resolve_translation(root, _info: ResolveInfo):
         return translation
 
 
-class TranslationCreated(ObjectType, TranslationBase):
+class TranslationCreated(SubscriptionObjectType, TranslationBase):
     class Meta:
+        dry_run_model_name = "Translation"
```
zedzior, line 1448: This one will not work, because root type depends of model wich was translated, ie. `AttributeTranslation` instead of `Translation`. Should we skip it and treat like `TYPE_NOT_SUPPORTED` or write some additional logic?

saleor/graphql/webhook/subscription_types.py
```diff
@@ -1157,8 +1237,9 @@ def resolve_shipping_zone(root, _info: ResolveInfo, channel=None):
         return ChannelContext(node=shipping_method.shipping_zone, channel_slug=channel)
 
 
-class ShippingPriceCreated(ObjectType, ShippingPriceBase):
+class ShippingPriceCreated(SubscriptionObjectType, ShippingPriceBase):
     class Meta:
+        dry_run_model_name = "ShippingMethod"
```
zedzior, line 1242: This one currently doesn't work. `get_node_or_error` function returns error instead of node, due to:
`graphene_type = info.schema.get_type("ShippingMethod).graphene_type`
`getattr(graphene_type, "get_node", None) = None`
Do you know how to handle it? I see that we have 2 options for `ShippingMethod` like `PRICE_BASED` and `WEIGHT_BASE`. Maybe it is somehow related...

saleor/graphql/webhook/subscription_types.py
```diff
@@ -124,24 +124,27 @@ def resolve_address(root, _info: ResolveInfo):
         return address
 
 
-class AddressCreated(ObjectType, AddressBase):
+class AddressCreated(SubscriptionObjectType, AddressBase):
     class Meta:
+        dry_run_model_name = "Address"
```
zedzior, line 129: Let's discuss here, a naming for current `dry_run_model_name`

saleor/graphql/webhook/subscription_types.py
maarcingebala: For now I would skip it and consider adding support for it later.

saleor/graphql/webhook/subscription_types.py
maarcingebala: My suggestion is `root_type` or `resolver_root_type`.

saleor/graphql/webhook/subscription_types.py
korycins: AFAIK ShippingMethod is a dynamic type, that can not be related to the DB model. The shipping method can come from the external apps. In the middle we're creating the dataclass that stores the object. Maybe this is something that causes the issue here 🤔

saleor/graphql/webhook/mutations.py
```diff
@@ -306,3 +321,96 @@ def perform_mutation(cls, _root, info: ResolveInfo, **data):
         manager = get_plugin_manager_promise(info.context).get()
         manager.event_delivery_retry(delivery)
         return EventDeliveryRetry(delivery=delivery)
+
+
+class WebhookDryRun(BaseMutation):
```
maarcingebala, line 326: BTW we try to use the pattern that every mutation is a separate file, once we merge dry-run and manual-trigger, we could quickly refactor this file. Just a note.

saleor/graphql/webhook/subscription_types.py
korycins: `root_type` is ok, for attaching the model name to the subscription types. But we need something that will tell developers that this flag controls also enabling dry run and trigger. 
So, we should somehow connect `root_type`+ something that will tell developer that setting this value activates the dry run and trigger. In other case, there is a risk, that someone who is not familiar with this future, will treat a missing value for `root_type` as a bug in the code. Setting the comment is a workaround as I would prefer to have the variable name that explains everything.
If there is a possibility that some types will have only dry run or only trigger (probably a valid case for TransactionActionRequest), then maybe we should have a flags there?
```
class Meta:
    root_type = "Address"
    enable_dry_run = True
    enable_test_trigger = False
```
There is more lines but it is suuper clean and readable

saleor/graphql/webhook/tests/mutations/test_webhook_dry_run.py
```diff
@@ -0,0 +1,591 @@
...
+def test_webhook_dry_run_root_type(
+    superuser_api_client,
+    async_subscription_webhooks_with_root_objects,
+):
+    # given
+    query = WEBHOOK_DRY_RUN_MUTATION
+
+    for event_name, event_type in WEBHOOK_TYPES_MAP.items():
+        if not event_type._meta.enable_dry_run:
+            continue
```
zedzior, line 577: This gonna force user to add new webhook fixture, whenever new event with `enable_dry_run` flag will be created. But I think it's good

saleor/graphql/webhook/mutations.py
zedzior: Will be handled in separate PR

saleor/graphql/webhook/subscription_types.py
zedzior: Currently, ShippingPrice related events will be not supported

saleor/graphql/webhook/mutations.py
zedzior: Currently, a single event (the first if multiple provided) will be supported

maarcingebala (approved):

saleor/graphql/webhook/mutations.py
```diff
@@ -306,3 +321,100 @@ def perform_mutation(cls, _root, info: ResolveInfo, **data):
...
+            required=True,
+        )
+        object_id = graphene.ID(
+            description="The ID of an object to serialize.", required=True
+        )
+
+    class Meta:
+        description = (
+            "Performs a dry run of a webhook event. "
+            "Supports a single event (the first if multiple provided)."
```
korycins, line 344: Maybe?
```suggestion
            "Supports a single event (the first if multiple provided in the `query`)."
```

korycins (approved):

szczecha (approved):

PR #11561, opened 2022-12-25
Support Python 3.11

patrys:
This change vendors the Authorize.net SDK and its PyXB dependency to fix its imports. It then switches the Python version from 3.9 to 3.11.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


korycins (approved):
🍾 🍾 🍾 
(Black is failing and missing changelog entry/`skip changelog` label)

fowczarek (approved):

saleor/payment/gateways/authorize_net/__init__.py
```diff
@@ -20,6 +16,16 @@
     PaymentMethodInfo,
 )
 
+# vendored dependencies
+sys.path.append(os.path.join(os.path.dirname(__file__), "vendored"))
```
NyanKiyoshi, line 20: The [`setup.py`](https://github.com/AuthorizeNet/sdk-python/blob/11f5f69dc965c72c0e96d4bc6a4f98ca84727fa3/setup.py#L1) from AuthorizeNet could be used instead, Python will take care of the rest

Docs:
https://python-poetry.org/docs/dependency-specification/#path-dependencies

pyproject.toml
```diff
@@ -126,27 +126,16 @@ documentation = "https://docs.saleor.io/"
   fakeredis = "^1.8"
 
 [tool.black]
-target_version = [ "py35", "py36", "py37", "py38" ]
+target_version = [ "py39" ]
```
NyanKiyoshi, line 129: `black` [22.6.0](https://github.com/psf/black/releases/tag/22.6.0) adds support for [py11](https://github.com/psf/black/blob/aafc21aa77f5c4d2ebcb833aa60faba6c2138b94/src/black/mode.py#L125), any reason for not using Python 3.11 mode?

setup.cfg
```diff
@@ -30,6 +30,7 @@ exclude =
     __pycache__/,
     node_modules/,
     */migrations/
+    */vendored/
```
NyanKiyoshi, line 33: Why not simply fork their repository and include it as [git dependency in poetry](https://python-poetry.org/docs/dependency-specification/#git-dependencies) (with full hash, not partial hash like they suggest in docs) or a sub-module? It would make it much easier to review and see diffs in the future as well

pyproject.toml
patrys: I did not want to reformat any files in the process.

setup.cfg
patrys: I don't want us to maintain a fork or create an illusion that we do.

saleor/payment/gateways/authorize_net/__init__.py
patrys: We don't rely on poetry for running Saleor and we don't require poetry for installation.

patrys:
Superseded by #15357 as we're still blocked by Django 4.x adoption

PR #11573, opened 2022-12-27
Add mutation to create an report event

korycins:
I want to merge this change because it adds new mutation for creating the `TransactionEvent`. As described in section `2.1.3` of #10350.
Section `6.3` of #10350  contains more details about the way of working. 
Future PRs will extend the mutation to use the newly created event to re-calculate the transactionItem's amounts

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/payment/mutations.py
```diff
@@ -1229,3 +1223,153 @@ def perform_mutation(cls, root, info, **data):
...
+            description=(
+                "The url that will allow to redirect user to "
+                "payment provider page with event details."
+            )
+        )
+        message = graphene.String(description="The message related to the event.")
+        available_actions = graphene.List(
+            graphene.NonNull(TransactionActionEnum),
+            description="List of all possible actions for the transaction",
+        )
```
IKarbowiak, line 1275: Usually, when we have so many input fields we create a particular input type for that, what do you think about that?

IKarbowiak: After that the tests could be simplified - the constant could be created for the query.

korycins: We were thinking with @patrys about having flat input fields here, instead of having `input`. But agree with the `tests` argument.

IKarbowiak: What were the arguments for flat inputs?

korycins: @IKarbowiak Sorry for the late response. I had some things that I had to handle first. 
Honestly, I don't remember all arguments and prons/cons, but we were generally thinking about using a flat structure instead of injected input type. In that case, the inputs are related to the transactionEvent and TransactionItem (some of them are used to create new transaction event, but some of them will be used to make changes to the transaction item). The main task for the mutation is not creating a new TransactionEvent, but reporting the payment action.

saleor/graphql/payment/mutations.py
```diff
@@ -1229,3 +1223,153 @@ def perform_mutation(cls, root, info, **data):
...
+        ),
+    )
+
+    class Arguments:
+        id = graphene.ID(
+            description="The ID of the transaction.",
+            required=True,
+        )
+        psp_reference = graphene.String(
+            description=("PSP Reference of the event to report."), required=True
```
IKarbowiak, line 1249: Not needed brackets.

IKarbowiak (approved):

fowczarek (approved):

PR #11648, opened 2023-01-05
Rework checkout total calculation

maarcingebala:
Change the way of `checkout.total` calculation when taxes are disabled, to always sum line totals and shipping. This change is because the voucher is already propagated in line totals. We included this voucher once again in the total calculation, which was causing small rounding issues.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/calculations.py
+++ b/saleor/checkout/calculations.py
@@ -7,12 +7,11 @@ from prices import Money, TaxedMoney
 
 from ..checkout import base_calculations
 from ..core.prices import quantize_price
-from ..core.taxes import TaxData, zero_taxed_money
+from ..core.taxes import TaxData, zero_money, zero_taxed_money
 from ..discount import DiscountInfo
 from ..tax import TaxCalculationStrategy
 from ..tax.calculations.checkout import update_checkout_prices_with_flat_rates
 from ..tax.utils import (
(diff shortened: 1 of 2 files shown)


korycins (approved):

saleor/checkout/calculations.py
```diff
@@ -490,30 +490,26 @@ def _get_checkout_base_prices(
         unit_price = base_calculations.apply_checkout_discount_on_checkout_line(
             checkout_info, lines, line_info, discounts, unit_price
         )
-        line_total_price = quantize_price(unit_price * quantity, unit_price.currency)
+        line_total_price = quantize_price(unit_price, unit_price.currency) * quantity
```
patrys, line 493: I can't speak about the rest of this PR, but this particular change seems incorrect. While Saleor does not currently support fractional quantities, if we did add support for them, a quantity of `0.01` would result in USD line totals with four decimal places. Line totals need to be quantized after multiplication, so if adding quantization to unit prices helps, it's probably indicating a bug in another area in our logic.

maarcingebala: Revert this one.

patrys (approved):

Issue #11009, opened 2022-10-21
Allow querying orders by checkoutID


mmiszy:
We're using `orderCreateFromCheckout` to create an order. Later on, we want to be able to fetch that order by the checkout ID used for its creation.

Quoting Patrys:
> we should add a query to find an order using the checkout token, either as a separate `Query.orderByCheckoutToken` (we already have `orderByToken` so it would be consistent) or by adding another parameter to `Query.order`

korycins:
@patrys  `orderCreateFromCheckout` has flag `removeCheckout`. It means that we can have multiple orders created from single checkout.
We need to have an interface that would handle such case. 
- We could create `Query.ordersByCheckoutToken` that will return a list of orders created from the same checkout. 
- We could also create `Query.orderByCheckoutToken` that will return the latest/oldest order.

PR #11689, opened 2023-01-10
Add filter to query orders by checkout token

kadewu:
Fix https://github.com/saleor/saleor/issues/11009

Added new filter to query orders by checkout token.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/filters.py
+++ b/saleor/graphql/order/filters.py
@@ -100,6 +100,12 @@ def filter_channels(qs, _, values):
     return qs
 
 
+def filter_checkouts(qs, _, value):
+    if value:
+        qs = qs.filter(checkout_token=value)
+    return qs
+
+
 def filter_is_click_and_collect(qs, _, values):
     if values is not None:
         lookup = Q(collection_point__isnull=False) | Q(
@@ -201,6 +207,7 @@ class OrderFilter(DraftOrderFilter):
     numbers = ListObjectTypeFilter(
         input_class=graphene.String, method=filter_by_order_number
     )
+    checkout_token = django_filters.CharFilter(method=filter_checkouts)
 
     class Meta:
         model = Order
--- a/saleor/graphql/order/filters.py
+++ b/saleor/graphql/order/filters.py
@@ -100,9 +100,10 @@ def filter_channels(qs, _, values):
     return qs
 
 
-def filter_checkouts(qs, _, value):
-    if value:
-        qs = qs.filter(checkout_token=value)
+def filter_checkouts(qs, _, values):
+    if values:
+        _, checkout_ids = resolve_global_ids_to_primary_keys(values, "Checkout")
+        qs = qs.filter(checkout_token__in=checkout_ids)
     return qs
 
 
@@ -207,7 +208,7 @@ class OrderFilter(DraftOrderFilter):
     numbers = ListObjectTypeFilter(
         input_class=graphene.String, method=filter_by_order_number
     )
-    checkout_token = django_filters.CharFilter(method=filter_checkouts)
+    checkout_ids = GlobalIDMultipleChoiceFilter(method=filter_checkouts)
 
     class Meta:
         model = Order
--- a/saleor/graphql/order/tests/queries/test_order_with_filter.py
+++ b/saleor/graphql/order/tests/queries/test_order_with_filter.py
@@ -1249,3 +1249,79 @@ def test_order_query_with_filter_not_allow_numbers_and_ids_together(
     # then
     assert content["errors"][0]["message"] == error_message
     assert not content["data"]["orders"]
+
+
+def test_order_query_with_filter_by_checkout_token(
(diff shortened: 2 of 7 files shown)


saleor/graphql/order/tests/queries/test_order_with_filter.py
```diff
@@ -1249,3 +1249,79 @@ def test_order_query_with_filter_not_allow_numbers_and_ids_together(
...
+        orders_query_with_filter, variables, permissions=(permission_manage_orders,)
+    )
+
+    # then
+    content = get_graphql_content(response)
+    order_data = content["data"]["orders"]["edges"]
+
+    assert content["data"]["orders"]["totalCount"] == 4
+    for order in orders_from_checkout:
+        if order.checkout_token == checkout.token:
```
korycins, line 1279: This `if` can cause false positive. If you don't get any order with checkout_token, the test will pass.
Also personally I am not a fan of having `if` condition inside the test. For me it means, that the test have more than single way to finish.

saleor/graphql/order/tests/queries/test_order_with_filter.py
```diff
@@ -1249,3 +1249,79 @@ def test_order_query_with_filter_not_allow_numbers_and_ids_together(
     # then
     assert content["errors"][0]["message"] == error_message
     assert not content["data"]["orders"]
+
+
+def test_order_query_with_filter_by_checkout_token(
+    orders_query_with_filter,
+    staff_api_client,
+    permission_manage_orders,
+    orders_from_checkout,
```
korycins, line 1258: I would also add some orders here that doesn't have a given token. Just to confirm that we don't return all(), but only the one that have a checkout.token provided.

saleor/graphql/order/tests/queries/test_order_with_filter.py
korycins: IMO, you could do something like this:
```
order_data = content["data"]["orders"]["edges"]
orders_ids = set([edge["id"] for edge in order_data])
assert len(orders_ids.intersection(set([graphene.Node.to_global_id("Order", order.id) for order in orders_from_checkout]))) == 4
```
(needs to be refactored but I think that you could use `set` here)

fowczarek (approved):

saleor/graphql/order/tests/queries/test_order_with_filter.py
kadewu: Actually it was done already, but thanks for that comment, now I realized that I shouldn't have included orders without token in fixture named this way.

saleor/graphql/order/filters.py
```diff
@@ -201,6 +207,7 @@ class OrderFilter(DraftOrderFilter):
     numbers = ListObjectTypeFilter(
         input_class=graphene.String, method=filter_by_order_number
     )
+    checkout_token = django_filters.CharFilter(method=filter_checkouts)
```
maarcingebala, line 210: Did you consider making it a list filter that can return multiple results when multiple tokens are passed? We have the same for `ids` and `numbers`.

kadewu: Yeah I was actually wondering about that. But then I believe it would make sense to add `checkout_token` to the `order` type. And I'm not sure about that one.

maarcingebala: Yep, that would make sense. I think we could do that but maybe we should discuss. @fowczarek @korycins What do you think?

fowczarek: Yea we could show this field to Order Managers. It's could be helpful in debugging payment errors. 
I think we should filter by multiple tokens like in other places. 
I also consider that we should use `CheckoutID's` instead of tokens. What do you think? As far as I remember we deprecate fetching checkout by tokens and we require ids instead.

kadewu: I also wondered about using ID's instead of tokens and I'm fine with it. @mmiszy What do you think, will it be still good for you guys?

mmiszy: As long as I'll be able to query orders by checkout id - it's fine :)

saleor/graphql/order/types.py
```diff
@@ -1140,6 +1141,10 @@ class Order(ModelObjectType[models.Order]):
     external_reference = graphene.String(
         description=f"External ID of this order. {ADDED_IN_310}", required=False
     )
+    checkout_id = graphene.ID(
+        description=f"Id of the checkout that order was created from. {ADDED_IN_311}",
```
maarcingebala, line 1145:
```suggestion
        description=f"ID of the checkout that the order was created from. {ADDED_IN_311}",
```

maarcingebala (approved):

IKarbowiak (approved):

szczecha (approved):

PR #11736, opened 2023-01-16
Inherit events from query

zedzior:
https://github.com/saleor/saleor/issues/11456

I want to merge this change, because I want to get events directly from `query`, instead of providing it in separate fields, like `asyncEvents` and `syncEvents`.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/webhook/mutations.py
+++ b/saleor/graphql/webhook/mutations.py
@@ -79,11 +79,17 @@ class WebhookCreateInput(graphene.InputObjectType):
 
 def clean_webhook_events(_info, _instance, data):
     # if `events` field is not empty, use this field. Otherwise get event types
-    # from `async_events` and `sync_events`.
+    # from `async_events` and `sync_events`. If the fields are also empty, parse events
+    # from `query`.
     events = data.get("events", [])
     if not events:
         events += data.pop("async_events", [])
         events += data.pop("sync_events", [])
+
+    query = data.get("query", [])
+    if not events and query:
+        events = get_event_type_from_subscription(query)
+
     data["events"] = events
     return data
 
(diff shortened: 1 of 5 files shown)


saleor/graphql/webhook/mutations.py
```diff
@@ -359,14 +365,15 @@ def validate_input(cls, info, **data):
...
-        if not event_type:
+        event_types = get_event_type_from_subscription(query) if query else None
+        if not event_types:
             raise_validation_error(
                 field="query",
                 message="Can't parse an event type from query.",
                 code=WebhookDryRunErrorCode.UNABLE_TO_PARSE,
             )
 
+        event_type = event_types[0] if event_types else None
```
maarcingebala, line 376: Here we assume that if there are multiple events we always take first? I would add a comment that we have this assumption.

saleor/graphql/webhook/utils.py
```diff
@@ -1,58 +1,69 @@
...
+        if not events:
+            return []
+
+        fragments = get_event_fragment_names_from_query(ast)
+        if fragments:
+            events = [fragments.get(event, event) for event in events]
+
+        return list(map(to_snake_case, events))
+
+    except Exception:
```
maarcingebala, line 29: Which part exactly can throw an exception? It's better to wrap small chunks with try..except and handle a specific exception instead of `Exception` which might hide some other errors.

saleor/graphql/webhook/mutations.py
zedzior: For `webhookDryRun` and `webhookTrigger` mutation there is already an info in description like: _Supports a single event (the first, if multiple provided)_

For `webhookCreate` and `webhookUpdate` we are able to parse and handle multiple events

maarcingebala (approved):

Issue #11759, opened 2023-01-19
Validate subscription query from Webhook


lkostrowski:
### What I'm trying to achieve
During apps development, its easy to write invalid `query` for a webhook.

For example 
```
query: "
  subscirption { event { ...MyFragment on SomeEvent } }
"
```

`MyFragment` should be added inline to the query, but its not forced during local development. 

Saleor should validate correctness of query and return exhaustive error during installation.

### Describe a proposed solution
Validate query during
- App installation from Manifest
- `webhookCreate` mutation

PR #11797, opened 2023-01-24
Subscription query validation

zedzior:
https://github.com/saleor/saleor/issues/11759
I want to merge this change, because I want to improve webhook's subscription query validation, while running webhook mutations. This PR also apply the validation to manifest checks, during `appInstall` mutation.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/app/manifest_validations.py
```diff
@@ -188,47 +186,21 @@ def clean_extensions(manifest_data, app_permissions, errors):
...
-            try:
-                webhook["events"].append(sync_types[e_type])
-            except KeyError:
-                errors["webhooks"].append(
-                    ValidationError(
-                        "Invalid synchronous event.",
-                        code=AppErrorCode.INVALID.value,
-                    )
-                )
+        webhook["events"] = subscription_query.events
```
zedzior, line 203: With this change, we don't need `asyncEvents` and `syncEvents` fields at all. If it is ok, where can I find some description to mention that?

saleor/graphql/schema.graphql
```diff
@@ -15963,7 +15963,6 @@ type WebhookDryRunError {
 """An enumeration."""
 enum WebhookDryRunErrorCode {
   GRAPHQL_ERROR
-  UNABLE_TO_PARSE
```
zedzior, line 15966: WIll be now under `GRAPHQL_ERROR`, but the message will point accurately to error reason

saleor/graphql/webhook/subscription_query.py
```diff
@@ -0,0 +1,138 @@
...
+
+        if errors:
+            return errors
+
+        try:
+            self.events = self.get_events_from_subscription()
+        except Exception as err:
+            return [err]
+
+        self.is_valid = not bool(errors)
```
zedzior, line 57: This means, that subscription query is valid only if events can be parsed.

saleor/graphql/webhook/subscription_query.py
```diff
@@ -0,0 +1,138 @@
...
+
+        if errors:
+            return errors
+
+        try:
+            self.events = self.get_events_from_subscription()
+        except Exception as err:
+            return [err]
+
+        self.is_valid = True
```
zedzior, line 57: This means, that subscription query is valid only if events can be parsed. Any objections?

saleor/graphql/schema.graphql
maarcingebala: Actually, it's not a bad idea to return a more specific error, message is only for dev, while the code is for the frontend to know which error to display. The more accurate code, the easier it is for them to handle the error.

saleor/app/manifest_validations.py
maarcingebala: I think we can deprecate these fields. In the description we could say that these fields can be skipped whenever the `query` param is passed. But that would also mean that we would deprecate the old static payloads in favor of subscription ones.

maarcingebala: Hmm, but do we support subscriptions for EACH webhook there is in Saleor? Only then we could deprecate these input fields.

saleor/graphql/webhook/mutations/webhook_trigger.py
```diff
@@ -10,9 +10,11 @@
 from ...core.mutations import BaseMutation
 from ...core.types.common import WebhookTriggerError
 from ...core.utils import raise_validation_error
+from ..subscription_query import SubscriptionQuery
 from ..subscription_types import WEBHOOK_TYPES_MAP
 from ..types import EventDelivery
-from ..utils import get_event_type_from_subscription
+
+# from ..utils import get_events_from_subscription
```
maarcingebala, line 17: To remove?

saleor/graphql/webhook/mutations/webhook_create.py
```diff
@@ -113,12 +98,38 @@ def clean_input(cls, info: ResolveInfo, instance, data, **kwargs):
...
+        # if `events` field is not empty, use this field. Otherwise get event types
+        # from `async_events` and `sync_events`. If the fields are also empty,
+        # parse events from `query`.
+        events = data.get("events", [])
+        if not events:
+            events += data.pop("async_events", [])
+            events += data.pop("sync_events", [])
+
+        if not events and subscription_query:
+            events = subscription_query.events
```
maarcingebala, line 128: Maybe if the subscription query is passed, we should ignore `events`, `asyncEvents`, and `syncEvents`? If the query is present, it should have precedence over other fields, I guess.

saleor/graphql/webhook/subscription_query.py
```diff
@@ -0,0 +1,140 @@
...
+
+class IsFragment(Flag):
+    TRUE = True
+    FALSE = False
+
+
+class SubscriptionQueryError(Enum):
+    MISSING_SUBSCRIPTION = "Subscription operation can't be found."
+    MISSING_EVENT_FIELD = "Event field can't be found."
+    MISSING_EVENTS = "Can't find a single event."
```
maarcingebala, line 26: Maybe instead of returning generic errors in API, we should return these specific errors to give the consumer more precise info about what is wrong. We would have to add these to schema with some descriptions then.

saleor/graphql/webhook/subscription_query.py
maarcingebala: I think it makes sense.

saleor/graphql/webhook/mutations/webhook_create.py
maarcingebala: Ignore this comment, when any of those "events" fields are passed, they should take precedence.

jakubkuc (approved):

saleor/app/manifest_validations.py
```diff
@@ -230,6 +222,27 @@ def clean_webhooks(manifest_data, errors):
...
+            errors["webhooks"].append(
+                ValidationError(
+                    "Subscription query is not valid: " + subscription_query.error_msg,
+                    code=AppErrorCode.INVALID.value,
+                )
+            )
+
+        if not webhook["events"]:
+            webhook["events"] = subscription_query.events
+        else:
```
korycins, line 236: You can use `elif`here

saleor/graphql/webhook/subscription_query.py
```diff
@@ -0,0 +1,146 @@
...
+from graphql.language.ast import (
+    Document,
+    Field,
+    FragmentDefinition,
+    FragmentSpread,
+    InlineFragment,
+    OperationDefinition,
+)
+
+from saleor.webhook.error_codes import WebhookErrorCode
```
korycins, line 17: Absolute import

saleor/graphql/webhook/subscription_query.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self, query: str):
+        self.query: str = query
+        self.is_valid: bool = False
+        self.ast: Document = Document("")
+        self.events: List[str] = []
+        self.error_code: Optional[str] = None
+        self.errors = self.validate_query()
+        self.error_msg: str = ";".join(set([err.message for err in self.errors]))
+
+    def validate_query(self):
```
korycins, line 35: Maybe would be good to add typing here, for the data that we return here

saleor/graphql/webhook/subscription_query.py
```diff
@@ -0,0 +1,146 @@
...
+                continue
+            event_definition = fragment_definitions[event_name]
+            self._get_events_from_field(event_definition, unpacked_events)
+        events_and_fragments.update(unpacked_events)
+        events = [k for k, v in events_and_fragments.items() if not v]
+
+        if not events:
+            raise ValidationError(
+                message="Can't find a single event.",
+                code=WebhookErrorCode.UNABLE_TO_PARSE.value,
```
korycins, line 92: Wouldn't be good to have a separate error code here? We were able to determine that we don't have any events here, so we were able to parse the query, right?

zedzior: MISSING_EVENT error added

szczecha (approved):

korycins (approved):

maarcingebala (approved):

Issue #11755, opened 2023-01-18
Bug: App looks like installed, but installation is still pending.


orzechdev:
### What are you trying to achieve?

When installing app via `appInstall` mutation, everything is fine, mutation returns app installation with status pending. But after that, when I query for `apps`, I see that the app which I wanted to install is already there, also it is marked as active. It could mean that the app has been installed, but… when I query for `appsInstallations` I see that the corresponding installation is still pending.

I want to have only installed app be present in `apps` query when the installation is successfully completed. Otherwise, dashboard can fetch inconsistent apps list with their installations list.

### Steps to reproduce the problem

1. Call `appInstall` mutation.
2. Call `apps` query.
3. Call `appsInstallations` mutation.

You'll see that although the app in 2nd step is installed already, in 3rd step installation still exists as pending.

### What did you expect to happen?

App shouldn't exist in the `apps` query when installation is still pending.

### Environment

Saleor version: v3.10.0-a
OS and version: MacOS 12.4

PR #11809, opened 2023-01-25
Exclude apps with pending installations from app queries.

zedzior:
https://github.com/saleor/saleor/issues/11755
I want to merge this change, because I want to have only installed app be present in `app` and `apps` query, when the installation is successfully completed. Otherwise, dashboard can fetch inconsistent apps list with their installations list.

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/app/installation_utils.py
```diff
@@ -73,6 +74,8 @@ def install_app(app_installation: AppInstallation, activate: bool = False):
         type=AppType.THIRDPARTY,
         audience=manifest_data.get("audience"),
     )
+
+    app.store_value_in_private_metadata({"pending_installation": True})
```
maarcingebala, line 78: This is fine, but why not store it in the model as a regular field?

zedzior: I didn't want to make the model messy, as this is needed for couple of seconds. Any pros for regular field in this case?

fowczarek: The field on the model will be a better solution. (If we introduce filtering via this field, it's cheaper to optimize it.)

maarcingebala: Generally, model fields are better for filtering and DB operations, while metadata is mostly for external integrations to write arbitrary data we don't need to know about. We have the AppInstallation model, which is dedicated to tracking the installation process, could we store this info there perhaps?

fowczarek: We filter through this field in resolvers. For sure we should add a field on the model for storing this information.

zedzior: OK, so I will add field to App model, because AppInstallation has no relationship with App

saleor/site/models.py
```diff
@@ -116,7 +116,7 @@ def default_from_email(self) -> str:
         sender_address: Optional[str] = self.default_mail_sender_address
 
         if not sender_address:
-            sender_address = settings.DEFAULT_FROM_EMAIL  # type: ignore[misc]
+            sender_address = settings.DEFAULT_FROM_EMAIL
```
zedzior, line 119: This caused mypy checks to fail.

fowczarek (approved):

korycins (approved):

szczecha (approved):

maarcingebala (approved):

Issue #11831, opened 2023-01-27
Bug: draftOrderUpdate mutation doesn't set shipping price


kadewu:
### What are you trying to achieve?

When updating shipping through `draftOrderUpdate` I want to update `shippingPrice`

### Steps to reproduce the problem

For mutation:

```
mutation {
  draftOrderUpdate(id: "orderID", input: {shippingMethod: "shippingID"}) {
    order {
      shippingMethodName
      shippingPrice {
        net {
          amount
        }
      }
    }
  }
}
```

Output:
```
{
  "data": {
    "draftOrderUpdate": {
      "order": {
        "shippingMethodName": "Name",
        "shippingPrice": {
          "net": {
            "amount": 0
          }
        }
      }
    }
  },
  "extensions": {
    "cost": {
      "requestedQueryCost": 0,
      "maximumAvailable": 50000
    }
  }
}
```

### What did you expect to happen?

Expected output:
```
{
  "data": {
    "draftOrderUpdate": {
      "order": {
        "shippingMethodName": "Name",
        "shippingPrice": {
          "net": {
            "amount": 12.0
          }
        }
      }
    }
  },
  "extensions": {
    "cost": {
      "requestedQueryCost": 0,
      "maximumAvailable": 50000
    }
  }
}
```

### Environment

Saleor version: … 3.11
OS and version: …

PR #11832, opened 2023-01-27
Fix shipping price update

kadewu:
Fix https://github.com/saleor/saleor/issues/11831

Beside fixing bug:
- add shipping method validation to `draftOrderCreate` and `draftOrderUpdate`
- remove redundant code definition

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/mutations/draft_order_create.py
+++ b/saleor/graphql/order/mutations/draft_order_create.py
@@ -286,12 +286,6 @@ class DraftOrderCreate(ModelMutation, I18nMixin):
             instance.billing_address = billing_address.get_copy()
 
     @staticmethod
-    def _parse_shipping_method_name(instance: models.Order, cleaned_input):
-        shipping_method = cleaned_input.get("shipping_method")
-        if shipping_method:
-            instance.shipping_method_name = shipping_method.name
-
-    @staticmethod
     def _save_lines(info, instance, lines_data, app, manager):
         if lines_data:
             lines = []
@@ -315,7 +309,7 @@ class DraftOrderCreate(ModelMutation, I18nMixin):
     def _commit_changes(
         cls, info: ResolveInfo, instance, cleaned_input, is_new_instance, app
     ):
-        if shipping_method := cleaned_input["shipping_method"]:
+        if shipping_method := cleaned_input.get("shipping_method"):
             instance.shipping_method_name = shipping_method.name
             tax_class = shipping_method.tax_class
             if tax_class:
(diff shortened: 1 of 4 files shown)


zedzior (approved):

saleor/graphql/order/mutations/draft_order_create.py
```diff
@@ -331,9 +328,56 @@ def _commit_changes(
...
+                instance.shipping_method,
+                shipping_channel_listing,
+            )
+            manager = get_plugin_manager_promise(info.context).get()
+            error = get_shipping_method_availability_error(
+                instance, shipping_method_data, manager
+            )
+            if error:
+                raise ValidationError({"shipping_method": error})
+            instance.base_shipping_price = shipping_channel_listing.price
```
korycins, line 356: Do we really need to determine if the field should be included in saving? I would say that we could simplify the body of the method, and set `zero_money` by default and then if we should calculate the shipping prices we can overwrite the value. (We don't save here any db queries as we always make a .save())

saleor/graphql/order/mutations/draft_order_create.py
```diff
@@ -331,9 +328,56 @@ def _commit_changes(
...
+        else:
+            instance.base_shipping_price = zero_money(instance.currency)
+            return False
+
+    @classmethod
+    def _get_shipping_channel_listing(cls, instance, cleaned_input):
+        if shipping_method := cleaned_input["shipping_method"]:
+            shipping_channel_listing = (
+                shipping_models.ShippingMethodChannelListing.objects.filter(
+                    shipping_method=shipping_method, channel=instance.channel
```
korycins, line 367: If you don't have prefetched `channel`, AFAIk this will raise an additional DB query for fetching a channel. I would suggest to use `channel__id=instance.channel_id`

saleor/graphql/order/mutations/draft_order_create.py
```diff
@@ -285,12 +288,6 @@ def _save_addresses(instance: models.Order, cleaned_input):
             billing_address.save()
             instance.billing_address = billing_address.get_copy()
 
-    @staticmethod
-    def _parse_shipping_method_name(instance: models.Order, cleaned_input):
-        shipping_method = cleaned_input.get("shipping_method")
-        if shipping_method:
-            instance.shipping_method_name = shipping_method.name
```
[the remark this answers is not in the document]
zedzior: It is done in line 316 (`_commit_changes` function)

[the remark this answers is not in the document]
kadewu: As @zedzior mentioned, this is removing duplication of the code.

saleor/graphql/order/mutations/draft_order_create.py
kadewu: You're right, that logic can be changed here.

saleor/graphql/order/mutations/draft_order_create.py
kadewu: Thanks for catching!

korycins (approved):

PR #11873, opened 2023-01-31
Partial DB replica emulation

magul:
It doesn't actually restrict connection to read-only, but having two connections open (even to the same database) allow us to locally recreate bug connected with using two request contexts in one Django request (I mean synchronous webhooks' subscriptions here).

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -86,19 +86,20 @@ DATABASE_CONNECTION_DEFAULT_NAME = "default"
 # TODO: For local envs will be activated in separate PR.
 # We need to update docs an saleor platform.
 # This variable should be set to `replica`
-DATABASE_CONNECTION_REPLICA_NAME = "default"
+DATABASE_CONNECTION_REPLICA_NAME = "replica"
 
 DATABASES = {
     DATABASE_CONNECTION_DEFAULT_NAME: dj_database_url.config(
         default="postgres://saleor:saleor@localhost:5432/saleor",
         conn_max_age=DB_CONN_MAX_AGE,
     ),
-    # TODO: We need to add read only user to saleor platfrom, and we need to update
-    # docs.
-    # DATABASE_CONNECTION_REPLICA_NAME: dj_database_url.config(
-    #     default="postgres://saleor_read_only:saleor@localhost:5432/saleor",
-    #     conn_max_age=DB_CONN_MAX_AGE,
-    # ),
+    DATABASE_CONNECTION_REPLICA_NAME: dj_database_url.config(
+        default="postgres://saleor:saleor@localhost:5432/saleor",
+        # TODO: We need to add read only user to saleor platform,
+        # and we need to update docs.
+        # default="postgres://saleor_read_only:saleor@localhost:5432/saleor",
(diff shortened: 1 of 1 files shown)


saleor/settings.py
```diff
@@ -86,19 +86,20 @@ def get_bool_from_env(name, default_value):
...
         conn_max_age=DB_CONN_MAX_AGE,
     ),
-    # TODO: We need to add read only user to saleor platfrom, and we need to update
-    # docs.
-    # DATABASE_CONNECTION_REPLICA_NAME: dj_database_url.config(
-    #     default="postgres://saleor_read_only:saleor@localhost:5432/saleor",
-    #     conn_max_age=DB_CONN_MAX_AGE,
-    # ),
+    DATABASE_CONNECTION_REPLICA_NAME: dj_database_url.config(
+        default="postgres://saleor:saleor@localhost:5432/saleor",
```
NyanKiyoshi, line 97: You can make the session read-only using this (should theoretically work with Django):
```suggestion
        default="postgres://saleor:saleor@localhost:5432/saleor?options=-cdefault_transaction_read_only%3Dtrue&target_session_attrs=read-only",
```

magul: `?options=-cdefault_transaction_read_only%3Dtrue&target_session_attrs=read-only` I guess we can replace `%3D` with `=` here (as it's going to be url encoded anyway)?

Also - would that mean, that it will behave as read only connection and we coud just remove `TODO` statement there?

magul: > ?options=-cdefault_transaction_read_only%3Dtrue&target_session_attrs=read-only I guess we can replace %3D with = here (as it's going to be url encoded anyway)?

nvm, I've just spotted that this `%3D` is another layer of `=`

NyanKiyoshi: > Also - would that mean, that it will behave as read only connection and we coud just remove `TODO` statement there?

Yes it would be read-only. It's not a real replica though thus the replication lag is not there but that's good enough for now. For local development we may want to have a real logical-replication going on to detect issues but that needs to be tackled in a ticket.

`TODO` appear to only want read-only session without replication lag thus I do believe it would resolve them.

fowczarek (approved):

korycins (approved):

saleor/settings.py
magul: It looks that `dj_database_url` disregard parameters set after `?`, so for the time being I'm leaving/merging it as it is @NyanKiyoshi .

```python
>>> dj_database_url.config(default="postgres://saleor:saleor@localhost:5432/saleor?options=-cdefault_transaction_read_only%3Dtrue&target_session_attrs=read-only", conn_max_age=10)
{'NAME': 'saleor', 'USER': 'saleor', 'PASSWORD': 'saleor', 'HOST': 'db', 'PORT': '', 'CONN_MAX_AGE': 10, 'CONN_HEALTH_CHECKS': False, 'ENGINE': 'django.db.backends.postgresql'}
```

Issue #11906, opened 2023-02-02
Bug: Two request context in single Django request.


zedzior:
### What are you trying to achieve?

When CalculateTaxes event is triggered during `odrerDiscountAdd` mutation (within transaction), subscription payload is prepared base on replica db. This leads to uncommitted changes not being visible in the payload.

### Steps to reproduce the problem

Steps to reproduce:
1. Create new saleor.cloud instance ~~(locally (docker+gunicorn) it is not reproducible for me - which can indicate some setting impacts this)~~.
2. Configure webhook with subscription query provided below and syncEvent `CalculateTaxes`
3. Through API create Draft Order (`draftOrderCreate`)
4. Add any non-free lines (`orderLinesCreate`)
5. Add discount (`orderDiscountAdd`)
6. Tax webhook will be sent without discount in payload.

```grapql
subscription {
  event {
    ... on CalculateTaxes {
      taxBase {
        sourceObject {
          ... on Order {
            discounts {
              amount {
                amount
              }
              reason
            }
          }
        }
      }
    }
  }
}
```

update:
To reproduce locally change settings.py:
```
DATABASE_CONNECTION_DEFAULT_NAME = "default"
DATABASE_CONNECTION_REPLICA_NAME = "replica"

DATABASES = {
    DATABASE_CONNECTION_DEFAULT_NAME: dj_database_url.config(
        default="postgres://saleor:saleor@localhost:5432/saleor",
        conn_max_age=DB_CONN_MAX_AGE,
    ),
    DATABASE_CONNECTION_REPLICA_NAME: dj_database_url.config(
        default="postgres://saleor_read_only:saleor@localhost:5432/saleor",
        conn_max_age=DB_CONN_MAX_AGE,
    ),
}
```

### What did you expect to happen?

Discount visible in webhook payload

### Environment

Saleor version: 3.7+
Only saleor.cloud instances

PR #11920, opened 2023-02-06
Use default db for synchronous tax webhooks

zedzior:
https://github.com/saleor/saleor/issues/11906
I want to merge this change because when tax webhooks are triggered during mutation, subscription payload is generated based on replica db, so changes done by the mutation are not visible in the payload.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/webhook/subscription_payload.py
+++ b/saleor/graphql/webhook/subscription_payload.py
@@ -90,11 +90,14 @@ def generate_payload_from_subscription(
     app_id = app.pk if app else None
 
     request.app = app
+    is_mutation = request.is_mutation if request.is_mutation else False
+    context = get_context_value(request)
+    context.is_mutation = is_mutation
 
     results = document.execute(
         allow_subscriptions=True,
         root=(event_type, subscribable_object),
-        context=get_context_value(request),
+        context=context,
     )
     if hasattr(results, "errors"):
         logger.warning(
--- a/saleor/plugins/webhook/plugin.py
+++ b/saleor/plugins/webhook/plugin.py
@@ -1479,6 +1479,7 @@ class WebhookPlugin(BasePlugin):
             parse_tax_data,
             checkout_info.checkout,
             self.requestor,
+            is_mutation=True,
         )
 
     def get_taxes_for_order(
@@ -1490,6 +1491,7 @@ class WebhookPlugin(BasePlugin):
             parse_tax_data,
             order,
             self.requestor,
+            is_mutation=True,
         )
 
     def get_shipping_methods_for_checkout(
--- a/saleor/plugins/webhook/tasks.py
+++ b/saleor/plugins/webhook/tasks.py
@@ -254,6 +254,7 @@ def trigger_all_webhooks_sync(
     parse_response: Callable[[Any], Optional[R]],
     subscribable_object=None,
     requestor=None,
+    is_mutation=False,
 ) -> Optional[R]:
     """Send all synchronous webhook request for given event type.
 
@@ -272,6 +273,8 @@ def trigger_all_webhooks_sync(
                 request_context = initialize_request(
                     requestor, event_type in WebhookEventSyncType.ALL
                 )
+                request_context.is_mutation = is_mutation
+
             delivery = create_delivery_for_subscription_sync_event(
                 event_type=event_type,
                 subscribable_object=subscribable_object,


maarcingebala (approved):

saleor/graphql/webhook/subscription_payload.py
```diff
@@ -90,11 +90,14 @@ def generate_payload_from_subscription(
     app_id = app.pk if app else None
 
     request.app = app
+    is_mutation = request.is_mutation if hasattr(request, "is_mutation") else False
```
korycins, line 93: Isn't something that should be inside `get_context_value` ?

saleor/plugins/webhook/plugin.py
```diff
@@ -1479,6 +1479,7 @@ def get_taxes_for_checkout(
             parse_tax_data,
             checkout_info.checkout,
             self.requestor,
+            is_mutation=True,
```
korycins, line 1482: Are you sure that this is a correct way for handling this? Maybe I am missing something, but these hooks `get_taxes_for_checkout`, `get_taxes_for_order` are also called inside the query (in case when expiration time is exceeded). Maybe we could set this flag when we initialise the manager? Like `get_plugin_manager_promise` could determine if the request is query or mutation and pass this info directly to the `PluginsManager(is_mutation=<value>)`?

saleor/plugins/webhook/tasks.py
```diff
@@ -272,6 +273,8 @@ def trigger_all_webhooks_sync(
                 request_context = initialize_request(
                     requestor, event_type in WebhookEventSyncType.ALL
                 )
+                request_context.is_mutation = is_mutation
```
korycins, line 276: Why we don't set this inside `initialize_request`? It looks like this is a perfect function to set some flags on the request.

saleor/plugins/webhook/plugin.py
maarcingebala: I think it is the right way, just the name `is_mutation` is confusing. IMO this entire flag should be called `enforce_master_db` or something like this.

saleor/plugins/webhook/tasks.py
fowczarek: We shouldn't set this flag for all webhooks. It's required only for synchronous webhooks. 
Maybe we should add the arg to `initialize_request` e.g. `use_replicas`, and set it by default to `True`.

fowczarek (approved):

saleor/plugins/webhook/tasks.py
korycins: Yeah, but `is_mutation` is boolean, so you can by default set it to `False` and change it only for sync webhooks to `True`.

korycins (approved):

saleor/graphql/webhook/subscription_payload.py
```diff
@@ -18,7 +18,9 @@
 logger = get_task_logger(__name__)
 
 
-def initialize_request(requestor=None, sync_event=False) -> SaleorContext:
+def initialize_request(
+    requestor=None, sync_event=False, is_mutation=False
```
maarcingebala, line 22: I would change `is_mutation=False` to `allow_replica=True` everywhere to keep it consistent.

saleor/plugins/webhook/tasks.py
```diff
@@ -270,8 +271,9 @@ def trigger_all_webhooks_sync(
         if webhook.subscription_query:
             if request_context is None:
                 request_context = initialize_request(
-                    requestor, event_type in WebhookEventSyncType.ALL
+                    requestor, event_type in WebhookEventSyncType.ALL, not allow_replica
```
maarcingebala, line 274: When we change is_mutation to allow_replica this should become: 
```suggestion
                    requestor, event_type in WebhookEventSyncType.ALL, allow_replica
```

saleor/graphql/webhook/subscription_payload.py
zedzior: You are right, this function is only used by subscription stuff.

saleor/graphql/context.py
```diff
@@ -17,7 +17,7 @@
 def get_context_value(request: HttpRequest) -> SaleorContext:
     request = cast(SaleorContext, request)
     request.dataloaders = {}
-    request.is_mutation = False
+    request.is_mutation = getattr(request, "is_mutation", False)
```
maarcingebala, line 20: We should change this as well to replace is_mutation with allow_replica entirely:
```suggestion
    request.is_mutation = getattr(request, "allow_replica", True)
```

szczecha (approved):

maarcingebala (approved):

PR #11938, opened 2023-02-07
Force to return proper result for transaction webhooks

korycins:
I want to merge this change because it introduces a protection for recieving a correct `result` type for a request. 
For example, if Saleor sends a webhook with information that refund should be executed, the optional response can return `result` - "refund_success", or "refund_fail"

> **Warning**
> Please ignore failing CI. The tests are passing there but there are some issues with pre-commit but I will solve this later on main feature branch.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


fowczarek (approved):

saleor/payment/utils.py
```diff
@@ -693,11 +693,37 @@ def payment_owned_by_user(payment_pk: int, user) -> bool:
     )
 
 
+def get_correct_event_types_based_on_request_type(request_type: str) -> List[str]:
```
kadewu, line 696:
```suggestion
def get_correct_event_types_based_on_request_type(request_type: str) -> list[str]:
```
From python 3.9 onward using `list` over `typing.List` is preferred way.

saleor/payment/utils.py
```diff
@@ -693,11 +693,37 @@ def payment_owned_by_user(payment_pk: int, user) -> bool:
...
+    elif request_type == TransactionEventType.REFUND_REQUEST:
+        return [
+            TransactionEventType.REFUND_FAILURE,
+            TransactionEventType.REFUND_SUCCESS,
+        ]
+    elif request_type == TransactionEventType.CANCEL_REQUEST:
+        return [
+            TransactionEventType.CANCEL_FAILURE,
+            TransactionEventType.CANCEL_SUCCESS,
+        ]
```
IKarbowiak, line 717: The map with the request type to correct events types can be created and used here, instead of multiple `if, elif`

IKarbowiak (approved):

Issue #11257, opened 2022-11-21
Canceling the orders that haven't been paid.


korycins:
### What I'm trying to achieve
With new payment's transaction flow, there is a possibility to create an order then paid for it. If customer cancels the payment process, we will end up with orders that haven't been paid and allocated stock. We need to have a mechanism to release stock and cancel the orders.

### General assumptions

- The `OrderSettings` in `Channel` type should include a new field `expireOrdersAfter` of type `Minute`.
- The `channelCreate` and `channelUpdate` mutations should include a new input field `expireOrdersAfter` of type `Minute`.
- The `Channel` model should have a new field `expire_orders_after` to store the expiration delta. The default value for this field should be None, which means do not expire any orders.
- Only users with `MANAGE_ORDERS` or `MANAGE_CHANNELS` permissions should be able to update `OrderSettings` (Keep in mind that `MANAGE_ORDERS` can be used **only** to update `orderSettings`. That means, that user cannot use MANAGE_ORDERS permission on mutation that will update `OrderSettings` and other `Channel` field).
- When creating an order from checkout, Saleor should create an `unconfirmed` order, and only confirm it when a transaction is attached to the order (but only when `auto-confirm` is active).
- When creating a transaction for an order, Saleor should mark the order as `confirmed` (but only when `auto-confirm` is active).
- The `expired order` should be treated in the same way as a `canceled order`, that means: 
    - any update action on `expired order` should not be allowed
    - cannot generate `invoice` for `expired_order`
    - `orders_total` from `Order` should not include `expired order`
    - `expired order` cannot be captured or confirmed
    - reporting product sales should not contain `expired orders`
- New celery task that will be triggered by `celery_beat` to convert `unconfirmed` and not paid orders into `expired orders`.
- `expired orders` will have their stock released and `vouchers` will be available again.
- When order is being expired, we will trigger the `order_expired` and `order_updated` event.
- New `CeleryTask` object should be added as a lock for celery tasks.
- The `Order.status` field new value: `EXPIRED`.

### Describe a proposed solution
The proposed solution relays on the changes proposed in https://github.com/saleor/saleor/issues/11337.
Staff user will have a possibility to provide the time, that Saleor will use to change the unconfirmed and not paid orders into expired orders. We will release a stock allocation for these orders

##### `OrderSettings` type
```graphql
type OrderSettings {
	"""
	Expiration time in minutes. Default null - means do not expire any orders
	"""
	expireOrdersAfter: Minute <New scalar to add>
[...]

PR #11960, opened 2023-02-09
Cancelling orders

kadewu:
Fix: https://github.com/saleor/saleor/issues/11257

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/channel/mutations/channel_create.py
```diff
@@ -43,6 +44,12 @@ class OrderSettingsInput(graphene.InputObjectType):
         description="When enabled, all non-shippable gift card orders "
         "will be fulfilled automatically. By defualt set to True.",
     )
+    expire_orders_after = Minute(
+        required=False,
+        description=(
+            "Expiration time in minutes. Default null - means do not expire any orders"
```
korycins, line 50: preview+ added_in

saleor/graphql/channel/types.py
```diff
@@ -178,6 +179,13 @@ class OrderSettings(ObjectType):
             "will be fulfilled automatically."
         ),
     )
+    expire_orders_after = Minute(
+        required=False,
+        description=(
+            "Expiration time in minutes. Default null - means do not expire any orders"
+            + ADDED_IN_312
```
korycins, line 186: I think that we can add preview also

saleor/graphql/channel/types.py
```diff
@@ -426,4 +434,5 @@ def resolve_order_settings(root: models.Channel, _info):
             automatically_fulfill_non_shippable_gift_card=(
                 root.automatically_fulfill_non_shippable_gift_card
             ),
+            expire_orders_after=(root.expire_orders_after),
```
korycins, line 437: No need to have `()`

saleor/graphql/checkout/tests/mutations/test_order_create_from_checkout.py
```diff
@@ -207,6 +207,49 @@ def test_order_from_checkout(
     _recalculate_order_prices_mock.assert_not_called()
 
 
+def test_order_from_checkout_with_transaction(
```
korycins, line 210: I see that you updated the rest of the tests, but I think that it would be good to have the same test as this one but for `channel.automatically_confirm_all_new_orders` is False. (Maybe parametrize can work here).
Also I see that explicit test for checkout with transaction is missing (except the one that just added), so I would propose to split this into three tests:
`test_order_from_checkout_with_transaction` <- the one that handles a scenario when checkout has transaction attached
`test_order_from_checkout_with_unconfirmed_status` <- the one that controls the flag on channel and confirm the expected status
`test_order_from_checkout_with_confirmed_status` <- the one that controls the flag on channel and confirm the expected status

The last two can be probably done with pytest.mark.parametrize

@kadewu  WDYT?

saleor/graphql/payment/mutations.py
```diff
@@ -824,16 +833,22 @@ def create_transaction_event(
...
+        authorized_amount = transaction_data.get("authorized_value", Decimal(0))
+        charged_amount = transaction_data.get("charged_value", Decimal(0))
+        updated_fields = {}
+        if authorized_amount or charged_amount:
+            updated_fields = order_total_authorized_and_total_charged(
+                authorized_amount_to_add=authorized_amount,
+                charged_amount_to_add=charged_amount,
+            )
+        if (
+            order.channel.automatically_confirm_all_new_orders
```
korycins, line 847: Have you considered extending `update_order_status` to handle such scenario? I think that there is multiple places where we use channel to calculate the order status, so I think maybe it would be worth to add it there?

saleor/order/tasks.py
```diff
@@ -21,3 +32,58 @@ def send_order_updated(order_ids):
...
+            orders_to_update = []
+            for order in orders:
+                if _process_order(order, manager, now):
+                    orders_to_update.append(order)
+            Order.objects.bulk_update(orders_to_update, ["status", "updated_at"])
+
+
+def _process_order(order, manager, now):
+    if (
+        order.channel.expire_orders_after is not None
```
korycins, line 67: What will happen If i provide expire_orders_after < 0 or expire_orders_after == 0?

saleor/order/tasks.py
```diff
@@ -21,3 +32,58 @@ def send_order_updated(order_ids):
...
+
+def _expire_orders(manager, now):
+    qs = Order.objects.filter(
+        status=OrderStatus.UNCONFIRMED,
+        channel__expire_orders_after__isnull=False,
+        total_charged_amount=Decimal(0),
+    ).select_related("channel")
+    for ids_batch in _queryset_in_batches(qs):
+        with traced_atomic_transaction():
+            orders = Order.objects.filter(id__in=ids_batch).prefetch_related("user")
```
korycins, line 57: Shouldn't be `select_related("user")` instead of `prefetch_related` ?

saleor/order/tasks.py
```diff
@@ -21,3 +32,58 @@ def send_order_updated(order_ids):
...
+    for ids_batch in _batch_ids(ids, 1000):
+        yield ids_batch
+
+
+def _expire_orders(manager, now):
+    qs = Order.objects.filter(
+        status=OrderStatus.UNCONFIRMED,
+        channel__expire_orders_after__isnull=False,
+        total_charged_amount=Decimal(0),
+    ).select_related("channel")
```
korycins, line 54: No need to have select here. You use the `qs` only to fetch order's IDs.

korycins: Also I am not conviced that the query is safe enough. 
We should take into account only the orders that are unconfirmed, without any transaction and without any active payment attached to it. This should give us an orders created by "order-create-from-checkout" mutations.
You should also double check, what kind of the status received the draft order when it is marked as a completed (with and without active auto-confirm ).

saleor/order/tasks.py
```diff
@@ -21,3 +32,58 @@ def send_order_updated(order_ids):
...
+            for order in orders:
+                if _process_order(order, manager, now):
+                    orders_to_update.append(order)
+            Order.objects.bulk_update(orders_to_update, ["status", "updated_at"])
+
+
+def _process_order(order, manager, now):
+    if (
+        order.channel.expire_orders_after is not None
+        and order.created_at
```
korycins, line 68: Why we can't filter it out during the fetching all expired orders? `qs` from line 50.  With current approach and big amount of `expire_after` this will cause that we will need to fetch a lot of unneeded orders.

korycins: Thanks to it, you don't need this method, you can just call `expire_orders` to all fetched orders.

saleor/order/tasks.py
```diff
@@ -21,3 +32,58 @@ def send_order_updated(order_ids):
...
+        <= now - timezone.timedelta(minutes=order.channel.expire_orders_after)
+    ):
+        expire_order(order, order.user, manager, save_order_object=False)
```
korycins, line 71: Have you tested it with bigger database? each call of `deallocate_stock_for_order` produces 4 db queries for each order. We need to be sure that this will not be a bottle neck. If yes, then we need to rewrite `deallocate_stock_for_order` to accept data like `lines` of prefetched allocations for the order-lines

saleor/order/tasks.py
korycins: Also, we should test this with the database that doesn't have any order with transaction to be sure, that we don't break anything related to the old payments or order created from draft, or order created from replacment

saleor/settings.py
```diff
@@ -599,6 +599,10 @@ def get_host():
         "schedule": timedelta(seconds=20),
         "options": {"expires": BEAT_UPDATE_SEARCH_EXPIRE_AFTER_SEC},
     },
+    "expire-orders": {
+        "task": "saleor.order.tasks.expire_orders_task",
+        "schedule": crontab(minute="*/5"),
```
korycins, line 604: Why not `timedelta(minutes=5),`? It should be more readable with `timedelta`

korycins: Also, I would create a settings variable to control it. In case when the interval is too low or too big, we will be able to control it.

saleor/graphql/checkout/tests/mutations/test_order_create_from_checkout.py
kadewu: Make sense, will do that.

saleor/graphql/payment/mutations.py
kadewu: Yeah, actually I had that in mind, but forget to change that later. Will do that.

saleor/order/tasks.py
kadewu: In both scenarios, the order will be cancelled the moment the job is triggered.

saleor/order/tasks.py
kadewu: `user` can be null and also idea is to not make those `Order` objects even bigger.
I still wonder if I should do the same with `Channel`

saleor/order/tasks.py
kadewu: Yes, I did test it, but my benchmarks on local machines may have been not enough to test performance.

 Though I agree, that It would be better to optimize it, no matter of results.

saleor/core/tasks.py
```diff
@@ -1,11 +1,25 @@
...
 from django.db.models import Exists, OuterRef, Q
 from django.utils import timezone
 
 from ..celeryconf import app
-from .models import EventDelivery, EventDeliveryAttempt, EventPayload
+from .models import CeleryTask, EventDelivery, EventDeliveryAttempt, EventPayload
+
+
+@contextmanager
+def celery_task_lock(task_name):
```
fowczarek, line 14: I think we should add some docstring here.

saleor/graphql/payment/mutations.py
```diff
@@ -59,15 +60,23 @@
...
+
+def update_order(
+    order_id: uuid.UUID,
+    **update_fields,
 ):
-    Order.objects.filter(id=order_id).update(
-        total_authorized_amount=F("total_authorized_amount") + authorized_amount_to_add,
-        total_charged_amount=F("total_charged_amount") + charged_amount_to_add,
-    )
+    Order.objects.filter(id=order_id).update(**update_fields)
```
fowczarek, line 79: IMO we could get rid of this function.

saleor/order/tests/test_order_actions.py
```diff
@@ -377,6 +378,30 @@ def test_cancel_order(
...
+    ).exists()
+
+    # when
+    expire_order(order, None, manager)
+
+    # then
+    assert order.status == OrderStatus.EXPIRED
+    assert not Allocation.objects.filter(
+        order_line__order=order, quantity_allocated__gt=0
+    ).exists()
```
fowczarek, line 402: We should also check `Voucher` usage limit.

saleor/order/tests/test_order_actions.py
```diff
@@ -377,6 +378,30 @@ def test_cancel_order(
     )
 
 
+def test_expire_order(unconfirmed_order_with_lines, voucher):
```
fowczarek, line 381: We should also add a test for vouchers applied once per customer.

saleor/order/actions.py
```diff
@@ -195,6 +196,26 @@ def cancel_order(
         call_event(send_order_canceled_confirmation, order, user, app, manager)
 
 
+def expire_order(
```
fowczarek, line 199: In PR description we have `Saleor will use to change the unconfirmed and not paid orders into expired orders` maybe we should add validation for this function?

saleor/order/tasks.py
fowczarek: `quetset_in_batches` from migration has `Input queryset should be sorted be pk.` as a requirement. 
Should we also add the same requirement to yours `_queryset_in_batches`?

saleor/order/actions.py
kadewu: I was also wondering about that. The idea of not having validation here is just for future development, maybe we will want to expire with different conditions. But I guess it will not harm to add that validation now.

kadewu:
@fowczarek 
> quetset_in_batches from migration has Input queryset should be sorted be pk. as a requirement.
> Should we also add the same requirement to yours _queryset_in_batches?

`Order` model has already defined `ordering` on unique field `number`. So I believe, that we are already preventing the issue that goes with random order of data.

saleor/order/tasks.py
kadewu: Changed:
Values <0 are now not valid, and 0 will disable expiration of orders (the same as null value would do)

fowczarek: FYI. We have the same strategy for reservations.

saleor/order/tasks.py
```diff
@@ -21,3 +39,75 @@ def send_order_updated(order_ids):
...
+
+def _expire_orders(manager, now):
+    qs = Order.objects.filter(
+        ~Exists(TransactionItem.objects.filter(order=OuterRef("pk"))),
+        ~Exists(Payment.objects.filter(order=OuterRef("pk"))),
+        status=OrderStatus.UNCONFIRMED,
+        channel__expire_orders_after__isnull=False,
+        channel__expire_orders_after__gt=0,
+        channel__expire_orders_after__lte=Extract(now - F("created_at"), "epoch"),
+        total_charged_amount=Decimal(0),
```
korycins, line 89: `total_charged_amount=Decimal(0),` I think this is not needed. If you don't have any payments and any TransactionItem, then we don't have anything that could be charged.

korycins (approved):

fowczarek (approved):

saleor/order/tasks.py
```diff
@@ -1,10 +1,27 @@
...
+from ..payment.models import Payment, TransactionItem
 from ..plugins.manager import get_plugins_manager
+from ..warehouse.management import deallocate_stock_for_orders
+from . import OrderStatus
 from .models import Order
 from .utils import invalidate_order_prices
 
+logger = logging.getLogger(__name__)
+
+ORDER_BATCH_SIZE = 1000
```
8r2y5, line 23: Please provide doc string with explanation why this is `1000`, have you checked memory usage?

saleor/order/tasks.py
```diff
@@ -21,3 +38,81 @@ def send_order_updated(order_ids):
...
+
+
+@app.task(soft_time_limit=60 * 30)
+def expire_orders_task():
+    now = timezone.now()
+    task_name = "expire_orders"
+    manager = get_plugins_manager()
+    try:
+        with celery_task_lock(task_name) as (lock_obj, acquired):
+            if lock_obj.created_at < now - timezone.timedelta(hours=1):
```
8r2y5, line 112: Is there a case when `acquired` is `True` and this here will be also `True`?

saleor/graphql/webhook/subscription_types.py
```diff
@@ -465,6 +465,16 @@ class Meta:
         )
 
 
+class OrderExpired(SubscriptionObjectType, OrderBase):
+    class Meta:
+        root_type = "Order"
+        enable_dry_run = True
+        interfaces = (Event,)
+        description = (
+            "Event sent when order is expired." + ADDED_IN_312 + PREVIEW_FEATURE
```
tomaszszymanski129, line 474: Maybe `Event sent when order becomes expired`? Seems a bit clearer to me, when I was reading that I wondered if this is used in a conditional while checking some orders(?), but that's only triggered when a task marks it as such under some conditions automatically. wdyt?

tomaszszymanski129 (approved):

saleor/graphql/webhook/subscription_types.py
kadewu: Sure, that's sounds better.

saleor/order/tasks.py
kadewu: I think that's such case should not happen, but added that logging just in case. Do you think this check unnecessary?

8r2y5: Logging is fine, I'm thinking of case where both `if` would return `True`. `CeleryTask` has `unique=True` on task name so there can't be race condition case where there would be 2 records. I think I would be ok to just do
```suggestion
            if not acquired:
                if lock_obj.created_at < now - timezone.timedelta(hours=1):
                    logger.error("%s task exceeded 1h working time.", [task_name])
            else:
                _expire_orders(manager, now)
```

8r2y5 (approved):

szczecha:
After QA: 

- The worker is not always starting:

  a) On deployment: since 14.03, no order has changed its status to "Expired". Only webhook events are coming through, such as "Order Created" and "Order Updated" (e.g., canceled). For "Expired" nothing is being received, even though the "ORDER_EXPIRED" event has been added to the asyncEvents and a subscription for this event has been set up.

  b) Locally: yesterday, it seemed that something came through after rebuilding, but today, when testing, orders are not being updated. I’ve changed the setup and rebuilt docker:
```
# Expire orders task setting
BEAT_EXPIRE_ORDERS_AFTER_TIMEDELTA = timedelta(
    seconds=parse(os.environ.get("BEAT_EXPIRE_ORDERS_AFTER_TIMEDELTA", "1 minutes"))
)
```
  In the channel, I set the expiration to 1 minute. However, the task did not start even after a few minutes. After restarting the worker, it runs the task and processes all the orders that were in the queue.


- Creating OrderEvent of type EXPIRED.
- An expired order should be treated in the same way as a canceled order, meaning that no update actions should be allowed on it. This will be reported as a separate task. - Since there are many issues related to handling both canceled and expired orders. Generally, all actions, except for generating an invoice and canceling the expired order, do not work on canceled and expired orders.
- The query "reportProductSales" returns the gross revenue amount, which includes revenue from expired and canceled products. This is likely related to the previous point.

PR #11978, opened 2023-02-10
Webhook custom headers

zedzior:
I want to merge this change, to allow adding custom headers to HTTP webhook requests.

The PR adds `custom_headers` field to `Webhook` model. Only headers in accordance with [RFC5322](https://www.rfc-editor.org/rfc/rfc5322#section-2.2) and `X-*` or `Authorization*` key pattern are allowed. There is a limitation of 5 headers per webhook and 998 characters per header.

# Impact

* [x] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/site/models.py
```diff
@@ -16,6 +16,9 @@
 from .error_codes import SiteErrorCode
 from .patch_sites import patch_contrib_sites
 
+if TYPE_CHECKING:
+    from ..settings import DEFAULT_FROM_EMAIL
+
```
zedzior, line 21: Does it make sense to solve weird fail of mypy for `sender_address = settings.DEFAULT_FROM_EMAIL  # type: ignore` line?

saleor/webhook/validators.py
```diff
@@ -0,0 +1,46 @@
+import re
+from typing import Dict
+
+from django.core.exceptions import ValidationError
+
+HEADERS_NUMBER_LIMIT = 5
+HEADERS_LENGTH_LIMIT = 78
```
zedzior, line 7: Limits to discuss.

saleor/site/models.py
korycins: You placed it under the TYPE_CHECKING, I don't think that this will be accesible in line 120.

saleor/webhook/validators.py
```diff
@@ -0,0 +1,46 @@
...
+        )
+
+    key_chars_allowed = "".join([chr(i) for i in range(33, 127) if i != ord(":")])
+    value_chars_allowed = key_chars_allowed + chr(32) + chr(9)
+
+    for key, value in headers.items():
+        try:
+            header = ": ".join([key, value])
+        except TypeError:
+            raise ValidationError("One of the header can't be converted to string.")
```
korycins, line 27: Can't we try to add here `key`?  I think, that it would be helpful in case of understanding what is wrong

korycins (commented):
@zedzior 
What about extending manifest logic to also handle custom headers from app installation flow?
Also, keep in mind that this changes should be also included in the docs.

saleor/site/models.py
IKarbowiak: Yep, and it also means that there is a lack of a test for the scenario without `sender_address`. Could you add one?

saleor/webhook/validators.py
zedzior: If it can't be converted to string, how can I add it to string message?

saleor/site/models.py
zedzior: I added the test, however can't see any usage of the function in the codebase.

saleor/webhook/validators.py
korycins: So in which scenario, the `key` cannot be converted to string? 
You're getting headers as a JSONString from API, and from manifest  (probably also as a json or jsonstring). So in both scenarios, the key & value need to be serializable

The current implementation will fail, when you will have numeric value for `key` or `value` will be for example a `dict` or `list`, but I don't know which case doesn't allow us to provide a `key` here.

saleor/webhook/validators.py
```diff
@@ -0,0 +1,46 @@
+import re
+from typing import Dict
+
+from django.core.exceptions import ValidationError
+
+HEADERS_NUMBER_LIMIT = 5
+HEADERS_LENGTH_LIMIT = 78
+
+
+def custom_headers_validator(headers) -> Dict[str, str]:
```
korycins, line 10: Would be good to add a typing here, as someone could use the method with `None` (by mistake).

saleor/plugins/webhook/tasks.py
```diff
@@ -324,6 +333,18 @@ def send_webhook_using_http(
         AppHeaders.SIGNATURE: signature,
         AppHeaders.API_URL: build_absolute_uri(reverse("api"), domain),
     }
+
+    if custom_headers:
+        try:
+            custom_headers_validator(custom_headers)
```
maarcingebala, line 339: I think we can skip validation at this point, since we have validation at DB field level and mutation level.

maarcingebala (approved):

IKarbowiak (approved):

saleor/webhook/validators.py
```diff
@@ -0,0 +1,46 @@
...
+
+    https://www.rfc-editor.org/rfc/rfc5322#section-2.2
+    """
+    if len(headers) > HEADERS_NUMBER_LIMIT:
+        raise ValidationError(
+            f"Number of headers exceeds the limit: {HEADERS_NUMBER_LIMIT}."
+        )
+
+    key_chars_allowed = "".join([chr(i) for i in range(33, 127) if i != ord(":")])
+    value_chars_allowed = key_chars_allowed + chr(32) + chr(9)
```
NyanKiyoshi, line 21: Rather than generating the string every time, you could make it a constant

saleor/webhook/validators.py
```diff
@@ -0,0 +1,46 @@
...
+
+    for key, value in headers.items():
+        try:
+            header = ": ".join([key, value])
+        except TypeError:
+            raise ValidationError(f'Header with "{key}" can\'t be converted to string.')
```
NyanKiyoshi, line 27: I don't personally get the point of this string concat? It seems unnecessary

saleor/webhook/validators.py
```diff
@@ -0,0 +1,46 @@
...
+
+        if not set(value).issubset(set(value_chars_allowed)):
+            raise ValidationError(f'Value "{value}" contains invalid character.')
+
+        if len(header) > HEADERS_LENGTH_LIMIT:
+            raise ValidationError(
+                f'"{header}" exceeds the limit of characters: {HEADERS_LENGTH_LIMIT}.'
+            )
+
+        if not re.search(r"(^X-\S*)|(^Authorization\S*)", key):
```
NyanKiyoshi, line 40: Regex doesn't seem necessary, on top of this it doesn't check if it starts with the value (e.g. it would count `Foo-X-` as valid which doesn't seem intended)

```suggestion
        if not (key.startswith("X-") or key.startswith("Authorization")):
```

saleor/webhook/validators.py
```diff
@@ -0,0 +1,46 @@
...
+        if not set(key).issubset(set(key_chars_allowed)):
+            raise ValidationError(f'Key "{key}" contains invalid character.')
+
+        if not set(value).issubset(set(value_chars_allowed)):
+            raise ValidationError(f'Value "{value}" contains invalid character.')
+
+        if len(header) > HEADERS_LENGTH_LIMIT:
+            raise ValidationError(
+                f'"{header}" exceeds the limit of characters: {HEADERS_LENGTH_LIMIT}.'
+            )
```
NyanKiyoshi, line 38: Since you are scanning a N chars string in `issubset`, I would suggest checking length first as it's a cheaper check

korycins (approved):

saleor/webhook/validators.py
zedzior: It is to check if both key and value can be converted to string and the length of the header.

saleor/webhook/validators.py
zedzior: "^" character in regex stands for "starts with"

maarcingebala (approved):

saleor/graphql/webhook/mutations/webhook_update.py
```diff
@@ -52,6 +59,15 @@ class WebhookUpdateInput(graphene.InputObjectType):
         + PREVIEW_FEATURE,
         required=False,
     )
+    custom_headers = JSONString(
+        description=f"Custom headers, which will be added to HTTP request. "
+        f"There is a limitation of {HEADERS_NUMBER_LIMIT} headers per webhook "
+        f"and {HEADERS_LENGTH_LIMIT} characters per header."
+        f'Only "X-*" and "Authorization*" keys are allowed.'
+        + ADDED_IN_312
+        + PREVIEW_FEATURE,
```
NyanKiyoshi, line 68: Why not use f-string?

```suggestion
        description=(
            f"Custom headers, which will be added to HTTP request. "
            f"There is a limitation of {HEADERS_NUMBER_LIMIT} headers per webhook "
            f"and {HEADERS_LENGTH_LIMIT} characters per header."
            f'Only "X-*" and "Authorization*" keys are allowed.'
            f"{ADDED_IN_312}{PREVIEW_FEATURE}"
        ),
```

saleor/graphql/webhook/tests/mutations/test_webhook_create.py
```diff
@@ -286,3 +291,31 @@ def test_webhook_create_inherit_events_from_query(
         data["webhook"]["syncEvents"][0]["eventType"]
         == WebhookEventTypeSyncEnum.PAYMENT_LIST_GATEWAYS.name
     )
+
+
+def test_webhook_create_invalid_custom_headers(app_api_client):
```
NyanKiyoshi, line 296: You are missing a test checking if updating a webhook validates the inputs

saleor/app/tests/test_installation_utils.py
```diff
@@ -409,6 +412,30 @@ def test_install_app_webhook_incorrect_query(
...
+    with pytest.raises(ValidationError) as excinfo:
+        install_app(app_installation, activate=True)
+
+    error_dict = excinfo.value.error_dict
+    assert "webhooks" in error_dict
+    assert (
+        error_dict["webhooks"][0].message == 'Invalid custom headers: "InvalidKey" '
+        'does not match allowed key pattern: "X-*" or "Authorization*".'
+    )
+
```
NyanKiyoshi, line 437:
```suggestion
    assert (
        error_dict["webhooks"][0].message == (
            'Invalid custom headers: "InvalidKey" '
            'does not match allowed key pattern: "X-*" or "Authorization*".'
        )
    )

```

NyanKiyoshi (approved):

PR #11996, opened 2023-02-13
Default to serving 4K images instead of originals

patrys:
This changes the default behavior to return a 4K thumbnail instead of the original which could be tens of MiB. Not requesting an appropriate thumbnail is among API users' most common mistakes.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/account/types.py
+++ b/saleor/graphql/account/types.py
@@ -488,7 +488,7 @@ class User(ModelObjectType[models.User]):
         if not root.avatar:
             return
 
-        if not size:
+        if size == 0:
             return Image(url=root.avatar.url, alt=None)
 
         format = get_thumbnail_format(format)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ All notable, unreleased changes to this project will be documented in this file.
        }
     ```
 
-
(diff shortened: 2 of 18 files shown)


IKarbowiak (approved):

saleor/graphql/account/types.py
```diff
@@ -475,25 +475,29 @@ def _resolve_orders(orders):
...
-    def resolve_avatar(root: models.User, info: ResolveInfo, size=None, format=None):
+    def resolve_avatar(
+        root: models.User, info: ResolveInfo, size: Optional[int] = None, format=None
+    ):
         if not root.avatar:
             return
 
-        if not size:
+        if size is None:
+            size = 4096
```
fowczarek, line 485: Maybe we could store this value in settings(e.g. `DEFAULT_THUMBNAIL_SIZE`)?

fowczarek (approved):

maarcingebala (approved):

saleor/graphql/account/types.py
patrys: I can turn it into a constant, but I don't think it belongs in settings. We don't want users to change it.

PR #12045, opened 2023-02-16
Fix order filtering by charge status

### LINKED ISSUE #12012: Bug: Order filter not working with payment status

I'm trying to filter orders with `Fully refunded` payment status. 

1. Create an order and fully refund it

I should get orders all orders with fully refund payment status

### Logs

_No response_

### Environment

Saleor version: 3.9+

jakubkuc:
I want to merge this change because filter by payment status was filtering only active payments so it skipped `FULLY_REFUNDED` orders since they are deactivated when they are fully refunded.

It fixes #12012

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/filters.py
+++ b/saleor/graphql/order/filters.py
@@ -28,7 +28,6 @@ from .enums import OrderAuthorizeStatusEnum, OrderChargeStatusEnum, OrderStatusF
 
 def filter_payment_status(qs, _, value):
     if value:
-        print(value)
         lookup = Q(payments__is_active=True, payments__charge_status__in=value)
         if ChargeStatus.FULLY_REFUNDED in value:
             lookup |= Q(payments__charge_status=ChargeStatus.FULLY_REFUNDED)
--- a/saleor/graphql/order/filters.py
+++ b/saleor/graphql/order/filters.py
@@ -28,13 +28,11 @@ from .enums import OrderAuthorizeStatusEnum, OrderChargeStatusEnum, OrderStatusF
 
 def filter_payment_status(qs, _, value):
     if value:
+        print(value)
+        lookup = Q(payments__is_active=True, payments__charge_status__in=value)
         if ChargeStatus.FULLY_REFUNDED in value:
-            qs = qs.filter(
-                Q(payments__is_active=True, payments__charge_status__in=value)
-                | Q(payments__charge_status=ChargeStatus.FULLY_REFUNDED)
-            )
-        else:
-            qs = qs.filter(payments__is_active=True, payments__charge_status__in=value)
+            lookup |= Q(payments__charge_status=ChargeStatus.FULLY_REFUNDED)
+        qs = qs.filter(lookup)
     return qs
 
 
--- a/saleor/graphql/order/filters.py
+++ b/saleor/graphql/order/filters.py
@@ -27,7 +27,7 @@ from .enums import OrderAuthorizeStatusEnum, OrderChargeStatusEnum, OrderStatusF
 
 def filter_payment_status(qs, _, value):
     if value:
-        qs = qs.filter(payments__is_active=True, payments__charge_status__in=value)
+        qs = qs.filter(payments__charge_status__in=value)
     return qs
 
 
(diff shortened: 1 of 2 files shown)


fowczarek (approved):

kadewu (approved):

saleor/graphql/order/filters.py
```diff
@@ -27,7 +27,7 @@
 
 def filter_payment_status(qs, _, value):
     if value:
-        qs = qs.filter(payments__is_active=True, payments__charge_status__in=value)
+        qs = qs.filter(payments__charge_status__in=value)
```
IKarbowiak, line 30: @korycins what do you think? Won't we end up with plenty of payments that are irrelevant? Maybe instead of removing `payments__is_active=True` condition, we should handle this specific case?

IKarbowiak: Also, it could be rewritten using `Exists` and `OuterRef`

korycins: If you will filter out by `NOT_CHARGED` you will probably get incorrect orders (for example, order with two payments, one active&captured, second disabled&NOT_CHARGED (the default status). 
I would suggest to extract  `FULLY_REFUNDED` to separate if condition.

saleor/graphql/order/filters.py
```diff
@@ -27,7 +28,13 @@
...
 def filter_payment_status(qs, _, value):
     if value:
-        qs = qs.filter(payments__is_active=True, payments__charge_status__in=value)
+        if ChargeStatus.FULLY_REFUNDED in value:
+            qs = qs.filter(
+                Q(payments__is_active=True, payments__charge_status__in=value)
+                | Q(payments__charge_status=ChargeStatus.FULLY_REFUNDED)
+            )
+        else:
+            qs = qs.filter(payments__is_active=True, payments__charge_status__in=value)
```
IKarbowiak, line 37: LGTM 👍 

Could be simplified a little:
```suggestion
        lookup = Q(payments__is_active=True, payments__charge_status__in=value)
        if ChargeStatus.FULLY_REFUNDED in value:
            lookup |=  Q(payments__charge_status=ChargeStatus.FULLY_REFUNDED)
        qs = qs.filter(lookup)
```

saleor/graphql/order/filters.py
```diff
@@ -27,7 +28,13 @@
 
 def filter_payment_status(qs, _, value):
     if value:
-        qs = qs.filter(payments__is_active=True, payments__charge_status__in=value)
+        if ChargeStatus.FULLY_REFUNDED in value:
+            qs = qs.filter(
+                Q(payments__is_active=True, payments__charge_status__in=value)
```
korycins, line 33: Is it correct?
You are going to filter by `ChargeStatus.FULLY_REFUNDED` and also by `payments__is_active=True, payments__charge_status__in=value`. But the filter `ChargeStatus.FULLY_REFUNDED` will return you all records, including the ones from the `payments__is_active=True, payments__charge_status__in=value`.
Maybe I am missing something, but I would say that `qs = qs.filter(payments__charge_status=ChargeStatus.FULLY_REFUNDED)` should be enough here

saleor/graphql/order/filters.py
```diff
@@ -27,7 +28,13 @@
 
 def filter_payment_status(qs, _, value):
     if value:
-        qs = qs.filter(payments__is_active=True, payments__charge_status__in=value)
+        if ChargeStatus.FULLY_REFUNDED in value:
```
korycins, line 31: Shouldn't be `if ChargeStatus.FULLY_REFUNDED == value`?

saleor/graphql/order/filters.py
jakubkuc: `value` is a list that can look like this for example: `['fully-refunded', 'partially-refunded']` . That is why I do it this way so filter will return both orders `FULLY_REFUNDED` and others.

saleor/graphql/order/filters.py
jakubkuc: like I commented above. value is a list that can look like this for example: `['fully-refunded', 'partially-refunded']`

korycins (approved):

szczecha (approved):

PR #12058, opened 2023-02-17
Make custom header name case-insensitive.

zedzior:
I want to merge this change, because currently custom header names are case-sensitive. This PR also ensure the header names are stored in database as lower-case strings.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/webhook/validators.py
+++ b/saleor/webhook/validators.py
@@ -5,7 +5,8 @@ from django.core.exceptions import ValidationError
 HEADERS_NUMBER_LIMIT = 5
 HEADERS_LENGTH_LIMIT = 998
 KEY_CHARS_ALLOWED = (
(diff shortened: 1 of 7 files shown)


IKarbowiak (approved):

saleor/webhook/validators.py
```diff
@@ -24,6 +23,7 @@ def custom_headers_validator(headers: Dict[str, str]) -> Dict[str, str]:
             f"Number of headers exceeds the limit: {HEADERS_NUMBER_LIMIT}."
         )
 
+    headers = {k.lower(): v for k, v in headers.items()}
```
korycins, line 26: In line 28, you're checking if the `key` is `str`, but before this you're calling `.lower()` which means, that this can fail, right?

maarcingebala (approved):

Issue #12056, opened 2023-02-17
Bug: Specific products voucher is not included in Checkout.discount


maarcingebala:
### What are you trying to achieve?

When a voucher for specific products is applied, it is propagated into checkout lines, but `Checkout.discount` is set to `0`. This is incorrect, and we should calculate the discount amount in the `Checkout.discount` field as well. This has been broken [since Saleor 3.6](https://github.com/saleor/saleor/pull/9699). This may be considered a breaking change as `Checkout.discount` might have been used by frontend to display discounts.

To fix this we can use a function that was originally used to calculate that in Saleor 3.1: https://github.com/saleor/saleor/blob/3.0/saleor/checkout/utils.py#L373

- Double-check if order calculations work as expected after this change - (Checkout.discount value might have been used in internal calculations) - especially with different types of vouchers.
- Verify with Avatax.

### Steps to reproduce the problem

1. Create a checkout with a voucher for specific products.
2. Inspect the `Checkout.discount` value.

### What did you expect to happen?

`Checkout.discount` amount should be set when "specific products" voucher is used.

### Environment

Saleor version: 3.9

PR #12106, opened 2023-02-23
Include specific products voucher in checkout discount

IKarbowiak:
Attach the discount from `specific product` voucher, and `apply once per order` vouchers to `checkout.discount`. 

Fixes [#12056](https://github.com/saleor/saleor/issues/12056)

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/checkout/base_calculations.py
```diff
@@ -192,7 +192,12 @@ def base_checkout_total(
     discounts: Iterable[DiscountInfo],
     lines: Iterable["CheckoutLineInfo"],
 ) -> Money:
-    """Return the total cost of the checkout."""
+    """Return the total cost of the checkout.
+
+    The price includes sales, specific voucher and applied once per order
+    voucher discounts.
+    The price does not include the entire order discount.
```
korycins, line 199: Don't know if this is already done, but maybe it would be worth having a test, that will fail if someone change the `base_checkout_total` to include voucher.entire order?

korycins: What about voucher for shipping? Maybe would be worth to mention it here also

saleor/checkout/base_calculations.py
```diff
@@ -237,12 +238,49 @@ def base_checkout_subtotal(
     return sum(line_totals, zero_money(currency))
 
 
+def checkout_total(
+    checkout_info: "CheckoutInfo",
+    discounts: Iterable[DiscountInfo],
+    lines: Iterable["CheckoutLineInfo"],
+) -> Money:
+    """Return the total cost of the checkout including discounts and vouchers."""
```
korycins, line 246: Is this a function that should be called, when we don't have any active flat rate/tax plugin/tax app? If yes, maybe it would be worth mentioning this in the description. This could be hard to understand what kind of function should be used in a specific scenario.

saleor/checkout/base_calculations.py
```diff
@@ -256,32 +294,32 @@ def apply_checkout_discount_on_checkout_line(
         or voucher.apply_once_per_order
         or voucher.type in [VoucherType.SHIPPING, VoucherType.SPECIFIC_PRODUCT]
     ):
-        return line_unit_price
+        return line_price
 
     line_quantity = checkout_line_info.line.quantity
     total_discount_amount = checkout_info.checkout.discount_amount
-    line_total_price = line_unit_price * line_quantity
+    line_total_price = line_price * line_quantity if is_unit_price else line_price
```
korycins, line 301: Question: Wouldn't be better to always pass `line_total_price` to `apply_checkout_discount_on_checkout_line` instead of using `is_unit_price` flag? It looks like the perfect place for making the mistake 😄  In places where we call this function, and where it uses a unit price, we could simply call unit_price* quantity, right?

saleor/checkout/utils.py
```diff
@@ -445,22 +446,33 @@ def get_prices_of_discounted_specific_product(
...
+    return line_prices
+
 
-    for line_info in discounted_lines:
+def get_base_lines_prices(
+    checkout_info: "CheckoutInfo",
+    lines: Iterable["CheckoutLineInfo"],
+    discounts: Optional[Iterable[DiscountInfo]],
+):
+    """Get base total price of provided checkout lines without discount applied."""
```
korycins, line 463: "without voucher discount applied." right? Sale discount is applied.

saleor/plugins/manager.py
```diff
@@ -383,6 +383,16 @@ def calculate_checkout_line_total(
             checkout_info.channel,
             discounts,
         )
+        # apply entire order discount
+        default_value = base_calculations.apply_checkout_discount_on_checkout_line(
```
korycins, line 387: Is it expected? The `checkout_total` description says, that sale and voucher are applied to the amount.

fowczarek (approved):

saleor/plugins/manager.py
IKarbowiak: It doesn't include the `entire_order` voucher with `apply_once_per_order` set to `False`. I can update the descriptions of the `calculate_base_line_total_price` as it's not clear.

saleor/checkout/utils.py
IKarbowiak: Yes, you are right. Thanks. I will change that.

saleor/checkout/base_calculations.py
IKarbowiak: I don't understand what kind of test you want to add 🤔.

IKarbowiak: I will update the description

saleor/checkout/base_calculations.py
IKarbowiak: You're right! I'll change that!

saleor/plugins/manager.py
IKarbowiak: I updated the docstring of `calculate_base_line_total_price`

saleor/checkout/base_calculations.py
IKarbowiak: Ok, I get what you mean. But there is already a test that will fail because of that

maarcingebala (approved):

IKarbowiak:
[Branch with docs update](https://github.com/saleor/saleor-docs/tree/discount-page) - still in progress

szczecha (approved):

8r2y5 (approved):

Issue #12093, opened 2023-02-21
[RFC]Address validation for country area


jakubkuc:
The proposed solution for #11789 and some other problems with saving country area in addresses.

We want to be able to save country areas in Address objects even if they are not allowed by `i18n` plugin. 
Also, we would like to be able to pass alternative namings for country areas and map them to unify output seven in db.  

### General assumptions
- we will create a mapping for known country areas for different countries
- we will leave the i18n validation as it is now 
- additionally, if `countryArea` is provided in the input it will check the created map if country area provided is valid

If `i18n` is not allowing country area for a given country it will be saved in db if it is present in our map. 
If `i18n` is allowing country area for given country but it doesn't pass the validation - it will also be put through our map to check if we consider it valid. 

### Notes and issues found 

1.  AddressUpdate mutation does not work optimal. It Overwrites all the fields instead of only given ones. 
e.g : `AddressUpdate` mutation if given fields in input will update those fields with new values and update the ones that are not provided with "Null". ATM it is handled in the dashboard by always sending all fields values. 

2. The second problem (but connected to the first) is old way (connected to templates used in the past) of handling Address objects by using django forms. We should think about a new way of handling all Address creation/updates without django forms. This way saleor would be more flexible during validation and all address operations could be more consistent than now. 

3. We would need a country areas and their allowed alternatives handed to us so we can add them to saleor core internal mapping mechanism. 

DOD:
- If country area is allowed by the core country area map, user is able to add country area even if country area field is not allowed by `i18n` validation.
- If country area is allowed by core country area map, user is able to add country area even if it's considered invalid by `i18n` validation
- If user is providing all fields according to `i18n` rules everything works the same as before

szymi-:
So if my understanding is correct, if I input a country area that matches what's in Saleor's map of valid country areas, but does not match exactly what is in `i18n`, then the value from `i18n` is saved in Saleor's database. What will I get when I read the value later, e.g. via a graphql query? Will the original value, or value from `i18n`?

jakubkuc:
@szymi- the main point is to keep the output consistent. So in this case it will be what is stored in db. We want to be flexible in terms of entering input, to figure out which country area the user has in mind, without enforcing him to spell it correctly. E.g. like inputting `California` will be saved and fetched as `CA` since that is expected by the local post office.

szymi-:
It would be good have a way to retrieve the originally input value along with what's in the database.

PR #12147, opened 2023-02-28
Address validation country area additional mapping

jakubkuc:
I want to merge this change because it resolves issue #12093

The full map is not yet provided. There is only one example for `CH` country area `Graubünden`.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/account/i18n.py
+++ b/saleor/account/i18n.py
@@ -40,7 +40,9 @@ AREA_TYPE = {
 
 CountryAreaValues = namedtuple("CountryAreaValues", ["value", "choices"])
 
-COUNTRY_AREA_MAP = {"CH": CountryAreaValues("Graubünden", ["Graubünden"])}
+COUNTRY_AREA_MAP = {"CH": [
+    CountryAreaValues("Graubünden", ["Graubünden"]),
+]}
 
 
 class PossiblePhoneNumberFormField(forms.CharField):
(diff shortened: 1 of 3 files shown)


saleor/account/i18n.py
```diff
@@ -38,6 +38,10 @@
     "zip": "ZIP code",
 }
 
+CountryAreaValues = namedtuple("CountryAreaValues", ["value", "choices"])
+
+COUNTRY_AREA_MAP = {"CH": CountryAreaValues("Graubünden", ["Graubünden"])}
```
korycins, line 43: I may not fully understand the requirements, but do we really need to have and maintain such a map on the core side? It seems like we would have to cover a large number of countries and regions, and manually paste them into the Saleor code. Additionally, we would have to update the list of choices whenever someone reports a missing value, and create a new release with an extended map that would need to be ported to all Saleor versions greater than the requested one.

IKarbowiak (approved):

saleor/account/i18n.py
maarcingebala, line 43: There should be another mapping nested per country, as multiple country areas will exist. The current structure supports only a single country area per country.

IKarbowiak (approved):

saleor/account/i18n.py
```diff
@@ -38,6 +38,12 @@
     "zip": "ZIP code",
 }
 
+CountryAreaValues = namedtuple("CountryAreaValues", ["value", "choices"])
+
+COUNTRY_AREA_MAP = {"CH": [
```
8r2y5, line 43: Please run black or pre-commit

8r2y5 (approved):

maarcingebala:
We still didn't get any specific examples of exceptions that should be included in the map. I think let's close this PR for now and revisit if this is really needed by someone, as for now I believe it's not a priority.

PR #12155, opened 2023-02-28
Add TrasactionEventType.INFO for TransactionEvent

korycins:
I want to merge this change because it adds a new type: INFO. The INFO type will be assigned to all existing events that are created from the `transactionCreate` and `transactionUpdate` mutations, by using `transactionEventInput`.

It also fixes the problem with the status field. Saleor raises an exception when status was not provided. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/order/types.py
```diff
@@ -521,7 +521,7 @@ def resolve_discount(root: models.OrderEvent, info):
 
     @staticmethod
     def resolve_status(root: models.OrderEvent, _info):
-        return root.parameters.get("status")
+        return root.parameters.get("status") or ""
```
kadewu, line 524: I believe this is not working correctly, cause empty string is not part of `TransactionEventStatus`.
Also I think adding one test for that would be good.

fowczarek: In other resolvers, we return None. Why do we return `""` here?

fowczarek (approved):

saleor/graphql/order/types.py
korycins: @kadewu  good catch. Thanks. Will extend the tests also 👍

kadewu (approved):

Issue #12151, opened 2023-02-28
Bug: Unable to calculate taxes for JPY in Avatax plugin


IKarbowiak:
### What are you trying to achieve?

The error is raised for Avatax tax calculation for JPY currency.

### Steps to reproduce the problem

1. Create a channel for Japan, assign the shipping zone for this channel that support Japan
2. Configure Avatax for Japan
3. Configure taxes for Japan - Use the tax app for this channel
4. Chose a product and add the availability for Japan channel for this product
5. Create checkout with the product available in Japan
6. Try to assign the delivery method to this checkout - the error will be raised (`"'Money' object has no attribute 'gross'"`)

### What did you expect to happen?

The taxes should be correctly calculated.

### Logs

https://saleor.sentry.io/issues/3963210588/?project=6417854&referrer=slack

### Environment

Saleor version: 3.9

PR #12158, opened 2023-02-28
Fix Avalara tax calculation for JPY currency

IKarbowiak:
Fix problem with calculating taxes by Avalara plugin for `JPY` currency.

Fixes https://github.com/saleor/saleor/issues/12151

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/plugins/avatax/plugin.py
+++ b/saleor/plugins/avatax/plugin.py
@@ -424,9 +424,11 @@ class AvataxPlugin(BasePlugin):
             net = Decimal(line["lineAmount"])
 
             if currency == "JPY" and prices_entered_with_tax():
-                base_value._setup()
+                if isinstance(base_value, SimpleLazyObject):
+                    base_value = base_value._setupfunc()  # type: ignore
+
                 line_gross = Money(
-                    base_value._wrapped.amount - discount_amount, currency=currency
+                    base_value.amount - discount_amount, currency=currency
                 )
                 line_net = Money(amount=line_gross.amount - tax, currency=currency)
             else:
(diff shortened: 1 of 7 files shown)


fowczarek (approved):

saleor/plugins/avatax/plugin.py
```diff
@@ -421,15 +424,20 @@ def _calculate_checkout_line_total_price(
             net = Decimal(line["lineAmount"])
 
             if currency == "JPY" and prices_entered_with_tax():
-                line_gross = base_value.gross
+                base_value._setup()
+                line_gross = Money(
+                    base_value._wrapped.amount - discount_amount, currency=currency
```
patrys, line 429: Why call `._setup()` instead of using `base_value.amount` here and letting the simple object do its thing? If we need to deal with SimpleLazyObject directly, why do we even pass it as a lazy value?

maarcingebala (approved):

saleor/plugins/avatax/plugin.py
IKarbowiak: We are passing this as a lazy value, because we are using this `base_value` only for `JPY` currency or if there are no lines in response. So to avoid calling this method if it's not needed, the value is provided by `SimpleLazyObject`.

patrys: But the whole point of `SimpleLazyObject` is that it's transparent. Why do we call `_setup()` and reach into the `_wrapped` object directly?

patrys (approved):

Issue #12148, opened 2023-02-28
Bug: Can't resolve order from OrderEvent


fowczarek:
### What are you trying to achieve?

GraphQL was unable to resolve the Order from the OrderEvent created before changing the `order.id` from `int` to `UUID`.

### Steps to reproduce the problem

1. Create OrderEvent with Order.id as `int` (e.g. Saleor 3.1) 
2. Migrate to post 3.2 ([PR#9324](https://github.com/saleor/saleor/pull/9324/files))
3. Try to fetch an order from this event. 

### What did you expect to happen?

I should correctly receive an Order from this OrderEvent. 

### Logs

```
AttributeError: 'int' object has no attribute 'replace'
  File "graphql/execution/executor.py", line 452, in resolve_or_error
    return executor.execute(resolve_fn, source, info, **args)
  File "graphql/execution/executors/sync.py", line 16, in execute
    return fn(*args, **kwargs)
  File "saleor/graphql/order/types.py", line 428, in resolve_related_order
    return OrderByIdLoader(info.context).load(UUID(order_pk))
  File "uuid.py", line 174, in __init__
    hex = hex.replace('urn:', '').replace('uuid:', '')
```

### Environment

Saleor version: Above 3.2
OS and version: Cloud

fowczarek:
Internal discussion:
https://saleorcommerce.slack.com/archives/C03GQ5FD7FY/p1677572494863789

PR #12161, opened 2023-02-28
Fix loading related_order when related id is the old order_id

kadewu:
Fix https://github.com/saleor/saleor/issues/12148

If provided `related_order_pk` is not parsable by UUID, then we can assume that's old integer id.
In migration `order_0128` all old integer id's were copied to the `number` field, so we can use `number` to search for order.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/dataloaders.py
+++ b/saleor/graphql/order/dataloaders.py
@@ -35,6 +35,19 @@ class OrderByIdLoader(DataLoader):
         return [orders.get(order_id) for order_id in keys]
 
 
+class OrderByNumberLoader(DataLoader):
+    context_key = "order_by_number"
+
+    def batch_load(self, keys):
+        orders = Order.objects.using(self.database_connection_name).filter(
+            number__in=keys
+        )
+        orders_by_number = defaultdict(Order)
+        for order in orders:
+            orders_by_number[order.number] = order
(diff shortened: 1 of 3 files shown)


korycins (approved):

jakubkuc (approved):

saleor/graphql/order/dataloaders.py
```diff
@@ -35,6 +35,19 @@ def batch_load(self, keys):
...
+class OrderByNumberLoader(DataLoader):
+    context_key = "order_by_number"
+
+    def batch_load(self, keys):
+        orders = Order.objects.using(self.database_connection_name).filter(
+            number__in=keys
+        )
+        orders_by_number = defaultdict(Order)
+        for order in orders:
+            orders_by_number[order.number] = order
```
IKarbowiak, line 47: Instead of that, you can just do: `Order.objects.in_bulk(field_name="number")`

IKarbowiak (approved):

saleor/graphql/order/dataloaders.py
kadewu: Much cleaner, thanks!

szczecha (approved):

PR #12276, opened 2023-03-09
Add directive to group API objects by categories

maarcingebala:
This PR changes the schema printer to include directives that allow grouping objects in the [API reference](https://docs.saleor.io/docs/3.x/api-reference/) by a given category. To begin with this PR only covers the "account" module, the rest will be added in the following PRs.

After this change, the API reference contains meaningful groups Authentication, Users and Miscellaneous:

https://user-images.githubusercontent.com/5421321/224074021-bda8ee5e-fae7-4976-a22b-700f42aa2c91.mov

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/core/mutations.py
```diff
@@ -157,6 +158,7 @@ def __init_subclass_with_meta__(
         cls,
         auto_permission_message=True,
         description=None,
+        doc_category=None,
```
korycins, line 161: Wouldn't be good to extend the `ModelMutation` to automatically assign the correct `doc_category` based on model that we provided? This should simplify the creation of the mutation and the amount of work to update all existing ModelMutations. The input `doc_category` could be there and if not provided we could use map.
Something like this one:
```
{
   "User": DOC_CATEGORY_USERS,
   "Group": DOC_CATEGORY_USERS
   "Product": DOC_CATEGORY_PRODUCTS
}

maarcingebala: I'll try to add something like that; it should help with mutations. Maybe something similar could be used for model-based types. For many other types, it will have to be added manually anyway.

jakubkuc (approved):

IKarbowiak (approved):

korycins (approved):

PR #12286, opened 2023-03-10
Change the way of storing the transaction amounts

korycins:
I want to merge this change because this pr changes the way of handling the transaction's amount.

Currently, the main branch flow uses a “bucket” approach for transaction amounts. This means that we transfer funds between authorized, charged, refunded and canceled amounts, for example, by moving 10 USD from authorized to charged. The sum of all amounts (authorized, charged, refunded, canceled) gives us the total amount of the transaction.
The feature branch (#10350), however, records the status of the amounts. For example, if a transaction is authorized for 25 USD and we call charge for 20 USD, the transaction will store `(authorizedAmount=25USD, chargedAmount=20USD)`.

We decided to return to the “bucket flow” currently implemented on the main branch, but with some improvements. Specifically, we will add pending amounts to the flow. The sum of all transaction item amounts, including pending amounts, will give us the total amount of the transaction. The transactionEvents will be used to always return correct amounts assigned to the transaction.

For example, if we authorize an Adyen transaction for 25 USD and a staff user calls to charge 15 USD from the Saleor dashboard panel, the result will be `TransactionItem(authorizedAmount=10USD, pendingChargeAmount=15USD, chargedAmount=0)`. When we receive a report via `transactionEventReport` mutation, we will transfer 15 USD to `chargedAmount` if the charge is successful, or to `authorizedAmount` if the charge fails.
The table below provides two examples that should clarify this further. Columns 1-3 show different approaches, while column 0 defines the action taken for the transaction.
image

The changes for recalculation have impact on `transactionCreate`, `transactionUpdate` and `transactionEventReport`.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/order/migrations/0164_auto_20230310_1118.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def rename_order_events(apps, _schema_editor):
+    OrderEvent = apps.get_model("order", "OrderEvent")
+    OrderEvent.objects.filter(type="transaction_void_requested").update(
+        type="transaction_cancel_requested"
+    )
+    OrderEvent.objects.filter(type="transaction_capture_requested").update(
+        type="transaction_charge_requested"
+    )
```
IKarbowiak, line 13: We should ensure 0 downtime here cc. @kadewu

saleor/order/utils.py
```diff
@@ -875,16 +875,19 @@ def update_order_charge_status(order: Order):
...
-    if total_charged == 0 and total_gross == Decimal(0):
+    if total_charged == 0 and current_total_gross == Decimal(0):
         order.charge_status = OrderChargeStatus.FULL
     elif total_charged <= 0:
         order.charge_status = OrderChargeStatus.NONE
-    elif total_charged < total_gross:
+    elif total_charged < current_total_gross:
         order.charge_status = OrderChargeStatus.PARTIAL
-    elif total_charged == total_gross:
+    elif total_charged == current_total_gross:
```
IKarbowiak, line 890: This condition can be mixes with the first one

IKarbowiak (approved):

saleor/order/migrations/0164_auto_20230310_1118.py
kadewu: Comment: we are going to re-do this migration in the next version after the release, instead of the previous version patch approach, as we don't want to change `Event.type` in the previous version.

kadewu (approved):

saleor/order/migrations/0164_auto_20230310_1118.py
IKarbowiak: Ok, but the `update` will lock the whole table AFAIK, and it should be improved.

kadewu: Yes right, it should be run in batches. Thanks!

szczecha (approved):

Issue #12024, opened 2023-02-14
[Support for payment apps] Add changes to checkout


korycins:
Required permission: -
```graphql
type Checkout {
  """
  Deprecated, use updatedAt instead.
  """
  lastChange: DateTime! 
  updatedAt: DateTime!
  """
  The difference between the paid and the checkout total amount.
  """
  totalBalance: Money
  
    """
  The authorize status of the checkout.
  
  Added in Saleor 3.13.
  
  Note: this API is currently in Feature Preview and can be subject to changes at later point.
  """
  authorizeStatus: CheckoutAuthorizeStatusEnum!

  """
  The charge status of the checkout.
  
  Added in Saleor 3.13.
  
  Note: this API is currently in Feature Preview and can be subject to changes at later point.
  """
  chargeStatus: CheckoutChargeStatusEnum!
}

enum CheckoutChargeStatusEnum {
  NONE
  PARTIAL
  FULL
  OVERCHARGED
}

enum CheckoutAuthorizeStatusEnum {
  NONE
  PARTIAL
  FULL
}

input CheckoutFilterInput {
  ...
  updatedAt: DateRangeInput
  authorizeStatus: [CheckoutAuthorizeStatusEnum!]
  chargeStatus: [CheckoutChargeStatusEnum!] 
}
```

`checkout.availablePaymentGateways` should include the app that are subscribed to `transaction-initialize-session`

PR #12300, opened 2023-03-14
Add checkout payment statuses. Connect availablePaymentGateway with transaction flow

korycins:
I want to merge this change because it covers #12024

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/checkout/migrations/0058_auto_20230313_1602.py
```diff
@@ -0,0 +1,41 @@
...
+            model_name="checkout",
+            name="authorize_status",
+            field=models.CharField(
+                choices=[
+                    ("none", "The checkout is not authorized"),
+                    ("partial", "The checkout is partially authorized"),
+                    ("full", "The checkout is fully authorized"),
+                ],
+                db_index=True,
+                default="none",
```
IKarbowiak, line 22: This will set `none` value only on existing rows, we should also add the SQL operation for setting the default `none` value on this column

saleor/checkout/migrations/0058_auto_20230313_1602.py
```diff
@@ -0,0 +1,41 @@
...
+            name="charge_status",
+            field=models.CharField(
+                choices=[
+                    ("none", "The checkout is not charged"),
+                    ("partial", "The checkout is partially charged"),
+                    ("full", "The checkout is fully charged"),
+                    ("overcharged", "The checkout is overcharged"),
+                ],
+                db_index=True,
+                default="none",
```
IKarbowiak, line 37: Same here

IKarbowiak (approved):

fowczarek (approved):

saleor/checkout/migrations/0058_auto_20230313_1602.py
korycins: - checkout on main
- populatedb on main
- checkout to this branch
- ./manage.py migrate
- and psql returns data as below:
image

So it looks like, the status is populated for all existing fields. 
Also we will handle a zero downtime migrations in separate PR so we will double check this.

cc @kadewu

saleor/checkout/actions.py
```diff
@@ -0,0 +1,25 @@
...
+    discounts: Optional[Iterable["DiscountInfo"]],
+    manager: "PluginsManager",
+):
+    if not transaction.checkout_id:
+        return
+    checkout = cast(Checkout, transaction.checkout)
+    lines, _ = fetch_checkout_lines(checkout)
+    discounts = discounts or []
+    checkout_info = fetch_checkout_info(checkout, lines, discounts, manager)
+    fetch_checkout_data(
```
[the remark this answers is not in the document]
korycins: `discounts: Iterable["DiscountInfo"] = None,` Wrong typing here. It needs to be Optional as you assign `None` here. So I will still need to to make `discounts or []` as I don't want to assign `[]` as a default parameter.

PR #12319, opened 2023-03-16
Use replicas on order queries

tomaszszymanski129:
I want to merge this change because order queries used to be executed on the writer which shouldn't be the case.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/resolvers.py
+++ b/saleor/graphql/order/resolvers.py
@@ -9,32 +9,36 @@ from ...order.events import OrderEvents
 from ...order.models import OrderEvent
 from ...order.utils import sum_order_totals
 from ..channel.utils import get_default_channel_slug_or_graphql_error
+from ..core.context import get_database_connection_name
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [CLOUD-2743 use read replica when querying orders](https://app.clickup.com/t/2549495/CLOUD-2743)

IKarbowiak (approved):

fowczarek (approved):
Do you consider passing to resolvers only connection_name instead of the whole context?

tomaszszymanski129:
> Do you consider passing to resolvers only connection_name instead of the whole context?

in some other resolvers we use that exact pattern, however that's a good point, we don't need an entire object.

tomaszszymanski129:
@fowczarek @IKarbowiak changed the passing context pattern to pass only connection names. Refactored also product resolvers where we had the same thing.

8r2y5 (approved):

PR #12364, opened 2023-03-20
Fix slug validation and generation for Attribute create

jakubkuc:
I want to merge this change because it fixes a problem where creating Attribute with two similar values returned ValidationError when it shouldn't. 

Example valid input that was raising error: 
```{
  "input": {
    "availableInGrid": true,
    "entityType": null,
    "filterableInDashboard": true,
    "filterableInStorefront": true,
    "inputType": "DROPDOWN",
    "name": "test",
    "slug": "test",
    "storefrontSearchPosition": null,
    "type": "PRODUCT_TYPE",
    "valueRequired": true,
    "visibleInStorefront": true,
    "values": [
      {
        "name": "1.5"
      },
      {
        "name": "15"
      }
    ]
  }
}
```

Acceptance criteria: 
- Mutation with given input works properly and creates Attribute with given values.
- Adding values with similar names (the ones that would have the same result when they are slugified) would result in generating a unique slug for all the values. 
- Names that are different only by upper/lower case or spaces are not allowed.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/attribute/mutations/mixins.py
+++ b/saleor/graphql/attribute/mutations/mixins.py
@@ -121,7 +121,8 @@ class AttributeMixin:
     @classmethod
     def check_values_are_unique(cls, values_input: dict, attribute: models.Attribute):
         # Check values uniqueness in case of creating new attribute.
-        existing_names = attribute.values.values_list("slug", flat=True)
+        existing_names = attribute.values.values_list("name", flat=True)
+        existing_names = [name.lower().strip() for name in existing_names]
         for value_data in values_input:
             name = unidecode(value_data["name"]).lower().strip()
             if name in existing_names:
--- a/saleor/graphql/attribute/mutations/mixins.py
+++ b/saleor/graphql/attribute/mutations/mixins.py
@@ -43,10 +43,11 @@ class AttributeMixin:
             )
 
         is_swatch_attr = attribute_input_type == AttributeInputType.SWATCH
-        values_list = list(attribute.values.values_list("slug", flat=True))
+
+        slug_list = list(attribute.values.values_list("slug", flat=True))
 
         for value_data in values_input:
-            cls._validate_value(attribute, value_data, is_swatch_attr, values_list)
+            cls._validate_value(attribute, value_data, is_swatch_attr, slug_list)
 
         cls.check_values_are_unique(values_input, attribute)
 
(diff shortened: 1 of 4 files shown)


maarcingebala (commented):
Can we also add a test that ensures that the case for which the bug was reported is working correctly?

fowczarek (approved):

saleor/graphql/attribute/tests/mutations/test_attribute_update.py
IKarbowiak, line 1: Could you also add a test for creating attribute values with names that give the same slug with slugify?

jakubkuc: added 👍🏻

saleor/graphql/attribute/tests/mutations/test_attribute_create.py
```diff
@@ -1255,3 +1255,43 @@ def test_create_attribute_and_attribute_values_errors(
     assert errors[0]["field"] == "values"
     assert errors[0]["message"] == error_msg
     assert errors[0]["code"] == error_code.name
+
+
+def test_create_attribute_similar_names(
```
IKarbowiak, line 1260: Could we also add a similar one for `update_attribute` mutation, when the value with `15` already exists and we are adding a new `1.5` value?

szczecha (approved):

maarcingebala (approved):

saleor/graphql/attribute/mutations/mixins.py
```diff
@@ -115,10 +121,10 @@ def validate_swatch_attr_value(cls, value_data: dict):
     @classmethod
     def check_values_are_unique(cls, values_input: dict, attribute: models.Attribute):
         # Check values uniqueness in case of creating new attribute.
-        existing_values = attribute.values.values_list("slug", flat=True)
+        existing_names = attribute.values.values_list("slug", flat=True)
```
IKarbowiak, line 124:
```suggestion
        existing_names = attribute.values.values_list("name", flat=True)
```

IKarbowiak (approved):

slawekmarszalek:
Reported in: https://saleorcommerce.slack.com/archives/C038XPD8HLK/p1678892312013259

PR #12373, opened 2023-03-20
[Support for payment apps] zero downtime migrations

korycins:
I want to merge this change because it provides logic in the previous saleor version that will support zero downtime migrations. Covers: https://github.com/saleor/saleor/issues/12305

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/payment/migrations/0042_auto_20230320_1252.py
```diff
@@ -0,0 +1,30 @@
...
+        ),
+        migrations.AddField(
+            model_name="transactionitem",
+            name="use_old_id",
+            field=models.BooleanField(default=True),
+        ),
+        migrations.RunSQL(
+            """
+            ALTER TABLE payment_transactionevent
+            ALTER COLUMN include_in_calculations
```
korycins, line 25: Change incorrect field

saleor/payment/migrations/tasks/saleor3_12.py
```diff
@@ -310,3 +311,27 @@ def set_default_currency_for_transaction_event_task():
...
+
+
+def update_transaction_token_field(transaction_item_class, batch_size=None):
+    transactions = transaction_item_class.objects.filter(token__isnull=True)
+    fully_processed = True
+    if batch_size:
+        transactions = transactions[:BATCH_SIZE]
+
+    for transaction in transactions:
+        transaction.token = RandomUUID()
```
IKarbowiak, line 323: You can do this in a single update query, take a look [here](https://github.com/saleor/saleor/blob/main/saleor/checkout/migrations/0044_fulfill_checkout_line_new_fields.py#L15)

saleor/payment/migrations/0043_populate_transaction_item_token.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def on_migrations_complete(sender=None, **kwargs):
+        update_transaction_token_field_task.delay()
+
+    if __version__.startswith("3.12"):
+        sender = registry.get_app_config("account")
+        post_migrate.connect(on_migrations_complete, weak=False, sender=sender)
+    else:
+        update_transaction_token_field(transaction_item_class=TransactionItem)
+
```
IKarbowiak, line 25: Shouldn't it be done in the next version? If we update it here, we'll need to run it in the next version again, as we will have instances with empty `token` value.

IKarbowiak: Or maybe we should change the `token` definition to set the default value to fulfill new instances with `UUID` token value.

korycins: @IKarbowiak  3.13 version  changes `token` from null=True to `False` and sets the `default`.
In 3.12, all places where the transaction is created have an additional implementation to set the token.
Assuming that the flow of upgrading from 3.11 to 3.12 for 0 downtime looks like below:
- turn off celery worker
- call migrations for 3.12
- turn on workers for 3.12
- turn off workers for 3.11
- turn on celery for 3.12

Celery will set the token for all missing fields, and all new transaction created from 3.12 will have token. 
But to be 100% sure, I also added additional migration in 3.13, to double check that we didn't have any empty token. Maybe it is not needed, but I would like to be sure all transactions are ready for change from `null=True` to `False`

saleor/payment/migrations/tasks/saleor3_12.py
korycins: Right, thanks

korycins: @IKarbowiak Hhm, I am not sure if this the best way for handling this. By calling `.update` and `Case` I will lock the whole table with `transactionItem`. And I don't think that I can use `update` with slice `[:BATCH_SIZE]`. I believe that the celery worker should be fast, take part of the records, update them and move to the next batch.
Am I wrong?

IKarbowiak (approved):

PR #12432, opened 2023-03-28
Support channel permissions in OIDC

IKarbowiak:
Add support for OIDC.

All new staff users will be assigned to the default staff group. 

# 3. OIDC changes
## 3.1 Allows configuring staff user's default permission group via OIDC settings
We want to allow the OIDC plugin to assign newly created staff users to the chosen permission group. To be able to define if the user should be marked as a staff user we want to introduce a domain whitelist. 
Add new fields OIDC settings:
- `Staff users domains`: In this wield we store a list of domains separated by `,`(Store them as the lower case). If the email newly created user will be matched to any of these domains then the user will be marked as a staff user instead of a customer. 
- `Default group name for new staff users`: When any user will be created as staff then will be assigned to the permission group matching to group from this field. If the provided group doesn't exist Saleor we create a such group without any assigned permissions and channels.
> :warning: Warning :warning: 
> If `Default group name for new staff users` will be empty and user domain will be matched to any `Staff users domains`. Then Saleor creates a new staff user without any permission group assigned. 

## 3.2 Flow of new login process via OIDC:
```mermaid
graph TD
 A[User login via OIDC]
 A --> B{Check if the user has Saleor account}
 B -->|Yes| C[The user is successfully logged in]
 B -->|No| D{Check if the email domain match to<br> domain from OIDC settings}
 D -->|Yes| E[Create staff user in Saleor]
 D -->|No| C
 E --> F{Default permission group set in OIDC settings}
 F -->|Yes| G[Assing newly created staff user to default group<br> If the group does not exist create them without<br> any assigned permissions and channels]
 G -->C
 F -->|No| C
```

### 3.3.1 Migration process 
To do not break the currently working environments with RBAC, in the migration process we should create a permission group without any permission and all channels assigned to this group. Assign all staff users to this group. This migration should execute code only if OIDC is turned on. 

### 3.3.2 Logged-in user permission should be the union of permissions from OIDC and from the Saleor groups.
In pre Channel permissions we treat OIDC as a source of truth about user permission. In a new way, we want to create a union of permissions from OIDC and Saleor groups. 
After this change, OIDC will be provided only permissions and Saleor groups provide permissions and channel restrictions. 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


patrys:
Task linked: [SALEOR-7880 Channel permissions [Orders]](https://app.clickup.com/t/2549495/SALEOR-7880)

saleor/account/migrations/tasks/saleor3_13.py
```diff
@@ -0,0 +1,23 @@
...
+
+from ....celeryconf import app
+from ...models import User, Group
+
+# The batch of size 5000 took about 0.5 s to assign users to groups
+BATCH_SIZE = 5000
+
+
+@app.task
+def create_full_channel_access_group_task(group_name):
```
korycins, line 11: I would double-check if this task should go to 3.13 or maybe to 3.12. If you will merge this task to 3.13, then we will have a moment, when the data will not be migrated correctly, which means, that you could block the access to dashboard.

saleor/plugins/openid_connect/utils.py
```diff
@@ -391,6 +455,16 @@ def _update_user_details(
         user.save(update_fields=fields_to_save)
 
 
+def get_staff_user_domains(
+    config: "OpenIDConnectConfig",
+):
+    """Return staff user domains for given gateway configuration."""
+    staff_domains = config.staff_user_domains
+    return (
+        [domain.strip() for domain in staff_domains.split(",")] if staff_domains else []
```
korycins, line 464: Just to confirm. Shouldn't we also call `.lower()` here?

saleor/plugins/openid_connect/utils.py
```diff
@@ -245,12 +270,39 @@ def get_user_from_oauth_access_token(
...
+
     return user
 
 
+def assign_staff_to_default_group_and_update_permissions(
+    user: "User", default_group_name: str
+):
+    """Assign staff user to the default permission group. and update user permissions.
+
+    If the group doesn't exist, the new group without any assigned permissions or
```
korycins, line 295: Are you sure that `get_or_create` is the correct approach here? Consider the following cases:

- What if the provided default group is an empty string?
- What if the provided `default_group_name` has spaces in the name, like `" my-group"`?
- What if someone makes a typo when trying to assign a default group? For example, assuming that you have a group named `OpenID group`, but a staff member provides `openid group` or `OpenID-group`. This would result in the creation of a new group. Is this expected behavior? Shouldn't we raise an exception that group doesn’t exist during the validation of the data to save on the OpenID plugin side?

korycins: If we would rise an exception in that case, then we should also consider a case when the group has been removed.

saleor/account/migrations/tasks/saleor3_13.py
IKarbowiak: Thanks! Yes, it probably should go to 3.12. I will handle it after merging to feature branch

saleor/plugins/openid_connect/utils.py
IKarbowiak: So for sure I should do the `strip` before, and check if the group name is not empty. Regarding checking if the group exists I can't entirely agree. In the feature, there is an assumption that when the group doesn't exist we should create a new one. It will be mentioned in the docs.

IKarbowiak: I updated the method - added a strip on value and I don't create the group when the value is empty

saleor/account/utils.py
```diff
@@ -129,3 +130,13 @@ def retrieve_user_by_email(email):
...
+
+
+def get_user_groups_permissions(user: User):
+    GroupUser = User.groups.through
+    group_users = GroupUser.objects.filter(user_id=user.id)
+    GroupPermissions = Group.permissions.through
+    group_permissions = GroupPermissions.objects.filter(
+        group_id__in=group_users.values("group_id")
+    )
+    return Permission.objects.filter(id__in=group_permissions.values("permission_id"))
```
fowczarek, line 142: Should we rewrite those queries to use `Exists`?

fowczarek (approved):

fowczarek (approved):

saleor/plugins/openid_connect/utils.py
```diff
@@ -245,12 +270,43 @@ def get_user_from_oauth_access_token(
...
+    channels.
+    """
+    default_group_name = (
+        default_group_name.strip() if default_group_name else default_group_name
```
kadewu, line 299: Just want to check for sure, that this `if` here is purposely. Cause right now we will accept names like "   ", but not "".

kadewu (approved):

saleor/plugins/openid_connect/dataclasses.py
```diff
@@ -13,3 +13,5 @@ class OpenIDConnectConfig:
     user_info_url: str
     audience: str
     use_scope_permissions: bool
+    staff_user_domains: str
+    default_group_name: str
```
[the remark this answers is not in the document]
korycins: @IKarbowiak  I meant a situation like this:
staffUserA: sets `default_group_name` as a blank string
staffUserB: logs in with OIDC
<what happen with `staffUserB` in that case?>

PR #12441, opened 2023-03-30
[RFC] Add app events log

przlada:
We want to add app events history log mechanism. This is the proposed schema and initial implementation. 
Comments are welcome as this approach differs from other event logs in use, like `OrderEvent` or `CustomerEvent`.
Main differences:
- Every event type has its own graphQL object type that implements `AppEvent` interface instead of one object type for all event types. 
  ```graphql
  union AppEventRequestor = User | App
  
  interface AppEvent {
    id: ID!
    createdAt: DateTime!
    requestor: AppEventRequestor
  }
  ```
  For now, we only support three app event types: `AppEventInstalled`, `AppEventActivated` and `AppEventDeactivated`, but we plan to add more.  
- Custom event type data won't be stored in one `JSONField` field. Instead, custom (nullable) fields will be added to the `AppEvent` model when needed.
- App events can be accessed using the `events` field of the `App` object type. `FilterConnectionField` is used instead of `NonNullList` as there could possibly be an unlimited amount of app events, and we would like to be able to filter them. 

After we add the automatic deactivation of failing apps functionality to Saleor, we could extend `AppEventDeactivated` event with custom event data. It could look, for example, like this: 
- New nullable field in the `AppEvent` model:
  ```python
  class AppEvent(models.Model):
      app = models.ForeignKey(App, related_name="events", on_delete=models.CASCADE)
      ...
      deactivation_reason = models.TextField(blank=True, null=True)
  ```
- New field in `AppEventDeactivated` object type
  ```graphql
  type AppEventDeactivated implements Node & AppEvent {
    id: ID!
    createdAt: DateTime!
    requestor: AppEventRequestor
    reason: String
  }
  ```

### Todo
* [ ] Validate implementation idea
* [ ] Add schema descriptions
* [ ] Invoke app event creation
* [ ] Add unit tests

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


korycins:
`union AppEventRequestor = User | App`
There is already common union for `User|App`. I think that you should use it, instead of adding a new one. `UserOrApp`.

saleor/app/events.py
```diff
@@ -0,0 +1,34 @@
...
+def app_event_activated(app: App, *, requestor: Optional[Union[App, User]]):
+    return AppEvent.objects.create(
+        app=app, type=AppEventType.ACTIVATED, **_requestor(requestor)
+    )
+
+
+def app_event_deactivated(app: App, *, requestor: Optional[Union[App, User]]):
+    return AppEvent.objects.create(
+        app=app, type=AppEventType.DEACTIVATED, **_requestor(requestor)
+    )
```
korycins, line 34: Don't know if this is something for MVP but: Maybe would be worth adding the events for webhook to `AppEvent`? Like `webhook_created`, `updated`, `deleted`? This would be usefull as it could show what happens with the webhooks assigned to the specific app. WDYT?

saleor/graphql/api.py
```diff
@@ -116,7 +117,7 @@ class Mutation(
 schema = build_federated_schema(
     Query,
     mutation=Mutation,
-    types=unit_enums + list(WEBHOOK_TYPES_MAP.values()),
+    types=unit_enums + list(WEBHOOK_TYPES_MAP.values()) + list(APP_EVENTS_MAP.values()),
```
korycins, line 120: Are you sure that you need this? 
AFAIK it should not be necessary. The app events should be imported through the `App` type.

saleor/graphql/app/events.py
```diff
@@ -0,0 +1,20 @@
+import graphene
+
+from ...account.models import User as UserModel
+from ...app.models import App as AppModel
+from ..account.types import User
+from ..app.types import App
+from ..core import ResolveInfo
+
+
+class AppEventRequestor(graphene.Union):
```
korycins, line 10: `saleor.graphql.core.types.user_or_app.UserOrApp`

saleor/graphql/app/sorters.py
```diff
@@ -22,3 +22,20 @@ class Meta:
         doc_category = DOC_CATEGORY_APPS
         sort_enum = AppSortField
         type_name = "apps"
+
+
+class AppEventSortField(BaseEnum):
+    CREATION_DATE = ["created_at", "pk"]
```
korycins, line 28: I think that it should be `CREATED_AT`. At least for the `OrderSort` we deprecate the `CREATION_DATE` in favor of `CREATED_AT`

saleor/app/events.py
lkostrowski: @korycins we will think about it in post-mvp. We don't want to accidentally end-up having monitoring/observability implementation here. Initially I was thinking about something user-facing. 

For app adding, upading and deleting events will be "daily job". For example, app should not register itself with webhook, because it will not work in many cases (config needed). So each time user changes configuration,webhook likely changes (disable/enable, create/delete).

lkostrowski:
From API perspective ✅ from me

PR #12442, opened 2023-03-30
Validate positive value of weight fields directly in `WeightScalar` object.

zedzior:
I want to merge this change because I want to move validation of positive weight value, directly to `WeightScalar` object, instead of every single mutation, containing weight field.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/product/tests/mutations/test_product_create.py
+++ b/saleor/graphql/product/tests/mutations/test_product_create.py
@@ -19,6 +19,7 @@ from .....plugins.manager import PluginsManager
 from .....product.error_codes import ProductErrorCode
 from .....product.models import Product
 from .....tests.utils import dummy_editorjs
+from ....tests.utils import assert_graphql_error_with_message
 
 CREATE_PRODUCT_MUTATION = """
        mutation createProduct(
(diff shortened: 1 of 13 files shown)


kadewu (approved):

jakubkuc (approved):

IKarbowiak (approved):

saleor/graphql/product/tests/mutations/test_product_create.py
```diff
@@ -883,13 +884,12 @@ def test_create_product_with_negative_weight(
...
+        query,
+        variables,
+        permissions=[permission_manage_products],
+        check_no_permissions=False,
     )
-    content = get_graphql_content(response)
-    data = content["data"]["productCreate"]
-    error = data["errors"][0]
-    assert error["field"] == "weight"
-    assert error["code"] == ProductErrorCode.INVALID.name
```
korycins, line 892: What about checking the error code?
Same question for `field`. 
IMHO these tests should return the same error codes as previously. Checking the error msg is an additional validation but we should be sure that we didn't change the behavior of the API and we return the same errors as previously.

szczecha (approved):

zedzior:
This change is still under the discussion. Proposed generic errors are unacceptable in this form.

PR #12402, opened 2023-03-25
Use mailhog smtp server on Dev Container

carlosa54:
I saw that dev container support was recently added in [Dev Container support for local development](https://github.com/saleor/saleor/pull/12391) which is nice 🚀 

I want to merge this change because it adds mailhog local smtp server on the dev container

### Test
1.  Open the dev container in VSCode
2. After the container is loaded go to http://localhost:8025

mailhog-screenshot

### M1 Mac `arm64` users
The `mailhog` docker image is not built for `arm64`. Because of this you could get a warning/error. You'll need to enable the `Use Rosetta for x86/amd64 emulation on Apple Silicon` in the Docker Dashboard settings.

<details>
<summary>See Screenshots</summary>
Make sure `Use Virtualization framework` is enabled in `Settings -> General`

image
</details>

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #12447, opened 2023-04-01
Use mailpit in Dev Container

carlosa54:
I want to merge this change because it replaces mailhog with mailpit.

I recently pushed https://github.com/saleor/saleor/pull/12402 for Mailhog support in the Dev Container setup with a workaround for `arm64` users since the mailhog docker image don't have support for it. I was going to add a new issue in the Mailhog repo but realized that the project seem unmaintained.

Found a well maintained alternative [Mailpit](https://github.com/axllent/mailpit) that was inspired by `Mailhog` and started using it and it works pretty nice. It also has a docker image that supports multiple archs so no need to emulate

### Screenshots
mailpit

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/.devcontainer/backend.env
+++ b/.devcontainer/backend.env
@@ -3,5 +3,5 @@ DASHBOARD_URL=http://localhost:9000/
 DEFAULT_FROM_EMAIL=noreply@example.com
 CELERY_BROKER_URL=redis://redis:6379/1
 SECRET_KEY=changeme
-# Mailhog
-EMAIL_URL=smtp://mailhog:1025
+# Mailpit
+EMAIL_URL=smtp://mailpit:1025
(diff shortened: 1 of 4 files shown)


maarcingebala:
@carlosa54 Thanks for opening this. Indeed, it seems like Mailhog is not maintained, so we would be keen on replacing it with Mailpit. Would you also be able to look into updating this in [saleor-platform repo](https://github.com/saleor/saleor-platform/)? It's another way of setting up Saleor locally which uses Mailhog, so if we want to replace, we need to change it in both places.

carlosa54:
@maarcingebala no problem 👍 PR added in https://github.com/saleor/saleor-platform/pull/261

maarcingebala (approved):

IKarbowiak (approved):

szczecha (approved):

carlosa54:
> @carlosa54 Now I noticed we're missing changelog here, can you add an entry about this change?

@maarcingebala updated now

PR #12503, opened 2023-04-06
Fix deadlock in order events celery migrations task

kadewu:
I want to merge this change because it fixes the issue when the migration triggered the data migration as celery tasks and there is more than one celery worker that can process the tasks. Previously there was a possibility of having a deadlock during the update, currently, we lock the batch of products.

Additionaly increase the batch size and improve query performance to decrease overall migration time.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/order/migrations/tasks/saleor3_13.py
+++ b/saleor/order/migrations/tasks/saleor3_13.py
@@ -1,14 +1,18 @@
 from ....celeryconf import app
 from ...models import OrderEvent
 
+from django.db import transaction
 from django.db.models import QuerySet
 
-# Batch size of size 1000 is about 1MB memory usage in task
-BATCH_SIZE = 1000
+# Batch size of size 5000 is about 1MB memory usage in task
+BATCH_SIZE = 5000
 
 
 def update_type_to_transaction_cancel_requested(qs: QuerySet[OrderEvent]):
-    qs.update(type="transaction_cancel_requested")
+    with transaction.atomic():
+        # lock the batch of objects
+        _events = list(qs.select_for_update(of=(["self"])))
+        qs.update(type="transaction_cancel_requested")
 
 
 @app.task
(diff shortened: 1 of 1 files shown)


fowczarek (approved):

saleor/order/migrations/tasks/saleor3_13.py
```diff
@@ -1,36 +1,47 @@
...
     if ids:
         update_type_to_transaction_cancel_requested(qs)
         order_events_rename_transaction_void_events_task.delay()
 
 
 def update_type_to_transaction_charge_requested(qs: QuerySet[OrderEvent]):
-    qs.update(type="transaction_charge_requested")
+    with transaction.atomic():
+        # lock the batch of objects
+        _events = list(qs.select_for_update(of=(["self"])))
```
tomaszszymanski129, line 34: cast to list and thus evaluation of extra qs can be avoided and moved to subquery which will be much faster, see here on how it's done:
https://github.com/saleor/saleor/blob/main/saleor/graphql/attribute/mutations/attribute_value_update.py#L83
It also does the batching and in your case I believe you can also skip`with transaction.atomic():` if you use that approach. Update operations are atomic and since we're gonna do things in subquery that will be the only statement.

tomaszszymanski129 (changes requested):

saleor/order/migrations/tasks/saleor3_13.py
IKarbowiak: In the example you sent, we block all related products until all batches will be finished, I'm not sure if it's right. Let's consider the migration will take an hour, the access to those products will be blocked for this time AFAIK

korycins: Also keep in mind that this task will called in a celery worker. There is a possibility that we will have a different tasks that will try to update the same objects, which means that we need to lock a batch of objects to be sure that DB will not raise the deadlock exception.
This is mandatory when you have more that one celery worker that processes the data

korycins (approved):

saleor/order/migrations/tasks/saleor3_13.py
```diff
@@ -1,36 +1,47 @@
 from ....celeryconf import app
 from ...models import OrderEvent
 
+from django.db import transaction
 from django.db.models import QuerySet
 
-# Batch size of size 1000 is about 1MB memory usage in task
-BATCH_SIZE = 1000
+# Batch size of size 5000 is about 1MB memory usage in task
```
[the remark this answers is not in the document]
kadewu: You are right, I forgot to update the value here.

tomaszszymanski129 (approved):

PR #12269, opened 2023-03-09
Order bulk create

zedzior:
RFC: https://github.com/saleor/saleor/issues/11870
I want to merge this change because, I want to allow user to create multiple orders in single mutation.

Following missing items will be handled in separate PRs:
- ~~fulfillments~~
- ~~transactions~~
- ~~invoices~~
- ~~gift cards / vouchers / discounts~~
- ~~ORDER_BULK_CREATED webhook~~
- ~~order number~~
- ~~fix weight validators~~

# Impact

* [ ] New migrations
* [x] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

Issue #11870, opened 2023-01-30
[RFC] Bulk order create / order import


maarcingebala:
# Order Import RFC

This document describes a mutation allowing users to create multiple orders in Saleor. Purpose of this mutation could be importing orders from other system. The main assumption is to allow the user to specify as many fields as possible and skip most of the logic and calculations that are done when creating orders from draft or checkout. Among the other things, the mutation allows specifying things like: order creation date, order number, order lines with net and gross line prices, payment transactions, and fulfillments. 

### Permissions

- The mutation requires a new permission: `MANAGE_ORDERS_IMPORT`. Since this operations is more powerful than regular order management, it is not automatically assigned to staff users or apps with `MANAGE_ORDERS` permission. Instead it is assigned only to “full access” users, if such a group exists in the system, or to `MANGE_SETTINGS` users.

### GraphQL Schema

```graphql
type Mutation {
  orderBulkCreate(
    errorPolicy: ErrorPolicyEnum,
    stockUpdatePolicy: StockUpdatePolicyEnum,
    orders: [OrderBulkCreateInput!]!
  ): OrderBulkCreate!
}

type OrderBulkCreate {
  count: Int!
  results: [OrderBulkCreateResult!]!
}

type OrderBulkCreateResult {
  order: Order!
  errors: [OrderBulkCreateError!]
}

type OrderBulkCreateError {
  path: String
  message: String
  code: OrderBulkCreateErrorCode!
}

enum OrderBulkCreateErrorCode {
    GRAPHQL_ERROR
    REQUIRED
    INVALID
    NOT_FOUND
    UNIQUE
    BULK_LIMIT
    TOO_MANY_IDENTIFIERS
    FUTURE_DATE
    INVALID_QUANTITY
    PRICE_ERROR
    NOTE_LENGTH
    INSUFFICIENT_STOCK
    NON_EXISTING_STOCK
    NO_RELATED_ORDER_LINE
    NEGATIVE_INDEX
    ORDER_LINE_FULFILLMENT_LINE_MISMATCH
    METADATA_KEY_REQUIRED
    INCORRECT_CURRENCY
}

enum OrderOriginEnum {
  BULK_CREATE
}

enum StockUpdatePolicyEnum {
  SKIP
  UPDATE
  FORCE
}

"""
If `id` is provided, try to get the user by ID. If `email`
is provided try to lookup the user by email; if not found,
set only the email in the Order instance.
"""
input OrderBulkCreateUserInput {
  id: ID
[...]

PR #12536, opened 2023-04-13
Complete bulk order create

zedzior:
RFC: https://github.com/saleor/saleor/issues/11870
I want to merge this change, because this PR add missing logic to https://github.com/saleor/saleor/pull/12269.

The PR includes:
- process discounts
- process vouchers
- process gift cards
- process order.number
- reduce API calls by collecting all models keys from input and retrieving all instances at once
- add missing validations
- add OrderBulkCreated event
- simplify code
- replace `OrderBulkCreateError.field` with `path`

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/order/bulk_mutations/order_bulk_create.py
```diff
@@ -518,6 +543,132 @@ class Meta:
...
+                        line.get("variant_external_reference")
+                    )
+                    keys["Warehouse.id"].append(line.get("warehouse"))
+            for gift_card_code in order["gift_cards"]:
+                keys["GiftCard.code"].append(gift_card_code)
+
+        # Convert global ids to model ids and get rid of Nones
+        for key, values in keys.items():
+            keys[key] = [value for value in values if value is not None]
+            if key.split(".")[1] == "id":
```
korycins, line 602: If you always have a structure of `key` like `<type>.<identifier>`, You could do something like this 👇  Then you don't need to call `key.split(".")[0]` for each global_id in values.
```
object_type, key_field = key.split(".")
if key_field == "id":
    model_ids = []
    for global_id in values:
        try:
            _, id = from_global_id_or_error(
                str(global_id), object_type, raise_error=True
            )
            model_ids.append(id)
        except GraphQLError:
            pass
    keys[key] = model_ids
```

korycins (commented):
Is there a limit to how many orders can be imported at once? Importing a large number of orders or objects attached to the order can consume a lot of memory and cause the Saleor process to shut down.

If we don't want to impose any limits, we can implement a logic that processes the data by batch instead of loading all data based on the input. This will prevent loading a lot of objects into memory at once.
cc @zedzior @maarcingebala

zedzior:
> Is there a limit to how many orders can be imported at once?

@korycins There is a PR in progress, to determine limit of objects to process, for all bulk mutations.

saleor/graphql/order/bulk_mutations/order_bulk_create.py
```diff
@@ -543,27 +703,113 @@ def validate_order_input(cls, order_input, errors: List[OrderBulkError]):
...
             )
 
-        # TODO validate status (wait for fulfillments)
-        # TODO validate number (?)
-        return errors
+    @classmethod
+    def validate_order_numbers(cls, orders: List[OrderBulkClass]):
+        numbers = []
+        for order in orders:
+            if order.order and order.order.number in numbers:
```
maarcingebala, line 772: `order.order` looks strange, maybe to differentiate `Order` instance and the `OrderBulkClass` instance it should be `order_data.order`?

saleor/graphql/order/bulk_mutations/order_bulk_create.py
```diff
@@ -79,36 +96,35 @@ class OrderBulkOrderLine:
 
 
 @dataclass
-class OrderWithErrors:
+class OrderBulkClass:
```
maarcingebala, line 99: Maybe rename it to `OrderBulkCreateData`? Should be clear name suggesting that this is only a data class gathering different fields.

saleor/graphql/order/bulk_mutations/order_bulk_create.py
```diff
@@ -517,6 +546,135 @@ class Meta:
...
+                    try:
+                        _, id = from_global_id_or_error(
+                            str(global_id), object_type, raise_error=True
+                        )
+                        model_ids.append(id)
+                    except GraphQLError:
+                        pass
+                keys[key] = model_ids
+
+        # Make API calls
```
maarcingebala, line 621: "Make DB calls"?

saleor/graphql/order/bulk_mutations/order_bulk_create.py
zedzior: good points, thanks

saleor/graphql/order/bulk_mutations/order_bulk_create.py
```diff
@@ -930,55 +1232,85 @@ def create_single_note(
...
+                    code=OrderBulkCreateErrorCode.INVALID,
+                )
+            )
+
+        return OrderDiscount(
+            order=order_data.order,
+            value_type=discount_input["value_type"],
+            value=discount_input["value"],
+            reason=discount_input.get("reason"),
+        )
```
fowczarek, line 1299: Shouldn't we provide also `amount`?

zedzior: According to RFC discount input is type of `OrderDiscountCommonInput` and there is no info about amount. In general, for `OrderBulkCreate` mutation, `lineTotalPrice` and `undiscountedLineTotalPrice` are the primary sources of truth about the order pricing.

fowczarek (approved):

saleor/graphql/order/bulk_mutations/order_bulk_create.py
```diff
@@ -1233,178 +1576,248 @@ def create_single_fulfillment(
...
         # get order related instances
-        instances = cls.get_instances_related_to_order(
+        cls.get_instances_related_to_order(
             order_input=order_input,
-            errors=order.order_errors,
+            order_data=order_data,
             object_storage=object_storage,
         )
         delivery_input = order_input["delivery_method"]
         delivery_method = cls.get_delivery_method(
```
korycins, line 1600: Just to confirm: Are you taking into account that some of the orders would not require a shipping? Right?

saleor/graphql/order/bulk_mutations/order_bulk_create.py
```diff
@@ -1233,178 +1576,248 @@ def create_single_fulfillment(
...
+            if order_data.is_critical_error:
+                order_data.order = None
+                return order_data
 
         # create transactions
         if transactions_input := order_input.get("transactions"):
+            transaction_index = 0
             for transaction_input in transactions_input:
                 try:
                     transaction = TransactionCreate.prepare_transaction_item_for_order(
```
korycins, line 1677: I am not convinced of having this method inside `TransactionCreate`. As far as I see, it is used only in this file.

korycins: Also in case of adding any transaction that is attached to the order. You need to proceed of few steps that are mandatory. You can check `TransactionCreate.perform_mutation`.
```
new_transaction = cls.create_transaction(transaction_data, user=user, app=app)
if money_data:
    create_manual_adjustment_events(
        transaction=new_transaction, money_data=money_data, user=user, app=app
    )
    recalculate_transaction_amounts(new_transaction)
if transaction_data.get("order_id") and money_data:
    order = cast(order_models.Order, new_transaction.order)
    cls.update_order(order, money_data, update_search_vector=True)

    order_info = fetch_order_info(order)
    order_transaction_updated(
        order_info=order_info,
        transaction_item=new_transaction,
        manager=manager,
        user=user,
        app=app,
        previous_authorized_value=Decimal(0),
        previous_charged_value=Decimal(0),
    )
```

korycins: `create_manual_adjustment_events` creates the events that will be used to re-calcualte the transaction's amounts. The amounts provided in `transaction` are based on the events' amounts. So without adding events, any update on imported transaction  will set amounts to 0.

korycins: `order_transaction_updated` probably should be skipped as it triggers the webhooks like order_fully_piad

saleor/graphql/order/bulk_mutations/order_bulk_create.py
```diff
@@ -1233,178 +1576,248 @@ def create_single_fulfillment(
                 quantity=line_input["quantity"],
             )
             lines.append(OrderBulkFulfillmentLine(fulfillment_line, warehouse))
+            line_index += 1
 
         return OrderBulkFulfillment(fulfillment=fulfillment, lines=lines)
 
     @classmethod
     def create_single_order(
```
korycins, line 1584: The function is quite long (1584-1800). Have you considered adding some helper methods to make it shorter?

saleor/graphql/order/bulk_mutations/order_bulk_create.py
zedzior: Good spot, I omitted this case. Implemented in last push.

saleor/graphql/order/bulk_mutations/order_bulk_create.py
zedzior: Reduced it a bit (216 -> 130). Let me know if you have other ideas how to organise the whole module.

saleor/graphql/webhook/subscription_types.py
```diff
@@ -438,6 +442,26 @@ class Meta:
         description = "Event sent when order metadata is updated." + ADDED_IN_38
 
 
+class OrderBulkCreated(SubscriptionObjectType):
+    orders = ConnectionField(
```
korycins, line 446: This adds `first` `last` etc to the `order` query.  This is something that we will not be able to use from the webhook subscription.  I believe that we should have a normal list of orders. 
Again the worst thing is that we could consume all system resources, but this should be solved by the limit for the mutation that we are going to introduce.

saleor/plugins/base_plugin.py
```diff
@@ -711,6 +711,12 @@ def __str__(self):
     # metadata is changed.
     order_metadata_updated: Callable[["Order", Any], Any]
 
+    # Trigger when order is imported.
```
korycins, line 714: Does this description is correct? It says `order is imported` it should be `orders` right?

korycins (approved):

PR #12561, opened 2023-04-17
Don't require query for sync webhooks in app manifest

### LINKED ISSUE #12559: Bug: Apps can't get full payloads for sync webhooks

Trying to create an app which registers sync payment/shipping webhooks automatically via the manifest.

1. Create an app project
2. Add a webhook like the following to the manifest
```
{
  name: "Payment: List Gateways",
  syncEvents: ["PAYMENT_LIST_GATEWAYS"],
  targetUrl: "...",
  isActive: true
}
```
3. Attempt to add the app to a Saleor instance
4. This results in an error: "Missing required fields for webhook: query"

Specifying query as null or an empty string also fails.

The only way to get the app to register is to set query to valid GraphQL, e.g. `"subscription { event { issuedAt } }"`

This causes problems because (as far as I can tell) not all information in the default sync webhook payloads is available in the GraphQL subscription payload.

I expected that query would only be required for `asyncEvents`, not `syncEvents`.

(When adding a webhook in the UI, there's no option to add a `query` for sync webhooks, so I assume the non-subscription version is what's meant to be used)

### Logs

_No response_

### Environment

Saleor version: main (`7465cc83b`)
OS and version: Docker 4.10.1 on macOS

binary-koan:
I want to merge this change because it fixes #12559

This is what I've been using locally to work around the problem and register sync webhooks with an app manifest. It works with the app I'm working on, but I haven't exhaustively tested different cases.

As always, not sure if this is the best solution but hopefully it can start a conversation or help things along at least!

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/app/installation_utils.py
+++ b/saleor/app/installation_utils.py
@@ -92,7 +92,7 @@ def install_app(app_installation: AppInstallation, activate: bool = False):
             app=app,
             name=webhook["name"],
             target_url=webhook["targetUrl"],
-            subscription_query=webhook["query"],
+            subscription_query=webhook.get("query", None),
             custom_headers=webhook.get("customHeaders", None),
         )
         for webhook in manifest_data.get("webhooks", [])
--- a/saleor/app/manifest_validations.py
+++ b/saleor/app/manifest_validations.py
@@ -246,17 +246,19 @@ def clean_webhooks(manifest_data, errors):
                     )
                 )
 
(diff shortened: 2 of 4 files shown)


binary-koan:
Probably not the right fix judging by the issue comments, back to draft until that's confirmed

binary-koan:
Not the right fix, see the explanation in the issue thread for why `query` should be required

Issue #11352, opened 2022-11-30
Possibility to create a checkout based on existing order


korycins:
## Problem

There is no easy way to create a checkout from an order. Having this possibility would allow you to provide users with the option to reorder products based on an existing order.

---

## 1. General Assumptions

We would like to have an API to create checkout based on  existing order. This should cover a case when a customer often buys the same items. The new checkout should contain the all order's lines. For now, we can skip porting the user data like: addresses.

### Acceptance Criteria
- [ ] Customers can create a checkout by calling `checkoutCreateFromOrder` with the order ID.
- [ ] The checkout should contain all available variants that were added to the order.
- [ ] The quantity of a specific line should be the same as the quantity of specific order lines.
- [ ] `checkoutCreateFromOrder` doesn't require any permissions.
- [ ] The checkout should have a `user` assigned if `checkoutCreateFromOrder` is called as a logged-in user.
- [ ] Any unavailable variants will be attached in the mutation response `unavailableVariants` (product_unavailable_for_purchase, unavailable_variant_in_channel, product_not_published, quantity_greater_than_limit ,insufficient_stock) 
- [ ] User that calls the mutation will be attached to the newly created checkout (`checkout.user`, `checkout.email`)

## 2. API changes

``` graphql
type Mutation {
	checkoutCreateFromOrder(
		"""ID of existing order that will be used to create a checkout"""
		id: ID!
	)
}

type CheckoutCreateFromOrder @doc(category: "Checkout") {
  """Variants that were not attached to the checkout."""
  unavailableVariants: [CheckoutCreateFromOrderUnavailableVariant!]

  """Created checkout."""
  checkout: Checkout
  errors: [CheckoutCreateFromOrderError!]!
}

type CheckoutCreateFromOrderUnavailableVariant @doc(category: "Checkout") {
  """The error message."""
  message: String!

  """The error code."""
  code: CheckoutCreateFromOrderUnavailableVariantErrorCode!

  """Variant ID that is unavailable."""
  variantId: ID!

  """Order line ID that is unavailable."""
  lineId: ID!
}

enum CheckoutCreateFromOrderUnavailableVariantErrorCode @doc(category: "Checkout") {
  NOT_FOUND
  PRODUCT_UNAVAILABLE_FOR_PURCHASE
  UNAVAILABLE_VARIANT_IN_CHANNEL
  PRODUCT_NOT_PUBLISHED
  QUANTITY_GREATER_THAN_LIMIT
[...]

PR #12628, opened 2023-04-20
Add mutation to create checkout from order

korycins:
I want to merge this change because it covers the feature explained here:  #11352

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


saleor/graphql/checkout/mutations/checkout_create_from_order.py
```diff
@@ -0,0 +1,425 @@
...
+                CheckoutLineData(
+                    variant_id=str(line.variant_id), quantity=line.quantity
+                )
+                for line in order_lines
+            ],
+        )
+        valid_order_lines: list[order_models.OrderLine] = []
+        if order.shipping_address:
+            country = order.shipping_address.country.code
+        else:
```
fowczarek, line 323: We should also handle the situation when the order is digital and have only `billing_address`. 
Maybe this function could be used here `saleor/order/utils.py:get_order_country`?

fowczarek (approved):
Could you add a benchmark for this new mutation?

IKarbowiak (approved):

saleor/graphql/checkout/mutations/checkout_create_from_order.py
fowczarek: I think we should use one logic in all places. Due to my knowledge, we should choose a country by priorities:
1. country from shipping address if it exists
2. country from billing address if it exists 
3. country default from the channel

korycins: @fowczarek  But proposed one is a different flow than currently exists in the checkout flow. If we agree to include the billing address also, then it should be done in a separate PR. IMHO, shipping address + default country sounds good to me. Especially since I don't think that we should use billing addresses in the logic related to stock/warehouses.

szczecha (approved):

PR #12646, opened 2023-04-21
fix checkout metadata fetches to not create checkout metadata objects

jakubkuc:
I want to merge this change because it prevents fetches of checkout metadata to create checkout metadata objects in db. Otherwise it can cause a thread race during shipping method and payment method fetching during `checkoutCreate`.

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -874,7 +874,7 @@ def clear_delivery_method(checkout_info: "CheckoutInfo"):
             "last_change",
         ]
     )
-    get_or_create_checkout_metadata(checkout).save(
+    get_checkout_metadata(checkout).save(
         update_fields=[
             "private_metadata",
         ]
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -975,4 +975,4 @@ def get_checkout_metadata(checkout: "Checkout"):
     if hasattr(checkout, "metadata_storage"):
         return checkout.metadata_storage
     else:
-        return CheckoutMetadata()
+        return CheckoutMetadata(checkout)
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -975,4 +975,4 @@ def get_checkout_metadata(checkout: "Checkout"):
     if hasattr(checkout, "metadata_storage"):
         return checkout.metadata_storage
     else:
-        return CheckoutMetadata(checkout)
+        return CheckoutMetadata(checkout=checkout)
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -953,7 +953,7 @@ def set_external_shipping_id(checkout: Checkout, app_shipping_id: str):
 
 def get_external_shipping_id(container: Union["Checkout", "Order"]):
     if type(container) == Checkout:
-        container = get_or_create_checkout_metadata(container)
+        container = get_checkout_metadata(container)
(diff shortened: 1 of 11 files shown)


saleor/checkout/utils.py
```diff
@@ -969,3 +969,10 @@ def get_or_create_checkout_metadata(checkout: "Checkout"):
         return checkout.metadata_storage
     else:
         return CheckoutMetadata.objects.create(checkout=checkout)
+
+
+def get_checkout_metadata(checkout: "Checkout"):
```
korycins, line 974: # 

In my opinion, we should double-check all occurrences of `get_or_create_checkout_metadata` and replace them with `get_checkout_metadata`. If the issue has already occurred once, it's only a matter of time before `get_or_create_checkout_metadata` causes the same issue again. Using `get_or_create_checkout_metadata` in one thread can create problems with saving metadata in a second thread. If your checkout doesn't have `metadata_storage`, you create a dummy object and maybe call `.save()` on it. If another thread calls `get_or_create_checkout_metadata` in the meantime, we will get an exception that the object already exists..

jakubkuc: You are right. I've missed some places that should be modified. Fixed it.

IKarbowiak (approved):

korycins (approved):

saleor/checkout/utils.py
```diff
@@ -969,3 +969,10 @@ def get_or_create_checkout_metadata(checkout: "Checkout"):
...
+    if hasattr(checkout, "metadata_storage"):
+        return checkout.metadata_storage
+    else:
+        return CheckoutMetadata()
```
maarcingebala, line 978: This is risky - this instance would not be associated with the given checkout instance, saving that into DB would probably fail.

saleor/checkout/utils.py
maarcingebala, line 974: Can we add unit tests for this function? They should check both scenarios - saving metadata to a checkout where storage already exists and to a checkout without it.

saleor/graphql/checkout/tests/mutations/test_checkout_complete.py
```diff
@@ -484,6 +484,84 @@ def test_checkout_complete_with_metadata_updates_existing_keys(
...
+
+    # when
+    response = user_api_client.post_graphql(MUTATION_CHECKOUT_COMPLETE, variables)
+    content = get_graphql_content(response)
+    data = content["data"]["checkoutComplete"]
+
+    # then
+    assert not data["errors"]
+    assert Order.objects.count() == orders_count + 1
+    order = Order.objects.first()
```
tomaszszymanski129, line 548: @IKarbowiak Can we trust it to always be the correct one? (same for other test)

tomaszszymanski129 (approved):

saleor/graphql/checkout/tests/mutations/test_checkout_complete.py
tomaszszymanski129: Agreed that refactor will be handled in separate ticket as it involves also other tests.

IKarbowiak: Issue created: https://github.com/saleor/saleor/issues/12728

saleor/checkout/utils.py
IKarbowiak: Done

PR #12692, opened 2023-04-26
Fix returning error messages using product names with % sign

SzymJ:
I want to merge this change because it fixes an issue where error messages containing product names with `%` sign were causing `TypeError` 

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/graphql/order/mutations/order_fulfill.py
+++ b/saleor/graphql/order/mutations/order_fulfill.py
@@ -105,7 +105,7 @@ class OrderFulfill(BaseMutation):
                 ) % {
                     "quantity": line_quantity_unfulfilled,
                     "item_pluralize": pluralize(line_quantity_unfulfilled),
-                    "order_line": order_line,
+                    "order_line": str(order_line).replace("%", "%%"),
                 }
                 order_line_global_id = graphene.Node.to_global_id(
                     "OrderLine", order_line.pk
--- a/saleor/graphql/order/mutations/order_fulfill.py
+++ b/saleor/graphql/order/mutations/order_fulfill.py
@@ -100,7 +100,7 @@ class OrderFulfill(BaseMutation):
 
             if line_total_quantity > line_quantity_unfulfilled:
                 msg = (
-                    "Only %(quantity)d item%(item_pluralize)s remaining " "to fulfill."
+                    "Only %(quantity)d item%(item_pluralize)s remaining to fulfill."
                 ) % {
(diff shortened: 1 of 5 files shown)


patrys:
Wouldn't it be better to stop including product names in exception messages? Is that ever more helpful than a GraphQL ID would be?

SzymJ:
> Is this message returned to API users?

yes

maarcingebala:
It's not a breaking change; these messages are more debug info for developers and should not be relied on in frontend clients.

saleor/graphql/order/utils.py
```diff
@@ -300,9 +300,12 @@ def prepare_insufficient_stock_order_validation_errors(exc):
             if item.warehouse_pk
             else None
         )
+
+        item_name = item.order_line or item.variant
+
         errors.append(
             ValidationError(
-                f"Insufficient product stock: {item.order_line or item.variant}",
+                f"Insufficient product stock: {str(item_name).replace('%', '%%')}",
```
maarcingebala, line 308: We could entirely drop rendering the product name or ID from the message since there is a dedicated field in error for that - `order_lines` - where we return IDs of items that throw this error, right?

SzymJ: yes, so I think this will be best idea

maarcingebala (approved):

IKarbowiak (approved):

saleor/graphql/order/mutations/order_fulfill.py
```diff
@@ -100,12 +100,10 @@ def clean_lines(cls, order_lines, quantities_for_lines):
 
             if line_total_quantity > line_quantity_unfulfilled:
                 msg = (
-                    "Only %(quantity)d item%(item_pluralize)s remaining "
-                    "to fulfill: %(order_line)s."
+                    "Only %(quantity)d item%(item_pluralize)s remaining " "to fulfill."
```
fowczarek, line 103:
```suggestion
                    "Only %(quantity)d item%(item_pluralize)s remaining to fulfill."
```

saleor/tests/fixtures.py
```diff
@@ -3936,7 +3936,7 @@ def order_with_lines(
     default_tax_class,
 ):
     product = Product.objects.create(
-        name="Test product",
+        name="Test product 100%",  # "100%" added to test handling error messages with %
```
fowczarek, line 3939: Is this still needed?

PR #12714, opened 2023-04-28
Fix setting webhook activity using `isActive` field in the App manifest

przlada:
I want to merge this change because it uses the `isActive` field from App manifest to set the initial webhook activity state. 
Resolves https://github.com/saleor/apps/issues/435

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/app/installation_utils.py
+++ b/saleor/app/installation_utils.py
@@ -91,6 +91,7 @@ def install_app(app_installation: AppInstallation, activate: bool = False):
         Webhook(
             app=app,
             name=webhook["name"],
+            is_active=webhook["isActive"],
             target_url=webhook["targetUrl"],
             subscription_query=webhook["query"],
             custom_headers=webhook.get("customHeaders", None),
--- a/saleor/app/manifest_validations.py
+++ b/saleor/app/manifest_validations.py
@@ -224,6 +224,15 @@ def clean_webhooks(manifest_data, errors):
     )
 
     for webhook in webhooks:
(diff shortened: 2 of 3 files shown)


patrys:
@przlada Why is it useful to support apps installing inactive webhooks? Why do we need `isActive` in manifest at all?

przlada:
> @przlada Why is it useful to support apps installing inactive webhooks? Why do we need `isActive` in manifest at all?

@patrys In many cases, apps require additional configuration after they are installed. Without that configuration, the app might receive webhook events that it cannot handle correctly. To address this with @lkostrowski we came up with an idea to install those apps with webhooks disabled by default and leave them for the app to enable when the configuration is provided by the user (to make that possible, we enabled `webhookUpdate` mutation for apps)

lkostrowski:
@patrys we want apps to control webhooks enabled/disabled. Many apps initially will have inactive webhooks. We can either:
1. Have them statically declared in manifest as `isActive: false` and app will make it active in runtime
2. Dont register it at all - app will create webhook in runtime with `webhookCreate`

IMO option 1 is slightly better, because its more explicit what webhooks app will register (but app can still hide some webhooks. Also, this field was already added so removing it is an API change

kadewu (approved):

IKarbowiak (approved):

korycins (approved):

maarcingebala (approved):

PR #12716, opened 2023-04-30
Add cache for SHIPPING_LIST_METHODS_FOR_CHECKOUT event

maarcingebala:
`SHIPPING_LIST_METHODS_FOR_CHECKOUT` sync event can set up to 3 requests to external service within one mutation call. This PR introduces a cache that stores responses from the external service. Details:
- Static checkout payload is used as a cache key (generated with `generate_checkout_payload` function); with two datetime fields removed from the original payload, that change between requests (`last_change` and `issued_at`).
- any change to the checkout instance will result in a different payload which invalidates the cache
- cache stores responses for 5 minutes

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/plugins/webhook/plugin.py
+++ b/saleor/plugins/webhook/plugin.py
@@ -1580,14 +1580,14 @@ class WebhookPlugin(BasePlugin):
                 )
                 response_data = cache.get(cache_key)
 
-                if not response_data:
+                if response_data is None:
                     response_data = trigger_webhook_sync(
                         event_type=WebhookEventSyncType.SHIPPING_LIST_METHODS_FOR_CHECKOUT,
                         payload=payload,
                         webhook=webhook,
                         subscribable_object=checkout,
                     )
-                    if response_data:
+                    if response_data is not None:
                         cache.set(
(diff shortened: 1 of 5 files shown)


IKarbowiak (approved):

kadewu (approved):

przlada (approved):

szczecha (approved):

saleor/plugins/webhook/plugin.py
```diff
@@ -1566,12 +1575,25 @@ def get_shipping_methods_for_checkout(
...
+                        webhook=webhook,
+                        subscribable_object=checkout,
+                    )
+                    if response_data:
+                        cache.set(
+                            cache_key,
+                            response_data,
+                            CACHE_TIME_SHIPPING_LIST_METHODS_FOR_CHECKOUT,
+                        )
+
```
NyanKiyoshi, line 1596: Having `[]` is a valid response, shouldn't it be cached? Thus instead of `if [not] response: ...` do `if response is [not] None: ...`

Issue #12733, opened 2023-05-02
Bug: The `completeCheckout` is crashing when lines changed in the meantime


IKarbowiak:
### What are you trying to achieve?

The `IntegrityError` is raised when the lines have changed in the middle of the checkout complete process.

### Steps to reproduce the problem

1. Start the `checkoutcomplete` process for valid checkout with payment
2. Remove the line in the meantime

### What did you expect to happen?

The checkout should be processed for actual line data.

### Logs

https://saleor.sentry.io/issues/4135442525/?project=6417854

### Environment

Saleor version: 3.12

PR #12734, opened 2023-05-02
Fix crashing `CheckoutComplete` when the line is deleted in the meantime

IKarbowiak:
Fix the `IntegrityError` error when the lines have changed in the middle of the checkout complete process.

Fixes https://github.com/saleor/saleor/issues/12733

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested

### DIFF
--- a/saleor/checkout/complete_checkout.py
+++ b/saleor/checkout/complete_checkout.py
@@ -1307,7 +1307,7 @@ def complete_checkout(
             redirect_url=redirect_url,
         )
 
-        reservations = _reserve_stocks_without_availability_check(checkout_info, lines)
+        _reserve_stocks_without_availability_check(checkout_info, lines)
 
     # Process payments out of transaction to unlock stock rows for another user,
     # who potentially can order the same product variants.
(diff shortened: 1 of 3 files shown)


saleor/checkout/complete_checkout.py
```diff
@@ -1291,6 +1296,11 @@ def complete_checkout(
             order = Order.objects.get_by_checkout_token(checkout_info.checkout.token)
             return order, False, {}
 
+        # We need to refetch the checkout lines to ensure that we process checkout
+        # for correct data.
+        # Lines are selected for update by checkout select for update in line 1421
+        _refetch_checkout_lines(checkout, checkout_info, discounts, manager)
```
NyanKiyoshi, line 1302: Aren't we reintroducing a slow 3rd party call inside a transaction? From what I see `_refetch_checkout_lines` involves synchronous webhooks which do not belong inside DB transactions

IKarbowiak: Hm, so maybe it should be refetched just before the transaction block. What do you think?

IKarbowiak: Yes, it will be a little chance that something will change. We cannot add the lock for the entire checkout process as the stocks will be frozen too.

saleor/checkout/complete_checkout.py
```diff
@@ -1339,63 +1345,29 @@ def complete_checkout(
...
 
-        # Run pre-payment checks to make sure, that nothing has changed to the
-        # checkout, during processing payment.
+        # We need to refetch the checkout info to ensure that we process checkout
+        # for correct data.
+        lines, _ = fetch_checkout_lines(checkout, skip_recalculation=True)
+        checkout_info = fetch_checkout_info(checkout, lines, discounts, manager)
+
         checkout_info.checkout.voucher_code = None
-        _, _, post_payment_order_data = complete_checkout_pre_payment_part(
```
korycins, line 1345: Are you sure of this? We have two separate blocks of locks. It means that between these blocks something can happen to the checkout, like adding more lines to the checkout etc. Pre_payment_part has checks like validate if checkout is fully_paid, etc.

IKarbowiak: We were discussing this with @patrys and we agreed that the second part should only fetch the checkout again without doing any calculations or validations. Any mismatch should be done by the shop owner. @patrys could you confirm? Or maybe I misunderstood something?

jakubkuc (approved):

saleor/checkout/complete_checkout.py
patrys: @korycins At this stage we've already collected the payment, we can either continue with the current state of checkout, or attempt a refund (which can also fail). I think a good first approach is to create the order and let CX deal with any discrepancies as those will be edge cases and any outstanding unpaid balance will be visible.

fowczarek (approved):

maarcingebala (approved):

szczecha (approved):

PR #12788, opened 2023-05-10
Add migration for recalculating minimal prices

IKarbowiak:
Recalculate discounted product and variant prices.

The migration plan:
- 3.13 (https://github.com/saleor/saleor/pull/12805):
  - add nullable `discounted_price` field in `ProductVariantChannelListings`
  - add a migration with the use of tasks for recalculating all product prices for which any variant has an empty `discounted_price` field. It means that the price hasn't been updated.
  - update the code to call product recalculation in all places that change variants, products, collections, categories, or sales
  - move the celery task that will call product recalculation every time the sale toggle.
- 3.14 (current main):
   - move the logic from the migration task to the migration code 
   ❗  The migration code is base on the task code that is responsible for updating the prices: method `update_products_discounted_price` in `saleor/product/utils/variant_prices.py` 
   - drop the migration task
   - use the new `discounted_price` in prices recalculations

# Impact

* [ ] New migrations

# Pull Request Checklist

* [ ] The changes are tested


korycins (approved):

kadewu (approved):

saleor/product/migrations/0183_calculate_discounted_prices.py
```diff
@@ -0,0 +1,395 @@
...
+        update_products_discounted_price(
+            product_model,
+            collection_product_model,
+            product_listing_model,
+            variant_listing_model,
+            variant_model,
+            products,
+            discounts,
+        )
+        update_discounted_prices_task(
```
kadewu, line 175: I would suggest to test this migration on some bigger dataset to check if we can manage that recursion.

IKarbowiak: I tested it on a DB with 1.5 M of products and it goes smoothly, each `update_discounted_prices_task` is run for about 1 second.

korycins (approved):

kadewu (approved):

saleor/product/migrations/0183_calculate_discounted_prices.py
```diff
@@ -0,0 +1,402 @@
...
+    for sale_pk, category_pks in category_map.items():
+        subcategory_map[sale_pk] = set(
+            category_model.tree.filter(pk__in=category_pks)
+            .get_descendants(include_self=True)
+            .values_list("pk", flat=True)
+        )
+    return subcategory_map
+
+
+def update_discounted_prices_task(
```
fowczarek, line 157: Should we remove `task` from the name of this function?

fowczarek: Maybe we should store the original name in the comment if needed? But I don't know the use case when we could need to use the old name of this function.

IKarbowiak: See `0041_add_calculation_transaction_events`, we are doing that in other migrations that have tasks in the previous version.

PR #12864, opened 2023-05-19
Fix recalculating tax in draft order on country change

### LINKED ISSUE #12861: Bug: Draft order uses incorrect tax rate for lines on changing order's country

Create a draft order with taxes for one country, and changing the country to another one, and have the taxes correctly recalculated with a new rate.

The easiest way to recreate the issue is be using Saleor Dashboard:
1. Create channel A with the default country e.g. Germany.
2. Enable flat rates in channel A; create a tax class; configure tax rates for Germany by setting some value for the tax class, and leaving the "Country default rate" empty.
3. Add tax configuration for some other country, e.g. Poland, and provide tax rates for both the tax class as well as the "Country default rate".
4. Create a draft order without an address in channel A.
5. Add an order line with a product without a tax class (also product type cannot have a tax class). Taxes should be 0, since there is no default country rate or tax class.
6. Set order addresses to some valid addresses from Poland. Default rate for Poland is set and Saleor should recalculate taxes with the Polish rate. Currently, it doesn't happen and taxes stay 0.

Taxes should be correctly recalculated when changing addresses in draft order.

### Logs

_No response_

### Environment

Saleor version: 3.12.22

maarcingebala:
Fixes #12861

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/tax/calculations/order.py
+++ b/saleor/tax/calculations/order.py
@@ -51,10 +51,13 @@ def update_order_prices_with_flat_rates(
             default_tax_rate,
             country_code,
         )
-    elif order.shipping_tax_rate is not None:
-        # Use order.shipping_tax_rate if it was ever set before (it's non-null now).
-        # This is a valid case when recalculating shipping price and the tax class is
-        # null, because it was removed from the system.
+    elif (
+        order.shipping_tax_class_name is not None
+        and order.shipping_tax_rate is not None
+    ):
+        # Use order.shipping_tax_rate if it was ever set before (it's non-null now and
+        # the name is non-null). This is a valid case when recalculating shipping price
+        # and the tax class is null, because it was removed from the system.
         shipping_tax_rate = denormalize_tax_rate_from_db(order.shipping_tax_rate)
     else:
         shipping_tax_rate = default_tax_rate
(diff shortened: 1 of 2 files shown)


saleor/tax/calculations/order.py
```diff
@@ -135,10 +138,11 @@ def update_taxes_for_order_lines(
                 default_tax_rate,
                 country_code,
             )
-        elif line.tax_rate is not None:
-            # line.tax_class can be None when the tax class was removed from DB. In
-            # this case try to use line.tax_rate which stores the denormalized tax rate
-            # value that was originally used.
+        elif line.tax_class_name is not None and line.tax_rate is not None:
```
korycins, line 141: Do we need to have here `line.tax_class_name is not None and line.tax_rate is not None`?
Can't we replace it with `line.tax_rate is not None`. This is the value that you're using in 146. So no need to check `line.tax_class_name`, right?

szczecha (approved):

saleor/tax/calculations/order.py
maarcingebala: Yes, tax_class_name determines that the tax class was ever set here before; in case of draft orders if it was removed from the system (edge-case) we still want to use saved `line.tax.rate`.

PR #12900, opened 2023-05-23
Introduce a command to fix undiscounted total prices

tomaszszymanski129:
Manage command that fixes order and order lines undiscounted prices.
⚠️ Do not merge!

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/core/management/commands/fix_undiscounted_total_prices.py
+++ b/saleor/core/management/commands/fix_undiscounted_total_prices.py
@@ -1,0 +1,157 @@
+from collections import defaultdict
+
+from django.core.management.base import BaseCommand
+from django.db import connection
+from django.db.models import Exists, F, OuterRef
+
+from ....celeryconf import app
+from ....order.models import Order, OrderLine
+from ....order.tasks import send_order_updated
+
+ORDERS_UPDATE_BATCH_SIZE = 500
+SEND_ORDER_UPDATED_BATCH_SIZE = 3500
+RAW_0139_UPDATE_SQL = """
+    UPDATE "order_orderline"
+    SET
+        "undiscounted_total_price_gross_amount" = CASE
+            WHEN NOT (
+                ("order_orderline"."undiscounted_unit_price_gross_amount" * "order_orderline"."quantity") =
+                                                    "order_orderline"."undiscounted_total_price_gross_amount")
+            THEN ("order_orderline"."undiscounted_unit_price_gross_amount" * "order_orderline"."quantity")
+            ELSE "order_orderline"."undiscounted_total_price_gross_amount" END,
+
+        "undiscounted_total_price_net_amount"   = CASE
+            WHEN NOT (
+                ("order_orderline"."undiscounted_unit_price_net_amount" * "order_orderline"."quantity") =
+                                                    "order_orderline"."undiscounted_total_price_net_amount")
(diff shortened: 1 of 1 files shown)


saleor/core/management/commands/fix_undiscounted_total_prices.py
```diff
@@ -0,0 +1,157 @@
...
+            lines_discount_data[data["order_id"]] = (
+                current_discount_gross + discount_amount_gross,
+                current_discount_net + discount_amount_net,
+            )
+
+        for order in orders:
+            discount_amount_gross, discount_amount_net = lines_discount_data.get(
+                order.id
+            )
+            if discount_amount_gross > 0 or discount_amount_net > 0:
```
8r2y5, line 110: Is there a case where `discount_amount_gross` could be negative but `discount_amount_net` positive value?

saleor/core/management/commands/fix_undiscounted_total_prices.py
```diff
@@ -0,0 +1,157 @@
...
+        records = cursor.fetchall()
+    updated_orders_pks_from_lines = set(
+        [record[0] for record in records] if records else []
+    )
+
+    # 0140_fix_order_undiscounted_total
+    updated_orders_pks_from_orders = update_order_undiscounted_price()
+
+    # send webhooks for unique order ids
+    if all_order_pks := updated_orders_pks_from_lines | updated_orders_pks_from_orders:
```
8r2y5, line 149: I don't think variables `updated_orders_pks_from_lines` and `updated_orders_pks_from_orders` needed separately since at the end you need just combined. Could you create one variable for this?

saleor/core/management/commands/fix_undiscounted_total_prices.py
```diff
@@ -0,0 +1,157 @@
...
+
+from django.core.management.base import BaseCommand
+from django.db import connection
+from django.db.models import Exists, F, OuterRef
+
+from ....celeryconf import app
+from ....order.models import Order, OrderLine
+from ....order.tasks import send_order_updated
+
+ORDERS_UPDATE_BATCH_SIZE = 500
```
8r2y5, line 11: Please do not just magic numbers

saleor/core/management/commands/fix_undiscounted_total_prices.py
tomaszszymanski129: It doesn't make sense, a discount is a discount, you would have to probably do something weird like set the percentage to negative but I don't think that's even possible.

saleor/core/management/commands/fix_undiscounted_total_prices.py
tomaszszymanski129: I am, right on this line. Both ID lists come from different queries so we combine them together to shoot hooks only for unique ones. We delete partial sets afterward to save memory.

saleor/core/management/commands/fix_undiscounted_total_prices.py
tomaszszymanski129: It's a copy of already existing code and this PR is not gonna be merged. As for the memory usage I measured it and it looks fine (below 70MB for everything in this script with a set of 400k orders and 450k order lines)

PR #12926, opened 2023-05-24
Fix draft order shipping method update

### LINKED ISSUE #12712: Bug: draft order API doesn't validate provided shipping method

I want to be sure that the shipping method data stored in a draft order is valid.

First scenario:
1. Create a draft order and provide a shipping method ID that doesn't belong to the `Order.shippingMethods` list of valid methods for this order.
2. Draft order is saved.
3. Problem - when only the channel and shipping address are known, the list of shipping methods is still empty (shipping is not required at this point, so we don't return any available methods). Products need to be set first to get the available options. Should we run validation only if the products and shipping addresses are known?

Second scenario:
1. Create a draft order and provide a valid shipping method, address, and products.
2. Change the shipping address to another country where this shipping method is not applicable.
3. Shipping method should be removed from the draft order.

- Draft order API should validate provided shipping method IDs.
- It should clear the shipping method, shipping method name, and shipping tax rate when the method becomes invalid.
- Shipping tax rate should remain `null` when shipping method is not yet set.

### Logs

_No response_

### Environment

Saleor version: 3.9

jakubkuc:
I want to merge this change because it fixes #12712. This PR unifies logic and refactors order shipping method updating via 'DrafOrderUpdate' mutation and 'OrderUpdateShipping' mutation. 
It also fixes that if shipping method is removed via DraftOrderUpdate the shipping prices were not cleared or updated. 

Further investigation and discussion revealed some new conclusions for this bug (more info in comment section #12712).

Acceptance Criteria: 

- Using `draftOrderUpdate` for updating shipping method should update all order fields connected to shipping ( `shippingPrice`, `shippingTaxRate`, `shippingMethodName`, `shippingTaxClass`, `shippingTaxRate` )
- Passing `shippingMethod: null` in the `input` of `draftOrderUpdate` mutation should clear all of the above fields.
- we are still able to change `shipping_address` to one that is not valid for saved shipping method - validation will not let order be finalized in this state. But using either `draftOrderUpdate` or `OrderUpdateShipping` to set a valid shipping address and shipping method will recalculate all the shipping prices and let you finalize order. 

EDIT: 
- We also decided that we want both `draftOrderUpdate` and  'OrderUpdateShipping' to not validate shipping methods availability for draft orders to extend flexibility on draft orders for users. Finalizing draft order will still validate all the data. Updating shipping with 'OrderUpdateShipping'  for unconfirmed orders will still validate the shipping method availability.

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/mutations/draft_order_create.py
+++ b/saleor/graphql/order/mutations/draft_order_create.py
@@ -35,7 +35,6 @@ from ..types import Order
 from ..utils import (
     OrderLineData,
     validate_product_is_published_in_channel,
-    validate_shipping_method,
     validate_variant_channel_listings,
 )
 from .utils import SHIPPING_METHOD_UPDATE_FIELDS, ShippingMethodUpdateMixin
--- a/saleor/graphql/order/mutations/draft_order_create.py
+++ b/saleor/graphql/order/mutations/draft_order_create.py
@@ -5,6 +5,7 @@ import graphene
 from django.core.exceptions import ValidationError
 from graphene.types import InputObjectType
 
+from .utils import ShippingMethodUpdateMixin, SHIPPING_METHOD_UPDATE_FIELDS
 from ....account.models import User
 from ....checkout import AddressType
 from ....core.permissions import OrderPermissions
@@ -35,7 +36,7 @@ from ..types import Order
 from ..utils import (
     OrderLineData,
     validate_product_is_published_in_channel,
-    validate_variant_channel_listings,
+    validate_variant_channel_listings, validate_shipping_method
 )
 
 
@@ -95,7 +96,7 @@ class DraftOrderCreateInput(DraftOrderInput):
     )
 
 
-class DraftOrderCreate(ModelMutation, I18nMixin):
+class DraftOrderCreate(ModelMutation, ShippingMethodUpdateMixin, I18nMixin):
     class Arguments:
         input = DraftOrderCreateInput(
             required=True, description="Fields required to create an order."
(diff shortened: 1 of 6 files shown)


saleor/graphql/order/mutations/draft_order_create.py
```diff
@@ -345,10 +346,22 @@ def save(cls, info, instance, cleaned_input):
...
+            # Process shipping method
+            method = cleaned_input["shipping_method"]
+            if method:
+                shipping_channel_listing = cls.validate_shipping_channel_listing(
+                    method, instance
+                )
+                cls.update_shipping_method(instance, method, shipping_channel_listing)
+                updated_fields.extend(SHIPPING_METHOD_UPDATE_FIELDS)
+            else:
+                cls.clean_shipping_method_from_order(instance)
```
maarcingebala, line 363: It should be called `clear_shipping_method_from_order` IMO, because "clean" refers to a different part of mutations flow, where we're "cleaning" input data for further processing.

maarcingebala: Now it will clear the shipping method when it's not provided in the input, right? Is that expected? I guess we should clear it only when the shipping method is explicitly set to `null` in the input.

saleor/graphql/order/mutations/utils.py
```diff
@@ -25,6 +44,56 @@ def validate_order(cls, order):
             )
 
 
+class ShippingMethodUpdateMixin:
+    class Meta:
+        abstract = True
+
+    @classmethod
+    def clean_shipping_method_from_order(cls, order):
+
```
maarcingebala, line 53:
```suggestion
```

saleor/graphql/order/mutations/utils.py
```diff
@@ -25,6 +44,56 @@ def validate_order(cls, order):
...
+        order.shipping_method = None
+        order.base_shipping_price = zero_money(order.currency)
+        order.shipping_price = zero_taxed_money(order.currency)
+        order.shipping_method_name = None
+        order.shipping_tax_class = None
+        order.shipping_tax_class_name = None
+        order.shipping_tax_class_private_metadata = {}
+        order.shipping_tax_class_metadata = {}
+        invalidate_order_prices(order)
+        order.save(update_fields=SHIPPING_METHOD_UPDATE_FIELDS)
```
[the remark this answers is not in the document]
jakubkuc: I agree. It should be saved outside the method.

saleor/graphql/order/mutations/draft_order_create.py
jakubkuc: It is fixed now. Shipping method will be cleared when it's explicitly set as `None` in input. Also changed the name.

saleor/graphql/order/mutations/utils.py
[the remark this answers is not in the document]
jakubkuc: fixed it

IKarbowiak (approved):

fowczarek (approved):

8r2y5 (approved):

szczecha (approved):

Issue #12539, opened 2023-04-13
Promotion rules base CRUD


IKarbowiak:
## Problem

Introduce the promotion rules that allow applying discounts based on certain conditions. 
Conditional discounts are based on basket predicates such as Product, Variant, Collection, and Category.
Later also Attributes and Quantity. 

The promotion rule will replace the current `Sale`.

## 1. General Assumptions

- Allow creating, updating, and deleting promotion rules
- The promotion rule functionality must cover all sale functionalities:
  - Allow defining conditions based on Product, Variant, Collection, and Category.
  - Allow defining percentage or fixed discount
  - Allow defining discount value per channel
  - Allow setting start and end date
  - Support metadata
- Migrate old sales to new promotions, drop old sale model
- Ensure that old sales and queries can be handled as previously
- Deprecate the old sale mutations and queries
- When the old sale is modified with the use of mutation for promotion rules, the sale will become the promotion rule and couldn't be handled by old sale migrations and queries anymore
- Ensure zero downtime support
  - In the previous version:
    - Create a new promotion rule model
    - Migrate existing sales to a new model, set the flag for those sales that it's a sale in the old style with use of celery task
    - Ensure that the new sales creates the promotion rule object
- Ensure that the `CheckoutLineDisocunt`, `CheckoutShippingDiscount`  and `OrderLineDisocunt`, `OrderShippingDiscount` objects are created and used in the calculations.

### Acceptance Criteria

- [ ] Support for old versions: the old sales should work as before
- [ ] Allow CRUD of promotion rules
- [ ] The Promotion rules support all functionalities as old sales

### Mutations

- Add `promotionCreate` mutation - requires `MANAGE_DISCOUNTS` 
```graphql
type Mutation {
  ...
  promotionCreate(
    input: PromotionCreateInput!
  ): PromotionCreate @doc(category: "Discounts")

}

input PromotionCreateInput @doc(category: "Discounts") {
  name: String!
  startDate: DateTime
  endDate: DateTime
  rules: [PromotionRuleInput!]
}

input PromotionRuleInput @doc(category: "Discounts") {
  type: DiscountValueTypeEnum
  products: [ID!]
  variants: [ID!]
  categories: [ID!]
  collections: [ID!]
  channel_listings: [PromotionRuleChannelListingInput!]
}

input PromotionRuleChannelListingInput @doc(category: "Discounts") {
  channel_id: ID!
  discount_value: PositiveDecimal
}

[...]

PR #12980, opened 2023-05-31
Add promotion rules base CRUD

IKarbowiak:
Implement the promotion rules CRUD.

RFC: https://github.com/saleor/saleor/issues/12539

Docs: https://github.com/saleor/saleor-docs/pull/874

### Acceptance Criteria
- [ ] The promotion can be created, updated, and deleted only by users, and apps with manage discounts permission.
- [ ] The promotion rule can be created, updated, and deleted with manage discounts permission.
- [ ] Catalogue predicate can have one operator on each level. Operators cannot be mixed with another input field

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] Privileged queries and mutations are either absent or guarded by proper permission checks
* [x] Database migrations are either absent or optimized for zero downtime
* [x] The changes are covered by test cases


saleor/graphql/discount/mutations/promotion/promotion_update.py
```diff
@@ -0,0 +1,140 @@
...
+
+        notification_date = instance.last_notification_scheduled_at
+        start_date = clean_input.get("start_date")
+        end_date = clean_input.get("end_date")
+
+        if not start_date and not end_date:
+            return
+
+        send_notification = False
+        for date in [start_date, end_date]:
```
korycins, line 121: I know that the PR is in draft but I saw it in different PR unrelated to this one. So I decided to leave a comment here.
What is the purpose of this `FOR`? I see we iterate over two dates: `start` and `end`, and set send_notification to True, but few lines below, we have a condition for setting `send_notification` for  `end_date`. Which means that end_date requires additional action. So maybe instead of `For` + additional `if` condition for `end_date`, we could have two `if`s - one for `start`, second one for `end`?

fowczarek:
We should use `UUID` instead of enumerated id in all new models.

saleor/graphql/discount/mutations/promotion/promotion_update.py
IKarbowiak: We introduced `promotion_started` and `promotion_ended` webhook (instead of `promotion_toggle`), so I marked this comment as resolved.

szczecha (approved):

michal-macioszczyk (approved):

Air-t (approved):

zedzior (approved):

SzymJ (approved):

PR #12994, opened 2023-06-01
Use `webhookDryRun` mutation's context instead of creating dummy one

zedzior:
I want to merge this change, because it uses mutation's context with user data to generate payload from subscription.

It also disables dry run option for `GiftCardSent` webhook, as current implementation doesn't allow to perform it due to its specific root type. Currently dry run works for webhooks with subscriptable object being model instance.

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/webhook/subscription_types.py
+++ b/saleor/graphql/webhook/subscription_types.py
@@ -561,8 +561,8 @@ class GiftCardSent(SubscriptionObjectType, GiftCardBase):
     )
 
     class Meta:
-        root_type = "GiftCard"
-        enable_dry_run = True
+        root_type = None
+        enable_dry_run = False
         interfaces = (Event,)
         description = (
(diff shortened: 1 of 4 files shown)


fowczarek (approved):

kadewu (approved):

saleor/graphql/webhook/subscription_types.py
```diff
@@ -561,8 +561,8 @@ class GiftCardSent(SubscriptionObjectType, GiftCardBase):
     )
 
     class Meta:
-        root_type = "GiftCard"
-        enable_dry_run = True
+        root_type = None
```
maarcingebala, line 564: This results in removing the `doc_category` in the GraphQL schema. To fix this, we need to add the following to the meta class as well:

```
doc_category = DOC_CATEGORY_GIFT_CARDS
```

8r2y5 (approved):

PR #13029, opened 2023-06-06
Promotion events

zedzior:
It is part of: https://github.com/saleor/saleor/issues/12539
Resolves https://github.com/saleor/saleor/issues/13304

I want to merge this change, because it adds `PromotionEvent` model, which stores log of promotion lifecycle.

**DB changes:**
Add `PromotionEvent` model:
```python
class PromotionEvent:
    date: DateTimeField
    type: CharField
    user: ForeignKey
    app: ForeignKey
    promotion: ForeignKey
    ruleId: CharField
```

**API changes:**
```graphql
type PromotionEvent implements Node @doc(category: "Discounts") {
  id: ID!
  date: DateTime
  type: PromotionEventsEnum
  createdBy: UserOrApp
}

enum PromotionEventsEnum @doc(category: "Discounts") {
  PROMOTION_CREATED
  PROMOTION_UPDATED

  RULE_CREATED
  RULE_UPDATED
  RULE_DELETED
  
  PROMOTION_STARTED
  PROMOTION_ENDED
}

type Promotion{
    ....
    events: [PromotionEvent!]
}
```

# Impact

* [x] New migrations

# Pull Request Checklist


saleor/graphql/discount/types/promotions.py
```diff
@@ -99,3 +115,65 @@ def resolve_promotion(root: models.PromotionRule, info: ResolveInfo):
     @staticmethod
     def resolve_channels(root: models.PromotionRule, info: ResolveInfo):
         return ChannelsByPromotionRuleIdLoader(info.context).load(root.id)
+
+
+class PromotionEvent(ModelObjectType[models.PromotionEvent]):
```
korycins, line 120: Wouldn't be better to have small types dedicated to specific actions instead of generic one?
The current type doesn't contain a lot of fields but probably it will going to grow in the future. We should consider using `UNION` here - like `PromotionCreatedEvent`, `PromotionRuleCrated` etc. 
With union here, we give frontend apps more context on which fields are related to specific events. 
With the current approach, frontend apps need to know how to combine the fields. For example by calling
```graphql
{
    promotions{
        events{
            ruleId
        }
    }
}
```
Frontend needs to be aware that the `ruleId` field will have a value only for specific type of events (related to the rules).  Also frontend needs to be aware that `ruleId` doesn't provide here any value when type of the event is `PROMOTION_CREATED`

saleor/graphql/discount/types/promotions.py
```diff
@@ -99,3 +115,65 @@ def resolve_promotion(root: models.PromotionRule, info: ResolveInfo):
...
+    @staticmethod
+    def resolve_user(root: models.PromotionEvent, info):
+        def _resolve_user(user):
+            requester = get_user_or_app_from_context(info.context)
+            if not requester:
+                return None
+            check_is_owner_or_has_one_of_perms(
+                requester,
+                user,
+                AccountPermissions.MANAGE_USERS,
```
korycins, line 144: Shouldn't we validate here only `MANAGE_STAFF`? Is it possible that the `user` will not be a `Staff`?

saleor/graphql/discount/types/promotions.py
```diff
@@ -99,3 +115,65 @@ def resolve_promotion(root: models.PromotionRule, info: ResolveInfo):
     @staticmethod
     def resolve_channels(root: models.PromotionRule, info: ResolveInfo):
         return ChannelsByPromotionRuleIdLoader(info.context).load(root.id)
+
+
+class PromotionEvent(ModelObjectType[models.PromotionEvent]):
+    id = graphene.GlobalID()
+    date = graphene.DateTime(description="Date when event happened.")
+    type = PromotionEventsEnum(description="Promotion event type.")
+    user = graphene.Field(User, description="User who performed the action.")
```
korycins, line 124: The field contains additional permissions, we should mention it in the description.
Have you considered using `PermissionsField` here? It could help with handling the perms and extending the description

saleor/graphql/discount/types/promotions.py
```diff
@@ -99,3 +115,65 @@ def resolve_promotion(root: models.PromotionRule, info: ResolveInfo):
...
+    app = graphene.Field(App, description="App that performed the action.")
```
korycins, line 125: Have you considered using union for `app` and `user`? 
Something as we have in `saleor.graphql.payment.types.TransactionEvent` - `created_by` field

saleor/discount/models.py
```diff
@@ -388,13 +396,40 @@ class PromotionRule(models.Model):
...
 
+class PromotionEvent(models.Model):
+    date = models.DateTimeField(auto_now_add=True, db_index=True, editable=False)
+    type = models.CharField(max_length=255, choices=PromotionEvents.CHOICES)
+    parameters = JSONField(blank=True, default=dict, encoder=CustomJsonEncoder)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        blank=True,
+        null=True,
+        related_name="promotion_events",
```
przlada, line 415: Is there a specific requirement for a backward relation here? Do we have any plans to access promotion events differently than throughout `Promotion` objects?

saleor/graphql/discount/types/promotions.py
```diff
@@ -99,3 +115,90 @@ def resolve_promotion(root: models.PromotionRule, info: ResolveInfo):
...
+            return app
+
+        if root.user_id:
+            return (
+                UserByUserIdLoader(info.context).load(root.user_id).then(_resolve_user)
+            )
+        if root.app_id:
+            return AppByIdLoader(info.context).load(root.app_id).then(_resolve_app)
+
+        return None
```
przlada, line 165: Where the event creator is purged from the database, both `app` and `user` fields are set to null. As a consequence, the `created_by` resolver will return `None`. This is acceptable, but it also means we lose data regarding the event creator type (if it was an `App` or `User`). Are we comfortable with this loss, or should we contemplate introducing an additional enum field, say, `creator_type`? I faced a similar issue while working on the `App` events, hence my concern.

saleor/discount/models.py
```diff
@@ -388,13 +396,40 @@ class PromotionRule(models.Model):
...
     )
 
     class Meta:
         ordering = ("name", "pk")
 
 
+class PromotionEvent(models.Model):
+    date = models.DateTimeField(auto_now_add=True, db_index=True, editable=False)
+    type = models.CharField(max_length=255, choices=PromotionEvents.CHOICES)
+    parameters = JSONField(blank=True, default=dict, encoder=CustomJsonEncoder)
```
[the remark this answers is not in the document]
zedzior: As discussed, for now we keep `parameters` field. All event related objects will be refactored in nearly future.

saleor/graphql/discount/types/promotions.py
zedzior: No need for an extra field.

IKarbowiak (approved):

saleor/graphql/discount/mutations/promotion/promotion_create.py
```diff
@@ -208,21 +209,31 @@ def _save_m2m(
...
-    def send_promotion_webhooks(
-        cls, manager: "PluginsManager", instance: models.Promotion
+    def post_save_actions(
+        cls,
+        manager: "PluginsManager",
+        instance: models.Promotion,
+        info: ResolveInfo,
+        rules: List[models.PromotionRule],
     ):
         cls.call_event(manager.promotion_created, instance)
```
korycins, line 222: Are you sure that you have the correct order of function calls here? 

222 triggers the promotion_created (and webhook if someone is subscribed)
223 optionally triggers `manager.promotion_started` (and probably second webhook).
224 creates new promotion events. It means that the webhooks triggered from 222&223 will not have details from 224

saleor/graphql/discount/mutations/promotion/promotion_update.py
```diff
@@ -142,3 +145,22 @@ def send_promotion_toggle_webhook(
...
             instance.last_notification_scheduled_at = now
             instance.save(update_fields=["last_notification_scheduled_at"])
+            if event == manager.promotion_started:
+                return "started"
+            if event == manager.promotion_ended:
+                return "ended"
+        return None
+
+    @classmethod
+    def save_events(
```
korycins, line 155: The method name is misleading. It suggests making only save, but you also optionally trigger webhook

saleor/graphql/discount/mutations/promotion/promotion_update.py
```diff
@@ -142,3 +145,22 @@ def send_promotion_toggle_webhook(
...
+        cls,
+        info: ResolveInfo,
+        instance: models.Promotion,
+        sent_webhook_type: Optional[str],
```
korycins, line 159: Also to be precise. the name `sent_webhook_type` is not correct. You don't directly trigger the webhook. You trigger the event on PluginsManager side. WebhookPlugin is one of many plugins that could be subscribed to the event triggered by this method.

saleor/graphql/discount/types/promotion_events.py
```diff
@@ -0,0 +1,193 @@
...
+        return PROMOTION_EVENT_MAP.get(instance.type)
+
+    @staticmethod
+    def resolve_created_by(root: models.PromotionEvent, info):
+        requester = get_user_or_app_from_context(info.context)
+        if not requester:
+            return None
+
+        def _resolve_user(user):
+            check_is_owner_or_has_one_of_perms(
```
korycins, line 49: You don't say anything about `owner` in `created_by` description

saleor/graphql/discount/types/promotion_events.py
```diff
@@ -0,0 +1,193 @@
...
+    events.PROMOTION_UPDATED: PromotionUpdatedEvent,
+    events.PROMOTION_STARTED: PromotionStartedEvent,
+    events.PROMOTION_ENDED: PromotionEndedEvent,
+    events.RULE_CREATED: PromotionRuleCreatedEvent,
+    events.RULE_UPDATED: PromotionRuleUpdatedEvent,
+    events.RULE_DELETED: PromotionRuleDeletedEvent,
+}
+
+
+class UnionPromotionEvent(graphene.Union):
```
korycins, line 187: Where do you use this? 
Also, wouldn't be better to have `PromotionEventInterface` and rename `UnionPromotionEvent` to `PromotionEvent`. Then connect `PromotionEvent` (union) to the `Promotion`? 
Then you will have connected `union` as `Promotion.event`. It should be much more easier to use. 
(You will be able to make something like this:
```graphql
events {
    ...on PromotionUpdatedEvent{
        id
    }
    ...on PromotionStartedEvent{
        id
   }
```

Currently, as far as I understand the tests, you can't do that ☝️ , you have generic separation based on the interface. 
```graphql
    query Promotion($id: ID!) {
         promotion(id: $id) {
             id
             events {
                 ... on PromotionEvent {
                     type
                     createdBy {
                         ... on User {
                             id
                         }
                     }
                 }
                 ... on PromotionRuleEvent {
                     ruleId
                 }
             }
         }
     }
```
And this is the reason why you had to include `PROMOTION_EVENT_MAP` in `api.py`. You don't use it currently.

korycins (approved):

PR #13033, opened 2023-06-07
[Test] Unlogged customer should be able to order digital product

fowczarek:
I want to merge this change because of adding the first implementation of the API tests. 

Test case covered in this PR:
[Unlogged customer should be able to order digital product](https://saleor.testmo.net/repositories/5?group_id=125&case_id=747)

> **Warning**
> In separate PR I will:
> - add helper for assigning permissions to API clients
> - create separate API clients 

To clean:
- [x] Drop assert is not None from `test_process_checkout_with_digital_product` 
- [x] Add asserts for that response from utils don't have errors inside. 
- [x] Change catalog name from products to product

# Impact

* [ ] New migrations

# Pull Request Checklist


saleor/tests/API/checkout/test_unlogged_customer_should_be_able_to_order_digital_product.py
```diff
@@ -0,0 +1,109 @@
...
+def test_process_checkout_with_digital_product(
+    api_client,
+    staff_api_client,
+    permission_manage_product_types_and_attributes,
+    permission_manage_channels,
+    permission_manage_products,
+    permission_manage_shipping,
+    media_root,
+):
+    # given
```
tomaszszymanski129, line 30: 👮🏻

IKarbowiak (approved):

tomaszszymanski129 (approved):

saleor/tests/API/__init__.py
maarcingebala, line 1: I was thinking about naming this `e2e` instead of `API` since we do have API tests in other places, so it might be slightly confusing what is the purpose of this catalog.

saleor/tests/API/checkout/test_unlogged_customer_should_be_able_to_order_digital_product.py
```diff
@@ -0,0 +1,109 @@
...
+from ..warehouse.utils import create_warehouse
+from .utils import (
+    checkout_billing_address_update,
+    checkout_complete,
+    checkout_create,
+    checkout_dummy_payment_create,
+)
+
+
+def test_process_checkout_with_digital_product(
```
maarcingebala, line 21: Maybe each test should come with a comment or a docstring explaining the purpose or the test scenario? Could be useful for maintenance of such tests if we have more of them, and many will be similar but testing slightly different things? @saleor/qa What do you think?

maarcingebala: Or not to copy the scenario, paste a link to Testmo? So that there is some connection between the test and the scenario.

8r2y5 (approved):

saleor/tests/API/checkout/test_unlogged_customer_should_be_able_to_order_digital_product.py
szczecha: Not everyone has access to testmo, so comments may be more useful here. I think it should at least have a test case ID (e.g for this test CORE_0101) and in case the title isn't descriptive enough we can provide more context in a comment.

PR #13079, opened 2023-06-15
Fix draft order shipping method update

jakubkuc:
I want to merge this change because it is partially a port of #12926
Changes that were introduced in 3.11 forced some changes in the original PR. 

Here are updated Acceptance Criteria: 
- Using `draftOrderUpdate` for updating shipping method should update all order fields connected to shipping ( `shippingPrice`, `shippingTaxRate`, `shippingMethodName`, `shippingTaxClass`, `shippingTaxRate` )
- Passing `shippingMethod: null` in the `input` of `draftOrderUpdate` mutation should clear all of the above fields.
- we are still able to change `shipping_address` to one that is not valid for saved shipping method - validation will not let order be finalized in this state. But using either `draftOrderUpdate` or `OrderUpdateShipping` to set a valid shipping address and shipping method will recalculate all the shipping prices and let you finalize order. 
- We also decided that we want both `draftOrderUpdate` and  'OrderUpdateShipping' to not validate shipping methods availability for draft orders to extend flexibility on draft orders for users. Finalizing draft order will still validate all the data. Updating shipping with 'OrderUpdateShipping'  for unconfirmed orders will still validate the shipping method availability.
❗ UPDATED ❗ 
- If shipping method is not applicable (e.g. order does not require shipping, has no lines or there is no address) the shipping price will be updated to zero. (this is how it is working right now in 3.11 so I didn't want to change that behaviour)

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/mutations/order_update_shipping.py
+++ b/saleor/graphql/order/mutations/order_update_shipping.py
@@ -1,8 +1,7 @@
 import graphene
 from django.core.exceptions import ValidationError
 
-from ....order import OrderStatus
-from ....order import models
+from ....order import OrderStatus, models
 from ....order.error_codes import OrderErrorCode
 from ....permission.enums import OrderPermissions
 from ....shipping import models as shipping_models
@@ -125,6 +124,7 @@ class OrderUpdateShipping(
         if order.status != OrderStatus.DRAFT:
             clean_order_update_shipping(order, shipping_method_data, manager)
         cls.update_shipping_method(order, method, shipping_channel_listing)
+        cls._update_shipping_price(order, shipping_channel_listing)
 
         order.save(update_fields=SHIPPING_METHOD_UPDATE_FIELDS)
         # Post-process the results
--- a/saleor/graphql/order/mutations/order_update_shipping.py
+++ b/saleor/graphql/order/mutations/order_update_shipping.py
@@ -1,10 +1,9 @@
 import graphene
 from django.core.exceptions import ValidationError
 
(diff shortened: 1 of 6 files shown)


fowczarek (approved):

IKarbowiak (approved):

saleor/graphql/order/mutations/order_update_shipping.py
```diff
@@ -133,54 +114,19 @@ def perform_mutation(  # type: ignore[override]
...
         shipping_method_data = convert_to_shipping_method_data(
             method,
             shipping_channel_listing,
         )
         manager = get_plugin_manager_promise(info.context).get()
-        clean_order_update_shipping(order, shipping_method_data, manager)
-
-        order.shipping_method = method
-        order.shipping_method_name = method.name
+        if order.status != OrderStatus.DRAFT:
```
korycins, line 124: Can you explain, why we clean up the order shipping only for the orders that are not draft? It is not clear. Would be also good to probably left a comment here

8r2y5 (approved):

saleor/graphql/order/mutations/order_update_shipping.py
jakubkuc: `clean_order_update_shipping` validates shipping method. We don't want to do that for drafts - this enables users to be more flexible with drafts (e.g. add shipping method to draft without address). This mutation can be used not only on draft orders (also unconfirmed) so we want to skip this validation for orders with status `DRAFT`

szczecha (approved):

Issue #13078, opened 2023-06-14
Bug: `accountAddressCreate` – some fields are not saved but there's no error


mmiszy:
### What are you trying to achieve?

`accountAddressCreate` with some of the input fields invalid results in the address being saved but the invalid fields are saved as empty strings. No errors are returned. Particularly this is true for `postalCode` and `countryArea` (maybe other fields too, I don't know).

### Steps to reproduce the problem

https://master.staging.saleor.cloud/graphql/#saleor/N4IgjiBcILYK4BcCGCCWB7AdgAgIIGN905MFcATcgJwFMBnOgYVpRoAoASVTAB0UjyVaDAJK9EAQgA02DggCePGgIrV6dACqKaAUUxwYASmzAAOjmxJCxUquFMWCdtz4IBXcQhkKl7nzWMzC2xsGioqdCo6E3MQuOwAM1QaABtyWPjsGHUkAHMaDPiicgLguKQhdS0lQuwAX1qKtQYYspDUdLbE1CiEADkkbNqQlKQ6fsHSzOwiGB4kTHkBoa7x2hoySoYARmHsNZoNu3UAJj38VAVzy-lcFnObBCpb+66edHGkFMZ0Er2eAAWWCmmSIJCe8la0xCxRB0zBpGeewaZRR8TgdDCUMyTXs9Gx0w6exCSV6yzhmVG43JxJm6DmCyWk1pByOWzouy6IVZm2adDOXJmN1pFwUdxoSBFj2e4slgven2+vwp8UBwKl4OeBOhsNpMOl8lpaMyxpCxpRdRAUhA6CgIAICN59mYEqcVpAADc7UEQqYQC5EH6BD64n7RYaoNg-YwbthMMyrbUwwag1GQBoNH6pEmQI6ZfdI36sznSdSEwI-QApBbFsp+qkTIaFkCMIG1+J+hXIJV-ZsAZhOAFoTn3dom6yAeccdqm-dt52PsxOp+yzs3tiOixY6kvfSB-LOQABlAASIgACueRH0AOJby11IA

### What did you expect to happen?

Error

### Environment

Saleor version: master

FremahA:
@mmiszy I do not have the permissions to recreate the problem but I want to link a pull request that hopefully solves this.

korycins:
The issue reported here is related to the country code that were used in the example: TT
Saleor uses https://github.com/mirumee/google-i18n-address to fetch the validation rules for address format.
[...]

PR #13086, opened 2023-06-15
Fix the bug related to invalid fields in account address creation

FremahA:
I want to merge this change because fixes the bug related to invalid fields in account address creation.

It solves #13078 

# Impact

* [ ] New migrations

# Pull Request Checklist


FremahA:
> @FremahA Looks good, but we would also need some unit tests for these changes.

Great. Working on this.

CHANGELOG.md
```diff
@@ -29,6 +30,8 @@ All notable, unreleased changes to this project will be documented in this file.
 ### Other changes
 
 - Expand metric units to support more types of products. - #13043 by @FremahA
+
+- Fix Error Cannot return null for non-nullable field Webhook.name. - #12989 by @cyborg7898
```
maarcingebala, line 34: It looks unrelated to this PR so let's remove this change.

saleor/graphql/account/tests/test_account_handle_invalid_fields.py
```diff
@@ -0,0 +1,53 @@
...
+    valid_input = {
+        "firstName": "John",
+        "lastName": "Doe",
+        "streetAddress1": "123 Main St",
+        "city": "New York",
+        "postalCode": "12345",
+    }
+
+    valid_fields = AccountAddressCreate.get_invalid_fields(valid_input)
+    assert len(valid_fields) == 0
```
fowczarek, line 31: It should be a separate test.

fowczarek (approved):

maarcingebala (approved):

FremahA:
@maarcingebala I see the build is failing because of five failed tests in the test_address_create.py file. The `graphql_address_data` fixture used in those two have two empty fields which the `handle_invalid_fields` function is catching. Adding a commit that fixes this.

saleor/tests/fixtures.py
```diff
@@ -818,11 +818,9 @@ def graphql_address_data():
         "lastName": "Doe Mirumee",
         "companyName": "Mirumee Software",
         "streetAddress1": "Tęczowa 7",
-        "streetAddress2": "",
```
maarcingebala, line 821: Do we have to remove these fields? I would keep them; they can be empty (at least for PL) so they should be valid data.

maarcingebala:
After the recent change, there are still some tests failing. I think we should not change this fixture and keep those fields. If needed, we should provide some new fixtures or modify the existing ones inside tests.

saleor/graphql/account/mutations/account/account_address_create.py
```diff
@@ -75,6 +80,29 @@ def perform_mutation(  # type: ignore[override]
                 utils.change_user_default_address(user, address, address_type, manager)
         return AccountAddressCreate(user=user, address=address)
 
+    @classmethod
+    def get_invalid_fields(cls, cleaned_input):
+        invalid_fields = []
+        for field, value in cleaned_input.items():
+            if not value and field not in (
```
kadewu, line 87: I'm not sure if this is really correct way to handle the issue. I would rather expect that there should be problem within `clean` method where validation of the mentioned field in issue is happening.
Didn't we also allow providing e.g empty `firstName` before, is this expected to remove such possibility? @maarcingebala

saleor/graphql/account/tests/test_account_handle_invalid_fields.py
```diff
@@ -0,0 +1,55 @@
+import pytest
```
kadewu, line 1: Do we need separate file for those tests? Shouldn't this be part of `test_account_address_create`?

saleor/graphql/account/mutations/account/account_address_create.py
```diff
@@ -67,6 +68,10 @@ def perform_mutation(  # type: ignore[override]
         cleaned_input = cls.clean_input(info=info, instance=Address(), data=input)
         with traced_atomic_transaction():
             address = cls.validate_address(cleaned_input, address_type=address_type)
+            # Check for invalid fields and return errors if necessary
```
kadewu, line 71: And just small comment, it's nothing big, but I personally do not like to write comments that are redundant. By that I mean, you named your methods correctly, and the code is already self explanatory.

saleor/graphql/account/tests/test_account_handle_invalid_fields.py
FremahA: @kadewu I created a separate test file for easy review. However, I'm open to adding the tests to the main test file if you prefer.

saleor/graphql/account/mutations/account/account_address_create.py
FremahA: @kadewu Thanks for the feedback. I'll keep this in in mind.

szczecha:
Sending empty not required field raising an error: Screenshot 2023-07-03 at 12 58 55

On main branch, the above mutation creates an address object with an empty first name

saleor/graphql/account/mutations/account/account_address_create.py
kadewu: I'm not sure if I follow. The bug is happening in core code, as pointed in raised issue, what does frontend have to do with this? And what package exactly you want to introduce?

My concern is, that this code here feels hacky (e.g. you change behaviour of `firstName` field), and I'm not sure if this is right place for fix. Did you check what is happening within the mixture for address validation?

saleor/graphql/account/tests/test_account_handle_invalid_fields.py
kadewu: There is no reason to create separate file for just review purpose, as I believe there is no difference anyway. If that was the only reason, I would suggest to move the tests to the proper test file. It may be also a good idea then to follow common pattern for testing mutations.

FremahA: Thanks for the feedback. Working on this.

saleor/graphql/account/mutations/account/account_address_create.py
FremahA: @kadewu, I apologise if my previous response was unclear. I wanted to better understand the issue by exploring the frontend validation and data formatting aspects.

I acknowledge that the bug is occurring in the core code, specifically within the clean method responsible for address validation. I understand your concern that modifying the behaviour of the "firstName" field may seem like a quick fix rather than addressing the root cause of the issue.

I appreciate your suggestion to investigate the clean method for a more appropriate fix. I will focus on analysing that part of the code to identify any problems or potential improvements. 

Thank you for your patience and valuable input regarding this issue. 

Please feel free to share any additional questions or suggestions you may have regarding this matter.

FremahA:
> Sending empty not required field raising an error: Screenshot 2023-07-03 at 12 58 55
> 
> On main branch, the above mutation creates an address object with an empty first name

@szczecha @kadewu Can you help with the part of the code where the required fields are defined? The `AddressInput` class has no required fields so it is hard to know which fields can be empty.

szczecha:
> @szczecha @kadewu Can you help with the part of the code where the required fields are defined? The `AddressInput` class has no required fields so it is hard to know which fields can be empty.

There is a query addressValidationRules that returns the required fields for the selected country. 
Screenshot 2023-07-05 at 08 16 55

saleor/graphql/account/mutations/account/account_address_create.py
kadewu: No need to apologize, cause nothing happened. ;)

Just side note, it may happen that fixing it inside mentioned method may not be possible, please also remember that the `I18nMixin` is used by other mutations.

korycins (commented):
I don't think that this is the correct approach. 
- The address validation rules should be the same for all mutations that use addresses. For example checkoutCreate, accountAddressCreate, warehouse etc
- Sometimes the empty field is a valid response: Company address, empty first/last name (I might be wrong now, but I think that in some countries you could use only first/last name.)

---

The issue reported here is related to the country code that were used in the example: `TT`
Saleor uses https://github.com/mirumee/google-i18n-address to fetch  the validation rules for address format. 
Here you have an example for PL: https://www.gstatic.com/chrome/autofill/libaddressinput/chromium-i18n/ssl-address/data/PL
(see the structure of it)
Then see the file for TT https://www.gstatic.com/chrome/autofill/libaddressinput/chromium-i18n/ssl-address/data/TT
TT doesn't have validation rules. This is why we don't receive the error. I believe that we should investigate why we overwrite the input fian eld with empty value.
The fix for this should be inside the address validation logic, not inside a specific mutation as we want to have the same address validation rules in all mutations.

aniav:
I will close this PR because of inactivity. Feel free to reopen when you decide to work on it further 🙌

PR #13108, opened 2023-06-17
add function return type and utilised list comprehension

AjmalPonneth:
I want to merge this change because the change reduced the line of code by using list comprehension and providing a function return type.

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/permission/utils.py
+++ b/saleor/permission/utils.py
@@ -10,7 +10,7 @@ if TYPE_CHECKING:
 
 def one_of_permissions_or_auth_filter_required(
     context, permissions: Iterable[BasePermissionEnum]
-):
+) -> bool:
     """Determine whether user or app has rights to perform an action.
 
     The `context` parameter is the Context instance associated with the request.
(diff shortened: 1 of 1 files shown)


saleor/permission/utils.py
```diff
@@ -32,9 +32,7 @@ def one_of_permissions_or_auth_filter_required(
     requestor = get_user_or_app_from_context(context)
 
     if requestor and permissions:
-        perm_checks_results = []
-        for permission in permissions:
-            perm_checks_results.append(requestor.has_perm(permission))
+        perm_checks_results = [requestor.has_perm(perm) for perm in permissions]
```
8r2y5, line 35: This variable could be a generator. Passing it to `any` would result in stopping the iteration on the first `True` returned by it.

AjmalPonneth: > Current implementation also checking the exact same thing. Checking any permission from the list after the iteration and here we are doing the same thing with list comprehension.

maarcingebala (approved):

IKarbowiak (approved):

PR #13128, opened 2023-06-19
Added metadata to Shop object

cyborg7898:

# Impact

* [ ] New migrations

# Pull Request Checklist


saleor/graphql/shop/mutations.py
```diff
@@ -169,6 +186,17 @@ def clean_input(cls, _info, _instance, data):
...
+            metadata = data.pop("metadata", [])
+            metadata_dict = {item["key"]: item["value"] for item in metadata}
+            data["metadata"] = metadata_dict
+
+        if "private_metadata" in data:
+            private_metadata = data.pop("private_metadata", [])
+            private_metadata_dict = {
+                item["key"]: item["value"] for item in private_metadata
+            }
+            data["private_metadata"] = private_metadata_dict
```
maarcingebala, line 199: I think this can be skipped since this is a `BaseMutation` we should be able to use set `support_meta_field=True` in the meta class and use `validate_and_update_metadata` in the `perform_mutation`. See other usages of `validate_and_update_metadata` for reference.

cyborg7898:
How we are storing metadata `{"kye":"foo","value":"bar"}` like this or `[
      {
        "key": "key1",
        "value": "value1"
      },
      {
        "key": "key2",
        "value": "value2"
      },
      {
        "key": "key3",
        "value": "value3"
      }
    ]`? I'm referring user metadata.

saleor/graphql/shop/types.py
```diff
@@ -551,3 +555,13 @@ def resolve_include_taxes_in_prices(_, _info, site):
     @load_site_callback
     def resolve_display_gross_prices(_, _info, site):
         return site.settings.display_gross_prices
+
+    @staticmethod
+    @load_site_callback
+    def resolve_metadata(root: ModelWithMetadata, info: ResolveInfo, site):
```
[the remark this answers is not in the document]
maarcingebala: I meant that we have three fields for metadata: `metadata`, `metafield` and `metafields`, and the same for private metadata. I think we need to add resolvers for all these fields.

maarcingebala:
@Smit-Parmar `metadata` input can be provided as a list or as a dict, so, both of these inputs are valid:
```
accountRegister(
    input: {
      email: "test1@example.com",
      password: "password",
      metadata: [
        {key: "example-key", value: "example-value"},
        {key: "example-key-2", value: "example-value-2"}
      ],
      redirectUrl: "http://example.com/", 
      channel: "default-channel"
    }
)
```

```
accountRegister(
    input: {
      email: "test2@example.com",
      password: "password",
      metadata: {key: "example-key", value: "example-value"},
      redirectUrl: "http://example.com/", 
      channel: "default-channel"
    }
  )
```

The only difference is the first one will allow you to provide one metadata item, and with the second one you can define as many as you want. In both cases, we store this structure as a dict in the database:

```
In [9]: User.objects.filter(email="test1@example.com").first().metadata
Out[9]: {'example-key': 'example-value', 'example-key-2': 'example-value-2'}

In [10]: User.objects.filter(email="test2@example.com").first().metadata
Out[10]: {'example-key': 'example-value'}
```

Smit-Parmar:
Thank you @maarcingebala

saleor/graphql/shop/mutations/shop_settings_update.py
```diff
@@ -94,6 +101,16 @@ class ShopSettingsInput(graphene.InputObjectType):
...
+    metadata = common_types.NonNullList(
+        MetadataInput,
+        description="Shop public metadata." + ADDED_IN_315,
+        required=False,
+    )
+    private_metadata = common_types.NonNullList(
+        MetadataInput,
+        description=("Shop private metadata." + ADDED_IN_315),
+        required=False,
+    )
```
maarcingebala, line 113: Let's move these declarations above the `#deprecated` comment in line 82. This comment groups the deprecated fields together for better visibility. Metadata fields obviously don't belong to this group.

maarcingebala:
@Smit-Parmar Please also update the `CHANGELOG.md` file with an entry about this change.

maarcingebala (approved):

saleor/graphql/meta/resolvers.py
```diff
@@ -101,6 +104,7 @@ def check_private_metadata_privilege(root: ModelWithMetadata, info: ResolveInfo)
             "Make sure that model exists inside MODEL_TO_TYPE_MAP."
         )
 
+    # print("107",item_type.__name__)
```
IKarbowiak, line 107: Print in comment to remove

IKarbowiak (changes requested):
We should also add the tests for meta mutations, as when the metadata is added to the new type it should be also supported in meta mutations.

Smit-Parmar:
> We should also add the tests for meta mutations, as when the metadata is added to the new type it should be also supported in meta mutations.

@IKarbowiak Can you please specify more about new test case? There is already one test for mutation with metadata.

IKarbowiak:
@Smit-Parmar sure! Let's take a look at [saleor/graphql/meta/tests/mutations](https://github.com/saleor/saleor/tree/main/saleor/graphql/meta/tests/mutations), there are tests for meta mutations for each module that supports metadata. You should add here a new file for shop testing add a tests for meta mutations for update, delete private and public metadata. Also, [here saleor/graphql/meta/tests/queries](https://github.com/saleor/saleor/tree/main/saleor/graphql/meta/tests/queries) are the tests for fetching the metadata and you should add a new file with tests here as well.

Smit-Parmar (commented):
@IKarbowiak I was trying to add test cases in meta but here scenario is different . Type we are using is `Shop` and model behind `Shop` is `SiteSettings`. We don't have any query for `SiteSettings` and `Shop` don't have attribute `Id` because of this test case are failing. 
```
def test_delete_private_metadata_for_shop(
    staff_api_client, permission_manage_settings, site_settings
):
    # given
    site_settings.store_value_in_private_metadata({PRIVATE_KEY: PRIVATE_VALUE})
    site_settings.save(update_fields=["private_metadata"])
    site_setting_id = graphene.Node.to_global_id("SiteSettings", site_settings.pk)

    # when
    response = execute_clear_private_metadata_for_item(
        staff_api_client, permission_manage_settings, site_setting_id, "SiteSettings"
    )

    # then
    assert item_without_private_metadata(
        response["data"]["deletePrivateMetadata"]["item"], site_settings, site_setting_id
    )
```
To make this work I think we need to add resolver for `SiteSettings` .

szczecha:
@Smit-Parmar 
There is an error when providing metadata dict in graphql variables 
Screenshot 2023-06-28 at 13 35 13

When it is provided in request works fine. 
Screenshot 2023-06-28 at 13 38 05

Smit-Parmar:
@szczecha Whenever we are passing metadata via input variable then we are passing metadata inside the array something like this 
```
"input": {
    "metadata": [
      {
        "key": "key_one",
        "value": "value_one"
      }
    ]
  }
```
But while passing input direct to mutation query we are passing like 
```
shopSettingsUpdate(
    input: {metadata: {key: "foo1", value: "bar1"}, privateMetadata: {key: "foo1", value: "bar1"}
  ) 
```
That's same thing happening in `productUpdate`.

Smit-Parmar:
@szczecha 
product_mutation
product_mutation_1

szczecha (approved):

maarcingebala:
@Smit-Parmar We found one more thing that needs to be added - currently, it's impossible to delete metadata from the Shop object. We have a generic API for metadata for that `deleteMetadata` as well as a few others (see `/graphql/meta/mutations/` module). It's impossible to use them since the shop object doesn't have the `id` field.

We thought we could add the `Shop.id` field that would allow this. In every other type, the `id` value is a Base64 decoded form of a string "<type>:<pk>", for example look at the following product ID:

```python
In [1]: import base64

In [2]: base64.b64decode("UHJvZHVjdDoxNTI=")
Out[2]: b'Product:152'
```
`Product` is the type and `152` is the database primary key of that object.

We could have the same for shop, but instead of using the database PK use a fixed number (since there is always one instance of shop) generated from the following string:

```python
In [5]: base64.b64encode(b"Shop:1")
Out[5]: b'U2hvcDox'
```

Could you look into adding this change? I think the steps to implement this are the following:
- add `relay.Node` interface to the `Shop` type
- add the `id: ID` field declaration
- add a resolver `resolve_id` to always return encoded `Shop:1`? 
- override `get_node` method in the `Shop` type to always return `Shop.objects.first()`

Smit-Parmar:
@maarcingebala Please review above changes. Shouldn't it be `SiteSettings.objects.first()` as we are using SiteSettings behind the Type `Shop`. Is it possible to return `graphene.Node.to_global_id("SiteSettings","1")` from `resolve_id`?
 I tried returning both `Shop:1` and `SiteSettings:1` but not working. So I return "1" it is working but might not work while adding test case in `meta/test/mutations`. I'm bit confused here.

saleor/graphql/shop/types.py
```diff
@@ -352,8 +353,17 @@ class Meta:
...
             "Represents a shop resource containing general shop data and configuration."
         )
-        interfaces = [ObjectWithMetadata]
-
+        interfaces = [graphene.relay.Node,ObjectWithMetadata]
+        model = site_models.SiteSettings
+    
+    @staticmethod
+    def resolve_id(_, _info):
+        return "1"
```
maarcingebala, line 361: It should return Base64 encoded string "Shop:1".

maarcingebala:
@Smit-Parmar Yes that's correct, I meant the `SiteSettings` model ;) Added a comment about the resolver.

Smit-Parmar:
> @Smit-Parmar Yes that's correct, I meant the `SiteSettings` model ;) Added a comment about the resolver.

@maarcingebala Please check comment

saleor/graphql/shop/types.py
maarcingebala: I think this should work:
```
@staticmethod
def resolve_id(_, _info):
    return base64.b64encode(b"Shop:1").decode()
```

Smit-Parmar: It's not working . Unable to delete.

Smit-Parmar: It's working with return "1". Returning string like `U2hvcDox`

maarcingebala: The resolver needs to return Base64 encoded version. To support deleting and other metadata mutations I think we also need to make changes to `BaseMetadataMutation`; not sure now which methods need changing but there is the entire logic for resolving objects for these mutations. Since we have a custom ID for the shop object, I think we need to adjust some of the methods in `BaseMetadataMutation` to properly translate from `U2hvcDox` to `SiteSettings.objects.first()`. For example this mutation should work and remove metadata from the Shop:

```
mutation {
  deleteMetadata(id: "U2hvcDox", keys: ["test"]) {
    errors {
      message
      field
    }
  }
}
```

Smit-Parmar: Above mutation is working .

maarcingebala:
@Smit-Parmar I think that for now, it's fine if we only support `shopSettingsUpdate` mutation. Can you remove the last commit https://github.com/saleor/saleor/pull/13128/commits/1b5710bf777c52d18be0eec95b9d100cd1d1ffd9 and we will be good to merge this PR?

Smit-Parmar:
@maarcingebala Reverted

saleor/graphql/shop/tests/queries/test_shop.py
```diff
@@ -830,3 +830,80 @@ def test_query_countries_filter_shiping_zones_attached_none(
 
     # then
     assert len(data) == len(countries)
+
+
+SHOP_METADATA_QUERY = """
```
IKarbowiak, line 835: Could you move does tests to `saleor/graphql/meta/tests/queries/` directory? You should create there a new `test_shop.py` file and move those tests there :)

IKarbowiak: And we should also add tests for mutations in `saleor/graphql/meta/tests/mutations/` directory.

saleor/graphql/meta/tests/mutations/test_shop.py
```diff
@@ -0,0 +1,46 @@
...
+                value
+            }
+            privateMetadata {
+                key
+                value
+            }
+            }
+        }
+    }
+"""
```
IKarbowiak, line 22: Thank you for moving the tests to the meta directory.

Could you right now make it consistent with other tests in `meta/tests/mutations`? For example take a look at this file: https://github.com/saleor/saleor/blob/main/saleor/graphql/meta/tests/mutations/test_gift_card.py. 

We need tests 4 tests, and please use the helper methods there:
- for updating public metadata (helper: `execute_update_public_metadata_for_item`)
- for updating private metadata (helper: `execute_clear_private_metadata_for_item`)
- for deleting public metadata (helper: `execute_clear_public_metadata_for_item`)
- for deleting private metadata (helper: `execute_clear_private_metadata_for_item`).

maarcingebala:
@Smit-Parmar Some linters are failing in CI due to unformatted code. I recommend installing [pre-commit](https://docs.saleor.io/docs/3.x/developer/community/contributing#coding-style), which runs the same linters as our CI before each commit.

IKarbowiak (approved):

Smit-Parmar:
> @Smit-Parmar Some linters are failing in CI due to unformatted code. I recommend installing [pre-commit](https://docs.saleor.io/docs/3.x/developer/community/contributing#coding-style), which runs the same linters as our CI before each commit.

@maarcingebala pre-commit is already setup on my machine but somehow it's not running automatic on commit .

mirekm:
Great first contributions, congrats, and thank you @Smit-Parmar!

Smit-Parmar:
> Great first contributions, congrats, and thank you @Smit-Parmar!

Thanks @mirekm . BTW it's my 4th or 5th PR

mirekm:
> > Great first contributions, congrats, and thank you @Smit-Parmar!
> 
> Thanks @mirekm . BTW it's my 4th or 5th PR

Right, I saw the others too, I meant contributions plural! 🙃

maarcingebala:
@Smit-Parmar FYI, I've added the following PR that supports the generic metadata API with the Shop object, which includes adding the `id` field you were initially working on: https://github.com/saleor/saleor/pull/13364.

Smit-Parmar:
Thanks @maarcingebala for sharing.

PR #13132, opened 2023-06-20
Added createdByEmail filter in GiftCard

### LINKED ISSUE #13098: Gift card filter by `createdByEmail`
## Problem

Customer support needs to find customers giftcards to be able to resend code, correct balance or inquiry about balance

Smit-Parmar:
I want to merge this change because It closes #13098 

# Impact

* [ ] New migrations

# Pull Request Checklist


saleor/graphql/giftcard/tests/queries/test_gift_card_filtering.py
```diff
@@ -495,3 +496,26 @@ def test_query_filter_gift_cards_by_metadata(
...
+
+    # when
+    response = staff_api_client.post_graphql(
+        query, variables, permissions=[permission_manage_gift_card]
+    )
+
+    # then
+    content = get_graphql_content(response)
+    data = content["data"]["giftCards"]["edges"]
+    assert data[0]["node"]["createdByEmail"] == "test@example.com"
```
maarcingebala, line 521: Maybe for the sake of this test, we should first update one of the gift cards with some unique email address (in the "given" section) and then expect only one result in the response.

maarcingebala (commented):
Please also extend the changelog with an information about added field.

saleor/graphql/giftcard/tests/queries/test_gift_card_filtering.py
Smit-Parmar: @maarcingebala  Even after updating gift_cards getting assert error 
```
gift_card.created_by_email = "random@example.com"
gift_card.save()
```
Because gift card is returning `test@example.com` which is customer email.

maarcingebala: Maybe we should then reassign the user instance that has the email updated.

Smit-Parmar:
@maarcingebala Just added new `gift_card2` fixtures with different email as updating `gift_card` with different user in "given" was still returning "test@example.com" .

saleor/tests/fixtures.py
```diff
@@ -3972,6 +3972,20 @@ def gift_card(customer_user):
     return gift_card
 
 
+@pytest.fixture
+def gift_card2(customer_user2):
```
maarcingebala, line 3976: That's not an obvious name for a fixture; I would not add any new ones but try to modify existing ones in the test. You can debug the resolver for that field to see why it's returning the other email (see the `resolve_created_by_email` function).

Smit-Parmar: It's happening because first it returning user.email instead of instance.created_by_email and that user.email is customer_user.email (`test@example.com`)from fixtures. 
```
return user.email if user else root.created_by_email
```
I also tried
```
    query = QUERY_GIFT_CARDS
    gift_card.user = customer_user2
    gift_card.created_by_email = customer_user2.email
    gift_card.save()
    variables = {"filter": {"createdByEmail": gift_card2.created_by_email}}
```
but it was pointing to customer_user instead of customer_user2.

maarcingebala (approved):

saleor/tests/fixtures.py
```diff
@@ -3959,11 +3959,11 @@ def gift_card_tag_list(db):
...
 
 @pytest.fixture
-def gift_card(customer_user):
+def gift_card(customer_user2):
     gift_card = GiftCard.objects.create(
         code="never_expiry",
-        created_by=customer_user,
-        created_by_email=customer_user.email,
+        created_by=customer_user2,
+        created_by_email=customer_user2.email,
```
maarcingebala, line 3966: I guess it would be better to change these fields only in the test, but this is OK with me too, we'll see what other reviewers think ;)

8r2y5 (approved):

8r2y5:
Overall looks good so I approved it but please check this comment https://github.com/saleor/saleor/pull/13132#discussion_r1236550085

saleor/tests/fixtures.py
Smit-Parmar: I tried that to change fields only in particular test case but it' not working it was still returning `customer_user1` email instead of `customer_user2`. That's why I changed globally.

maarcingebala: @Smit-Parmar Did you debug it in the resolver? There is some logic around that; maybe some fields were left unchanged and it was returning the wrong value. Usually, we try to keep fixtures more general and adjust them inside the tests.

Smit-Parmar: @maarcingebala I checked the resolver [https://github.com/saleor/saleor/pull/13132#discussion_r1235171805](cause). I'm looking into this and see what can be done.

Smit-Parmar:
@maarcingebala I revert the fixtures file and re run the test cases it's working now. I think no changes are required.

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ All notable, unreleased changes to this project will be documented in this file.
 
 ### Other changes
 - Expand metric units to support more types of products. - #13043 by @FremahA
+- Fix Error Cannot return null for non-nullable field Webhook.name. - #12989 by @Smit-Parmar
+- Added createdByEmail filter in GiftCard and Test case. - #13132 by @Smit-Parmar
+
```
maarcingebala, line 22: @Smit-Parmar I have slightly redacted the changelog entries for consistency. Also we can move both of them to "GraphQL API" section - they're both strictly related to the API.

```suggestion
- Fix error "Cannot return null for non-nullable field Webhook.name" - #12989 by @Smit-Parmar
- Added `GiftCardFilterInput.createdByEmail` filter - #13132 by @Smit-Parmar
```

maarcingebala (approved):

szczecha (approved):

Issue #13141, opened 2023-06-20
Drop unused packages


maarcingebala:
Review packages and try to identify unused ones and remove them. Some examples:
- [x] `google-measurement-protocol` for Google Analytics - it should be removed from Saleor as it's been deprecated for a long time
- [x] `django-versatileimagefield` - it's not used in logic anymore, but it's still referred to in migrations. We should look into how it can be removed.
- [x] `django-prices-openexchangerates`
  - it's not usable in cloud so we can drop the package
  - drop the `to_local_currency` function
  - mark all `*localCurrency` fields as deprecated and make them always return `null` (they're nullable already)
- [x] `django-prices-vatlayer`
  - drop usages of `VAT` models 
  - make `vat` fields in API that rely on VAT model return `null` 
  - `0005_migrate_vatlayer.py` migration uses the `VAT` model to migrate Vatlayer's taxes to flat rates; we could wrap the [model import](https://github.com/saleor/saleor/blob/b1b673819007af0f9fee83b60723ec52495661f8/saleor/tax/migrations/0005_migrate_vatlayer.py#L66) with `try..except` and in case the package is not installed provide an empty list as the [`vat_rates` queryset](https://github.com/saleor/saleor/blob/b1b673819007af0f9fee83b60723ec52495661f8/saleor/tax/migrations/0005_migrate_vatlayer.py#L69).

maarcingebala:
Added `django-prices-openexchangerates` to the list.

maarcingebala:
Added `django-prices-vatlayer`.

PR #13146, opened 2023-06-21
Remove unused google-measurement-protocol package

SzymJ:
I want to merge this change because it removes unused `google-measurement-protocol package`.

Part of #13141

# Impact

* [ ] New migrations

# Pull Request Checklist


maarcingebala (commented):
There is also `OrderBulkCreateInput.trackingClientId` that we should deprecate as well. But I think we could even remove it since it's part of the order bulk API, which is a preview feature. We would have to mention it as a breaking change.

maarcingebala (approved):

Issue #13143, opened 2023-06-20
Drop TranslationProxy


maarcingebala:
Drop the `TranslationProxy`, which is no longer useful as we don't rely on localization middleware. Any references to the `translated` property should be fixed to return proper translation in the current context. For checkout and order usages we should rely on `language_code` that can be set by mutations to provide the locale.

PR #13156, opened 2023-06-21
Drop translation proxy

zedzior:
Issue: https://github.com/saleor/saleor/issues/13143

I want to merge this change, because TranslationProxy is no longer useful as we don't rely on localisation middleware. Any references to the translated property should be fixed to return proper translation in the current context, ie. for checkout and order usages we should rely on language_code that can be set by mutations to provide the locale.

Essentially, this PR replaces `translated` property with `get_translation(language_code)` function and remove `translated` property from models, which do not utilise it at all.

# Impact

* [ ] New migrations

# Pull Request Checklist


IKarbowiak (approved):

kadewu (approved):

saleor/core/utils/translations.py
```diff
@@ -51,3 +45,9 @@ def get_translated_keys(self) -> Dict[str, Any]:
 
     def get_translation_context(self) -> Dict[str, Any]:
         return {}
+
+
+def get_translation(instance, language_code) -> TranslationWrapper:
+    if not language_code:
+        language_code = settings.LANGUAGE_CODE
```
maarcingebala, line 52: I'm wondering if this is needed. IMO we should set translations only when explicitly requested, in this case, when `language_code` is known, so when it was provided in checkout mutations. In the cloud, the language code is always `en`. I think we could return `None` when language code is not provided.

maarcingebala (approved):

PR #13159, opened 2023-06-22
Unify country fetching

### LINKED ISSUE #13140: Unify fetching addresses for checkouts and orders
Review the logic for patterns used to get users' address or their parts in the logic and possibly unify them to follow the same rules. One example is `get_tax_country`, which follows this logic: 
- get the country from the shipping address if it exists and shipping is required
- get the country from the billing address if it exists and shipping is not required
- if none of the above returned a value, fallback to channel's default country

If possible everywhere we should use the same logic, or at least have some unified functions to reuse in different cases.

jakubkuc:
I want to merge this change because it unifies how active country is fetched for orders and checkouts. 
Country code is fetched in the following order:
1. shipping address
2. billing address
3. default country from channel

fixes #13140

the function used to unify this can also be used with address data from mutation input by passing address_data in parameters.

# Impact

* [ ] New migrations

# Pull Request Checklist


maarcingebala (approved):

IKarbowiak (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ All notable, unreleased changes to this project will be documented in this file.
     - `ProductPricingInfo.priceRangeLocalCurrency`
     - `VariantPricingInfo.discountLocalCurrency`
     - `VariantPricingInfo.priceLocalCurrency`
+- Unify country fetching - #13159 by @jakubkuc
+    - Note: this change allows to add `shippingAddress` to checkout or order that has no shipping required
```
maarcingebala, line 18: If this is a breaking change we need to be more explicit how it affect our users.

Issue #12317, opened 2023-03-16
Implement missing events for user/account operations


lkostrowski:
# What I'm trying to achieve

Re-implement user-plugin to an App. To achieve that, missing events must be implemented

# Describe a proposed solution

Proposed APIs

## New `ACCOUNT_CONFIRMATION_REQUESTED`

Triggered when [customerCreate](https://docs.saleor.io/docs/3.x/api-reference/mutations/customer-create) is invoked

### Payload

```graphql
subscription {
  event {
    ... on AccountConfirmation {
      token: String!,
      user: User!
      shop: Shop!
      redirectUrl: String!
    }
  }
}
```

## New `ACCOUNT_PASSWORD_RESET_REQUESTED`

Triggen when [requestPasswordReset](https://docs.saleor.io/docs/3.x/api-reference/mutations/request-password-reset)

### Payload

```graphql
subscription {
  event {
    ... on AccountPasswordReset {
      token: String!,
      user: User!
      redirectUrl: String!
      channel: Channel!
      shop: Shop!
    }
  }
}
```

Storefront will use [setPassword mutation](https://docs.saleor.io/docs/3.x/api-reference/mutations/set-password)

# New `ACCOUNT_CHANGE_EMAIL_REQUESTED`
[rel old event](https://docs.saleor.io/docs/3.x/developer/available-plugins/user-emails#customer-requested-email-change)

Triggered after mutation [requestEmailChange](https://docs.saleor.io/docs/3.x/api-reference/mutations/request-email-change)

### Payload

```graphql
subscription {
  event {
    ... on AccountChangeEmailRequest {
      token: String!,
      user: User!
      redirectUrl: String!
      channel: Channel!
      shop: Shop!
      newEmail: String!
    }
  }
}
```

Storefront will call https://docs.saleor.io/docs/3.x/api-reference/mutations/confirm-email-change

## New `ACCOUNT_CHANGE_EMAIL_CONFIRMED`

Triggered after [confirmEmailChange mutation](https://docs.saleor.io/docs/3.x/api-reference/mutations/confirm-email-change)

### Payload

```graphql
subscription {
  event {
    ... on AccountChangeEmailConfirmation {
      user: User!
      channel: Channel!
      shop: Shop!
    }
  }
}
```

## New `ACCOUNT_DELETE_REQUESTED`

Triggered after [accountRequestDeletion](https://docs.saleor.io/docs/3.x/api-reference/mutations/account-request-deletion)

### Payload

```graphql
[...]

PR #13162, opened 2023-06-22
Add ACCOUNT_CONFIRMATION_REQUESTED event

SzymJ:
I want to merge this change because it adds `ACCOUNT_CONFIRMATION_REQUESTED` event.

Part of #12317 

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/schema.graphql
+++ b/saleor/graphql/schema.graphql
@@ -27451,13 +27451,13 @@ type AccountConfirmationRequested implements Event @doc(category: "Users") {
   """The URL to redirect the user after he accepts the request."""
   redirectUrl: String
 
-  """The user assigned to the webhook."""
+  """The user the event relates to."""
   user: User
 
-  """The channel assigned to the webhook."""
+  """The channel data."""
   channel: Channel
 
-  """The token required to confirm the user."""
+  """The token required to confirm request."""
   token: String
 
   """Shop data."""
--- a/saleor/graphql/schema.graphql
+++ b/saleor/graphql/schema.graphql
@@ -1687,6 +1687,7 @@ type WebhookEvent @doc(category: "Webhooks") {
 enum WebhookEventTypeEnum @doc(category: "Webhooks") {
   """All the events."""
   ANY_EVENTS
+  ACCOUNT_CONFIRMATION_REQUESTED
 
   """A new address created."""
   ADDRESS_CREATED
@@ -2318,6 +2319,7 @@ type WebhookEventAsync @doc(category: "Webhooks") {
 enum WebhookEventTypeAsyncEnum @doc(category: "Webhooks") {
   """All the events."""
   ANY_EVENTS
+  ACCOUNT_CONFIRMATION_REQUESTED
 
   """A new address created."""
   ADDRESS_CREATED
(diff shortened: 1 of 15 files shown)


maarcingebala (approved):

IKarbowiak (approved):

patrys:
Since this is a new event, here are my thoughts:

1. We should always send the confirmation requested event.
2. The event should contain information if confirmation is mandatory or not.
3. Even if confirmation is not required to log in, it should be required to claim previously placed orders (and orders using the same email address placed without logging in)

SzymJ:
> Since this is a new event, here are my thoughts:
> 
> 1. We should always send the confirmation requested event.
> 2. The event should contain information if confirmation is mandatory or not.
> 3. Even if confirmation is not required to log in, it should be required to claim previously placed orders (and orders using the same email address placed without logging in)

@patrys so:
1. should we send an event without checking `enable_account_confirmation_by_email` flag?
2. If point `1` is true then we can get this flag in the payload from `Shop` object.
3. Do you mean to add a task for matching orders with an email to this mutation?

patrys:
1. Yes, I believe we should always generate allow email confirmation, whether it's required or not.
2. Great!
3. The mutation that confirms your account should also claim any orders placed using the same email, not the one sending the token.

SzymJ:
@patrys it looks like `ConfirmAccount` already run matching orders with email

saleor/graphql/account/mutations/account.py
```diff
@@ -162,12 +164,26 @@ def save(cls, info: ResolveInfo, user, cleaned_input):
...
+                params = urlencode({"email": user.email, "token": token})
+                confirm_url = prepare_url(params, cleaned_input["redirect_url"])
+
+                cls.call_event(
+                    manager.account_confirmation_requested,
+                    user,
+                    cleaned_input["channel"],
+                    token,
+                    confirm_url,
+                )
```
patrys, line 186: It's still only sent if account confirmation is required.

SzymJ: Done, additionally, as @fowczarek mentioned in #13170 I make notify and webhook event to use the same token.

patrys (approved):

saleor/graphql/schema.graphql
```diff
@@ -1687,6 +1687,7 @@ type WebhookEvent @doc(category: "Webhooks") {
 enum WebhookEventTypeEnum @doc(category: "Webhooks") {
   """All the events."""
   ANY_EVENTS
+  ACCOUNT_CONFIRMATION_REQUESTED
```
patrys, line 1690: Description missing!

saleor/graphql/schema.graphql
```diff
@@ -2318,6 +2319,7 @@ type WebhookEventAsync @doc(category: "Webhooks") {
 enum WebhookEventTypeAsyncEnum @doc(category: "Webhooks") {
   """All the events."""
   ANY_EVENTS
+  ACCOUNT_CONFIRMATION_REQUESTED
```
patrys, line 2322: Description missing!

saleor/graphql/schema.graphql
```diff
@@ -27426,6 +27429,40 @@ enum VolumeUnitsEnum {
   ACRE_FT
 }
 
+"""
+Event sent when account confirmation requested.
+
+Added in Saleor 3.15.
+"""
+type AccountConfirmationRequested implements Event @doc(category: "Users") {
```
patrys, line 27437: It does not include information if the confirmation is required to log in or optional.

patrys, line 27437: It does not include information if the confirmation is required to log in or optional.

szczecha:
Should we add a description of the event? 
Screenshot 2023-06-26 at 11 34 20

yellowee (approved):

Issue #13082, opened 2023-06-15
Search query for giftcards


timuric:
## Problem

Customer support needs to able to find gift cards by the name or email of customers. 

`usedBy` filter field doesn't help as it only accepts IDs of existing accounts, which doesn't cover anonymous users.

Another pain point is that admins want to use generic search that accepts any input instead of filters, the same way as product search works
```[tasklist]
### Tasks
```

maarcingebala:
- `GiftCard` model changes:
  - Add `search_index_dirty = models.BooleanField(default=False, db_index=True)` field
  - Add `search_vector = SearchVectorField(blank=True, null=True)` field.
    - Fields to include in the search vector:
      - `GiftCard.code`
      - `GiftCard.used_by_email`
      - `GiftCard.created_by_email`
      - `GiftCard.used_by` - user's email, first name and last name
      - `GiftCard.createdBy` - user's email, first name and last name
    - See `prepare_product_search_vector_value` for a reference
- Update logic to mark `search_index_dirty=True` whenever parts of the search vector change.
- Add a Celery task that looks for dirty search indexes and updates the search vector value. The task should update a batch of gift cards in a single run and check if there are still gift cards to update and schedule itself in the queue to process the remaining ones in separate tasks.

PR #13173, opened 2023-06-23
Add `search` to giftcard query

zedzior:
Issue: https://github.com/saleor/saleor/issues/13082
I want to merge this change, because I want to add full text search to `giftCards` query.

# Impact

* [x] New migrations

# Pull Request Checklist


saleor/giftcard/tasks.py
```diff
@@ -4,8 +4,10 @@
 from ..celeryconf import app
 from .events import gift_cards_deactivated_event
 from .models import GiftCard
+from .search import update_gift_cards_search_vector
 
 task_logger = get_task_logger(__name__)
+GIFT_CARD_BATCH_SIZE = 10
```
korycins, line 10: Is it expected that the batch is so small? Why?

saleor/graphql/giftcard/schema.py
```diff
@@ -49,6 +50,9 @@ class GiftCardQueries(graphene.ObjectType):
         filter=GiftCardFilterInput(
             description=("Filtering options for gift cards." + ADDED_IN_31)
         ),
+        search=graphene.String(
+            description="Look up gift cards by multiple fields." + ADDED_IN_315
```
korycins, line 54: Is it a preview feature or we are sure of the current way of working?

saleor/giftcard/migrations/0019_auto_20230623_0857.py
```diff
@@ -0,0 +1,35 @@
...
+            field=models.BooleanField(db_index=True, default=True),
+        ),
+        migrations.AddField(
+            model_name="giftcard",
+            name="search_vector",
+            field=django.contrib.postgres.search.SearchVectorField(
+                blank=True, null=True
+            ),
+        ),
+        AddIndexConcurrently(
```
kadewu, line 27: I believe there is something wrong here. AddIndexConcurrently needs `atomic=False`, but I can't see it. If the migrations is running correctly then probably we need to remove statement from line 18 `db_index=True`

saleor/settings.py
```diff
@@ -602,6 +602,11 @@ def get_host():
         "schedule": timedelta(seconds=20),
         "options": {"expires": BEAT_UPDATE_SEARCH_EXPIRE_AFTER_SEC},
     },
+    "update-gift-cards-search-vectors": {
```
kadewu, line 605: Maybe we should ask SRE team if they are fine with that?

saleor/giftcard/tasks.py
zedzior: My bad, it was for testing purpose

saleor/graphql/giftcard/schema.py
```diff
@@ -49,6 +50,11 @@ class GiftCardQueries(graphene.ObjectType):
         filter=GiftCardFilterInput(
             description=("Filtering options for gift cards." + ADDED_IN_31)
         ),
+        search=graphene.String(
+            description="Look up gift cards by multiple fields."
```
maarcingebala, line 54: Let's make it more explicit - "Search gift cards by email and name of user who created or used the gift card" - maybe something like that.

maarcingebala (commented):
We're missing marking the search index dirty whenever the parts of the vector depending on relationships are changed: email, first and last name of `used_by` and `created_by` users. We should also look into mutations that change these fields for user models, check if gift cards are assigned to the user, and mark the search index as dirty in such case.

saleor/graphql/account/mutations/base.py
```diff
@@ -338,6 +340,10 @@ def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
             manager = get_plugin_manager_promise(info.context).get()
             cls.call_event(manager.customer_metadata_updated, instance)
 
+        if user_gift_cards := get_user_gift_cards(instance):
+            if cleaned_input.get("first_name") or cleaned_input.get("last_name"):
```
maarcingebala, line 344: It would be better to reverse these conditions to first check inputs and then `get_user_gift_cards`, since the second hits the database, which won't have to be necessary if the first/last name is not changed. We can save some queries this way.

zedzior: good point

maarcingebala (approved):

korycins (approved):

kadewu (approved):

szczecha:
I've two gift cards:
first created in the dashboard and then used by customer in checkout 
second user bought product gift card via checkout
```
{
  "data": {
    "giftCards": {
      "totalCount": 2,
      "edges": [
        {
          "node": {
            "id": "R2lmdENhcmQ6Mg==",
            "displayCode": "341A",
            "created": "2023-06-30T09:47:09.971152+00:00",
            "createdByEmail": "testers+dashboard@saleor.io",
            "createdBy": {
              "__typename": "User",
              "email": "testers+dashboard@saleor.io"
            },
            "usedByEmail": "customer+1@example.com",
            "boughtInChannel": null,
            "lastUsedOn": "2023-06-30T10:12:35.269088+00:00"
          }
        },
        {
          "node": {
            "id": "R2lmdENhcmQ6Mw==",
            "displayCode": "1F52",
            "created": "2023-06-30T10:31:03.375165+00:00",
            "createdByEmail": "customer+1@example.com",
            "createdBy": null,
            "usedByEmail": null,
            "boughtInChannel": "default-channel",
            "lastUsedOn": null
          }
        }
      ]
    }
  }
}
```


I tried to use search and find gift cards by code or email but got no results 
Screenshot 2023-06-30 at 12 43 30

Screenshot 2023-06-30 at 12 44 25

Screenshot 2023-06-30 at 12 44 46

maarcingebala:
Hmm maybe it's a matter of this test env- I think it doesn't have Cron that would reindex search vectors. Not sure how we tested that with other searches. But maybe it's something wrong with the code as well.

szczecha:
> Hmm maybe it's a matter of this test env- I think it doesn't have Cron that would reindex search vectors. Not sure how we tested that with other searches. But maybe it's something wrong with the code as well.

ok, I will check it on local env

szczecha (approved):

PR #13244, opened 2023-06-30
Refactor E2E tests

fowczarek:
I want to merge this change because of refactoring E2E tests.

TODO:
- [x] Reuse default address in warehouse create util
- [x] Drop unused parameter from E2eApiClient
- [x] Add assert for chekcoutId in created order via CheckoutCreate
- [x] Add a short description to steps
- [x] Split PyTest workflow to separate workflow for E2E tests

# Impact

* [ ] New migrations

# Pull Request Checklist


maarcingebala (approved):

.github/workflows/pytest.yml
```diff
@@ -99,10 +108,7 @@ jobs:
...
-      # Run linters and Django related checks
-      # `git config` command is a workaround for https://github.com/actions/runner-images/issues/6775
-      - name: Run Linters and Checks
+      - name: Run E2E tests
         run: |
-          git config --global --add safe.directory "$GITHUB_WORKSPACE"
-          pre-commit run --all
-        if: ${{ always() }}
+          pytest \
+            -m "e2e"
```
NyanKiyoshi, line 114: Is it intended that it is excluded from coverage?

IKarbowiak (approved):

korycins (approved):

.github/workflows/pytest.yml
fowczarek: Yes. In my opinion, we shouldn't sum coverage from the E2E test and unit tests. 
If we want to have coverage on the E2E test we should ad it as a separate metric.

PR #9806, opened 2022-05-22
Fix memory consumption of delete_event_payloads_task

przlada:
I want to merge this change because it optimizes memory consumption of `delete_event_payloads_task` by switching to raw SQL queries to delete event data.

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #13246, opened 2023-06-30
Delete event deliveries in batches

### LINKED ISSUE #13235: Handle event deliveries delete in batches
Event deliveries, event delivery attempts and payloads are deleted in a `delete_event_payloads_task` without using batches.
This should be done in batches as it can lead to timeouts on db in case of large number of event delivery fails in two weeks period.

jakubkuc:
I want to merge this change because it adds batches to event deliveries delete task. Fixes #13235
I'm switching back to using django `delete()` instead od raw SQL query (which was introduced in #9806 ) and use the logic that is used in zero-downtime migrations to make the task more optimal. 
It will now delete all the `EventDelivery` objects and by `CASCADE` all `EventDeliveryAttempt`s (max 5 attempts per delivery) and when it is done it will take care of unused `EventPayload`s

# Impact

* [ ] New migrations

# Pull Request Checklist


saleor/core/tasks.py
```diff
@@ -16,18 +17,16 @@ def delete_from_storage_task(path):
...
-    attempts._raw_delete(attempts.db)
-    deliveries._raw_delete(deliveries.db)
-    payloads._raw_delete(payloads.db)
+    ids = deliveries.values_list("pk", flat=True)[:BATCH_SIZE]
+    qs = EventDelivery.objects.filter(pk__in=ids)
+    if ids:
+        qs.delete()
+        delete_event_payloads_task.delay()
+    else:
+        delete_unused_payloads.delay()
```
korycins, line 29: Do we really need additional task here? 
It looks like it has the same body as `delete_event_payloads_task`, with an additional filter argument. Can't we make a single filter query with `Q()`? 
Also, I think, that there is a mistake in the flow. In the case of calling  `delete_event_payloads_task` -> `delete_unused_payloads`, `delete_unused_payloads` will not re-trigger the task. Which means that we will just delete 100 event payloads no matter how many we have them in DB.

fowczarek (approved):

maarcingebala (approved):

maarcingebala:
@jakubkuc Linters are failing.

IKarbowiak (approved):

saleor/core/tasks.py
```diff
@@ -16,18 +18,15 @@ def delete_from_storage_task(path):
...
-    payloads._raw_delete(payloads.db)
+    valid_deliveries = EventDelivery.objects.filter(created_at__gt=delete_period)
+    payloads_to_delete = EventPayload.objects.filter(
+        ~Exists(valid_deliveries.filter(payload_id=OuterRef("id")))
+    ).order_by('-pk')
+    ids = payloads_to_delete.values_list("pk", flat=True)[:BATCH_SIZE]
+    qs = EventPayload.objects.filter(pk__in=ids)
+    if ids:
+        qs.delete()
+        delete_event_payloads_task.delay()
```
NyanKiyoshi, line 29: Give up if it takes too long, e.g. if the first task started 1h ago, then give up (1h is a random value out of my head).

We don't want to end-up in a case where we have concurrent delete tasks against the same rows, we will end-up with a lock queue. On top of not wanting to run forever a single task.

jakubkuc:
This PR is closed due to moving it to #13368

Issue #13073, opened 2023-06-14
[RFC] Add a flag to channel to allow placing unpaid orders from checkoutComplete


kadewu:
## Problem

https://github.com/saleor/saleor/issues/13016
We want to have possibility to complete checkout with order before successful payment.

## 1. General Assumptions

- New `Channel` flag `allow_unpaid_orders`.
- When flag `allow_unpaid_orders` on channel is `False` flow will not be affected.
- `allow_unpaid_orders` will result in creating order in TRANSACTION_FLOW
- Changes affects only `checkoutComplete` mutation

### Acceptance Criteria
- [ ] Add new boolean flag to `Channel`: `allow_unpaid_orders`
- [ ] User with `MANAGE_CHANNELS` or `MANAGE_ORDERS`  can update flag `allow_unpaid_orders`.
- [ ] When `allow_unpaid_orders` is disabled, completeCheckout requires successful payment to process.
- [ ] When `allow_unpaid_orders` is enabled, user can create order from checkout even if the payment isn't successful.

### Types

```graphql
input: OrderSettingsInput {
  ...
  """Allows to create order before payment"""
  allowUnpaidOrders: Boolean!

```

type OrderSettings{
  ...
  allowUnpaidOrders: Boolean!
}

### Other

Update logic regardless processing the payment and creating the order from checkout. 

## 3. Database changes

New boolean field in `Channel` model: allow_unpaid_orders with default value `False`

## 5. To Do list

```[tasklist]
### Tasks
- [ ] Sync with checkout/dashboard teams
- [ ] Update docs - new subchapter with new flow for finalizing checkout
- [ ] Document order expiration settigs
```

korycins:
I would suggest different name for `checkout_complete_create_order_first`. It should be something like: `allow-to-create-order-without-payment` or something with `checkout-complete` name but with something that suggests that we will allow you to create order without a processed payment.

---
Also worth mentioning, the flag will have an impact only on the transactionAPI. It will not change the behaviour of the old PaymentAPI

---
We should also update the docs  for finalizing the checkout process, where we place the subchapter that will describe the flow where we create the order without payment. The chapter should also mention, that at the beginning we lock the prices, stocks, etc for the order, then based on the settings for expireOrder, we mark them as expired and release the stocks (if not have any payment), and delete it after X days.

korycins:
I believe that the new flag should be able to control by staff user with MANAGE_ORDERS

maarcingebala:
[...]

PR #13339, opened 2023-07-07
Allow creating order without payment

kadewu:
Fix https://github.com/saleor/saleor/issues/13073

# Impact

* [ ] New migrations

# Pull Request Checklist


saleor/checkout/complete_checkout.py
```diff
@@ -665,7 +665,13 @@ def _prepare_checkout_with_transactions(
 ):
     """Prepare checkout object with transactions to complete the checkout process."""
     clean_billing_address(checkout_info, CheckoutErrorCode)
-    if checkout_info.checkout.authorize_status != CheckoutAuthorizeStatus.FULL:
+    if (
+        checkout_info.checkout.authorize_status != CheckoutAuthorizeStatus.FULL
+        and not (
```
korycins, line 670: We shouldn't have a condition like `checkout_info.checkout.authorize_status == CheckoutAuthorizeStatus.NONE` here. 
You may want to have an order with two payments, first payment could be processed before creating the order, next one could be added later.
Similar case, when you require to cover 30% of the order total  for pre-orders, etc. 
I am saying that if `allow_to_create_order_without_payment` is set to True, we shouldn't validate any payment's satuses

saleor/checkout/complete_checkout.py
```diff
@@ -1299,10 +1305,18 @@ def complete_checkout(
     # paid is used. In case when we have TRANSACTION_FLOW we use transaction flow to
     # finalize the checkout.
     checkout_is_zero = checkout_info.checkout.total.gross.amount == Decimal(0)
-    if transactions or (
-        checkout_is_zero
-        and checkout_info.channel.order_mark_as_paid_strategy
-        == MarkAsPaidStrategy.TRANSACTION_FLOW
+    if (
```
korycins, line 1308: I think that we should have here something like this:
```python
    is_transaction_flow = (
            checkout_info.channel.order_mark_as_paid_strategy
            == MarkAsPaidStrategy.TRANSACTION_FLOW
    )
    use_transaction_flow = (
        # handle with transaction if we have assigned transaction to the checkout
        transactions or
        # handle with transaction if we allow to place the unpaid order from checkout
        checkout_info.channel.allow_to_create_order_without_payment or
        # handle with transaction if we have zero total and we use transaction flow
        # to mark order as paid
        (checkout_is_zero and is_transaction_flow)
    )
    if handle_with_transaction:
        order = complete_checkout_with_transaction(
```
I don't think that we should combine `checkout_info.channel.allow_to_create_order_without_payment` and `checkout_info.channel.order_mark_as_paid_strategy` into `and` condition. It would mean, that you would have to set up 2 different flag to be able to create unpaid order.

saleor/graphql/channel/mutations/channel_create.py
```diff
@@ -96,6 +97,13 @@ class OrderSettingsInput(BaseInputObjectType):
             "requested action for the transaction." + ADDED_IN_313 + PREVIEW_FEATURE
         ),
     )
+    allow_to_create_order_without_payment = graphene.Boolean(
+        required=False,
+        description=(
+            "Determine if it is possible to create order before finalizing payment "
```
korycins, line 103: - I would mention here that we are describing here usage of `checkoutComplete` mutation (We also have orderCreateFromCheckout).  "Determine if it is possible to place unpaid order by calling `checkoutComplete` mutation"
- `TRANSACTION_FLOW` doesn't say anything. It is a value of unrelated ENUM `MarkAsPaidStrategyEnum`.  I think that we can skip this part.  Creating unpaid order is not strictly related with the payment flow that you're using. For transactionAPI you can process payment later. For paymentAPI you can mark payment as paid.

saleor/graphql/channel/mutations/channel_create.py
```diff
@@ -221,6 +229,13 @@ def clean_input(cls, info: ResolveInfo, instance, data, **kwargs):
                     "default_transaction_flow_strategy"
                 ] = default_transaction_strategy
 
+            if allow_to_create_order_without_payment := order_settings.get(
```
korycins, line 232: What if `allow_to_create_order_without_payment` will be `False`?
It looks like we are not able to set up the `False` value. It means that we're missing unit tests.

saleor/graphql/channel/mutations/channel_update.py
```diff
@@ -140,6 +140,14 @@ def clean_input(cls, info: ResolveInfo, instance, data, **kwargs):
                 cleaned_input[
                     "default_transaction_flow_strategy"
                 ] = default_transaction_strategy
+
+            if allow_to_create_order_without_payment := order_settings.get(
```
korycins, line 144: Same as above

saleor/graphql/channel/types.py
```diff
@@ -221,6 +222,13 @@ class OrderSettings(ObjectType):
             + PREVIEW_FEATURE
         ),
     )
+    allow_to_create_order_without_payment = graphene.Boolean(
+        required=True,
+        description=(
+            "Determine if it is possible to create order before finalizing payment "
```
korycins, line 228: We can unify it to the description that we will add to mutation input

saleor/channel/models.py
```diff
@@ -32,6 +32,7 @@ class Channel(models.Model):
     )
 
     automatically_confirm_all_new_orders = models.BooleanField(default=True, null=True)
+    allow_to_create_order_without_payment = models.BooleanField(default=False)
```
korycins, line 35: I think, that we could figure out better name here & for Type field. 
maybe something like: 
a) `allow_to_create_unpaid_order_from_checkout`, 
b) `allow_to_create_unpaid_order_from_checkout_complete`
IMO `a` is OK as it explicitly says about creating the order from checkout. but still, it could be misleading, as we also have orderCreateFromCheckout mutation. In that case `b` sounds better.
WDYT? @kadewu  @maarcingebala

maarcingebala: I guess we could name it `allow_unpaid_orders` or something simple and just mention in the description and the docs in what mutations/cases it applies. What if we changed this behavior in the future? Would we have to change the name of this field then?

saleor/checkout/complete_checkout.py
IKarbowiak, line 1: I see only changes for `checkout_create`, shouldn't we also handle it for `create_order_from_checkout`?

korycins: Yes, `create_order_from_checkout` is a protected mutation that allows you to create unpaid orders, no matter what is the value of `allow_unpaid_orders`

korycins (approved):

saleor/channel/models.py
korycins: > What if we changed this behavior in the future? Would we have to change the name of this field then?

`allow_to_create_unpaid_order_from_checkout` is a field dedicated exactly to single action, if we will want to change the behavior or extend it in the future, we could add a new field and/or deprecate the previous one.

jakubkuc (approved):

szczecha (approved):

Issue #13248, opened 2023-06-30
Add metadata to checkoutShippingAddressUpdate & checkoutBillingAddressUpdate


timuric:
It would be convenient to pass extra metadata upon adding shipping & billing addresses for the checkout using `checkoutShippingAddressUpdate` & `checkoutBillingAddressUpdate` this would allow us to store non-standard address fields without calling separate `metadataUpdate` mutation

maarcingebala:
Proposed schema:
```graphql
checkoutShippingAddressUpdate(
  id: ID
  metadata: [MetadataInput!]
  privateMetadata: [MetadataInput!]
  shippingAddress: AddressInput!
  validationRules: CheckoutAddressValidationRules
): CheckoutShippingAddressUpdate
```

For readability, I've skipped the deprecated input fields in the example. `metadata` and `privateMetadata` fields would need a description that makes it clear that this metadata would be set for the address, not the checkout. `checkoutBillingAddressUpdate` would follow the same schema changes.

Question - addresses can be added to the checkout also in the `checkoutCreate` mutation which doesn't support setting any metadata. We we want also to provide support for setting metadata there? It's possible, but we would have to differentiate checkout metadata and addresses' metadata.

korycins:
@maarcingebala Have you considered adding metadata and privateMetada to the `AddressInput` type? This would solve the issue:
```
[we] would need a description that makes it clear that this metadata would be set for the address, not the checkout.
```
We would also have a nice support for checkoutCreate. Also what about permissions for retrieving the privateMetadata? Who can access to it?

maarcingebala:
Yes, that would be a much cleaner API, except these fields would have no effect in some queries where AddressInput is used, but I think that's fine. We already have permission rules for address metadata: https://github.com/saleor/saleor/blob/main/saleor/graphql/meta/permissions.py#L121

maarcingebala:
So the schema would look like this:
```graphql
input AddressInput {
  firstName: String
  lastName: String
  companyName: String
  streetAddress1: String
  streetAddress2: String
  city: String
  cityArea: String
  postalCode: String
  country: CountryCode
  countryArea: String
  phone: String

  # new fields
  metadata: [MetadataInput!]
  privateMetadata: [MetadataInput!]
}
```

Air-t:
@timuric For now using `updatePrivateMetadata` mutation for `Address` we are checking if updated address:

1. belongs to staff member then `AccountPermissions.MANAGE_STAFF`
[...]

PR #13361, opened 2023-07-10
Add `metadata` and `privateMetadata` to `AddressInput` (#13248)

Air-t:
Add ability to send extra metadata on `checkoutShippingAddressUpdate` mutation.
Extend `AddressInput` to pass `metadata` and `privateMetadata` if needed without a separate call to `updatePrivateMetadata` and `updateMetadata` mutations.

[(#13248)](https://github.com/saleor/saleor/issues/13248)

# Impact

* [ ] New migrations

# Pull Request Checklist


saleor/checkout/models.py
```diff
@@ -250,6 +247,30 @@ def get_country(self):
             self.set_country(country_code, commit=True)
         return country_code
```
Air-t, line 249: Not sure if setting `CheckoutMetadata` should be part of a `Checkout` model or rather a helper method.
Also I was wondering if proper permission checks are made in order to save `privateMetadata`

saleor/graphql/meta/mutations/utils.py
```diff
@@ -1,11 +1,13 @@
...
 from django.core.exceptions import FieldDoesNotExist, ValidationError
 from django.db import DatabaseError
 
 from ....core.error_codes import MetadataErrorCode
 
 
-def save_instance(instance, metadata_field: str):
-    fields = [metadata_field]
+def save_instance(instance, metadata_field: Union[str, list[str]]):
+    fields = [metadata_field] if isinstance(metadata_field, str) else metadata_field
```
fowczarek, line 10: In my opinion, we should refactor this function so that it always receives a list of fields. 
Currently, `metadata_field` is not a clear name for the variable. The name suggests that we should pass one field but typing suggests it could be one field or a list of fields. It could be confusing in the future.

fowczarek (commented):
In the issue, we have a requirement to also add Medatada to the checkout billing address. Do you plan to do it in a separate PR?

saleor/checkout/models.py
```diff
@@ -250,6 +250,30 @@ def get_country(self):
...
 
+    def set_metadata_from_shipping_address(self, shipping_address):
+        from ..graphql.meta.mutations.utils import save_instance
+        from ..graphql.meta.types import get_valid_metadata_instance
```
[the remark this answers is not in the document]
Air-t: hm, that way I will need to think of putting this method to a separate file or a helper file - this imports were made due to circular imports - but now I know I should avoid this. Do you have any suggestions where to put this set method? or is there any rule to follow?

saleor/graphql/meta/mutations/utils.py
Air-t: right, naming is misleading - it would be better to pass a list in all cases

saleor/checkout/models.py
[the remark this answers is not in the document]
IKarbowiak: I see that currently is used only in `CheckoutShippingAddressUpdate` mutation, so it could be the class method or in `saleor/graphql/checkout/mutations/utils.py` file. 

If you move it to `CheckoutShippingAddressUpdate` and make a class method, you will be able to use the `BaseMutation` methods: `validate_and_update_metadata` and `validate_metadata_keys`.

saleor/graphql/account/tests/mutations/staff/test_address_create.py
```diff
@@ -38,6 +38,8 @@ def test_create_address_mutation(
     # given
     query = ADDRESS_CREATE_MUTATION
     graphql_address_data["city"] = "Dummy"
+    graphql_address_data["metadata"] = [{"key": "key", "value": "public"}]
+    graphql_address_data["privateMetadata"] = [{"key": "key", "value": "private"}]
```
IKarbowiak, line 42: We should also check the response if the metadata are properly set.

saleor/graphql/account/types.py
```diff
@@ -85,6 +86,17 @@ class AddressInput(BaseInputObjectType):
         )
     )
 
+    metadata = graphene.List(
+        graphene.NonNull(MetadataInput),
+        description="Checkout public metadata.",
```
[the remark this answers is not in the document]
IKarbowiak: We should also add `ADDED_IN_X` label. We are adding this label to any new fields that are added to API.

saleor/graphql/account/types.py
```diff
@@ -85,6 +86,17 @@ class AddressInput(BaseInputObjectType):
...
+        required=False,
+    )
+    private_metadata = graphene.List(
+        graphene.NonNull(MetadataInput),
+        description="Checkout private metadata.",
```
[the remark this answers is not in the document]
IKarbowiak: Same here. Missing `ADDED_IN_X` label

saleor/checkout/models.py
```diff
@@ -250,6 +250,30 @@ def get_country(self):
             self.set_country(country_code, commit=True)
         return country_code
 
+    def set_metadata_from_shipping_address(self, shipping_address):
```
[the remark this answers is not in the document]
Air-t: finally changed a bit and only had to fix `CheckoutShippingAddressUpdate.construct_instance`

saleor/graphql/account/mutations/staff/address_create.py
```diff
@@ -73,3 +73,11 @@ def perform_mutation(cls, root, info: ResolveInfo, /, **data):
...
         manager = get_plugin_manager_promise(info.context).get()
         cls.call_event(manager.address_created, instance)
+
+    @classmethod
+    def construct_instance(cls, instance, cleaned_data):
+        cls.update_metadata(instance, cleaned_data.pop("metadata", list))
+        cls.update_metadata(
+            instance, cleaned_data.pop("private_metadata", list), is_private=True
+        )
+        super().construct_instance(instance, cleaned_data)
```
IKarbowiak, line 83:
``` suggestion
        return super().construct_instance(instance, cleaned_data)
```

saleor/graphql/account/tests/mutations/staff/test_address_create.py
```diff
@@ -50,7 +60,12 @@ def test_create_address_mutation(
     data = content["data"]["addressCreate"]
     assert data["address"]["city"] == "DUMMY"
     assert data["address"]["country"]["code"] == "PL"
+    assert data["address"]["metadata"] == [{"key": "key", "value": "public"}]
+    assert data["address"]["privateMetadata"] == [{"key": "key", "value": "private"}]
```
IKarbowiak, line 64: Instead of copying it in both places we could assign it to the variable and reuse

saleor/graphql/meta/mutations/utils.py
```diff
@@ -4,19 +4,22 @@
 from ....core.error_codes import MetadataErrorCode
 
 
-def save_instance(instance, metadata_field: str):
-    fields = [metadata_field]
-
+def save_instance(instance, metadata_fields: list, commit: bool = False):
```
IKarbowiak, line 7: Do we need this `commit` parameter? Are we using `False` somewhere?

saleor/graphql/account/tests/mutations/staff/test_address_create.py
IKarbowiak: Similar in other tests

saleor/graphql/checkout/mutations/checkout_shipping_address_update.py
```diff
@@ -199,3 +199,11 @@ def perform_mutation(
...
 
         return CheckoutShippingAddressUpdate(checkout=checkout)
+
+    @classmethod
+    def construct_instance(cls, instance, cleaned_data):
+        cls.update_metadata(instance, cleaned_data.pop("metadata", list))
+        cls.update_metadata(
+            instance, cleaned_data.pop("private_metadata", list), is_private=True
+        )
+        super().construct_instance(instance, cleaned_data)
```
IKarbowiak, line 209:
```suggestion
        return super().construct_instance(instance, cleaned_data)
```

saleor/graphql/checkout/tests/mutations/test_checkout_billing_address_update.py
```diff
@@ -148,7 +148,7 @@ def test_checkout_billing_address_update_by_id_without_street_address_2(
 )
 def test_checkout_billing_address_update(
     mocked_invalidate_checkout_prices,
-    user_api_client,
+    superuser_api_client,
```
maarcingebala, line 151: `superuser_api_client` is used only in some special cases, since in most cases Saleor tries to relies on permission and not on the `is_superuser` setting (e.g. it's not available to any staff users in Saleor Cloud). I guess we should keep `user_api_client` here, and if we test private metadata we should have the second test with `staff_api_client` and proper permissions granted.

saleor/graphql/meta/mutations/utils.py
Air-t: thought it might be helpful in future use to avoid multiple saves, but in this implementation it is not used - do you think it should be removed? what is the approach - better not to add not used features?

fowczarek: This function is responsible only for saving mutation instances. This function does nothing without committing. We use this pattern with `commit` mostly in functions inside Models.

Air-t: ok so I will remove commit flag if it is redundant here

saleor/graphql/checkout/tests/mutations/test_checkout_billing_address_update.py
Air-t: there might be a wider discussion if there is a use case for updating private metadata in this call. Also there is a new situation where Address do not exist yet as in `updatePrivateMetadata` mutation

CHANGELOG.md
```diff
@@ -74,6 +74,7 @@ Shipping methods can be removed by the user after it has been assigned to a chec
 - Add `search` to `giftCards` query - #13173 by @zedzior
 - Add `ProductBulkTranslate` mutation - #13329 by @SzymJ
 - Add `ProductVariantBulkTranslate` mutation - #13329 by @SzymJ
+- Add `metadata` to `AddressInput` field - #13248 by @Air-t
```
maarcingebala, line 77: Now as I read it I think that adding metadata to this input is the technical detail, while the significant information from the user point of view is that we added the ability to set public address metadata in checkout address mutation. So maybe something like this:
```
- Add the ability to set address public metadata in the following mutations: <mutation A>, <mutation B>, ... - #13248 by @Air-t
  - Add `metadata` to `AddressInput` field
```

maarcingebala: Listing the affected mutation will be also useful for QA team when they're testing it.

saleor/graphql/account/tests/mutations/staff/test_address_create.py
```diff
@@ -33,15 +37,15 @@
 
 
 def test_create_address_mutation(
-    staff_api_client, customer_user, permission_manage_users, graphql_address_data
+    user_api_client, customer_user, permission_manage_users, graphql_address_data
```
maarcingebala, line 40: Do we need to change the API client here? `staff_api_client` represents a staff user so it is fine in this case.

Air-t: good catch

maarcingebala (approved):

saleor/graphql/checkout/tests/mutations/test_checkout_shipping_address_update.py
```diff
@@ -48,6 +48,40 @@
...
+                    }
+                }
+            }
+            errors {
+                field
+                message
+                code
+            }
+        }
+    }"""
```
fowczarek, line 82:
```suggestion
    }
"""
```

saleor/graphql/meta/permissions.py
```diff
@@ -108,14 +108,12 @@ def public_address_permissions(
...
-            company_address_id=address.id
-        ).exists()
-    ):
-        raise PermissionDenied()
+    elif warehouse_models.Warehouse.objects.filter(address_id=address.id).exists():
+        return [ProductPermissions.MANAGE_PRODUCTS]
+    elif site_models.SiteSettings.objects.filter(
+        company_address_id=address.id
+    ).exists():
+        return [SitePermissions.MANAGE_SETTINGS]
```
fowczarek, line 116: Could you describe this change in the PR description and `CHANGELOG.md`?

fowczarek (approved):

yellowee (changes requested):
@Air-t  Please address issues below, thanks:
1. ``privateMetadata`` field has not been added to AddressInput
2. ``addressUpdate`` mutation with new ``metadata`` [https://add-metadata-to-checkout-shipping-mutations.api.saleor.rocks/graphql/#saleor/N4IgjiBcILYK4BcCGCCWB7AdgA[…]0IS3YUQlW2IvmDFPENPqchrHkgI61ECBIA](https://add-metadata-to-checkout-shipping-mutations.api.saleor.rocks/graphql/#saleor/N4IgjiBcILYK4BcCGCCWB7AdgAiQEzwCcBTAZ1IFUAHPFYgCgBJU9JsBJAEQEIAabZpiqI2AQQIly7IYm4BKbMAA6OXBLKUadeizbM8-VDIR6jwhHOWrsxQoXSFSwAGapiAGzzYYGpAHNiAF8VbFC1Ig1gFmwAY3Q4TARCAE9gOLwg72JkWmRgAGtiZOwANyR3OCDg1WrAkF4QdCgQfAjyalziepAS5qtQpRAWQbZBgEUAdQAlfJiYADUALxjF9AAmADkAdwBeHcHeEOxBs0QRxSOB2Gz8FCRzgG1lEELk88GEMgQOugPjnvKlXeIE+pAQ80BxB+n0GgQAuodrP8YqgEG8oP8NgBRCbYACaAHkpgBpA6XZGo5KiEj3DGDMlIwZxGBUJCYZIbJA+YEMsLI+KJFLAigAZV5YSZAqSVJpwI2ePFV1cjgQnO5dJAABUvor-u4kGC1cRgaCELrBlQABZYY0agDUAEY1k6ACwAVgAbAB2AAcAE4AAzmkBUdBg8oAYXQGWBDoDAYdHuDYJI2XEbVIDtjAGYHdnsCKAHTYb0IS3YUQlW2IvmDFPENPqchrHkgI61ECBIA)

Air-t:
@yellowee thanks for checking
1.`privateMetadata` is not added in this ticket, it was removed from the scope. Sorry for misleading ticket description.
2. issue should be fixed now

yellowee (changes requested):
@Air-t 
1. Providing address metadata in most mutations causes Internal server error with Attribute error:
Screenshot 2023-07-20 at 13 18 06

Mutations affected:
customerBulkUpdate, customerUpdate, customerCreate, accountUpdate, draftOrderCreate, checkoutCreate, checkoutShippingAddressUpdate, checkoutBillingAddressUpdate

example [payload](https://add-metadata-to-checkout-shipping-mutations.api.saleor.rocks/graphql/#saleor/N4IgjiBcILYK4BcCGCCWB7AdgAgLIE9dEUNMAKAElUwAdFIBVAZwFMAnAYTZZRYElaiAIQBKbMAA6ObAGM4TBOhjsuPBCzLU6CSNiqCEIydOzZ2bdGybAAZqhYAbACbZlTJkgDmLAL5TTpvLsxgEBqE7+odhOLDZIcA4IAEKoDg7UngCCTk7c7sDKyE4oSMAA1iz42ABuSA5wvj7YkaExcQkIAMoAFqg0NBnZuSz5hUjFyOWVNXUNPvORfpg+IAA0IOhQIAREyGhYayDVWyHNIFqIElDiLWcsMEipV5BX6goAAkzd3TbffwB0MiUV1WkQkCCubXiiRSaUGOTyTGeNxMzQhIBkqAQ+GRVwAcgBRADq2AAmgB5ABKAGkQbcAlcxhMkM8ANqSEAVHFQK4AI1S6UwnhBZ1q9RYz1eI3RPlWN05lUlIH5cKFACYRVcxQ0lW8EBqQD4ALqg1Hgq6Y7GZbgs65SlgOh10s3ooEwGhITD4PFIZS4kDO0LmjHoOCYBBsbm6K4MTqBhmu0PhyPWnj+vGk+OmYN2KwIH1+u0gAAq0qz5vRDiQCgLEqLeqzaKuNG6WDr0ZAAGoAIxq3sAFgArAA2ADsAA4AJwABkbwZo6AUdQ46Bi-u70+n3eHc-RCm4LAQQ0R3fXAGZu2fsJ1-tgxwhuthMtU66ag3uIw6jwiRkwDR2rluWVbkhWJoS6Xp+nhYZ3GRU5s1dLEozOQkSQpGlGwZWBD3GEo2Q5LklS+PoBiFTUjlmdspQUK5jTfBMLSQ1NbQ7dRHSdNYQMTd1PW9X12zOXcLSTCNkJjONOJdYSw1E5j00zST3yuXMa34-1SxoxSEODKtVMLViyy07SQBbNt-R7Ps1SHMcp1nIymxMxdkAcFc1yLDctx3ezg33L9j1-U93IvK8bzvUcHyfF8hJAXzD389x-0EgNUSWUwlhWHwgA)

2. Metadata added to shipping and billing addresses I think should be visible under addresses as well - currently only metadata from billing address are included:
Screenshot 2023-07-20 at 12 11 28

Air-t:
@yellowee after discussion with @maarcingebala we will skip the topic of `User.addresses` field for now.
Other issues with missing mutations should be solved now

yellowee (approved):
All issues are resolved - approving

PR #13368, opened 2023-07-10
Delete event deliveries in batches

jakubkuc:
This is a port of changes from #13246 since we decided to merge it to main instead of 3.9.

I want to merge this change because it adds batches to event deliveries delete task. Fixes https://github.com/saleor/saleor/issues/13235
I'm switching back to using django delete() instead od raw SQL query (which was introduced in https://github.com/saleor/saleor/pull/9806 ) and use the logic that is used in zero-downtime migrations to make the task more optimal.
It will now delete all the EventDelivery objects and by CASCADE all EventDeliveryAttempts (max 5 attempts per delivery) and when it is done it will take care of unused EventPayloads

PERFORMANCE NOTE: 
`BATCH=1000` - was tested on large DB with multiple attempts per delivery and big JSON payloads. Tasks took less than 1s and no significant raise in memory usage (around 1%), tested also with multiple workers at once with no crashes. 

# Impact

* [ ] New migrations

# Pull Request Checklist


kadewu (approved):

IKarbowiak (approved):

zedzior (approved):

korycins (commented):
Looks ok. Have you checked it with a big database (at least a couple of millions of event deliveries, retries, and payloads)? What is the time of execution for a single task in that case?
(I am asking as we need to be sure that BATCH_SIZE=1000 is enough to clean up the database)

saleor/core/tasks.py
```diff
@@ -1,11 +1,13 @@
...
 from django.core.files.storage import default_storage
-from django.db.models import Exists, OuterRef, Q
+from django.db.models import Exists, OuterRef
 from django.utils import timezone
 
 from ..celeryconf import app
-from .models import EventDelivery, EventDeliveryAttempt, EventPayload
+from .models import EventDelivery, EventPayload
+
+BATCH_SIZE = 1000
```
fowczarek, line 10: We should move `PERFORMANCE NOTE:` here.

jakubkuc: I've added comment

jakubkuc:
@korycins Yes it was tested. I've added comment in code so it is more clear :)

tomaszszymanski129 (approved):

saleor/core/tasks.py
```diff
@@ -16,18 +21,15 @@ def delete_from_storage_task(path):
...
-    payloads._raw_delete(payloads.db)  # type: ignore[attr-defined] # raw access # noqa: E501
+    valid_deliveries = EventDelivery.objects.filter(created_at__gt=delete_period)
+    payloads_to_delete = EventPayload.objects.filter(
+        ~Exists(valid_deliveries.filter(payload_id=OuterRef("id")))
+    ).order_by("-pk")
+    ids = payloads_to_delete.values_list("pk", flat=True)[:BATCH_SIZE]
+    qs = EventPayload.objects.filter(pk__in=ids)
+    if ids:
+        qs.delete()
+        delete_event_payloads_task.delay()
```
NyanKiyoshi, line 32: Same comment as here: https://github.com/saleor/saleor/pull/13246#discussion_r1258298416

NyanKiyoshi (changes requested):

saleor/core/tasks.py
jakubkuc: We have decided that we should check all our tasks in search of such probability and make a separate PR with fix that will prevent that from happening in all of them. So that will be out of scope of this PR.

NyanKiyoshi: Why not fix it now instead of introducing technical debt in a bug fix? This would be highly problematic if we start queuing tons of concurrent writes on the same rows as it is a "recursive" function.

You can do something similar as here: https://github.com/saleor/saleor/blob/5e7c57dad9e64b09477ebdcee53f0277359bc598/saleor/checkout/tasks.py#L86-L96

NyanKiyoshi: You could also set an expiration date/timestamp and pass it to the task. If null then you need to give one to the sub tasks, otherwise you just check the value is lower than current time and pass it along the value.

jakubkuc: @NyanKiyoshi I've made changes. What do you think about it?

NyanKiyoshi: Looks good! 👌

NyanKiyoshi (approved):

saleor/core/tasks.py
```diff
@@ -14,20 +25,21 @@ def delete_from_storage_task(path):
 
 
 @app.task
-def delete_event_payloads_task():
+def delete_event_payloads_task(expiration_date=None):
+    expiration_date = expiration_date or timezone.now() + datetime.timedelta(minutes=60)
```
fowczarek, line 29: This task is executed once per day, are you sure that we have such a short expiration date for them?

PR #13380, opened 2023-07-12
Update Graphql Playground

mmiszy:


mmiszy:
https://github.com/saleor/saleor/pull/13379

Issue #13373, opened 2023-07-12
[RFC] Add Attributes bulk mutations


SzymJ:
The goal is to allow creating and modifying attributes and values in bulk.

## General assumptions
- add `AttributeBulkCreate` mutation
- add `AttributeBulkUpdate` mutation
- allow to set `externalReference` for attribute and value

### New inputs
```graphql

input AttributeBulkCreateValueInput @doc(category: "Attributes") {
  name: String!
  value: String
  externalReference: String
  fileUrl: String
  contentType: String
}


input AttributeBulkCreateInput @doc(category: "Attributes") {
  externalReference: String
  inputType: AttributeInputTypeEnum
  entityType: AttributeEntityTypeEnum
  name: String!
  slug: String
  type: AttributeTypeEnum!
  unit: MeasurementUnitsEnum
  values: [AttributeBulkCreateValueInput!]
  valueRequired: Boolean
  isVariantOnly: Boolean
  visibleInStorefront: Boolean
  filterableInDashboard: Boolean
}
```

### New types
```graphql

enum AttributeBulkCreateErrorCode @doc(category: "Attributes") {
  ALREADY_EXISTS
  BLANK
  GRAPHQL_ERROR
  INVALID
  NOT_FOUND
  REQUIRED
  UNIQUE
  DUPLICATED_INPUT_ITEM
  MAX_LENGTH
}

type AttributeBulkCreateError @doc(category: "Attributes") {
  path: String
  message: String
  code: AttributeBulkCreateErrorCode!
}


type AttributeBulkCreateResult @doc(category: "Attributes") {
  attribute: Attribute
  errors: [AttributeBulkCreateError!]
}
```

### New mutations
```graphql
type AttributeBulkCreate @doc(category: "Attributes") @webhookEventsInfo(asyncEvents: [ATTRIBUTE_CREATED], syncEvents: []) {
  count: Int!
  results: [AttributeBulkCreateResult!]!
  errors: [AttributeBulkCreateError!]!
}
```

```[tasklist]
### Tasks
- [ ] https://github.com/saleor/saleor/issues/13374
- [ ] https://github.com/saleor/saleor/issues/13375
```

PR #13398, opened 2023-07-13
Add AttributeBulkCreate mutation

### LINKED ISSUE #13374: Add AttributeBulkCreate mutation

SzymJ:
I want to merge this change because it adds `AttributeBulkCreate` mutation.

Part of #13373
Resolves #13374

# Impact

* [ ] New migrations

# Pull Request Checklist


SzymJ:
More test in progress

saleor/graphql/attribute/mutations/attribute_bulk_create.py
```diff
@@ -0,0 +1,525 @@
...
+from ...core.utils import WebhookEventInfo, get_duplicated_values
+from ...plugins.dataloaders import get_plugin_manager_promise
+from ..descriptions import AttributeDescriptions, AttributeValueDescriptions
+from ..enums import AttributeEntityTypeEnum, AttributeInputTypeEnum, AttributeTypeEnum
+from ..types import Attribute
+
+ONLY_SWATCH_FIELDS = ["file_url", "content_type", "value"]
+
+
+def get_results(instances_data_with_errors_list, reject_everything=False):
```
korycins, line 35:
```python
def get_results(instances_data_with_errors_list, reject_everything=False):
    return [
        AttributeBulkResult(
            attribute=None if reject_everything else data.get("instance"),
            errors=data.get("errors")
        )
        for data in instances_data_with_errors_list
    ]
```

saleor/graphql/attribute/mutations/attribute_bulk_create.py
```diff
@@ -0,0 +1,525 @@
...
+
+            if (
+                value_data.value
+                and input_type not in AttributeInputType.TYPES_WITH_CHOICES
+            ):
+                index_error_map[attribute_index].append(
+                    AttributeBulkCreateError(
+                        path="values",
+                        message=(
+                            "Values cannot be used with " f"input type {input_type}.",
```
korycins, line 347: Looks like it was originally in two separate lines (`f` in the middle)

saleor/graphql/attribute/mutations/attribute_bulk_create.py
```diff
@@ -0,0 +1,525 @@
...
+            index_error_map[attribute_index].append(
+                AttributeBulkCreateError(
+                    path=f"values.{value_index}",
+                    message=(
+                        "Cannot specify both value and file for swatch attribute."
+                    ),
+                    code=AttributeBulkCreateErrorCode.INVALID.value,
+                )
+            )
+        else:
```
korycins, line 390: Why not `elif`?

saleor/graphql/attribute/mutations/attribute_bulk_create.py
```diff
@@ -0,0 +1,525 @@
...
+        instances_data_and_errors_list = []
+
+        for index, cleaned_input in cleaned_inputs_map.items():
+            if not cleaned_input:
+                instances_data_and_errors_list.append(
+                    {"instance": None, "errors": index_error_map[index]}
+                )
+                continue
+            try:
+                values = []
```
korycins, line 430: Maybe we could limit the body of try/except here? I mean, can we leave here a part that already can raise the exception? For example: I don't think the line like `values_data = cleaned_input.pop("values",None)` or `instance = models.Attribute()` can raise an exception. correct?

saleor/graphql/attribute/mutations/attribute_bulk_create.py
```diff
@@ -0,0 +1,525 @@
...
+                    value.full_clean(exclude=["attribute"])
+
+                instances_data_and_errors_list.append(
+                    {
+                        "instance": instance,
+                        "errors": index_error_map[index],
+                        "values": values,
+                    }
+                )
+            except ValidationError as exc:
```
korycins, line 452: Looks like a lack of tests. At least codecov suggests that

saleor/graphql/attribute/mutations/attribute_bulk_create.py
```diff
@@ -0,0 +1,525 @@
...
+        # clean and validate inputs
+        cleaned_inputs_map = cls.clean_attributes(
+            info, data["attributes"], index_error_map
+        )
+        instances_data_with_errors_list = cls.create_attributes(
+            info, cleaned_inputs_map, index_error_map
+        )
+
+        # check error policy
+        if any([True if error else False for error in index_error_map.values()]):
```
korycins, line 508: It looks pretty unclear. Can we refactor `if` condition to be more readable?

saleor/graphql/attribute/mutations/attribute_bulk_create.py
```diff
@@ -0,0 +1,525 @@
...
+            if error_policy == ErrorPolicyEnum.REJECT_EVERYTHING.value:
+                results = get_results(instances_data_with_errors_list, True)
+                return AttributeBulkCreate(count=0, results=results)
+
+            if error_policy == ErrorPolicyEnum.REJECT_FAILED_ROWS.value:
+                for data in instances_data_with_errors_list:
```
korycins, line 514: Not sure if this clean up should be here. IMO `create_attributes` should set up `data["instance"]` as a None in case of the error. Or it should be done by `get_results`.  Now you have `create_attributes` that returns the data, for `reject_everything` the response is created in `get_results`, for `reject_failed_rows` the response data is initially prepared in `perform_mutation` and later it is converted in `get_results`

8r2y5 (approved):

maarcingebala (approved):

szczecha (approved):

saleor/graphql/attribute/mutations/attribute_bulk_create.py
```diff
@@ -0,0 +1,548 @@
...
+                type=WebhookEventAsyncType.ATTRIBUTE_CREATED,
+                description="An attribute was created.",
+            ),
+        ]
+
+    @classmethod
+    def clean_attributes(cls, info, attributes_data, index_error_map):
+        cleaned_inputs_map: dict = {}
+
+        attrs_existing_external_refs = set(
```
korycins, line 142: Are you aware that you're fetching here data from **ALL** Attribute, AttributeValue objects?This won't work in case of having a huge dataset of the attributes

korycins (changes requested):

saleor/graphql/attribute/mutations/attribute_bulk_create.py
SzymJ: Yep, but do you have any other ideas how to validate if external ref is unique before save? @korycins

maarcingebala: Maybe we could do it without loading the entire dataset into memory and check it using a database query with filter and "exists" query?

SzymJ: but this will generate query for every attribute and attribute value from input, so if there will be 200 objects with external refs we will have 200 queries just for checking external ref

kadewu: You can make one query using `IN`, or I'm missing something?

SzymJ: @kadewu but `exists()` returns only `True/False` so I won't be able to check which input has invalid external_ref

kadewu (approved):

PR #13404, opened 2023-07-14
Run metadata webhooks only when metadata has changed

maarcingebala:
### Problem

Currently, `*_METADATA_UPDATED` webhooks are triggered for every call of generic metadata API, even if the metadata hasn't been updated. It results in flooding the listeners with events that are not valuable and result in more noise and network overhead.

Assuming the following scenario:
1. `product.metadata = {}`
2. Call mutation:
```
updateMetadata(
  id: "UHJvZHVjdDox",
  input: [{key: "foo", value: "bar"}]
)
```
`PRODUCT_METADATA_UPDATED` is triggered.

3. Call the mutation again with the same input - the webhook is called again with the same payload as the previous one. This informs listeners that metadata was updated, but it wasn't in this case.

### Proposed solution

This PR changes the metadata API to only trigger webhook when there is an actual change in the metadata. This is done by comparing the metadata state from before the mutation with the mutated state.

# Impact

* [ ] New migrations

# Pull Request Checklist


8r2y5 (approved):

saleor/graphql/meta/mutations/base.py
```diff
@@ -165,10 +166,40 @@ def mutate(cls, root, info: ResolveInfo, **data):
...
+            if instance:
+                metadata_instance = get_valid_metadata_instance(instance)
+                new_metadata = metadata_instance.metadata
+                new_private_metadata = metadata_instance.private_metadata
+                has_changed = (
+                    old_metadata != new_metadata
+                    or old_private_metadata != new_private_metadata
+                )
+            else:
+                has_changed = False
```
aniav, line 198: Nitpick: I find it's cleaner to assign a default before performing the logic so e.g. here we'd have:
```python
            has_changed = False
            
            instance = result.item
            if isinstance(instance, ChannelContext):
                instance = instance.node

            if instance:
                metadata_instance = get_valid_metadata_instance(instance)
                new_metadata = metadata_instance.metadata
                new_private_metadata = metadata_instance.private_metadata
                has_changed = (
                    old_metadata != new_metadata
                    or old_private_metadata != new_private_metadata
                )
                
```

saleor/graphql/meta/utils.py
```diff
@@ -0,0 +1,13 @@
+from ...checkout.models import Checkout
```
aniav, line 1: I have two issues with this:
1. This seems to belong to checkout as it's only operating on the checkout piece. Would it make sense to move it to `graphql.checkout`?
2. We shouldn't be creating *new* `utils.py` files as they become buckets containing a lot of random stuff and instead create something more meaningful. `checkout.utils` is actually a good example of that because it has logic manipulating checkout in place, creating new objects, operating on discounts... all of it 😅 I'd suggest adding it to checkout `utils.py` for now and then splitting this file into specific modules, each responsible for a specific set of functions.

aniav (approved):
Two small nitpicks, but nothing really blocking 🙌

saleor/graphql/meta/mutations/base.py
maarcingebala: I agree; it's more readable this way.

saleor/graphql/meta/utils.py
maarcingebala: I missed the `graphql/meta/mutation/utils.py` file, so adding a new one was unnecessary - fixed. I agree that this is not a perfect pattern, as indeed utils files are already buckets of different stuff. We will have to refactor it at some point.

yellowee:
@maarcingebala 
✅ 1. PRODUCT_METADATA_UPDATED event is triggered on ``updateMetadata`` mutation with new data

✅  2. PRODUCT_METADATA_UPDATED is NOT triggered on ``updateMetadata`` mutation with the same values as before(ad 1), when no changes have been applied

❌ 3. Too many events received on calling updateMetadata mutation with new metadata: 

https://github.com/saleor/saleor/assets/24992893/f2e1a603-55a4-40de-a8ff-4c977fede540

- 3 events with metadata update details as in ad2 

Screenshot 2023-07-17 at 11 53 27

- 2 events with the following user(?) details:

Screenshot 2023-07-17 at 12 09 32

❌  4. Currently **no event is received** when I call ``updateProduct`` mutation with new metadata. 

Please let me know if any other mutations events are affected by the changes, thanks.


Test data:
https://webhook.site/84415625-3b38-4cef-87bd-d2fc6d7e898e

eg. mutation payload:

```
mutation updateMetadata {
  updateMetadata(
    id: "UHJvZHVjdDo3Mg=="
    input: {key: "vatlayer.description", value: "test01"}
  ) {
    errors {
      field
      message
    }
    item {
      metadata {
        key
        value
      }
      metafields
      privateMetadata {
        key
        value
      }
    }
  }
}
```

```
mutation productUpdate {
  productUpdate(
    input: {metadata: {key: "vatlayer.description", value: "test001b"}}
    id: "UHJvZHVjdDo3Mg=="
  ) {
    errors {
      field
      message
    }
    product {
      id
      metadata {
        key
        value
      }
      privateMetadata {
        key
        value
      }
    }
  }
}
```

maarcingebala:
Thanks @yellowee I'll check these cases!

yellowee (approved):
As discussed - please create a separate ticket for fixing event on productUpdate mutation, thanks.

Issue #12881, opened 2023-05-22
Attribute refactoring


maarcingebala:
The goal is to simplify the attribute models. The current attribute model relations are complex and really hard to understand. It causes a high entry threshold for this part of code and makes it hard to keep.

Right now we have available attributes for `Product`, `ProductVariant`, and `Page` models. For each of these models, we are having additional models to keep the relation to the attributes. For each model, the structure is the same. So the problem and potential solution will be shown only on `Product` model, but it's applicable to all of them.

The below chart shows what the current solution looks like:
```mermaid
classDiagram
    Attribute "1" -- "0..*" AttributeValue
    Attribute "1" -- "0..*" AttributeProduct
    AttributeProduct "0..*" -- "1" ProductType
    AssignedProductAttribute "0..*" -- AttributeProduct
    AssignedProductAttribute "0..*" -- "1" Product
    AssignedProductAttribute "0..*" -- "0..*" AssignedProductAttributeValue
    AssignedProductAttributeValue "0..*" -- "1" AttributeValue

  
    Product "0..*" -- "1" ProductType
    class AttributeProduct{
      sort_order
    }
    class AssignedProductAttributeValue{
      sort_order
    }
    class AttributeValue{
      sort_order
    }
```

The proposal is to simplify it to:
```mermaid
classDiagram
    Attribute "1" -- "0..*" AttributeValue
    Attribute "1" -- "0..*" AttributeProduct
    AttributeProduct "0..*" -- "1" ProductType
    AttributeValue "1" -- "0..*" AssignedProductAttributeValue
    AssignedProductAttributeValue "0..*" -- "1" Product
    Product "0..*" -- "1" ProductType
    Product "0..*" -- "0..*" Attribute
    class AttributeProduct{
      sort_order
    }
    class AssignedProductAttributeValue{
      sort_order
    }
    class AttributeValue{
      sort_order
    }
```

## Database changes
To apply proposed changes the following steps need to be performed.
1. Add `Product` - `Attribute` relation
2. Go through all `AttributeProduct` and based on `attribute` and `assigned_products` fields create relations between attributes and products.
3. Add `product` field to `AssignedProductAttributeValue` model.
4. Fulfill `AssignedProductAttributeValue.product` fields based on `AssignedProductAttributeValue.assignment.product` field values.
5. Drop `AttributeProduct.assigned_products` field.
6. Drop `AssignedProductAttribute` model.

The same for `ProductVariant` and `Page` relation.

## Dataloaders changes
[...]

PR #13407, opened 2023-07-14
Simplify Product - Attribute relation

aniav:
I want to merge this change because we're moving away from intermediate model `AssignedProductAttribute` in the Product - Attribute relation.

Part of https://github.com/saleor/saleor/issues/12881.

# What has changed
- the `AssignedProductAttribute` has been removed
- in order to know what attributes are assigned to a `Product` we should now check attributes that belong to a `ProductType` as a source of truth
- `AssignedProductAttributeValue` now holds a relation to `Product` without an intermediate `assignment`
- `SelectedAttributesByProductIdLoader` is now using a new dataloader: `AttributeValuesByProductIdLoader`

# Note
The db queries are not yet optimised but as this PR is quite extensive already I'm giving it away for reviews while I make updates to the db queries.

# Impact

* [x] New migrations

# Pull Request Checklist

* [x] Privileged queries and mutations are either absent or guarded by proper permission checks
* [x] Database migrations are either absent or optimized for zero downtime
* [x] The changes are covered by test cases


saleor/attribute/migrations/0031_extend_attribute_value_length.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 3.2.20 on 2023-08-21 14:50
+
+from django.db import migrations, models
```
aniav, line 3: Fixes #13735

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ All notable, unreleased changes to this project will be documented in this file.
 # 3.16.0 [Unreleased]
 
 ### Breaking changes
+- Change the Attribute - Product relation to decrease code complexity and make it easier to understand the relations - #13407 by @aniav
+- The CSV export will now use empty string for empty values instead of a whitespace value
```
korycins, line 9: Is it an expected changelog here?

saleor/attribute/migrations/tasks/saleor3_16.py
```diff
@@ -0,0 +1,49 @@
...
+
+def update_product_assignment():
+    """Assign product_id to a new field on assignedproductattributevalue.
+
+    Take the values from attribute_assignedproductattribute to product_id.
+    The old field has already been deleted in Django State operations so we need
+    to use raw SQL to get the value and copy the assignment from the old table.
+    """
+    with connection.cursor() as cursor:
+        cursor.execute(
```
korycins, line 19: Are you sure that this is the correct approach? Previously we were doing such actions in patch releases.
So assuming that the newest release is 3.15.0, and main is unreleased 3.16. 
- 3.15.1 should receive migrations, which will move the data to the new fields
- 3.16.0 could receive alterField to remove `null=True`.

---
With the current approach, we have a situation, in which we already removed the field from Django State but we haven't triggered the task yet. It means that 3.16 workers (which operate on new code) will not have details mandatory to correctly display the data. 

---

Some stores can have thousands of attributes, which means that until fully processing the celery task we could display corrupted data, am I correct?

saleor/attribute/models/base.py
```diff
@@ -322,9 +324,7 @@ def get_translation_context(self):
                 elif assigned_product_attribute_value := (
                     attribute_value.productvalueassignment.first()
                 ):
-                    if product_id := (
-                        assigned_product_attribute_value.assignment.product_id
-                    ):
+                    if product_id := (assigned_product_attribute_value.product_id):
```
korycins, line 327: You don't need `(`, `)`

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ All notable, unreleased changes to this project will be documented in this file.
 # 3.16.0 [Unreleased]
 
 ### Breaking changes
+- Change the Attribute - Product relation to decrease code complexity and make it easier to understand the relations - #13407 by @aniav
```
korycins, line 8: Is it breaking change? It shouldn't have any impact on the Saleor users, right? It is only backend refactoring

CHANGELOG.md
korycins: I see that you also touched the csv logic, and it looks like this changelog is valid. I am not sure that we should place such changes in the PR related to Attributes. I would never think that Attribute refactors could include the changes in CSV.
Also, this is a breaking change, that could have an impact on existing services that consume our CSV, is that correct? In that case, we should think about how we provide backward compatibility, I don't think that we can just change it. I know that this is a small change but still, it can be a breaking change.

aniav: While working on the changes in my PR I noticed we have two types of empty values in the CSV exports: whitespace for situations where the value was not assigned and an empty string for situations where the value is assigned and is effectively empty. If I wanted to keep this state it would require a lot of unwanted changes in my PR so I decided to also incorporate it here.

I agree it doesn't belong here. I'll create a separate PR for these changes to be reviewed before going back to this PR.

aniav: Moved to: https://github.com/saleor/saleor/pull/13779

saleor/graphql/product/tests/mutations/test_product_update.py
```diff
@@ -773,8 +782,9 @@ def test_update_product_clear_attribute_values(
 
     assert len(attributes) == 1
     assert not attributes[0]["values"]
-    with pytest.raises(product_attr._meta.model.DoesNotExist):
-        product_attr.refresh_from_db()
+    # Why do we need this?
+    # with pytest.raises(product_attr._meta.model.DoesNotExist):
+    #     product_attr.refresh_from_db()
```
IKarbowiak, line 787: I suppose that we deleted the `product_attribute` model when there were no more values assigned, but it looks like in current solution is not valid anymore

saleor/graphql/product/dataloaders/attributes.py
```diff
@@ -247,85 +167,64 @@ def map_assignment_to_values(values):
...
-            assigned_product_attribute_ids = [
-                a.id for attrs in product_attributes for a in attrs
-            ]
-            product_type_ids = list({p.product_type_id for p in products})
-            product_attributes = dict(zip(keys, product_attributes))
+        attribute_values = list(
+            AssignedProductAttributeValue.objects.using(self.database_connection_name)
+            .filter(product_id__in=keys)
+            .prefetch_related("value")
+            .iterator()
```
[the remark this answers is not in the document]
aniav: Yes, db `iterator()` is a bit different to python iterator and using it will in this case cause the db to not store the resultset in memory, which will result in a slight optimisation at the db level. I will add a comment about it 🙌

saleor/attribute/migrations/0038_remove_assignedproductattribute.py
```diff
@@ -0,0 +1,108 @@
...
+                WHERE attribute_assignedproductattributevalue.assignment_id = attribute_assignedproductattribute.id
+            )
+            WHERE id IN (
+                SELECT ID FROM attribute_assignedproductattributevalue
+                WHERE product_id IS NULL
+                ORDER BY ID DESC
+                FOR UPDATE
+            );
+            """,  # noqa
+        ),
```
IKarbowiak, line 57: Missing `reverse_sql=migrations.RunSQL.noop` to ensure that migration is reversible

saleor/attribute/migrations/0038_remove_assignedproductattribute.py
```diff
@@ -0,0 +1,108 @@
...
+            name="assignment",
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.SET_NULL,
+                related_name="productassignments",
+                to="attribute.AttributeProduct",
+                null=True,
+            ),
+        ),
+        # Remove FK constraints so that the DB doesn't try to truncate the references
+        migrations.RunSQL(
```
IKarbowiak, line 90: Same here 👮‍♀️

saleor/attribute/migrations/0038_remove_assignedproductattribute.py
```diff
@@ -0,0 +1,108 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("attribute", "0037_remove_assignedpageattribute"),
+    ]
+
+    state_operations = [
+        migrations.RemoveField(
```
IKarbowiak, line 13: Also, do we have a task for removing those tables from DB?

IKarbowiak (approved):

saleor/attribute/migrations/0038_remove_assignedproductattribute.py
aniav: https://github.com/saleor/saleor/issues/14594 🫡

korycins (approved):

Issue #3915, opened 2019-04-03
Refactor test fixtures


maarcingebala:
We're using `pytest` fixtures to provide data for our tests, however the `conftest.py` file grew pretty large and it became difficult to maintain. The biggest issue is that we have to duplicate a lot of fixtures that are often similar but produce slight differences that are required by particular tests.

E.g. we have three separate fixtures for orders: `order`, `order_with_lines`, order_list`

@korycins suggested that we could use [factory_boy](https://factoryboy.readthedocs.io/en/latest/) to create factories which, from what I've seen so far, are quite versatile. One advantage of this lib is that it allows creating both saved and unsaved model instances. We could also investigate how to refactor our `conftest.py` to have more versatile fixtures.

maarcingebala:
E.g. we have three similar fixtures for creating orders, all of them save instances in the database:
- `order`
- `order_with_lines`
- `orders_list`

Question is how could we rewrite them to simplify the code with either pure `pytest` or `factory_boy`. From what I saw, in `factory_boy` we could have only one factory to easily generate all three fixtures (@korycins correct me if I'm wrong). Is it possible with `pytest`?

korycins:
@maarcingebala If you want to have correct relation between objects you need to have factories for each related object like `lines`.
Before we will answer your question - 'how we can rewrite them' I would also ask about functionalities that we would like to have. Like - customizing given fields, with/without save object to DB, batch creation.

korycins:
@maarcingebala We should also consider splitting test files into smaller files. For example `test_product` has over 2k lines, it is hard to find anything there. It contains tests for product's forms, views, models utils as well as tests for categories and collections.

maarcingebala:
Related issue: https://github.com/mirumee/saleor/issues/2120

beatboxchad:
Out of a recent project, I can offer a PR that breaks the ice with the Pipfile/settings for factory_boy, a `UserFactory`, and an `AddressFactory`. Would you guys need something more in-depth? Maybe a few more factories? 

As to the bigger task of refactoring the tests to use them, perhaps retaining fixtures where appropriate, I don't have much to offer at this time, but hope to change that as time permits. 

I think in most cases, fixtures can be replaced with factories, which provide reasonable defaults and are configured in the test. I'll bet there will be a reason to use a pytest fixture here and there, perhaps to replicate complicated one-off real-world data. But in general, the factories are super ergonomic to write and use. I like them a lot.

aniav:
[...]

PR #13445, opened 2023-07-18
refactor channel factories

Air-t:
Adds channel factory

Continues https://github.com/saleor/saleor/issues/3915

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/channel/tests/fixtures.py
+++ b/saleor/graphql/channel/tests/fixtures.py
@@ -7,76 +7,84 @@ from ....tax import TaxCalculationStrategy
 from ....tax.models import TaxConfiguration
 
 
-def _create_channel_tax_configuration(channel):
-    # Use TAX_APP strategy, to enable calculations with plugins by default.
-    tax_configuration = TaxConfiguration.objects.create(
-        channel=channel,
-        metadata={"key": "value"},
-        private_metadata={"key": "value"},
-        tax_calculation_strategy=TaxCalculationStrategy.TAX_APP,
-    )
-    tax_configuration.country_exceptions.create(
-        country="PL",
-        tax_calculation_strategy=TaxCalculationStrategy.TAX_APP,
-    )
-    tax_configuration.country_exceptions.create(
-        country="DE",
-        tax_calculation_strategy=TaxCalculationStrategy.TAX_APP,
-    )
-
-
 @pytest.fixture
-def channel_USD(db):
-    slug = settings.DEFAULT_CHANNEL_SLUG
-    channel = Channel.objects.create(
-        name="Main Channel",
-        slug=slug,
-        currency_code="USD",
-        default_country="US",
+def channel_generator(db):
+    def _create_channel_tax_configuration(channel):
+        # Use TAX_APP strategy, to enable calculations with plugins by default.
+        tax_configuration = TaxConfiguration.objects.create(
+            channel=channel,
+            metadata={"key": "value"},
+            private_metadata={"key": "value"},
+            tax_calculation_strategy=TaxCalculationStrategy.TAX_APP,
+        )
+        tax_configuration.country_exceptions.create(
+            country="PL",
+            tax_calculation_strategy=TaxCalculationStrategy.TAX_APP,
+        )
+        tax_configuration.country_exceptions.create(
+            country="DE",
+            tax_calculation_strategy=TaxCalculationStrategy.TAX_APP,
+        )
+
+    def create_channel(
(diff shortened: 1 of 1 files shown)


8r2y5 (approved):

patrys (commented):
This increases complexity and adds more code. If we have time to refactor tests, we should refactor them to not use the ORM and instead mock dataloaders.

aniav (commented):
I have a few issues with the solution here.
1. We have added more code in order to simplify things which doesn't seem to really help
2. If we want to create a function doing things we should leave it as a function, instead of marking it as fixture. Not eveything that is in `fixtures.py` has to be a fixture. This fixture is doing stuff inside to actually act as a function. Why not make it a function instead?
3. We can achieve the same result with composable fixtures

In order not to repeat writing the same `objects.create` with a lot of defaults we could:

1. Create a base channel fixture
```python
@pytest.fixture
def base_channel():
   return Channel(
       name="Main Channel",
       slug=settings.DEFAULT_CHANNEL_SLUG,,
       currency_code="USD",
       default_country="US",
       is_active=is_active,
       allocation_strategy=AllocationStrategy.PRIORITIZE_HIGH_STOCK,
   )
```

2. Compose other fixtures on top of that
```python
@pytest.fixture
def channel_USD(base_channel):
    base_channel.save()
    return base_channel


@pytest.fixture
def other_channel_USD(base_channel):
    base_channel.name="Other Channel USD"
    base_channel.slug="other-usd"
    base_channel.save()
    _create_channel_tax_configuration(base_channel)



@pytest.fixture
def channel_JPY(base_channel):
   base_channel.name="Channel=JPY",
   base_channel.slug="c-jpy",
   base_channel.currency_code="JPY",
   base_channel.default_country="JP",
   base_channel.save()
   _create_channel_tax_configuration(base_channel)
  
   return base_channel
```

This way:
- we have a way to add a new field to `base_channel` when we add it to the model without having to add it in all fixtures. If the new field doesn't matter for a specific fixture it will have a default. If it does matter we would add it to a specific fixture for one of the countries.
- we don't run `save()` on the db multiple times and we only store the specific channel we need in a specific fixture
- we run the tax configurations setup in each of the defined fixtures so it's instantly clear if a fixture does create it or doesn't without having to analyse any additional function logic (tests should be as readable as possible because we write them once and them read them each time we introduce a regression)
- it's ok to repeat code in tests e.g. run the `_create_channel_tax_configuration` in each channel fixture; what we are aiming at is not having to write mundane repeatable code many times when something changes in the models and just calling a function in a fixture won't do us any harm

aniav (commented):
Actually updating my previous comment. The previous implementation would make all fixtures operate on the same db instance so that wouldn't work. But we could still have it with a helper function like so:

```python
def create_channel(
     name=None,
     slug=None,
     currency_code=None,
     default_country=None,
     is_active=True,
     allocation_strategy=AllocationStrategy.PRIORITIZE_HIGH_STOCK,
     create_tax_config=True,
 ):
     channel = Channel.objects.create(
         name=name,
         slug=slug,
         currency_code=currency_code,
         default_country=default_country,
         is_active=is_active,
         allocation_strategy=allocation_strategy,
     )
```

And then:
```python
def channel_USD(channel_generator):
    channel = create_channel(
         name="Main Channel",
         slug=settings.DEFAULT_CHANNEL_SLUG,
         currency_code="USD",
         default_country="US",
    )
    _create_channel_tax_configuration(channel)
    return channel
```

Similar to the code we had initially but still making sure the defaults can be added to the helper when needed without having to add them in each fixture.

Air-t:
thanks @patrys for explaining the difference in performance for calling fixtures in a test methods and just adding them as args - didn't know that.

I guess using a `channel_generator` makes it easier to customise each fixture (not sure if `channel_generator` should be fixture too but I was following the new implementation logic from @kadewu in this [PR](https://github.com/saleor/saleor/pull/13153). 

Assuming the factory will not be customised in test cases there should be no performance impact. 
I measured the test time for old and new setup and is almost the same, around 16mins for both with slight advantage for the new setup always around 10 sec better but in might be just a coincidence.

I heard that the test topic needs a wider discussion so I didn't mark this PR to be reviewed - thanks anyway - I know that implementation might be changed in a near future, or just after finding a proper way of using factories in project.

maarcingebala:
Closing for now, let's get back to this when we agree on a solution in RFC https://github.com/saleor/saleor/issues/13494 and we have some time dedicated to improving tests.

PR #13453, opened 2023-07-18
Refactor promotion models

zedzior:
I want to merge this change, because it replaces `Promotion` and `PromotionRule` id with UUID. The PR also alter `old_sale` field into `old_sale_id` field, which will be used by sales queries and mutations to identify old sale. 

# Impact

* [ ] New migrations

# Pull Request Checklist


saleor/graphql/discount/filters.py
```diff
@@ -154,7 +154,7 @@ def filter_start_date_range(qs, _, value):
 
     @staticmethod
     def filter_is_old_sale(qs, _, value):
-        return qs.filter(old_sale=value)
+        return qs.exclude(old_sale_id__isnull=value)
```
aniav, line 157: Shouldn't this be: `qs.exclude(old_sale_id__isnull=True)`? And it doesn't need a `value` passed in this case, I think?

zedzior: With `value=False` passed you can get promotion set which doesn't derive from old sale. I think it can be useful.

aniav: Ah, makes sense, for some reason I thought value is something else than a bool 😅

8r2y5 (approved):

SzymJ (approved):

PR #13455, opened 2023-07-18
Add pagination to set_description_plaintext_task

8r2y5:
I want to merge this change because it prevents infinite loop.

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/product/migrations/tasks/saleor3_9.py
+++ b/saleor/product/migrations/tasks/saleor3_9.py
@@ -18,8 +18,9 @@ def set_description_plaintext(qs: QuerySet[Product]):
 
 @app.task
 def set_description_plaintext_task(last_pk: int = 0):
+    filter_kwargs = {"pk__lt": last_pk} if last_pk else {}
     products = Product.objects.filter(
-        Q(description_plaintext="", pk__gt=last_pk) & ~Q(description=None)
+        Q(description_plaintext="", **filter_kwargs) & ~Q(description=None)
     ).order_by("-pk")
 
     ids = products.values_list("pk", flat=True)[:2000]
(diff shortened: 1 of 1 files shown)


patrys (changes requested):
It needs explicit ordering by PK to work. Otherwise, it could accidentally find the last product in the first batch.

saleor/product/migrations/tasks/saleor3_9.py
```diff
@@ -17,14 +17,14 @@ def set_description_plaintext(qs: QuerySet[Product]):
...
+        Q(description_plaintext="", pk__gt=last_pk) & ~Q(description=None)
     ).order_by("-pk")
 
     ids = products.values_list("pk", flat=True)[:2000]
     qs = Product.objects.filter(pk__in=ids)
 
     if ids:
         set_description_plaintext(qs)
-        set_description_plaintext_task.delay()
+        set_description_plaintext_task.delay(qs.last().pk)
```
patrys, line 30: Will `.last()` not result in a new DB call?

saleor/product/migrations/tasks/saleor3_9.py
```diff
@@ -17,14 +17,14 @@ def set_description_plaintext(qs: QuerySet[Product]):
 
 
 @app.task
-def set_description_plaintext_task():
+def set_description_plaintext_task(last_pk: int = 0):
```
kadewu, line 20: Don't we need to change the direction of ordering if we start from 0 and use `greater_than`?

saleor/product/migrations/tasks/saleor3_9.py
kadewu, line 30: And probably `ids[-1]` will be safer if applied what I mention above.

8r2y5: It won't work as slices does not provide that functionality

patrys: `max(ids)` should work, though, no? The QuerySet is already evaluated.

kadewu (approved):

patrys (approved):

PR #13495, opened 2023-07-21
Move managers out from models.py in product

aniav:
I want to merge this change because it's making `models.py` in product cleaner. All code related to managers lives in `managers.py` now. This is just a code cleanup, no logic was changed.

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/product/constants.py
+++ b/saleor/product/constants.py
@@ -1,0 +1,9 @@
+from ..permission.enums import DiscountPermissions, OrderPermissions, ProductPermissions
+
+ALL_PRODUCTS_PERMISSIONS = [
+    # List of permissions, where each of them allows viewing all products
+    # (including unpublished).
+    OrderPermissions.MANAGE_ORDERS,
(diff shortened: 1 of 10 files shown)


saleor/product/constants.py
```diff
@@ -0,0 +1,9 @@
+from ..permission.enums import DiscountPermissions, OrderPermissions, ProductPermissions
```
aniav, line 1: Should this file be called `constants`? Or do we have another preferred naming?

Moving this out from `models` as it's not even used there and instead it's imported all around the place everywhere else.

maarcingebala: We tend to put such constants in `__init__` files of the modules where they are commonly used or belong by domain.

maarcingebala (approved):

8r2y5 (approved):

fowczarek (approved):

SzymJ (approved):

PR #13497, opened 2023-07-23
Add support for unhandled field trackInventoryByDefault .

### LINKED ISSUE #13492: Add suport for unhandled field `trackInventoryByDefault`
The `Shop` type defines the field `trackInventoryByDefault`, which is also available in the `ShopSettingsInput` input, but it's not used in the code anywhere. Most likely the field was added a long time ago in one of the first versions of API and removed before we started properly using field deprecation. It sounds useful, however, to have a field that would be used as a default value for `ProductVariant.trackInventory`.

TODO:
- Add logic that would use the `Shop.trackInventoryByDefault` value (if it is set) to populate `trackInventory` value (if not provided) in the following mutations:
  - `productVariantCreate`
  - `productVariantBulkCreate`
- Add a description to `Shop.trackInventoryByDefault` explaining that this field is used as a default value for ProductVariant.trackInventory
- Extend the description of all `trackInventory` fields to mention that if the field is not provided, `Shop.trackInventoryByDefault` would be used.

Smit-Parmar:
it resolves #13492 

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,7 +84,7 @@ Shipping methods can be removed by the user after it has been assigned to a chec
   - `accountUpdate`, `accountAddressCreate`, `accountAddressUpdate`, `addressCreate`, `AccountAddressUpdate`,
   - `checkoutShippingAddressUpdate`, `checkoutBillingAddressUpdate`, `shopAddressUpdate`, `warehouseUpdate`
   - Add `metadata` to `AddressInput` field
-- Add suport for unhandled field `trackInventoryByDefault` for `productVariantCreate` and `productVariantBulkCreate` - #13492 by @Smit-Parmar
+- Add support for unhandled field `Shop.trackInventoryByDefault` for `productVariantCreate` and `productVariantBulkCreate` - #13492 by @Smit-Parmar
 ### Saleor Apps
 
 - Introduce `Saleor-Schema-Version` HTTP header in app manifest fetching and app installation handshake requests. - #13075 by @przlada
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,7 +84,7 @@ Shipping methods can be removed by the user after it has been assigned to a chec
   - `accountUpdate`, `accountAddressCreate`, `accountAddressUpdate`, `addressCreate`, `AccountAddressUpdate`,
   - `checkoutShippingAddressUpdate`, `checkoutBillingAddressUpdate`, `shopAddressUpdate`, `warehouseUpdate`
   - Add `metadata` to `AddressInput` field
-
+- Add suport for unhandled field `trackInventoryByDefault` for `productVariantCreate` and `productVariantBulkCreate` - #13492 by @Smit-Parmar
 ### Saleor Apps
 
 - Introduce `Saleor-Schema-Version` HTTP header in app manifest fetching and app installation handshake requests. - #13075 by @przlada
--- a/saleor/graphql/product/bulk_mutations/product_variant_bulk_create.py
+++ b/saleor/graphql/product/bulk_mutations/product_variant_bulk_create.py
@@ -7,8 +7,6 @@ from django.core.exceptions import ValidationError
 from django.db.models import F
 from graphene.utils.str_converters import to_camel_case
 
-from saleor.graphql.shop.utils import get_track_inventory_by_default
-
 from ....attribute import AttributeType
 from ....core.tracing import traced_atomic_transaction
 from ....permission.enums import ProductPermissions
@@ -49,6 +47,7 @@ from ...core.types import (
 from ...core.utils import get_duplicated_values
 from ...core.validators import validate_price_precision
 from ...plugins.dataloaders import get_plugin_manager_promise
+from ...shop.utils import get_track_inventory_by_default
 from ..mutations.channels import ProductVariantChannelListingAddInput
 from ..mutations.product.product_create import StockInput
 from ..mutations.product_variant.product_variant_create import ProductVariantInput
--- a/saleor/graphql/product/bulk_mutations/product_variant_bulk_create.py
+++ b/saleor/graphql/product/bulk_mutations/product_variant_bulk_create.py
@@ -866,11 +866,12 @@ class ProductVariantBulkCreate(BaseMutation):
             if not variant:
                 continue
             track_inventory_by_default = get_track_inventory_by_default(info)
+            track_inventory = variant_data["cleaned_input"].get("track_inventory")
             if track_inventory_by_default is not None:
                 variant.track_inventory = (
                     track_inventory_by_default
-                    if variant_data["cleaned_input"].get("track_inventory") is None
-                    else variant_data["cleaned_input"].get("track_inventory")
+                    if track_inventory is None
+                    else track_inventory
                 )
             variants_to_create.append(variant)
             cleaned_input = variant_data["cleaned_input"]
--- a/saleor/graphql/product/bulk_mutations/product_variant_bulk_create.py
+++ b/saleor/graphql/product/bulk_mutations/product_variant_bulk_create.py
@@ -887,7 +887,6 @@ class ProductVariantBulkCreate(BaseMutation):
 
             if not variant.name:
                 cls.set_variant_name(variant, cleaned_input)
-        print(variants_to_create)
         models.ProductVariant.objects.bulk_create(variants_to_create)
 
         for variant, attributes in attributes_to_save:
--- a/saleor/graphql/product/bulk_mutations/product_variant_bulk_create.py
+++ b/saleor/graphql/product/bulk_mutations/product_variant_bulk_create.py
@@ -865,11 +865,13 @@ class ProductVariantBulkCreate(BaseMutation):
 
             if not variant:
                 continue
-            variant.track_inventory = (
-                get_track_inventory_by_default()
-                if variant_data.get("track_inventory") is None
-                else variant_data.get("track_inventory")
-            )
+            track_inventory_by_default = get_track_inventory_by_default()
+            if track_inventory_by_default is not None:
+                variant.track_inventory = (
+                    track_inventory_by_default
+                    if variant_data.get("track_inventory") is None
+                    else variant_data.get("track_inventory")
+                )
             variants_to_create.append(variant)
             cleaned_input = variant_data["cleaned_input"]
 
--- a/saleor/graphql/product/bulk_mutations/product_variant_bulk_create.py
+++ b/saleor/graphql/product/bulk_mutations/product_variant_bulk_create.py
@@ -7,6 +7,8 @@ from django.core.exceptions import ValidationError
 from django.db.models import F
 from graphene.utils.str_converters import to_camel_case
 
+from saleor.graphql.shop.utils import get_track_inventory_by_default
+
 from ....attribute import AttributeType
 from ....core.tracing import traced_atomic_transaction
 from ....permission.enums import ProductPermissions
@@ -864,7 +866,11 @@ class ProductVariantBulkCreate(BaseMutation):
 
             if not variant:
                 continue
-
+            variant.track_inventory = (
+                get_track_inventory_by_default()
+                if variant_data.get("track_inventory") is None
+                else variant_data.get("track_inventory")
+            )
             variants_to_create.append(variant)
             cleaned_input = variant_data["cleaned_input"]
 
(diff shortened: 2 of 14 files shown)


saleor/graphql/product/mutations/product_variant/product_variant_create.py
```diff
@@ -5,6 +5,8 @@
 from django.core.exceptions import ValidationError
 from django.utils.text import slugify
 
+from saleor.graphql.shop.utils import get_track_inventory_by_default
```
maarcingebala, line 8: We should use relative imports for consistency.

saleor/graphql/product/mutations/product_variant/product_variant_create.py
```diff
@@ -297,6 +300,12 @@ def check_for_duplicates_in_stocks(cls, stocks_data):
     @classmethod
     def save(cls, info: ResolveInfo, instance, cleaned_input):
         new_variant = instance.pk is None
+        instance.track_inventory = (
+            get_track_inventory_by_default()
+            if cleaned_input.get("track_inventory") is None
+            else cleaned_input.get("track_inventory")
+        )
+        instance.save()
```
maarcingebala, line 308: There are two things we need to take care of here.

1. `ProductVariantUpdate` mutation inherits from `ProductVariantCreate`, so this logic would run there as well (which we don't want). One option I see here is to create a method `set_track_inventory` in both mutations and provide different logic. In `ProductVariantCreate` we would have:

```
def set_track_inventory(cls, _info, instance, cleaned_input):
    instance.track_inventory = (
        get_track_inventory_by_default()
        if cleaned_input.get("track_inventory") is None
        else cleaned_input.get("track_inventory")
    )
```

In `ProductVariantUpdate`:
```
def set_track_inventory(cls, _info, instance, cleaned_input):
    track_inventory = cleaned_input.get("track_inventory")
    if track_inventory is not None:
        instance.track_inventory = track_inventory
```

Then in the `save` method, we could only call `set_track_inventory` and in both mutations, the proper implementation of this method would be called.

2. `instance.save()` in line 308 is not needed  as there is a save in line 310.

saleor/graphql/product/mutations/product_variant/product_variant_create.py
```diff
@@ -60,6 +62,7 @@ class ProductVariantInput(BaseInputObjectType):
         description=(
             "Determines if the inventory of this variant should be tracked. If false, "
             "the quantity won't change when customers buy this item."
+            "If the field is not provided, Shop.trackInventoryByDefault will be used."
```
maarcingebala, line 65:
```suggestion
            "If the field is not provided, `Shop.trackInventoryByDefault` will be used."
```

saleor/graphql/product/types/products.py
```diff
@@ -276,7 +276,12 @@ class ProductVariant(ChannelContextTypeWithMetadata[models.ProductVariant]):
         description="The product to which the variant belongs.",
     )
     track_inventory = graphene.Boolean(
-        required=True, description="Whether the inventory of the variant is tracked."
+        required=True,
+        description=(
+            "Determines if the inventory of this variant should be tracked. If false, "
+            "the quantity won't change when customers buy this item."
+            "If the field is not provided, Shop.trackInventoryByDefault will be used."
```
maarcingebala, line 283:
```suggestion
            "If the field is not provided, `Shop.trackInventoryByDefault` will be used."
```

saleor/graphql/shop/mutations/shop_settings_update.py
```diff
@@ -29,7 +29,10 @@ class ShopSettingsInput(graphene.InputObjectType):
     header_text = graphene.String(description="Header text.")
     description = graphene.String(description="SEO description.")
     track_inventory_by_default = graphene.Boolean(
-        description="Enable inventory tracking."
+        description=(
+            "This field is used as a default value for "
+            "ProductVariant.trackInventory."
```
maarcingebala, line 34:
```suggestion
            "`ProductVariant.trackInventory`."
```

saleor/graphql/shop/types.py
```diff
@@ -248,7 +248,10 @@ class Shop(graphene.ObjectType):
         required=True,
     )
     track_inventory_by_default = graphene.Boolean(
-        description="Enable inventory tracking."
+        description=(
+            "This field is used as a default value for "
+            "ProductVariant.trackInventory."
```
maarcingebala, line 253:
```suggestion
            "`ProductVariant.trackInventory`."
```

saleor/graphql/shop/utils.py
```diff
@@ -26,3 +28,12 @@ def get_countries_codes_list(attached_to_shipping_zones: Optional[bool] = None):
...
     return all_countries_codes
+
+
+def get_track_inventory_by_default():
+    track_inventory_by_default = (
+        SiteSettings.objects.all().values("track_inventory_by_default").first()
+    )
+    if track_inventory_by_default is not None:
+        return track_inventory_by_default["track_inventory_by_default"]
+    raise NotImplementedError("Site settings not found.")
```
maarcingebala, line 39: Raising an error here means we need to handle it in places where this function is called, but I think we have a better option - return `None` and change usages of `get_track_inventory_by_default` to not set anything on the variant instance if `None` is returned.

CHANGELOG.md
```diff
@@ -84,7 +84,7 @@ Shipping methods can be removed by the user after it has been assigned to a chec
   - `accountUpdate`, `accountAddressCreate`, `accountAddressUpdate`, `addressCreate`, `AccountAddressUpdate`,
   - `checkoutShippingAddressUpdate`, `checkoutBillingAddressUpdate`, `shopAddressUpdate`, `warehouseUpdate`
   - Add `metadata` to `AddressInput` field
-
+- Add suport for unhandled field `trackInventoryByDefault` for `productVariantCreate` and `productVariantBulkCreate` - #13492 by @Smit-Parmar
```
maarcingebala, line 87:
```suggestion
- Add support for unhandled field `Shop.trackInventoryByDefault` for `productVariantCreate` and `productVariantBulkCreate` - #13492 by @Smit-Parmar
```

saleor/graphql/product/mutations/product_variant/product_variant_create.py
```diff
@@ -294,9 +296,18 @@ def check_for_duplicates_in_stocks(cls, stocks_data):
                 }
             )
 
+    @classmethod
+    def set_track_inventory(cls, _info, instance, cleaned_input):
+        instance.track_inventory = (
+            get_track_inventory_by_default()
```
maarcingebala, line 302: After recent changes, `get_track_inventory_by_default` can return `None`, an invalid value for a boolean field. We would have to rewrite this part to ensure that we set `instance.track_inventory` when we get a non-null result from `get_track_inventory_by_default`. This applies to all usages of this function.

saleor/graphql/shop/utils.py
```diff
@@ -2,6 +2,8 @@
 
 from django_countries import countries
 
+from saleor.site.models import SiteSettings
```
maarcingebala, line 5: For importing from Saleor modules use relative imports for consistency :)

saleor/graphql/product/mutations/product_variant/product_variant_create.py
Smit-Parmar: Yeah that's why I raised error. NVM now added code to handle None.

saleor/graphql/shop/utils.py
Smit-Parmar: Ops silly me :)

maarcingebala (approved):

maarcingebala (commented):
I would also add tests where we explicitly set `SiteSettings.track_inventory_by_default` to one value, then in a mutation passed the opposite value, and checked that the mutation result is as we expected. Could be also a parametrized test where we pass different combinations.

saleor/graphql/product/mutations/product_variant/product_variant_create.py
```diff
@@ -294,9 +296,20 @@ def check_for_duplicates_in_stocks(cls, stocks_data):
                 }
             )
 
+    @classmethod
+    def set_track_inventory(cls, _info, instance, cleaned_input):
+        track_inventory_by_default = get_track_inventory_by_default()
```
korycins, line 301: Instead of calling SiteSettings.objects.all().first(), you should use `site = get_site_promise(info.context).get()`
Then you have access to site settings by making `site.settings`.
`get_site_promise` uses data loader to fetch the site settings, it means that we will fetch siteSettings only once per request (no matter if you need siteSettings in mutation or in query response). 
This will also resolve the issue in bulk mutation. You're using `get_track_inventory_by_default()` in loop. It means that you're adding additional db queryset for each variant in bulk

Smit-Parmar: @korycins Thanks for the suggestion. I've implemented the changes so basically use of data loaders is to reduce db queries . It will only query per requests. Is it something similar to caching but valid for one request only?

korycins: Yes, correct. It is a batching algorithm connected with an in-memory cache. It solves the problem of fetching the same from the database in different graphql resolvers for a single request.

saleor/graphql/product/types/products.py
```diff
@@ -276,7 +276,12 @@ class ProductVariant(ChannelContextTypeWithMetadata[models.ProductVariant]):
         description="The product to which the variant belongs.",
     )
     track_inventory = graphene.Boolean(
-        required=True, description="Whether the inventory of the variant is tracked."
+        required=True,
+        description=(
+            "Determines if the inventory of this variant should be tracked. If false, "
+            "the quantity won't change when customers buy this item."
```
korycins, line 282: Missing space. It should be `but this item. "` (see the description generated to `schema.graphql` the space there is missing.

Smit-Parmar: @korycins Please review.

korycins (approved):

saleor/graphql/product/bulk_mutations/product_variant_bulk_create.py
```diff
@@ -864,7 +865,13 @@ def save_variants(cls, variants_data_with_errors_list, product):
...
             if not variant:
                 continue
-
+            track_inventory_by_default = get_track_inventory_by_default(info)
+            if track_inventory_by_default is not None:
+                variant.track_inventory = (
+                    track_inventory_by_default
+                    if variant_data["cleaned_input"].get("track_inventory") is None
+                    else variant_data["cleaned_input"].get("track_inventory")
+                )
```
IKarbowiak, line 874: It could be simplified
```suggestion
                track_inventory = variant_data["cleaned_input"].get("track_inventory")
                variant.track_inventory = track_inventory or track_inventory_by_default
```

saleor/graphql/product/mutations/product_variant/product_variant_create.py
```diff
@@ -294,9 +296,20 @@ def check_for_duplicates_in_stocks(cls, stocks_data):
...
 
+    @classmethod
+    def set_track_inventory(cls, _info, instance, cleaned_input):
+        track_inventory_by_default = get_track_inventory_by_default(_info)
+        if track_inventory_by_default is not None:
+            instance.track_inventory = (
+                track_inventory_by_default
+                if cleaned_input.get("track_inventory") is None
+                else cleaned_input.get("track_inventory")
+            )
```
IKarbowiak, line 307: Same here:
```suggestion
                track_inventory = variant_data["cleaned_input"].get("track_inventory")
                variant.track_inventory = track_inventory or track_inventory_by_default
```

saleor/graphql/shop/utils.py
```diff
@@ -26,3 +27,10 @@ def get_countries_codes_list(attached_to_shipping_zones: Optional[bool] = None):
             return all_countries_codes - covered_countries_codes
 
     return all_countries_codes
+
+
+def get_track_inventory_by_default(info):
```
IKarbowiak, line 32: Could you also add tests to this function?  🙂

saleor/graphql/product/mutations/product_variant/product_variant_create.py
IKarbowiak: Oh, right. The `track_inventory` is boolean. So:
```suggestion
            track_inventory = variant_data["cleaned_input"].get("track_inventory")
            variant.track_inventory = (
                track_inventory_by_default if track_inventory is None
                else track_inventory_by_default
            )
```

Smit-Parmar: @IKarbowiak I think then there is no change needed we can directly use the `cleaned_input.get("track_inventory")` instead of storing it on `track_inventory`.

IKarbowiak: Yep, but you are doing it twice, and IMO it easier to read in a way I proposed. But it's just a suggestion. 🙂

szczecha (approved):

Issue #13315, opened 2023-07-05
Handle the sale queries - ensure it works only for old sales (5 days)


aniav:

zedzior:
Resolves in https://github.com/saleor/saleor/pull/13521

PR #13521, opened 2023-07-25
Refactor sales queries

zedzior:
I want to merge this change, because it makes `sale` and `sales` queries to read from `Promotion` model instead of `Sale`.

The queries will be only used for old sales (migrated ones and those created by by sale mutations).

This PR breaks sales mutations tests, what will be fixed in upcoming PR, which refactors sales mutations.

Note: Previous sale architecture had `type` and `catalogue` associated with `Sale` (parent node), while promotions stores the info in rules (child node). Therefore we need to ensure that:

Issue: https://github.com/saleor/saleor/issues/13315

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/discount/dataloaders.py
+++ b/saleor/graphql/discount/dataloaders.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 from typing import List, Optional
 
 from django.db.models import Exists, F, OuterRef
+from promise import Promise
 
 from ...channel.models import Channel
 from ...discount import DiscountInfo
@@ -23,6 +24,7 @@ from ...discount.utils import (
     fetch_sale_channel_listings,
     fetch_variants,
 )
+from ..channel.dataloaders import ChannelBySlugLoader
 from ..core.dataloaders import DataLoader
 
 
@@ -330,3 +332,108 @@ class PromotionByRuleIdLoader(DataLoader):
         )
         promotion_map = {rule.id: promotions.get(rule.promotion_id) for rule in rules}
         return [promotion_map.get(rule_id) for rule_id in keys]
+
+
+class SaleChannelListingByPromotionIdLoader(DataLoader):
+    context_key = "sale_channel_listing_by_promotion_id"
+
+    def batch_load(self, keys):
+        from .types.sales import SaleChannelListing
+
+        def with_rules(rules):
+            rule_ids = [rule.id for item in rules for rule in item]
+
+            def with_channels(channels):
+                rule_channels = dict(zip(rule_ids, channels))
+                promotion_listing_map = defaultdict(list)
+                for promotion_id, promotion_rules in zip(keys, rules):
+                    for rule in promotion_rules:
+                        channels = rule_channels[rule.id]
+                        for channel in channels:
+                            promotion_listing_map[promotion_id].append(
+                                SaleChannelListing(
+                                    id=rule.old_channel_listing_id,
+                                    channel=channel,
+                                    discount_value=rule.reward_value,
+                                    currency=channel.currency_code,
+                                )
+                            )
+                return [promotion_listing_map[key] for key in keys]
+
+            return (
+                ChannelsByPromotionRuleIdLoader(self.context)
+                .load_many(rule_ids)
+                .then(with_channels)
+            )
+
+        return (
+            PromotionRulesByPromotionIdLoader(self.context)
+            .load_many(keys)
+            .then(with_rules)
+        )
+
+
+class PromotionRulesByPromotionIdAndChannelSlugLoader(DataLoader):
+    context_key = "promotion_rules_by_promotion_id_and_channel_slug"
+
+    def batch_load(self, keys):
+        promotion_ids = [key[0] for key in keys]
+        channel_slug = keys[0][1]
+        channel = ChannelBySlugLoader(self.context).load(channel_slug)
+
+        def with_channel(data):
+            channel, promotion_ids = data
+            promotions = Promotion.objects.using(self.database_connection_name).filter(
+                id__in=promotion_ids
+            )
+            PromotionRuleChannel = PromotionRule.channels.through
+            rule_channels = PromotionRuleChannel.objects.using(
+                self.database_connection_name
+            ).filter(channel_id=channel.id)
+
+            rules = PromotionRule.objects.using(self.database_connection_name).filter(
+                Exists(promotions.filter(id=OuterRef("promotion_id"))),
+                Exists(rule_channels.filter(promotionrule_id=OuterRef("id"))),
+            )
+
+            promotion_rule_map = defaultdict(list)
+            for rule in rules:
+                promotion_rule_map[rule.promotion_id].append(rule)
+
+            return [promotion_rule_map[promotion_id] for promotion_id, _ in keys]
+
+        return Promise.all([channel, promotion_ids]).then(with_channel)
+
+
+class PredicateByPromotionIdLoader(DataLoader):
+    context_key = "predicate_by_promotion_id_and_channel_slug"
+
+    def batch_load(self, keys):
+        def with_rules(rules):
+            from .utils import convert_migrated_sale_catalogue_predicate
+
+            rules = [rule for item in rules for rule in item]
+            promotion_predicated_map = defaultdict(list)
+            for rule in rules:
+                converted_predicate = convert_migrated_sale_catalogue_predicate(
+                    rule.catalogue_predicate
+                )
+                promotion_predicated_map[rule.promotion_id].append(converted_predicate)
+
+            promotion_merged_predicated_map = {}
+            for promotion_id, predicates in promotion_predicated_map.items():
+                merged_predicates: dict = defaultdict(list)
+                for predicate in predicates:
+                    if not predicate:
+                        continue
+                    for key, ids in predicate.items():
+                        merged_predicates[key].extend(ids)
+                promotion_merged_predicated_map[promotion_id] = merged_predicates
+
+            return [promotion_merged_predicated_map[key] for key in keys]
+
+        return (
+            PromotionRulesByPromotionIdLoader(self.context)
+            .load_many(keys)
+            .then(with_rules)
+        )
--- a/saleor/graphql/discount/resolvers.py
+++ b/saleor/graphql/discount/resolvers.py
@@ -1,5 +1,6 @@
-from django.db.models import QuerySet
+from django.db.models import Exists, OuterRef, QuerySet
 
+from ...channel.models import Channel
 from ...discount import models
 from ..channel import ChannelContext, ChannelQsContext
 from .filters import filter_sale_search, filter_voucher_search
@@ -23,14 +24,21 @@ def resolve_vouchers(info, channel_slug, **kwargs) -> ChannelQsContext:
 
 
 def resolve_sale(id, channel):
-    sale = models.Sale.objects.filter(id=id).first()
-    return ChannelContext(node=sale, channel_slug=channel) if sale else None
+    promotion = models.Promotion.objects.filter(old_sale_id=id).first()
+    return ChannelContext(node=promotion, channel_slug=channel) if promotion else None
 
 
-def resolve_sales(info, channel_slug, **kwargs) -> ChannelQsContext:
-    qs = models.Sale.objects.all()
+def resolve_sales(_info, channel_slug, **kwargs) -> ChannelQsContext:
+    qs = models.Promotion.objects.filter(old_sale_id__isnull=False)
     if channel_slug:
-        qs = qs.filter(channel_listings__channel__slug=channel_slug)
+        channel = Channel.objects.filter(slug=channel_slug)
+        rule_channel = models.PromotionRule.channels.through.objects.filter(
+            channel__in=channel
+        )
+        rules = models.PromotionRule.objects.filter(
+            Exists(rule_channel.filter(promotionrule_id=OuterRef("id")))
+        )
+        qs = qs.filter(Exists(rules.filter(promotion_id=OuterRef("pk"))))
 
     # DEPRECATED: remove filtering by `query` argument when it's removed from the schema
     if query := kwargs.get("query"):
--- a/saleor/graphql/discount/sorters.py
+++ b/saleor/graphql/discount/sorters.py
@@ -1,5 +1,5 @@
 import graphene
-from django.db.models import Min, Q, QuerySet
+from django.db.models import F, Min, Q, QuerySet
 
 from ..core.descriptions import CHANNEL_REQUIRED
 from ..core.doc_category import DOC_CATEGORY_DISCOUNTS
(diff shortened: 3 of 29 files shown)


saleor/graphql/discount/resolvers.py
```diff
@@ -23,16 +23,19 @@ def resolve_vouchers(info, channel_slug, **kwargs) -> ChannelQsContext:
 
 
 def resolve_sale(id, channel):
-    sale = models.Sale.objects.filter(id=id).first()
-    return ChannelContext(node=sale, channel_slug=channel) if sale else None
+    # TODO what about channel
+    # TODO use dataloaders
+    promotion = models.Promotion.objects.filter(old_sale_id=id).first()
+    return ChannelContext(node=promotion, channel_slug=channel) if promotion else None
```
zedzior, line 29: Should we still return `ChannelContext` ?
Do we need dataloaders here?

saleor/graphql/discount/types/sales.py
```diff
@@ -121,81 +112,101 @@ class Meta:
...
+        if rule := root.node.rules.first():
+            if channel := rule.channels.first():
+                return [
+                    SaleChannelListing(
+                        # TODO what about id???
+                        id=graphene.Node.to_global_id("SaleChannelListing", rule.id),
+                        channel=channel,
+                        discount_value=rule.reward_value,
+                        currency=channel.currency_code,
+                    )
```
zedzior, line 153: We will not have `PromotionChannelListing`, so it is kind of imitation, but what can we return here as `id` to not have a breaking change?

saleor/discount/migrations/0049_migrate_sales_to_promotions.py
```diff
@@ -1,231 +1,10 @@
 # Generated by Django 3.2.18 on 2023-07-06 09:43
-from dataclasses import dataclass
-from typing import List
-
-import graphene
 from django.db import migrations
+from saleor.discount.tasks import SaleToPromotionConverter
```
IKarbowiak, line 3: We shouldn't importing anything from the code, as we'll be not able to remove it. We should copy the function to the migration code

IKarbowiak (changes requested):

saleor/graphql/discount/resolvers.py
```diff
@@ -23,16 +23,19 @@ def resolve_vouchers(info, channel_slug, **kwargs) -> ChannelQsContext:
 
 
 def resolve_sale(id, channel):
-    sale = models.Sale.objects.filter(id=id).first()
-    return ChannelContext(node=sale, channel_slug=channel) if sale else None
+    # TODO what about channel
+    # TODO use dataloaders
```
fowczarek, line 27: What about those TODO's?

saleor/graphql/discount/tests/queries/test_sales.py
```diff
@@ -1,13 +1,17 @@
+from saleor.discount.tasks import SaleToPromotionConverter
```
fowczarek, line 1: Invalid import type.

saleor/graphql/discount/tests/queries/test_sales.py
```diff
@@ -108,6 +122,7 @@ def test_sales_query(
     channel_USD,
     permission_manage_products,
 ):
+    SaleToPromotionConverter.convert_sales_to_promotions()
```
fowczarek, line 125: I don't think we should use `SaleToPromotionConverter` to convert any data in tests. Currently, tests create sales and save them to the database. In the next steps, we convert nearly create sales to promotions. We should fix it in all tests.

saleor/graphql/discount/types/sales.py
```diff
@@ -52,12 +47,8 @@ class SaleChannelListing(ModelObjectType[models.SaleChannelListing]):
 
     class Meta:
         description = "Represents sale channel listing."
-        model = models.SaleChannelListing
         interfaces = [relay.Node]
-
-    @staticmethod
-    def resolve_channel(root: models.SaleChannelListing, info: ResolveInfo):
-        return ChannelByIdLoader(info.context).load(root.channel_id)
```
fowczarek, line 60: Why did we drop this?

saleor/graphql/discount/types/sales.py
```diff
@@ -121,81 +112,101 @@ class Meta:
...
-        return SaleChannelListingBySaleIdLoader(info.context).load(root.node.id)
+    def resolve_channel_listings(
+        root: ChannelContext[models.Promotion], info: ResolveInfo
+    ):
+        if rule := root.node.rules.first():
+            if channel := rule.channels.first():
+                return [
+                    SaleChannelListing(
+                        # TODO what about id???
+                        id=graphene.Node.to_global_id("SaleChannelListing", rule.id),
```
fowczarek, line 149: We should return the old IDs here. If someone has integration based on those IDs there we shouldn't break previous assumptions. 
Note:
`rule.id != sale_channel_listing.id`

saleor/graphql/discount/types/sales.py
```diff
@@ -121,81 +112,101 @@ class Meta:
...
+                        channel=channel,
+                        discount_value=rule.reward_value,
+                        currency=channel.currency_code,
+                    )
+                ]
 
     @staticmethod
     def resolve_collections(
-        root: ChannelContext[models.Sale], info: ResolveInfo, **kwargs
+        root: ChannelContext[models.Promotion], info: ResolveInfo, **kwargs
```
fowczarek, line 158: I'm not sure about typing in sales resolver.
What do we have in the `root` object? 
`ChennelContext` is required only for resolving channel-specific files e.g. `pricing` on the Product level.

saleor/graphql/discount/types/sales.py
```diff
@@ -121,81 +112,101 @@ class Meta:
...
     def resolve_products(
-        root: ChannelContext[models.Sale], info: ResolveInfo, **kwargs
+        root: ChannelContext[models.Promotion], info: ResolveInfo, **kwargs
     ):
-        qs = root.node.products.all()
-        qs = ChannelQsContext(qs=qs, channel_slug=root.channel_slug)
-        return create_connection_slice(qs, info, kwargs, ProductCountableConnection)
+        from ..utils import convert_migrated_sale_catalogue_predicate
+
+        if predicates := convert_migrated_sale_catalogue_predicate(root.node):
```
fowczarek, line 176: This function causes additional database queries. We should fetch required data via dataloader and pass it to any required functions inside resolvers.

saleor/graphql/discount/types/sales.py
```diff
@@ -121,81 +112,101 @@ class Meta:
...
 
     @staticmethod
     def resolve_collections(
-        root: ChannelContext[models.Sale], info: ResolveInfo, **kwargs
+        root: ChannelContext[models.Promotion], info: ResolveInfo, **kwargs
     ):
-        qs = root.node.collections.all()
-        qs = ChannelQsContext(qs=qs, channel_slug=root.channel_slug)
-        return create_connection_slice(qs, info, kwargs, CollectionCountableConnection)
+        from ..utils import convert_migrated_sale_catalogue_predicate
```
fowczarek, line 160: Could we fix it to not import this function in many resolvers?

saleor/graphql/discount/types/sales.py
```diff
@@ -121,81 +112,101 @@ class Meta:
...
-            .load((root.node.id, root.channel_slug))
-            .then(
-                lambda channel_listing: channel_listing.currency
-                if channel_listing
-                else None
-            )
-        )
+    def resolve_currency(root: ChannelContext[models.Promotion], _info: ResolveInfo):
+        if rule := root.node.rules.first():
+            if channel := rule.channels.first():
```
fowczarek, line 208: Missing dataloaders.

saleor/graphql/discount/types/sales.py
```diff
@@ -121,81 +112,101 @@ class Meta:
...
-            .then(
-                lambda channel_listing: channel_listing.discount_value
-                if channel_listing
-                else None
-            )
-        )
+    def resolve_discount_value(
+        root: ChannelContext[models.Promotion], _info: ResolveInfo
+    ):
+        if rule := root.node.rules.first():
```
fowczarek, line 202: Missing dataloader.

fowczarek (approved):

saleor/graphql/discount/types/sales.py
IKarbowiak: That's true. So it means that we will have to also keeps old sale channel listings ids somewhere in promotion models

saleor/graphql/discount/types/sales.py
zedzior: ChannelContext dropped

saleor/graphql/discount/types/sales.py
zedzior: There will be no related `PromotionChannelListing` model, so we need to kind of imitate it.

saleor/discount/tasks.py
```diff
@@ -94,3 +104,237 @@ def get_sales_to_notify_about():
         )
     ).distinct()
     return sales
+
+
+class SaleToPromotionConverter:
```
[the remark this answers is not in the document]
zedzior: I moved it to separate file. What is wrong with keeping it in class?

saleor/graphql/discount/types/sales.py
zedzior: I will add old sale channel listings ids to promotion models.

saleor/graphql/discount/resolvers.py
zedzior: It breaks over 100 tests, let do this in separate PR

saleor/discount/migrations/0049_migrate_sales_to_promotions.py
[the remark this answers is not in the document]
zedzior: Moved to https://github.com/saleor/saleor/pull/13553

saleor/discount/models.py
```diff
@@ -398,6 +398,9 @@ def is_active(self, date=None):
             date = datetime.now(pytz.utc)
         return (not self.end_date or self.end_date >= date) and self.start_date <= date
 
+    def __str__(self):
+        return self.name
+
```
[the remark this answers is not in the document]
zedzior: I used it for debugging purpose. Deleted.

saleor/graphql/discount/resolvers.py
zedzior: UPDATE: we cannot drop the channel context, because it used by `sale.currency` and `sale.discount_value` queries

saleor/graphql/discount/sorters.py
```diff
@@ -11,7 +11,7 @@ class SaleSortField(BaseEnum):
     START_DATE = ["start_date", "name", "pk"]
     END_DATE = ["end_date", "name", "pk"]
     VALUE = ["value", "name", "pk"]
-    TYPE = ["type", "name", "pk"]
+    TYPE = ["rules__reward_value_type", "name", "pk"]
```
IKarbowiak, line 14: It should be done in similar wat as sorting by `value`. We should have a dedicated function for that

saleor/graphql/discount/types/sales.py
```diff
@@ -127,81 +125,112 @@ class Meta:
...
+            predicates := PredicateByPromotionIdLoader(info.context)
+            .load(root.node.id)
+            .get()
+        ):
+            if collection_ids := predicates.get("collectionPredicate"):
+                qs = Collection.objects.filter(id__in=collection_ids)
+                qs = ChannelQsContext(qs=qs, channel_slug=root.channel_slug)
+                return create_connection_slice(
+                    qs, info, kwargs, CollectionCountableConnection
+                )
```
IKarbowiak, line 179: I think we should use then that will be calling the helper function like `return PredicateByPromotionIdLoader(info.context).load(root.node.id).then(_resolve_predicate)`.
I think that `get` might resolves the dataloader immediately. 
So IMO it should looks like:
```suggestion
       def _resolve_predicate(predicate):
            if not predicate:
                return None
            if collection_ids := predicates.get("collectionPredicate"):
                qs = Collection.objects.filter(id__in=collection_ids)
                qs = ChannelQsContext(qs=qs, channel_slug=root.channel_slug)
                return create_connection_slice(
                    qs, info, kwargs, CollectionCountableConnection
                )
        return PredicateByPromotionIdLoader(info.context).load(root.node.id).then(_resolve_predicate)
```

IKarbowiak: Same for `products`, `variants`, `categories`

saleor/graphql/discount/types/sales.py
```diff
@@ -127,81 +125,112 @@ class Meta:
...
+        if rules := (
+            PromotionRulesByPromotionIdAndChannelSlugLoader(info.context)
             .load((root.node.id, root.channel_slug))
-            .then(
-                lambda channel_listing: channel_listing.discount_value
-                if channel_listing
-                else None
-            )
-        )
+            .get()
```
IKarbowiak, line 223: Same here. We should use `then`

saleor/graphql/discount/dataloaders.py
```diff
@@ -330,3 +332,101 @@ def batch_load(self, keys):
...
+        channel_slug = keys[0][1]
+        rules = PromotionRulesByPromotionIdLoader(self.context).load_many(promotion_ids)
+        channel = ChannelBySlugLoader(self.context).load(channel_slug)
+
+        def with_rules_and_channel(data):
+            rules, channel = data
+            rule_ids = [rule.id for item in rules for rule in item]
+            rules = PromotionRule.objects.prefetch_related("channels").filter(
+                id__in=rule_ids, channels__id=channel.id
+            )
```
IKarbowiak, line 390: It such scenario it's no reason for using `PromotionRulesByPromotionIdLoader` and `ChannelBySlugLoader` for the start 😄 .

Probably we should use here the `through` model between `PromotionRule - Channel` to get the matching objects, instead.

saleor/graphql/discount/types/sales.py
IKarbowiak: Also, `PredicateByPromotionIdLoader` does not take channel into consideration but it should. IMO we should have datalaoder only for fetching promotion rule for given promotion and channel id, and use helper function for chaning predicate into different structure and then fetch the interesting data

zedzior: I agree with using `then`. However:

1. we can't fetch the data with `channel`, because it would be breaking change
2. I would also keep the "helper functionality" in dataloader, because this way we call it once instead of four times.

saleor/graphql/discount/dataloaders.py
zedzior: refactored

saleor/graphql/discount/types/sales.py
```diff
@@ -127,80 +125,132 @@ class Meta:
...
     @staticmethod
     def resolve_categories(
-        root: ChannelContext[models.Sale], info: ResolveInfo, **kwargs
+        root: ChannelContext[models.Promotion], info: ResolveInfo, **kwargs
     ):
-        qs = root.node.categories.all()
-        return create_connection_slice(qs, info, kwargs, CategoryCountableConnection)
+        def _get_categories(predicates):
+            if category_ids := predicates.get("categoryPredicate"):
+                qs = Category.objects.filter(id__in=category_ids)
```
korycins, line 153: Shouldn't be dataloader?

korycins: Same question for `Collection`, `Product` and `Variant` 👇

korycins (approved):

saleor/graphql/discount/types/sales.py
zedzior: I need to have queryset here in order to pass it to `create_connection_slice` function.

michal-macioszczyk (approved):

kadewu (approved):

PR #13532, opened 2023-07-26
Add AttributeBulkUpdate mutation

### LINKED ISSUE #13375: Add AttributeBulkUpdate mutation

SzymJ:
I want to merge this change because it adds `AttributeBulkUpdate` mutation.

Notes:
1. We decided to use the same inputs as in single mutations that's why `AttributeBulkCreateInput` and `AttributeBulkCreateInput` were changed to `AttributeCreateInput`
2. Deprecated fields from those inputs are not allowed in bulk mutations. An error will be returned if any deprecated field is used in bulk input.

Part of #13373
Resolves #13375

# Impact

* [ ] New migrations

# Pull Request Checklist


saleor/graphql/attribute/mutations/attribute_update.py
```diff
@@ -69,6 +69,12 @@ class AttributeUpdateInput(BaseInputObjectType):
 
     class Meta:
         doc_category = DOC_CATEGORY_ATTRIBUTES
+        description = (
+            "Represents an input for update of attribute.\n\n"
+            "NOTE: Deprecated fields 'filterableInStorefront', "
+            "'storefrontSearchPosition' and 'availableInGrid` are not supported in "
+            "bulk mutation."
```
maarcingebala, line 76:
```suggestion
            "NOTE: Deprecated fields `filterableInStorefront`, "
            "`storefrontSearchPosition` and `availableInGrid` are not supported in "
            "bulk mutations: `attributeBulkCreate`, `attributeBulkUpdate`."
```

saleor/graphql/attribute/mutations/attribute_bulk_update.py
```diff
@@ -0,0 +1,604 @@
...
+                description="An attribute value was created.",
+            ),
+            WebhookEventInfo(
+                type=WebhookEventAsyncType.ATTRIBUTE_VALUE_DELETED,
+                description="An attribute value was deleted.",
+            ),
+        ]
+
+    @classmethod
+    def clean_attributes(
```
IKarbowiak, line 120: Could we split it into smaller methods?

saleor/graphql/attribute/mutations/attribute_bulk_update.py
IKarbowiak, line 1: For `AttributeBulkUpdate` we are missing few tests as well

saleor/graphql/schema.graphql
```diff
@@ -17448,13 +17448,33 @@ type Mutation {
...
 
+  """
+  Updates attributes.
+  
+  Added in Saleor 3.15.
+  
+  Note: this API is currently in Feature Preview and can be subject to changes at later point.
+  
+  Triggers the following webhook events:
+  - ATTRIBUTE_UPDATED (async): An attribute was updated.
```
korycins, line 17466: These webhooks are called in specific conditions, right? The description should contain this info. 
Something like:
`An attribute value was created.` -> 
`Optionally called when  new attribute value was created.` 
(Same for delete)

jakubkuc (approved):

szczecha:
IntegrityError when creating an attribute value with an existing value name

1. Create an attribute 
2. Update attribute by adding new value (use existing value name) 

Result: IntegrityError. If I try to add the same value to an attribute by attributeValueCreate mutation it creates a new value with the same name as the existing one but a different slug    
Screenshot 2023-07-31 at 15 16 02
Screenshot 2023-07-31 at 15 15 51

szczecha:
REJECT_FAILED_ROWS throws IntegrityError while REJECT_EVERYTHING shows max length error

1. Create attribute
2. Try to update attribute by adding values first with too-long name and second with valid name using REJECT_EVERYTHING Policy 
Result: "code": "MAX_LENGTH"
3. Try to update attribute by adding values first with too-long name and second with valid name using REJECT_FAILED_ROWS Policy 
Result: IntegrityError 

Screenshot 2023-08-01 at 10 08 40
Screenshot 2023-08-01 at 10 08 23

IKarbowiak (approved):

maarcingebala (approved):

szczecha (approved):

saleor/graphql/attribute/mutations/attribute_bulk_update.py
```diff
@@ -0,0 +1,636 @@
...
+                    {"instance": None, "errors": index_error_map[index]}
+                )
+                continue
+
+            attr = cleaned_input["instance"]
+            fields = cleaned_input["fields"]
+            remove_values = fields.pop("remove_values", [])
+            add_values = fields.pop("add_values", [])
+
+            try:
```
korycins, line 420: What is a part of the data that can raise an exception here?  methods called in `if fields:`? If yes, why are wrapping everything in the try-except?
The try: except is quite big. Do we need to create a `data` structure inside it? 

What do you think about this one 👇 
```python
            if fields:
                try:
                    attr = cls.construct_instance(attr, fields)
                    cls.clean_instance(info, attr)
                except ValidationError as exc:
                    for key, errors in exc.error_dict.items():
                        for e in errors:
                            index_error_map[index].append(
                                AttributeBulkUpdateError(
                                    path=to_camel_case(key),
                                    message=e.messages[0],
                                    code=e.code,
                                )
                            )
                    instances_data_and_errors_list.append(
                        {"instance": None, "errors": index_error_map[index]}
                    )
                    continue

            data = {
                "instance": attr,
                "errors": index_error_map[index],
                "remove_values": remove_values,
                "add_values": [],
                "attribute_updated": True if fields else False,
            }
```
The scope of try-except is small. the `if fields` gives you immediately context that this exception can be raised only when `if fields`. The `data` is outside as it is not part of try-except block

korycins (commented):
What about concerns from this comment: https://github.com/saleor/saleor/pull/13532/files#r1279076947 ? I still see some coverage reports. It looks like some error handlers don't have a test. For example:
`saleor/graphql/attribute/mutations/attribute_bulk_update.py:221`
`saleor/graphql/attribute/mutations/attribute_bulk_update.py:371

saleor/graphql/attribute/mutations/attribute_bulk_update.py
SzymJ: Good idea thanks!

saleor/graphql/attribute/mutations/attribute_bulk_update.py
```diff
@@ -0,0 +1,636 @@
...
+                type=WebhookEventAsyncType.ATTRIBUTE_UPDATED,
+                description=(
+                    "An attribute was updated. "
+                    "Optionally called when  new attribute value was created or "
+                    "deleted."
+                ),
+            ),
+            WebhookEventInfo(
+                type=WebhookEventAsyncType.ATTRIBUTE_VALUE_CREATED,
+                description="An attribute value was created.",
```
korycins, line 115: You could also place here a info that the events are optionally called. Right? VALUE_CREATED will be called when you add new values, VALUE_DELETED will be called when you delete values. Both events are optionally, I think it is worth to mention.

korycins (approved):

saleor/graphql/attribute/mutations/attribute_bulk_update.py
SzymJ: Added.

PR #13539, opened 2023-07-27
Handle promotion toggle

IKarbowiak:
Get rid of `handle_sale_toggle` add `handle_promotion_toggle` task instead.

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/discount/tasks.py
+++ b/saleor/discount/tasks.py
@@ -1,96 +1,115 @@
-from collections import defaultdict
 from datetime import datetime
-from typing import Dict
 
-import graphene
 import pytz
 from celery.utils.log import get_task_logger
-from django.db.models import F, Q
+from django.db.models import Exists, F, OuterRef, Q, QuerySet
 
 from ..celeryconf import app
-from ..graphql.discount.mutations.utils import CATALOGUE_FIELD_TO_TYPE_NAME
+from ..graphql.discount.utils import get_variants_for_predicate
 from ..plugins.manager import get_plugins_manager
-from ..product.tasks import update_products_discounted_prices_of_catalogues_task
-from .models import Sale
-from .utils import CATALOGUE_FIELDS, CatalogueInfo
+from ..product.models import Product, ProductVariant
+from ..product.tasks import update_products_discounted_prices_for_promotion_task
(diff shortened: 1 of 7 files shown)


fowczarek (approved):

zedzior (approved):

saleor/settings.py
```diff
@@ -589,9 +589,9 @@ def get_host():
         "task": "saleor.csv.tasks.delete_old_export_files",
         "schedule": crontab(hour=1, minute=0),
     },
-    "handle-sale-toggle": {
-        "task": "saleor.discount.tasks.handle_sale_toggle",
-        "schedule": initiated_sale_webhook_schedule,
+    "handle-promotion-toggle": {
```
korycins, line 592: I am not convinced that we should replace the sale-toggle in the current way. 
I would expect that sale webhooks will be deprecated and removed in 4.0. I would also expect that all of my apps subscribed to the sale's webhooks will be receiving the events in the same form as previously. In that case, I would have time till 4.0 to apply changes in my apps and add implementation to support promotion events.
Stoping sending the Sale events sounds like a not intended breaking change 🤔

IKarbowiak: Fixed

saleor/discount/tasks.py
```diff
@@ -1,96 +1,152 @@
...
+    )
 
-    if catalogue_infos:
+    for staring_promo in staring_promotions:
+        manager.promotion_started(staring_promo)
+
+    for ending_promo in ending_promotions:
+        manager.promotion_ended(ending_promo)
+
+    # DEPRECATED: will be removed in Saleor 4.0.
```
korycins, line 49: Maybe it would be worth adding a gh issue about this, and including to [Saleor4.0 mileston](https://github.com/saleor/saleor/milestone/80)?

korycins (approved):

saleor/discount/tasks.py
IKarbowiak: Added: https://github.com/saleor/saleor/issues/13573#issue-1828825452

zedzior (approved):

Issue #13384, opened 2023-07-12
[RFC] Programmatic payments - PaymentMethod API


korycins:
## Problem
Unable to fetch a list of stored payment methods. That can be used to pay for the checkout or order.

---

## 1. General Assumptions
We will add new sync webhook that will allow external App to provide the stored payment
methods.

### Acceptance Criteria

- [ ] Saleor has new sync webhook `LIST_STORED_PAYMENT_METHODS`
- [ ] Webhook is called when user request `checkout.storedPaymentMethods` field
- [ ] Webhook is called when user request `user.storedPaymentMethods` field
- [ ] stored payment method can be fetched from `User` type and `Checkout` type

### Status

```[tasklist]
## PaymentMethod API
- [ ] Add `LIST_STORED_PAYMENT_METHODS` webhook - https://github.com/saleor/saleor/pull/13491
- [ ] Add PaymentMethod types - https://github.com/saleor/saleor/pull/13541
- [ ] Add `User.storedPaymentMethod` query - https://github.com/saleor/saleor/pull/13541
- [ ] Add `Checkout.storedPaymentMethod` query - https://github.com/saleor/saleor/pull/13541
- [ ] Update the docs after introducing the changes - https://github.com/saleor/saleor-docs/pull/852
- [ ] make sure ADDED_IN labels are correct before merging feature branch to main
```
#### PaymentMethod
```graphql

"""
List of available flows that can be used to process payment with this payment method
"""
enum TokenizedPaymentFlowEnum {
  INTERACTIVE
}

"""
Represents a payment source stored for user in payment gateway, such as credit card.

@deprecated in Saleor 3.xx - use PaymentMethod instead
"""
type PaymentSource {
  # ...
}

type PaymentGateway {
  name: String!
  id: String!
  config: [GatewayConfigLine!]!
  currencies: [String!]!
}

"""
Represents a payment method stored for user (tokenized) in payment gateway, such as credit card or SEPA direct debit

Can be accessed only by user the PaymentMethod is attached to
or by staff members and app with permission `HANDLE_CHARGE`
"""
type PaymentMethod @doc(category: "Payments") {
  """Payment gateway that stores this payment method"""
  gateway: PaymentGateway!

  """
  ID of stored payment method used to make charges
  Note: method ID is unique only within the payment gateway
  """
  paymentMethodId: String!

  """
  Unique ID of payment method (gateway.id + paymentMethodId)
  """
  id: ID!

  """Stored credit card details if available."""
[...]

PR #13541, opened 2023-07-27
[Programmatic payments] storedPaymentMethods field in `Checkout` and `User` type

korycins:
I want to merge this change because it adds a new field to `Checkout` and `User` type. The field  returns a list of stored payment methods fetched by `list-stored-payment-methods` webhook. 
More details: https://github.com/saleor/saleor/issues/13384

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/types.py
+++ b/saleor/graphql/account/types.py
@@ -393,7 +393,10 @@ class User(ModelObjectType[models.User]):
     )
     stored_payment_sources = NonNullList(
         "saleor.graphql.payment.types.PaymentSource",
-        description="List of stored payment sources.",
+        description=(
+            "List of stored payment sources. The field returns a list of payment "
+            "sources stored for payment plugins."
+        ),
         channel=graphene.String(
             description="Slug of a channel for which the data should be returned."
         ),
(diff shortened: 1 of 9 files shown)


kadewu (approved):

saleor/graphql/account/types.py
```diff
@@ -415,6 +421,21 @@ class User(ModelObjectType[models.User]):
...
+            "provided channel. When `amount` is not provided, 0 will be used as "
+            "default value." + ADDED_IN_315 + PREVIEW_FEATURE
+        ),
+        channel=graphene.String(
+            description="Slug of a channel for which the data should be returned.",
+            required=True,
+        ),
+        amount=PositiveDecimal(
+            description="Amount that will be used to fetch stored payment methods."
+        ),
```
maarcingebala, line 437: What is the purpose of passing `amount` when resolving the list of saved payment methods? Maybe it should be briefly explained in the description, as it's unclear now.

saleor/graphql/account/types.py
```diff
@@ -415,6 +421,21 @@ class User(ModelObjectType[models.User]):
         required=True,
         description="The data when the user last update the account information.",
     )
+    stored_payment_methods = NonNullList(
```
maarcingebala, line 424: What about the `stored_payment_sources` field? What's the difference between these fields?

saleor/graphql/account/types.py
korycins: Some stored payment methods could not be allowed when you want to use is with amount less or greater than X.
I will fix the description

saleor/graphql/account/types.py
korycins: stored_payment_sources is dedicated to the PaymentPlugin and will be deprecated with PaymentPlugins in the future.
Also, we didn't want to use it as it would mean that we would have to have a single type that would need to work for previous solutions and the new one - it means, that new fields which are currently - `required` would need to be optional, or fulfilled with placeholder data (if even possible). 
Providing a new interface, for a new flow is much more cleaner and flexible than trying to use the old one.

IKarbowiak (approved):

saleor/graphql/account/types.py
maarcingebala: Maybe that should also be mentioned in the description, these two fields seem nearly identical for the users.

Issue #13486, opened 2023-07-20
Add `ACCOUNT_SET_PASSWORD_REQUESTED` webhook


maarcingebala:
Add the `ACCOUNT_SET_PASSWORD_REQUESTED` webhook:
- should be triggered when:
  - customer users request resetting their password in the `requestPasswordReset` mutation
  - a new customer is created using the `customerCreate` mutation
- required permission: `MANAGE_USERS`

Payload:
```graphql
subscription {
  event {
    ... on AccountSetPasswordRequested {
      redirectUrl: String!  # link to a view to set password
      user: User!
      channel: Channel
      token: String!  # one-time token required to set password
    }
  }
}
```

Note: this event replaces the two notification events: 
- `NotifyEventType.ACCOUNT_PASSWORD_RESET`
- `NotifyEventType.ACCOUNT_SET_CUSTOMER_PASSWORD`

With the new webhook, we don't differentiate between cases of setting or resetting the password as they both lead to the same mutation `setPassword`. If needed, we can extend the payload later with a flag informing which of these two cases it is.

PR #13555, opened 2023-07-28
Add `ACCOUNT_SET_PASSWORD_REQUESTED` webhook event

Air-t:
Adds `ACCOUNT_SET_PASSWORD_REQUESTED` event sent after `requestPasswordReset` or `customerCreate` mutation.

Resolves https://github.com/saleor/saleor/issues/13486

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/mutations/authentication/request_password_reset.py
+++ b/saleor/graphql/account/mutations/authentication/request_password_reset.py
@@ -1,12 +1,15 @@
+from urllib.parse import urlencode
+
 import graphene
 from django.conf import settings
+from django.contrib.auth.tokens import default_token_generator
 from django.core.exceptions import ValidationError
 from django.utils import timezone
 
 from .....account.error_codes import AccountErrorCode
 from .....account.notifications import send_password_reset_notification
 from .....account.utils import retrieve_user_by_email
-from .....core.utils.url import validate_storefront_url
+from .....core.utils.url import prepare_url, validate_storefront_url
 from .....webhook.event_types import WebhookEventAsyncType
 from ....channel.utils import clean_channel, validate_channel
 from ....core import ResolveInfo
@@ -47,7 +50,11 @@ class RequestPasswordReset(BaseMutation):
             WebhookEventInfo(
                 type=WebhookEventAsyncType.NOTIFY_USER,
                 description="A notification for password reset.",
-            )
+            ),
+            WebhookEventInfo(
+                type=WebhookEventAsyncType.ACCOUNT_SET_PASSWORD_REQUESTED,
+                description="Setting a new password for the account is requested.",
+            ),
         ]
 
     @classmethod
@@ -99,8 +106,10 @@ class RequestPasswordReset(BaseMutation):
     def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
         email = data["email"]
         redirect_url = data["redirect_url"]
-        channel_slug = data.get("channel")
         user = cls.clean_user(email, redirect_url, info)
+        channel_slug = data.get("channel")
+        token = default_token_generator.make_token(user)
+        params = urlencode({"email": user.email, "token": token})
 
         if not user.is_staff:
             channel_slug = clean_channel(
@@ -110,6 +119,7 @@ class RequestPasswordReset(BaseMutation):
             channel_slug = validate_channel(
                 channel_slug, error_class=AccountErrorCode
             ).slug
+
         manager = get_plugin_manager_promise(info.context).get()
         send_password_reset_notification(
             redirect_url,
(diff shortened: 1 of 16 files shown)


saleor/graphql/account/mutations/authentication/request_password_reset.py
```diff
@@ -99,25 +106,41 @@ def clean_user(cls, email, redirect_url, info: ResolveInfo):
     def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
         email = data["email"]
         redirect_url = data["redirect_url"]
-        channel_slug = data.get("channel")
         user = cls.clean_user(email, redirect_url, info)
+        user.last_password_reset_request = timezone.now()
+        user.save(update_fields=["last_password_reset_request"])
```
maarcingebala, line 111: I'm not sure if we should save `last_password_reset_request` here; we run some validations in `post_save_action`, so wouldn't that lead to a situation where the new timestamp is saved, and later we get a validation error, which would effectively not send the notification?

saleor/graphql/webhook/enums.py
```diff
@@ -41,6 +41,9 @@
     WebhookEventAsyncType.ACCOUNT_CHANGE_EMAIL_REQUESTED: (
         "Account email change is requested."
     ),
+    WebhookEventAsyncType.ACCOUNT_SET_PASSWORD_REQUESTED: (
+        "Account set password is requested."
```
maarcingebala, line 45: It sounds unnatural, I would rephrase it to something like "Setting a new password for the account is requested."

saleor/graphql/webhook/subscription_types.py
```diff
@@ -214,6 +214,14 @@ def resolve_new_email(root, _info: ResolveInfo):
         return data["new_email"]
 
 
+class AccountSetPasswordRequested(SubscriptionObjectType, AccountOperationBase):
+    class Meta:
+        root_type = None
+        enable_dry_run = False
+        interfaces = (Event,)
+        description = "Event sent when set password is requested." + ADDED_IN_315
```
maarcingebala, line 222: "Event sent when setting a new password is requested." maybe like this?

saleor/graphql/account/mutations/authentication/request_password_reset.py
```diff
@@ -47,7 +50,11 @@ class Meta:
             WebhookEventInfo(
                 type=WebhookEventAsyncType.NOTIFY_USER,
                 description="A notification for password reset.",
-            )
+            ),
+            WebhookEventInfo(
+                type=WebhookEventAsyncType.ACCOUNT_SET_PASSWORD_REQUESTED,
+                description="An Account set password is requested.",
```
IKarbowiak, line 56:
```suggestion
                description="A notification for account set password is requested.",
```

saleor/graphql/account/mutations/staff/customer_create.py
```diff
@@ -33,4 +33,8 @@ class Meta:
                 type=WebhookEventAsyncType.NOTIFY_USER,
                 description="A notification for setting the password.",
             ),
+            WebhookEventInfo(
+                type=WebhookEventAsyncType.ACCOUNT_SET_PASSWORD_REQUESTED,
+                description="An Account set password is requested.",
```
IKarbowiak, line 38:
```suggestion
                description="A notification for account set password is requested.",
```

IKarbowiak (approved):

saleor/graphql/account/mutations/authentication/request_password_reset.py
Air-t: If there are validations which I missed, while changing the order of sending notifications with instance save, I will need to send notifications before save is done. However this will not be post_save_actions anymore (triggered before the actual save)

Air-t: @maarcingebala I have added changes, however I'm not sure if sending a notification and triggering webhooks should still be a part of post_save_actions - where they are triggered before save. Should I change it, moving them to separate function or just to save itself?

maarcingebala: Sending the event is the main point of this mutation, not a save. So I think only the naming "post_save" is confusing here. Maybe we don't need this function at all and everything should happen perform_mutation.

maarcingebala: We're overriding `perform_mutation` anyway, and `post_save_action` is mostly useful when you want to override just that "post save" part of the flow.

Air-t: My idea was to keep additional actions in one place, however maybe for this case it does not make sense to add extra method which naming is misleading just to keep all events calls in `post_save_action` as a pattern

Air-t: anyways, I added all the logic to the `perform_mutation` method cause it seems to be more convenient

korycins (approved):

maarcingebala (approved):

maarcingebala:
@Air-t Looks good, but the CI is failing here.

szczecha (approved):

Issue #13577, opened 2023-07-31
Bug: `NOTIFY_USER` is not triggered if it was not the only one event added to the same app


Air-t:
### What are you trying to achieve?

The `NOTIFY_USER` webhook works if it is a stand alone event added to the app. If user tries to add multiple `asyncEvent`'s to the same webhook including `NOTIFY_USER` this webhook will not be triggered and no error will be produced. Other events will work as expected.

The goal is to throw an error if someone tries to create a webhook which contains `NOTIFY_USER` with any other event, or adds it as a next event for the same webhook

Mutations affected:
`webhookUpdate`, `webhookCreate`

### Steps to reproduce the problem

1. create app with permissions to `MANAGE_USERS`
2. create webhook `NOTIFY_USER` event to app with any other event (add your `targetUrl` where webhook should be sent)
3. make mutation which triggers `NOTIFY_USER` event (`accountRegister` or any other)
4. observe that `NOTIFY_USER` event should not be sent

### What did you expect to happen?

`NOTIFY_USER` failure remains silent - there is no sign that event was not sent, besides it isn't sent where it should. This may lead to confusion.

### Environment

Saleor version: 3.15

PR #13584, opened 2023-08-01
Fix `NOTIFY_USER` with another event issue

Air-t:
* This fixes the issue when too `NOTIFY_USER` event is added
* with any other event to the same webhook. In this case
* `NOTIFY_USER` will not be triggered. Fix prevents to save
* multiple events along with `NOTIFY_USER` event.

Resolves https://github.com/saleor/saleor/issues/13577

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/webhook/mixins.py
+++ b/saleor/graphql/webhook/mixins.py
@@ -1,0 +1,21 @@
+from django.core.exceptions import ValidationError
+
+from ...webhook.error_codes import WebhookErrorCode
+from ...webhook.event_types import WebhookEventAsyncType
+
+
+class NotifyUserEventValidationMixin:
+    @classmethod
+    def validate_events(cls, events):
+        # NOTIFY_USER needs to be the only one event registered per webhook.
+        # This solves issue temporarily. NOTIFY_USER will be deprecated in the future.
+        if WebhookEventAsyncType.NOTIFY_USER in events and len(events) > 1:
+            raise ValidationError(
+                {
+                    "async_events": ValidationError(
+                        "The NOTIFY_USER webhook cannot be combined with other events.",
+                        code=WebhookErrorCode.INVALID_NOTIFY_WITH_SUBSCRIPTION.value,
+                    )
+                }
+            )
+        return events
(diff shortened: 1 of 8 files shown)


saleor/graphql/webhook/mixins.py
```diff
@@ -0,0 +1,21 @@
...
+class NotifyUserEventValidationMixin:
+    @classmethod
+    def validate_events(cls, events):
+        # NOTIFY_USER needs to be the only one event registered per webhook.
+        # This solves issue temporarily. NOTIFY_USER will be deprecated in the future.
+        if WebhookEventAsyncType.NOTIFY_USER in events and len(events) > 1:
+            raise ValidationError(
+                {
+                    "async_events": ValidationError(
+                        "Notify User event has to be the only one event per webhook.",
```
maarcingebala, line 16: I would rephrase a bit to something like this:
"The `NOTIFY_USER` webhook cannot be combined with other events."

saleor/graphql/webhook/mixins.py
```diff
@@ -0,0 +1,21 @@
...
+                        code=WebhookErrorCode.INVALID.value,
```
maarcingebala, line 17: I also think this is a very specific error, and providing a dedicated error code might be better. Then we could open a ticket in the dashboard to handle this specific error and provide a correct error message. Otherwise, the INVALID code may result in a too-generic error message, not clear to the user.

saleor/graphql/webhook/tests/mutations/test_webhook_create.py
```diff
@@ -322,3 +322,33 @@ def test_webhook_create_invalid_custom_headers(app_api_client):
...
+        query,
+        variables=variables,
+    )
+    content = get_graphql_content(response)
+    # then
+    data = content["data"]["webhookCreate"]
+    assert not data["webhook"]
+    error = data["errors"][0]
+    assert error["field"] == "asyncEvents"
+    assert error["message"] == (
```
maarcingebala, line 352: If we had a specific error code for this error, we could rely on checking the code and not the message in the test (which is a bit better as the code never or rarely changes).

saleor/graphql/webhook/mixins.py
Air-t: do we have any convention for bug fix error codes? Or just something like `COMBINE_FORBBIDEN` is enough?

maarcingebala: You can see examples in `error_codes.py` files, some codes are simple and some are very specific. Here I would go with something descriptive like: `INVALID_NOTIFY_WITH_SUBSCRIPTION` or `CANNOT_COMBINE_NOTIFY_AND_SUBSCRIPTION`.

8r2y5 (approved):

szczecha (approved):

PR #12284, opened 2023-03-10
Add taxes to undiscounted order prices

jakubkuc:
I want to merge this change because it fixes #12185

It adds calculating taxes for undiscounted prices for `OrderLine`

It contains:
- fix that will make new order lines have fields `undiscounted_unit_price` and `undiscounted_total_price` properly calculated gross and net values.
- Zero downtime migration with celery task that calculates and updates all OrderLines that have `tax_rate` > 0 and equal net and gross undiscounted prices.  

Acceptance criteria: 
- All queries For orders should contain proper `undiscounted_unit_price` and `undiscounted_total_price` fields 
- All new created and updated orders should also have  `undiscounted_unit_price` and `undiscounted_total_price` calculated correctly. 
 

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migration files are up to date
* [ ] The changes are tested
* [ ] GraphQL schema and type definitions are up to date
* [ ] Changes are mentioned in the changelog

PR #13585, opened 2023-08-01
Add taxes to undiscounted order prices

jakubkuc:
I want to merge this change because it is a port of #12284 but we decided to release it for Saleor `3.14` and up instead of `3.9` so it needed some extra adjustments. 

All the comments from the old PR are addressed, but It still needs a review since the base is different. 

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/tests/test_utils.py
+++ b/saleor/checkout/tests/test_utils.py
@@ -1,0 +1,52 @@
+from decimal import Decimal
+
+import pytest
+from prices import Money, TaxedMoney
+
+from ..utils import get_taxed_undiscounted_price
+
+BASE = Money("35.00", "USD")
+
+
+@pytest.mark.parametrize(
+    "price, tax_rate, prices_entered_with_tax, result",
+    [
+        # result should not be calculated but taken from price
+        (
+            TaxedMoney(gross=Money("43.74", "USD"), net=BASE),
+            Decimal(0.25),
+            False,
+            TaxedMoney(gross=Money("43.74", "USD"), net=BASE),
+        ),
+        # result should be calculated and different from price
+        (
+            TaxedMoney(gross=Money("43.74", "USD"), net=Money("36.00", "USD")),
+            Decimal(0.25),
+            False,
+            TaxedMoney(gross=Money("43.75", "USD"), net=BASE),
+        ),
+        # result should not be calculated and taken from price
+        (
+            TaxedMoney(gross=BASE, net=Money("26.26", "USD")),
+            Decimal(0.25),
+            True,
+            TaxedMoney(gross=BASE, net=Money("26.26", "USD")),
+        ),
+        # result should be calculated and different from price
+        (
+            TaxedMoney(gross=Money("36.00", "USD"), net=Money("26.26", "USD")),
+            Decimal(0.25),
+            True,
+            TaxedMoney(gross=BASE, net=Money("28.00", "USD")),
+        ),
+    ],
+)
+def test_get_taxed_undiscounted_price(price, tax_rate, prices_entered_with_tax, result):
+    result_price = get_taxed_undiscounted_price(
+        undiscounted_base_price=BASE,
+        price=price,
+        tax_rate=tax_rate,
+        prices_entered_with_tax=prices_entered_with_tax,
+    )
+
+    assert result_price == result
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -5,10 +5,11 @@ from typing import TYPE_CHECKING, Iterable, List, Optional, Tuple, Union, cast
 import graphene
 from django.core.exceptions import ValidationError
 from django.utils import timezone
-from prices import Money
+from prices import Money, TaxedMoney
 
 from ..account.models import User
 from ..core.exceptions import ProductNotPublished
+from ..core.prices import quantize_price
 from ..core.taxes import zero_taxed_money
 from ..core.utils.promo_code import (
     InvalidPromoCode,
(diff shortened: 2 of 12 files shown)


saleor/checkout/tests/test_utils.py
```diff
@@ -0,0 +1,52 @@
...
+            Decimal(0.25),
+            False,
+            TaxedMoney(gross=Money("43.74", "USD"), net=BASE),
+        ),
+        # result should be calculated and different from price
+        (
+            TaxedMoney(gross=Money("43.74", "USD"), net=Money("36.00", "USD")),
+            Decimal(0.25),
+            False,
+            TaxedMoney(gross=Money("43.75", "USD"), net=BASE),
```
IKarbowiak, line 26: Why this gross is different than gross from line 23?

saleor/checkout/tests/test_utils.py
```diff
@@ -0,0 +1,52 @@
...
+)
+def test_get_taxed_undiscounted_price(price, tax_rate, prices_entered_with_tax, result):
+    result_price = get_taxed_undiscounted_price(
+        undiscounted_base_price=BASE,
+        price=price,
+        tax_rate=tax_rate,
+        prices_entered_with_tax=prices_entered_with_tax,
+    )
+
+    assert result_price == result
```
IKarbowiak, line 52: IMO, instead of hardcoding the result, it should be calculated here based on the `tax_rate`. Otherwise the test is kind of magic and hard to maintenance.

jakubkuc: It kinda is, I agree. Though I had no other Idea how to test with `parameterize` all the scenarios for this function  - it checks and takes into consideration the corner-case when `tax_app` is returning slightly different results than local saleor calculations based on `tax_rate` ([this comment](https://github.com/saleor/saleor/pull/13585/files/8368e1944d6cd2b4ad8f741962864974b5203f00#diff-2f35ab82d7f669cedfbd9a04be1c39c1a1c3ee299730d7f0470254f5cc37721fR969)) 🤔 .

IKarbowiak: Firstly this test probably should be split into two separates - one for checking when `prices_entered_with_tax` is `True`, when when it's `False`. Then as a prameters you can provide the `price` (I see that `tax_rate` is the same in all examples so there is no reason to parametrize it) and based on `tax_rate` we should calculate the `net` or `gross` amount and compare it with `result_price`

saleor/checkout/tests/test_utils.py
jakubkuc: It is connected to the functionality described [here](https://github.com/saleor/saleor/pull/13585/files/8368e1944d6cd2b4ad8f741962864974b5203f00#diff-2f35ab82d7f669cedfbd9a04be1c39c1a1c3ee299730d7f0470254f5cc37721fR969) that takes into consideration that if local tax calculations for undiscounted prices should be the same, the price is always taken from the tax_app.

jakubkuc: I suppose that since I have to explain it it is not properly described in docstring and comments so I will need to fix that (unless you have a concern regarding this whole solution)

saleor/plugins/manager.py
```diff
@@ -448,7 +448,7 @@ def calculate_checkout_line_unit_price(
             default_value * quantity,
         )
         default_taxed_value = TaxedMoney(
-            net=total_value / quantity, gross=default_value
+            net=total_value / quantity, gross=total_value / quantity
```
korycins, line 451: Isn't something that can cause incorrect values on our side? Shouldn't we patch them also in previous versions?

saleor/order/migrations/tasks/saleor3_14.py
```diff
@@ -15,3 +26,131 @@ def order_propagate_expired_at_task():
...
+    order_to_prices_with_tax_map = get_order_prices_entered_with_tax_mapping(
+        order_lines,
+        order_model=Order,
+        channel_model=Channel,
+        tax_configuration=TaxConfiguration,
+    )
+
+    lines_to_update = set()
+    for line in order_lines:
+        prices_entered_with_tax = order_to_prices_with_tax_map[line.order.id]
```
korycins, line 138: This won't work. The prefetch_related is incorrect. It will raise:
```
ValueError: 'order_id' does not resolve to an item that supports prefetching - this is an invalid parameter to prefetch_related()
```
Have you tested it locally?

korycins: btw. IMHO The best way would be `line.order_id`  (and fixing incorrect prefetch_related) + solid test on local instance.

saleor/checkout/utils.py
```diff
@@ -954,3 +956,27 @@ def get_checkout_metadata(checkout: "Checkout"):
...
+    price: "TaxedMoney",
+    tax_rate: "Decimal",
+    prices_entered_with_tax: bool,
+):
+    """Apply taxes to undiscounted base price.
+
+    This function also prevents rounding difference between prices from tax-app and
+    local calculations based on tax_rate that might occur in orders without discounts.
+    """
+    price_to_compare = price.gross if prices_entered_with_tax else price.net
```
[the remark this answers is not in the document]
jakubkuc: yes it does. Sometimes when there are no discounts tax app can return `tax_rate` and taxed `price` that are not "matching" (are rounded differently) so the prices from `TAX_APP` and saleors calculations were slightly different. This prevents those small differences.

saleor/plugins/manager.py
jakubkuc: It is a bug that was found during development of this PR... maybe the best way would be to fix this particular line in separate PR and port it to versions affected

maarcingebala: To me it looks risky, I would remove and fix in a separate PR, with proper testing and QA only of this single case.

maarcingebala: Also with an GH issue first that proves and shows why and when this calculation is incorrect.

maarcingebala:
Closing in favor of https://github.com/saleor/saleor/pull/14095.

Issue #13557, opened 2023-07-28
Bug: Searching for products with `sort.field` set to `RANK` fails


krzysztofzuraw:
### What are you trying to achieve?

I want to use a new `search` argument on `Query.products`  with a `sort` argument set to `RANK`. This is a common case when searching for products in the dashboard.

### Steps to reproduce the problem

1. Try to execute the following query in the GraphiQL playground on the `main` branch:
```graphql
query ProductList($sort: ProductOrder, $search: String) {
  products(first: 10, sortBy: $sort, search: $search) {
    edges {
      node {
        id
      }
    }
  }
}
```
with variables:
```json
{
  "search": "bean",
  "sort": {
    "direction": "DESC",
    "field": "RANK"
  }
}
```
2. Error with `Sorting by RANK is available only when using a search filter.` appears

### What did you expect to happen?

I want to be able to perform query with search and sort.field set to `RANK`

### Environment

Saleor version: 3.15-dev
OS and version: macOS Ventura 13.4.1

PR #13617, opened 2023-08-03
Fix using search argument with sorting by RANK

SzymJ:
I want to merge this change because it fixes using `search` argument with sorting by `RANK`.

Fix for #13557

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/schema.py
+++ b/saleor/graphql/product/schema.py
@@ -148,7 +148,9 @@ from .types import (
 
 
 def search_string_in_kwargs(kwargs: dict) -> bool:
-    return bool(kwargs.get("filter", {}).get("search", "").strip())
+    return bool(kwargs.get("filter", {}).get("search", "").strip()) or bool(
+        kwargs.get("search", "").strip()
+    )
 
 
 def sort_field_from_kwargs(kwargs: dict) -> Optional[List[str]]:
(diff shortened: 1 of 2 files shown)


Air-t (approved):

kadewu (approved):

saleor/graphql/product/schema.py
```diff
@@ -453,9 +453,12 @@ def resolve_products(
     ):
         if sort_field_from_kwargs(kwargs) == ProductOrderField.RANK:
             # sort by RANK can be used only with search filter
-            if not search_string_in_kwargs(kwargs):
+            if not search_string_in_kwargs(kwargs) and not search:
```
maarcingebala, line 456: I guess this solution is OK, but we have two identical definitions of `search_string_in_kwargs` in Saleor; the other one is in the orders module. I suppose the bug happens there as well. 

We could fix it in all places if we had one reusable function `search_string_in_kwargs` and we changed the signature to remove `search=None` and get it from kwargs:

```
def resolve_products(_root, info: ResolveInfo, *, channel=None, **kwargs):
    search = kwargs.get("search")
```

And fixed the `search_string_in_kwargs` function:
```
def search_string_in_kwargs(kwargs: dict) -> bool:
    return bool(kwargs.get("filter", {}).get("search", "").strip()) or "search" in kwargs
```

SzymJ: I added these changes to products module. Orders don't have`where` filters implemented yet,

maarcingebala (approved):

szczecha (approved):

Issue #13319, opened 2023-07-05
Update sale mutations to use the `Promotion` objects instead of `Sale` - allow using the old sale mutations only on promotions with `old_sale` set to `True`


IKarbowiak:
Part of https://github.com/saleor/saleor/issues/12539

Update sale mutations to use the Promotion objects instead of Sale - allow using the old sale mutations only on promotions with old_sale set to True

aniav:
@zedzior noticed we also need to fetch the Promotions migrated from Sales by their corresponding id, so we'll add sale id to be stored on the Promotion model instead of `old_sale` and we'll make sure only Promotions with sale id are returned in the old queries and mutations.

zedzior:
[saleCreate](https://github.com/saleor/saleor/pull/13624)
[saleUpdate](https://github.com/saleor/saleor/pull/13633)
[saleDelete](https://github.com/saleor/saleor/pull/13635)
[saleChannelListingUpdate](https://github.com/saleor/saleor/pull/13659)
[saleAddCatalogues & saleRemoveCatalogues
](https://github.com/saleor/saleor/pull/13701)
[saleBulkDelete](https://github.com/saleor/saleor/pull/13716#event-10120185971)
[saleTranslate](https://github.com/saleor/saleor/pull/13721#pullrequestreview-1584672147)
[metadata mutations](https://github.com/saleor/saleor/pull/13717)

PR #13624, opened 2023-08-04
Refactor sale create mutation

zedzior:
I want to merge this change, because it makes `saleCreate` mutation compatible with new Promotion model.

Promotions created by the mutation must be treated as "old sale", therefor `old_sale_id` is assigned.

This PR also corrects `sale_converter` and `0049_migrate_sales_to_promotions` , to return empty predicate instead of `{"OR": []}` for old sales without catalogue.

Important note: Previous sale architecture had type and catalogue associated with Sale (parent node), while promotions stores the info in rules (child node). Therefore we need to ensure that:

# Impact

* [ ] New migrations

# Pull Request Checklist


SzymJ (approved):

saleor/discount/migrations/0050_old_sale_id_sequence.py
```diff
@@ -0,0 +1,27 @@
...
+    operations = [
+        migrations.RunSQL(
+            sql=[
+                """
+            CREATE SEQUENCE discount_promotion_old_sale_id_seq
+            OWNED BY discount_promotion.old_sale_id;
+
+            SELECT setval(
+                'discount_promotion_old_sale_id_seq',
+                 coalesce(max(old_sale_id), 0) + 1, false
```
IKarbowiak, line 20: IMO it should be more than + 1 as the new sales might be created it meantime

IKarbowiak (changes requested):

IKarbowiak (approved):

Issue #13560, opened 2023-07-28
Add `FULFILLMENT_UPDATED` webhook


maarcingebala:
Add `FULFILLMENT_UPDATED` webhook. The event should be run in the `fulfillmentUpdateTracking` mutation to notify listeners about the changed tracking number of the fulfillment. 

Payload: 

```graphql
type FulfillmentUpdated implements Event @doc(category: "Orders") {
  issuedAt: DateTime
  version: String
  issuingPrincipal: IssuingPrincipal
  recipient: App
  fulfillment: Fulfillment
  order: Order

  """Informs whether a notification should be sent to the customer."""
  notifyCustomer: Boolean!
}
```

Required permission: `MANAGE_ORDERS`
Note: this webhook will replace the `ORDER_FULFILLMENT_UPDATE` action event in the `NOTIFY` webhook.

Air-t:
@maarcingebala Should `ORDER_FULFILLMENT_UPDATE` action be removed as the part of this PR  or it should remain untouched cause some clients may relay on this notification? If so, should we mark it somehow in changelog or any other places?

maarcingebala:
We will mark NOTIFY webhook as deprecated when all notify events are covered with dedicated webhook events. We cannot remove them until Saleor 4.0 as many apps currently rely on it will need time to migrate.

maarcingebala:
We can close this issue in favor of another event - https://github.com/saleor/saleor/issues/13670 - which will make more sense in the `fulfillmentUpdateTracking`. There is no other mutation where `FULFILLMENT_UPDATED` would be required now so I think we can skip it for now.

PR #13630, opened 2023-08-07
Add notify user flag to fullfilment updated webhook

Air-t:
I want to merge this change because it adds missing `FULLFILMENT_UPDATED` updated event which replaces `NotifyEventType.ORDER_FULFILLMENT_UPDATE` event

Resolves: https://github.com/saleor/saleor/issues/13560

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/mutations/fulfillment_update_tracking.py
+++ b/saleor/graphql/order/mutations/fulfillment_update_tracking.py
@@ -3,14 +3,17 @@ from typing import cast
 import graphene
 
 from ....account.models import User
+from ....core.notify_events import NotifyEventType
 from ....order.actions import fulfillment_tracking_updated
 from ....order.notifications import send_fulfillment_update
 from ....permission.enums import OrderPermissions
+from ....webhook.event_types import WebhookEventAsyncType
 from ...app.dataloaders import get_app_promise
 from ...core import ResolveInfo
 from ...core.doc_category import DOC_CATEGORY_ORDERS
 from ...core.mutations import BaseMutation
 from ...core.types import OrderError
+from ...core.utils import WebhookEventInfo
 from ...plugins.dataloaders import get_plugin_manager_promise
 from ..types import Fulfillment, Order
 from .order_fulfill import FulfillmentUpdateTrackingInput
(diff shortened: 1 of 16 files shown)


saleor/graphql/order/mutations/fulfillment_update_tracking.py
```diff
@@ -63,7 +71,9 @@ def perform_mutation(  # type: ignore[override]
         fulfillment_tracking_updated(fulfillment, user, app, tracking_number, manager)
 
         notify_customer = input.get("notify_customer")
+        cls.call_event(manager.fulfillment_updated, fulfillment, notify_customer)
         if notify_customer:
+            # this event will be deprecated. Fulfillment_updated will replace it.
             send_fulfillment_update(order, fulfillment, manager)
```
korycins, line 77: You can also mention `NOTIFY` event in `webhook_events_info`. Some of the modules didn't receive `webhook_events_info` yet, but it would be much more easier to assume, that each mutation that already has `webhook_events_info` covers all events.

saleor/graphql/order/mutations/fulfillment_update_tracking.py
```diff
@@ -36,6 +38,12 @@ class Meta:
         permissions = (OrderPermissions.MANAGE_ORDERS,)
         error_type_class = OrderError
         error_type_field = "order_errors"
+        webhook_events_info = [
+            WebhookEventInfo(
+                type=WebhookEventAsyncType.FULFILLMENT_UPDATED,
```
korycins, line 43: I am not sure if the scope of the event here is correct. It is a single place where you added it. The mutation is not standard "update" action. It is an update of the tracking number.
If you don't plan to add more fulfillment_updated events, we should have an event for updating the tracking number instead of fulfillment_updated. (Also `tracking_number_updated` [I know, the name is not the best one]  is already implemented on PluginsManager side - see `fulfillment_tracking_updated`)

korycins: cc @maarcingebala

saleor/graphql/order/mutations/fulfillment_update_tracking.py
Air-t: sure, will add missing events if I see they are missing

saleor/graphql/order/mutations/fulfillment_update_tracking.py
Air-t: ok so you suggest that we should have 2 separate events, one for tracking number update, and the second for general fulfillment update, right? and add the possibility for user to choose which one of those two he would like to receive? also if the new `FULFILLMENT_UPDATE` event is used only in this place - for mutation confirming a tracking number - better is to rename it for `fulfillmeent_tracking_number_updated`

maarcingebala: After discussing the topic internally, we decided to introduce a new webhook event dedicated to setting tracking number: https://github.com/saleor/saleor/issues/13670.

PR #13633, opened 2023-08-07
Refactor sale update mutation

zedzior:
I want to merge this change, because it makes `saleUpdate` mutation compatible with new Promotion model.

This PR also corrects `sale_converter` and `0049_migrate_sales_to_promotions` by:

1. copy `notification_sent_datetime`
2. set rule's name as None instead of empty string (nullify `PromotionRule.name` field)

It adds `blank` flag to `PromotionRule.catalogue_predicate` field, since `{}` is a valid predicate. 

It adds missing validation for `start_date` and `end_date`.

Important note: Previous sale architecture had type and catalogue associated with Sale (parent node), while promotions stores the info in rules (child node). Therefore we need to ensure that:

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/discount/migrations/0049_migrate_sales_to_promotions.py
+++ b/saleor/discount/migrations/0049_migrate_sales_to_promotions.py
@@ -101,6 +101,7 @@ def convert_sale_into_promotion(Promotion, sale):
         updated_at=sale.updated_at,
         metadata=sale.metadata,
         private_metadata=sale.private_metadata,
+        last_notification_scheduled_at=sale.notification_sent_datetime,
     )
 
 
@@ -108,7 +109,6 @@ def create_promotion_rule(
     PromotionRule, sale, promotion, discount_value=None, old_channel_listing_id=None
 ):
     return PromotionRule(
-        name="",
         promotion=promotion,
         catalogue_predicate=create_catalogue_predicate_from_sale(sale),
         reward_value_type=sale.type,
--- a/saleor/discount/models.py
+++ b/saleor/discount/models.py
@@ -428,7 +428,7 @@ class PromotionRule(models.Model):
         Promotion, on_delete=models.CASCADE, related_name="rules"
     )
     channels = models.ManyToManyField(Channel)
-    catalogue_predicate = models.JSONField(blank=True)
+    catalogue_predicate = models.JSONField(blank=True, default=dict)
     reward_value_type = models.CharField(
         max_length=255, choices=RewardValueType.CHOICES, blank=True, null=True
     )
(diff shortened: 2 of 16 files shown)


saleor/discount/migrations/0049_migrate_sales_to_promotions.py
```diff
@@ -101,14 +101,15 @@ def convert_sale_into_promotion(Promotion, sale):
...
+        last_notification_scheduled_at=sale.notification_sent_datetime,
     )
 
 
 def create_promotion_rule(
     PromotionRule, sale, promotion, discount_value=None, old_channel_listing_id=None
 ):
     return PromotionRule(
-        name="",
+        name=promotion.name,
```
IKarbowiak, line 112: Hm, I'm not sure if we want to set the rule name the same as promotion name.

saleor/graphql/discount/mutations/sale/sale_update.py
```diff
@@ -53,57 +63,144 @@ class Meta:
...
-            cls.update_products_discounted_prices(
-                cleaned_input, previous_catalogue, current_catalogue
-            )
-        return cls.success_response(ChannelContext(node=instance, channel_slug=None))
+        return cls.success_response(ChannelContext(node=promotion, channel_slug=None))
+
+    @classmethod
+    def get_instance(cls, info: ResolveInfo, **data):
+        object_id = cls.get_global_id_or_error(data["id"], "Sale")
+        return models.Promotion.objects.get(old_sale_id=object_id)
```
IKarbowiak, line 99: Shouldn't we raise some specific error when the new promotion id is provided, to indicate that user should use the new promotion mutation?

saleor/discount/models.py
```diff
@@ -422,13 +422,13 @@ def get_translated_keys(self):
...
     id = models.UUIDField(primary_key=True, editable=False, unique=True, default=uuid4)
-    name = models.CharField(max_length=255)
+    name = models.CharField(max_length=255, blank=True, null=True)
     description = SanitizedJSONField(blank=True, null=True, sanitizer=clean_editor_js)
     promotion = models.ForeignKey(
         Promotion, on_delete=models.CASCADE, related_name="rules"
     )
     channels = models.ManyToManyField(Channel)
-    catalogue_predicate = models.JSONField()
+    catalogue_predicate = models.JSONField(blank=True)
```
IKarbowiak, line 431: I think we should set the default value here `default=dict`

IKarbowiak (approved):

fowczarek (approved):

fowczarek (approved):

Issue #13561, opened 2023-07-28
Add `notifyUser` flag to `FULFILLMENT_APPROVED` webhook payload


maarcingebala:
The `FULFILLMENT_APPROVED` webhook is triggered when the admin manually approves a fulfillment using the `fulfillmentApprove` mutation (when auto-approval of fulfillment is disabled). Mutation accepts the `notifyCustomer` flag, determining whether a notification should be sent to the customer. This flag needs to be passed to webhook so that apps can read it and based on that send notifications.

Extend the `FulfillmentApproved` type with the `notifyCustomer` flag:
```graphql
type FulfillmentApproved implements Event @doc(category: "Orders") {
  issuedAt: DateTime
  version: String
  issuingPrincipal: IssuingPrincipal
  recipient: App
  fulfillment: Fulfillment
  order: Order

  """Informs whether a notification should be sent to the customer."""
  notifyCustomer: Boolean!
}
```

PR #13637, opened 2023-08-08
Add `notify_customer` flag to `FulfillmentApproved` type

Air-t:
I want to merge this change because it allows to determine if customer should be informed when fulfilment is approved

Resolves https://github.com/saleor/saleor/issues/13561

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/mutations/order_fulfill.py
+++ b/saleor/graphql/order/mutations/order_fulfill.py
@@ -115,6 +115,10 @@ class OrderFulfill(BaseMutation):
                 type=WebhookEventAsyncType.ORDER_FULFILLED,
                 description="Order is fulfilled.",
             ),
+            WebhookEventInfo(
+                type=WebhookEventAsyncType.FULFILLMENT_APPROVED,
+                description="A fulfillment is approved.",
+            ),
         ]
 
     @classmethod
--- a/saleor/graphql/webhook/subscription_types.py
+++ b/saleor/graphql/webhook/subscription_types.py
@@ -1165,7 +1165,8 @@ class FulfillmentBase(AbstractType):
 class FulfillmentCreated(SubscriptionObjectType, FulfillmentBase):
     notify_customer = graphene.Boolean(
         description=(
-            "If true, send an email notification to the customer." + ADDED_IN_316
+            "If true, the app should send a notification to the customer."
+            + ADDED_IN_316
         ),
         required=True,
     )
@@ -1202,11 +1203,32 @@ class FulfillmentCanceled(SubscriptionObjectType, FulfillmentBase):
 
 
 class FulfillmentApproved(SubscriptionObjectType, FulfillmentBase):
+    notify_customer = graphene.Boolean(
+        description="If true, send a notification to the customer." + ADDED_IN_316,
+        required=True,
+    )
+
     class Meta:
+        doc_category = DOC_CATEGORY_ORDERS
         root_type = "Fulfillment"
-        enable_dry_run = True
+        enable_dry_run = False
         interfaces = (Event,)
         description = "Event sent when fulfillment is approved." + ADDED_IN_37
+
+    @staticmethod
+    def resolve_fulfillment(root, info: ResolveInfo):
+        _, data = root
+        return data["fulfillment"]
+
+    @staticmethod
+    def resolve_order(root, info: ResolveInfo):
+        _, data = root
+        return data["fulfillment"].order
+
+    @staticmethod
+    def resolve_notify_customer(root, _info: ResolveInfo):
+        _, data = root
+        return data["notify_customer"]
 
 
 class FulfillmentMetadataUpdated(SubscriptionObjectType, FulfillmentBase):
--- a/saleor/plugins/webhook/plugin.py
+++ b/saleor/plugins/webhook/plugin.py
@@ -958,14 +958,26 @@ class WebhookPlugin(BasePlugin):
                 fulfillment_data, event_type, webhooks, fulfillment, self.requestor
             )
 
-    def fulfillment_approved(self, fulfillment: "Fulfillment", previous_value):
+    def fulfillment_approved(
+        self,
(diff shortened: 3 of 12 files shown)


saleor/graphql/webhook/subscription_types.py
```diff
@@ -1197,11 +1197,26 @@ class Meta:
 
 class FulfillmentApproved(SubscriptionObjectType, FulfillmentBase):
     class Meta:
-        root_type = "Fulfillment"
-        enable_dry_run = True
+        root_type = None
+        enable_dry_run = False
```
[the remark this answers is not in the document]
Air-t: if I'm right - when we change the subscription to an object rather then a single model, we also drop the possibility of enabling dry run - not sure if my understanding is correct in this case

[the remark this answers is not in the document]
Air-t: good point, fixed

saleor/plugins/webhook/plugin.py
```diff
@@ -944,14 +944,23 @@ def fulfillment_canceled(self, fulfillment: "Fulfillment", previous_value):
...
             return previous_value
         event_type = WebhookEventAsyncType.FULFILLMENT_APPROVED
         if webhooks := get_webhooks_for_event(event_type):
             fulfillment_data = generate_fulfillment_payload(fulfillment, self.requestor)
             trigger_webhooks_async(
-                fulfillment_data, event_type, webhooks, fulfillment, self.requestor
+                fulfillment_data,
+                event_type,
+                webhooks,
+                {"notify_customer": notify_customer},
```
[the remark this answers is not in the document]
Air-t: ah, I did not test it manually - thought it will work if deliveries tests works - fixed

saleor/graphql/webhook/subscription_types.py
```diff
@@ -1200,12 +1200,35 @@ class Meta:
 
 
 class FulfillmentApproved(SubscriptionObjectType, FulfillmentBase):
+    notify_customer = graphene.Boolean(
+        description=(
+            "If true, send an email notification to the customer." + ADDED_IN_315
```
maarcingebala, line 1205: ADDED_IN_316

maarcingebala (approved):

IKarbowiak (approved):

saleor/graphql/order/mutations/order_fulfill.py
```diff
@@ -105,6 +106,16 @@ class Meta:
...
         error_type_class = OrderError
         error_type_field = "order_errors"
+        webhook_events_info = [
+            WebhookEventInfo(
+                type=WebhookEventAsyncType.FULFILLMENT_CREATED,
+                description="A new fulfillment is created.",
+            ),
+            WebhookEventInfo(
+                type=WebhookEventAsyncType.ORDER_FULFILLED,
+                description="Order is fulfilled.",
```
korycins, line 116: Isn't it like `OrderFulfill` can also trigger the event `FULFILLMENT_APPROVED`?
I mean this part: `saleor.order.actions.order_fulfilled`

saleor/graphql/webhook/subscription_types.py
```diff
@@ -1161,12 +1162,33 @@ def resolve_order(root, info: ResolveInfo):
 
 
 class FulfillmentCreated(SubscriptionObjectType, FulfillmentBase):
+    notify_customer = graphene.Boolean(
+        description="If true, send a notification to the customer." + ADDED_IN_316,
```
korycins, line 1166: Shouldn't be: `If true, the app should send a notification to the customer.` ? IMHO Now is little bit not clear who should send a notification 😅

saleor/graphql/webhook/subscription_types.py
```diff
@@ -1161,12 +1162,33 @@ def resolve_order(root, info: ResolveInfo):
...
+        required=True,
+    )
+
     class Meta:
+        doc_category = DOC_CATEGORY_ORDERS
         root_type = "Fulfillment"
-        enable_dry_run = True
+        enable_dry_run = False
```
korycins, line 1173: As far as I understand this means, that any fulfillment event that will have `notifyCustomer` will not work with dryRun. 
Instead of removing this part of the functionality, have you considered a change in `webhookDryRun` to keep a dry run for fulfillment? 

We could have a map, that for specific objects, changes the structure of the subscribable object. We could set `notifyCustomer` as a dummy value. 
Something like (`saleor/graphql/webhook/mutations/webhook_dry_run.py`):

```python
    # new method added
    @classmethod
    def get_subscribable_object(cls, db_object, event_type):
        if event_type == "fulfillment_created":
            return {
                "fulfillment": db_object,
                "notify_customer": True,
            }
        return db_object

    @classmethod
        event_type, object, query = cls.validate_input(info, **data)
        payload = None
        if all([event_type, object, query]):
            request = info.context
            # line below added
            subscribable_object = cls.get_subscribable_object(object, event_type)
            # replace usage of object with subscribable_object in line below
            payload = generate_payload_from_subscription(
                event_type, subscribable_object, query, request
            )
        return WebhookDryRun(payload=payload)
```
(Don't suggest with the structure, I don't think that the mapper should be placed in mutation code. I've placed it here only to show how we can solve the problem without turning off the dryRun)

saleor/graphql/order/mutations/order_fulfill.py
Air-t: yes it does, and it already sent this event - adding missing webhook info

saleor/graphql/webhook/subscription_types.py
Air-t: I believe English is not my best skill 🙈

saleor/graphql/webhook/subscription_types.py
Air-t: I think this is a good point, and there should be a task for that - to make all subscriptions work in dry run

PR #13645, opened 2023-08-08
Add note about OrderFullyPaid

mmiszy:
Fixes https://github.com/saleor/saleor-app-payment-adyen/issues/358

### DIFF
--- a/saleor/graphql/webhook/subscription_types.py
+++ b/saleor/graphql/webhook/subscription_types.py
@@ -533,7 +533,12 @@ class OrderFullyPaid(SubscriptionObjectType, OrderBase):
         root_type = "Order"
         enable_dry_run = True
         interfaces = (Event,)
-        description = "Event sent when order is fully paid." + ADDED_IN_32
+        description = (
+            "Event sent when order is fully paid."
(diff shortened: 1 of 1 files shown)


saleor/graphql/webhook/subscription_types.py
```diff
@@ -533,7 +533,12 @@ class Meta:
         root_type = "Order"
         enable_dry_run = True
         interfaces = (Event,)
-        description = "Event sent when order is fully paid." + ADDED_IN_32
+        description = (
+            "Event sent when order is fully paid."
+            "Note that in certain scenarios payments can still be cancelled even after this event is sent."
```
IKarbowiak, line 538: Lack of space at the end

korycins:
I don't agree that this is the correct place for such a description. The described actions are payment provider specific. It should be placed in AdyenApp docs. In such cases you should also include description there, what are suggested actions etc. 
To give more arguments, why I am not convinced to have this description on Saleor side: what doesn't it mean in case when user pays with cash in c&c point? Or on delivery? What does it mean, when you have a simple app that processes bank transfers?

Issue #13640, opened 2023-08-08
Multiple giftcards on one checkout clears giftcard balance


jakubkuc:
Adding two giftcards to a checkout that have higher `currentBalance` than checkouts `totalPrice` will use up whole balance on both giftcards even though one of the gift card should have some money left on it (like it is working when using single giftCard) 

Steps to reproduce:
- Create a checkout the shipping method added. 
- Add two gifcards with total amount higher than checkouts `totalPrice`
- Use `checkoutComplete`mutation to complete checkout, fetch giftcards andtheir balance

Expected results:
- Giftcards should have some balance left on them 

Results:
- Both giftcards balance is `0`

PR #13647, opened 2023-08-08
Fix how multiple giftcards are applied

jakubkuc:
I want to merge this change because it fixes [#13640](https://github.com/saleor/saleor/issues/13640) 
Multiple gift cards, that have overall balance bigger than total price, were both used up instead of having remaining change in giftcard. 

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/tests/test_order_from_checkout.py
+++ b/saleor/checkout/tests/test_order_from_checkout.py
@@ -188,8 +188,7 @@ def test_create_order_with_many_gift_cards_worth_more_than_total(
     gift_cards_balance_before_order = (
         gift_card_1.current_balance.amount + gift_card_2.current_balance.amount
     )
-    checkout.gift_cards.add(gift_card_2)
-    checkout.gift_cards.add(gift_card_1)
+    checkout.gift_cards.add(gift_card_2, gift_card_1)
     checkout.save()
     checkout_lines, unavailable_variant_pks = fetch_checkout_lines(checkout)
     checkout_info = fetch_checkout_info(checkout, checkout_lines, manager)
@@ -207,7 +206,7 @@ def test_create_order_with_many_gift_cards_worth_more_than_total(
     zero_price = zero_money(gift_card.currency)
 
     # then
-    assert order.gift_cards.count() > 0
+    assert order.gift_cards.count() == 2
     assert gift_card_1.current_balance == zero_price
     assert gift_card_2.current_balance.amount == gift_card_2_balance_halved
     assert price_without_gift_card.gross.amount == (
--- a/saleor/checkout/tests/test_order_from_checkout.py
+++ b/saleor/checkout/tests/test_order_from_checkout.py
@@ -151,6 +151,76 @@ def test_create_order_with_gift_card_partial_use(
     )
 
 
+def test_create_order_with_many_gift_cards_worth_more_than_total(
+    checkout_with_items_and_shipping,
+    gift_card_created_by_staff,
+    gift_card,
+    customer_user,
+    shipping_method,
+    app,
+):
+    # given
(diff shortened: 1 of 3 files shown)


saleor/order/tests/test_order_utils.py
```diff
@@ -306,7 +306,7 @@ def test_add_gift_cards_to_order(
 
     # when
     add_gift_cards_to_order(
-        checkout_info, order, Money(20, gift_card.currency), staff_user, None
+        checkout_info, order, Money(30, gift_card.currency), staff_user, None
```
maarcingebala, line 309: Why are we changing this?

saleor/order/utils.py
```diff
@@ -438,6 +440,7 @@ def update_gift_card_balance(
         total_price_left = total_price_left - gift_card.current_balance
         gift_card.current_balance_amount = 0
     balance_data.append((gift_card, previous_balance.amount))
+    return total_price_left
```
maarcingebala, line 443: We can update the typing in the signature now when the function returns a value.

saleor/order/tests/test_order_utils.py
jakubkuc: Because the test was checking wrong assumptions - there were two giftcards - with 10 and 20 balance, the test checked if both of the giftcards were zeroed after they were used even though on one card there should be 10 left. This would be valid when balance on both cards were equal to this number - that's why I raised it. 
And I've added the separate tet for case where two cards are above the total.

saleor/checkout/tests/test_order_from_checkout.py
```diff
@@ -151,6 +151,76 @@ def test_create_order_with_gift_card_partial_use(
...
+    )
+
+    gift_card_2.current_balance_amount = gift_card_2_new_balance
+    gift_card_2.initial_balance_amount = gift_card_2_new_balance
+    gift_card_2.save()
+    gift_cards_balance_before_order = (
+        gift_card_1.current_balance.amount + gift_card_2.current_balance.amount
+    )
+    checkout.gift_cards.add(gift_card_2)
+    checkout.gift_cards.add(gift_card_1)
```
IKarbowiak, line 192:
```suggestion
    checkout.gift_cards.set([gift_card_1, gift_card_2])
```

saleor/checkout/tests/test_order_from_checkout.py
```diff
@@ -151,6 +151,76 @@ def test_create_order_with_gift_card_partial_use(
...
+        user=None,
+        app=app,
+        tracking_code="tracking_code",
+    )
+    gift_card_1.refresh_from_db()
+    gift_card_2.refresh_from_db()
+    zero_price = zero_money(gift_card.currency)
+
+    # then
+    assert order.gift_cards.count() > 0
```
IKarbowiak, line 210: I guess we can check exact gift cards number here

saleor/order/tests/test_order_utils.py
```diff
@@ -347,6 +347,34 @@ def test_add_gift_cards_to_order(
...
 
 
+def test_add_gift_cards_to_order_with_more_than_total(
+    checkout_with_item, gift_card, gift_card_expiry_date, order, staff_user
+):
+    # given
+    checkout = checkout_with_item
+    checkout.user = staff_user
+    checkout.gift_cards.add(gift_card_expiry_date)
+    checkout.gift_cards.add(gift_card)
```
IKarbowiak, line 357:
```suggestion
    checkout.gift_cards.set([gift_card_expiry_date, gift_card])
```

IKarbowiak (approved):

8r2y5 (approved):

szczecha (approved):

PR #13655, opened 2023-08-09
Clear promotion `old_sale_id` when new promotion mutations is used

IKarbowiak:
Clear promotion `old_sale_id` when any new promotion mutations is used for old sale

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/discount/mutations/utils.py
+++ b/saleor/graphql/discount/mutations/utils.py
@@ -31,5 +31,5 @@ def clear_promotion_old_sale_id(promotion: Promotion, *, save=False):
     """Clear the promotion `old_sale_id` if set."""
     if promotion.old_sale_id:
         promotion.old_sale_id = None
-    if save:
-        promotion.save(update_fields=["old_sale_id"])
+        if save:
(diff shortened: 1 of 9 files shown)


saleor/graphql/discount/mutations/utils.py
```diff
@@ -24,3 +25,11 @@ def convert_catalogue_info_to_global_ids(
...
         )
     return converted_catalogue_info
+
+
+def clear_promotion_old_sale_id(promotion: Promotion, *, save=False):
+    """Clear the promotion `old_sale_id` if set."""
+    if promotion.old_sale_id:
+        promotion.old_sale_id = None
+    if save:
+        promotion.save(update_fields=["old_sale_id"])
```
korycins, line 35: In case when we receive `save=True`:
 It means that we always make a save action here. Whether you use this function a second or tenth time for a single promotion, we will always make a save.

IKarbowiak: Good point, will change

korycins (approved):

zedzior (approved):

PR #13701, opened 2023-08-16
Refactor sale catalogues mutations.

zedzior:
I want to merge this change because, because it makes `saleAddCatalogues` and `saleRemoveCatalogues` mutations compatible with new Promotion model.

# Impact

* [ ] New migrations

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/discount/mutations/sale/sale_base_catalogue.py
+++ b/saleor/graphql/discount/mutations/sale/sale_base_catalogue.py
@@ -31,7 +31,7 @@ class SaleBaseCatalogueMutation(BaseDiscountCatalogueMutation):
             raise_validation_error(
                 field="id",
                 message="Provided ID refers to Promotion model. "
-                "Please use 'promotionUpdate' mutation instead.",
+                "Please use 'promotionRuleCreate' mutation instead.",
                 code=DiscountErrorCode.INVALID.value,
             )
         object_id = cls.get_global_id_or_error(id, "Sale")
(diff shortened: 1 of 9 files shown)


8r2y5 (approved):

saleor/graphql/discount/mutations/sale/sale_base_catalogue.py
```diff
@@ -19,3 +26,30 @@ class Arguments:
...
+        new_product_ids = get_product_ids_for_predicate(new_predicate)
+
+        if previous_product_ids != new_product_ids:
+            if len(new_product_ids) > len(previous_product_ids):
+                product_ids = new_product_ids - previous_product_ids
+            else:
+                product_ids = previous_product_ids - new_product_ids
+            update_products_discounted_prices_for_promotion_task.delay(
+                list(product_ids)
+            )
```
IKarbowiak, line 55: We should recalculate prices for both new and removed product ids

zedzior: Hmm, logic behind is:

-  `if len(new_product_ids) > len(previous_product_ids) is True` it means this is `saleAddCatalogue` mutation, so we only need to recalculate products we just added. Products, which were previously in catalogue don't need to be recalculated, thus `product_ids = new_product_ids - previous_product_ids`
- `if len(new_product_ids) > len(previous_product_ids) is False` it means this is `saleRemoveCatalogue` mutation, so we only need to recalculate products we just removed. There is no need to recalculate products, which are still in catalogue, thus `product_ids = previous_product_ids - new_product_ids`.

Does it make sense?

IKarbowiak: Ok, right now it has sense, but please add a comments to explain that :)

zedzior: added explicit flag `is_add_mutation`

IKarbowiak (approved):

Issue #13603, opened 2023-08-02
Use Poetry to install dependencies in CI


maarcingebala:
### Problem

Currently, CI installs dependencies using `pip`, but there is also the `pyproject.toml` file used by Poetry, which gets updated by the Dependabot bot every week. As a result, Dependabot PRs are using outdated dependency files, and we cannot see in CI whether the package upgrade caused any problems with running the test suite.

### Solution

One possible solution is to rewrite the workflow to use Poetry instead of `pip` to install packages. It may require changing the following steps:
- "Install system dependencies" - install poetry (see [docs](https://python-poetry.org/docs/#installing-with-the-official-installer))
- "Install dependencies" - change the commands to use poetry to install packages with `poetry install`

PR #13736, opened 2023-08-21
Add poetry to workflow actions (#13603)

rafiwts:
I want to merge this change because it is trying to solve issue #13603 
#13603 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


maarcingebala:
This looks good but the workflow that would install the dependencies was not run in this PR, since there are no changes in `*.py` files or requirements. Maybe we could make some dummy changes just to see if the workflow works correctly, that would be later removed? Another option is to bump some package - this workflow is triggered when there is a change in `requirements_dev.txt` file. One candidate I see is `boto` (https://github.com/saleor/saleor/pull/13730).

maarcingebala:
@rafiwts This is OK, `gunicorn` was bumped in `pyproject.toml` but new versions of requirements.txt files were not updated yet.

.github/workflows/pytest.yml
```diff
@@ -42,23 +42,22 @@ jobs:
...
 
     steps:
-      - uses: actions/checkout@v3
-
-      - name: Install system dependencies
-        run: apt-get install -y libpq-dev
+       - name: Install and configure poetry
+        run: |
+          pipx install poetry==1.5.0
+          poetry -m config virtualenvs.create true
```
NyanKiyoshi, line 48: `virtualenvs.create = true` seems redundant as it's already the default config value. Or is this here for some other reason?

rafiwts: @NyanKiyoshi Thank you for the review. The whole point of the command was to change the poetry config value. But as you rightly suggested, since the default config is already `true`, there is no point of using the command in workflow actions.

pyproject.toml
```diff
@@ -12,7 +12,7 @@ documentation = "https://docs.saleor.io/"
   [tool.poetry.dependencies]
   python = "~3.9"
   babel = ">=2.8,<2.13"
-  boto3 = "^1.28"
+  boto3 = "1.28.30"
```
NyanKiyoshi, line 15: Probably accidental change:

```suggestion
  boto3 = "^1.28"
```

NyanKiyoshi:
I noticed something else that looks wrong (sorry!), we are missing `paths` for `poetry.lock` and `pyproject.toml` here:


https://github.com/saleor/saleor/blob/0a89093aa725428064520bfe3e496382e8d15e45/.github/workflows/pytest.yml#L4-L20

rafiwts:
The paths were have been for both push and pr.

.github/workflows/pytest.yml
```diff
@@ -42,23 +46,21 @@ jobs:
           --health-retries 5
 
     steps:
-      - uses: actions/checkout@v3
-
-      - name: Install system dependencies
-        run: apt-get install -y libpq-dev
```
NyanKiyoshi, line 48: Checkout step is missing and `libpq-dev` is probably needed

.github/workflows/test-migrations-compatibility.yml
```diff
@@ -30,25 +30,21 @@ jobs:
           --health-retries 5
 
     steps:
-      - name: Checkout branch
-        uses: actions/checkout@v3
-
-      - name: Install system dependencies
-        run: apt-get install -y libpq-dev
```
NyanKiyoshi, line 37: Checkout step is missing and `libpq-dev` is probably needed

.github/workflows/pytest.yml
```diff
@@ -42,23 +46,21 @@ jobs:
...
+       - name: Install and configure poetry
+        run: |
+          pipx install poetry==1.5.0
```
NyanKiyoshi, line 51: There is a syntax error, the indentation has one extra space compared to the next items in the `steps` array:
```suggestion
      - name: Install and configure poetry
        run: |
          pipx install poetry==1.5.0
```

.github/workflows/test-migrations-compatibility.yml
rafiwts: The deleted step has been restored for both files but as a separate step or perhaps `poetry` and `libpq-dev` installation should be treated as a one step

.github/workflows/pytest.yml
rafiwts: Syntax error has been corrected as well

rafiwts:
@NyanKiyoshi It seems that I may have made a basic error of not adding `pipx` as a dependency to `poetry`. I was wondering if I should add a new dependency or maybe replace `pipx` with a direct installation by `install.python-poetry.org`

NyanKiyoshi:
> @NyanKiyoshi It seems that I may have made a basic error of not adding `pipx` as a dependency to `poetry`. I was wondering if I should add a new dependency or maybe replace `pipx` with a direct installation by `install.python-poetry.org`

You can use `python -m pip install poetry==1.5.0` and use poetry via `python -m poetry`, probably no need for `pipx`

Quite odd though as `pipx` should be part of the workflow image https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#package-management. Could be an issue with `PATH`, maybe an action issue or image issue.

rafiwts:
@NyanKiyoshi Thank you for your help and a thorough explanation of that matter. I have updated both files with `pip` installation command. Hopefully, everything is going to work right now. Thus the virtual environment will be created inside a container run during the workflow.

rafiwts:
@maarcingebala The PR has been rebased. Now, the GitHub workflow has V4. I hope that it is going to work right now.

maarcingebala:
Fails with this error:
```
Run git config --global --add safe.directory "$GITHUB_WORKSPACE"
/__w/_temp/70e76b5a-1413-4b85-bd4f-777d25538[9](https://github.com/saleor/saleor/actions/runs/6085389190/job/16510091295?pr=13736#step:8:10)14.sh: 2: pre-commit: not found
Error: Process completed with exit code 127.
```
`pre-commit` is part of the dev dependencies; it looks like they're not installed. I think they should be installed by default just with `poetry install`.

rafiwts:
Hi @maarcingebala ! Thank you for your advice. The `-no-root` command has been deleted for both files.

maarcingebala:
I meant that dev dependencies should be installed without any special flag. The `--no-root` is actually fine, but it shouldn't affect it. I see in the logs that `pre-commit` is getting installed so maybe it's a problem with the PATH or Python environment? E.g. the next step that's supposed to run pre-commit is using a different Python environment?

NyanKiyoshi:
> I see in the logs that `pre-commit` is getting installed so maybe it's a problem with the PATH or Python environment? E.g. the next step that's supposed to run pre-commit is using a different Python environment?

`python -m pre_commit` should do the trick

rafiwts:
> I meant that dev dependencies should be installed without any special flag. The `--no-root` is actually fine, but it shouldn't affect it. I see in the logs that `pre-commit` is getting installed so maybe it's a problem with the PATH or Python environment? E.g. the next step that's supposed to run pre-commit is using a different Python environment?

Sorry, I must have misunderstood what you meant by that. I was also analyzing the failed workflow and I could notice that in fact all dependencies are getting installed and tried to figure out what is the reason for that. Maybe it is a problem with the environment as it does not see the dependencies installed. I will follow the advice of @NyanKiyoshi  and replace `pre-commit run --all` with `python -m pre-commit run --all`. I apologize for any inconvenience cause by so many issues with this PR.

maarcingebala:
@rafiwts No worries, testing CI changes is often quite cumbersome, but this PR adds changes that are very useful so we will try to assist here. Anyway, with this change suggested by @NyanKiyoshi it's still failing. TBH I don't know why, but it looks like path/env problem.

NyanKiyoshi:
> @rafiwts No worries, testing CI changes is often quite cumbersome, but this PR adds changes that are very useful so we will try to assist here. Anyway, with this change suggested by @NyanKiyoshi it's still failing. TBH I don't know why, but it looks like path/env problem.

`python3 -m pre_commit` without the dash (`-`), `python -m` takes a Python module name, behind the scene it does `from pre_commit import __main__`

rafiwts:
> `python3 -m pre_commit` without the dash (`-`), `python -m` takes a Python module name, behind the scene it does `from pre_commit import __main__`

@NyanKiyoshi Could you please explain exactly what you meant by `without a dash ('-')`. I can see that you used the command `pre_commit` instead of `pre-commit`. Does it concern this issue or maybe you meant that I should get rid of `--all` at the end, leaving just `python3 -m pre-commit run` command. Thank you in advance for your help.

NyanKiyoshi:
> > `python3 -m pre_commit` without the dash (`-`), `python -m` takes a Python module name, behind the scene it does `from pre_commit import __main__`
> 
> @NyanKiyoshi Could you please explain exactly what you meant by `without a dash ('-')`. I can see that you used the command `pre_commit` instead of `pre-commit`. Does it concern this issue or maybe you meant that I should get rid of `--all` at the end, leaving just `python3 -m pre-commit run` command. Thank you in advance for your help.

The command options are fine to leave as is, thus: `python -m pre_commit run --all`.

When you do `python -m my_module arg1 arg2`, python pretty much does that:

1. `sys.argv = ["arg1", "arg2"]`
2. `from my_module import __main__`
3. `__main__` gets executed (because `__name__` is `__main__`)

Python expects a module rather than a PyPI package name, thus `pre_commit` resolves to that module: https://github.com/pre-commit/pre-commit/tree/e2c6a822c79e3a7dbb206bfa1ab87928ec25848b/pre_commit

patrys:
@rafiwts Thanks for your work, I managed to get it working with a few changes.

What was happening is that `poetry` creates a virtualenv for every installation, and those packages and binaries are only visible through `poetry run` and `poetry shell`.

We can think about whether this is what we want or not. Creating and activating an environment prior to using `poetry install` will have it use the currently active environment instead.

NyanKiyoshi (approved):

rafiwts:
I can see that the changes have been already applied so no action required from me. Thank you for your support and exact  explanations of the matter. In fact, I did not resolve the issue but I am glad that, at least, I could contribute to resolve the issue in some way.

maarcingebala (approved):

maarcingebala:
@rafiwts Thanks a lot for your contribution to Saleor 🎉

PR #13758, opened 2023-08-22
Lazy legacy webhooks payload generation

maarcingebala:
The legacy webhook payloads are generated for each webhook call, even if a subscription query is defined. Some payloads (e.g. checkout or order) have large static payloads with lots of DB relations, which adds unnecessary dozens of DB queries to the webhook execution.

This PR refactors legacy payload generation to be called only when a subscription query is not defined, and the static payload should be generated.

Some examples of saved queries for event and payload generator function used:
- `checkout_created` event / `generate_checkout_payload` - 14 queries for checkout with 1 line
- `order_created` event / `generate_order_payload` - 3 queries for order with 1 line
- `fulfillment_approved` event / `generate_fulfillment_payload`  - 26 queries for order with 1 line and 1 fulfillment

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/plugins/webhook/tasks.py
+++ b/saleor/plugins/webhook/tasks.py
@@ -209,7 +209,7 @@ def trigger_webhooks_async(
         if legacy_data_generator:
             data = legacy_data_generator()
         elif data is None:
-            raise ValueError("No payload was provided for regular webhooks.")
+            raise NotImplementedError("No payload was provided for regular webhooks.")
 
         payload = EventPayload.objects.create(payload=data)
         deliveries.extend(
--- a/saleor/plugins/webhook/tasks.py
+++ b/saleor/plugins/webhook/tasks.py
@@ -185,19 +185,32 @@ def create_delivery_for_subscription_sync_event(
 
 
 def trigger_webhooks_async(
-    data, event_type, webhooks, subscribable_object=None, requestor=None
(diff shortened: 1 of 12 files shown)


8r2y5 (approved):

saleor/plugins/webhook/tasks.py
```diff
@@ -185,19 +185,32 @@ def create_delivery_for_subscription_sync_event(
...
     :param requestor: used in subscription webhooks to generate meta data for payload.
+    :param legacy_data_generator: used to generate payload for regular webhooks.
     """
     regular_webhooks, subscription_webhooks = group_webhooks_by_subscription(webhooks)
     deliveries = []
     if regular_webhooks:
+        if legacy_data_generator:
+            data = legacy_data_generator()
+        elif data is None:
+            raise ValueError("No payload was provided for regular webhooks.")
```
fowczarek, line 212: Shouldn't we raise InmpelentationError here?

fowczarek (approved):

saleor/plugins/webhook/tasks.py
maarcingebala: Yes, I wasn't sure, maybe `NotImplementedError` would be better.

fowczarek: That sounds good to me.

PR #13776, opened 2023-08-23
Extend AttributeValueTranslation.name to 250 chars

### LINKED ISSUE #13735: Extend length of attribute translation fields
There is a mismatch between the length of char fields for attribute values. The max lengths are as follows:

`AttributeValue` model:
- `name`: 250
- `value`: 100

`AttributeValueTranslation` model:
- `name`: 100

Limit on the `AttributeValueTranslation.name` doesn't allow saving translations for long attribute value names. To fix it we can make all these fields use a maximum available length of 255.

aniav:
I want to merge this change because we need to extend the length of `AttributeValueTranslation.name` to allow for longer translation names. I also want to make name and value on `AttributeValue` longer so it accommodates longer values.

The fix will be merged to 3.15, so added it to the changelog there.

Fixes #13735.

# Impact

- [x] New migrations

# Docs

Link to documentation not needed.

# Pull Request Checklist


korycins (approved):

saleor/attribute/models/base.py
```diff
@@ -214,9 +214,9 @@ def get_translated_keys(self):
 
 
 class AttributeValue(SortableModel, ModelWithExternalReference):
-    name = models.CharField(max_length=250)
+    name = models.CharField(max_length=255)
```
fowczarek, line 217: The name should be shorter than a slug. 
We generate a slug from the name and add a suffix to the slug in case of a duplicate. 
If someone uses a name with 255 chars suffix exceed a maximum number of characters in the slug field. 
Sufix generator - `saleor/saleor/core/utils/__init__.py::generate_unique_slug`

aniav: Maybe I could change the function so that it cuts the name and adds a suffix?

If I make the name shorter than a slug it will be inconsistent with the other models' fields and that's what we are aiming at here.

IKarbowiak (approved):

fowczarek (approved):

SzymJ (approved):

szczecha (approved):

PR #13777, opened 2023-08-23
Don't run Saleor or Celery by default inside a devcontainer

patrys:
This requires people to run `uvicorn saleor.asgi:application` manually in a terminal but it makes it easier to debug and/or restart. It's also closer to the flow of working without devcontainers.

Update: Celery is now also not started manually. In exchange VSCode is preconfigured to launch either with a single click (and debugger support).

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -32,23 +32,10 @@ services:
       - POSTGRES_PASSWORD=saleor
 
   redis:
-    image: library/redis:5.0-alpine
+    image: library/redis:7.0-alpine
     restart: unless-stopped
     volumes:
       - saleor-redis:/data
-
-  worker:
-    image: saleor
-    command: celery -A saleor --app=saleor.celeryconf:app worker --loglevel=info -B
-    restart: unless-stopped
-    env_file:
-      - common.env
-      - backend.env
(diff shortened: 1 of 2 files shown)


korycins (approved):
@patrys  Don't we need to mention somewhere about this? Looks like, we are adding an additional step that might not be clear to everyone.

fowczarek (approved):

patrys:
@korycins AFAIK, we don't have any docs yet for dev containers?

korycins:
So maybe at least a task on saleor-docs, to add this part when we will have more details about dev containers. Or a blank chapter with a note about this. 
It looks like a possible annoying thing, that we don't explain anywhere (it worked previously, now we need to run an additional command).

kadewu:
What if we add page to docs and explain that you can change command for better debugging experience instead of doing this change permanently in code? 

I do personally believe, that leaving the command has more benefits, as e.g. when trying to run someone code I tend to check docker-compose for that information.

patrys:
I've added VSCode launch configuration, so it now takes a single click to get Saleor running, including debugger and breakpoint support.

PR #13792, opened 2023-08-25
Optimize products bulk create

tomaszszymanski129:
I want to merge this change because it optimizes product bulk create mutation.
- `attribute.values.all()` is no longer called every single time for each attribute.
- Instead it calls a filter (only for attribute types that need it), uses `slug` field (does utilize an index).
- reduced core side memory consumption drastically.
- reduced DB returned row count drastically.
- improved mutation speed significantly.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/core/utils/__init__.py
+++ b/saleor/core/utils/__init__.py
@@ -14,6 +14,9 @@ from django_prices_openexchangerates import exchange_currency
 from prices import Money, MoneyRange, TaxedMoney, TaxedMoneyRange
 from text_unidecode import unidecode
 
+if TYPE_CHECKING:
+    from ...attribute.models import Attribute
+
 task_logger = get_task_logger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


saleor/core/utils/__init__.py
```diff
@@ -145,3 +148,10 @@ def prepare_unique_slug(slug: str, slug_values: Iterable):
         unique_slug = f"{slug}-{extension}"
 
     return unique_slug
+
+
+def prepare_unique_attribute_value_slug(attribute: "Attribute", slug: str):
+    value_slugs = attribute.values.filter(slug__startswith=slug).values_list(
+        "slug", flat=True
+    )
```
fowczarek, line 156: This implementation performs additional database query for each slug. Are we sure that is the proper solution?

tomaszszymanski129: Unlike previous implementation not for every but three specific types of attributes and now it's limited to a slug - not fetching every value. I will later see if it's possible to push it further but even now it's enormously faster than it was.

korycins (approved):

maarcingebala (approved):

IKarbowiak (approved):

PR #13778, opened 2023-08-23
Introduce a new field to store product on AssignedProductAttributeValue

aniav:
I want to merge this change because it's introducing a new field for a FK to Product on `AssignedProductAttributeValue`. This change is part of the work needed to simplify Product - Attribute relation as described in https://github.com/saleor/saleor/issues/12881

# Impact

- [x] New migrations
- [ ] New/Updated API fields or mutations
- [ ] Deprecated API fields or mutations
- [ ] Removed API types, fields, or mutations

# Docs

No changes to the docs.

# Pull Request Checklist

- [x] Privileged queries and mutations are either absent or guarded by proper permission checks
- [x] Database queries are optimized and the number of queries is constant
- [ ] Database migrations are either absent or optimized for zero downtime
- [x] The changes are covered by test cases

PR #13814, opened 2023-08-28
Add new field to store page in AssignedPageAttributeValue

michal-macioszczyk:
I want to merge this change because it is part of refactoring work https://github.com/saleor/saleor/issues/12881
This is first step to add page field and migration, secondly will be add logic which will simplify Page - Attribute relation.  

Small update regarding to zero downtime migrations documentation page
https://github.com/saleor/saleor-docs/pull/921

# Impact

- [x] New migrations

# Docs

- [x] Link to documentation:

# Pull Request Checklist


saleor/attribute/migrations/tasks/saleor3_16.py
```diff
@@ -0,0 +1,57 @@
...
+
+    Update a batch of 'AssignedPageAttributeValue' rows by setting their 'page' based
+    on their related 'assignment'.
+
+    The number of rows updated in each batch is determined by the BATCH_SIZE.
+    Rows are locked during the update to prevent concurrent modifications.
+    """
+    with transaction.atomic():
+        with connection.cursor() as cursor:
+            cursor.execute(
```
korycins, line 22: Have you tested the update logic with DB that has +/- million attributes?

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ in 3.17. Use `PaymentSettings.defaultTransactionFlowStrategy` instead.
 ### Other changes
 - Fix error in variant available stock calculation - 13593 by @awaisdar001
 - Fix giftcard code description - #13728 by @rafiwts
+- Add a new `page` field on `AssignedPageAttributeValue`. First stage (migration) of a simplification of Attribute - Page relation from #13403. by michal-macioszczyk
```
korycins, line 40: Are you aware that merging this to main, means that the data migration will happen in 3.16. It means that you will be able to use it in 3.17. So the refactor logic that you're going to introduce will need to be delivered to 3.17
As far as I understand, it currently looks like this:
- main (future 3.16) received changes from this PR
- 3.16 release and migration data, trigger the celery task to update the attributes
- <wait until it ends>
- 3.17 receives the logic that can be sure that all "new" attribute  values are already propagated

I would say that this PR should be added to 3.15 (as a patch release). Then your main logic can be delivered to 3.16.
So the flow would look:
- 3.15.X received a data migration
- celery task updates the attributes
- (main) 3.16 recieve (+ update PR that was merged to 3.15*) the logic that can be sure that all "new" attribute values are already propagated

* The celery task introduced in 3.15, should be converted to standard migration in 3.16

saleor/attribute/migrations/tasks/saleor3_16.py
fowczarek, line 1: Missing loop. Migration update only first `BATCH_SIZE` number of items.

saleor/attribute/migrations/tasks/saleor3_16.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    assigned_values = (
+        AssignedPageAttributeValue.objects.filter(page__isnull=True)
+        .values_list("pk", flat=True)
+        .exists()
+    )
+    # If we found data, queue next execution of the task
+    if assigned_values:
+        update_page_assignment()
+        assign_pages_to_attribute_values_task.delay()
```
fowczarek, line 57: The task should be called in post migrate.

saleor/attribute/models/page.py
```diff
@@ -24,6 +33,10 @@ class Meta:
     def get_ordering_queryset(self):
         return self.assignment.pagevalueassignment.all()
 
+    def save(self, *args, **kwargs):
+        self.page = self.assignment.page
+        super(AssignedPageAttributeValue, self).save(*args, **kwargs)
```
fowczarek, line 38: This code adds 1 additional db query per each save.

saleor/attribute/models/page.py
```diff
@@ -66,3 +79,17 @@ class Meta:
...
+    if action == "post_add":
+        # Get the newly added AttributeValues
+        added_values = kwargs.get("pk_set", [])
+
+        for value_id in added_values:
+            # Find the corresponding AssignedPageAttributeValue instance
+            apa_value = AssignedPageAttributeValue.objects.get(
+                value_id=value_id, assignment=instance
+            )
+            apa_value.save()
```
fowczarek, line 95: This code executes 2 database queries for each element.

saleor/graphql/page/tests/mutations/test_page_attributes_migration.py
fowczarek, line 1: Invalid file localization. We should test the migration code in `graphlq` module.
`graphql` module should only contain only graphql-related things.

saleor/tests/fixtures.py
```diff
@@ -5664,6 +5664,26 @@ def page(db, page_type):
     return page
 
 
+@pytest.fixture
+def two_pages(page):
```
fowczarek, line 5668: Maybe `secound_page` instead of `two_pages` should be created here? 
Currently, this fixture is used by one test and is unwrapped at the start of the test.

CHANGELOG.md
aniav: We discussed that outside of github so I'll just add for visibility. For both this PR and #13778 we're aiming at adding the task in 3.16 and the rest in 3.17. So this is similar to the above strategy just moved by 1 version 👯

saleor/attribute/models/page.py
michal-macioszczyk: fixed by changing the strategy, right now there is one place where page from AssignedPageAttributeValue is updated, no need for signals and override save method.

saleor/attribute/models/page.py
michal-macioszczyk: I removed and change the approach for more efficient solution.

saleor/tests/fixtures.py
michal-macioszczyk: done

saleor/attribute/migrations/tasks/saleor3_16.py
michal-macioszczyk: Yes and single task range for 1_000_000 rows was min/max 0.15/0.52 second so based on that I have increased `BATCH_SIZE` to 5000. 
It was tested on my local machine.

saleor/attribute/utils.py
```diff
@@ -51,6 +51,13 @@ def associate_attribute_values_to_instance(
             assignment_id=assignment.pk
         ).update(product_id=instance.pk)
 
+    # This code will be deleted in new release, it is temporary solution between
+    # releases to keep database in sync
```
IKarbowiak, line 55: IMO we should write the specific version number here

IKarbowiak (approved):

korycins (approved):

saleor/attribute/utils.py
michal-macioszczyk: done

saleor/attribute/migrations/0031_assignedpageattributevalue_page.py
```diff
@@ -0,0 +1,38 @@
...
+    dependencies = [
+        ("page", "0028_add_default_page_type"),
+        ("attribute", "0030_assignedproductattributevalue_product"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="assignedpageattributevalue",
+            name="page",
+            field=models.ForeignKey(
```
kadewu, line 29: By default Django creates index when creating `ForeignKey`. We need to explictly turn off that behaviour by doing `db_index=False` in `ForeignKey` field creation. And then create index if needed. 
(and I believe we should have that index in db.)
Example can be found here: https://github.com/saleor/saleor/pull/13707/files

Note: it's not yet in documentation I will try to update it asap.

saleor/attribute/migrations/0031_assignedpageattributevalue_page.py
```diff
@@ -0,0 +1,38 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+from django.db.models.signals import post_migrate
+from django.apps import apps as registry
+
+from .tasks.saleor3_16 import assign_pages_to_attribute_values_task
+
+
+def data_migration(apps, _schema_editor):
```
kadewu, line 11: It's good idea to separate data migration with schema changes.

saleor/attribute/migrations/0031_assignedpageattributevalue_page.py
michal-macioszczyk: done

saleor/attribute/migrations/0031_assignedpageattributevalue_page.py
michal-macioszczyk: done

kadewu (approved):

Issue #13788, opened 2023-08-25
Turn on webhook dry runs if they are disabled.


Air-t:
## Problem

Enable all missing webhook dry runs. 
In cases where subscription is not a model object, some dry runs are disabled.

### Acceptance Criteria

working dry runs.

PR #13816, opened 2023-08-29
Enable subscription webhook dry runs which were initially disabled.

Air-t:
I want to merge this change because it enables missing webhook dry runs

Resolves https://github.com/saleor/saleor/issues/13788

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,9 +32,12 @@ in 3.17. Use `PaymentSettings.defaultTransactionFlowStrategy` instead.
   - Event sent when CSV export for gift cards is completed.
 - Add `PRODUCT_EXPORT_COMPLETED` webhook - #13787, by @Air-t
   - Event sent when CSV export for products is completed.
-
 - Add `FULFILLMENT_TRACKING_NUMBER_UPDATED` webhook - #13708, by @Air-t
   - Called after `fulfillmentUpdateTracking` or `orderFulfill` mutation if tracking number is updated.
+- Enable webhook dry runs - #13816, by @Air-t
+  - Dry runs enabled: `AccountConfirmed`, `AccountConfirmationRequested`, `AccountChangeEmailRequested`
+  - `AccountEmailChanged`, `AccountSetPasswordRequested`, `AccountDeleteRequested`, `GiftCardSent`,
+  - `FulfillmentCreated`, `FulfillmentApproved`, `StaffSetPasswordRequested`
 
 ### Other changes
 - Fix error in variant available stock calculation - 13593 by @awaisdar001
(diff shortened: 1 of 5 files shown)


saleor/webhook/event_types.py
```diff
@@ -722,6 +722,50 @@ class WebhookEventAsyncType:
...
     ALL = [event[0] for event in CHOICES]
 
+    @classmethod
+    def get_subscribable_object(
+        cls,
+        db_object: Any,
+        event_type: str,
+        **kwargs: Any,
+    ):
+        subscribable_objects_map = {
```
maarcingebala, line 732: This part looks a bit difficult to read. I'm wondering if there would be any way that we could define these context dicts per event. Not sure where in code at this point, but that would be ideal. Also, I would consider naming this "subscription context" instead of "subscribbable object", but that's a secondary change.

Air-t: this can looks much simpler for just dry runs, if we skip the kwargs. The idea was to also use this function for `PluginManager` methods and replace `subscribbable object` with it. This way we had a logic in one place which is in `WebhookEventAsyncType`

korycins: This method, adds you an additional step to cover in case of adding a new webhook. This means, that everyone needs to remember to also check this method when adding new webhook. 
IMHO we should consider using `EVENT_MAP` ☝️  if possible.
Maybe we could add a new field like `dry-run-payload` to EVENT_MAP or something, and include it only for the 👇  events. 

```python
    EVENT_MAP: dict[str, dict[str, Any]] = {
        FULFILLMENT_CREATED: {
            "name": "Fulfillment created",
            "permission": OrderPermissions.MANAGE_ORDERS,
            "dry_run_payload": lambda f: {"fulfillment": f, "notify_customer": False},
        },
    }
```
Then In dry run mutation, we would have something similar to:

```python
    @classmethod
        event_type, object, query = cls.validate_input(info, **data)
        payload = None
        if all([event_type, object, query]):
            request = info.context
             if "dry_run_payload" in WebhookEventAsyncType.EVENT_MAP[event_type]:
                 object = WebhookEventAsyncType.EVENT_MAP[event_type]["dry_run_payload"](object)
            payload = generate_payload_from_subscription(
                event_type, object, query, request
            )
        return WebhookDryRun(payload=payload)
```
(the changes in perform_mutation are only to show the way how we could use it).

@maarcingebala  @Air-t  WDYT? I don't like that we have `lambda` here, but I still prefer this way as we have an event configuration in single place.

Air-t: I have changed a approach a bit so please first look at new one - we can discouss which one should we follow.
Now if webhook is a standard one - have a subsciption object as a model object, there is no need to add extra methods. if someone needs to configure a dict object, then a `sunscription_type` needs also to override the `get_subscription_context` method

maarcingebala (approved):

korycins (approved):

PR #13921, opened 2023-09-07
Fix managers calculate checkout line unit price default value

### LINKED ISSUE #13920: Bug: Invalid default taxed value for calculate_checkout_line_unit_price in manager

In `calculate_checkout_line_unit_price` in manager.py the `default_taxed_value ` is invalid. 
It is set to: 
`net = <discounted unit price>` and `gross= <undiscounted_unit_price>` 
That problem occurs only when there are discounts on checkout, there are no plugins configured, or (e.g for avatax) when `charge_taxes` is set to False cause then, the default value is returned.

1. Create checkout with discounts and products with quantity greater than one
2. Configure avatax plugin with charge_taxes as False
3. Fetch Checkout line unit price - it will be invalid 

Price should be valid and the same for net and gross (since the taxes should not be charged)

### Logs

_No response_

### Environment

Saleor version: 3.14
OS and version: …

jakubkuc:
I want to merge this change because it fixes #13920 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/plugins/manager.py
+++ b/saleor/plugins/manager.py
@@ -448,7 +448,7 @@ class PluginsManager(PaymentInterface):
             default_value * quantity,
         )
         default_taxed_value = TaxedMoney(
-            net=total_value / quantity, gross=default_value
+            net=total_value / quantity, gross=total_value / quantity
         )
         unit_price = self.__run_method_on_plugins(
             "calculate_checkout_line_unit_price",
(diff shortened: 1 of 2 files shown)


8r2y5 (approved):

korycins (approved):

IKarbowiak (approved):

maarcingebala:
Closing this for now, as we need to still confirm whether this problem is affecting anyone. We'll reopen if it's confirmed.

PR #13953, opened 2023-09-12
Fix amount value comparison in transaction webhooks

korycins:
I want to merge this change because it fixes a case, when `transactionEventReport` was called first, and response from the transactionInitialize was returned as a second. In that case, there was a possibility that we will have a decimal value with different precision, which could cause a problem with de-duplication of the events.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/payment/utils.py
+++ b/saleor/payment/utils.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, Optional, Union, cast, overload
 import graphene
 from aniso8601 import parse_datetime
 from babel.numbers import get_currency_precision
+from django.conf import settings
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import transaction
 from django.db.models import Q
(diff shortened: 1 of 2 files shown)


8r2y5 (approved):

saleor/payment/utils.py
```diff
@@ -977,6 +989,7 @@ def deduplicate_event(
     already_existing_event = get_already_existing_event(event)
     if already_existing_event:
         if already_existing_event.amount != event.amount:
+            # FIXME: add amounts and psp reference
```
IKarbowiak, line 992: Should it be fixed in this PR?

saleor/payment/utils.py
```diff
@@ -780,9 +781,20 @@ def parse_transaction_event_data(
...
+            parsed_event_data["amount"] = amount
         except decimal.DecimalException:
+            amount_valid = False
+
+        if amount and not amount.is_finite():
+            amount_valid = False
+
+        if not amount_valid:
             logger.warning(invalid_msg, "amount", amount_data)
             error_field_msg.append(invalid_msg % ("amount", amount_data))
```
IKarbowiak, line 799: Can we extract it to separate function?

Air-t (approved):

IKarbowiak (approved):

szczecha (approved):

PR #13956, opened 2023-09-12
Fix amount value comparison in transaction webhooks

korycins:
I want to merge this change because it fixes a case, when transactionEventReport was called first, and response from the transactionInitialize was returned as a second. In that case, there was a possibility that we will have a decimal value with different precision, which could cause a problem with de-duplication of the events.

Port of changes: #13953 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


8r2y5 (approved):

IKarbowiak:
I added comments in main PR

PR #13971, opened 2023-09-14
Raise PaymentError instead of Exception

korycins:
I want to merge this change because is uses PaymentError instead of normal Exception. PaymentError is handled and should generate proper error in mutation response.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


maarcingebala:
Sentry issue for context: https://saleor.sentry.io/issues/4205743338/events/1e5639899de5499ab9e845b722ad27f3/
I understand this change will make this error disappear from Sentry, but it will still be returned to the user. What is the reason for raising this error exactly?

korycins:
@maarcingebala  Yes, this will remove the error from sentry. 
Currently it raises unhandled exception. The paymentError is better here as someone tries to execute payment action on the plugin/app which is not accessible/or doesn't have a proper webhook.

PR #14006, opened 2023-09-19
Update sale to promotion migration

IKarbowiak:
Update sale migartion:
 - call `update_products_discounted_price` to create `VariantChannelListingPromotionRule` instances
 - update creation of `OrderLineDiscount`
 - add locks to prevent potential Promotion creation during sales to promotion migration

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/discount/migrations/tasks/saleor3_16.py
+++ b/saleor/discount/migrations/tasks/saleor3_16.py
@@ -20,12 +20,11 @@ from ...models import (
 )
 from babel.numbers import get_currency_precision
 
-
 # The batch of size 100 takes ~0.9 second and consumes ~30MB memory at peak
 BATCH_SIZE = 100
 
-# Results in memory usage of ~40MB for 500 products
-DISCOUNTED_PRICES_RECALCULATION_BATCH_SIZE = 500
+# Results in memory usage of ~40MB for 20K products
+DISCOUNTED_PRICES_RECALCULATION_BATCH_SIZE = 100
 
 
 @dataclass
--- a/saleor/discount/migrations/tasks/saleor3_16.py
+++ b/saleor/discount/migrations/tasks/saleor3_16.py
@@ -1,9 +1,11 @@
 from dataclasses import dataclass
 from typing import Dict, List
 from ....celeryconf import app
+from decimal import Decimal
 
 import graphene
 from django.db.models import Exists, OuterRef
+from ....order.models import OrderLine
 from ...models import (
     Promotion,
     PromotionRule,
@@ -14,6 +16,7 @@ from ...models import (
     CheckoutLineDiscount,
     OrderLineDiscount,
 )
+from babel.numbers import get_currency_precision
 
 
 # The batch of size 100 takes ~0.9 second and consumes ~30MB memory at peak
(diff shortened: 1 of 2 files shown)


saleor/discount/migrations/tasks/saleor3_16.py
```diff
@@ -1,9 +1,14 @@
 from dataclasses import dataclass
 from typing import Dict, List
 from ....celeryconf import app
```
michal-macioszczyk, line 3: https://docs.saleor.io/docs/3.x/developer/community/zero-downtime-migrations#additional-autodiscovery-with-specified-related-names

additional configuration is needed in this case

michal-macioszczyk (approved):

saleor/discount/migrations/tasks/saleor3_16.py
IKarbowiak: It's already added.

zedzior (approved):

saleor/discount/migrations/tasks/saleor3_16.py
```diff
@@ -253,26 +281,57 @@ def migrate_checkout_line_discounts(sales_pks, rule_by_channel_and_sale):
...
+    start_pk: int = 0,
+):
+    products = list(
+        Product.objects.filter(pk__gt=start_pk)
+        .prefetch_related("channel_listings", "collections")
+        .order_by("pk")[:DISCOUNTED_PRICES_RECALCULATION_BATCH_SIZE]
+    )
+
+    if products:
+        update_products_discounted_price(products)
```
kadewu, line 334: Correct me if I'm wrong, but doesn't we fetch products again in this method. If so, there is no need to fetch all data about products here.

saleor/discount/migrations/tasks/saleor3_16.py
```diff
@@ -253,26 +281,57 @@ def migrate_checkout_line_discounts(sales_pks, rule_by_channel_and_sale):
...
-        for order_line_discount in order_line_discounts:
-            if order_line := order_line_discount.line:
-                channel_id = order_line.order.channel_id
-                sale_id = order_line_discount.sale_id
-                lookup = f"{channel_id}_{sale_id}"
-                order_line_discount.type = "promotion"
-                if promotion_rule := rule_by_channel_and_sale.get(lookup):
-                    order_line_discount.promotion_rule = promotion_rule
+    global_pks = [graphene.Node.to_global_id("Sale", pk) for pk in sales_pks]
+    if order_lines := OrderLine.objects.filter(sale_id__in=global_pks).select_related(
```
kadewu, line 285: `prefetch_related` may result in overall better performance here (not sure though)

But anyway, I think we can specify here that we don't need to join whole tables, as we just use `order.channel_id` from order only.

IKarbowiak: I checked, and we cannot fetch just specific field with `prefetch_related` we need to fetch `channel` and in such case we're fetching both `order` and `channel`, so `select_related` seems better as we are fetching only order and then using `channel_id`

fowczarek (approved):

PR #14021, opened 2023-09-20
[TEST] order deletion after expiration

yellowee:
I want to merge this change because it covers [CORE_0216](https://saleor.testmo.net/repositories/5?group_id=112&case_id=2399) - expired order is deleted after specified time.
I've also modified test for expiring an order and creating an order from checkout.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/tests/e2e/orders/test_expired_order_is_deleted_after_specified_time.py
+++ b/saleor/tests/e2e/orders/test_expired_order_is_deleted_after_specified_time.py
@@ -1,14 +1,20 @@
-import pytest
+import datetime
 
+import pytest
+from django.utils import timezone
+from freezegun import freeze_time
+
+from ....order.tasks import delete_expired_orders_task, expire_orders_task
+from ..channel.utils import update_channel
 from ..checkout.utils import checkout_create, checkout_delivery_method_update
 from ..product.utils.preparing_product import prepare_product
 from ..shop.utils.preparing_shop import prepare_shop
 from ..utils import assign_permissions
-from .utils import order_create_from_checkout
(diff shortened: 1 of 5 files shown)


saleor/tests/e2e/orders/test_expired_order_is_deleted_after_specified_time.py
```diff
@@ -0,0 +1,124 @@
...
+    data = order_query(e2e_staff_api_client, order_id)
+    assert data["status"] == "EXPIRED"
+    assert data["paymentStatus"] == "NOT_CHARGED"
+
+    # Step 5 - Wait and check the order has been deleted
+    time.sleep(80)
+    delete_expired_orders_task()
+    data = order_query(e2e_staff_api_client, order_id)
+    assert data["status"] == "EXPIRED"
+    assert data["paymentStatus"] == "NOT_CHARGED"
```
szczecha, line 124: I have some doubts here.
Shouldn't we expect that the order query does not return any data? 
time.sleep(80) means that we wait 80 seconds but deleteExpiredOrdersAfter is expressed in days so we should use something different here to make sure that the expiration date passed and check if the order does not exist.

yellowee: You're right - working on it now

IKarbowiak (approved):

michal-macioszczyk (approved):

fowczarek (approved):

Issue #13770, opened 2023-08-23
Use separate socket and read timeouts when making external calls


patrys:
Throughout the codebase, we pass a single number as the` requests.*`'s `timeout` argument. The number is used as the socket connection timeout, and the remainder is used as the read timeout. We never want to wait for 20 or 30 seconds for a socket connection to be established, so I'd like to see Saleor always pass a tuple, explicitly splitting the time budget into a connection and read time limit.

Example:

```python
# allow up to 2 seconds for a connection to be established
# and up to 18 seconds for the response to be sent over
requests.get(
    manifest_url, headers=headers, timeout=(2, 18), allow_redirects=False
)
```

While at it, we should also ensure all timeouts are named constants instead of magic numbers.

NyanKiyoshi:
[RFC 8305](https://www.rfc-editor.org/rfc/rfc8305) recommends two seconds as the connect timeout which should be more than enough in most cases, I thus suggest using 2s as the default value (users should be able to adjust the value as needed through `settings.py` for flexibility).
In most cases, the request would get acknowledged within that connect timeout and be put in queue in the remote target's system/infra.

patrys:
Agreed, that's precisely why I want them separate. If we cannot get the connection accepted within 2 seconds, something is broken, and there's no point in waiting for the entire timeout.

maarcingebala:
Not sure if this is the same problem but, seems related (it was reported on Clickup):

Some synchronous webhooks don't always respect the timeout depending on which layer failed (see below screenshots for the errors):
- https://webhook.site:10/ - times out after 20s
- https://master.staging.saleor.cloud:10/ - times out after 1:20mn

Steps to reproduce:
- Add webhook for "list payment methods"
- Endpoint: "https://master.staging.saleor.cloud:10/"
- Query will return HTTP 504:
```
{
  shop {
    availablePaymentGateways {
      name
      id
      config {
        value
      }
      currencies
    }
  }
}
```

PR #13891, opened 2023-09-05
Add IP filter for HTTP requests

NyanKiyoshi:
This adds an IP filtering feature to the project allowing users to disallow private IPs and loopback IPs. This is specifically useful for administrators that wish for the staff users to be unable to talk to internal systems.

# Impact

- [ ] New migrations
- [ ] New/Updated API fields or mutations
- [ ] Deprecated API fields or mutations
- [ ] Removed API types, fields, or mutations

# Docs

- [x] Link to documentation: https://github.com/saleor/saleor-docs/pull/933

# Pull Request Checklist

- [ ] Privileged queries and mutations are either absent or guarded by proper permission checks
- [ ] Database queries are optimized and the number of queries is constant
- [ ] Database migrations are either absent or optimized for zero downtime
- [ ] The changes are covered by test cases

PR #14057, opened 2023-09-22
Strict connection timeout settings

tomaszszymanski129:
I want to merge this change because it sets consistent settings for connection establishment timeout.
More context in issue https://github.com/saleor/saleor/issues/13770
Resolves https://github.com/saleor/saleor/issues/13770

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/app/management/commands/create_app.py
```diff
@@ -13,6 +14,8 @@
 from ...models import App
 from .utils import clean_permissions
 
+REQUEST_TIMEOUT = (settings.REQUESTS_CONN_EST_TIMEOUT, 15)
```
maarcingebala, line 17: We talked about it with @patrys that at this point we could set all timeouts to the same value and have 20s total, but split as follows:
- connection timeout - 2s
- read timeout - 18s

patrys: Yes, I suggest we set the default timeout of hardened requests to (2, 18) and not provide any overrides unless strictly required (and those cases need a big fat comment explaining why it needs to be special).

patrys: I believe there is a PR to add https://github.com/saleor/requests-hardened to our stack.

NyanKiyoshi: > I believe there is a PR to add https://github.com/saleor/requests-hardened to our stack.

Added in 3.16.0 via #13891 


You can override the global default here: https://github.com/saleor/saleor/blob/cc0ff0f7a533ddf5fb2a0ad4e500fa5c5e3b88ea/saleor/core/http_client.py#L8?0, when `HTTPClient.send_request()` doesn't define `timeout=...` (thus `timeout` is `None`) then it will fallback to `default_timeout`.

I recommend to define multiple clients if you want to split timeouts between kinds, e.g.:
- `AsyncHTTPClient = requests_hardened.Manager(...)`
- `SyncHTTPClient = requests_hardened.Manager(...)`

Then in the code-base you can use the adequate client for the use case.

.github/workflows/notify/notify-slack.py
NyanKiyoshi, line 1: That script runs in CIs without Django being installed

tomaszszymanski129:
PR is reworked for the assumptions, only two things left with timeouts they used to (but added connection estab. timeout in those)
- NP atobarai - left 30s as it was in the comment that it was recommended by NP.
- AppFetchManifest mutation 5s for media as the brand data is not essential and mutation can proceed without it.

Let me know if it's advised to change these too or it's fine as is.

patrys (approved):

.github/workflows/notify/notify-slack.py
```diff
@@ -90,7 +91,11 @@ def make_slack_message(self) -> dict:
     def send_notification(self) -> None:
         post_data = self.make_slack_message()
         print(f"Notifying slack with payload: {post_data!r}", file=sys.stderr)
-        response = requests.post(self.slack_endpoint, json=post_data)
+        response = requests.post(
+            self.slack_endpoint,
+            json=post_data,
+            timeout=settings.COMMON_REQUESTS_TIMEOUT,
```
maarcingebala, line 97: This is not part of Django, so should not use these settings IMO.

maarcingebala (commented):
Let's also add a changelog to inform that the default timeout was changed and split into two values.

saleor/app/installation_utils.py
```diff
@@ -96,7 +98,8 @@ def fetch_icon_image(
                 content.write(chunk)
                 if content.tell() > max_file_size:
                     raise ValidationError(size_error_msg, code=code)
-                if (time.monotonic() - fetch_start) > timeout:
+                timeout_in_secs = sum(timeout)
```
NyanKiyoshi, line 101: Here it would make more sense to use `timeout[1]` since it's the read timeout and we are reading

maarcingebala (approved):

Issue #13911, opened 2023-09-07
Set User-Agent header in webhooks


maarcingebala:
Currently, outgoing webhooks have the `User-Agent` header set to `python-requests/2.31.0`, which is the name and version of the Python library used underneath. We should change it to include the "Saleor" name and the version, e.g. `Saleor/3.16`. The current Saleor version can be taken from `saleor.__version__` (stored in `saleor/__init__.py`). The "patch" part of the version number can be dropped.

PR #14094, opened 2023-09-26
change `User-Agent` header for webhooks to `Saleor/version`

Air-t:
I want to merge this change because it changes the `User-Agent` header for HTTP and HTTPS webhook requests

Resolves: https://github.com/saleor/saleor/issues/13911

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


maarcingebala (commented):
This looks good, but I just realized that we should set this header for every outgoing request, otherwise it will be `requests/version` which is not very helpful. We could check all usages of `HTTPClient.send_request` and fix them. In this case maybe the header should be returned via some simple utility function to not construct it manually in each place.

fowczarek (approved):

NyanKiyoshi:
> This looks good, but I just realized that we should set this header for every outgoing request, otherwise it will be `requests/version` which is not very helpful. We could check all usages of `HTTPClient.send_request` and fix them. In this case maybe the header should be returned via some simple utility function to not construct it manually in each place.

Instead of doing this, I suggest adding this to `requests-hardened` project so we are sure it's always set. This is also useful for security (thus in scope of `requests-hardened`) as it allows to not tell what is the server using and which version it is.

You could something similar to https://github.com/saleor/requests-hardened/blob/cb1197d8c0f443121e51f83668b9a6ea4b96ede0/requests_hardened/config.py#L17 where it would be `user_agent_override: Optional[str]` (None => don't set it, str => set a custom one).

This would also avoid having to add PR checks that the user-agent is set everywhere (e.g. semgrep).

maarcingebala (approved):

IKarbowiak (approved):

PR #14123, opened 2023-09-28
Allow add multiple codes per voucher 

SzymJ:
I want to merge this change because it adds the possibility of creating multiple codes for vouchers.

Part of #14409
Docs: [`1000`](https://github.com/saleor/saleor-docs/pull/1000)

### Acceptance Criteria
- [x] The voucher can be created, updated, and deleted only by users and apps with manage discounts permission.
- [x] Allow creating vouchers with multiple codes using the new field `addCodes`.
- [x] Voucher/s queries return a paginated list of codes.
- [x] Ensure that old voucher API(creating/updating voucher code using `code` input) and queries can be handled as previously.
- [x] Voucher usage is calculated as the sum of used codes.
- [x] Allow to set `singleUse` flag to `true` to make codes single-use and deactivate code when it is used.
- [x] Allow to export voucher codes via `exportVoucherCodes` mutation.
- [x] Allow to delete codes using `voucherCodeBulkDelete` mutation.
- [x] Ensure that old vouchers are migrated to new db models.

# Impact

- [x] New migrations
- [x] Deprecated API fields or mutations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


yellowee (approved):

IKarbowiak:
`Test migrations compatibility / build (pull_request) ` is failing but it's handled in https://github.com/saleor/saleor/pull/14150.

IKarbowiak (approved):

jakubkuc (approved):

kadewu (approved):

saleor/csv/utils/export.py
```diff
@@ -89,6 +90,37 @@ def export_gift_cards(
...
+):
+    file_name = get_filename("voucher_code", file_type)
+
+    qs = VoucherCode.objects.all()
+    if voucher_id:
+        qs = VoucherCode.objects.filter(voucher_id=voucher_id)
+    if ids:
+        qs = VoucherCode.objects.filter(id__in=ids)
+
+    export_fields = ["code"]
```
korycins, line 108: Sorry if asking something obvious: 
Don't we want to provide more data here than only `code`?  Like `voucherId`, `usage` etc?
For example, I am generating the CSV file. See one suspicious code. I don't have voucher ID, I have only `code`. 
How can I easily find the voucher on the dashboard/graphql side? I might be wrong but I don't see any filter on `vouchers` query that would allow mi to find the voucher with the specific `code`.

saleor/graphql/core/types/common.py
```diff
@@ -297,6 +298,23 @@ class DiscountError(ProductWithoutVariantError):
         description="List of channels IDs which causes the error.",
         required=False,
     )
+    voucher_codes = NonNullList(
+        graphene.String,
+        description="List of voucher codes which causes the error.",
```
korycins, line 303: Shouldn't we add ADDED_IN to all new fields that extend existing types?

saleor/graphql/csv/mutations/export_voucher_codes.py
```diff
@@ -0,0 +1,113 @@
...
+
+class ExportVoucherCodes(BaseExportMutation):
+    class Arguments:
+        input = ExportVoucherCodesInput(
+            required=True, description="Fields required to export voucher codes."
+        )
+
+    class Meta:
+        description = (
+            "Export active voucher codes to csv/xlsx file."
```
korycins, line 48: `active voucher codes` - is it correct? Currently csv logic exports `all()` codes. without checking `is_active` flag on VoucherCode model.

saleor/graphql/discount/mutations/voucher/voucher_create.py
```diff
@@ -68,6 +84,14 @@ class VoucherInput(BaseInputObjectType):
     only_for_staff = graphene.Boolean(
         description="Voucher can be used only by staff user."
     )
+    single_use = graphene.Boolean(
+        description=(
+            "When set to 'True', each voucher is limited to a single use; "
+            "otherwise, usage remains unrestricted."
```
korycins, line 90: > usage remains unrestricted

This is not clear to me. If the usage remains unrestricted, how `usage_limit` behave in that case?
So assuming, I have `single_use`:False, and `usageLimit`:5. The usage should be limited to 5, right?

saleor/graphql/discount/mutations/voucher/voucher_create.py
```diff
@@ -98,10 +122,19 @@ class Meta:
...
-        code = data.get("code", None)
-        if code == "":
+        cls.clean_codes(data)
+        cleaned_input = super().clean_input(info, instance, data, **kwargs)
+
+        return cleaned_input
+
+    @classmethod
+    def _clean_old_code(cls, data):
+        # Deprecated in 3.0, remove in 4.0
```
korycins, line 132: Have you created a gh issue for that? It should be assigned to Saleor4.0 Milestione: https://github.com/saleor/saleor/milestone/80

saleor/graphql/discount/mutations/voucher/voucher_create.py
```diff
@@ -110,27 +143,152 @@ def clean_input(cls, info: ResolveInfo, instance, data, **kwargs):
...
+            cls._clean_old_code(data)
+
+        if "add_codes" in data:
+            cls._clean_new_codes(data)
+
+    @classmethod
+    def construct_codes_instances(
+        cls, code, codes_data, cleaned_input, voucher_instance
+    ):
+        if code:
```
korycins, line 209: Are you sure, that the order is correct here? 
I would say that we should use `codes_data` first. If missing, then use it as a fallback `code`. IMHO We should always use deprecated fields at the end.

korycins: Also, you should create single codes list based on `code`+`add_codes` OR provide the proper description in the input that will tell, when `code` or `add_codes` are ignored.

saleor/graphql/discount/mutations/voucher/voucher_delete.py
```diff
@@ -38,6 +38,24 @@ def success_response(cls, instance):
...
+    @classmethod
+    def perform_mutation(  # type: ignore[override]
+        cls, _root, info: ResolveInfo, /, *, external_reference=None, id=None
+    ):
+        """Perform a mutation that deletes a model instance."""
+        instance = cls.get_instance(info, external_reference=external_reference, id=id)
+
+        cls.clean_instance(info, instance)
+        db_id = instance.id
+        code = instance.codes.first().code
```
korycins, line 54: What if the voucher will have multiple codes?

saleor/graphql/discount/mutations/voucher/voucher_update.py
```diff
@@ -33,6 +39,96 @@ class Meta:
...
+                            code=DiscountErrorCode.VOUCHER_ALREADY_USED.value,
+                        )
+                    }
+                )
+
+    @classmethod
+    def construct_codes_instances(
+        cls, code, codes_data, cleaned_input, voucher_instance
+    ):
+        if codes_data:
```
korycins, line 88: Here you have different order of picking up the codes than in `VoucherCreate`. It means that it will behave differently for Create&Update.

saleor/graphql/discount/mutations/voucher/voucher_delete.py
korycins: I understand that you're passing the code to match the static webhook payload, but what about subscription `VoucherDeleted`? The codes don't exist as you deleted the `Voucher`, so I will not be able to fetch the codes that were removed.  Is it something that we should take into account?
Also, I think we should double-check (if not yet checked) the subscription query for `VoucherDeleted` and requested `codes` field

saleor/graphql/discount/types/vouchers.py
```diff
@@ -60,12 +63,41 @@ def resolve_channel(root: models.VoucherChannelListing, info: ResolveInfo):
...
+    id = graphene.GlobalID(required=True, description="The ID of the voucher code.")
+    code = graphene.String(description="Code to use the voucher.")
+    used = graphene.Int(description="Number of times a code has been used.")
+    is_active = graphene.Boolean(description="Whether a code is active or not.")
+    created_at = graphene.DateTime(
+        required=True, description="Date time of code creation."
+    )
+
+    class Meta:
+        description = "Represents voucher code." + ADDED_IN_318
```
korycins, line 76: Question: Shouldn't be in preview_feature?

saleor/graphql/discount/types/vouchers.py
```diff
@@ -81,6 +113,13 @@ class Voucher(ChannelContextTypeWithMetadata[models.Voucher]):
         description="Determine if the voucher usage should be limited to one use per "
         "customer.",
     )
+    single_use = graphene.Boolean(
+        required=True,
+        description=(
+            "Determine if the voucher is limited to a single use or unlimited use."
```
korycins, line 119: > unlimited use
What does it mean in terms of having `single_use`:False & `usage_limit`:5?

saleor/graphql/discount/mutations/voucher/voucher_create.py
SzymJ: we have `validate_one_of_args_is_in_mutation` do you can't send `code` and `add_codes` together,  so there will always be one list

saleor/graphql/discount/mutations/voucher/voucher_update.py
SzymJ: API doesn't allow to send `code` and `add_codes` together so the behavior will be the same

saleor/graphql/csv/mutations/export_voucher_codes.py
SzymJ: yes this is misleading information, we should not use `active` word in description

saleor/csv/utils/export.py
SzymJ: @korycins As an MVP we just decided to keep it simple and allow to export only codes. If it will be needed we will add filters later.

saleor/graphql/discount/mutations/voucher/voucher_create.py
IKarbowiak: We have planty of places with deprecated data in code. All fields with `DEPRECATED_IN` label so I don't think it's required to create a ticket for each deprecation. I suppose we will drop all deprecated fields at once. 
And this method is for cleaning deprecated field data.

saleor/graphql/discount/mutations/voucher/voucher_delete.py
SzymJ: A talk with @IKarbowiak and we will add events related to voucher codes as separate PR(https://github.com/saleor/saleor/issues/14495) after we merge this. I don't think there is an easy way to reassign deleted related objects to voucher. So we think a dedicated webhook is the best option. What do you think @korycins ?

korycins: I just wanted to confirm that we have taken into account this. If you have addressed this then it is ok for me.

saleor/graphql/discount/mutations/voucher/voucher_update.py
korycins: Yeah, miss the validation of arguments!

michal-macioszczyk (approved):

saleor/checkout/complete_checkout.py
```diff
@@ -1516,7 +1551,11 @@ def complete_checkout_with_payment(
         lines, _ = fetch_checkout_lines(checkout, skip_recalculation=True)
         checkout_info = fetch_checkout_info(checkout, lines, manager)
 
-        checkout_info.checkout.voucher_code = None
+        # reassign voucher data that was used during payment process to allow voucher
```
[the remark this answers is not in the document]
IKarbowiak: Done

Issue #14112, opened 2023-09-27
Optimize loading webhooks for event


maarcingebala:
The `get_webhooks_for_event` should be rewritten into a dataloader. Currently, there is a performance issue visible in bulk mutations where webhooks are triggered in a loop for each instance modified in the bulk mutation, and for each instance, it's running `get_webhooks_for_event`. Each call triggers a database query:

```
SELECT * FROM webhook_webhook INNER JOIN app_app ON ( webhook_webhook . app_id = ? . id ) WHERE ( webhook_webhook . app_id IN ( SELECT U0. id FROM app_app U0 INNER JOIN app_app_permissions U1 ON ( U0. id = U1. app_id ) INNER JOIN permission_permission U2 ON ( U1. permission_id = U2. id ) INNER JOIN django_content_type U3 ON ( U2. content_type_id = U3. id ) WHERE ( U0. is_active AND U2. codename = ? AND U3. app_label = ? ) ) AND webhook_webhook . is_active AND EXISTS ( SELECT ( ? ) FROM webhook_webhookevent U0 WHERE ( U0. event_type IN ( ? ) AND U0. webhook_id = ? . id ) LIMIT ? ) ) ORDER BY webhook_webhook . id ASC
```

maarcingebala:
Example trace: https://app.datadoghq.com/apm/trace/15459061078129668117?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=4396032202316836865&spanViewType=metadata

maarcingebala:
Instead of adding a dataloader at this point, we optimized this function by calling it at the resolver level and passing webhooks all the way down to webhook trigger functions. This way they are fetched once per webhook event which highly reduces number of duplicated queries for bulk mutations. Refactoring this function into a dataloader can be done in a separate PR (https://github.com/saleor/saleor/issues/14267).

PR #14240, opened 2023-10-10
Optimize webhooks calls in bulk mutations

Air-t:
I want to merge this change because it reduces the number of webhook queries for bulk mutations

resolves: https://github.com/saleor/saleor/issues/14112

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/bulk_mutations/customer_bulk_delete.py
+++ b/saleor/graphql/account/bulk_mutations/customer_bulk_delete.py
@@ -3,6 +3,7 @@ import graphene
 from ....account import models
 from ....permission.enums import AccountPermissions
 from ....webhook.event_types import WebhookEventAsyncType
+from ....webhook.utils import get_webhooks_for_event
 from ...core import ResolveInfo
 from ...core.doc_category import DOC_CATEGORY_USERS
 from ...core.mutations import ModelBulkDeleteMutation
@@ -49,6 +50,7 @@ class CustomerBulkDelete(CustomerDeleteMixin, UserBulkDelete):
     def bulk_action(cls, info: ResolveInfo, queryset, /):
         instances = list(queryset)
         queryset.delete()
-        manager = get_plugin_manager_promise(info.context).get()
-        for instance in instances:
-            manager.customer_deleted(instance)
+        if webhooks := get_webhooks_for_event(WebhookEventAsyncType.CUSTOMER_DELETED):
+            manager = get_plugin_manager_promise(info.context).get()
+            for instance in instances:
+                manager.customer_deleted(instance, webhooks=webhooks)
--- a/saleor/graphql/account/bulk_mutations/customer_bulk_delete.py
+++ b/saleor/graphql/account/bulk_mutations/customer_bulk_delete.py
@@ -50,7 +50,7 @@ class CustomerBulkDelete(CustomerDeleteMixin, UserBulkDelete):
     def bulk_action(cls, info: ResolveInfo, queryset, /):
         instances = list(queryset)
         queryset.delete()
-        if webhooks := get_webhooks_for_event(WebhookEventAsyncType.CUSTOMER_DELETED):
-            manager = get_plugin_manager_promise(info.context).get()
-            for instance in instances:
-                manager.customer_deleted(instance, webhooks=webhooks)
+        webhooks = get_webhooks_for_event(WebhookEventAsyncType.CUSTOMER_DELETED)
+        manager = get_plugin_manager_promise(info.context).get()
+        for instance in instances:
+            manager.customer_deleted(instance, webhooks=webhooks)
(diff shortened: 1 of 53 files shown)


saleor/graphql/account/bulk_mutations/customer_bulk_update.py
```diff
@@ -556,8 +557,16 @@ def post_save_actions(cls, info, manager, instances, old_instances):
         app = get_app_promise(info.context).get()
         staff_user = info.context.user
         users_with_name_or_email_updated = []
+        webhooks_meta = get_webhooks_for_event(
+            WebhookEventAsyncType.CUSTOMER_METADATA_UPDATED
+        )
```
IKarbowiak, line 562: Can we move it to line 632 to call it only if needed?

saleor/graphql/account/bulk_mutations/customer_bulk_update.py
```diff
@@ -556,8 +557,16 @@ def post_save_actions(cls, info, manager, instances, old_instances):
...
+        webhooks_updated = get_webhooks_for_event(
+            WebhookEventAsyncType.CUSTOMER_UPDATED
+        )
```
IKarbowiak, line 565: Similar here to line 567

zedzior (approved):

saleor/graphql/account/bulk_mutations/customer_bulk_delete.py
```diff
@@ -49,6 +50,7 @@ def perform_mutation(cls, root, info: ResolveInfo, /, **data):
     def bulk_action(cls, info: ResolveInfo, queryset, /):
         instances = list(queryset)
         queryset.delete()
-        manager = get_plugin_manager_promise(info.context).get()
-        for instance in instances:
-            manager.customer_deleted(instance)
+        if webhooks := get_webhooks_for_event(WebhookEventAsyncType.CUSTOMER_DELETED):
+            manager = get_plugin_manager_promise(info.context).get()
+            for instance in instances:
+                manager.customer_deleted(instance, webhooks=webhooks)
```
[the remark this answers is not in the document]
SzymJ: same, I think is a good idea, because as I remember when `call_event` is used we are sure that webhook is sent after the transaction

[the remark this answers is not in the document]
Air-t: sure, but this seems to be a part of a [task](https://github.com/orgs/saleor/projects/16/views/13?pane=issue&itemId=39527121)

saleor/graphql/account/bulk_mutations/customer_bulk_update.py
Air-t: moving this to line 632 may result in using this query multiple times cause it will be in `for` loop - I have no better idea for now - but maybe getting all webhooks for both events at once, then filtering them might be a better approach (this might be a next step improvement)

saleor/graphql/account/bulk_mutations/customer_bulk_delete.py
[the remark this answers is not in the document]
maarcingebala: Yes, we want to tackle that separately.

saleor/graphql/account/bulk_mutations/customer_bulk_update.py
maarcingebala: As a next step we could make `get_webhooks_for_event` a dataloader where webhook event names would be the keys.

maarcingebala: @Air-t Maybe open a task for this, and let's add it to Q4 roadmap so we could try to tackle it later.

SzymJ (approved):

maarcingebala (approved):

IKarbowiak (approved):

saleor/graphql/warehouse/tests/benchmark/test_stock_bulk_update.py
```diff
@@ -95,7 +95,7 @@ def test_stocks_bulk_update_queries_count(
     ]
 
     # Test number of queries when multiple objects are updated
-    with django_assert_num_queries(9):
+    with django_assert_num_queries(10):
```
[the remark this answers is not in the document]
Air-t: I'm not sure why the query to get all webhooks was not visible before, however now it is visible

saleor/graphql/account/bulk_mutations/customer_bulk_delete.py
```diff
@@ -49,6 +50,7 @@ def perform_mutation(cls, root, info: ResolveInfo, /, **data):
     def bulk_action(cls, info: ResolveInfo, queryset, /):
         instances = list(queryset)
         queryset.delete()
-        manager = get_plugin_manager_promise(info.context).get()
-        for instance in instances:
-            manager.customer_deleted(instance)
+        if webhooks := get_webhooks_for_event(WebhookEventAsyncType.CUSTOMER_DELETED):
```
korycins, line 53: Any `if` like this one directly in the Saleor code (not in plugin part) will cause, that you will not be able to use these handlers inside your own the custom plugin without declaring dummy webhook.

maarcingebala (approved):

PR #14242, opened 2023-10-10
Move migration tasks to migration code

IKarbowiak:
Move migration tasks from 3.16 to migration code

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/attribute/migrations/0034_assignedpageattributevalue_page_data_migration.py
```diff
@@ -1,18 +1,16 @@
 # Generated by Django 3.2.20 on 2023-08-24 11:25
 
-from django.db import migrations
-from django.db.models.signals import post_migrate
-from django.apps import apps as registry
+from django.db import migrations, transaction, connection
 
-from .tasks.saleor3_16 import assign_pages_to_attribute_values_task
 
+PAGE_BATCH_SIZE = 5000
```
8r2y5, line 6: Could you leave comment why 5000 is a good number here?

IKarbowiak: I do not have such info. I'm only moved the task to migration code and in the task the info was:

```
# batch size to make sure that task is completed in 1 second
# and we don't use too much memory
```

So I don't think it's useful 😄

michal-macioszczyk: yes, agree this comment is not useful ;) nonetheless it was tested :) 
https://github.com/saleor/saleor/pull/13814/files#r1310035818

michal-macioszczyk (approved):

saleor/attribute/migrations/0034_assignedpageattributevalue_page_data_migration.py
8r2y5: Please do not use magic numbers and please leave info about ram usage

tomaszszymanski129: This is a magic number as it is, we shouldn't leave it like that. As this part is only moving files I would skip measuring the RAM (although it should have been tested in original PR). But I would at least leave the comment about why this is 5k (because it completes within 1s). If it ever causes issues with the performance it's good to know at least what the author had in mind when setting it to 5k.

kadewu (approved):

kadewu (approved):

szczecha (approved):

PR #14257, opened 2023-10-11
Enforce decision about migrations performance test

fowczarek:
I want to merge this change because it enforces the decision of whether a change should have a migration performance test if migrations have been changed.

After this change, anyone who modifies or creates migration files should explicitly add  one for labels:
- `skip migrations perf test` - whenever migrations don't require performance tests(e.g. when migrations are run in celery tasks)
- `migrations perf test` - whenever migrations require performance tests 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/.github/workflows/migrations-perf-test-check.yml
+++ b/.github/workflows/migrations-perf-test-check.yml
@@ -1,0 +1,24 @@
+name: "Migration performance test required"
+on:
+  pull_request:
+    branches:
+      - "main"
+    paths:
+      - "**/migrations/**"
+    types: [labeled, unlabeled, opened, closed]
+
+jobs:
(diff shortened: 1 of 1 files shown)


IKarbowiak (approved):

.github/workflows/migrations-test-check.yml
NyanKiyoshi, line 1: I don't understand the point of this workflow: it does nothing. Can you explain?

fowczarek: I added more info in the Job description in the workflow file.

NyanKiyoshi (approved):

Air-t (approved):

PR #14270, opened 2023-10-12
Move update product search vector to celery task

jakubkuc:
I want to merge this change because it changes how the product search vector is updated in mutations. Instead of updating it synchronously in mutation the field `search_index_dirty` will be set to `True` and the search vector will be updated asynchronously with celery beat task `update_products_search_vector_task`. 

Synchronous updates caused multiple locks on db when updating many products in a short amount of time causing in long waiting time for locks to be freed.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/mutations/product/product_create.py
+++ b/saleor/graphql/product/mutations/product/product_create.py
@@ -9,7 +9,6 @@ from .....core.utils.editorjs import clean_editor_js
 from .....permission.enums import ProductPermissions
 from .....product import models
 from .....product.error_codes import ProductErrorCode
-from .....product.search import update_product_search_vector
 from .....product.tasks import update_product_discounted_price_task
 from ....attribute.types import AttributeValueInput
 from ....attribute.utils import AttributeAssignmentMixin, AttrValuesInput
@@ -220,7 +219,8 @@ class ProductCreate(ModelMutation):
     @classmethod
     def post_save_action(cls, info: ResolveInfo, instance, _cleaned_input):
         product = models.Product.objects.prefetched_for_webhook().get(pk=instance.pk)
-        update_product_search_vector(instance)
+        product.search_index_dirty = True
+        product.save(update_fields=["search_index_dirty"])
         update_product_discounted_price_task.delay(instance.id)
         manager = get_plugin_manager_promise(info.context).get()
         cls.call_event(manager.product_created, product)
--- a/saleor/graphql/product/mutations/product/product_update.py
+++ b/saleor/graphql/product/mutations/product/product_update.py
@@ -5,7 +5,6 @@ import graphene
 from .....attribute import models as attribute_models
 from .....permission.enums import ProductPermissions
 from .....product import models
-from .....product.search import update_product_search_vector
 from .....product.tasks import update_product_discounted_price_task
 from ....attribute.utils import AttributeAssignmentMixin, AttrValuesInput
 from ....core import ResolveInfo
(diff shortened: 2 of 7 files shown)


saleor/graphql/product/tests/mutations/test_product_variant_bulk_update.py
```diff
@@ -104,8 +104,10 @@ def test_product_variant_bulk_update(
     content = get_graphql_content(response)
     flush_post_commit_hooks()
     data = content["data"]["productVariantBulkUpdate"]
+    product_with_single_variant.refresh_from_db()
```
NyanKiyoshi, line 107: Slightly faster:

```suggestion
    product_with_single_variant.refresh_from_db(fields=["search_index_dirty"])
```

zedzior (approved):

NyanKiyoshi (approved):
The function `update_product_search_vector` is now no longer used (except in a test case that it is testing `update_product_search_vector()`), can't we remove the function?

jakubkuc:
@NyanKiyoshi I was thinking about that but I wasn't sure if it shouldn't be dropped in a separate PR since it will remove some tests also and would be more clear this way.

saleor/graphql/product/mutations/product/product_create.py
```diff
@@ -220,7 +219,8 @@ def _save_m2m(cls, _info: ResolveInfo, instance, cleaned_data):
     @classmethod
     def post_save_action(cls, info: ResolveInfo, instance, _cleaned_input):
         product = models.Product.objects.prefetched_for_webhook().get(pk=instance.pk)
-        update_product_search_vector(instance)
+        product.search_index_dirty = True
+        product.save(update_fields=["search_index_dirty"])
```
maarcingebala, line 223: For the product mutations we already run the `save` function above. It should be enough to add:

```
product.search_index_dirty = True
```

in line 208 before `instance.save()` happens.

saleor/graphql/product/mutations/product/product_update.py
```diff
@@ -53,7 +52,8 @@ def clean_attributes(
     @classmethod
     def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
         product = models.Product.objects.prefetched_for_webhook().get(pk=instance.pk)
-        update_product_search_vector(instance)
+        product.search_index_dirty = True
+        product.save(update_fields=["search_index_dirty"])
```
maarcingebala, line 56: This might be not needed if we changed the `save` method of `ProductCreate` mutation (see comment above).

NyanKiyoshi:
> @NyanKiyoshi I was thinking about that but I wasn't sure if it shouldn't be dropped in a separate PR since it will remove some tests also and would be more clear this way.

Any thought @saleor/core?

jakubkuc:
Edit: Dropped the function and moved setting flag to `save` method to prevent multiple calls to db

maarcingebala (approved):

saleor/graphql/product/mutations/product/product_update.py
```diff
@@ -53,7 +52,6 @@ def clean_attributes(
     @classmethod
     def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
         product = models.Product.objects.prefetched_for_webhook().get(pk=instance.pk)
-        update_product_search_vector(instance)
```
NyanKiyoshi, line 56: Can you add a test case for this mutation? It's quite hidden that `is_dirty` is set to `True`

jakubkuc: I've added a check for this field to `product_update` and `product_create tests`.

NyanKiyoshi (approved):

PR #14286, opened 2023-10-13
Drop recursion from migrations and move migration tasks to migration code

IKarbowiak:
Port of https://github.com/saleor/saleor/pull/14269 and https://github.com/saleor/saleor/pull/14242

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/attribute/migrations/0034_assignedpageattributevalue_page_data_migration.py
```diff
@@ -1,18 +1,52 @@
 # Generated by Django 3.2.20 on 2023-08-24 11:25
 
-from django.db import migrations
-from django.db.models.signals import post_migrate
-from django.apps import apps as registry
+from django.db import migrations, transaction, connection
 
-from .tasks.saleor3_16 import assign_pages_to_attribute_values_task
 
+PAGE_BATCH_SIZE = 5000
```
8r2y5, line 6: Please do not use magic numbers. Could you please provide information about ram usage on that batch size?

IKarbowiak: https://github.com/saleor/saleor/pull/14242#discussion_r1352279239

IKarbowiak: It's the same case as before, it's port of the changes that were moved from task

SzymJ (approved):

saleor/attribute/migrations/0034_assignedpageattributevalue_page_data_migration.py
```diff
@@ -1,18 +1,52 @@
...
+                FOR UPDATE
+            )
+            UPDATE attribute_assignedpageattributevalue AS av
+            SET page_id = apa.page_id
+            FROM attribute_assignedpageattribute AS apa
+            WHERE av.id IN (SELECT id FROM limited)
+            AND av.assignment_id = apa.id;
+            """,  # noqa
+                [PAGE_BATCH_SIZE],
+            )
```
tomaszszymanski129, line 49: Instead of making extra SELECT each loop in line 13, cannot we just determine here if there is a need to continue looping? You have a variable here `cursor.rowcount` which can tell you how many rows were affected. When that is zero we can break, WDYT? same would go with `update_product_assignment` below.

IKarbowiak: I've only moved those mutations from the task code, and I don't want to adjust them, unless it's really necessary.

maarcingebala: Depends on what is the performance gain. If it's a cosmetic change, I think we can keep it as it is right now, but if the performance gain is significant, then we should look into making this improvement here and in other places.

IKarbowiak: For me it sounds more like cosmetic change. Shouldn't affect the performance much

aniav (approved):

Issue #13657, opened 2023-08-09
Reduce the number of database fetch count from PluginManager


NyanKiyoshi:
The plugin configurations are fetched for every GraphQL requests which scales poorly and wastes resources: https://github.com/saleor/saleor/blob/65349faa97dc6d4666813719c7fb03b51389bf3e/saleor/plugins/manager.py#L158-L162

Figure out a way to reduce the number of calls (e.g. caching)

maarcingebala:
Much of that time is spent in Python code, not DB calls.

- PluginsManager fetches all channels from the (writer) database during initialization and then, once again, fetches channels from the (reader) database in `_get_db_plugin_configs`:
        1. https://github.com/saleor/saleor/blob/0d21c08210ea9e1785d0f19638840cca35047a23/saleor/plugins/manager.py#L143
        2. https://github.com/saleor/saleor/blob/0d21c08210ea9e1785d0f19638840cca35047a23/saleor/plugins/manager.py#L163

- We need to profile PluginsManager with representative data (e.g. has 100 plugin configurations across 25 channels, it’s fast when we test it with just one plugin and one channel)

maarcingebala:
- Drop prefetch_related in favor of map in Python.
- Use the reader database instead of the writer.

PR #14340, opened 2023-10-17
Optimize queries to database when initializing PluginsManager

8r2y5:
I want to merge this change because it fixes https://github.com/saleor/saleor/issues/13657 by fetching all channels and creating map of them, so it could be used instead of fetching channel from database each time it is needed.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/plugins/mutations.py
+++ b/saleor/graphql/plugins/mutations.py
@@ -99,7 +99,9 @@ class PluginUpdate(BaseMutation):
         plugin_id = cleaned_data["plugin"].PLUGIN_ID
         channel_slug = cleaned_data["channel_slug"]
         input_data = cleaned_data["data"]
+
         manager = get_plugin_manager_promise(info.context).get()
         manager.save_plugin_configuration(plugin_id, channel_slug, input_data)
-        manager = get_plugins_manager()
+
+        manager = get_plugins_manager(allow_replica=info.context.allow_replica)
         return PluginUpdate(plugin=resolve_plugin(plugin_id, manager))
--- a/saleor/graphql/plugins/mutations.py
+++ b/saleor/graphql/plugins/mutations.py
@@ -103,5 +103,5 @@ class PluginUpdate(BaseMutation):
         manager = get_plugin_manager_promise(info.context).get()
         manager.save_plugin_configuration(plugin_id, channel_slug, input_data)
 
-        manager = get_plugins_manager(allow_replica=info.context.allow_replica)
+        manager = get_plugins_manager(allow_replica=False)
         return PluginUpdate(plugin=resolve_plugin(plugin_id, manager))
--- a/saleor/plugins/manager.py
+++ b/saleor/plugins/manager.py
@@ -101,6 +101,14 @@ class PluginsManager(PaymentInterface):
     global_plugins: List["BasePlugin"] = []
     all_plugins: List["BasePlugin"] = []
 
+    @property
+    def database(self):
+        return (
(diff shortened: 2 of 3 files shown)


saleor/core/utils/random_data.py
```diff
@@ -574,7 +574,7 @@ def create_fake_payment(mock_notify, order):
         total=order.total.gross.amount,
         currency=order.total.gross.currency,
     )
-    manager = get_plugins_manager()
+    manager = get_plugins_manager(allow_replica=False)
```
maarcingebala, line 577: Why are we disallowing replicas here? Of all the places that this PR touches, I think only the `PluginUpdate` mutation can affect the list of loaded plugins, so there we should disallow replica, but in other places, we should be able to assume that the replica is up-to-date and can be used.

saleor/graphql/plugins/mutations.py
```diff
@@ -99,7 +99,9 @@ def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
         plugin_id = cleaned_data["plugin"].PLUGIN_ID
         channel_slug = cleaned_data["channel_slug"]
         input_data = cleaned_data["data"]
+
         manager = get_plugin_manager_promise(info.context).get()
         manager.save_plugin_configuration(plugin_id, channel_slug, input_data)
-        manager = get_plugins_manager()
+
+        manager = get_plugins_manager(allow_replica=info.context.allow_replica)
```
maarcingebala, line 106: I'm wondering if here we shouldn't explicitly pass `False` but maybe let's see what others think.

saleor/core/utils/random_data.py
8r2y5: I assumed that whenever write to database we should have up-to-date results.

saleor/graphql/plugins/mutations.py
8r2y5: In such case, should we also force `allow_replica=False` [here](https://github.com/saleor/saleor/pull/14340/files/204cab3c7a82ace09dc0bab2261e7b0524db3788#diff-8c0b159a7dafb59bbe23bc4b82fb8e294a547e5171900b41bd3c696386d474afR103) too?

fowczarek (approved):

IKarbowiak (approved):

maarcingebala (approved):

saleor/graphql/plugins/mutations.py
```diff
@@ -99,7 +99,9 @@ def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
         plugin_id = cleaned_data["plugin"].PLUGIN_ID
         channel_slug = cleaned_data["channel_slug"]
         input_data = cleaned_data["data"]
+
         manager = get_plugin_manager_promise(info.context).get()
         manager.save_plugin_configuration(plugin_id, channel_slug, input_data)
-        manager = get_plugins_manager()
+
+        manager = get_plugins_manager(allow_replica=False)
```
maarcingebala, line 106: @8r2y5 One more thing, I checked with @fowczarek, and all DB calls in mutation should be done using writer. Can you verify that both calls to `get_plugin_manager_promise` are using the writer here? I think this will require enabling replicas locally.

8r2y5: Tested locally with replica enabled, it is selecting writer when using `get_plugin_manager_promise`

PR #14355, opened 2023-10-18
Automate Docker metadata calculation

patrys:
This should give us much nicer tags and labels for Docker images

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/.github/workflows/publish-containers.yml
+++ b/.github/workflows/publish-containers.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       # Outputs the name of the repository (owner/repo)
       - name: Build Image Name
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-containers.yml
```diff
@@ -24,68 +24,17 @@ jobs:
...
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ${{ steps.image.outputs.image_name }}
+          tags: |
+            type=ref,event=branch,prefix=unstable-
+            type=pep440,pattern={{version}}
+            type=pep440,pattern={{major}}.{{minor}}
```
NyanKiyoshi, line 36: Shouldn't it be like so?
```suggestion
            type=pep440,event=tag,pattern={{version}}
            type=pep440,event=tag,pattern={{major}}.{{minor}}
```

.github/workflows/publish-containers.yml
```diff
@@ -24,68 +24,17 @@ jobs:
...
+            type=sha
```
NyanKiyoshi, line 37: I don't believe we need `sha-...` tags, it will cause to have a lot of tags

.github/workflows/publish-containers.yml
patrys: `event=tag` is a parameter for `type=ref`, it generates a label when a new reference type is added (tag, head, etc.). `type=pep440` only works on tags.

aniav (approved):

fowczarek (approved):

maarcingebala (approved):

Issue #14326, opened 2023-10-17
Bug: Incorrect chargeStatus when using transactions and grantedRefunds


korycins:
### What are you trying to achieve?

I want to have the correct status of the order.chargeStatus. Currently, the status is not updated when modifying/creating the grantedRefund. When refund is requested, the status is updated but the order.chargeStatus shows incorrect status 

### Steps to reproduce the problem

1. Create order. Order needs to have transactions attached to it
2. Create granted refund with any amount. 
3. See order.chargeStatus. The sum of order.transactions[].chargedAmount is bigger than order.total - sum of grantedRefunds, the status should be overcharged. 

---
4. Call send refund button, and provide the same amount as the one attached to the granted refund.
5. When paymentApp returns the refund_success event with the same amount, the chargeStatus should be FULL (as sum of the order.transactions[].chargedAmount  is equal to order.total - sum of grantedRefunds), currently it shows overcharged

### What did you expect to happen?

I expect that all chargeStatus will display the correct status when handling granted refunds and refund requests:

chargeStatus == OVERCHARGED when order.total-grantedRefund amount <= transactions[].chargedAmount
chargeStatus == FULL when order.total-grantedRefund amount == transactions[].chargedAmount
chargeStatus == PARTIAL when order.total-grantedRefund amount > transactions[].chargedAmount and bigger than 0
chargeStatus == None when order.total-grantedRefund amount == 0

### Environment

Saleor version: 3.13+
OS and version: …

PR #14371, opened 2023-10-19
Fix order.chargeStatus calculated incorrectly

korycins:
I want to merge this change because it fixes the issue reported here: https://github.com/saleor/saleor/issues/14326

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/tests/mutations/test_order_grant_refund_create.py
+++ b/saleor/graphql/order/tests/mutations/test_order_grant_refund_create.py
@@ -1,7 +1,6 @@
 from decimal import Decimal
 
-from saleor.order.utils import update_order_charge_data
-
+from .....order.utils import update_order_charge_data
 from ....core.utils import to_global_id_or_none
 from ....tests.utils import assert_no_permission, get_graphql_content
 from ...enums import OrderChargeStatusEnum, OrderGrantRefundCreateErrorCode
--- a/saleor/graphql/order/tests/mutations/test_order_grant_refund_create.py
+++ b/saleor/graphql/order/tests/mutations/test_order_grant_refund_create.py
@@ -1,8 +1,10 @@
 from decimal import Decimal
 
+from saleor.order.utils import update_order_charge_data
+
 from ....core.utils import to_global_id_or_none
 from ....tests.utils import assert_no_permission, get_graphql_content
-from ...enums import OrderGrantRefundCreateErrorCode
+from ...enums import OrderChargeStatusEnum, OrderGrantRefundCreateErrorCode
 
 ORDER_GRANT_REFUND_CREATE = """
 mutation OrderGrantRefundCreate(
@@ -26,6 +28,7 @@ mutation OrderGrantRefundCreate(
     }
     order{
       id
+      chargeStatus
       grantedRefunds{
         id
         amount{
@@ -66,7 +69,6 @@ def test_grant_refund_by_user(staff_api_client, permission_manage_orders, order)
     response = staff_api_client.post_graphql(ORDER_GRANT_REFUND_CREATE, variables)
 
     # then
-
     content = get_graphql_content(response)
     data = content["data"]["orderGrantRefundCreate"]
     errors = data["errors"]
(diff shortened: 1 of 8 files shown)


saleor/order/migrations/tasks/saleor3_13.py
```diff
@@ -46,3 +51,62 @@ def order_events_rename_transaction_capture_events_task():
     if ids:
         update_type_to_transaction_charge_requested(qs)
         order_events_rename_transaction_capture_events_task.delay()
+
+
+@app.task
+def update_orders_charge_statuses_task():
```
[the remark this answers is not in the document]
IKarbowiak: You should leave this comment in code. 5-7 sec for single celery task or for whole update?

saleor/order/migrations/tasks/saleor3_13.py
```diff
@@ -46,3 +51,63 @@ def order_events_rename_transaction_capture_events_task():
...
+    orders = Order.objects.filter(
+        Exists(OrderGrantedRefund.objects.filter(order_id=OuterRef("pk")))
+    )
+    orders = orders.annotate(
+        refunded_amount=Coalesce(
+            Sum("granted_refunds__amount_value"),
```
kadewu, line 63: Did you check maybe if doing `Subquery` and annotate there have better performance?

saleor/order/migrations/tasks/saleor3_13.py
```diff
@@ -46,3 +51,63 @@ def order_events_rename_transaction_capture_events_task():
...
+            0,
+            output_field=DecimalField(),
+        ),
+        current_total_gross=F("total_gross_amount") - F("refunded_amount"),
+    )
+    trigger_task = False
+
+    overcharged_orders = orders.filter(
+        total_charged_amount__gt=F("current_total_gross")
+    ).exclude(charge_status=OrderChargeStatus.OVERCHARGED)[:BATCH_SIZE]
```
kadewu, line 73: I wonder if hitting database every time for each case is correct approach here. `orders` query above has some cost already. If we just load `orders[:BATCH_SIZE]` into memory and iterate through it in python to deduce witch `order_id`'s should use to update it may be safer if I'm not wrong. WDYT?

saleor/order/migrations/tasks/saleor3_13.py
```diff
@@ -46,3 +51,70 @@ def order_events_rename_transaction_capture_events_task():
...
+
+
+@app.task
+def update_orders_charge_statuses_task(stop_granted_refund_pk=-1):
+    """Update the charge status for orders with granted refunds.
+
+    Task takes around 0.3 seconds for 1000 orders and around 5 MB of memory.
+    """
+    if stop_granted_refund_pk == -1:
+        stop_granted_refund_pk = OrderGrantedRefund.objects.count()
```
IKarbowiak, line 97: The `pk` might not start from `0` and then you will miss some objects.

kadewu: I believe you should hit `0`, cause we iterate here backwards.

saleor/order/migrations/tasks/saleor3_13.py
```diff
@@ -46,3 +51,70 @@ def order_events_rename_transaction_capture_events_task():
...
+        order.charge_status = OrderChargeStatus.OVERCHARGED
+
+
+@app.task
+def update_orders_charge_statuses_task(stop_granted_refund_pk=-1):
+    """Update the charge status for orders with granted refunds.
+
+    Task takes around 0.3 seconds for 1000 orders and around 5 MB of memory.
+    """
+    if stop_granted_refund_pk == -1:
```
kadewu, line 96: I can see one edge-case with this approach, e.g: we have 999 `OrderGrantedRefund`s , we go through that task with: `start_pk = -1`, this is going to work here, but we will then hit infinity loop.

So we need to add condition to not reshedule task if start_pk < 0.

IKarbowiak (approved):

kadewu (approved):

saleor/graphql/order/tests/mutations/test_order_grant_refund_create.py
```diff
@@ -1,8 +1,10 @@
 from decimal import Decimal
 
+from saleor.order.utils import update_order_charge_data
```
fowczarek, line 3: Invalid import type 🚓

saleor/order/migrations/tasks/saleor3_13.py
```diff
@@ -46,3 +51,72 @@ def order_events_rename_transaction_capture_events_task():
...
+
+    orders_to_update = []
+    last_number = number
+    for o in orders:
+        granted_refund_amount = sum(
+            [refund.amount.amount for refund in o.granted_refunds.all()], Decimal(0)
+        )
+        update_order_charge_status(o, granted_refund_amount)
+        orders_to_update.append(o)
+        last_number = o.number
```
fowczarek, line 118: Maybe?
```suggestion
    last_number = orders_to_update[-1].number
```

fowczarek (approved):

szczecha (approved):

PR #14378, opened 2023-10-20
Fix generating slug for product create mutation

SzymJ:
I want to merge this change because it fixes generating `slug` for cases when `slug` is not provided and `name` contains characters that return an empty string after `slugify`. 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/core/utils/__init__.py
+++ b/saleor/core/utils/__init__.py
@@ -119,6 +119,11 @@ def generate_unique_slug(
     """
     slug = slugify(unidecode(slugable_value))
 
+    # in case when slugable_value contains only not allowed in slug characters, slugify
+    # function will return empty string, so we need to provide some default value
+    if slug == "":
+        slug = "-"
+
     ModelClass = instance.__class__
 
     search_field = f"{slug_field_name}__iregex"
--- a/saleor/core/utils/__init__.py
+++ b/saleor/core/utils/__init__.py
@@ -150,7 +150,11 @@ def prepare_unique_slug(slug: str, slug_values: Iterable):
 
     while unique_slug in slug_values:
         extension += 1
-        unique_slug = f"{slug}-{extension}"
+
+        if slug == "-":
+            unique_slug = f"{slug}{extension}"
+        else:
+            unique_slug = f"{slug}-{extension}"
 
     return unique_slug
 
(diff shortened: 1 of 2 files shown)


saleor/graphql/product/tests/mutations/test_product_create.py
```diff
@@ -190,6 +190,32 @@ def test_create_product(
     update_product_discounted_price_task_mock.assert_called_once_with(product.id)
 
 
+def test_create_product_without_slug_and_not_allowed_characters_for_slug_in_name(
```
IKarbowiak, line 193: Maybe let's create second test when the product with `-` slug already exist

IKarbowiak (approved):

saleor/core/utils/__init__.py
```diff
@@ -145,7 +150,11 @@ def prepare_unique_slug(slug: str, slug_values: Iterable):
 
     while unique_slug in slug_values:
         extension += 1
-        unique_slug = f"{slug}-{extension}"
+
+        if slug == "-":
```
korycins, line 154: I might be wrong, but I am not convinced that this will work. 
See for example pattern from line 130. 
If I will have a products with slug: "-1", "-2", "-3":
```python
p = Product.objects.create(slug="-2", product_type_id=10)
p = Product.objects.create(slug="-3", product_type_id=10)
slug = "-"
slug_field_name = "slug"
search_field = f"{slug_field_name}__iregex"
pattern = rf"{slug}-\d+$|{slug}$"
lookup = {search_field: pattern}
slug_values = (
        Product.objects.filter(**lookup)
        .exclude(pk=instance.pk)
        .values_list(slug_field_name, flat=True)
    )
```
```python
slug_values
Out[27]: <ProductsQueryset ['-']>
```

Where the same lookup for `slug` == `3` returns different list:
```python
p = Product.objects.create(slug="3-1", product_type_id=10)
p = Product.objects.create(slug="3-2", product_type_id=10)
p = Product.objects.create(slug="3-3", product_type_id=10)

slug_values
Out[37]: <ProductsQueryset ['3-1', '3-2', 'polo-shirt-3', 't-shirt-3']>
```

SzymJ: thanks, I will check it

SzymJ: @korycins You were right. I removed line `unique_slug = f"{slug}{extension}"` to keep the pattern we are using so slugs for this case will look like `-`, `--2`, `--3` etc.

michal-macioszczyk (approved):

szczecha (approved):

Issue #14382, opened 2023-10-20
Enhance `DraftOrderComplete` with voucher validation


IKarbowiak:
We should validate here basically everything that is validated during `checkoutComplete`.

We should create `VoucherCustomer` instance here.

- validate min spent
- validate quantity
- validate once per customer
- validate is staff
- validate usage limit

Maybe we can use [this](https://github.com/saleor/saleor/blob/main/saleor/discount/utils.py#L239) method

Issue #14376, opened 2023-10-20
Update `draftOrderCreate` and `draftOrderUpdate` voucher validation


IKarbowiak:
Remove `CustomerVoucher` creation as the user might change (we should create `VoucherCustomer` during `draftOrderComplete` - will be done in https://github.com/saleor/saleor/issues/14382)

PR #14397, opened 2023-10-23
Remove CustomerVoucher creation from draftOrderCreate

michal-macioszczyk:
I want to merge this change because I want to remove CustomerVoucher creation as the user might change (we should create VoucherCustomer during draftOrderComplete - will be done in https://github.com/saleor/saleor/issues/14382)

https://github.com/saleor/saleor/issues/14376

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/discount/utils.py
+++ b/saleor/discount/utils.py
@@ -59,8 +59,6 @@ def increase_voucher_usage(
 ) -> None:
     if voucher.usage_limit:
         increase_voucher_code_usage_value(code)
-    if voucher.apply_once_per_customer:
-        add_voucher_usage_by_customer(code, customer_email)
     if voucher.single_use:
         deactivate_voucher_code(code)
 
--- a/saleor/discount/utils.py
+++ b/saleor/discount/utils.py
@@ -55,10 +55,15 @@ CATALOGUE_FIELDS = ["categories", "collections", "products", "variants"]
 
 
 def increase_voucher_usage(
-    voucher: "Voucher", code: "VoucherCode", customer_email: str
+    voucher: "Voucher",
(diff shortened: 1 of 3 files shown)


saleor/discount/utils.py
```diff
@@ -59,8 +59,6 @@ def increase_voucher_usage(
 ) -> None:
     if voucher.usage_limit:
         increase_voucher_code_usage_value(code)
-    if voucher.apply_once_per_customer:
-        add_voucher_usage_by_customer(code, customer_email)
```
IKarbowiak, line 63: We cannot remove it from this function as the function is used in different places as well.

IKarbowiak (changes requested):

saleor/discount/utils.py
michal-macioszczyk: changed to new solution.

saleor/discount/utils.py
```diff
@@ -55,11 +55,14 @@
 
 
 def increase_voucher_usage(
-    voucher: "Voucher", code: "VoucherCode", customer_email: str
+    voucher: "Voucher",
+    code: "VoucherCode",
+    customer_email: str,
+    allow_customer_voucher_creation: bool = True,
```
IKarbowiak, line 61: I would suggest changing the name to `increase_voucher_customer_usage` or smth like this, as we're creating the `VoucherCustomer` under the hood but it's not visible in this function

michal-macioszczyk: done

SzymJ (approved):

Issue #13700, opened 2023-08-16
Bug: `productBulkCreate` optimization issue


maarcingebala:
### What are you trying to achieve?

Running the `productBulkCreate` mutation to create a single product with 5-10 variants, where each has around 20 attributes, takes over 20 seconds.

### Steps to reproduce the problem

```
mutation($products:[ProductBulkCreateInput!]!){
  productBulkCreate(
    errorPolicy:IGNORE_FAILED,
    products:$products
  ) {
    errors{
      code
      message
      values
    }
    results{
      product{
        id
      }
      errors{
        code
        message
        values
      }
    }
  }
}
```

### What did you expect to happen?

20 seconds is an unacceptable time for this mutation.

### Logs

Context: https://saleorcommerce.slack.com/archives/C02LMJV6GGM/p1691754728176669
Trace: https://app.datadoghq.com/apm/trace/4092933743730849626?colorBy=service&env=production&filteredType=web&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=8068242235606102362&spanViewType=metadata&timeHint=1692268081392.753

### Environment

Saleor version: 3.14.9

tomaszszymanski129:
Most of it was handled in https://github.com/saleor/saleor/pull/13792 and https://github.com/saleor/saleor/pull/13850, there is still a case when it is rather slow - richtext attributes and product media.

PR #14485, opened 2023-10-26
Add max_sort_order field as cache to reduce database operations

8r2y5:
I want to merge this change because it fixes https://github.com/saleor/saleor/issues/13700 by reducing number of database operations to get current `sort_order`.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


fowczarek (changes requested):
We should use `F` expression when updating `max_sort_order`.
https://docs.djangoproject.com/en/4.2/ref/models/expressions/#f-expressions

IKarbowiak (approved):

saleor/attribute/tasks.py
```diff
@@ -29,3 +30,12 @@ def update_associated_products_search_vector(attribute_value_pk: int):
         Q(Exists(instance.productassignments.filter(product_id=OuterRef("id"))))
         | Q(Exists(variants.filter(product_id=OuterRef("id"))))
     ).update(search_index_dirty=True)
+
+
+@app.task()
+def setup_max_sort_order():
+    for attribute in Attribute.objects.filter(max_sort_order=None).iterator():
```
maarcingebala, line 37: I think it needs to use batching as well. If you had say a million attributes, it would work well with the memory when we have an iterator, but the task could still take a long time. Better would be to have smaller tasks that queue themselves when needed and operate on batches.

Another question is if we could do that in a single `update` without looping?

saleor/attribute/tasks.py
maarcingebala, line 1: For zero-downtime migrations that use Celery tasks, we define the tasks inside migrations directories. Could we do the same here? The task is used only in migrations.

jakubkuc (approved):

tomaszszymanski129 (approved):

fowczarek (approved):

Issue #14498, opened 2023-10-27
Bug: Impossible to determine the refund shipping cost for payment plugins when using webhook subscription


korycins:
### What are you trying to achieve?

The static webhook payload contains the details about `fulfillment.total_refund_amount` and `fulfillment.shipping_refund_amount`.  These fields are used when calling the refund action with return for fulfillment. It is related only to the "old" payment API flow. 
In case of migrating from old-webhook-payload to webhook subscription, these fields are missing and we are not able to determine if the specific refund for payment API included the shipping cost. 

### Steps to reproduce the problem

1. Create an order subscription webhook and include fulfillment, try to find the fields for a refund on the Fulfillment.

### What did you expect to happen?

I expect that Fulfillment will recieve the same fields as the static payload has. This field should contain the value when staff user triggers the refund for payment API flow.

### Environment

Saleor version: 3.14+
OS and version: …

PR #14499, opened 2023-10-27
Add missing refund fields to fulfillment

korycins:
I want to merge this change because it fixes the problem described here: https://github.com/saleor/saleor/issues/14498

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/types.py
+++ b/saleor/graphql/order/types.py
@@ -573,6 +573,17 @@ class Fulfillment(ModelObjectType[models.Fulfillment]):
         required=False,
         description="Warehouse from fulfillment was fulfilled.",
     )
+    shipping_refunded_amount = graphene.Field(
+        Money,
+        description="Amount of refunded shipping price." + ADDED_IN_314,
+        required=False,
+    )
+    total_refunded_amount = graphene.Field(
+        Money,
+        description="Total refunded amount assigned to this fulfillment."
+        + ADDED_IN_314,
+        required=False,
+    )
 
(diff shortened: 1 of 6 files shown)


saleor/graphql/order/types.py
```diff
@@ -615,6 +626,34 @@ def _resolve_stock(fulfillment_lines: List[models.FulfillmentLine]):
...
+            .then(_resolve_shipping_refund)
+        )
+
+    @staticmethod
+    def resolve_total_refunded_amount(root: models.Fulfillment, info):
+        if root.total_refund_amount is None:
+            return None
+
+        def _resolve_total_refund_amount(order):
+            return prices.Money(root.total_refund_amount, currency=order.currency)
```
IKarbowiak, line 649: Shouldn't we add `currency` on `Fulfillment` model instead fetching it from order each time? (I mean on `main` branch, I don't suggest to introduce it from 3.14)

IKarbowiak (approved):

saleor/graphql/order/types.py
korycins: Honestly, these changes are only to support the old payment flow & refund on webhook subscriptions. We will not need it when we will drop Payment API and fully move to transactionAPI. So I would say, that it doesn't make sense ta add this to Fulfillment model.

kadewu (approved):

szczecha (approved):

Issue #14125, opened 2023-09-28
Bump django-stubs


maarcingebala:
`django-stubs` is currently at 1.16.0, and to bump it to versions above 4.2.4, code changes are required. As per 4.2.4 release notes:
> We no longer assume the objects attribute to be present on generic Model classes. As per the [django documentation](https://docs.djangoproject.com/en/4.2/topics/db/managers/#django.db.models.Model._default_manager), code working with generic models should use the _default_manager attribute instead.

Effectively, it means we need to update occurrences of `Queryset.objects` to `Queryset._default_manager`.

PR #14510, opened 2023-10-27
Update django-stubs

patrys:
This updates django-stubs to the latest version and hopefully stabilizes mypy behavior.

Fixes: https://github.com/saleor/saleor/issues/14125

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


aniav (approved):

saleor/graphql/core/mutations.py
```diff
@@ -1173,6 +1173,8 @@ def perform_mutation(  # type: ignore[override]
         cls, _root, info: ResolveInfo, /, file
     ):
         file_data: UploadedFile = cast(UploadedFile, info.context.FILES[file])
+        if not file_data.file:
+            raise ValueError("Received an empty file.")
```
korycins, line 1177: Are you sure that ValueError is ok here? Isn't it like this will be returned as an unhandled error? Shouldn't it be ValidationError?

saleor/graphql/account/utils.py
```diff
@@ -108,8 +108,8 @@ def can_manage_app(requestor: Union["User", "App", None], app: "App") -> bool:
 def get_group_permission_codes(group: Group) -> "QuerySet":
     """Return group permissions in the format '<app label>.<permission codename>'."""
     return group.permissions.annotate(
-        formated_codename=Concat("content_type__app_label", Value("."), "codename")
-    ).values_list("formated_codename", flat=True)
+        formatted_codename=Concat("content_type__app_label", Value("."), "codename")
+    ).values_list("formatted_codename", flat=True)  # type: ignore[misc]
```
maarcingebala, line 112: There are still some usages of `formated_codename` in the codebase; maybe we should fix all of them on this occasion. I'm wondering if changing it only here doesn't break it somewhere else where `formated_codename` is expected.

saleor/graphql/core/mutations.py
maarcingebala: Yep, this should raise ValidationError, otherwise, this will be reported as "Internal Server Error" in the GraphQL response and thrown in Sentry as an unhandled exception.

Issue #13330, opened 2023-07-06
Add `checkoutAndOrderPredicate` to Promotion Rules for checkout/order total and subtotal


IKarbowiak:
Part of https://github.com/saleor/saleor/issues/12889

Implement all required logic from https://github.com/saleor/saleor/issues/12889 to allow defining only `totalPrice ` and `subtotalPrice` in `discountedObjectPredicate ` in `CheckoutAndOrderPredicateInput`

## 1. General Assumptions
- Allow define `rewardType` on `PromotionRule`
- Allow defining `rewardType` only for rules with `checkoutAndOrderPredicate`. In case od catalogue discounts the reward type is implied by `rewardValuType` (fixed/percentage)
- Allow defining `checkoutAndOrderPredicate` in promotion and promotion rules mutations
- Allow defining predicates based on `totalPrice` and `subtotalPrice` of checkout and order
- Apply the discounts in checkout and order calculations
- Ensure the tax are correctly calculated
- Add a validation in `PromotionRule` mutations to not allow providing both `checkoutAndOrderPredicate` and `cataloguePredicates`. The promotion rule discount should be applied on catalogue level or on order/checkout level. If we would allow to mix those it will be not clear where we should apply the discount - on catalogue or on order level. So to make it clear the rule should provide catalogue discount (that will reduce the variant price) or checkout/order discount that will be applied during order creation.
- One promotion can only have checkout rules OR catalogue rules - add a validation in `Promotion` and `PromotionRule` mutations to not allow mixing those types of rules
- During calculation, we should choose the `promotion` that gives the best discount for the user.

**For calculations**
- Implement a solution to apply the promotion that gives the greatest saving from all promotions that applied on the order.
- When the manual discount is given ignore discounts from promotions - include such info in API field description and docs
- For total and subtotal price conditions we should take into consideration price that is default for user (which depend on the configuration flag `prices_entered_with_tax`)
- Vouchers and promotion discounts are not mixes, when the voucher is given, do not calculate the promotion discounts

### Examples:
20% discount on subtotal when you spend at least $100
```json
{
    "promotionRule": {
        "checkoutAndOrderPredicate": {
            "discountedObjectPredicate": {
                "subtotalPrice": {
                    "range": {
                        "gte": 100
                     }
                  },
              }
        },
        "rewardType": "SUBTOTAL_DISCOUNT"
        "rewardValueType": "PERCENTAGE",
        "rewardValue": 20
    }
}
```

### Mutations
[...]

PR #14696, opened 2023-11-10
Add checkout and order promotions

IKarbowiak:
Allow creating promotions that applies on checkout/order level

Covers https://github.com/saleor/saleor/issues/13330

- The solution allows the creation of promotion rules that define `checkoutAndOrder` discount applied on checkout/order level.
- Current implementation applies the discount only on `Checkout`, it's not applicable on `Order` yet.
- The limit of possible `PromotionRule` with `checkoutAndOrder` predicate was introduced. It's set to 100 for the default.

### Measurement:
In case checkout applies for all 100 rules, the checkout calculation time is extended a bit.

**Case 1**
- checkout with one line
- 100 promotion rules with `checkoutAndOrder` predicate
- all rules applies on the checkout

| Mutation        | Time in s    |
| ------------- | ------------- |
| CheckoutCreate  | 0.638  |
| Checkout delivery method update  | 0.896  |
| Checkout complete  | 0.775  |

**Case 2**
- checkout with 1 line
- no promotion rules

| Mutation        | Time in s    |
| ------------- | ------------- |
| CheckoutCreate  | 0.426  |
| Checkout delivery method update  | 0.569  |
| Checkout complete  | 0.713  |

**Case 3**
- checkout with 100 lines
- 100 promotion rules with `checkoutAndOrder` predicate
- all rules applies on the checkout

| Mutation        | Time in s    |
| ------------- | ------------- |
| Checkout delivery method update  | 0.947  |
| Checkout complete  | 1.64  |

**Case 4**
- checkout with 100 lines
- no promotion rules

| Mutation        | Time in s    |
| ------------- | ------------- |
| Checkout delivery method update  | 0.604  |
| Checkout complete  | 2.2  |

# Impact

- [x] New migrations

# Docs

- [ ] Link to documentation: https://github.com/saleor/saleor-docs/pull/1067

# Pull Request Checklist


yellowee (approved):

saleor/graphql/discount/types/promotions.py
```diff
@@ -82,19 +88,34 @@ class PromotionRule(ModelObjectType[models.PromotionRule]):
             AuthorizationFilters.AUTHENTICATED_STAFF_USER,
         ],
     )
+    reward_value = PositiveDecimal(
```
korycins, line 91: Shoiuldn't we add ADDED_IN + PREVIEW_FEATURE?

korycins: Same for `type` and `predicate_type`

saleor/graphql/discount/utils.py
```diff
@@ -28,27 +29,17 @@
 PREDICATE_OPERATOR_DATA_T = list[dict[str, Union[list, dict, str, bool]]]
 
 
+class PredicateObjectType(Enum):
+    CATALOGUE = "catalogue"
+    CHECKOUT = "checkout"
+    ORDER = "order"
```
korycins, line 35: Do we use all of three? I might be wrong, but I think that we reduced it to two, right?

korycins: Also, as we have `predicate` for catalogue, and for order, this enum should operate on `CATALOGUE` and `ORDER`. Is it correct?

korycins:
From the benchmark above:
`add checkout lines with reservations` queries increased from `157` to `181`, where duplicates are increased by `9` queries. It may be worth checking why the number is so high. The rest of the tests increase the queries by 2, and it looks correct

saleor/graphql/discount/utils.py
IKarbowiak: Those values are used for filtering. For `checkout` we are using `CheckoutDiscountedObjectWhere`, for `order` we will be using `OrderDiscountedObjectWhere`

saleor/discount/utils.py
```diff
@@ -556,6 +729,41 @@ def get_variants_to_promotions_map(
...
+        )
+        .exclude(order_predicate={})
+        .prefetch_related("channels")
+    )
+    rule_to_channel_ids_map = _get_rule_to_channel_ids_map(rules)
+
+    checkout_channel_id = checkout.channel_id
+    currency = checkout.channel.currency_code
+    checkout_qs = Checkout.objects.filter(pk=checkout.pk)
+    for rule in list(rules.iterator()):
```
korycins, line 751: This means, that we will load ALL existing order rules to memory.

saleor/checkout/utils.py
```diff
@@ -743,6 +746,12 @@ def add_voucher_to_checkout(
     checkout_info.voucher = voucher
     checkout_info.voucher_code = voucher_code
 
+    # delete discounts from order promotions as cannot be mixed with vouchers
```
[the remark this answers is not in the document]
IKarbowiak: Done

saleor/graphql/discount/types/promotions.py
IKarbowiak: Done

saleor/discount/utils.py
korycins: I've made a test, on 5 mln of checkouts. With `https://github.com/pythonprofilers/memory_profiler`
```
In [5]: from memory_profiler import memory_usage
   ...: from time import sleep
   ...:
   ...: def f():
   ...:     # a function that with growing
   ...:     # memory consumption
   ...:     print(Checkout.objects.count())
   ...:     for a in list(Checkout.objects.iterator()):
   ...:         b = a.pk
   ...:     return None
   ```
   vs 
   ```
   In [5]: from memory_profiler import memory_usage
   ...: from time import sleep
   ...:
   ...: def f():
   ...:     # a function that with growing
   ...:     # memory consumption
   ...:     print(Checkout.objects.count())
   ...:     for a in Checkout.objects.iterator():
   ...:         b = a.pk
   ...:     return None
   ```
   
   `max(memory_usage(f))` returned `3737.390625` MiB for `list(iterator())`, and `235.671875` for `iterator()`
   
   Using `iterator` allows db to stream the data (ref:https://docs.djangoproject.com/en/5.0/ref/models/querysets/#with-server-side-cursors), but having `list(iterator())` will cause that we will load all data into worker memory.

File: saleor/discount/interface.py [hunk 5]
```diff
@@ -88,3 +76,25 @@ def fetch_variant_rules_info(
...
+    promotion: "Promotion", rule: "PromotionRule", translation_language_code: str
+):
+    promotion_translations = [
+        translation
+        for translation in promotion.translations.all()
+        if translation.language_code == translation_language_code
+    ]
+    promotion_translation = (
+        promotion_translations[0] if promotion_translations else None
+    )
```
kadewu, line 91: Maybe something like:
```suggestion
    promotion_translation = promotion.translations.filter(language_code=translation_language_code).first()
```
And same for rule_translation
WDYT?

saleor/discount/utils.py
```diff
@@ -535,6 +721,41 @@ def get_variants_to_promotion_rules_map(
...
+    PromotionRuleChannels = PromotionRule.channels.through.objects.filter(
+        channel_id=checkout_channel_id
+    )
+    rules = (
+        PromotionRule.objects.filter(
+            Exists(promotions.filter(id=OuterRef("promotion_id"))),
+            Exists(PromotionRuleChannels.filter(promotionrule_id=OuterRef("id"))),
+        )
+        .exclude(order_predicate={})
+        .prefetch_related("channels")
```
kadewu, line 741: Is this needed? (I just glanced through the function calls, so I might have missed though)

If it is used, then please note, that iterator is compatible with `prefetch_related` if `chunk_size` is set.

kadewu (commented):
Just some possible performance improvments from my site, overall LGTM!

saleor/discount/interface.py
IKarbowiak: `promotion.transaltions` are prefetched, `promotion.translations.filter` will cause extra db query

saleor/discount/utils.py
IKarbowiak: I removed `prefetch_related` as it's not needed anymore :)

kadewu (approved):

tomaszszymanski129 (approved):

yellowee (approved):

PR #14711, opened 2023-11-10
Move sale migration celery task.

zedzior:
I want to merge this change, because the celery task for recalculating discount prices during sale to promotion migration is going to stay asynchronous (usually we convert migration celery tasks to migration). In `migration/tasks/saleor{saleor version}.py` directory we only keep tasks, which should be converted to migrations in next version, therefore this PR moves it to `discount/tasks` directory. 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/discount/tasks.py
```diff
@@ -214,3 +220,36 @@ def disconnect_voucher_codes_from_draft_orders_task(order_ids):
         ).filter(type=DiscountType.VOUCHER).delete()
         if remaining_ids := list(set(order_ids) - set(ids)):
             disconnect_voucher_codes_from_draft_orders_task.delay(remaining_ids)
+
+
+@app.task
+def update_discounted_prices_task():
+    """Recalculate discounted prices during sale to promotion migration."""
+    # WARNING: this function is run during `0047_migrate_sales_to_promotions` migration,
+    # so please be careful while updating.
```
aniav, line 229: Would be good to add a comment along the lines:

```
# This task can be deleted after we introduce a different process for calculating discounted prices for promotions.
```

aniav (approved):

saleor/discount/tasks.py
```diff
@@ -214,3 +220,38 @@ def disconnect_voucher_codes_from_draft_orders_task(order_ids):
         ).filter(type=DiscountType.VOUCHER).delete()
         if remaining_ids := list(set(order_ids) - set(ids)):
             disconnect_voucher_codes_from_draft_orders_task.delay(remaining_ids)
+
+
+@app.task
```
patrys, line 225: If you pass `name` here, you can keep it backwards-compatible.

zedzior: I kept the name unchanged.

8r2y5 (approved):

saleor/discount/tasks.py
maarcingebala: But Celery uses a full module path for automatic task names, doesn't it? So names are:
- old: `saleor.discount.migrations.tasks.saleor3_17.update_discounted_prices_task`
- new: `saleor.discount.tasks.update_discounted_prices_task`
So without explicitly setting the name, after this upgrade it will fail to resolve the task name (if the old tasks are still in the queue).

korycins (approved):

PR #14787, opened 2023-11-17
FIX for migration on VoucherCode created_at field

michal-macioszczyk:
I want to merge this change because there is mismatch between migrations from 3.17 to main branch and we want to fix that. 

# Impact

- [x] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/discount/migrations/0065_vouchercode_created_at_check.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("discount", "0064_alter_promotionrule_catalogue_predicate"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql=(
```
kadewu, line 13: I would suggest to simplify that query to just:
`ALTER TABLE discount_vouchercode ADD COLUMN IF NOT EXISTS created_at TIMESTAMP;`

I now see one problem here, I'm not convinced if we should set default on the database, it's a function so the migration will be costly here. Do we know the impact of that table? (cause the data migration with batches may be needed to propagate the default value)

kadewu (approved):

maarcingebala (approved):

Air-t (approved):

Air-t (approved):

korycins (approved):

PR #14820, opened 2023-11-21
[Test] Use external shipping methods in the checkout

szczecha:
I want to merge this change because it covers test case [CORE_1652](https://saleor.testmo.net/repositories/5?group_id=591&case_id=4758) Should be able to use external shipping method in the checkout

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/tests/e2e/checkout/shipping/test_use_external_shiping_methods_in_checkout.py
```diff
@@ -0,0 +1,136 @@
...
+        email="testEmail@example.com",
+        set_default_billing_address=True,
+        set_default_shipping_address=True,
+    )
+    checkout_id = checkout_data["id"]
+    assert checkout_data["isShippingRequired"] is True
+    calculated_subtotal = product_variant_price * 4
+    assert checkout_data["totalPrice"]["gross"]["amount"] == calculated_subtotal
+    shipping_methods = checkout_data["shippingMethods"]
+    assert len(shipping_methods) == 3
```
yellowee, line 98: You have 3 external shipping methods, but you assert only 2 of them - is it intended?

yellowee (changes requested):

saleor/tests/e2e/checkout/shipping/test_use_external_shiping_methods_in_checkout.py
szczecha: I have two external shipping methods that I want to check as the should have come from webhook and one created in Saleor that's why I assert only externals :) I will add an assertion for that one too

yellowee (approved):

SzymJ (approved):

saleor/tests/e2e/utils.py
```diff
@@ -18,3 +20,22 @@ def assign_permissions(api_client, permissions):
         app = api_client.app
         if app:
             app.permissions.add(*permissions)
+
+
+def request_matcher(r1, r2):
+    body1 = json.loads(r1.body.decode("utf-8"))
+    body2 = json.loads(r2.body.decode("utf-8"))
+
+    # Check if all key-value pairs in body1 are present in body2
```
korycins, line 29:
```
if set(body1.keys()) != set(body2.keys()):
    return False
```
Then you are sure, that body1 and body2 have the same keys. So only need to check b1.value vs b2.value

SzymJ (approved):

IKarbowiak (approved):

PR #14843, opened 2023-11-22
Add an LRU cache for GraphQL document parser and validator

patrys:
Parsing and validating queries costs precious milliseconds. We could avoid doing it for popular queries.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/api.py
+++ b/saleor/graphql/api.py
@@ -1,8 +1,21 @@
+from functools import partial
+
 import graphql
 from django.urls import reverse
 from django.utils.functional import SimpleLazyObject
-from graphql import GraphQLScalarType
+from graphql import (
+    GraphQLCachedBackend,
+    GraphQLCoreBackend,
+    GraphQLScalarType,
+    GraphQLSchema,
+    execute,
+    parse,
+    validate,
+)
+from graphql.backend.base import GraphQLDocument
+from graphql.execution import ExecutionResult
 
+from ..core.utils.cache import CacheDict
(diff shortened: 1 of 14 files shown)


saleor/graphql/api.py
```diff
@@ -168,3 +181,35 @@ def serialize_webhook_event(value):
...
+    ) -> GraphQLDocument:
+        # validate eagerly so we can cache the result
+        document_ast = parse(document_string)
+        validation_errors = validate(schema, document_ast)
+        if validation_errors:
+            return GraphQLDocument(
+                schema=schema,
+                document_string=document_string,
+                document_ast=document_ast,
+                execute=partial(_fail, validation_errors),
```
maarcingebala, line 204: So the purpose here is to have a different `execute` function that returns early (when we know there are validation issues), without going into all logic that the regular `execute` does?

patrys: Yes, the previous implementation, `validate_and_execute` from `graphql-core`, did precisely that: it pretended to be `execute`, but internally first ran `validate` and if that failed, immediately returned a result without calling `execute`.

patrys: Ideally, we would not even bother with caching incorrect queries (as they are unlikely to be repeated), but that would require writing an even more elaborate backend implementation. Anything uncommon will be pushed out by more popular queries so it should be fine as it is.

maarcingebala (approved):

Issue #14472, opened 2023-10-26
Take into account the allow_replica in WebhookPlugin


korycins:
## Problem
The main problem is described here: 
The linked issue set the allow_replica to always False. It means that we always use main DB to generate the payload. It was only a quick fix for previous Saleor's verions.

---

## 1. General Assumptions
- The WebhookPlugin should take into account the `allow_replica`, provided by PluginsManager.
- Each payload generated by WebhookPlugin should take into account `allow_replica` value. 
- The `allow_replica` input should be mandatory argument to all "trigger" methods/functions used in WebhookPlugin handlers.  Example:
```python
    def order_created(self, order: "Order", previous_value: Any) -> Any:
        if not self.active:
            return previous_value
        event_type = WebhookEventAsyncType.ORDER_CREATED
        if webhooks := get_webhooks_for_event(event_type):
            order_data_generator = partial(
                generate_order_payload, order, self.requestor
            )
            trigger_webhooks_async(
                None,
                event_type,
                webhooks,
                order,
                self.requestor,
                legacy_data_generator=order_data_generator,
                allow_replica=self.allow_replica # <---
            )
```
Probably we could introduce the wrapper on WebhookPlugin side that will set this value by default
- All get_plugins_manager calls (Explictly any methods/functions that creates the instance of PluginsManager) should explicitly pass the value `allow_replica`. (we should consider setting up this value as a mandatory - if we do, we should probably notify the cloud team. It sounds as the safest solution for future new webhook events).
- The calls of async webhook triggered by mutation should have set allow_replica to False
- The calls of sync webhook triggered by mutation should have set allow_replica to False
- The calls of sync webhook triggered by query should have set allow_replica to True ( keep in mind, that it is only for the query request, not the response of the mutation query)

### Acceptance Criteria

- [ ] The calls of async webhook triggered by mutation have set allow_replica to False
- [ ] The calls of sync webhook triggered by mutation have set allow_replica to False
- [ ] The calls of sync webhook triggered by query have set allow_replica to True 

### Mutations

No changes

### Types

No changes

## 3. Database changes

No changes

## 5. To Do list

```[tasklist]
### Tasks
```

PR #14844, opened 2023-11-22
Use replicas in webhook payloads

tomaszszymanski129:
I want to merge this change because it adds usage of replicas to webhook payloads.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/webhook/payloads.py
+++ b/saleor/webhook/payloads.py
@@ -16,6 +16,7 @@ from typing import (
 )
 
 import graphene
+from django.conf import settings
 from django.db.models import F, QuerySet, Sum
 from django.utils import timezone
 from graphene.utils.str_converters import to_camel_case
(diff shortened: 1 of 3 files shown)


saleor/webhook/payloads.py
```diff
@@ -550,14 +553,18 @@ def generate_checkout_payload(
     quantize_price_fields(checkout, checkout_price_fields, checkout.currency)
     user_fields = ("email", "first_name", "last_name")
 
-    lines_dict_data = serialize_checkout_lines(checkout)
+    lines_dict_data = serialize_checkout_lines(checkout, allow_replica=True)
```
maarcingebala, line 556: Are we sure about this one to use replica? This webhook is called on checkout mutations where checkout lines can be modified, so are we 100% sure that the replica is up-to-date at this point?

tomaszszymanski129: That's the fattest, most expensive call (because of the amount of prefetches it does deeper in `fetch_checkout_lines`) but yep, unfortunately I see it may cause issues as deeper described [here](https://github.com/saleor/saleor/issues/14472). One of ACs on that is that `The calls of async webhook triggered by mutation have set allow_replica to False` and this is direct example. I'm closing this one.

Issue #14834, opened 2023-11-22
Fix missing subscription payload handling


jakubkuc:
At the moment configuring subscription queries in webhook incorrectly for sync events is causing Saleor to crash and raise `PaymentError`due to empty payload generated from subscription.

Way to reproduce: 
1. Create a Webhook with registered two events but only one defined in subscription query:
e.g. 

```
mutation {
  webhookCreate(
    input: {
      app: "<appid>"
      name: "test_subscription"
      targetUrl: "<URL>"
      syncEvents: [SHIPPING_LIST_METHODS_FOR_CHECKOUT, CHECKOUT_FILTER_SHIPPING_METHODS]
      query: "subscription{event{...on CheckoutFilterShippingMethods{__typename}}}"
    }
  ) {
    errors {
      field
      message
    }
    webhook {
      id
    }
  }
}
```
2. Fetch checkout querying the event missing from subscription query

Result: Saleor crashes and query returns internal error 

Expected result: 
We should handle empty payloads (caused by either incorrect subscription queries or crashing apps) without crashing. 
- If the subscription results in an empty payload, we want to ignore the webhook and proceed as if it didn't exist.
- If the subscription is correct but the app is broken (down, returns an incorrect result etc.), proceed as if it didn't exist (see above)

maarcingebala:
A [thread](https://saleorcommerce.slack.com/archives/C0373831YBC/p1700652262051419) with more context.

PR #14858, opened 2023-11-24
Fix handling subscription with empty data

jakubkuc:
I want to merge this change because it fixes [#14834](https://github.com/saleor/saleor/issues/14834)
This should not break how any sync webhook work at the moment. The only change will be that if something with subscription was incorrect and payload generation resulted in empty data - it will be logged and returned as None instead of raising an Error. It will work the same as the event was not subscribable. 
Additionally some redundant lines from tests were removed. 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


maarcingebala (approved):
Looks OK, but Black is failing, please check linters.

Air-t (approved):

korycins (approved):
Looks good 👍 
@jakubkuc  can you confirm, that you double-checked the logic that using the `create_delivery_for_subscription_sync_event(` implicitly and explicitly, and we don't have any unneeded try, except for `PaymentError`?

saleor/plugins/webhook/tasks.py
```diff
@@ -164,12 +164,12 @@ def create_delivery_for_subscription_sync_event(
         app=webhook.app,
     )
     if not data:
-        # PaymentError is a temporary exception type. New type will be implemented
-        # in separate PR to ensure proper handling for all sync events.
-        # It was implemented when sync webhooks were handling payment events only.
-        raise PaymentError(
+        logger.info(
             f"No payload was generated with subscription for event: {event_type}"
```
8r2y5, line 168: Please use % string formatting when dealing with loggers

jakubkuc:
@korycins Yes, I double-checked, this PaymentError was not handled anywhere apparently.

digital_contents/sample_image_as_digital_content.jpg
szczecha, line 1: These files should be omitted from PR

export_files/temp_file.csv
szczecha, line 1: These files should be omitted from PR

szczecha (approved):
@jakubkuc

saleor/plugins/webhook/tasks.py
jakubkuc: done 👍🏻

export_files/temp_file.csv
jakubkuc: of course, it was a mistake. fixed

IKarbowiak (approved):

PR #14876, opened 2023-11-29
Update filter_payment_status to fix filtering orders by Unpaid

simon-online:
Some Unpaid orders don't have payment records such as those manually created from a draft order or those created via bulk import. These types of Unpaid orders aren't included when filtering orders by Unpaid payment status which I consider a bug.

This update fixes this issue and ensures all Unpaid orders are included when filtering orders by Unpaid payment status.

# Impact

No impacts.

# Docs

No changes to docs required.

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/filters.py
+++ b/saleor/graphql/order/filters.py
@@ -8,6 +8,7 @@ from graphql.error import GraphQLError
 
 from ...giftcard import GiftCardEvents
 from ...giftcard.models import GiftCardEvent
+from ...order import OrderChargeStatus
 from ...order.models import Order, OrderLine
 from ...order.search import search_orders
 from ...payment import ChargeStatus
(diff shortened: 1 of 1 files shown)


korycins (commented):
The problem here is much bigger than the proposed fix.

The order's `paymentStatus` is calculated based on `Payment`/`TransactionItem` objects assigned to Order. These objects represent the payments assigned to the order.
The `paymentStatus` filter operates only on the `Payment`. It means that it is unuseful when order is paid with `TransactionItem` objects. During the implementation of TransactionAPI we added two new payment statuses: `AuthorizeStatus` and `ChargeStatus` with the assumption, that they will be a replacement for  `paymentStatus`.

The changes from this PR are mixing different enum classes, to filter out the orders. I am not sure that this is the correct approach. 
We need to clarify if we want to depreceate the `paymentStatus`,  or fully re-build it to use OrderChargeStatus, and also take into account the orders covered by `TransactionItem` payments.

I will pass this to the team.

aniav:
Hey @simon-online, thank you for your contribution, I'll go ahead and close this PR as it hasn't seen any actions for some time. Feel free to reopen when you have additional changes related to the feedback already provided.

PR #14883, opened 2023-11-30
Optimize products filtering

tomaszszymanski129:
I want to merge this change because it optimizes products filtering in various of places in product filters.
1. It no longer makes an exist within exist to just join a channel while explicitly assuming there exists one.
2. Handles case where there is no channel so further filtering by ProductChannelListing is redundant and can be skipped entirely.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


tomaszszymanski129:
Will be released differently, closing

Issue #14888, opened 2023-12-01
Bug: Performance issue with `handle_promotion_toggle` task.


zedzior:
### What are you trying to achieve?

`handle_promotion_toggle` celery task got stuck due to huge number of products covered by promotions. The task needs refactor to improve its performance.

### Steps to reproduce the problem

Not able to reproduce it locally.

### What did you expect to happen?

The `handle_promotion_toggle` task run succesfully no matter the number of products.

### Environment

Saleor version: …
OS and version: …

PR #14894, opened 2023-12-01
Enhance `handle_promotion_toggle` task performance.

zedzior:
I want to merge this change because it fixes performance issue with `handle_promotion_toggle` task. The celery queue got stuck with huge number of products covered by promotion. This PR make fetching product ids in chunks during `handle_promotion_toggle` task.

Issue: https://github.com/saleor/saleor/issues/14888

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/core/utils/__init__.py
+++ b/saleor/core/utils/__init__.py
@@ -153,19 +153,3 @@ def prepare_unique_attribute_value_slug(attribute: "Attribute", slug: str):
         "slug", flat=True
     )
     return prepare_unique_slug(slug, value_slugs)
-
-
-def queryset_in_batches(queryset, batch_size):
-    """Slice a queryset into batches."""
-    if batch_size <= 0:
-        return []
-
-    queryset = queryset.order_by("pk")
-    start_pk = 0
-    while True:
-        qs = queryset.filter(pk__gt=start_pk)[:batch_size]
-        pks = list(qs.values_list("pk", flat=True))
-        if not pks:
-            break
-        yield pks
-        start_pk = pks[-1]
--- a/saleor/core/utils/__init__.py
+++ b/saleor/core/utils/__init__.py
@@ -156,10 +156,8 @@ def prepare_unique_attribute_value_slug(attribute: "Attribute", slug: str):
 
 
 def queryset_in_batches(queryset, batch_size):
-    """Slice a queryset into batches.
-
-    Input queryset should be sorted be pk.
-    """
+    """Slice a queryset into batches."""
+    queryset = queryset.order_by("pk")
     start_pk = 0
     while True:
         qs = queryset.filter(pk__gt=start_pk)[:batch_size]
--- a/saleor/core/utils/__init__.py
+++ b/saleor/core/utils/__init__.py
@@ -153,3 +153,18 @@ def prepare_unique_attribute_value_slug(attribute: "Attribute", slug: str):
         "slug", flat=True
     )
     return prepare_unique_slug(slug, value_slugs)
+
+
+def queryset_in_batches(queryset, batch_size):
+    """Slice a queryset into batches.
+
+    Input queryset should be sorted be pk.
+    """
+    start_pk = 0
+    while True:
+        qs = queryset.filter(pk__gt=start_pk)[:batch_size]
+        pks = list(qs.values_list("pk", flat=True))
+        if not pks:
+            break
+        yield pks
+        start_pk = pks[-1]
--- a/saleor/core/utils/__init__.py
+++ b/saleor/core/utils/__init__.py
@@ -157,6 +157,9 @@ def prepare_unique_attribute_value_slug(attribute: "Attribute", slug: str):
 
 def queryset_in_batches(queryset, batch_size):
     """Slice a queryset into batches."""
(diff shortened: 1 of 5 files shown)


saleor/core/utils/__init__.py
```diff
@@ -153,3 +153,18 @@ def prepare_unique_attribute_value_slug(attribute: "Attribute", slug: str):
         "slug", flat=True
     )
     return prepare_unique_slug(slug, value_slugs)
+
+
+def queryset_in_batches(queryset, batch_size):
+    """Slice a queryset into batches.
```
korycins, line 159: Can you add here a note that this should be used only with the objects that have `int` as a `pk`?

saleor/core/utils/__init__.py
```diff
@@ -153,3 +153,18 @@ def prepare_unique_attribute_value_slug(attribute: "Attribute", slug: str):
...
+
+    Input queryset should be sorted be pk.
```
korycins, line 161: [Question] Maybe instead of writing that the query should be sorted, we can make this sort here?

saleor/discount/tasks.py
```diff
@@ -157,10 +158,18 @@ def fetch_promotion_variants_and_product_ids(promotions: "QuerySet[Promotion]"):
...
+    variants = variants.order_by("pk")
+    for variant_ids in queryset_in_batches(variants, 2000):
+        variants_batch = ProductVariant.objects.filter(id__in=variant_ids).all()
+        product_ids.update(
+            Product.objects.filter(
+                Exists(variants_batch.filter(product_id=OuterRef("id")))
+            ).values_list("id", flat=True)
+        )
+
+    return promotion_id_to_variants, list(product_ids)
```
NyanKiyoshi, line 172: We still fetch an infinite amount of products, this could cause:
- Payload too big here thus crashing: https://github.com/saleor/saleor/blob/c14ce5cfd8aad483f48b30e370dc0514309df286/saleor/discount/tasks.py#L81
- High memory usage

Also, are we sure Celery beat isn't going to trigger duplicate `update_products_discounted_prices_for_promotion_task` tasks?

saleor/core/utils/__init__.py
zedzior: Why only integers? You can compare strings, dates as well. ie. `Product.objects.filter(name__gt="abc")` is a valid query, right?

saleor/discount/tasks.py
zedzior: 1. It is not infinite. Max is a number of products in database. Scenario: I want to discount all of my products by 10% is a common case.
2. Yes, payload size can be troublesome.
3. We use `set()` on product_ids and update prices in batches. The `update_products_discounted_prices_for_promotion_task` fires itself, once the batch is processed, so I think we are good here.

saleor/discount/tasks.py
```diff
@@ -8,6 +8,7 @@
 from django.db.models import Exists, F, OuterRef, Q, QuerySet
 
 from ..celeryconf import app
+from ..core.utils import queryset_in_batches
```
NyanKiyoshi, line 11: test

saleor/discount/tasks.py
NyanKiyoshi: > It is not infinite. Max is a number of products in database. Scenario: I want to discount all of my products by 10% is a common case.

What I meant is that the value is infinite in code, we don't know how much data this code will have to accept, it can be 10, 10k, 100 billions

> We use set() on product_ids and update prices in batches. The update_products_discounted_prices_for_promotion_task fires itself, once the batch is processed, so I think we are good here.

`set()` doesn't affect how Celery beat works whatsoever, as long as there is no concurrency checks for `handle_promotion_toggle` then it may duplicate tasks

NyanKiyoshi (changes requested):
Limit the number of products that can be fetched and forwarded to a single task, the code shouldn't be allowed to get an infinite count of products but instead of maximum value that we determined is safe (payload size, memory usage, duration).

zedzior:
> Limit the number of products that can be fetched and forwarded to a single task, the code shouldn't be allowed to get an infinite count of products but instead a constant max count that we determined is safe (payload size, memory usage, duration).

Hmm, this would be a breaking change, as some apps may rely on all ids sent by `SALE_TOGGLE` webhook.

zedzior:
I tested the `handle_sale_toogle` task for:
- 60k products with 5 variants each
- single promotion with single rule
- I set very basic rule, like: “all products with category_id=x should be discunted by 10%”


Fetching product ids takes:
- old approach (single select call) : 10s
- new approach (selects in batches, batch_size=10k): 11s

For that specific conditions, the webhook payload has 9MB (300k global ids sent).

zedzior:
The changes will be tested on saleor-multitenant 3.17 first.

saleor/core/utils/__init__.py
korycins: Nevermind. I messed up this with something different 😅

zedzior:
Problem tackled in another PR: https://github.com/saleor/saleor/pull/14982

PR #14924, opened 2023-12-05
Fix attributes save performance

SzymJ:
I want to merge this change because it fixes the performance issue of `associate_attribute_values_to_instance`.
Benchmark for product with 3 attributes:
```
  test name             left count      right count     duplicate count
  -------------------   -----------     -----------     ---------------
- update product                 88             112                   6
```

For `productUpdate` with 103 attributes in input queries count dropped from ~960 -> ~340

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/attribute/utils.py
+++ b/saleor/attribute/utils.py
@@ -1,4 +1,5 @@
-from typing import TYPE_CHECKING, Iterable, Set, Union
+from collections import defaultdict
+from typing import Union
 
 from ..page.models import Page
 from ..product.models import Product, ProductVariant
@@ -9,7 +10,6 @@ from .models import (
     AssignedProductAttributeValue,
     AssignedVariantAttribute,
     AssignedVariantAttributeValue,
-    Attribute,
     AttributeValue,
 )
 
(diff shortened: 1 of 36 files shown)


saleor/attribute/utils.py
```diff
@@ -19,118 +19,154 @@
...
-    instance: T_INSTANCE,
-    attribute: Attribute,
-    *values: AttributeValue,
-) -> AttributeAssignmentType:
-    """Assign given attribute values to a product or variant.
-
-    Note: be aware this function invokes the ``set`` method on the instance's
-    attribute association. Meaning any values already assigned or concurrently
-    assigned will be overridden by this call.
-    """
```
IKarbowiak, line 36: I would suggest not removing this docsting. It contains important information that the attribute values are overridden

fowczarek (commented):
Could you add also benchmarks and put benchmark results in PR description?

saleor/attribute/utils.py
```diff
@@ -19,118 +22,169 @@
...
 
-    return assignment
+def _associate_attribute_to_instance(
+    instance: T_INSTANCE, attr_val_map: dict[int, list]
+):
+    instance_type = instance.__class__.__name__
+    variables = instance_to_function_variables_mapping.get(instance_type)
+
+    if not variables:
+        raise AssertionError(f"{instance_type} is unsupported")
```
tomaszszymanski129, line 65: maybe TypeError would be more accurate? If it's a big change we can skip it for now.

szczecha (approved):

PR #14945, opened 2023-12-07
Optimize products filtering

tomaszszymanski129:
I want to merge this change because it optimizes product filtering in various places.

1. It no longer makes an exist within exist just to join a channel while explicitly assuming there exists one.
2. Handles a case where there is no channel so further filtering by ProductChannelListing is redundant and can be skipped entirely.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/tests/benchmark/test_product.py
+++ b/saleor/graphql/product/tests/benchmark/test_product.py
@@ -741,7 +741,7 @@ def test_products_for_federation_query_count(
         ],
     }
 
-    with django_assert_num_queries(3):
+    with django_assert_num_queries(6):
         response = api_client.post_graphql(query, variables)
         content = get_graphql_content(response)
         assert len(content["data"]["_entities"]) == 1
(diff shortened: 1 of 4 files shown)


8r2y5 (approved):

saleor/graphql/product/tests/benchmark/test_product.py
```diff
@@ -763,7 +763,7 @@ def test_products_for_federation_query_count(
         ],
     }
 
-    with django_assert_num_queries(3):
+    with django_assert_num_queries(6):
```
IKarbowiak, line 766: Why has the number of queries increased?

saleor/graphql/product/tests/benchmark/test_variant.py
```diff
@@ -400,7 +400,7 @@ def test_products_variants_for_federation_query_count(
         ],
     }
 
-    with django_assert_num_queries(3):
+    with django_assert_num_queries(5):
```
IKarbowiak, line 403: Why has the number of queries increased?

tomaszszymanski129:
@IKarbowiak Check is right about it, it did increase, however, this is a much more inexpensive query than doing exists within an exists which caused issues in that particular case.

PR #14941, opened 2023-12-06
Optimize collection update and delete

tomaszszymanski129:
I want to merge this change because it optimizes collection create and delete mutations.
There is a big prefetch going on in these, which doesn't necessary needs to extend the mutation time.

QA tips: create a webhook for product_updated event, create / delete a collection, webhooks should fire for all products inside.

# Impact

- [ ] New migrations
- [ ] New/Updated API fields or mutations
- [ ] Deprecated API fields or mutations
- [ ] Removed API types, fields, or mutations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

- [ ] Privileged queries and mutations are either absent or guarded by proper permission checks
- [ ] Database queries are optimized and the number of queries is constant
- [ ] Database migrations are either absent or optimized for zero downtime
- [ ] The changes are covered by test cases
- [ ] All new fields/inputs/mutations have proper labels added (`ADDED_IN_X`, `PREVIEW_FEATURE`, etc.)
- [ ] All migrations have proper dependencies
- [ ] All indexes are added concurrently in migrations
- [ ] All RunSql and RunPython migrations have revert option defined

PR #14981, opened 2023-12-11
Optimize collections create and delete

tomaszszymanski129:
I want to merge this change because it optimizes collection create and delete mutations.
There is a big prefetch going on in these, which doesn't necessary needs to extend the mutation time. It could also make use of replicas which now it does (after deletion code, so should be safe replication-lag wise).

This is a port of https://github.com/saleor/saleor/pull/14941

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/graphql/product/mutations/collection/collection_create.py
```diff
@@ -111,14 +112,23 @@ def clean_input(cls, info: ResolveInfo, instance, data, **kwargs):
...
         return cleaned_input
 
+    @staticmethod
+    def batch_product_ids(ids):
+        # Batch size of 25k ids, assuming their pks are at least 7 digits each
+        # after json serialization, weights 225kB of payload.
+        BATCH_SIZE = 25000
+        _length = len(ids)
+        for i in range(0, _length, BATCH_SIZE):
+            yield ids[i : min(i + BATCH_SIZE, _length)]
```
korycins, line 122: Do you need a `min`  here?
```
ids = [9,8,7,6,5,4,3,2,1]

In [17]: def a(ids):
    ...:     BATCH_SIZE = 2
    ...:     _length = len(ids)
    ...:     for i in range(0, _length, BATCH_SIZE):
    ...:         yield ids[i : i + BATCH_SIZE]
    ...:

In [18]: print([q for q in a(ids)])
[[9, 8], [7, 6], [5, 4], [3, 2], [1]]


In [20]: ll = [1,2,3]

In [21]: ll[0:1000]
Out[21]: [1, 2, 3]
```

PS. (if not needed, I think it is ok, to only change it for `main` branch)

korycins (approved):

IKarbowiak (commented):
IMO we should at tests for those changes. We can override the `BATCH_SIZE` with some small number and check if the task was delayed for all ids

tomaszszymanski129:
@IKarbowiak Added tests for product batching.

IKarbowiak (approved):

IKarbowiak (approved):

Issue #14926, opened 2023-12-05
Idempotency for the transactions API


patrys:
## Problem

It's possible for a customer to accidentally pay twice for a checkout if they refresh the page during checkout. We need a mechanism for checkouts to enforce the "at most once" strategy for payments.

The proposed solution is to introduce an optional idempotency key that the storefront can pass when initiating payments that would result in both Saleor and the payment service provider (PSP) attempting to deduplicate the request.

The proposed flow would be as follows:

1. Storefront calls `transactionInitialize` with an optional idempotency key
2. If the key is not given, Saleor generates one
3. If a previous transaction for the same checkout matches the provided key, Saleor returns that transaction without further processing
4. Saleor emits `TransactionInitializeSession`, passing the idempotency key
5. The payment handling app passes the key to the PSP (for example, both Adyen and Stripe support it through the `idempotency-key` HTTP header)
6. If the PSP finds a previous transaction, it gets returned without further processing
7. Saleor proceeds as usual

---

### Acceptance Criteria

- Idempotency keys are present for all newly created transactions
- If provided, user-generated keys take precedence
- Automatically generated keys are unique at least within the last 24 hours
- It may be wise to wipe the keys after a few days to reduce the risk of collisions
- If present, the keys are forwarded to the apps handling payments
- A similar mechanism should guard `TransactionProcessSession`, but I'm not sure if it should allow the client to override the idempotency key. We need to evaluate the risk of duplicate processing versus the risk of implementors reusing the same key they used for session initialization and thus breaking some apps.

### Types

```graphql
type TransactionInitializeSession implements Event {
  # …
  idempotencyKey: String!
}

type TransactionProcessSession implements Event {
  # …
  idempotencyKey: String!
}
```

## 3. Database changes

<!-- New models or changes in existing models  -→

## 4. UML diagrams

<!-- You can render UML diagrams using [Mermaid](https://mermaidjs.github.io/). -→

```mermaid

```

## 5. To Do list

<!-- Everything from developer perspective -→

```[tasklist]
### Tasks
- [ ] Add a draft title or issue reference here
```

## 6. To Test

<!--
Put here additional info of what it would be good to check along with new changes.
e.g.:
- Check feature with tax systems
- Test with overrode prices
 -→

- [ ]
- [ ]

lkostrowski:
A few cents:

[...]

PR #15010, opened 2023-12-13
Idempotency key for transaction API

korycins:
Port of changes from #14948 
I want to merge this change because it should solve https://github.com/saleor/saleor/issues/14926

> [!IMPORTANT]  
> This PR introduces the logic for idempotency-key. Some migrations are introduced in separate PR

The implementation propagates the app.identifier with app.id when app.identifier is None (local apps). It was required to apply the unique constraint on `TransactionItem`. The unique constraint is also applied to `TransactionEvent`. 

The introduced constraints are used to block the possibility of recording the same transaction twice (in case of receiving the same requests twice in a short period of time). 
The implementation also doesn't allow to provide the idempotency key as an empty string. The mutation will raise an exception. 
In case of finding the transaction with the same idempotency key, the existing one will be used instead of creating a new one

The PR relays on changes from: https://github.com/saleor/saleor/pull/14991

The logic for idempotency-key is separated into multiple PRs to cover the zero-downtime approach:
- Add idempotency fields
- Add unique indexes concurrently
- Convert indexes into constraints
- Add logic that uses the logic

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation: https://github.com/saleor/saleor-docs/pull/1053

# Pull Request Checklist


zedzior (approved):

saleor/app/migrations/0027_set_identifier_when_missing.py
```diff
@@ -0,0 +1,70 @@
...
+
+def set_identifier_for_local_apps(apps, _schema_editor):
+    App = apps.get_model("app", "App")
+
+    while True:
+        apps = App.objects.filter(identifier__isnull=True)[:APP_BATCH_SIZE]
+        if not apps:
+            break
+
+        app_ids = []
```
IKarbowiak, line 49: Not needed I guess. You are not using anywhere else `app_ids`

IKarbowiak (approved):

witoszekdev:
I've tested it locally with Adyen App, and it works as expected: I've made multiple payments using drop-in at the same time and there was only one in Saleor

Great work 🚀

Issue #13840, opened 2023-08-31
Introduce the flag for products to mark the objects that needs recalculation


IKarbowiak:
We should refactor the way that the product prices are recalculated.

### Problem
Currently the discounted price recalculation method is called each time the related sale is changing. This may result in calling recalculation multiple times for single product what is not efficient.

## Solution
We should introduce similar solution as we already have for search index.
Add the flag on `Product` model to mark the products that needs recalculation. Then we should add a task that will be run each 20 seconds (as it's done for search vector, or maybe other frequency - to discuss) and recalculate the discounted prices only for those products.

maarcingebala:
Task functions to refactor:
- `update_products_discounted_prices_for_promotion_task`
- `update_products_discounted_prices_of_promotion_task`

maarcingebala:
We could try to use a custom Celery scheduler that will schedule the task only if there are products that need recalculating. An example custom schedule based on this logic was added in [this commit](https://github.com/saleor/saleor/pull/14823/commits/c4ed78c6fd984ef271878f811d3087bdf0bc753d#diff-6724102dbc2ea88d9de3ffdef217f9f4ac654211597696902f8aaf0d26bc5637R165).

Issue #15025, opened 2023-12-14
Optimize price recalculation


IKarbowiak:
We can introduce a new object to keeping denormalized prices for given promotion rules, channels, and variant channel listings. The new object will keep the discounted variant price.
- When the promotion is changed the related denormalized objects will be updated.
- In case any catalog changes appear (product/variant/collection/category) the denormalization needs to be updated for all promotion rules.
- The process of discounted price calculations for variants will be significantly simplified. We would just need to fetch related objects for a given variant and channel and get the lower price.
- It will also simplify the `handle_promotion_toggle` - the process of fetching the `product_ids`. It won't need to update promotion rules anymore. We will have to call only discounted price recalculation (`update_discounted_prices_task`)

IKarbowiak:
Resolved in https://github.com/saleor/saleor/pull/12713

PR #15051, opened 2023-12-15
Recalculate prices with celery beat

Air-t:
Calculate the prices in the background. In the previous solution any action on product, variant, sale, promotion, rule was triggering the action of recalculating prices in the background. In case of huge amount of actions this could trigger a lot of duplicated tasks. We change this approach to only set the `dirty` flag in the DB.  The celery workers defined in celery beat will pick up the PromotionRules/Products which needs price recalculation. With this approach, there is no risk, that multiple mutation calls could fully occupy the celery workers with price recalculations.

The changes provides below functions:
- `mark_active_promotion_rules_as_dirty(channel_ids: Iterable[int])` - mark rules with `variants_dirty`. Only rules with relations to `channel_ids` will be re-calculated - there are no reasons to re-calculated the promotion for channel X, where we update promotion/variant/etc for channel Y
- `mark_promotion_rules_as_dirty(promotion_pks: Iterable[UUID])`  - mark all rules assigned to specific promotions as dirty
- `mark_products_as_dirty(channel_to_product_ids: dict[int, set[int]]):` - accepts dictionary `channelID: {productIDs, ...}`. Based on the input the ProductChannelListing assigned to provided channelID will be marked as a dirty. So instead of processing all channelListings, we will process only the one that needs it. 

resolves: https://github.com/saleor/saleor/issues/13840

DoD:
- Adding/Updating/Removing promotion rules have an impact on variant prices
- Modifing Collection/Category/Product/Variant listed/filtered in category-predicate should have an impact on prices

For user, it should work in the same way as previously, one thing which is changed there is the way how we recalculate the prices. It can take maximal +/- 30 seconds after providing the changes to get updated prices.

# Impact

- [ x ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/product/migrations/0190_product_recalculate_discounted_price.py
```diff
@@ -0,0 +1,17 @@
...
+        ("product", "0189_merge_20230929_0857"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="product",
+            name="recalculate_discounted_price",
+            field=models.BooleanField(default=False),
+        ),
+    ]
```
IKarbowiak, line 17: You need to also set the default on database column

saleor/tests/e2e/checkout/discounts/test_checkout_with_promotion_and_voucher.py
```diff
@@ -112,10 +112,10 @@ def prepare_voucher(
...
     ),
     [
-        ("19.99", 13, "PERCENTAGE", 2.60, 13, "PERCENTAGE", 2.26, 15.13),
-        ("30", 10, "FIXED", 10, 5, "FIXED", 5, 17.5),
-        ("56.50", 19.99, "FIXED", 19.99, 33, "PERCENTAGE", 12.05, 24.46),
-        ("77.77", 17.5, "PERCENTAGE", 13.61, 25, "FIXED", 25, 51.66),
+        # ("19.99", 19.99, "PERCENTAGE", 0, 13, "PERCENTAGE", 2.6, 17.39),
+        # ("30", 30, "FIXED", 0, 5, "FIXED", 5, 27.5),
+        ("56.00", 56.00, "FIXED", 0, 30, "PERCENTAGE", 16.80, 39.2),
+        # ("77.77", 77.77, "PERCENTAGE", 0, 25, "FIXED", 25, 65.27),
```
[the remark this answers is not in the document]
Air-t: forgot to uncomment them, they should be uncommented if we still believe that e2e tests in this case are still valid.

saleor/graphql/discount/utils.py
```diff
@@ -54,6 +57,32 @@ def get_products_for_promotion(promotion: Promotion) -> ProductsQueryset:
...
+        return
+    products = get_products_for_promotion(promotion)
+    if products:
+        products.update(recalculate_discounted_price=True)
+
+
+def mark_products_for_recalculate_discounted_price(product_ids: Iterable[int]):
+    """Mark products for recalculate discounted prices."""
+    products = Product.objects.filter(pk__in=product_ids)
+    if products:
```
korycins, line 82: Do you need this `if` here? 
You could make 
```python
Product.objects.filter(pk__in=product_ids).update(recalculate_discounted_price=True)
```
Additionally you could skip the update action when `product_ids` is an empty list

saleor/product/tasks.py
```diff
@@ -92,14 +92,28 @@ def update_products_discounted_prices_of_promotion_task(promotion_pk: UUID):
...
 
+@app.task
+def recalculate_discounted_price_for_products_task():
+    """Recalculate discounted price for products"""
+    products_ids = list(
+        Product.objects.filter(recalculate_discounted_price=True)
+        .order_by("id")
+        .values_list("id", flat=True)
+    )[:DISCOUNTED_PRODUCT_BATCH]
+    if products := Product.objects.filter(id__in=products_ids):
```
korycins, line 111: Why not:
```suggestion
    products = Product.objects.filter(
        recalculate_discounted_price=True
    ).order_by("id")[:DISCOUNTED_PRODUCT_BATCH]
    if products:
```

saleor/settings.py
```diff
@@ -627,6 +627,10 @@ def get_url_from_env(name, *, schemes=None) -> Optional[str]:
         "task": "saleor.payment.tasks.transaction_release_funds_for_checkout_task",
         "schedule": timedelta(minutes=10),
     },
+    "recalculate-discounted-price-for-products": {
+        "task": "saleor.product.tasks.recalculate_discounted_price_for_products_task",
+        "schedule": timedelta(minutes=20),
```
korycins, line 632: Is 20 minutes correct here? 
I mean: It means that we will not update the prices for max 20 minutes (or even more if the amount of products to process is big).
By making a promotion, you could expect that undiscounted prices will show up immediately. Did we document it somewhere?

IKarbowiak: I agree, IMO is too rare

korycins: But if this is too rare,  we can end up with a huge amount of the same celery tasks which will occupy most of the available workers. 
We should have a validation that only one task is under processing. Or provide separate workers to only update the prices.  (we need to define which way is better and safer).

saleor/product/tasks.py
Air-t: there is an issue with `update_discounted_prices_for_promotion` it will not take sliced queryset

pyproject.toml
```diff
@@ -87,6 +87,8 @@ documentation = "https://docs.saleor.io/"
   urllib3 = "^1.26.18"
   uvicorn = {extras = ["standard"], version = "^0.23.1"}
   weasyprint = ">=53.0" # libpango >=1.44 is required
+  ipdb = "^0.13.13"
+  ipython = "^8.18.1"
```
IKarbowiak, line 91: Do you add this intentionally?

saleor/product/tasks.py
```diff
@@ -137,14 +137,28 @@ def update_products_discounted_prices_for_promotion_task(
...
+        update_discounted_prices_for_promotion(products)
```
IKarbowiak, line 157: It must be changed. We should do something similar like `update_products_discounted_prices_for_promotion_task` but without providing `product_ids`. In this solution without providing `rule_ids` we will have to fetch variants for each single `PromotionRule` for each batch!
We need to fetch this only once. 
Also, updating all `PromotionRules` each time the discounted task is called will kill us. Especially when the task is calls each 20 seconds. I'm afraid that this improvement cannot be merged without [this one](https://github.com/saleor/saleor/issues/15025)

IKarbowiak (changes requested):

pyproject.toml
Air-t: no this are added for debugging only - removing

saleor/product/utils/variants.py
```diff
@@ -54,11 +54,12 @@ def get_variant_selection_attributes(
 
 def fetch_variants_for_promotion_rules(
     rules: QuerySet[PromotionRule],
-):
+) -> List[int]:
```
IKarbowiak, line 57:
```suggestion
) -> list[int]:
```

saleor/product/utils/variants.py
```diff
@@ -54,11 +54,12 @@ def get_variant_selection_attributes(
...
     from ...graphql.discount.utils import get_variants_for_predicate
 
     PromotionRuleVariant = PromotionRule.variants.through
     promotion_rule_variants = []
+    variant_ids: List[int] = []
```
IKarbowiak, line 62:
```suggestion
    variant_ids: list[int] = []
```

saleor/product/utils/product.py
```diff
@@ -0,0 +1,8 @@
+from typing import Iterable
+
+from saleor.product.models import Product
```
IKarbowiak, line 3: Use relative import

saleor/product/tasks.py
```diff
@@ -137,6 +141,54 @@ def update_products_discounted_prices_for_promotion_task(
...
+        .filter(
+            Exists(promotions.filter(id=OuterRef("promotion_id"))),
+            variants_dirty=True,
+            **kwargs,
+        )[:PROMOTION_RULE_BATCH_SIZE]
+    )
+    if ids := list(rules.values_list("pk", flat=True)):
+        qs = PromotionRule.objects.filter(pk__in=ids)
+        product_ids = fetch_variants_for_promotion_rules(rules=qs)
+        update_promotion_rules_variants_dirty(False)
```
IKarbowiak, line 170: You already have queryset here so it's better to just do `qs.update(variants_dirty=False)`

saleor/product/tasks.py
```diff
@@ -137,6 +141,54 @@ def update_products_discounted_prices_for_promotion_task(
...
+        .using(settings.DATABASE_CONNECTION_REPLICA_NAME)
+        .filter(
+            Exists(promotions.filter(id=OuterRef("promotion_id"))),
+            variants_dirty=True,
+            **kwargs,
+        )[:PROMOTION_RULE_BATCH_SIZE]
+    )
+    if ids := list(rules.values_list("pk", flat=True)):
+        qs = PromotionRule.objects.filter(pk__in=ids)
+        product_ids = fetch_variants_for_promotion_rules(rules=qs)
```
IKarbowiak, line 169: Those ids are variant ids

saleor/discount/utils.py
```diff
@@ -581,3 +583,25 @@ def get_current_products_for_rules(rules: "QuerySet[PromotionRule]"):
...
     )
     return Product.objects.filter(Exists(variants.filter(product_id=OuterRef("id"))))
+
+
+def update_promotion_rules_variants_dirty(is_dirty):
+    rules_to_update = []
+    for rule in get_active_promotion_rules():
+        rule.variants_dirty = is_dirty
+        rules_to_update.append(rule)
+    PromotionRule.objects.bulk_update(rules_to_update, ["variants_dirty"])
```
IKarbowiak, line 593: Instead of iterating, we can just do `update` on the whole query set

saleor/discount/utils.py
```diff
@@ -581,3 +583,25 @@ def get_current_products_for_rules(rules: "QuerySet[PromotionRule]"):
...
+def get_active_promotion_rules():
+    promotions = Promotion.objects.using(
+        settings.DATABASE_CONNECTION_REPLICA_NAME
+    ).active()
+    return (
+        PromotionRule.objects.order_by("id")
+        .using(settings.DATABASE_CONNECTION_REPLICA_NAME)
+        .filter(
+            Exists(promotions.filter(id=OuterRef("promotion_id"))),
+            catalogue_predicate__isnull=False,
```
IKarbowiak, line 605: The default value for `catalogue_predicate` is empty dict

saleor/discount/utils.py
```diff
@@ -581,3 +583,21 @@ def get_current_products_for_rules(rules: "QuerySet[PromotionRule]"):
...
+    rules.update(variants_dirty=is_dirty)
+
+
+def get_active_promotion_rules():
+    promotions = Promotion.objects.using(
+        settings.DATABASE_CONNECTION_REPLICA_NAME
+    ).active()
+    return (
+        PromotionRule.objects.order_by("id")
+        .using(settings.DATABASE_CONNECTION_REPLICA_NAME)
```
IKarbowiak, line 599: When we are doing update on the queryset I suppose we shouldn't user replica here

saleor/product/tasks.py
IKarbowiak: Not resolved yet

saleor/product/utils/variants.py
```diff
@@ -69,6 +70,7 @@ def fetch_variants_for_promotion_rules(
                 for variant_id in set(variants.values_list("pk", flat=True))
             ]
         )
+        product_ids += variants.values_list("product", flat=True)
```
IKarbowiak, line 73: product_ids += variants.values_list("product_id", flat=True)

Air-t: that is exactly the same query no mater if you use `product_id` or just `product`. Is this just a convention to use `product_id` or something else?

IKarbowiak: I was afraid that using `product` causes fetching product instances. `product_id` explicitly says that we only want the `products_id` which is a column on the variant table. If `variants.values_list("product", flat=True)` is doing the same underhood, it's ok

saleor/product/tasks.py
```diff
@@ -140,6 +151,54 @@ def update_products_discounted_prices_for_promotion_task(
         update_discounted_prices_task.delay(product_ids)
 
 
+@app.task
+def update_promotion_rules_mark_products_for_price_recalculation_task(
+    product_ids: Iterable[int] = [],
+    start_id: Optional[UUID] = None,
+):
+    """Update PromotionRule and ProductVariant.
```
IKarbowiak, line 159: "Update relations between PromotionRule and ProductVariant.

IKarbowiak (commented):
Looks good for now 🚀

saleor/graphql/product/bulk_mutations/collection_bulk_delete.py
```diff
@@ -47,6 +47,4 @@ def bulk_action(cls, info, queryset):
         for product in products:
             cls.call_event(manager.product_updated, product, webhooks=webhooks)
 
-        update_products_discounted_prices_for_promotion_task.delay(
-            [product.id for product in products]
-        )
+        get_active_promotion_rules().update(variants_dirty=True)
```
korycins, line 50: This will make ALL existing rules marked as dirty, which means, that we will re-calculate everything.

saleor/product/tasks.py
```diff
@@ -100,6 +101,16 @@ def update_products_discounted_prices_of_promotion_task(promotion_pk: UUID):
     update_discounted_prices_task.delay(list(products.values_list("id", flat=True)))
 
 
+@app.task
+def mark_promotion_rules_to_recalculate_variants(promotion_pk: UUID):
```
korycins, line 105: Do we need to have this as celery task? It looks like you're calling it as normal function.

saleor/graphql/product/bulk_mutations/collection_bulk_delete.py
IKarbowiak: Yep, but we did it already and it was done each time any catalogue have been changed. No, we will mark the rules with flag and we will significantly reduce the number of proceeded promotion rules.
When something have changed in catalogues we have no clue which promotion rules should be marked to recalculate, as rules include filters and we do not know if the given product, variant, category, or collection is included in any of this rule until we filter be the rule predicate

saleor/discount/tasks.py
```diff
@@ -308,7 +309,7 @@ def update_discounted_prices_task():
     if products_ids:
         products = Product.objects.filter(id__in=products_ids)
         update_discounted_prices_for_promotion(products)
-        update_discounted_prices_task.delay()
+        mark_products_for_recalculate_discounted_price(products_ids)
```
korycins, line 312: I might be wrong but it looks like we will end up in infitie loop. We just processed the `products_ids`, and now we will mark it for recalculate

zedzior (commented):
So far I found some cosmetic drawbacks only. Will look further into it on Monday

saleor/discount/utils.py
```diff
@@ -1011,16 +1012,53 @@ def fetch_promotion_rules_for_checkout(
...
+        settings.DATABASE_CONNECTION_REPLICA_NAME
+    ).filter(Exists(rules.filter(id=OuterRef("promotionrule_id"))))
     for promotion_rule_channel in promotion_rule_channels:
         rule_id = promotion_rule_channel.promotionrule_id
         channel_id = promotion_rule_channel.channel_id
         rule_to_channel_ids_map[rule_id].append(channel_id)
     return rule_to_channel_ids_map
 
 
+def get_channel_to_products_map_from_rules(
```
IKarbowiak, line 1025: IMO, it's hard to understand what the function is doing from the name.
As I see it must be called before `mark_products_as_dirty`. So maybe both methods should be mixed? 
Right now you need to prepare specific mapping for `mark_products_as_dirty` but maybe it could be done inside the method? WDYT?

saleor/discount/utils.py
```diff
@@ -1036,6 +1074,21 @@ def get_current_products_for_rules(rules: "QuerySet[PromotionRule]"):
...
+
+    Collect all channels for are assigned to promotion rules.
+    """
+    PromotionRuleChannel = PromotionRule.channels.through
+    rule_channels = PromotionRuleChannel.objects.filter(
+        Exists(rules.filter(pk=OuterRef("promotionrule_id")))
+    )
+    channels = Channel.objects.filter(
+        Exists(rule_channels.filter(channel_id=OuterRef("id")))
+    )
```
IKarbowiak, line 1088: We can add `using` with replica, right?

saleor/discount/utils.py
```diff
@@ -1076,3 +1129,53 @@ def update_rule_variant_relation(
         PromotionRuleVariant.objects.bulk_create(
             rules_variants_to_add, ignore_conflicts=True
         )
+
+
+def get_active_promotion_rules(
```
IKarbowiak, line 1134:
```suggestion
def get_active_catalogue_promotion_rules(
```

saleor/discount/utils.py
```diff
@@ -1076,3 +1129,53 @@ def update_rule_variant_relation(
...
+            Exists(promotions.filter(id=OuterRef("promotion_id"))),
+        )
+        .exclude(catalogue_predicate={})
+    )
+    if allow_replica:
+        rules = rules.using(settings.DATABASE_CONNECTION_REPLICA_NAME)
+    return rules
+
+
+def mark_active_promotion_rules_as_dirty(channel_ids: Iterable[int]):
```
IKarbowiak, line 1152:
```suggestion
def mark_active_catalogue_promotion_rules_as_dirty(
```

saleor/discount/utils.py
```diff
@@ -1076,3 +1129,53 @@ def update_rule_variant_relation(
...
+    rules = get_active_promotion_rules()
+    PromotionRuleChannel = PromotionRule.channels.through
+    promotion_rules = PromotionRuleChannel.objects.filter(channel_id__in=channel_ids)
+    rules = rules.filter(
+        Exists(promotion_rules.filter(promotionrule_id=OuterRef("id")))
+    )
+    rules.update(variants_dirty=True)
+
+
+def mark_promotion_rules_as_dirty(promotion_pks: Iterable[UUID]):
```
IKarbowiak, line 1171:
```suggestion
def mark_catalogue_promotion_rules_as_dirty(promotion_pks: Iterable[UUID]):
```

saleor/graphql/discount/mutations/utils.py
```diff
@@ -52,3 +53,26 @@ def update_variants_for_promotion(
...
+    the rule has all details required to calculate the discount.
+    Keep in mind that in case of update applied on rule, this function should not be
+    called, previous version of rule could have an impact on products.
+    """
+    if promotion_type != PromotionType.CATALOGUE:
+        return False
+    if rule.reward_value is None:
+        return False
+    if rule.reward_value <= 0:
+        return False
```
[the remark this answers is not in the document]
korycins: The current logic responsible for triggering the re-calculation has:
```python
    rules = (
        PromotionRule.objects.order_by("id")
        .using(settings.DATABASE_CONNECTION_REPLICA_NAME)
        .filter(Exists(promotions.filter(id=OuterRef("promotion_id"))), **kwargs)
        .exclude(
            Q(reward_value__isnull=True) | Q(reward_value=0) | Q(catalogue_predicate={})
        )[:PROMOTION_RULE_BATCH_SIZE]
    )
```
This is why I assumed that I will also validate it here.

saleor/discount/utils.py
korycins: Looks like we can fully drop this function.

zedzior (approved):

IKarbowiak (approved):

saleor/graphql/product/mutations/category/category_update.py
```diff
@@ -36,6 +38,15 @@ def construct_instance(cls, instance, cleaned_data):
         return super().construct_instance(instance, cleaned_data)
 
     @classmethod
-    def post_save_action(cls, info: ResolveInfo, instance, _cleaned_input):
+    def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
         manager = get_plugin_manager_promise(info.context).get()
         cls.call_event(manager.category_updated, instance)
+
+        if "metadata" in cleaned_input:
```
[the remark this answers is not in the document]
korycins: Right. Added the comment to Category/Collection where input.

IKarbowiak (approved):

yellowee (approved):

PR #15150, opened 2024-01-02
Fix return value type in `send_webhook_using_aws_sqs`

przlada:
I want to merge this change because it addresses the issue related to the return value type from the `send_webhook_using_aws_sqs` function. The problem is resolved by JSON encoding the SQS response.

As it is [described in the docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs/client/send_message.html#send-message), the Boto3 SQS client's `send_message` method returns a dictionary, which was incorrectly treated in `send_webhook_using_aws_sqs` function as a string. This resulted in a failure in the Observability reporter, which went previously undetected as Django's `TextField` automatically converted the dictionary to a string under the hood.

This change should not be a breaking-change, as we do not store successful `EventDeliverAttempt`s (so SQS responses weren't accessible through GraphQL API), and Observability events failed to be created due to the type mismatch.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/webhook/transport/tests/test_utils.py
+++ b/saleor/webhook/transport/tests/test_utils.py
@@ -1,8 +1,7 @@
 import datetime
 import json
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
-import boto3
 import pytest
 from celery import Task
 from celery.exceptions import Retry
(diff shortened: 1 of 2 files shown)


maarcingebala (approved):

jakubkuc (approved):

saleor/webhook/transport/tests/test_utils.py
```diff
@@ -74,3 +78,90 @@ def test_webhook_retry_redirect(webhook, event_delivery):
...
+            "SaleorApiUrl": {
+                "DataType": "String",
+                "StringValue": f"http://{domain}/graphql/",
+            },
+            "EventType": {"DataType": "String", "StringValue": event_type},
+            "Signature": {"DataType": "String", "StringValue": signature},
+        },
+        "MessageBody": message,
+    }
+    if is_fifo:
```
aniav, line 163: As a rule of thumb if you need to use an `if` statement in your tests this means you should split the test into more than one. In your case `parametrize` is used to incorporate the logic that should be represented in the name/body of the test and the single True/False in the parametrize statement is an additional overhead someone will have to take on when they read the test. So this test should be split into two, one for fifo and one without.

aniav (approved):
Minor test changes and it's good to go.

saleor/webhook/transport/tests/test_utils.py
przlada: Great point, thank you. I've split this test as suggested

Issue #15145, opened 2023-12-29
Improve creation of promotion rule - variant relation


IKarbowiak:
Refactor `fetch_variants_for_promotion_rules`.
We shouldn't delete and create a new one `PromotionRuleVariant` each time the method is called. Instead, we should check if something has changed, remove non-valid relations, and create new ones if any appear.
[Here](https://github.com/saleor/saleor/blob/main/saleor/product/utils/variants.py#L73) is the part that needs refactoring.

maarcingebala:
As part of this task, we should estimate whether the `discount_promotionrule_variants.id` field needs to be an `integer` or `bigint` type in Postgres. We changed it to `bigint` in one of the deployments, as the limit of regular integer was exceeded due to many inserts to that table. We should answer these questions: 
- how many rules vs how many variants can be supported if we used integer type
- decide whether there should be any API limits for the number of supported variants/rules
- decide whether we should change this field to `bigint` in code, and create a migration for that

PR #15176, opened 2024-01-05
Improve creation of PromotionRule-ProductVariant relation

zedzior:
I want to merge this change, because we shouldn't delete and create a new one PromotionRuleVariant each time the method is called. Instead, we should check if something has changed, remove non-valid relations, and create new ones if any appear.

Issue: https://github.com/saleor/saleor/issues/15145

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/discount/utils.py
+++ b/saleor/discount/utils.py
@@ -12,6 +12,7 @@ from typing import (
 from uuid import UUID
 
 import graphene
+from django.db import transaction
 from django.db.models import Exists, F, OuterRef, QuerySet
 from django.utils import timezone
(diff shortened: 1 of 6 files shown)


saleor/discount/utils.py
IKarbowiak, line 1: I'll suggest adding one more test with the use of `before_after` to ensure that the deletion and creation are working as expected when this method is called two times in parallel

IKarbowiak (approved):

SzymJ (approved):

yellowee (approved):

PR #15179, opened 2024-01-05
Optimize price calculation

Air-t:
New discount prices for products are no longer calculated after a certain action is done. 
First `PromotionRule`s are marked to be recalculated. The celery beat task every 10 seconds recalculates all or a chosen `PromotionRule`s. Once they are recalculated, `Product`s affected by that change are marked to recalculate discounted price. A celery beat task is executed every 10 seconds to recalculate product prices.

This is a POC PR done for a `CollectionDelete` mutation. To be extended when approved.

resolves: https://github.com/saleor/saleor/issues/13840

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/discount/models.py
```diff
@@ -384,6 +384,7 @@ class PromotionRule(models.Model):
         blank=True,
     )
     old_channel_listing_id = models.IntegerField(blank=True, null=True, unique=True)
+    recalculate_rule = models.BooleanField(default=False)
```
IKarbowiak, line 387: I think we should find out a better name for this field. For example to mark that product `search_document` needs recalculation we have a field `search_index_dirty`. Maybe we should use something like `variants_dirty`? We can discuss the field name with the team as well

saleor/product/models.py
```diff
@@ -207,6 +207,7 @@ class Product(SeoModel, ModelWithMetadata, ModelWithExternalReference):
         null=True,
         on_delete=models.SET_NULL,
     )
+    recalculate_discounted_price = models.BooleanField(default=False)
```
IKarbowiak, line 210: Similar here, for consistency we can consider changing the name to `discounted_price_dirty`

saleor/discount/utils.py
```diff
@@ -581,3 +582,8 @@ def get_current_products_for_rules(rules: "QuerySet[PromotionRule]"):
         Exists(rule_variants.filter(productvariant_id=OuterRef("id")))
     )
     return Product.objects.filter(Exists(variants.filter(product_id=OuterRef("id"))))
+
+
+def mark_products_for_recalculate_discounted_price(product_ids: Iterable[int]):
+    """Mark products for recalculate discounted prices."""
+    Product.objects.filter(pk__in=product_ids).update(recalculate_discounted_price=True)
```
IKarbowiak, line 589: IMO this should be in `product` rather than in `discount` module

saleor/product/tasks.py
```diff
@@ -137,6 +140,54 @@ def update_products_discounted_prices_for_promotion_task(
         update_discounted_prices_task.delay(product_ids)
 
 
+@app.task
+def update_promotion_rules_mark_products_for_price_recalculation_task(
+    product_ids: Iterable[int] = [],
+    start_id: Optional[UUID] = None,
+):
+    """Mark the product discounted prices for recalculation when PromotionRule
+    is updated.
```
IKarbowiak, line 149: This should be a short description of what the function is doing, and I would say that the main purpose of the function is updating` PromotionRule - ProductVariant` relations, and marking the product for recalculation is less important so it should be in the next part of docstring if doesn't fit in the main sentence.

saleor/product/tasks.py
```diff
@@ -137,6 +140,54 @@ def update_products_discounted_prices_for_promotion_task(
...
+        .filter(
+            Exists(promotions.filter(id=OuterRef("promotion_id"))),
+            recalculate_rule=True,
+            **kwargs,
+        )[:PROMOTION_RULE_BATCH_SIZE]
+    )
+    if ids := list(rules.values_list("pk", flat=True)):
+        qs = PromotionRule.objects.filter(pk__in=ids)
+        product_ids = fetch_variants_for_promotion_rules(rules=qs)
+        update_products_discounted_prices_for_promotion_task.delay(product_ids, ids[-1])
```
IKarbowiak, line 170: We should also change the flag `recalculate_rule` of `PromotionRule`s to `False`

saleor/product/tasks.py
```diff
@@ -198,6 +249,7 @@ def collection_product_updated_task(product_ids):
                 replica_products_count,
                 len(products),
             )
+    PromotionRule.objects.all().update(recalculate_rule=True)
```
IKarbowiak, line 252: We should filter by active and those that have not empty `catalogue_predicate`

saleor/product/utils/variants.py
```diff
@@ -69,6 +70,7 @@ def fetch_variants_for_promotion_rules(
                 for variant_id in set(variants.values_list("pk", flat=True))
             ]
         )
+        products_ids += variants.values_list("id", flat=True)
```
IKarbowiak, line 73: It's `variant_ids` not `product_ids`.

saleor/product/models.py
Air-t: sure, if we go with `dirty` convention this one also needs rename

saleor/product/tasks.py
Air-t: good catch

PR #15241, opened 2024-01-17
Fix Deprecation: datetime.datetime.utcnow()

almamun-turing:

As of python 3.12, datetime.utcnow is deprecated (since it returns a naive datetime rather than a timezone-aware datetime).
The suggested fix is to use datetime.now with a specific timezone info of UTC instead.\
https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

# Impact

- [ ] New migrations

# Docs
This is an internal code change I think we don't need to update the documentation.

# Pull Request Checklist

### DIFF
--- a/saleor/core/jwt.py
+++ b/saleor/core/jwt.py
@@ -1,5 +1,5 @@
 from collections.abc import Iterable
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from typing import Any, Optional
 
 import graphene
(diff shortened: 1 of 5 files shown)


korycins (approved):

zedzior (dismissed):

zedzior:
Some tests are failing

aniav:
Hey @almamun-turing you need to sign the Contributor License Agreement. We cannot merge these changes without it.

Aside from that let us know if you will have time to fix the broken tests 🙌

aniav:
Hey @almamun-turing, sorry for us taking so long to reply and thank you for your contribution.

I see the deprecation has already been addressed in other changes on the repository in the meantime. I am going to go ahead and close this PR.

Issue #15233, opened 2024-01-12
Bug: `CheckoutDiscount.type` is always set to `VOUCHER`


zedzior:
### What are you trying to achieve?

During `completeCheckout` mutation, when discount derives from promotion, I want to see the proper `CheckoutDiscount.type` ('promotion') 

### Steps to reproduce the problem

1. Fetch `test_promotion_applied_on_order_with_specified_total` branch.
2. Run `test_promotion_applied_on_checkout_with_specific_subtotal_CORE_2133`
3. Place breakpoint at the end of the test
4. Get `Order.objects.first().discounts.first().type`
5. It is `voucher` instead of `promotion`

### What did you expect to happen?

`CheckoutDiscount.type` should be set to 'promotion'

### Environment

Saleor version: …
OS and version: …

zedzior:
https://github.com/saleor/saleor/pull/15248

PR #15248, opened 2024-01-18
Fix `OrderDiscount.type` when converting checkout to order.

zedzior:
I want to merge this change because it fixes https://github.com/saleor/saleor/issues/15233.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/complete_checkout.py
+++ b/saleor/checkout/complete_checkout.py
@@ -1058,16 +1058,10 @@ def _create_order_discount(order: "Order", checkout_info: "CheckoutInfo"):
     if checkout.discount:
         checkout_discount = checkout.discounts.first()
         if checkout_discount and checkout_discount.type == DiscountType.ORDER_PROMOTION:
-            order.discounts.create(
-                type=checkout_discount.type,
-                value_type=checkout_discount.value_type,
-                value=checkout_discount.value,
-                name=checkout_discount.name,
-                translated_name=checkout_discount.translated_name,
-                currency=checkout_discount.currency,
-                amount_value=checkout_discount.amount_value,
-                promotion_rule=checkout_discount.promotion_rule,
-            )
+            discount_data = model_to_dict(checkout_discount)
+            discount_data["promotion_rule"] = checkout_discount.promotion_rule
+            del discount_data["checkout"]
+            order.discounts.create(**discount_data)
 
         if not checkout_discount:
             # store voucher as a fixed value as it this the simplest solution for now.
--- a/saleor/checkout/complete_checkout.py
+++ b/saleor/checkout/complete_checkout.py
@@ -1055,31 +1055,38 @@ def _handle_allocations_of_order_lines(
 
 def _create_order_discount(order: "Order", checkout_info: "CheckoutInfo"):
     checkout = checkout_info.checkout
-    if checkout.discount:
-        checkout_discount = checkout.discounts.first()
-        if checkout_discount and checkout_discount.type == DiscountType.ORDER_PROMOTION:
-            discount_data = model_to_dict(checkout_discount)
-            discount_data["promotion_rule"] = checkout_discount.promotion_rule
-            del discount_data["checkout"]
-            order.discounts.create(**discount_data)
+    checkout_discount = checkout.discounts.first()
+    is_voucher_discount = checkout.discount and not checkout_discount
+    is_promotion_discount = (
+        checkout_discount and checkout_discount.type == DiscountType.ORDER_PROMOTION
+    )
 
-        if not checkout_discount:
-            # store voucher as a fixed value as it this the simplest solution for now.
-            # This will be solved when we refactor the voucher logic to use .discounts
-            # relations
-            order.discounts.create(
-                type=DiscountType.VOUCHER,
-                value_type=DiscountValueType.FIXED,
-                value=checkout.discount.amount,
-                name=checkout.discount_name,
-                translated_name=checkout.translated_discount_name,
-                currency=checkout.currency,
-                amount_value=checkout.discount_amount,
-                voucher=checkout_info.voucher,
-                voucher_code=checkout_info.voucher_code.code
-                if checkout_info.voucher_code
-                else None,
-            )
+    if is_promotion_discount:
+        discount_data = model_to_dict(checkout_discount)
+        discount_data["promotion_rule"] = checkout_discount.promotion_rule
+        del discount_data["checkout"]
+        order.discounts.create(**discount_data)
+
+    if is_voucher_discount:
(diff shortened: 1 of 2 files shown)


saleor/checkout/complete_checkout.py
```diff
@@ -1053,21 +1053,39 @@ def _handle_allocations_of_order_lines(
...
+            order.discounts.create(
+                type=checkout_discount.type,
+                value_type=checkout_discount.value_type,
+                value=checkout_discount.value,
+                name=checkout_discount.name,
+                translated_name=checkout_discount.translated_name,
+                currency=checkout_discount.currency,
+                amount_value=checkout_discount.amount_value,
+                promotion_rule=checkout_discount.promotion_rule,
+            )
```
IKarbowiak, line 1070: As we already have the object that is almost the same. We can do something like that:
```
discount_data = model_to_dict(checkout_discount)
order.discount.create(**discount_data)
```

IKarbowiak (approved):

IKarbowiak (approved):

saleor/checkout/complete_checkout.py
```diff
@@ -1053,21 +1053,33 @@ def _handle_allocations_of_order_lines(
...
-            type=DiscountType.VOUCHER,
-            value_type=DiscountValueType.FIXED,
-            value=checkout.discount.amount,
-            name=checkout.discount_name,
-            translated_name=checkout.translated_discount_name,
-            currency=checkout.currency,
-            amount_value=checkout.discount_amount,
-        )
+        checkout_discount = checkout.discounts.first()
+        if checkout_discount and checkout_discount.type == DiscountType.ORDER_PROMOTION:
```
korycins, line 1060: Instead of two `if`'s, where bot of them create the `OrderDiscount`. Can't we make a single point of calling `order.discounts.create`, and above this, have a logic that will prepare the data?
Something like this:
```python
    if checkout.discount:
        order_discou_data = {}
        if voucher:
            order_discout_data = {...}
        elif checkout_discount.type == DiscountType.ORDER_PROMOTION:
            order_discout_data = {...}
        # create OrderDiscount
```

saleor/checkout/complete_checkout.py
```diff
@@ -1053,21 +1053,33 @@ def _handle_allocations_of_order_lines(
...
+            discount_data = model_to_dict(checkout_discount)
```
korycins, line 1061: From one type of object, you're creating different type of object. If `CheckoutDiscount` receives a new field, which will not be available on `OrderDiscount`, the `order.discount.create` will raise an exception, as you don't explicitly define which fields should be passed. 
If you don't want to explicitly provide each field, you can always define the fields that you want to receive from `model_to_dict` (`fields=`).
Then probably `del discount_data["checkout"]` will not be needed

saleor/checkout/complete_checkout.py
```diff
@@ -1053,21 +1053,33 @@ def _handle_allocations_of_order_lines(
...
+            discount_data["promotion_rule"] = checkout_discount.promotion_rule
+            del discount_data["checkout"]
+            order.discounts.create(**discount_data)
+
+        if not checkout_discount:
```
korycins, line 1066: This could be tricky to understand for someone who is not fully familiar with promotions, and checkout-flow. 
Honestly, I am not sure, why in case of having `not checkout_discount` it means that we should prepare the voucher discount. This should be determined by some kind of the flag, or based on object existence.
Maybe adding `voucher` as an additional parameter `voucher` to `_create_order_discount`, could be helpful here?

saleor/checkout/complete_checkout.py
korycins: Because you explicitly say what fields you want to fetch. Defining the included fields will result in the same received data, no matter if the CheckoutDiscount object will be extended or not. Without this, all fields will be returned always.
Additionally, someone who checks the code, see what fields are exactly returned and needed.

zedzior: I get your point. However, I was thinking that way: if we extend `CheckoutDiscount` with some field, which doesn't fit to `OrderDiscount`, we will catch this error in tests. But if we extend `CheckoutDiscount` with a field, which fit to `OrderDiscount` we will need to update the list of fields, but no test will remind us about it. Does it make sense or not?

saleor/checkout/complete_checkout.py
korycins: I was just thinking about having a place to build the payload for `OrderDiscount`, and one place for calling `order.discount.create`. It is a cosmetic detail, but IMHO it is simpler to read the code. 
Something like this:
```python

def _get_data_for_order_discount(checkout_info: "CheckoutInfo"):
    """Prepare order discount data based on discounts assigned to Checkout"""

    checkout_discount = checkout_info.checkout.discounts.first()

    if checkout_discount and checkout_discount.type == DiscountType.ORDER_PROMOTION:
        data = model_to_dict(checkout_discount, fields=[
            "type",
            "value_type",
            "value",
            "name",
            "translated_name",
            "currency",
            "amount_value",
        ])
        data["promotion_rule"] = checkout_discount.promotion_rule
        return data

    # Is it correct? Or maybe it should be `not checkout_discount`
    if checkout_info.voucher:
        # return the data based on the voucher data
        return {
            "type" :DiscountType.VOUCHER,
            "value_type" : DiscountValueType.FIXED,
            "value" : checkout_info.checkout.discount.amount,
            "name" : checkout_info.checkout.discount_name,
            "translated_name" : checkout_info.checkout.translated_discount_name,
            "currency" : checkout_info.checkout.currency,
            "amount_value" : checkout_info.checkout.discount_amount,
            "voucher" : checkout_info.voucher,
            "voucher_code" : checkout_info.voucher_code.code
            if checkout_info.voucher_code
            else None,
        }




def _create_order_discount(order: "Order", checkout_info: "CheckoutInfo"):
    if not checkout.discount:
        return None
    discount_data = _get_data_for_order_discount(checkout_info)
    if not discount_data:
        return None
    order.discounts.create(**discount_data)
```
But this is only a suggestion.

saleor/checkout/complete_checkout.py
korycins: > if we extend CheckoutDiscount with some field, which doesn't fit to OrderDiscount, we will catch this error in tests.
But in that case doesn't that mean, that you don't want to have this field in OrderDiscount? You have field A in CheckoutDiscount but not in OrderDiscount. 
If I am adding a new field only to CheckoutDiscount, the logic will be working as previously. If I expect that new field will be also propagated on OrderDiscount, I assume that I will add a new test or modify the existing one which will fail as the field is not propagated. 

My comment was only suggestions, so if you don't agree with me, we can also leave this as it is 😄

korycins (approved):

Issue #15215, opened 2024-01-10
PromotionCreate raise an error when lack of channels


IKarbowiak:
Issue: https://saleor.sentry.io/issues/4837696200/?alert_rule_id=5529758&alert_type=issue&notification_uuid=120c0f64-c05f-49b9-b3a2-d4b63985e8b0&project=5381657&referrer=slack

PR #15249, opened 2024-01-18
Fix `promotionCreate` mutation to allow creating rules with no channels specified.

zedzior:
Port to: https://github.com/saleor/saleor/pull/15220
Issue: https://github.com/saleor/saleor/issues/15215

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/discount/mutations/promotion/promotion_create.py
+++ b/saleor/graphql/discount/mutations/promotion/promotion_create.py
@@ -183,7 +183,8 @@ class PromotionCreate(ModelMutation):
             models.PromotionRule.objects.bulk_create(rules)
 
         for rule, channels in rules_with_channels_to_add:
-            rule.channels.set(channels)
+            if channels:
+                rule.channels.set(channels)
 
         return rules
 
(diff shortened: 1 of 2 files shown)


IKarbowiak (approved):

saleor/graphql/discount/mutations/promotion/promotion_create.py
```diff
@@ -183,7 +183,8 @@ def _save_m2m(
             models.PromotionRule.objects.bulk_create(rules)
 
         for rule, channels in rules_with_channels_to_add:
-            rule.channels.set(channels)
+            if channels:
```
korycins, line 186: Name `rules_with_channels_to_add` suggest me that I have a `rule` with `channels`. But it looks like, there is a possibility that `channels` will be None. 
So my question is: why we add rule without a channel to `rules_with_channels_to_add` ? For me it looks like this `if` should be place between lines: 180 and 181.
I know, that this PR is port of changes, but maybe we could clean it up on main at least.

korycins (approved):

saleor/graphql/discount/mutations/promotion/promotion_create.py
zedzior: https://github.com/saleor/saleor/pull/15260

PR #14985, opened 2023-12-11
Bulk create attribute value

8r2y5:
I want to merge this change because it fixes https://github.com/saleor/saleor/issues/13700

Measurements were done by performing mutation `productBulkCreate` and logging db queries (counting them).
| SELECT | 1 product | 100 products |  optimized 1 products | optimized 100 products |
| --- | --- | --- | --- | --- | 
| account_user | 3 |  3 | 3 |  3 |
| channel_channel | 2 | 2 | 2 |  2 |
| product_producttype | 12 | 411 | 12 |  286 |
| product_category | 3 | 85 | 3 |  85 |
| product_product | 5 | 168 | 5 |  168 |
| warehouse_stock | 1 | 1 | 1 |  1 |
| attribute_attributeproduct | 4 | 52 | 4 | 100 |
| product_productvariant | 4 | 131 | 4 |  131 |
| attribute_attributevalue | 20 | 670 | 20 |  318 |
| warehouse_warehouse | 2 | 2 | 2 |  2 |
| attribute_attribute | 11 | 366 | 11 |  366 |
| attribute_attributevariant | 4 | 144 | 4 |  144 |
| attribute_assignedproductattribute | 4 | 60 | 4 |  33 |
| attribute_assignedproductattributevalue | 4 | 84 | 4 |  9 |
| attribute_assignedvariantattribute | 3 | 152 | 3 |  125 |
| attribute_assignedvariantattributevalue | 6 | 268 | 6 |  9 |

| UPDATE | 1 product | 100 products | optimized 1 products | optimized 100 products |
| --- | --- | --- | --- | --- |
| account_user | 1 | 1 | 1 | 1 |
| attribute_assignedproductattributevalue | 1 | 33 | 1 | 33 |
| attribute_assignedvariantattributevalue | 3 | 133 | 3 | 125 |

| INNER JOIN | 1 product | 100 products | optimized 1 product | optimized 100 products |
| --- | --- | --- | --- | --- |
| warehouse_warehouse | 1 | 1 | 1 | 1 |
| warehouse_channelwarehouse | 1 | 1 | 1 | 1 |
| channel_channel | 1 | 1 | 1 | 1 |
| attribute_attributevariant | 13 | 506 | 13 | 506 |
| attribute_attributeproduct | 3 | 77 | 3 | 77 | 
| attribute_assignedproductattributevalue | 3 | 51 | - | - |
| attribute_attributevalue | 6 | 194 | - | - | 
[...]

PR #15269, opened 2024-01-23
Fix saving attribute values in order

jakubkuc:
I want to merge this change because it fixes bug introduced in #14985 with attribute values not been saved in correct order on update.  and fixes test that was modified by it but shouldn't have been. 
The mentioned PR was already merged but not ported to higher versions so this fix should be implemented within the ports.

To reproduce problem: 

1. Use Saleor 3.14 
2. create product type with reference values 
3. create product
4. assign 3 or more references 
5. save product
6. rearrange reference order and save 
7. refresh

Result: Products references will be in initial order instead of a new one
Expected result: Products references will be in a new order

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/attribute/models/base.py
+++ b/saleor/attribute/models/base.py
@@ -242,23 +242,27 @@ class AttributeValueManager(models.Manager):
         results = []
         objects_not_in_db: List[AttributeValue] = []
 
+        # prepare a list that will save order index of attribute values
+        objects_enumerated = [
+            (index, obj_data) for index, obj_data in enumerate(objects_data)
+        ]
         query = self._prepare_query_for_bulk_operation(objects_data)
 
         # iterate over all records in db and check if they match any of objects data
         for record in query.iterator():
             # iterate over all objects data and check if they match any of records in db
-            for obj in objects_data:
+            for index, obj in objects_enumerated:
                 if self._is_correct_record(record, obj):
                     # upon finding existing record add it to results
-                    results.append(record)
-
-                    # remove it from objects data, so it won't be added to new records
-                    objects_data.remove(obj)
-
-                    break
+                    results.append((index, record))
+                    # remove it from objects list, so it won't be added to new records
+                    objects_enumerated.remove((index, obj))
 
         # add what is left to the list of new records
-        self._add_new_records(objects_data, objects_not_in_db, results)
+        self._add_new_records(objects_enumerated, objects_not_in_db, results)
+        # sort results to keep order of attribute values
+        results.sort()
+        results = [obj for index, obj in results]
 
         if objects_not_in_db:
             self.bulk_create(objects_not_in_db)  # type: ignore[arg-type]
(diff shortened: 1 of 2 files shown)


saleor/attribute/models/base.py
```diff
@@ -242,23 +242,27 @@ def bulk_get_or_create(self, objects_data):
...
+        objects_enumerated = [
+            (index, obj_data) for index, obj_data in enumerate(objects_data)
+        ]
         query = self._prepare_query_for_bulk_operation(objects_data)
 
         # iterate over all records in db and check if they match any of objects data
         for record in query.iterator():
             # iterate over all objects data and check if they match any of records in db
-            for obj in objects_data:
+            for index, obj in objects_enumerated:
```
korycins, line 254: Instead of creating `objects_enumerated`, can't you just call here:
```python
for index, obj in enumerate(objects_data):
```
?

saleor/attribute/models/base.py
```diff
@@ -242,23 +242,27 @@ def bulk_get_or_create(self, objects_data):
...
                     # upon finding existing record add it to results
-                    results.append(record)
-
-                    # remove it from objects data, so it won't be added to new records
-                    objects_data.remove(obj)
-
-                    break
+                    results.append((index, record))
+                    # remove it from objects list, so it won't be added to new records
+                    objects_enumerated.remove((index, obj))
```
korycins, line 259: I don't think that removing the object from the list while iterating is correct approach.
```
In [17]: aa = ["a", "b", "c", "d", "e", "f"]

In [18]: enumerated_aa = list(enumerate(aa))

In [19]: print(enumerated_aa)
[(0, 'a'), (1, 'b'), (2, 'c'), (3, 'd'), (4, 'e'), (5, 'f')]

In [20]: for index, ob in enumerated_aa:
    ...:     print(index, ob)
    ...:     if index==1:
    ...:         print("removing ", ob)
    ...:         enumerated_aa.remove((index, ob))
    ...:
0 a
1 b
removing  b
3 d
4 e
5 f
```
See that the one iteration call is missing.

saleor/attribute/models/base.py
```diff
@@ -242,23 +242,27 @@ def bulk_get_or_create(self, objects_data):
         results = []
         objects_not_in_db: List[AttributeValue] = []
 
+        # prepare a list that will save order index of attribute values
+        objects_enumerated = [
+            (index, obj_data) for index, obj_data in enumerate(objects_data)
```
korycins, line 247: `list(enumerate(objects_data))` will do the same.

saleor/attribute/models/base.py
8r2y5: This case here is different, whole goal of removing record from `objects_enumerated` is to reduce calls in the next iterations when looking for correct record

saleor/attribute/models/base.py
```diff
@@ -242,23 +242,27 @@ def bulk_get_or_create(self, objects_data):
...
-            for obj in objects_data:
+            for index, obj in objects_enumerated:
                 if self._is_correct_record(record, obj):
                     # upon finding existing record add it to results
-                    results.append(record)
-
-                    # remove it from objects data, so it won't be added to new records
-                    objects_data.remove(obj)
-
-                    break
```
8r2y5, line 258: This line should be reverted. There won't be more matching record, so there is no point in iterating over rests of the records.

8r2y5 (approved):

saleor/attribute/models/base.py
jakubkuc: additionally `objects_enumerated` is passed later to `_add_new_records`

saleor/attribute/models/base.py
IKarbowiak: But according to what @korycins sent, there is a chance that you will skip the validation for the record, when the record before this record will be removed

8r2y5: At https://github.com/saleor/saleor/pull/15269/files#diff-14660fe9dd560e3979262f1c821c635bd3d367592a9f85822a1237235b606249R259 there is a `break` that will stop iteration and move to next db record and validation won't be skipped.

IKarbowiak (approved):

yellowee (approved):

Issue #12942, opened 2023-05-26
Possibility to block checkout/draft order when tax app doesn't return proper response.


korycins:
## Problem

Saleor can finalize the checkout/draft order process when Tax app doesn't provide the proper response during tax calculation.

---

## 1. General Assumptions

We will provide input for picking up the app that is responsible for processing the taxes for specific channel.
Based on the new setting and picked-up tax app we will be able to determine if we should allow to finalize the checkout process or not.
If the tax app will not be assigned to the tax configuration Saleor will work as before, that means will iterate over Apps to get tax data.

### Acceptance Criteria

- [ ] I can assign a tax app to specific taxConfiguration
- [ ] I can assign a tax app to taxConfigurationPerCountry
- [ ] Saleor doesn't allow to finalize the checkout/draft order when tax app doesn't return a correct response and taxAppId is set on taxConfiguration
- [ ] Saleor does allow to finalize the checkout/draft order when tax app is not assigned to tax configuration and settings for fail proces

#### Updated
```graphql
input TaxConfigurationUpdateInput @doc(category: "Taxes") {
  ...

  """
  The default strategy to use for tax calculation in the given channel. Taxes can be calculated either using user-defined flat rates or with a tax app. Empty value means that no method is selected and taxes are not calculated.
  """
  taxCalculationStrategy: TaxCalculationStrategy

  """
  !New fields below!
  """

  """
  The tax app id that will be used to calculate the taxes for the given channel.
  Empty value for `TAX_APP` set as `taxCalculationStrategy` means that Saleor will
  iterate over all installed tax apps.
  """
  taxAppId: String!
  
```

#### Updated
```graphql
type TaxConfiguration implements Node & ObjectWithMetadata @doc(category: "Taxes") {

  """
  The tax app id that will be used to calculate the taxes for the given channel.
  Empty value for `TAX_APP` set as `taxCalculationStrategy` means that Saleor will
  iterate over all installed tax apps.
  """
  taxAppId: String

}
```

```graphql
  type Shop {
    """
    List of tax apps that can be assigned to the channel. The list will be calculated by Saleor 
    based on the apps that are subscribed to webhooks related to tax calculations: CHECKOUT_CALCULATE_TAXES
    """
    availableTaxApps: [App!]!
  }
```

## 3. Database changes

```python
class TaxConfiguration(ModelWithMetadata):
    channel = models.OneToOneField(
        Channel, related_name="tax_configuration", on_delete=models.CASCADE
    )
[...]

PR #15292, opened 2024-01-25
Block when tax failed

kadewu:
https://github.com/saleor/saleor/issues/12942

Needs fixes from: https://github.com/saleor/saleor/pull/15378
Zero downtime will be provided with those fixes.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation: https://github.com/saleor/saleor-docs/pull/1083

# Pull Request Checklist


IKarbowiak (approved):

saleor/plugins/webhook/plugin.py
```diff
@@ -2995,30 +2995,89 @@ def process_payment(
...
+        subscriptable_object=None,
+    ):
+        app = (
+            App.objects.filter(
+                identifier=app_identifier,
+                is_active=True,
+            )
+            .order_by("-created_at")
+            .first()
+        )
```
tomaszszymanski129, line 3012: we could possibly prefetch webhooks since we call `app.webhooks.all()` below

tomaszszymanski129 (approved):

saleor/checkout/calculations.py
```diff
@@ -315,7 +359,13 @@ def _calculate_and_add_tax(
...
         _apply_tax_data_from_plugins(checkout, manager, checkout_info, lines, address)
         # Get the taxes calculated with apps and apply to checkout.
-        tax_data = manager.get_taxes_for_checkout(checkout_info, lines)
+        tax_data = manager.get_taxes_for_checkout(
+            checkout_info, lines, tax_app_identifier
+        )
+        # If taxAppId is not configured we will for now allow to finalize process for
+        # backward compatibility.
+        if tax_data is None and tax_app_identifier is not None:
+            raise ValueError("Empty tax data")
```
maarcingebala, line 368: Maybe it's not a problem for now, but I can imagine it could be easier to handle in code in more places, if we had a dedicated exception class for this.

saleor/graphql/tax/mutations/tax_configuration_update.py
```diff
@@ -149,13 +209,15 @@ def update_countries_configuration(cls, instance, countries_configuration):
             obj.charge_taxes = data["charge_taxes"]
             obj.display_gross_prices = data["display_gross_prices"]
             obj.tax_calculation_strategy = data.get("tax_calculation_strategy")
+            obj.tax_app_id = data.get("tax_app_id")
```
maarcingebala, line 212: Is it possible (should it be?) to clear the tax app and provide `null` as `taxAppId`?

kadewu: As for now I'm leaving an option to set `NULL`, just in case someone will want to turn off implemented behaviour.
Said field will become mandatory for Saleor 4.0.

saleor/app/migrations/0028_set_app_identifier.py
```diff
@@ -0,0 +1,45 @@
+from django.db import migrations
+from django.db.models import F, QuerySet
+
+from ..models import App
+
+BATCH_SIZE = 1000
```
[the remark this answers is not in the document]
kadewu: The batch is here just in case, as we expect that clients have ~10 apps.
The migration also changed, for that case I will leave the batch as for now.

saleor/checkout/calculations.py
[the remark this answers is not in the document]
kadewu: Added separate new `Error` as per other sugestion.

saleor/plugins/webhook/plugin.py
kadewu: I will stick with no prefetch, as I believe it's more clear in code and consistent with other functions in file.

saleor/checkout/calculations.py
kadewu: Added new Error

saleor/checkout/calculations.py
```diff
@@ -315,7 +359,13 @@ def _calculate_and_add_tax(
         # Call the tax plugins.
         _apply_tax_data_from_plugins(checkout, manager, checkout_info, lines, address)
         # Get the taxes calculated with apps and apply to checkout.
-        tax_data = manager.get_taxes_for_checkout(checkout_info, lines)
+        tax_data = manager.get_taxes_for_checkout(
+            checkout_info, lines, tax_app_identifier
+        )
+        # If taxAppId is not configured we will for now allow to finalize process for
+        # backward compatibility.
+        if tax_data is None and tax_app_identifier is not None:
```
korycins, line 367: Shouldn't we validate the recieved payload from the app? Currently this will pass if received tax_data will be empty list.

saleor/checkout/calculations.py
```diff
@@ -258,37 +288,50 @@ def _fetch_checkout_prices_if_expired(
...
+        "shipping_tax_rate",
+        "translated_discount_name",
+        "discount_amount",
+        "discount_name",
+        "currency",
+        "last_change",
+    ]
+
+    if tax_app_responed:
+        checkout.price_expiration = timezone.now() + settings.CHECKOUT_PRICES_TTL
```
korycins, line 330: Are you sure of this? 
In case the app is down, this will cause the re-calculation of the prices for each query and mutation. It means that workers will wait for the response from the app. You already added the logic to raise an exception when taxAppID is provided, so why not leave this as it was?

saleor/graphql/app/mutations/app_create.py
```diff
@@ -17,6 +17,12 @@
 
 class AppInput(BaseInputObjectType):
     name = graphene.String(description="Name of the app.")
+    identifier = graphene.String(
+        description=(
+            "Canonical app ID. If not provided, "
+            "the identifier will be generated based on app.id."
```
korycins, line 23: ADDED_IN_X

saleor/graphql/tax/mutations/tax_configuration_update.py
```diff
@@ -133,6 +153,46 @@ def clean_input(cls, info: ResolveInfo, instance, data, **kwargs):
...
+    @classmethod
+    def clean_tax_app_id(cls, info: ResolveInfo, data):
+        active_tax_apps = list(get_active_tax_apps())
+        if app_identifier := data.get("tax_app_id"):
+            cls.__verify_tax_app_id(info, app_identifier, active_tax_apps)
+
+        update_countries_configuration = data.get("update_countries_configuration", [])
+        for country in update_countries_configuration:
+            if app_identifier := country.get("tax_app_id"):
+                cls.__verify_tax_app_id(
```
korycins, line 165: `__verify_tax_app_id` makes:
```        app = (
            app_models.App.objects.filter(
                identifier=app_identifier,
                is_active=True,
            )
            .order_by("-created_at")
            .first()
        )
```
And it is called inside loop

saleor/plugins/webhook/plugin.py
```diff
@@ -2995,30 +2995,89 @@ def process_payment(
...
-            checkout_info.checkout,
+    def __run_tax_webhook(
+        self,
+        event_type: str,
+        app_identifier: str,
+        payload: str,
+        subscriptable_object=None,
+    ):
+        app = (
+            App.objects.filter(
```
korycins, line 3006: Can't you provide `app_identifier` to `get_webhooks_for_event` to get the same result?

saleor/plugins/webhook/plugin.py
```diff
@@ -2995,30 +2995,89 @@ def process_payment(
...
+                identifier=app_identifier,
+                is_active=True,
+            )
+            .order_by("-created_at")
+            .first()
+        )
+        if app is None:
+            return None
```
korycins, line 3014: Any logger could be usefull in case of debbuging

saleor/plugins/webhook/plugin.py
```diff
@@ -2995,30 +2995,89 @@ def process_payment(
...
-        return trigger_all_webhooks_sync(
-            WebhookEventSyncType.ORDER_CALCULATE_TAXES,
-            lambda: generate_order_payload_for_tax_calculation(order),
-            parse_tax_data,
-            order,
-            self.requestor,
-        )
+        event_type = WebhookEventSyncType.ORDER_CALCULATE_TAXES
+        if app_identifier:
+            payload = generate_order_payload_for_tax_calculation(order)
```
korycins, line 3066: This means that we will always generate the static payload. No matter if we need it or not.

saleor/plugins/webhook/plugin.py
```diff
@@ -2995,30 +2995,89 @@ def process_payment(
...
+            request=request_context,
         )
+        return parse_tax_data(response)
+
+    def get_taxes_for_checkout(
+        self, checkout_info, lines, app_identifier, previous_value
+    ) -> Optional["TaxData"]:
+        event_type = WebhookEventSyncType.CHECKOUT_CALCULATE_TAXES
+        if app_identifier:
+            payload = generate_checkout_payload_for_tax_calculation(
```
korycins, line 3040: This means that we will always generate the static payload. No matter if we need it or not.

saleor/checkout/calculations.py
korycins: I think that this will cause to explode Saleor 😄 
If you have a query which fetches a prices, the first thing which is validating is the expiration. If your query has more than single price field to fetch, each resolver will try to call the tax app.

korycins (approved):

saleor/plugins/webhook/plugin.py
kadewu: Actually changed in fixes: switched to app id

PR #15274, opened 2024-01-23
Gift promotions API.

zedzior:
Issue: https://github.com/saleor/saleor/issues/15224
This PR adds API for gift promotions with necessary validation.

Following requirements from the https://github.com/saleor/saleor/issues/15224 will be updated in separate PR:
- Update `CheckoutLinesUpdate` - don’t allow to update quantity of lines marked as gift line
- Update `OrderLineUpdate` - don’t allow to update quantity of lines marked as gift line
- Update `checkoutLineDelete` and `CheckoutLinesDelete` - do not allow deleting gift lines

### Mutations

- Update the promotion inputs with `gifts` field
- Update the following mutations: `promotionCreate`, `PromotionRuleCreate`, `promotionRuleUpdate`:
  - Require the `gifts` field when the `GIFT` reward type is chosen
  - `rewardValueType`, `rewardValue`, `rewardPredicate` should be empty - raise an `ValidationError` otherwise
  - Validate `gifts` field - the provided IDs must be the `ProductVariant` ids

```graphql
input PromotionRuleInput @doc(category: "Discounts") {
  ...
  gifts: [ID!]
}

input PromotionRuleCreateInput @doc(category: "Discounts") {
  ...
  gifts: [ID!]
}

input PromotionRuleUpdateInput @doc(category: "Discounts") {
  ...
   gifts: [ID!]
}
```

### Types

- Extend `PromotionRule` with `giftIds`
- Extend `PromotionRule` with`giftsLimit` - will define the maximum number of gifts to choose from the gifts list. For now, we will always return 1, and it won't be configurable for now
- Extend `RewardTypeEnum` with GIFT 
- Extend `OrderLine` and `CheckoutLine` with `isGift` flag
- Consider adding filter for getting lines that are gifts

```graphql


type PromotionRule implements Node @doc(category: "Discounts"){
  ...
  giftIds: [ID!]
  giftsLimit: Int
}

enum RewardTypeEnum @doc(category: "Discounts") {
  ...
  GIFT
}

type OrderLine implements Node & ObjectWithMetadata @doc(category: "Orders") {
  ...
  isGift: Boolean!
}

type CheckoutLine implements Node & ObjectWithMetadata @doc(category: "Checkout") {
  ...
  isGift: Boolean!
}

```

# Impact

- [ ] New migrations
- [ ] New/Updated API fields or mutations
- [ ] Deprecated API fields or mutations
- [ ] Removed API types, fields, or mutations

# Docs

- [ ] Link to documentation:

[...]

Issue #15224, opened 2024-01-11
Adjust API for gift promotion rule


IKarbowiak:
Part of https://github.com/saleor/saleor/issues/15075

### Mutations

- Update the promotion inputs with `gifts` field
- Update the following mutations: `promotionCreate`, `PromotionRuleCreate`, `promotionRuleUpdate`:
  - Require the `gifts` field when the `GIFT` reward type is chosen
  - `rewardValueType`, `rewardValue`, `rewardPredicate` should be empty - raise an `ValidationError` otherwise
  - Validate `gifts` field - the provided IDs must be the `ProductVariant` ids
- Update `CheckoutLinesUpdate` - don't allow to update quantity of lines marked as gift line
- Update `OrderLineUpdate` - don't allow to update quantity of lines marked as gift line
- Update `checkoutLineDelete` and `CheckoutLinesDelete` - do not allow deleting gift lines

```graphql
input PromotionRuleInput @doc(category: "Discounts") {
  ...
  gifts: [ID!]
}

input PromotionRuleCreateInput @doc(category: "Discounts") {
  ...
  gifts: [ID!]
}

input PromotionRuleUpdateInput @doc(category: "Discounts") {
  ...
   gifts: [ID!]
}
```

### Types

- Extend `PromotionRule` with `giftIds`
- Extend `PromotionRule` with`giftsLimit` - will define the maximum number of gifts to choose from the gifts list. For now, we will always return 1, and it won't be configurable for now
- Extend `RewardTypeEnum` with GIFT 
- Extend `OrderLine` and `CheckoutLine` with `isGift` flag
- Consider adding filter for getting lines that are gifts

```graphql


type PromotionRule implements Node @doc(category: "Discounts"){
  ...
  giftIds: [ID!]
  giftsLimit: Int
}

enum RewardTypeEnum @doc(category: "Discounts") {
  ...
  GIFT
}

type OrderLine implements Node & ObjectWithMetadata @doc(category: "Orders") {
  ...
  isGift: Boolean!
}

type CheckoutLine implements Node & ObjectWithMetadata @doc(category: "Checkout") {
  ...
  isGift: Boolean!
}

```

PR #15326, opened 2024-01-26
Validate gift lines.

zedzior:
Part of: https://github.com/saleor/saleor/issues/15224
Follow up to: https://github.com/saleor/saleor/pull/15274
This PR adds following validations:
- `checkoutLinesUpdate` - don't allow to update quantity of lines marked as gift line
- `orderLineUpdate` - don't allow to update quantity of lines marked as gift line
- `checkoutLineDelete` and `checkoutLinesDelete` - do not allow deleting gift lines
- `orderLineDelete`  - do not allow deleting gift lines

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/graphql/checkout/mutations/checkout_lines_update.py
```diff
@@ -221,3 +224,30 @@ def _get_variants_from_lines_input(cls, lines: list[dict]) -> list[ProductVarian
     @classmethod
     def _get_grouped_lines_data(cls, lines, existing_lines_info):
         return group_lines_input_data_on_update(lines, existing_lines_info)
+
+    @classmethod
+    def _validate_gift_line(
```
IKarbowiak, line 229: Should we allow updating the gift line? I mean, maybe we shouldn't check the quantity and return a validation error when somebody tries to update the gift line?

IKarbowiak (commented):
Just one thing to consider, otherwise looks good!

saleor/graphql/checkout/mutations/checkout_lines_delete.py
```diff
@@ -47,16 +48,20 @@ class Meta:
...
         lines = checkout.lines.all()
         all_lines_ids = [str(line.id) for line in lines]
-        invalid_line_ids = list()
-        for line_to_delete in lines_to_delete:
-            if line_to_delete not in all_lines_ids:
-                line_to_delete = graphene.Node.to_global_id(
-                    "CheckoutLine", line_to_delete
+        gift_line_ids = [str(line.id) for line in lines if line.is_gift]
+        invalid_line_ids, gift_line_to_delete_ids = list(), list()
+        for line_to_delete_id in lines_to_delete_ids:
```
korycins, line 56: You could use `set` instead of loop. 
```python
invalid_line_ids = set(lines_to_delete_ids).difference(all_lines_ids)
gift_line_to_delete_ids = set(gift_line_ids).intersection(lines_to_delete_ids)
```

saleor/graphql/checkout/mutations/checkout_lines_update.py
```diff
@@ -221,3 +224,30 @@ def _get_variants_from_lines_input(cls, lines: list[dict]) -> list[ProductVarian
...
+        cls, lines_data: list[CheckoutLineData], lines_info: list[CheckoutLineInfo]
+    ):
+        gift_line_ids = [
+            str(line_info.line.id) for line_info in lines_info if line_info.line.is_gift
+        ]
+        gift_lines_with_quantity_to_update = [
+            line
+            for line in lines_data
+            if line.line_id in gift_line_ids and line.quantity_to_update
```
korycins, line 238: In line 233 you, are making `str(line_info.line.id)`, here you're checking `line.line_id in gift_line_ids`. This could mean, incorrect types, right? Like id in uuid vs str(uuid)

saleor/order/error_codes.py
```diff
@@ -36,6 +36,8 @@ class OrderErrorCode(Enum):
     CHANNEL_INACTIVE = "channel_inactive"
     INVALID_VOUCHER = "invalid_voucher"
     INVALID_VOUCHER_CODE = "invalid_voucher_code"
+    NON_EDITABLE_GIFT_LINE_QUANTITY = "non_editable_gift_line_quantity"
```
korycins, line 39: Wouldn't `NON_EDITABLE` and `NON_REMOVABLE` be better?
If you place `GIFT_LINE` here, this error will be related only to giftcards (it is ok for now, but if we extend saleor in the future, with functionality, where you can't edit the line for some reason (like bundle products etc), this error will not work).

---

Also don't think that `_QUNATITY` is useful here. Like the error that storefront receives will be:
```python
{
                     "quantity": ValidationError(
                         "Quantity of order line marked as gift can't be edited.",
                         code=CheckoutErrorCode.NON_EDITABLE_GIFT_LINE_QUANTITY.value,
                         params={"lines": global_ids},
                     )
                 }
```
You already have the error assigned to field `quantity`, you also have `quantity` in error message. 
The most important info for the storefront here is that the line is non editable/ non removable.

saleor/graphql/checkout/mutations/checkout_lines_delete.py
zedzior: Nice, thanks

saleor/graphql/checkout/mutations/checkout_lines_update.py
zedzior: `CheckoutLineInfo.line.id` is type of `UUID`, while `CheckoutLineData.line_id` is type of `str`

saleor/order/error_codes.py
zedzior: I deleted `_QUANTITY`, but kept `GIFT_LINE` part as we don't have `gift_line` field. So to determine the reason the line is not editable, would require additional check, like `line.is_gift=True`

korycins (approved):

IKarbowiak (approved):

Issue #14880, opened 2023-11-29
Bug: Order level discounts are not propagated to its lines, if tax calculations is based on plugins.


zedzior:
### What are you trying to achieve?

If tax startegy == `TAX_APP` and `manager.get_taxes_for_order` doesn't return any data, order level discount (entire order vouchers and manual discounts) are not reflected in order lines totals.

### Steps to reproduce the problem

Run `test_draft_order_complete_with_voucher` and check order line total prices do not sum up to order total.

### What did you expect to happen?

I expect to see total line prices decreased by order level discounts.

### Environment

Saleor version: …
OS and version: …

PR #15344, opened 2024-02-02
Propagate order discount when using tax plugins

korycins:
I want to merge this change because it should cover https://github.com/saleor/saleor/issues/14880

> [!IMPORTANT]  
> Scope of changes:
> - The avatax cassettes have been re-generated
> - I moved the discount propagation (`apply_order_discounts`) to `_recalculate_prices` - previously it was called for each method separately
> - I fixed the `default_value`s that are passed to plugins. They include the order discount. If not active any tax plugin, the default_value will be returned.
> - I split the logic from `apply_order_discounts` to separate methods. We will be able to only fetch discounted subtotal, without calling "assign" logic, etc. 
 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/plugins/manager.py
```diff
@@ -425,8 +426,21 @@ def calculate_order_line_total(
         order_line: "OrderLine",
         variant: "ProductVariant",
         product: "Product",
+        lines: Iterable["OrderLine"],
     ) -> OrderTaxedPricesData:
-        default_value = base_order_calculations.base_order_line_total(order_line)
+        base_subtotal = base_order_subtotal(order, lines)
+        subtotal, shipping_price = propagate_order_discount_on_order_prices(
+            order, lines
+        )
```
zedzior, line 434: With this implementation we fire `propagate_order_discount_on_order_prices` function for every single order line. The function call db, so it is not an optimal solution. Maybe we can pass `subtotal` and `shipping price` in args?

saleor/plugins/manager.py
```diff
@@ -485,17 +499,26 @@ def calculate_order_line_unit(
         order_line: "OrderLine",
         variant: "ProductVariant",
         product: "Product",
+        lines: Iterable["OrderLine"],
     ) -> OrderTaxedPricesData:
+        base_subtotal = base_order_subtotal(order, lines)
+        subtotal, shipping_price = propagate_order_discount_on_order_prices(
+            order, lines
+        )
```
zedzior, line 507: Same here

saleor/plugins/manager.py
korycins: It is okay. All db data is prefetched. I assumed, that calculating the values inside these method will allow to be "independent" method. Just to confirm I've made additional test when you're calling fetch_prices logic:
```
In [94]: def a():
    ...:     o = Order.objects.filter(currency="USD")[4]
    ...:     fetch_order_prices_if_expired(o, manager, o.lines.all())
    ...:

In [95]: a()
Without propagation
Without propagation
Without propagation
Number of queries 17

In [96]: a()
With propagation
With propagation
With propagation
Number of queries 17
```
Where "Without" is the logic with commented part of calculation inside the manager. 
"With" is the logic from this PR. 
Both results end up with 17db queries. Tested on order with 3 lines.

zedzior (approved):

IKarbowiak (approved):

yellowee (approved):

PR #15349, opened 2024-02-02
Limit update webhook triggers

maarcingebala:
This PR adds a comparison of pre-save and post-save webhook payloads in mutations, in order to limit sending webhooks where the payload has not changed as a result of the mutation.

The implementation assumes reusing data loaders and their cache for pre-save payload generation, in order to save database queries if the same fields are present in different webhook subscriptions (in case there are multiple webhooks for the event). 

- For now, the only affected mutation is `productVariantBulkUpdate`.
- The feature needs to be enabled with `ENABLE_LIMITING_WEBHOOKS_FOR_IDENTICAL_PAYLOADS` env variable.

https://linear.app/saleor/issue/PE-97/ir-150-limit-outgoing-webhooks-to-notify-listeners-only-when-data

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -64,24 +73,46 @@ def create_deliveries_for_subscriptions(
...
             continue
+
+        if (
+            settings.ENABLE_LIMITING_WEBHOOKS_FOR_IDENTICAL_PAYLOADS
+            and pre_save_payloads
+        ):
+            key = get_pre_save_payload_key(webhook, subscribable_object)
+            pre_save_payload = pre_save_payloads.get(key)
+            if pre_save_payload and pre_save_payload == data:
+                logger.info("Skipping webhook: %s - no data changes.", event_type)
```
przlada, line 113: It might be beneficial to include additional details in these logs, such as `event_type` and webhook's `target_url` (which can be helpful for debugging). Furthermore, we can consider aggregating all skipped webhooks into a single log message (with `extra` data).

przlada (approved):

saleor/webhook/transport/asynchronous/transport.py
maarcingebala: Good idea about including target URL 👍

przlada (approved):

kadewu (approved):

PR #15369, opened 2024-02-07
Remove `prefetched_for_webhook` to legacy payload generators

### LINKED ISSUE #15188: Move prefetched_for_webhook to payload serializers
Many product-related mutations use the method `prefeteched_for_webhook` to run prefetches needed by legacy webhook payload generators. This is inefficient since these prefetches are heavy and are unnecessarily run for webhook using subscription paylaods. To fix this, we should move the `prefetched_for_webhook` call to legacy payload generators.

Example: 
- [prefetched_for_webhook is called](https://github.com/saleor/saleor/blob/1df99ed89df3d4ee8db00e2280dcb98995e5551b/saleor/graphql/product/mutations/product/product_create.py#L222) in the `post_save_action` in the `productCreate` mutation
- Instead, it should be moved to the [legacy payload generator](https://github.com/saleor/saleor/blob/1df99ed89df3d4ee8db00e2280dcb98995e5551b/saleor/webhook/payloads.py#L713) where it's actually needed

To do:
- Analyze all usages of `prefeteched_for_webhook` (both for product and for variant) and move them to proper payload generators.
- Remove prefetch of [product_type__product_attributes__values](https://github.com/saleor/saleor/blob/1df99ed89df3d4ee8db00e2280dcb98995e5551b/saleor/product/managers.py#L230) which is not used by the product payload generator.
- Update tests that check the number of DB queries; after these changes, the DB query count should be lower when subscription webhooks are enabled (note: in tests, make sure to use products with some attributes and values assigned).
- Port changes to versions from 3.14 to latest.

Context: https://saleorcommerce.slack.com/archives/C06C6UWV45V/p1704470719634879

AjmalPonneth:
I want to merge this change because it fixes #15188

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


maarcingebala (commented):
Just one note from my side, but other than this it looks good.

AjmalPonneth:
@maarcingebala Updated the PR.

AjmalPonneth:
@maarcingebala Updated the PR. Please check and advise.

saleor/webhook/payloads.py
```diff
@@ -839,6 +835,9 @@ def generate_product_variant_payload(
     requestor: Optional["RequestorOrLazyObject"] = None,
     with_meta: bool = True,
 ):
+    product_variants = ProductVariant.objects.prefetched_for_webhook().filter(
```
korycins, line 838: Keep in mind, that this function is called inside `generate_product_payload`. It means that you will make a prefetch one by one for each variant assigned to the product. It won't give us anything. 
Instead of calling this always, we should have a helper method that will confirm if the prefetched is applied or not. This should solve the problem of calling this function at different stage of creating payload.

Something like this:
```python
# you can apply different name, I just pick up the first one
def should_apply_prefetch(object, requested_prefetches: list[str]):
    if prefetches:=getattr(product_variant, "_prefetched_objects_cache",None):
         # requested_prefetches should exists in prefeteches to return False
    return True
```
`_prefetched_objects_cache` will give you something dict in below form:
```
{'attributes': <QuerySet []>,
 'variant_media': <QuerySet [<VariantMedia: VariantMedia object (58)>]>,
 'stocks': <StockQuerySet [<Stock: Stock object (457)>, <Stock: Stock object (458)>, <Stock: Stock object (459)>, <Stock: Stock object (460)>, <Stock: Stock object (461)>, <Stock: Stock object (462)>, <Stock: Stock object (463)>, <Stock: Stock object (464)>]>,
 'channel_listings': <ProductVariantChannelListingQuerySet [<ProductVariantChannelListing: ProductVariantChannelListing object (342)>, <ProductVariantChannelListing: ProductVariantChannelListing object (343)>]>}
 ```

**Review by korycins — COMMENTED** [review]
The issue #15188 has also a task to drop some prefetched objects:
`Remove prefetch of product_type__product_attributes__values which is not used by the product payload generator.`
Is it something that you're going to handle in this PR?

**Comment by AjmalPonneth** [comment]
> The issue #15188 has also a task to drop some prefetched objects: `Remove prefetch of product_type__product_attributes__values which is not used by the product payload generator.` Is it something that you're going to handle in this PR?

Removed `product_type__product_attributes__values`

File: CHANGELOG.md [hunk 2]
```diff
@@ -30,6 +30,8 @@ All notable, unreleased changes to this project will be documented in this file.
 - Added caching of GraphQL documents for common queries to improve performance - #14843 by @patrys
 - Added `VOUCHER_CODES_CREATED` and `VOUCHER_CODES_DELETED` webhooks events. - #14652 by @SzymJ
 - Fixed validation for streetAddress1 or streetAddress2 are too long - #13973 by sonbui00
+- Remove `prefetched_for_webhook` to legacy payload generators - #15369 by @AjmalPonneth
```
maarcingebala, line 33: Since we released 3.19 this entry should be moved to 3.20 now (you need to rebase with main).

maarcingebala (approved):

saleor/webhook/payloads.py
korycins: Ahh, my bad. We can ignore it. I thought that the function accepts a single variant.

korycins (approved):

AjmalPonneth:
@maarcingebala Fixed mypy issues. Please check and advise.

IKarbowiak (approved):

saleor/graphql/product/mutations/product/product_update.py
```diff
@@ -48,10 +48,9 @@ def clean_attributes(
         return attributes
 
     @classmethod
-    def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
```
kadewu, line 51: Just small note: In productCreate mutation you did leave `instance` in function definition, I think we should be consistent in naming.

kadewu (approved):

maarcingebala:
One test is failing: `test_generate_product_variant_deleted_payload`.

zedzior:
Hey @AjmalPonneth.  Since we haven't heard from you for a while and the issue can't wait no longer, I decided to fix conflicts and merge the PR on my own. Thanks again for a great work.

PR #15373, opened 2024-02-09
Fix setting invalid identifier

kadewu:
Fix to the invalid set `App.identifier` in `AppCreate` mutation + missing setting `App.identifier` in `create_app` command. 
Command also contains optional new argument, I decided it should not be a problem to port this logic, please raise concerns if any.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/app/management/commands/create_app.py
+++ b/saleor/app/management/commands/create_app.py
@@ -25,7 +25,10 @@ class Command(BaseCommand):
             "--identifier",
             dest="identifier",
             default="",
-            help="Canonical app ID.",
+            help=(
+                "Canonical app ID. If not provided, "
+                "the identifier will be generated based on app.id."
+            ),
         )
         parser.add_argument(
             "--permission",
--- a/saleor/app/management/commands/create_app.py
+++ b/saleor/app/management/commands/create_app.py
@@ -1,6 +1,7 @@
 import json
 from typing import Any, Dict, Optional
 
+import graphene
 import requests
 from django.conf import settings
 from django.contrib.sites.models import Site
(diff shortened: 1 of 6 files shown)


saleor/app/management/commands/create_app.py
```diff
@@ -20,6 +21,12 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser: CommandParser) -> None:
         parser.add_argument("name", type=str)
+        parser.add_argument(
+            "--identifier",
+            dest="identifier",
+            default="",
+            help="Canonical app ID.",
```
korycins, line 28: I would add here: "If not provided, the identifier will be generated based on app.id"

korycins (approved):

saleor/graphql/app/mutations/app_create.py
```diff
@@ -76,7 +76,9 @@ def perform_mutation(cls, _root, info, /, **data):
 
     @classmethod
     def save(cls, info, instance, cleaned_input):
-        instance.identifier = graphene.Node.to_global_id("App", instance.pk)
         instance.save()
+        if not instance.identifier:
+            instance.identifier = graphene.Node.to_global_id("App", instance.pk)
```
maarcingebala, line 81: Where does the identifier come in this mutation, if not generated from ID? It's not available in the input.

kadewu: Good point. 
Context: it is added to the input in related PR (but only to the newest version)

I will remove that unnecessary if.

maarcingebala (approved):

PR #15406, opened 2024-02-14
Add disclaimer to not fork

timuric:
Many community members are doing mistake of forking the repository. Added disclaimers to attempt to raise awareness

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1,4 +1,8 @@
 #!/usr/bin/env python3
+
+# DO NOT FORK SALEOR TO EXTEND IT
+# Learn more https://docs.saleor.io/docs/3.x/developer/extending/overview#why-not-to-fork
+
 import os
 import sys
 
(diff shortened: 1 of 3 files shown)


korycins (commented):
Why do you apply this in .py files instead of the README?  I mean, in most cases someone will see it too late, or never. (I've never opened Saleor's manage.py or __main__ file 😄 ), you could also quickly skip the base_plugin in case if you relay on implementation of different plugin. So for me this note should be in readme, marked as important. If someone skip it, then I don't think that notes in .py files would help here.

We could add label like this one to the readme:
image

```
> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.
```

maarcingebala:
Closing for now since there was no agreement where this information should be put.

PR #15426, opened 2024-02-20
Fix `FLAT_RATE` tax calculation for Click and Collect option.

Air-t:
I want to merge this change because it fixes the calculation of a flat-rate tax for the click-and-collect option.

resolves: https://linear.app/saleor/issue/SHOPX-175/bug-billing-address-is-used-for-tax-calculation-for-clickandcollect

There is also a method called `get_tax_country` but it does not take into account the click-and-collect option. Maybe extending it and using it instead of adding a new one is a good idea.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
+++ b/saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
@@ -21,6 +21,7 @@ from ....plugins.webhook.utils import APP_ID_PREFIX
 from ....shipping import interface as shipping_interface
 from ....shipping import models as shipping_models
 from ....shipping.utils import convert_to_shipping_method_data
+from ....warehouse import WarehouseClickAndCollectOption
 from ....warehouse import models as warehouse_models
 from ...core import ResolveInfo
 from ...core.descriptions import ADDED_IN_31, ADDED_IN_34, DEPRECATED_IN_3X_INPUT
@@ -57,7 +58,9 @@ class CheckoutDeliveryMethodUpdate(BaseMutation):
     class Meta:
         description = (
             "Updates the delivery method (shipping method or pick up point) "
-            "of the checkout." + ADDED_IN_31
+            "of the checkout. "
+            "Updates the checkout shipping_address for click and collect delivery "
+            "for a warehouse address. " + ADDED_IN_31
         )
         doc_category = DOC_CATEGORY_CHECKOUT
         error_type_class = CheckoutError
@@ -223,6 +226,7 @@ class CheckoutDeliveryMethodUpdate(BaseMutation):
         external_shipping_method: Optional[shipping_interface.ShippingMethodData],
         collection_point: Optional[Warehouse],
     ) -> None:
+        checkout_fields_to_update = ["shipping_method", "collection_point"]
         checkout = checkout_info.checkout
         if external_shipping_method:
             set_external_shipping_id(
@@ -232,6 +236,14 @@ class CheckoutDeliveryMethodUpdate(BaseMutation):
             delete_external_shipping_id(checkout=checkout)
         checkout.shipping_method = shipping_method
         checkout.collection_point = collection_point
+        if (
+            collection_point is not None
+            and collection_point.click_and_collect_option
+            == WarehouseClickAndCollectOption.LOCAL_STOCK
+        ):
+            checkout.shipping_address = collection_point.address
+            checkout_info.shipping_address = collection_point.address
+            checkout_fields_to_update += ["shipping_address"]
         invalidate_prices_updated_fields = invalidate_checkout_prices(
             checkout_info, lines, manager, save=False
         )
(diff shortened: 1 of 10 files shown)


saleor/tax/utils.py
```diff
@@ -14,12 +14,26 @@
     from .models import TaxConfiguration, TaxConfigurationPerCountry
 
 
+def _get_country_code_for_checkout_for_tax_calculation(
```
tomaszszymanski129, line 17:
```suggestion
def get_country_code_for_checkout_tax_calculation(
```

tomaszszymanski129 (approved):

zedzior (approved):

saleor/tax/utils.py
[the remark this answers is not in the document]
Air-t: thanks, I missed that part where you proposed the solution

saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
```diff
@@ -57,7 +58,8 @@ class Arguments:
     class Meta:
         description = (
             "Updates the delivery method (shipping method or pick up point) "
-            "of the checkout." + ADDED_IN_31
+            "of the checkout. "
+            "Updates the checkout shipping_address if it was missing. " + ADDED_IN_31
```
maarcingebala, line 62: When the Click & Collect delivery is selected, it sets the warehouse address (pickup point) as the shipping address.

In the code below it sets the shipping_address to pickup point address always, not only when it's missing?

saleor/graphql/checkout/mutations/utils.py
```diff
@@ -45,6 +45,9 @@
 
 
 ERROR_DOES_NOT_SHIP = "This checkout doesn't need shipping"
+ERROR_CC_ADDRESS_CHANGE_FORBIDDEN = (
+    "Can't change shipping address for click and collect"
```
maarcingebala, line 49: Maybe we should inform here why it's forbidden? E.g., For Click&Collect delivery, the address is automatically set to the warehouse address and cannot be changed manually.

saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
Air-t: yes, I forgot to update the description to inform that it always updates the address for click-and-collect delivery for the warehouse address. Do you think it should just check a case where `shipping_address` is empty and only then update this field?

maarcingebala: Each collection point most likely will have a different address, so we need to update the address whenever a delivery method (collection point) is changed. Could be verified with product owner.

korycins (approved):

saleor/graphql/checkout/tests/mutations/test_checkout_delivery_method_update.py
```diff
@@ -485,6 +485,8 @@ def test_checkout_delivery_method_update_with_not_all_required_shipping_address_
             delivery_method,
             delivery_method.channel_listings.get(),
         )
+    elif attribute_name == "collection_point" and is_valid_delivery_method:
+        checkout_info.shipping_address = shipping_method_data.address
```
IKarbowiak, line 489: IMO this test should be split into two

saleor/graphql/checkout/tests/mutations/test_checkout_delivery_method_update.py
```diff
@@ -560,6 +562,8 @@ def test_checkout_delivery_method_update_with_not_valid_address_data(
...
```
IKarbowiak, line 566: Same here

IKarbowiak (approved):

saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
lkostrowski: I'm not sure if this isn't the feature we must still add? I agree that it makes sense that shipping address should be used for calculation (and collection point = shipping address here). 

On the other hand, some of the other places also use this warehouse logic (eg taxes plugin). Maybe first we can merge this, to unify the behavior (warehouse instead billing), then we can revisit this globally (plugins, apps, flat rates) and apply collection points on top of that?

cc @szczecha you were working on that

lkostrowski: >Each collection point most likely will have a different address, so we need to update the address whenever a delivery method (collection point) is changed. Could be verified with product owner.

@Air-t @maarcingebala to clarify - current solution will not update address when I switch collection point 1 to collection point 2?

So I can
1. Chose a collection point in state with lower tax
2. Change collection point to state with higher tax (where I live in)
3. Order with lower tax, because address was not recalculated?

saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
```diff
@@ -57,7 +58,9 @@ class Arguments:
     class Meta:
         description = (
             "Updates the delivery method (shipping method or pick up point) "
-            "of the checkout." + ADDED_IN_31
+            "of the checkout. "
+            "Updates the checkout shipping_address for click and collect delivery "
+            "for a warehouse address. " + ADDED_IN_31
```
maarcingebala, line 63:
```suggestion
            "Updates the checkout shipping address for click and collect delivery "
            "to a warehouse address. " + ADDED_IN_31
```

PR #15433, opened 2024-02-21
Add plugin to tax configuration

kadewu:
I want to merge this change because it adds possibility to set `plugin` directly on `taxConfiguration.taxAppId`.
Also contains:
- do not call Avalara using plugin if taxAppId set as Tax App
- reduce number of calls to Tax App if Tax App is not responding (in previous implementation, after failing checkoutComple/draftOrderComplete we will do potential call to Tax App when e.g. querying related object)

Changes will be ported to main branch.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation: https://github.com/saleor/saleor-docs/pull/1097

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/calculations.py
+++ b/saleor/checkout/calculations.py
@@ -367,13 +367,21 @@ def _call_plugin_or_tax_app(
     address: Optional["Address"] = None,
 ):
     if tax_app_identifier.startswith(PLUGIN_IDENTIFIER_PREFIX):
+        plugin_ids = [tax_app_identifier.replace(PLUGIN_IDENTIFIER_PREFIX, "")]
+        plugins = manager.get_plugins(
+            checkout_info.channel.slug,
+            active_only=True,
+            plugin_ids=plugin_ids,
+        )
+        if not plugins:
+            raise TaxEmptyData("Empty tax data.")
         _apply_tax_data_from_plugins(
             checkout,
             manager,
             checkout_info,
             lines,
             address,
-            plugin_ids=[tax_app_identifier.replace(PLUGIN_IDENTIFIER_PREFIX, "")],
+            plugin_ids=plugin_ids,
         )
         if checkout.tax_error:
             raise TaxEmptyData("Empty tax data.")
--- a/saleor/checkout/calculations.py
+++ b/saleor/checkout/calculations.py
@@ -3,7 +3,6 @@ from decimal import Decimal
 from typing import TYPE_CHECKING, Optional, cast
 
 from django.conf import settings
-from django.core.exceptions import ValidationError
 from django.utils import timezone
 from prices import Money, TaxedMoney
 
@@ -22,7 +21,6 @@ from ..tax.utils import (
     get_tax_calculation_strategy_for_checkout,
     normalize_tax_rate_for_db,
 )
-from .error_codes import CheckoutErrorCode
 from .fetch import find_checkout_line_info
 from .models import Checkout
 from .payment_utils import update_checkout_payment_statuses
@@ -39,7 +37,6 @@ def checkout_shipping_price(
     checkout_info: "CheckoutInfo",
     lines: Iterable["CheckoutLineInfo"],
     address: Optional["Address"],
-    need_tax_calculation: bool = False,
 ) -> "TaxedMoney":
     """Return checkout shipping price.
 
@@ -51,7 +48,6 @@ def checkout_shipping_price(
         manager=manager,
         lines=lines,
         address=address,
-        need_tax_calculation=need_tax_calculation,
     )
     return quantize_price(checkout_info.checkout.shipping_price, currency)
 
@@ -62,7 +58,6 @@ def checkout_shipping_tax_rate(
     checkout_info: "CheckoutInfo",
     lines: Iterable["CheckoutLineInfo"],
     address: Optional["Address"],
-    need_tax_calculation: bool = False,
 ) -> Decimal:
     """Return checkout shipping tax rate.
 
@@ -73,7 +68,6 @@ def checkout_shipping_tax_rate(
         manager=manager,
         lines=lines,
         address=address,
-        need_tax_calculation=need_tax_calculation,
     )
     return checkout_info.checkout.shipping_tax_rate
 
@@ -84,7 +78,6 @@ def checkout_subtotal(
     checkout_info: "CheckoutInfo",
     lines: Iterable["CheckoutLineInfo"],
     address: Optional["Address"],
-    need_tax_calculation: bool = False,
 ) -> "TaxedMoney":
     """Return the total cost of all the checkout lines, taxes included.
 
@@ -96,7 +89,6 @@ def checkout_subtotal(
         manager=manager,
         lines=lines,
         address=address,
-        need_tax_calculation=need_tax_calculation,
     )
     return quantize_price(checkout_info.checkout.subtotal, currency)
 
@@ -126,7 +118,6 @@ def checkout_total(
     checkout_info: "CheckoutInfo",
     lines: Iterable["CheckoutLineInfo"],
     address: Optional["Address"],
-    need_tax_calculation: bool = False,
 ) -> "TaxedMoney":
     """Return the total cost of the checkout.
 
@@ -141,7 +132,6 @@ def checkout_total(
         manager=manager,
         lines=lines,
         address=address,
-        need_tax_calculation=need_tax_calculation,
     )
     return quantize_price(checkout_info.checkout.total, currency)
 
@@ -152,7 +142,6 @@ def checkout_line_total(
     checkout_info: "CheckoutInfo",
     lines: Iterable["CheckoutLineInfo"],
     checkout_line_info: "CheckoutLineInfo",
-    need_tax_calculation: bool = False,
 ) -> TaxedMoney:
     """Return the total price of provided line, taxes included.
 
@@ -165,7 +154,6 @@ def checkout_line_total(
         manager=manager,
         lines=lines,
         address=address,
-        need_tax_calculation=need_tax_calculation,
     )
     checkout_line = find_checkout_line_info(lines, checkout_line_info.line.id).line
     return quantize_price(checkout_line.total_price, currency)
@@ -177,7 +165,6 @@ def checkout_line_unit_price(
     checkout_info: "CheckoutInfo",
     lines: Iterable["CheckoutLineInfo"],
     checkout_line_info: "CheckoutLineInfo",
-    need_tax_calculation: bool = False,
 ) -> TaxedMoney:
     """Return the unit price of provided line, taxes included.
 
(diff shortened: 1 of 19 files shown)


saleor/checkout/calculations.py
```diff
@@ -355,22 +328,52 @@ def _calculate_and_add_tax(
     address: Optional["Address"] = None,
 ):
     if tax_calculation_strategy == TaxCalculationStrategy.TAX_APP:
-        # Call the tax plugins.
+        # If taxAppId is not configured remain previous behaviour to call plugin and apps
```
maarcingebala, line 331: Consider that "previous" only makes sense when you read this PR; it doesn't when you read the code later when it's merged. I would improve the comment to say that:
- `taxAppId` can refer to an app or a plugin (plugin ID is supported only for backward compatibility and will be removed)
-  when taxAppId is not provided, we run (all?) tax plugins and tax apps (what if both plugin is active, and apps are active?)
- when taxAppId is provided, run a plugin or app with this identifier.

saleor/order/calculations.py
```diff
@@ -200,18 +189,42 @@ def _calculate_and_add_tax(
...
+                lines,
+                manager,
+                prices_entered_with_tax,
+            )
+    else:
+        # Get taxes calculated with flat rates and apply to order.
+        update_order_prices_with_flat_rates(order, lines, prices_entered_with_tax)
+
+
+def _call_plugin_or_app_tax(
```
maarcingebala, line 212: Nit: `_call_plugin_or_tax_app`

saleor/checkout/calculations.py
kadewu: Good point.
About the question, the logic was like this before, I mean to run plugins then go Apps if they are all active, I'm not sure if that behaviour is expected, but probably shouldn't be changed lightly.
That's also one of the reason why `taxAppId` was introduced, as this is not what many customers may expect/want.

maarcingebala: Yes, this old behavior should not be changed and I just wasn't sure what were the assumptions here in the old code. But since we're adding more logic here I think it's worth describing in a comment.

saleor/checkout/calculations.py
```diff
@@ -355,22 +328,52 @@ def _calculate_and_add_tax(
...
+    tax_app_identifier: Optional[str],
+    checkout: "Checkout",
+    manager: "PluginsManager",
+    checkout_info: "CheckoutInfo",
+    lines: Iterable["CheckoutLineInfo"],
+    address: Optional["Address"] = None,
+):
+    from ..plugins.avatax.plugin import AvataxPlugin
+
+    if tax_app_identifier == AvataxPlugin.PLUGIN_IDENTIFIER:
```
korycins, line 370: I don't think that this is correct approach. You're focusing here directly on AvataxPlugin. If you see in the code, you will see that all (or almost all) communication between Saleor and Plugins goes through the PluginsManager. 
With this approach, you will break any potential custom tax plugins in case when someone will set the plugin identifier as tax app. 
Instead of this, you should have a taxPlugin ID returned in form like (just an example): "taxPlugin:<plugin-identifier>". This should be passed to the plugin manager, and plugin manager should extract the plugin identifier and call the logic on the expected plugin (if exists). 

---

Additionally I think, that you don't need to modify the logic responsible for handling the taxes here. 
You could have something like this:
```python
def _calculate_and_add_tax(
    tax_calculation_strategy: str,
    tax_app_identifier: Optional[str],
    checkout: "Checkout",
    manager: "PluginsManager",
    checkout_info: "CheckoutInfo",
    prices_entered_with_tax: bool,
    address: Optional["Address"] = None,
    tax_identifier: Optional[str] = None,
):
    if tax_calculation_strategy == TaxCalculationStrategy.TAX_APP:
        # Call the tax plugins.
        _apply_tax_data_from_plugins(checkout, manager, checkout_info, lines, address, tax_identifier)
        # Get the taxes calculated with apps and apply to checkout.
        tax_data = manager.get_taxes_for_checkout(
            checkout_info, lines, tax_app_identifier
        )
        # If taxAppId is not configured we will for now allow to finalize process for
        # backward compatibility.
        if tax_data is None and tax_app_identifier is not None:
            raise EmptyTaxData("Empty tax data.")
        _apply_tax_data(checkout, lines, tax_data)
    else:
        # Get taxes calculated with flat rates and apply to checkout.
        update_checkout_prices_with_flat_rates(
            checkout, checkout_info, lines, prices_entered_with_tax, address
        )
```
(just added `tax_identifier` as argument), then you can pass it to each manager call, and make something like:
```python
        self,
        checkout_info: "CheckoutInfo",
        checkout_line_info: "CheckoutLineInfo",
        address: Optional["Address"],
            tax_identifier: Optional[str]
    ) -> TaxedMoney:
    ...
        if tax_identifier: 
            if not tax_identifier.startswith("taxPlugin"):
                return default_taxed_value
            else:
                tax_identifier = tax_identifier.split("taxPlugin")[1]
                return self.__run_method_on_single_plugin()

        line_total = self.__run_method_on_plugins(
```

---

Another way of handling it in calculations:
```python
def _calculate_and_add_tax(
    tax_calculation_strategy: str,
    tax_app_identifier: Optional[str],
    checkout: "Checkout",
    manager: "PluginsManager",
    checkout_info: "CheckoutInfo",
    prices_entered_with_tax: bool,
    address: Optional["Address"] = None,
    tax_identifier: Optional[str] = None,
):
    if tax_calculation_strategy == TaxCalculationStrategy.TAX_APP:
        if tax_app_identifier and tax_identifier.startswith("taxPlugin:"):
            # Call the tax plugins.
            _apply_tax_data_from_plugins(checkout, manager, checkout_info, lines, address, tax_identifier)
        # Get the taxes calculated with apps and apply to checkout.
        tax_data = manager.get_taxes_for_checkout(
            checkout_info, lines, tax_app_identifier
        )
```

Code snippets are only suggestions. The most important thing is that we don't want to have any custom logic for plugins inside the direct Saleor code. Saleor doesn't know anything about plugins. Saloer knows only PluginsManager

saleor/checkout/tests/test_calculations.py
```diff
@@ -521,3 +522,87 @@ def test_fetch_checkout_prices_when_tax_exemption_and_not_include_taxes_in_price
 
     assert checkout.total == shipping_price + all_lines_total_price
     assert checkout.subtotal == all_lines_total_price
+
+
+@freeze_time()
+@patch("saleor.plugins.avatax.plugin.AvataxPlugin")
```
korycins, line 528: The test should not be strictly related to AvataxPlugin. If you want to test single tax plugin call, you should mock PluginsManager, and confirm that the request was passed to single plugin (or just confirm that proper method for calling the method on single plugin was called)

korycins: Tests have own dummy plugins, which can be used: `saleor/plugins/tests/sample_plugins.py`

saleor/graphql/tax/mutations/tax_configuration_update.py
```diff
@@ -155,36 +156,42 @@ def clean_input(cls, info: ResolveInfo, instance, data, **kwargs):
...
 
         active_tax_apps = list(get_active_tax_apps(identifiers))
+        active_tax_app_identifiers = [app.identifier for app in active_tax_apps]
 
-        if app_identifier := data.get("tax_app_id"):
-            cls.__verify_tax_app_id(info, app_identifier, active_tax_apps)
+        # include plugin in list of possible tax apps
+        active_tax_app_identifiers.append(AvataxPlugin.PLUGIN_IDENTIFIER)
+
+        if (app_identifier := data.get("tax_app_id")) is not None:
```
korycins, line 173: You already fetched `app_identifier` in line 159, you can use it here, instead of taking it from `data` second time

saleor/graphql/tax/mutations/tax_configuration_update.py
```diff
@@ -155,36 +156,42 @@ def clean_input(cls, info: ResolveInfo, instance, data, **kwargs):
...
+            if (app_identifier := country.get("tax_app_id")) is not None:
                 identifiers.append(app_identifier)
 
         active_tax_apps = list(get_active_tax_apps(identifiers))
+        active_tax_app_identifiers = [app.identifier for app in active_tax_apps]
 
-        if app_identifier := data.get("tax_app_id"):
-            cls.__verify_tax_app_id(info, app_identifier, active_tax_apps)
+        # include plugin in list of possible tax apps
+        active_tax_app_identifiers.append(AvataxPlugin.PLUGIN_IDENTIFIER)
```
korycins, line 171: This should be fetched dynamically from PluginsManager. We can't hardcode it. 
Also the identifier for plugins should be returned with specific prefix, applied by Saleor. 
You could extend the manager in a similar way as it is done in `saleor.plugins.manager.PluginsManager.is_event_active_for_any_plugin`

saleor/order/calculations.py
```diff
@@ -200,18 +189,42 @@ def _calculate_and_add_tax(
...
-        # Get the taxes calculated with plugins.
+        # If taxAppId is not configured remain previous behaviour to call plugin and apps
+        if not tax_app_identifier:
+            # Get the taxes calculated with plugins.
+            _recalculate_with_plugins(manager, order, lines, prices_entered_with_tax)
+            # Get the taxes calculated with apps and apply to order.
+            tax_data = manager.get_taxes_for_order(order, tax_app_identifier)
+            _apply_tax_data(order, lines, tax_data)
+        else:
+            _call_plugin_or_app_tax(
```
korycins: Same sugestion as for `checkout/calculations`. I think that making something similar as in the posted example, could improve the readability in this logic (as we have a lot of if/esle conditions).

saleor/order/tests/test_calculations.py
```diff
@@ -1141,3 +1142,74 @@ def test_order_undiscounted_total(mocked_fetch_order_prices_if_expired):
     assert undiscounted_total == quantize_price(
         expected_undiscounted_total, order.currency
     )
+
+
+@patch("saleor.plugins.avatax.plugin.AvataxPlugin")
+@patch("saleor.order.calculations._recalculate_with_plugins")
+@patch("saleor.plugins.manager.PluginsManager.get_taxes_for_checkout")
+@override_settings(PLUGINS=["saleor.plugins.avatax.plugin.AvataxPlugin"])
```
korycins, line 1150: The test should not be strictly related to AvataxPlugin. If you want to test single tax plugin call, you should mock PluginsManager, and confirm that the request was passed to single plugin (or just confirm that proper method for calling the method on single plugin was called)

korycins (commented):
The tax plugin should be also returned in `shop.availableTaxApps`.  Without this, dashboard will not be able to proper list all available options.

kadewu:
> The tax plugin should be also returned in `shop.availableTaxApps`. Without this, dashboard will not be able to proper list all available options.

About this, we agreed with dashboard, that they will include plugin separately in the list (as we return `list[App]`).

saleor/checkout/calculations.py
kadewu: Remove direct usage of `AvataxPlugin`

Regarding the code snippets I decided to leave it, as it is, cause I believe it's easier to read code and logic regarding circuit break is more compact.

Air-t (approved):

korycins (approved):

szczecha (approved):

PR #15506, opened 2024-02-29
fix `_get_the_cheapest_line` when there is no lines

Air-t:
I want to merge this change because it fixes the case when there is no CheckoutLineInfo provided to find which line is the cheapest.

Fixes SHOPX-313

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/fetch.py
+++ b/saleor/checkout/fetch.py
@@ -410,8 +410,10 @@ def apply_voucher_to_checkout_line(
 
 
 def _get_the_cheapest_line(
-    lines_info: Iterable[CheckoutLineInfo],
-):
+    lines_info: Optional[Iterable[CheckoutLineInfo]],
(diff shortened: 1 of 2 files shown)


lkostrowski (approved):

saleor/checkout/fetch.py
```diff
@@ -410,8 +410,10 @@ def apply_voucher_to_checkout_line(
 
 
 def _get_the_cheapest_line(
-    lines_info: Iterable[CheckoutLineInfo],
+    lines_info: Optional[Iterable[CheckoutLineInfo]],
```
maarcingebala, line 413: We could also add typing for the return value to make it even better.

maarcingebala (approved):

IKarbowiak (approved):

PR #15532, opened 2024-03-04
Fix external shipping method calls on checkout lines add

tomaszszymanski129:
I want to merge this change because it fixes a bug that called external shipping methods on `checkoutLinesAdd` even when assigned one was no longer valid.
Steps to reproduce:
1. Add shippable item to the cart, go to checkout, fill in the address and select external shipping method.
2. Invalidate shipping method on app side (so app no longer returns assigned method).
3. Go back to product page and add another item to the cart.
4. `list_shipping_methods_for_checkout` webhook is called, this is expected
5. as the method was no longer valid (response from app didn't include it) it should be removed from checkout metadata - this is the issue, it was not happening.
6. Add another item to the cart
7. `list_shipping_methods_for_checkout` should no longer be called as Saleor deattatched the external method from checkout metadata.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/mutations/utils.py
+++ b/saleor/graphql/checkout/mutations/utils.py
@@ -56,15 +56,6 @@ class CheckoutLineData:
     metadata_list: Optional[list] = None
 
 
-def _is_external_shipping_valid(checkout_info: "CheckoutInfo"):
-    if external_shipping_id := get_external_shipping_id(checkout_info.checkout):
-        for method in checkout_info.valid_delivery_methods:
-            if method.id == external_shipping_id:
-                return True
-        return False
-    return True
-
-
 def clean_delivery_method(
     checkout_info: "CheckoutInfo",
     lines: Iterable[CheckoutLineInfo],
@@ -96,6 +87,15 @@ def clean_delivery_method(
 
     valid_methods = checkout_info.valid_delivery_methods
     return method in valid_methods
+
+
+def _is_external_shipping_valid(checkout_info: "CheckoutInfo"):
+    if external_shipping_id := get_external_shipping_id(checkout_info.checkout):
+        for method in checkout_info.valid_delivery_methods:
+            if method.id == external_shipping_id:
+                return True
+        return False
+    return True
 
 
 def update_checkout_external_shipping_method_if_invalid(
--- a/saleor/graphql/checkout/mutations/utils.py
+++ b/saleor/graphql/checkout/mutations/utils.py
@@ -23,6 +23,8 @@ from ....checkout.fetch import CheckoutInfo, CheckoutLineInfo
 from ....checkout.utils import (
     calculate_checkout_quantity,
     clear_delivery_method,
+    delete_external_shipping_id,
+    get_external_shipping_id,
     is_shipping_required,
 )
 from ....core.exceptions import InsufficientStock, PermissionDenied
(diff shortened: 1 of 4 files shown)


saleor/graphql/checkout/mutations/utils.py
```diff
@@ -54,6 +56,15 @@ class CheckoutLineData:
     metadata_list: Optional[list] = None
 
 
+def _is_external_shipping_valid(checkout_info: "CheckoutInfo"):
```
IKarbowiak, line 59: Could we add this method below `update_checkout_external_shipping_method_if_invalid`? It will be easier to read, than jumping function above

saleor/graphql/checkout/tests/mutations/test_checkout_lines_add.py
```diff
@@ -724,6 +727,80 @@ def test_checkout_lines_add_existing_variant_override_previous_custom_price(
     assert line.price_override == price
 
 
+@mock.patch("saleor.plugins.manager.PluginsManager.list_shipping_methods_for_checkout")
+def test_checkout_lines_add_deletes_external_shipping_method_if_invalid(
```
IKarbowiak, line 731: Missing `given, when, then`

saleor/graphql/checkout/tests/mutations/test_checkout_lines_add.py
```diff
@@ -724,6 +727,80 @@ def test_checkout_lines_add_existing_variant_override_previous_custom_price(
...
+    )
+    content = get_graphql_content(response)
+    data = content["data"]["checkoutLinesAdd"]
+    assert not data["errors"]
+    metadata.refresh_from_db()
+    assert metadata.private_metadata == {}
+
+
+@mock.patch("saleor.plugins.manager.PluginsManager.list_shipping_methods_for_checkout")
+def test_checkout_lines_add_doesnt_delete_external_shipping_method_if_valid(
```
IKarbowiak, line 762: Same here

IKarbowiak (approved):

saleor/graphql/checkout/mutations/utils.py
```diff
@@ -87,8 +89,25 @@ def clean_delivery_method(
     return method in valid_methods
 
 
-def update_checkout_shipping_method_if_invalid(
+def _is_external_shipping_valid(checkout_info: "CheckoutInfo"):
+    if external_shipping_id := get_external_shipping_id(checkout_info.checkout):
+        for method in checkout_info.valid_delivery_methods:
```
kadewu, line 94: Maybe just `return external_shipping_id in [method.id for method in checkout_info.valid_delivery_methods]`, WDYT?

kadewu (approved):

saleor/graphql/checkout/mutations/utils.py
```diff
@@ -87,8 +89,25 @@ def clean_delivery_method(
     return method in valid_methods
 
 
-def update_checkout_shipping_method_if_invalid(
+def _is_external_shipping_valid(checkout_info: "CheckoutInfo"):
```
Air-t, line 92: maybe it's also worth adding return value type?

Air-t (approved):

saleor/graphql/checkout/mutations/utils.py
tomaszszymanski129: merged accidentally but this is a good point, i'm preparing ports, will do so there.

PR #15546, opened 2024-03-06
Change `discount_promotionrule_variants` `id` to `bigint`

IKarbowiak:
Change `discount_promotionrule_variants` `id` to `bigint`

Resolves comment: https://github.com/saleor/saleor/issues/15145#issuecomment-1874093887

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


NyanKiyoshi (changes requested):
Why not use Django's `BigAutoField` instead of using raw SQL? https://docs.djangoproject.com/en/3.2/ref/models/fields/#bigautofield

IKarbowiak:
@NyanKiyoshi It's an ID on the cross-reference table so I don't want to define this table in the code only to change id to `BigAutoField`. 
Also, I checked what the SQL statements are under the django migration for changing the `id` to `BigAutoField` and it contains the same logic that is added in this migration.

korycins (approved):

NyanKiyoshi:
> @NyanKiyoshi It's an ID on the cross-reference table so I don't want to define this table in the code only to change id to `BigAutoField`. Also, I checked what the SQL statements are under the django migration for changing the `id` to `BigAutoField` and it contains the same logic that is added in this migration.

This sounds dangerous to me because Django may run checks (e.g., new relations added) and think it's int32 while the value is int64 which could lead to major bugs on some datasets. As well it hides the actual types from the developers, maintainers will also not check dozens of migration files and thus will be surprised if unexpected behaviors start happening due to using int64.

IMO the effort of defining the M2M table in Django ORM is worth it due to the risks the current migration brings.

IKarbowiak:
@NyanKiyoshi, thank you for the explenation, so I will change that :)

zedzior (approved):

korycins (approved):

PR #15556, opened 2024-03-06
Accept transaction item in order-granted-refund mutations

korycins:
I want to merge this change because it covers (internal link) https://linear.app/saleor/issue/MERX-179/extend-ordergrantedrefund-to-accept-assignment-of-transactionitem
Based on RFC: https://github.com/saleor/saleor/discussions/15458

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/graphql/order/mutations/order_grant_refund_create.py
```diff
@@ -81,6 +87,18 @@ class OrderGrantRefundCreateInput(BaseInputObjectType):
         + PREVIEW_FEATURE,
         required=False,
     )
+    transaction_id = graphene.ID(
+        description=(
+            "The ID of the transaction item related to the granted refund. "
+            "transaction.chargedAmount needs to be equal or greater than provided "
```
IKarbowiak, line 93:
```suggestion
            "The `transaction.chargedAmount` needs to be equal or greater than provided "
```

saleor/graphql/order/mutations/order_grant_refund_create.py
```diff
@@ -81,6 +87,18 @@ class OrderGrantRefundCreateInput(BaseInputObjectType):
...
+            "`amount`. If `amount` calculated automatically, the "
+            "min(amount, transaction.chargedAmount) will be used."
```
IKarbowiak, line 95: I have problem with understanding this part:
`If amount calculated automatically, the min(amount, transaction.chargedAmount) will be used.`
Do you mean in case `amount` is not provided? So in such case what the `amoung` in `min` is?

saleor/graphql/order/mutations/order_grant_refund_update.py
```diff
@@ -97,6 +103,17 @@ class OrderGrantRefundUpdateInput(BaseInputObjectType):
         + PREVIEW_FEATURE,
         required=False,
     )
+    transaction_id = graphene.ID(
+        description=(
+            "The ID of the transaction item related to the granted refund. "
+            "transaction.chargedAmount needs to be equal or greater than provided "
```
IKarbowiak, line 109:
```suggestion
            "The `transaction.chargedAmount` needs to be equal or greater than provided "
```

saleor/graphql/order/mutations/order_grant_refund_update.py
```diff
@@ -97,6 +103,17 @@ class OrderGrantRefundUpdateInput(BaseInputObjectType):
...
+            "`amount`. If `amount` calculated automatically, the "
+            "min(amount, transaction.chargedAmount) will be used."
```
IKarbowiak, line 111: Same here. IMO should be reworded

IKarbowiak (approved):

zedzior (approved):

saleor/graphql/order/mutations/order_grant_refund_create.py
korycins: If you don't provide the `amount` as an input, the `amount` is calculated automatically. 
So "If `amount` calculated automatically," then we will use the lower value from [`amount`, `chargedAmount`].
I can change it if you have a better suggestion. For now I see that I could change `min(amountCaclulatedAutomatically, transaction.chargedAmount)`

IKarbowiak: Maybe this sentence is not needed, as there is no difference in comparison for provided and calculated `amount`, and instead it could be:
```
"""The ID of the transaction item related to the granted refund.
transaction.chargedAmount needs to be equal or greater than amount.
"""
```

korycins: There is difference:
- if `amount` provided as input, then raise error when `amount > transaction.chargedAmount`
- if `amount` not provided as input, then set amount: `min(caclulated_amount,  transaction.chargedAmount)`
Will try to figure out a more meaningful description.

PR #15565, opened 2024-03-07
Validate numeric values in `atrribueValueCreate`.

zedzior:
I want to merge this change, because it adds missing validation for numeric values when running `atrribueValueCreate`/`atrribueValueUpdate`mutations.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/attribute/mutations/attribute_value_create.py
+++ b/saleor/graphql/attribute/mutations/attribute_value_create.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 import graphene
 from django.core.exceptions import ValidationError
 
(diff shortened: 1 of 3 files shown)


korycins (approved):

saleor/graphql/attribute/mutations/attribute_value_create.py
```diff
@@ -63,23 +65,34 @@ def clean_input(cls, info: ResolveInfo, instance, data, **kwargs):
...
+        if input_type == AttributeInputType.NUMERIC and "name" in cleaned_input:
+            try:
+                float(cleaned_input["name"])
+            except ValueError:
+                errors["name"].append(
+                    ValidationError(
+                        "Value of numeric attribute must be numeric.",
+                        code=AttributeErrorCode.INVALID.value,
+                    )
+                )
```
IKarbowiak, line 94: It shouldn't be here. This mutation is only for creating the values for attributes with choices. The `numeric` attribute is not this kind of attribute. If it's possible, the validation of that is missing, not for validating the numeric value.

IKarbowiak (changes requested):

saleor/graphql/attribute/mutations/attribute_value_create.py
zedzior: Thanks Iga. I can change the validation, but I would like to know:

1. Where did you get this info, that the mutation is for attribute with choices only?
2. What other mutation can be used, to create numeric attribute value?

I need to know that in order to display a proper error description.

IKarbowiak: 1. Unfortunately it's not described anyway directly, but the input for this mutation has a deprecation message for some fields that suggested this: https://github.com/saleor/saleor/blob/main/saleor/graphql/attribute/mutations/attribute_create.py#L32
Furthermore, attributes without choices are created per instance, so you need to specify instance when creating it, this mutation doesn't take such value, so you will end up with `AttributeValue` that is a trash.
2. You need to use the instance mutation for that, like `ProductUpdate`, `ProductVariantUpdate`, `ProductVariantCreate`. It depends on to which object you want to create an attribute value

Issue #15190, opened 2024-01-08
Investigate why multiple rows are created for the same value of numeric attribute


maarcingebala:
In one of the production databases, we saw a pattern where there are around 70k `AttributeValue` rows for a numeric attribute that most likely should have 10 or so unique values. Example:

SQL:
```
select attribute_attributevalue.slug, attribute_attributevalue.name from attribute_attributevalue join attribute_attribute on attribute_attribute.id = attribute_attributevalue.attribute_id where attribute_attribute.name = 'example';
```

Result:
```
    slug    | name
------------+------
 27512_1537 | 0
 70224_1537 | 0
 61989_1537 | 0
 58330_1537 | 0
 28033_1537 | 0
 14285_1537 | 0
 63169_1537 | 3
 70011_1537 | 0
 58070_1537 | 0
 58065_1537 | 0
```

Todo:
- Investigate why the same value is stored multiple times with different slugs, and if it can't be optimized not to duplicate the same value many times.
- Why the `value` field is empty for numeric attributes, and the actual value is only stored in the `name` field.

Context: https://saleorcommerce.slack.com/archives/C06C6UWV45V/p1704471134629569

zedzior:
> Why the value field is empty for numeric attributes, and the actual value is only stored in the name field.
We use `value` field only for swatch attributes. Others use `name`.

maarcingebala:
Closing as not planned for now.

PR #15566, opened 2024-03-07
Change slug pattern for numeric attribute values.

zedzior:
Issue: https://github.com/saleor/saleor/issues/15190

I want to merge this change, because it changes the slug of numeric attribute values from `{instance.id}_{attribute.id}` to `{actual_numeric_value}_{attribute.id}`. This will make numeric values reusable. Previous solution was resulting in redundant, duplicated attribute values created. 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/attribute/utils.py
+++ b/saleor/graphql/attribute/utils.py
@@ -594,7 +594,7 @@ class AttributeAssignmentMixin:
     @classmethod
     def _pre_save_numeric_values(
         cls,
-        instance: T_INSTANCE,
+        _,
         attribute: attribute_models.Attribute,
         attr_values: AttrValuesInput,
     ):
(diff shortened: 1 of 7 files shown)


saleor/graphql/attribute/utils.py
```diff
@@ -605,10 +605,13 @@ def _pre_save_numeric_values(
         else:
             return tuple()
 
+        slug = slugify(unidecode(f"{value}_{attribute.id}"))
```
IKarbowiak, line 608: It's not the right solution. The task what to investigate `why multiple rows are created` and not to change the number attribute slug pattern. 
We have two different types of attributes:
- `TYPES_WITH_CHOICES`
- `TYPES_WITH_UNIQUE_VALUES`
The numeric attribute is the type with unique values.

With your solution, two different instances of le's say `Product` may start sharing the `AttributeValue`. When one of these products is deleted, automatically `AttributeValue` is deleted, which means that we will end up with an empty value for attribute for the second `Product`.

IKarbowiak (changes requested):

saleor/graphql/attribute/utils.py
zedzior: Hmm, are you sure that we delete shared attribute values? I wrote a new test, which shows, that with the solution:

1. we don't create duplicated attribute values anymore
2. after deleting one of the instances, which share the same attribute value, the value is still associated with remaining instances

Do I miss something? Can you please check it again?

IKarbowiak: Here is the method that is responsible for that:
https://github.com/saleor/saleor/blob/main/saleor/graphql/product/mutations/product/product_delete.py#L85

zedzior:
We decided to extend the solution for other unique attribute types.

PR #15571, opened 2024-03-08
Add replica usages to checkout and order related queries

maarcingebala:
Add proper usages of replica in functions related to checkout and order calculations.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/calculations.py
+++ b/saleor/checkout/calculations.py
@@ -12,7 +12,6 @@ from ..core.taxes import TaxData, TaxEmptyData, zero_money, zero_taxed_money
 from ..discount.utils import (
     create_or_update_discount_objects_from_promotion_for_checkout,
 )
-from ..graphql.core.context import get_database_connection_name_from_flag
 from ..payment.models import TransactionItem
 from ..plugins import PLUGIN_IDENTIFIER_PREFIX
 from ..tax import TaxCalculationStrategy
@@ -39,6 +38,7 @@ def checkout_shipping_price(
     checkout_info: "CheckoutInfo",
     lines: Iterable["CheckoutLineInfo"],
     address: Optional["Address"],
+    database_connection_name: str = settings.DATABASE_CONNECTION_DEFAULT_NAME,
 ) -> "TaxedMoney":
     """Return checkout shipping price.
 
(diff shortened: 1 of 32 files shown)


fowczarek (approved):

IKarbowiak (approved):

saleor/checkout/calculations.py
```diff
@@ -222,6 +223,9 @@ def _fetch_checkout_prices_if_expired(
     last price update is greater than settings.CHECKOUT_PRICES_TTL.
     """
     checkout = checkout_info.checkout
+    database_connection_name = get_database_connection_name_from_flag(
+        manager._allow_replica
```
korycins, line 227: IMHO, not convinced of using `_allow_replica` in this way:
- You're using the variable marked as protected outside the PluginsManager. 
- You treat `PluginsManager` as a structure that stores the configuration for the rest of the Saleor's logic, but this is not a responsibility of the PluginsManager.

maarcingebala: I know it may seem a bit hacky, but in the context of each request, the plugin manager is always initialized with the proper value of allow_replica, and we don't have to pass it all the way to calculation functions. I prefer this instead of changing many signatures just for this reason.

korycins: Still I am not convinced that this is a proper way of using PluginsManager. IMHO we should extend all functions to pass the replica info in the same way in all places.  Some of the modified functions have defined input for accepting the database name (`database_connection_name`), and some treat the manager as a settings provider, to set up the Saleor's  logic. So we have two different ways of passing this info into the code. I think we should unify it and use `database_connection_name` in all places. 
Additionally: In the future, we are going to fully drop the plugins (after providing apps replacement) and fully drop the pluginsManager. It means that no matter what, we will have to finally modify the rest of the functions (which use pluginsManager as a configuration context)  to use `database_connection_name`. So why not make it in way that will not need any modification after dropping manager?

maarcingebala: I'll fix it

korycins (approved):

Issue #15124, opened 2023-12-26
Bug: Multiple CheckoutLineDiscount applied to single line


korycins:
### What are you trying to achieve?

In the case of thread race, there is a possibility that we will create multiple CheckoutLineDiscount for single checkoutLine. 
When line price is calculated, we apply all `line.discounts`. In the end, we could end up with a line discounted multiple times.
The issue here is that at the beginning of the mutations, we are creating the object of  `CheckoutLineInfo`. At the same moment, we're fetching all `CheckoutLineDiscount` assigned to the checkout. Later in `create_or_update_discount_objects_from_sale_for_checkout`, we' re using the `discounts` from `CheckoutLineInfo` to confirm if new `CheckoutLineDiscount` should be created. We don't fetch it from DB, so it means that there is a time range, where the different requests could create the `CheckoutLineDiscount`. At the end we have single line with multiple CheckoutLineDiscounts.

### Steps to reproduce the problem

I was reproducing it, by having to local servers and placing a sleep `create_or_update_discount_objects_from_sale_for_checkout` before calling a `bulk_create/update`.

### What did you expect to happen?

I expect, that we will always have single checkoutLineDiscount created from Sale. I also expect, that the logic will be able to handle potential thread races.

### Environment

Saleor version: 3.14+
OS and version: …

korycins:
The patches that we applied during christmas should be reverted when we solve the problem (3.14-3.18).

PR #15599, opened 2024-03-13
Protect catalogue discount handler from thread race.

zedzior:
I want to merge this change because it ensures no duplicated catalogue discounts applied in case of thread race.
Issue: https://github.com/saleor/saleor/issues/15124
https://linear.app/saleor/issue/MERX-49/multiple-checkoutlinediscount-applied-to-single-line

Most of the hanges from this PR will be handled in: https://github.com/saleor/saleor/pull/15767
Utilizing `unique_type` will be handled in separate PR

# Impact

- [x] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/discount/tests/test_utils/test_create_or_update_discount_objects_from_promotion_for_checkout.py
```diff
@@ -476,127 +477,6 @@ def test_create_percentage_discount_multiple_quantity_in_lines(
         assert not checkout_line_info.discounts
 
 
-def test_create_discount_multiple_rules_applied(
```
zedzior, line 479: Catalogue discounts shouldn't be stackable so the test is not valid.

saleor/graphql/checkout/tests/mutations/test_checkout_complete_with_payment.py
```diff
@@ -1679,186 +1678,6 @@ def test_checkout_complete_product_on_old_sale(
     ).exists(), "Checkout should have been deleted"
 
 
-def test_checkout_complete_multiple_rules_applied(
```
zedzior, line 1682: Catalogue discounts shouldn't be stackable so the test is not valid.

saleor/graphql/checkout/tests/mutations/test_order_create_from_checkout.py
```diff
@@ -1281,91 +1281,6 @@ def test_order_from_checkout_on_catalogue_and_gift_promotion(
     ).gross.amount == top_price + line_discount.amount_value
 
 
-def test_order_from_checkout_multiple_rules_applied(
```
zedzior, line 1284: Catalogue discounts shouldn't be stackable so the test is not valid.

saleor/discount/utils.py
```diff
@@ -407,11 +416,12 @@ def prepare_line_discount_objects_for_catalogue_promotions(
     for line_info in lines_info:
         line = line_info.line
 
-        # get the existing discounts for the line
-        discounts_to_update = line_info.get_catalogue_discounts()
-        rule_id_to_discount = {
-            discount.promotion_rule_id: discount for discount in discounts_to_update
-        }
```
zedzior, line 414: This was wrong and made our update in fact an add&delete

saleor/discount/utils.py
```diff
@@ -430,17 +440,9 @@ def prepare_line_discount_objects_for_catalogue_promotions(
             line_discounts_to_remove.extend(discounts_to_update)
             continue
 
-        # delete the discount objects that are not valid anymore
-        line_discounts_to_remove.extend(
-            _get_discounts_that_are_not_valid_anymore(
-                line_info.rules_info,
-                rule_id_to_discount,  # type: ignore[arg-type]
-            )
-        )
```
zedzior, line 439: This was wrong and in case of update, in fact, we made add&delete.

saleor/discount/migrations/0077_unique_type.py
```diff
@@ -0,0 +1,53 @@
...
+            field=models.CharField(blank=True, max_length=64, null=True),
+        ),
+        migrations.AlterUniqueTogether(
+            name="checkoutdiscount",
+            unique_together={
+                ("checkout_id", "unique_type"),
+                ("checkout_id", "promotion_rule_id"),
+            },
+        ),
+        migrations.AlterUniqueTogether(
```
kadewu, line 41: we need to add indexes that will be created for uniqueness concurrently.

saleor/discount/interface.py
```diff
@@ -59,10 +59,14 @@ def fetch_variant_rules_info(
         if variant_channel_listing
         else []
     )
+
     rules_info = []
-    for listing_promotion_rule in listings_rules:
+    if listings_rules:
+        listing_promotion_rule = max(
```
kadewu, line 65: Is there a reason why max? Maybe some comment here would be useful.

saleor/discount/tests/test_discounts.py
```diff
@@ -579,3 +580,43 @@ def test_get_discount_translated_name_no_translations(rule_info):
 
     # then
     assert translated_name is None
+
+
+def test_unique_together_order(order):
```
kadewu, line 585: Not sure if those tests are needed, for me it seems like we are testing here if the django works properly.

saleor/discount/interface.py
zedzior: Added: 

> Before introducing unique_type on discount models, there were possibility
   to have multiple catalogue discount associated with single line. In such a
   case, we should pick the best discount (with the highest discount amount)

saleor/discount/migrations/0077_unique_type.py
zedzior: According to postgres docs, it is created automatically:
https://www.postgresql.org/docs/9.4/indexes-unique.html#:~:text=PostgreSQL%20automatically%20creates%20a%20unique%20index%20when%20a%20unique%20constraint%20or%20primary%20key%20is%20defined%20for%20a%20table.%20The%20index%20covers%20the%20columns%20that%20make%20up%20the%20primary%20key%20or%20unique%20constraint%20(a%20multicolumn%20index%2C%20if%20appropriate)%2C%20and%20is%20the%20mechanism%20that%20enforces%20the%20constraint.

saleor/discount/models.py
```diff
@@ -454,6 +454,9 @@ class BaseDiscount(models.Model):
         choices=DiscountType.CHOICES,
         default=DiscountType.MANUAL,
     )
+    # `unique_type` is a field to ensure, that only single specific discount type
+    # can be associated with discounted object
+    unique_type = models.CharField(max_length=64, null=True, blank=True)
```
korycins, line 459: Maybe worth to use `choices` here?

saleor/discount/models.py
```diff
@@ -454,6 +454,9 @@ class BaseDiscount(models.Model):
         choices=DiscountType.CHOICES,
         default=DiscountType.MANUAL,
     )
+    # `unique_type` is a field to ensure, that only single specific discount type
+    # can be associated with discounted object
```
korycins, line 458: Maybe worth to mention that we are adding it as a separate field, as previously there was such possibility, and starting from Saleor 3.X we want to fully block it on database.

saleor/discount/migrations/0077_unique_type.py
zedzior: Got your point, we need to ensure it is added **CONCURRENTLY**

saleor/discount/utils.py
```diff
@@ -430,17 +445,9 @@ def prepare_line_discount_objects_for_catalogue_promotions(
...
-
-        for rule_info in line_info.rules_info:
+        if line_info.rules_info:
```
korycins, line 449: What will happen in the case when:
I will have some discount assigned to `line_info` == `discount_to_update` is not None but `line_info.rules_info` will be empty list or None?  As far as I understand in that case, the catalogue discount should be removed. Do we do that? Is it event possible?

saleor/discount/utils.py
```diff
@@ -376,17 +376,26 @@ def create_checkout_line_discount_objects_for_catalogue_promotions(
...
-    if discounts_to_create_inputs:
-        new_line_discounts = [
-            CheckoutLineDiscount(**input) for input in discounts_to_create_inputs
-        ]
-        CheckoutLineDiscount.objects.bulk_create(new_line_discounts)
+    with transaction.atomic():
+        # Protect against potential thread race. CheckoutLine object can have only
+        # single catalogue discount applied.
+        line_ids = [line_info.line.id for line_info in lines_info]
+        CheckoutLine.objects.filter(id__in=line_ids).only("id").select_for_update()
```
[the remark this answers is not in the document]
zedzior: I can lock whole checkout. I belive the reason we had a deadlock was: https://github.com/saleor/saleor/pull/15599#discussion_r1535712793

saleor/discount/utils.py
```diff
@@ -376,17 +376,31 @@ def create_checkout_line_discount_objects_for_catalogue_promotions(
...
-        new_line_discounts = [
-            CheckoutLineDiscount(**input) for input in discounts_to_create_inputs
-        ]
-        CheckoutLineDiscount.objects.bulk_create(new_line_discounts)
+    with transaction.atomic():
+        # Protect against potential thread race. CheckoutLine object can have only
+        # single catalogue discount applied.
+        line_ids = [line_info.line.id for line_info in lines_info]
+        _lines_lock = list(
+            CheckoutLine.objects.filter(id__in=line_ids).select_for_update(
```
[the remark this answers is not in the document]
zedzior: I belive the reason we had a deadlock was: https://github.com/saleor/saleor/pull/15599#discussion_r1535712793. Previously, instead of update, we were making add&delete.

saleor/discount/models.py
zedzior: Initially I used `DiscountType` here. But setting this char field gives us more freedom to handle uniqueness of discounts without changing the API or creating a migration. Does it make sense?

saleor/discount/interface.py
zedzior: @korycins I don't think it is necessary. Just checked and `update_discounted_prices_for_promotion` is currently designed to keep single `VariantChannelListingPromotionRule` relation (while db allows us to keep multiples, we don't use it now)

saleor/discount/models.py
zedzior: You are right, we should use `DiscountType` here

saleor/discount/utils.py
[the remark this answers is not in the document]
zedzior: I locked whole checkout instead of lines.

saleor/discount/interface.py
```diff
@@ -59,10 +59,14 @@ def fetch_variant_rules_info(
...
+            list(listings_rules),
+            key=lambda x: x.discount_amount,
+        )
         promotion = listing_promotion_rule.promotion_rule.promotion
```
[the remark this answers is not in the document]
zedzior: We operate here on prefetched data

saleor/discount/models.py
zedzior: Description updated

kadewu (approved):

PR #15636, opened 2024-03-20
Add decorator prevent_sync_event_circular_query to checkout.deliveryMethod

8r2y5:
I want to merge this change because this change will prevent circular queries when asking for `checkout.deliveryMethod`

# Impact

- [ ] New migrations

# Docs

- [x] Link to documentation:
https://github.com/saleor/saleor-docs/pull/1121

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/types.py
+++ b/saleor/graphql/checkout/types.py
@@ -855,6 +855,7 @@ class Checkout(ModelObjectType[models.Checkout]):
         )
 
     @staticmethod
+    @prevent_sync_event_circular_query
     def resolve_delivery_method(root: models.Checkout, info: ResolveInfo):
         return (
             CheckoutInfoByCheckoutTokenLoader(info.context)


saleor/graphql/checkout/types.py
```diff
@@ -855,6 +855,7 @@ def resolve_shipping_methods(root: models.Checkout, info: ResolveInfo):
         )
```
Air-t, line 855: shouldn't we add a test for that, that this decorator solved the actual problem? I believe that decorator is already tested.

Air-t (approved):

tomaszszymanski129 (approved):

korycins (commented):
Are you aware, that this can be a breaking change for existing subscriptions? 
What If I don't have external-shipping methods, and I am fetching this field with a different sync webhook?

PR #15650, opened 2024-03-20
Charge transaction items on orderConfirm

tomaszszymanski129:
I want to merge this change because it implements charging transaction items at the point of order confirmation.
I'm merging this to the feature branch because it relies on a couple more tickets (like the one in which there is a flag to control this behavior).

Ticket: https://linear.app/saleor/issue/SHOPX-193/orderconfirm-should-capture-transactionitems-that-have-action-charge

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/mutations/order_confirm.py
+++ b/saleor/graphql/order/mutations/order_confirm.py
@@ -1,3 +1,4 @@
+import uuid
 from typing import cast
 
 import graphene
@@ -11,7 +12,8 @@ from ....order.actions import order_charged, order_confirmed
 from ....order.error_codes import OrderErrorCode
 from ....order.fetch import fetch_order_info
 from ....order.utils import update_order_display_gross_prices
-from ....payment import gateway
+from ....payment import TransactionAction, TransactionEventType, gateway
+from ....payment.gateway import request_charge_action
 from ....permission.enums import OrderPermissions
 from ...app.dataloaders import get_app_promise
 from ...core import ResolveInfo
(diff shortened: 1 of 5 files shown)


saleor/graphql/order/mutations/order_confirm.py
```diff
@@ -60,6 +62,34 @@ def get_instance(cls, info: ResolveInfo, **data):
             )
         return instance
 
+    @staticmethod
+    def charge_transaction_items(user, app, order, manager):
+        for transaction_item in order.payment_transactions.filter(
+            available_actions__contains=[TransactionAction.CHARGE],
+            authorized_value__gt=0,
+        ):
+            charge_value = transaction_item.authorized_value
```
korycins, line 71: you could just make `min(charge_value, order.total.gross.amount)` instead of `if`

korycins: Also, you could have multiple transaction_items. So you need to reduce the value that you need to charge, based on the sum of the previously charged value.

saleor/graphql/order/mutations/order_confirm.py
```diff
@@ -60,6 +62,34 @@ def get_instance(cls, info: ResolveInfo, **data):
...
+            if transaction_item.authorized_value > order.total.gross.amount:
+                charge_value = order.total.gross.amount
+            event = transaction_item.events.create(
```
korycins, line 74: `saleor.graphql.payment.mutations.transaction.transaction_request_action.TransactionRequestAction` has `create_transaction_event_requested` method. It creates a new request event. Instead of doing the same thing here, we should extract this logic outside the mutation, and use it in both places. To keep the "create-event" logic in single place.

saleor/graphql/order/tests/mutations/test_order_confirm.py
```diff
@@ -103,6 +106,53 @@ def test_order_confirm(
     )
 
 
+@patch("saleor.graphql.order.mutations.order_confirm.request_charge_action")
+def test_order_confirm_charge_transaction_items(
```
korycins, line 110: IMHo, you should add three more tests here.
One with multiple transaction_items, where sum of authorized_value is equal to order.total
Second one with single transaction_item which has authorized_value bigger than order.total(to confirm that we always charge maximally `order.total` value)
Third one with multiple transaction_items, where sum of authorized_value is bigger than order.total.

saleor/graphql/order/mutations/order_confirm.py
```diff
@@ -60,6 +62,34 @@ def get_instance(cls, info: ResolveInfo, **data):
...
+                amount_value=charge_value,
+                currency=transaction_item.currency,
+                type=TransactionEventType.CHARGE_REQUEST,
+                user=user,
+                app=app,
+                app_identifier=app.identifier if app else None,
+                idempotency_key=str(uuid.uuid4()),
+            )
+            request_charge_action(
```
korycins, line 83: It can raise an exception. You should handle it. Also from the business perspective, we should clarify, what should we do in case of having more than a single transaction_item, where one of them raises an exception - do we want to stop the charging process? Stop orderConfirm? or maybe skip the transaction and proceed the rest of them?
I would say, that we should continue the process, and always mark the order as confirmed, as this action was explicitly requested by staff user, but I also think that it is worth to discuss internally with your team.

tomaszszymanski129:
I'm closing this one, not to hang forever since the priorities got switched but I'll include link to this one in the ticket description for future reference.

PR #15682, opened 2024-03-25
Fix relation between order and click and collect address

IKarbowiak:
Prevent warehouse addresses from being overwritten when changing shipping addresses for click and collect orders.

When completing checkout, the order should create own address instance it shouldn't be share with the cc warehouse.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/complete_checkout.py
+++ b/saleor/checkout/complete_checkout.py
@@ -152,6 +152,9 @@ def _process_shipping_data_for_order(
         if checkout_info.user.addresses.filter(pk=shipping_address.pk).exists():
             shipping_address = shipping_address.get_copy()
 
+    if shipping_address and delivery_method_info.warehouse_pk:
+        shipping_address = shipping_address.get_copy()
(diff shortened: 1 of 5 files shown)


korycins (approved):

zedzior (approved):

Air-t (approved):

patrys (commented):
Shouldn't we have a regression test that shows when addresses get duplicated and ensures they no longer do?

IKarbowiak:
Patrys I updated the tests to ensure that - previously the `shipping_address` was the same instance that cc warehouse address. I'm checking in tests if the ids are different.

patrys:
@IKarbowiak I'd rather have a separate test that explicitly checks for that problem so it's clear why we're doing it. Tests with a dozen different assertions are not self-documenting and if they stop passing, someone will just flip the equality check.

PR #15714, opened 2024-03-26
Refactor line weight calculation to reuse DB queries

maarcingebala:
Refactor weight calculation in `applicable_shipping_methods_for_instance` to use already prefetched data, instead of making separate DB queries. In benchmark it results in queries count going down from 88 to 72 for creating a checkout.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/shipping/models.py
+++ b/saleor/shipping/models.py
@@ -24,6 +24,7 @@ from . import PostalCodeRuleInclusionType, ShippingMethodType
 from .postal_codes import filter_shipping_methods_by_postal_code_rules
 
 if TYPE_CHECKING:
+    from ..account.models import Address
     from ..checkout.fetch import CheckoutLineInfo
     from ..checkout.models import Checkout
     from ..order.fetch import OrderLineInfo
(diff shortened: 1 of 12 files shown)


fowczarek (approved):

saleor/shipping/models.py
```diff
@@ -165,32 +166,44 @@ def applicable_shipping_methods_for_instance(
...
+            country_code = shipping_address.country.code
+
         if lines is None:
             # TODO: lines should comes from args in get_valid_shipping_methods_for_order
             lines = list(instance.lines.prefetch_related("variant__product").all())  # type: ignore[misc] # this is hack # noqa: E501
         instance_product_ids = {
             line.variant.product_id for line in lines if line.variant
         }
+
+        from ..checkout.models import Checkout
```
zedzior, line 188: Do we need this?

zedzior (approved):

saleor/shipping/models.py
maarcingebala: It's used in the `isinstance` check below. Also it's local import due to circular dependencies, but fixing them would require too much refactoring for now.

PR #15748, opened 2024-04-03
Optimize validate_draft_order function to run less queries

maarcingebala:
The `validate_draft_order` function is used in two queries and one mutation. In each case, it runs a lot of duplicated queries due to not properly using dataloaders and accessing not prefetched data via implicit DB queries. This PR partially improves it and loads order lines via dataloader, which are then passed around via arguments. 

There are still many queries to optimize here but it already improved a lot. Example: fetching 5 draft orders and `errors` field:
```
{
  draftOrders(first:5) {
    edges {
      node {
        id
        errors {
          field
          message
        }
      }
    }
  }
}
```

Before: 48 queries
After: 19 queries

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/types.py
+++ b/saleor/graphql/order/types.py
@@ -1930,21 +1930,26 @@ class Order(ModelObjectType[models.Order]):
     def resolve_can_finalize(root: models.Order, info):
         if root.status == OrderStatus.DRAFT:
 
-            def _validate_draft_order(manager):
+            @allow_writer_in_context(info.context)
+            def _validate_draft_order(data):
+                lines, manager = data
                 country = get_order_country(root)
                 database_connection_name = get_database_connection_name(info.context)
                 try:
                     validate_draft_order(
-                        root,
-                        country,
-                        manager,
+                        order=root,
+                        lines=lines,
+                        country=country,
+                        manager=manager,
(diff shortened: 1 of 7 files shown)


saleor/graphql/order/types.py
```diff
@@ -1930,21 +1930,26 @@ def resolve_status_display(root: models.Order, _info):
     def resolve_can_finalize(root: models.Order, info):
         if root.status == OrderStatus.DRAFT:
 
-            def _validate_draft_order(manager):
+            @allow_writer_in_context(info.context)
```
fowczarek, line 1933: Why do we need a writer to validate the draft order?

maarcingebala: Some fields that are needed to validate the draft order are fetched ad-hoc, without dataloaders. So that's why for mutations, it has to be allowed in this way. The proper solution is to add dataloaders and introduce a similar structure to `CheckoutInfo` but for orders. I can open a separate task for this, as this requires quite a lot refactoring to be done right.

jakubkuc (approved):

fowczarek (approved):

PR #15777, opened 2024-04-08
Fix checkout base prices when tax app is broken

tomaszszymanski129:
Solves a case when one uses a tax app and that app fails with an error (network or response) and the checkout ended up in broken state.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/calculations.py
+++ b/saleor/checkout/calculations.py
@@ -254,7 +254,7 @@ def _fetch_checkout_prices_if_expired(
                 address,
             )
         except TaxEmptyData as e:
-            _get_checkout_base_prices(checkout, checkout_info, lines)
+            _set_checkout_base_prices(checkout, checkout_info, lines)
             checkout.tax_error = str(e)
 
         if not should_charge_tax:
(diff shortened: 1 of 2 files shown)


saleor/checkout/calculations.py
```diff
@@ -278,6 +279,7 @@ def _fetch_checkout_prices_if_expired(
                     address,
                 )
             except TaxEmptyData as e:
+                _get_checkout_base_prices(checkout, checkout_info, lines)
```
Air-t, line 282: this is nonblocking comment, but I'm wondering why this function is called `_get...` when it is rather a setter for a checkout?

Air-t (approved):

saleor/checkout/calculations.py
tomaszszymanski129: Yeah it triggered me as well, just forgot about it. I'll change it.

8r2y5 (approved):

PR #15795, opened 2024-04-15
Add transaction items limit to TransactionSessionBase

tomaszszymanski129:
I want to merge this change because it adds a limit of transaction items to `TransactionSessionBase`.
This results in adding the limit to the following two mutations:
- `PaymentGatewayInitialize`
- `TransactionInitialize`

⚠️ This is a port of https://github.com/saleor/saleor/pull/15740

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/payment/mutations/base.py
+++ b/saleor/graphql/payment/mutations/base.py
@@ -8,6 +8,7 @@ from ....checkout import models as checkout_models
 from ....checkout.calculations import fetch_checkout_data
 from ....checkout.fetch import fetch_checkout_info, fetch_checkout_lines
 from ....order import models as order_models
+from ...core.enums import TransactionInitializeErrorCode
 from ...core.mutations import BaseMutation
 from ...core.utils import from_global_id_or_error
 
@@ -18,6 +19,8 @@ if TYPE_CHECKING:
 class TransactionSessionBase(BaseMutation):
     class Meta:
         abstract = True
+
+    TRANSACTION_ITEMS_LIMIT = 100
 
     @classmethod
     def clean_source_object(
@@ -74,6 +77,18 @@ class TransactionSessionBase(BaseMutation):
                     )
                 }
             )
+
+        if source_object.payment_transactions.count() >= cls.TRANSACTION_ITEMS_LIMIT:
+            raise ValidationError(
+                {
+                    "transactions": ValidationError(
+                        f"{source_object_type} transactions limit of "
+                        f"{cls.TRANSACTION_ITEMS_LIMIT} reached.",
+                        code=TransactionInitializeErrorCode.INVALID.value,
+                    )
+                }
+            )
+
         return source_object
 
     @classmethod
--- a/saleor/graphql/payment/tests/mutations/test_transaction_initialize.py
+++ b/saleor/graphql/payment/tests/mutations/test_transaction_initialize.py
@@ -18,10 +18,12 @@ from .....payment.interface import (
     TransactionSessionData,
     TransactionSessionResult,
 )
+from .....payment.models import TransactionItem
 from ....channel.enums import TransactionFlowStrategyEnum
 from ....core.enums import TransactionInitializeErrorCode
 from ....core.utils import to_global_id_or_none
 from ....tests.utils import assert_no_permission, get_graphql_content
+from ...mutations.base import TransactionSessionBase
 
 TRANSACTION_INITIALIZE = """
 mutation TransactionInitialize(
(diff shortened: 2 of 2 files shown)


saleor/graphql/payment/tests/mutations/test_transaction_initialize.py
```diff
@@ -234,6 +236,39 @@ def test_for_checkout_without_payment_gateway_data(
...
 
 
+def test_for_checkout_transactions_limit(user_api_client, checkout_with_prices):
+    # given
+    TransactionItem.objects.bulk_create(
+        [
+            TransactionItem(
+                checkout=checkout_with_prices, currency=checkout_with_prices.currency
+            )
+            for _ in range(TransactionSessionBase.TRANSACTION_ITEMS_LIMIT)
```
korycins, line 246: Why you didn't mock the TRANSACTION_ITEMS_LIMIT to reduce a limit? This test will be slow, as you're creating the 100 Transactions. The test will have the same result when you change the limit to 2 transactions instead of 100.

saleor/graphql/payment/mutations/base.py
```diff
@@ -74,6 +77,18 @@ def clean_source_object(
                     )
                 }
             )
+
+        if source_object.payment_transactions.count() >= cls.TRANSACTION_ITEMS_LIMIT:
+            raise ValidationError(
+                {
+                    "transactions": ValidationError(
```
korycins, line 84: AFAIK the `field` should be the one used in the `input`. Based on the `field` value, storefront can identify incorrect field. You don't have `transactions` as an input. For `transactionInitialize` there is `id` field. It is checkout or order id. It sounds more reasonable to return the error with `id`, as the input ID is the one which caused the problem.

korycins (commented):
Did you update the docs to include the info about limit? What about mutation description?

saleor/graphql/payment/mutations/base.py
```diff
@@ -19,6 +20,8 @@ class TransactionSessionBase(BaseMutation):
     class Meta:
         abstract = True
 
+    TRANSACTION_ITEMS_LIMIT = 100
```
fowczarek, line 23: How about move it to `settings.py`?

saleor/graphql/payment/tests/mutations/test_transaction_initialize.py
fowczarek: When we move this variable to settings we could use `@override_settings` instead of mock.

saleor/graphql/payment/tests/mutations/test_transaction_initialize.py
```diff
@@ -234,6 +236,39 @@ def test_for_checkout_without_payment_gateway_data(
...
+    # then
+    content = get_graphql_content(response)
+    data = content["data"]["transactionInitialize"]
+    assert data["errors"]
+    error = data["errors"][0]
+    assert error["code"] == TransactionInitializeErrorCode.INVALID.name
+    assert error["message"] == (
+        "Checkout transactions limit of "
+        f"{TransactionSessionBase.TRANSACTION_ITEMS_LIMIT} reached."
+    )
```
fowczarek, line 269: You should also check `field`.

tomaszszymanski129:
@korycins Updated mutations descriptions, will prepare docs PR right after i'm done with current urgencies.

saleor/graphql/payment/mutations/transaction/payment_gateway_initialize.py
```diff
@@ -65,7 +65,10 @@ class Meta:
             "Initializes a payment gateway session. It triggers the webhook "
             "`PAYMENT_GATEWAY_INITIALIZE_SESSION`, to the requested `paymentGateways`. "
             "If `paymentGateways` is not provided, the webhook will be send to all "
-            "subscribed payment gateways." + ADDED_IN_313 + PREVIEW_FEATURE
+            "subscribed payment gateways. There is a limit of transaction items per"
+            "checkout / order, configurable in settings by TRANSACTION_ITEMS_LIMIT."
```
korycins, line 69: can't you use `f"{settings.TRANSACTION_ITEMS_LIMIT}"` here ? I don't think that writing that this value is configurable from settings will give frontend developer any hint.

korycins (approved):

fowczarek (dismissed):

IKarbowiak (approved):

fowczarek (approved):

PR #15845, opened 2024-04-22
Add SMTP config validation in UserEmailPlugin

8r2y5:
I want to merge this change because plugin was missing SMTP config validation.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


korycins (commented):
Have you tested it with mailtimp/mailhog/ another "real" email provider?

What was the reason for not using `saleor.plugins.email_common.validate_email_config`?

EDIT: I see that we call `validate_email_config` inside UserEmailPlugin. Do we call it in proper places, when activating the plugin? It looks like it should be responsible for not allowing an activation when we don't have all details 🤔

8r2y5:
> Have you tested it with mailtimp/mailhog/ another "real" email provider?
> 
> What was the reason for not using `saleor.plugins.email_common.validate_email_config`?
> 
> EDIT: I see that we call `validate_email_config` inside UserEmailPlugin. Do we call it in proper places, when activating the plugin? It looks like it should be responsible for not allowing an activation when we don't have all details 🤔

I tested it with "real" provider. Plugin can be activated only if config is valid (all required fields are provided) and saleor is able to connect and authorize with mail provider.

Currently `saleor.plugins.email_common.validate_email_config` is still used to validate if with config values you can connect and authorize.
I added additional validation for config just for this plugin just to return nice error message instead of `Missing "use_tls" value`.

PR #15847, opened 2024-04-23
Optimize memory collection for plugin manager

maarcingebala:
During incident 168 we discovered that some instances of PluginsManager have reference cycles and are not collected by the garbage collector after each request. In cases where many channels are used (e.g. over 300), this leads to significant memory usage and eventually OOM worker kills.

Reference cycles were caused by properties on CheckoutInfo that are evaluated as lazy objects (`SimpleLazyObject`). When initialized, they were holding the reference to PluginsManager which was not collected after request execution. To fix this, these properties were refactored to not use lazy objects::
- `CheckoutInfo.all_shipping_methods` - changed into `functools.cached_property`, this field can call external webhook and should be evaluated only once. Fields that were needed to calculate it were moved to CheckoutInfo as class members. `update_delivery_method_lists_for_checkout_info` functions was changed to clean this property when any of the members needed to calculate it it changed.
- `CheckoutInfo.valid_pick_up_points` - changed into `functools.cached_property`, it only runs DB queries but they'll be cached, so we perform fewer in case the property is called again.
- `CheckoutInfo.delivery_method_info` - changed into `@property`, it uses cached property inside and doesn't need to be cached itself.

Additionally, some `__del__` methods were added to make sure that certain objects are cleared from memory on object finalization:
- `DataLoader.__del__`
- `BasePlugin.__del__`
- `PluginsManager.__del__`

Lastly, the GraphQLView has now "finally" block where we clear the dataloaders that were cached in the request.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/fetch.py
+++ b/saleor/checkout/fetch.py
@@ -475,7 +475,6 @@ def fetch_checkout_info(
     shipping_channel_listings: Optional[
         Iterable["ShippingMethodChannelListing"]
     ] = None,
-    fetch_delivery_methods=True,
     voucher: Optional["Voucher"] = None,
     voucher_code: Optional["VoucherCode"] = None,
 ) -> CheckoutInfo:
--- a/saleor/checkout/fetch.py
+++ b/saleor/checkout/fetch.py
@@ -1,7 +1,7 @@
 import itertools
 from collections.abc import Iterable
-from dataclasses import dataclass
-from functools import singledispatch
+from dataclasses import dataclass, field
+from functools import cached_property, singledispatch
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -11,7 +11,6 @@ from typing import (
 )
 from uuid import UUID
 
-from ..core.utils.lazyobjects import lazy_no_retry
 from ..discount import DiscountType, VoucherType
 from ..discount.interface import fetch_variant_rules_info, fetch_voucher_info
 from ..shipping.interface import ShippingMethodData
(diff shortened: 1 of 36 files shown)


przlada (approved):

saleor/tests/e2e/checkout/shipping/test_checkout_invalidate_price_based_shipping_on_voucher_add.py
```diff
@@ -0,0 +1,227 @@
...
+    )
+    assert order_data["status"] == "UNFULFILLED"
+    assert order_data["discounts"][0]["type"] == "VOUCHER"
+    assert order_data["discounts"][0]["value"] == voucher_amount
+    assert order_data["voucher"]["code"] == voucher_code
+    # assert order_data["total"]["gross"]["amount"] == subtotal_amount
+    assert order_data["deliveryMethod"]["id"] == shipping_method_id
+    assert order_data["shippingPrice"]["gross"]["amount"] == shipping_price
+    # order_line = order_data["lines"][0]
+    # assert order_line["unitPrice"]["gross"]["amount"] == float(product_variant_price)
```
IKarbowiak, line 227: 🚔

saleor/tests/e2e/checkout/shipping/test_checkout_invalidate_price_based_shipping_on_voucher_add.py
```diff
@@ -0,0 +1,227 @@
...
+    # Step 7 - Complete checkout.
+    order_data = checkout_complete(
+        e2e_not_logged_api_client,
+        checkout_id,
+    )
+    assert order_data["status"] == "UNFULFILLED"
+    assert order_data["discounts"][0]["type"] == "VOUCHER"
+    assert order_data["discounts"][0]["value"] == voucher_amount
+    assert order_data["voucher"]["code"] == voucher_code
+    # assert order_data["total"]["gross"]["amount"] == subtotal_amount
```
IKarbowiak, line 223: 🚔

IKarbowiak (approved):

saleor/tests/e2e/checkout/shipping/test_checkout_invalidate_price_based_shipping_on_voucher_add.py
maarcingebala: Oops, I forgot about these, thanks!

saleor/checkout/fetch.py
```diff
@@ -65,18 +64,84 @@ def get_promotion_discounts(self) -> list["CheckoutLineDiscount"]:
...
     billing_address: Optional["Address"]
     shipping_address: Optional["Address"]
-    delivery_method_info: "DeliveryMethodBase"
-    all_shipping_methods: list["ShippingMethodData"]
     tax_configuration: "TaxConfiguration"
-    valid_pick_up_points: list["Warehouse"]
+    lines: Iterable[CheckoutLineInfo]
+    shipping_channel_listings: list["ShippingMethodChannelListing"]
+    shipping_method: Optional["ShippingMethod"] = None
+    collection_point: Optional["Warehouse"] = None
```
fowczarek, line 77: Maybe? 
```suggestion
    _shipping_method: Optional["ShippingMethod"] = None
    _collection_point: Optional["Warehouse"] = None
```

saleor/tests/e2e/checkout/shipping/cassettes/test_use_external_shipping_methods_in_checkout/test_use_external_shipping_methods_in_checkout_core_1652.yaml
fowczarek, line 1: In this cassette, we have 2 requests. Is it expected for this test to perform 2 requesters?

fowczarek (approved):

korycins (commented):
@maarcingebala I don't remember the exact details, but I remember that when we were adding the webhook for shipping methods, we had some issues with the combination of sync shipping method webhook and a voucher. As a double-check I would suggest checking locally some cases with different types of vouchers.

saleor/checkout/fetch.py
```diff
@@ -428,109 +492,24 @@ def fetch_checkout_info(
...
         channel=channel,
         billing_address=checkout.billing_address,
         shipping_address=shipping_address,
-        delivery_method_info=delivery_method_info,
         tax_configuration=tax_configuration,
-        all_shipping_methods=[],
-        valid_pick_up_points=[],
+        lines=lines,
+        manager=manager,
+        shipping_channel_listings=list(shipping_channel_listings),
```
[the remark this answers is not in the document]
maarcingebala: When we construct CheckoutInfo in dataloaders we pass a list there. So this `list()` here just for consistency and for typing to match. If there is issue with large number of objects, that's a separate problem, not in the scope of this PR. Regarding second point, this code was run in `update_delivery_method_lists_for_checkout_info` before which was called immediately after checkout info was created. I just moved it but it's the same data that was set here before.

michalina-graczyk (approved):
QA approved.

saleor/tests/e2e/checkout/shipping/cassettes/test_use_external_shipping_methods_in_checkout/test_use_external_shipping_methods_in_checkout_core_1652.yaml
maarcingebala: I'll check that as it was moved from other PR, and open a separate ticket if needed.

saleor/checkout/fetch.py
```diff
@@ -65,18 +64,84 @@ def get_promotion_discounts(self) -> list["CheckoutLineDiscount"]:
...
     channel: "Channel"
     billing_address: Optional["Address"]
     shipping_address: Optional["Address"]
-    delivery_method_info: "DeliveryMethodBase"
-    all_shipping_methods: list["ShippingMethodData"]
     tax_configuration: "TaxConfiguration"
-    valid_pick_up_points: list["Warehouse"]
+    lines: Iterable[CheckoutLineInfo]
+    shipping_channel_listings: list["ShippingMethodChannelListing"]
+    shipping_method: Optional["ShippingMethod"] = None
```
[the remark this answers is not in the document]
maarcingebala: I think the type should be enough, also I moved it here from another place where it was called like that.

PR #15849, opened 2024-04-23
Fix race concurrency condition for completing checkout with the voucher that has a usage limit 

IKarbowiak:
Fix the race condition for checkout with the voucher that has a usage limit. 
Case:
- checkout with voucher that has usage limit set
- thread 1: checkout complete increases the voucher usage in pre-process part
- thread 2: checkout complete is trying to increase the voucher usage but the error is raised. The payment is refunded
- thread 1: finishing the checkout complete and returns order, that has refunded payment

To prevent such a situation, new `is_voucher_usage_increased` flag has been introduced on `Checkout` model. It marks if the voucher usage has been increased on not, and prevents rasing `NotApplicable` error for vouchers with the usage limit.

# Impact

- [x] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/checkout/complete_checkout.py
```diff
@@ -115,14 +115,50 @@ def _process_voucher_data_for_order(checkout_info: "CheckoutInfo") -> dict:
...
+
+
+@traced_atomic_transaction()
+def _release_checkout_voucher_usage(
+    checkout: "Checkout",
+    voucher_code: Optional["VoucherCode"],
+    voucher: Optional["Voucher"],
+    user_email: Optional[str],
+):
+    if not checkout.is_voucher_usage_increased or not voucher_code:
```
korycins, line 150: Shouldn't be a separate condition?
It will exit when checkout.is_voucher_usage_increased set to True, but no `voucher_code`. 
Shoudln't we have something like:
```python
if checkout.is_voucher_usage_increated:
    checkout.is_voucher_usage_increated = False
    checkout.save(..)
if not voucher_code:
    return

(rest of the logic)
```

korycins (approved):

saleor/checkout/complete_checkout.py
IKarbowiak: Good point, and I forgot to change the `is_voucher_usage_increased` when the voucher is removed from the checkout.

tomaszszymanski129 (approved):

Air-t (approved):

PR #15877, opened 2024-04-25
Valid address items extension.

zedzior:
I want to merge this change, because there is a need to accept some specific address values, which do not pass `i18naddress` package's validation.

The PR adds dedicated map, which will be used to replace input value with the one accepted by `i18naddress` library.

To discuss:
The map embedded in python module will be difficult to maintain. 

Issue: https://linear.app/saleor/issue/MERX-258/bug-location-codes-not-mapped-in-i18n-validation-failing-validation

Docs: https://github.com/saleor/saleor-docs/pull/1151

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/account/i18n.py
+++ b/saleor/account/i18n.py
@@ -202,9 +202,8 @@ class CountryAwareAddressForm(AddressForm):
         if custom_names := VALID_ADDRESS_EXTENSION_MAP.get(country_code):
             for field_name, mapping in custom_names.items():
                 actual_value = data.get(field_name, "").strip().lower()
-                mapping_adjusted = {k.strip().lower(): v for k, v in mapping.items()}
-                if actual_value in mapping_adjusted:
-                    data[field_name] = mapping_adjusted[actual_value]
+                if actual_value in mapping:
+                    data[field_name] = mapping[actual_value]
 
 
 def get_address_form_class(country_code):
--- a/saleor/account/i18n.py
+++ b/saleor/account/i18n.py
@@ -7,6 +7,7 @@ from django.forms import BoundField
 from django.forms.models import ModelFormMetaclass
 from django_countries import countries
 
+from .i18n_custom_names import custom_address_names_map
 from .models import Address
 from .validators import validate_possible_number
 from .widgets import DatalistTextWidget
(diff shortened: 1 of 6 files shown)


saleor/account/tests/test_account.py
```diff
@@ -312,3 +314,39 @@ def test_customers_show_staff_with_order(admin_user, channel_USD):
         channel=channel_USD,
     )
     assert User.objects.customers().count() == 1
+
+
+@pytest.mark.parametrize(
+    ("country_area_input", "country_area_output", "is_valid"),
+    [
+        ("Dublin", "Co. Dublin", True),
```
korycins, line 322: How will it work when I provide `dublin`? And how it was behaving previously when I was providing `co. dublin`?  (for both action, the result should be the same)

zedzior: Good remark, default validator does strip() and lower():
```
def _match_choices(value, choices):
    if value:
        value = value.strip().lower()
```
Let me quickly handle it.

saleor/account/i18n.py
```diff
@@ -191,6 +193,19 @@ def clean(self):
...
+    @staticmethod
+    def substitute_invalid_values(data):
+        """Map invalid address names to the values accepted by i18naddress package."""
+        country_code = data["country_code"]
+        if not country_code:
+            return
+        if custom_names := VALID_ADDRESS_EXTENSION_MAP.get(country_code):
+            for field_name, mapping in custom_names.items():
+                actual_value = data.get(field_name, "").strip().lower()
+                mapping_adjusted = {k.strip().lower(): v for k, v in mapping.items()}
```
korycins, line 205: Correct me if I am wrong, but `mapping` is the value that you fetched from `VALID_ADDRESS_EXTENSION_MAP`. Do we need to always make `strip().lower()` for the map declared by us? Can't we make this once, when initializing the map, instead of doing it always when validating any address?

korycins (commented):
I think, we should mention this change in docs for address-validation part: https://docs.saleor.io/docs/3.x/developer/address

8r2y5 (approved):

korycins (approved):

PR #15935, opened 2024-05-07
Add more fields to external shipping method payload parsing

jakubkuc:
I want to merge this change because it adds more fields to the external shipping method response parser.  
Most important part is adding metadata to the response 

**EDIT** 
Link to task for visibility 
https://linear.app/saleor/issue/SUPP-82/feature-request-adding-more-fields-to-shipping-webhooks-response

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/webhook/transport/shipping.py
```diff
@@ -64,13 +64,19 @@ def parse_list_shipping_methods_response(
...
 
         shipping_methods.append(
             ShippingMethodData(
                 id=to_shipping_app_id(app, method_id),
                 name=method_name,
                 price=Money(method_amount, method_currency),
                 maximum_delivery_days=method_maximum_delivery_days,
+                minimum_delivery_days=method_minimum_delivery_days,
+                description=method_description,
+                metadata=method_metadata,
```
maarcingebala, line 79: There should be some validation of metadata to check if what the app returned follows the format of metadata that Saleor uses. An app could return anything here e.g a null, string, or list. We need to make sure that we handle such cases. Also, I would add tests for resolvers that try to render the GraphQL response with external shipping methods, where app returns both valid and invalid structures as metadata (and other fields).

jakubkuc: I assume that when invalid structure is returned by the app it should be skipped by the parser and returned as `None` right?

maarcingebala: I think so, yes. Maybe we could also log a warning in such case, with name of the field that we failed to parsed and the app name.

jakubkuc:
potentially unlimited data that can be in metadata response is not trivial and there are some countermeasures that needs to be taken that's why it take some extra time - it needs to be internally discussed.

IKarbowiak (approved):

IKarbowiak (approved):

maarcingebala (approved):

PR #15975, opened 2024-05-13
Unassign shipping address on CheckoutDeliveryMethodUpdate with C&C

tomaszszymanski129:
I want to merge this change because it unassigns the checkout address after delivery method is changed from C&C to standard delivery method.

more context: https://linear.app/saleor/issue/SHOPX-568/switching-from-clickandcollect-to-shipping-automatically-pre-fills-the

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
+++ b/saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
@@ -104,6 +104,11 @@ class CheckoutDeliveryMethodUpdate(BaseMutation):
             checkout_info, lines, shipping_method=delivery_method, collection_point=None
         )
 
+        # Clear checkout shipping address if it was C&C before.
+        if checkout.collection_point_id:
+            checkout.shipping_address = None
+            checkout.save(update_fields=["shipping_address"])
+
         cls._update_delivery_method(
             manager,
             checkout_info,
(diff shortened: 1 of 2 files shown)


saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
```diff
@@ -104,6 +104,11 @@ def perform_on_shipping_method(
             checkout_info, lines, shipping_method=delivery_method, collection_point=None
         )
 
+        # Clear checkout shipping address if it was C&C before.
+        if checkout.collection_point_id:
+            checkout.shipping_address = None
+            checkout.save(update_fields=["shipping_address"])
+
```
IKarbowiak, line 111: Seems like we can move it at the beginning of `_update_delivery_method` to avoid the additional call to DB for saving the checkout.

aniav (approved):

Air-t (approved):

saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
```diff
@@ -228,6 +228,10 @@ def _update_delivery_method(
     ) -> None:
         checkout_fields_to_update = ["shipping_method", "collection_point"]
         checkout = checkout_info.checkout
+        # Clear checkout shipping address if it was switched from C&C.
+        if checkout.collection_point_id and not collection_point:
```
IKarbowiak, line 232: What if the collection point is changed to not CC with `ALL_WAREHOUSES` for click and collect option? We will end up with updated CC, but `shipping_address` for the previous `LOCAL_STOCK`. 
We should add a test for this case as well.

IKarbowiak (approved):

szczecha (approved):

PR #15995, opened 2024-05-14
Handle a celery task error when the job_id for an AppInstall is missing in the db

aniav:
I want to merge this change because it's adding a logger when celery would have a task crash on rare cases where `job_id` is missing.

# Pull Request Checklist

### DIFF
--- a/saleor/app/tasks.py
+++ b/saleor/app/tasks.py
@@ -18,7 +18,13 @@ logger = logging.getLogger(__name__)
 
 @celeryconf.app.task
 def install_app_task(job_id, activate=False):
-    app_installation = AppInstallation.objects.get(id=job_id)
+    try:
+        app_installation = AppInstallation.objects.get(id=job_id)
+    except AppInstallation.DoesNotExist:
+        logger.warning(
+            "Failed to install app. AppInstallation not found for job_id: %s.", job_id
+        )
+        return
     try:
         app, _ = install_app(app_installation, activate=activate)
         app_installation.delete()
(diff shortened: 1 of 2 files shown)


saleor/app/tests/test_app_tasks.py
```diff
@@ -23,6 +23,16 @@ def test_install_app_task(app_installation):
     assert app.is_installed
 
 
+@pytest.mark.vcr
+def test_install_app_task_job_id_does_not_exist():
+    # given & when
+    nonexistent_job_id = 5435435345
+    install_app_task(nonexistent_job_id, activate=True)
```
tomaszszymanski129, line 30: we could possibly mock the logger and check if it was called

saleor/app/tasks.py
```diff
@@ -18,7 +18,13 @@
 
 @celeryconf.app.task
 def install_app_task(job_id, activate=False):
-    app_installation = AppInstallation.objects.get(id=job_id)
+    try:
+        app_installation = AppInstallation.objects.get(id=job_id)
+    except AppInstallation.DoesNotExist:
```
tomaszszymanski129, line 23: Do we know a case when it happens (what causes it)?

aniav: Our test database in Sentry gets cleared every now and then and it causes Celery to crash. Shouldn't normally happen on production environments but wanted to clear this so Sentry doesn't show it as unhandled 😁

tomaszszymanski129 (approved):

korycins (approved):

kadewu (approved):

PR #15996, opened 2024-05-15
Fix TypeError when availableActions null is provided to transactionCreate

tomaszszymanski129:
I want to merge this change because it fixes `TypeError` on mutation `transactionCreate` if one passed `availableActions` as `null` in variables.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/payment/mutations.py
+++ b/saleor/graphql/payment/mutations.py
@@ -954,6 +954,9 @@ class TransactionCreate(BaseMutation):
                 if "cancel" not in available_actions:
                     available_actions.append("cancel")
                 transaction["available_actions"] = available_actions
+        elif "available_actions" in transaction:
+            # null available_actions is provided, removing it from keys
+            transaction.pop("available_actions")
         return instance
 
     @classmethod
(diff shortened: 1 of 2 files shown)


saleor/graphql/payment/mutations.py
```diff
@@ -954,6 +954,9 @@ def validate_input(
                 if "cancel" not in available_actions:
                     available_actions.append("cancel")
                 transaction["available_actions"] = available_actions
+        elif "available_actions" in transaction:
```
aniav, line 957: I think it's a bit hard to follow the code because you have to wrap your head around what the previous if is doing and later that elif implicitly makes a statement that if the above one has items this one doesn't even if the key is there.

Maybe we could change it to:
```python
available_actions = transaction.get("available_actions")

if available actions:
...
else:
...
```

aniav:
@tomaszszymanski129 AFAIK based on Sentry this is happening in 3.20 only, so no need to port to 3.14.

PR #15997, opened 2024-05-15
Fix order from draft confirmation

8r2y5:
I want to merge this change because it fixes issue with draft orders being confirmed regardless of value in `automatically_confirm_all_new_orders`

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/tests/mutations/test_draft_order_complete.py
+++ b/saleor/graphql/order/tests/mutations/test_draft_order_complete.py
@@ -167,18 +167,9 @@ def test_draft_order_complete_no_automatically_confirm_all_new_orders(
         assert allocation.quantity_allocated == line.quantity_unfulfilled
 
     # ensure there are only 1 event with correct type
-    event_params = {
-        "user": staff_user,
-        "type__in": [
-            order_events.OrderEvents.PLACED_FROM_DRAFT,
-            order_events.OrderEvents.CONFIRMED,
-        ],
-        "parameters": {},
-    }
-    matching_events = OrderEvent.objects.filter(**event_params)
-    assert matching_events.count() == 1
-    assert matching_events[0].type == order_events.OrderEvents.PLACED_FROM_DRAFT
-    assert not OrderEvent.objects.exclude(**event_params).exists()
+    event = OrderEvent.objects.get(user=staff_user)
+    assert event.type == order_events.OrderEvents.PLACED_FROM_DRAFT
+    assert not OrderEvent.objects.exclude(user=staff_user).exists()
 
 
 def test_draft_order_complete_by_user_no_channel_access(
--- a/saleor/graphql/order/tests/mutations/test_draft_order_complete.py
+++ b/saleor/graphql/order/tests/mutations/test_draft_order_complete.py
@@ -129,6 +129,58 @@ def test_draft_order_complete(
     )
 
 
+def test_draft_order_complete_no_automatically_confirm_all_new_orders(
+    staff_api_client,
+    permission_group_manage_orders,
+    staff_user,
+    draft_order,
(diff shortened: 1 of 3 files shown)


saleor/order/actions.py
```diff
@@ -120,7 +120,7 @@ def order_created(
         )
 
     channel = order_info.channel
-    if channel.automatically_confirm_all_new_orders or from_draft:
+    if channel.automatically_confirm_all_new_orders and from_draft:
```
andrzejewsky, line 123: does this mean that we only confirm the order when the flag is enabled AND it is a draft? shouldn't be: `if channel.automatically_confirm_all_new_orders:` ? to confirm in both cases (regular order and draft then the flag is enabled)?

saleor/order/actions.py
```diff
@@ -120,7 +120,7 @@ def order_created(
         )
 
     channel = order_info.channel
-    if channel.automatically_confirm_all_new_orders or from_draft:
+    if channel.automatically_confirm_all_new_orders:
```
kadewu, line 123: Did you maybe thought about adding additional flag to cover that behavior of draft orders?
I think it makes more sense to not have, but just leaving a note.

Anyway we are missing the changelog for that change and it should land there I believe, and we should update graphql description for that change.

saleor/order/tests/test_actions.py
```diff
@@ -0,0 +1,41 @@
...
+        channel=order.channel,
+        payment=order.get_last_payment(),
+        lines_data=[],
+    )
+
+    # when
+    order_created(order_info, user=customer_user, app=None, manager=plugins_manager)
+
+    # then
+    if flag is True:
```
kadewu, line 36: It should be splited into multiple tests instead of using if in test.

zedzior (approved):

saleor/graphql/order/tests/mutations/test_draft_order_complete.py
```diff
@@ -129,6 +129,58 @@ def test_draft_order_complete(
...
+        "type__in": [
+            order_events.OrderEvents.PLACED_FROM_DRAFT,
+            order_events.OrderEvents.CONFIRMED,
+        ],
+        "parameters": {},
+    }
+    matching_events = OrderEvent.objects.filter(**event_params)
+    assert matching_events.count() == 1
+    assert matching_events[0].type == order_events.OrderEvents.PLACED_FROM_DRAFT
+    assert not OrderEvent.objects.exclude(**event_params).exists()
```
korycins, line 181: Based on `.filter` and `.exclude` parts I assume, that you expect a single order event. In that case, why not:
```python
assert OrderEvent.objects.count() == 1
event = OrderEvent.objects.first()
assert event.type = PLACED_FROM_DRAFT
...
```
This would be more meaningful IMO. Also, you don't need inputs like `type__in`

korycins: @aniav But we're talking about unit-tests. I believe that we prefer better readability here, instead of something that would make a single query. But if we want to reduce the number of queries in tests, I think we can find a lot of more heavy places 😄

korycins (approved):

saleor/graphql/order/tests/mutations/test_draft_order_complete.py
aniav: I think it's easy to avoid two queries and still keep the code readable. We have multiple thousands of tests and we need to try and avoid stuff that's easy to avoid. `count()` and `first()` one after the other in tests is so common that I am thinking we should have a CI check failing if someone does it on purpose in tests 🤣

PR #16001, opened 2024-05-16
Allow query checkout for all users.

Air-t:
Allow for fetching checkouts or checkout if the channel is inactive with additional permission `HANDLE_PAYMENTS` for apps and staff users. 
Raises `PermissionDenied` error instead of returning `None` while the channel is inactive for the checkout or for checkouts query
Allow to query checkout if the user knows the ID of the checkout

Fixes: https://linear.app/saleor/issue/SHOPX-693/allow-fetching-all-checkouts-with-handle-payments-permission
Fixes: https://linear.app/saleor/issue/SHOPX-694/allow-fetching-checkout-if-id-is-known-even-if-checkout-was-assigned
Fixes: https://linear.app/saleor/issue/SHOPX-688/raise-an-error-in-checkout-query-when-the-permissions-are-missing-for

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/resolvers.py
+++ b/saleor/graphql/checkout/resolvers.py
@@ -1,5 +1,10 @@
 from ...checkout import models
-from ...permission.enums import AccountPermissions, CheckoutPermissions
+from ...core.exceptions import PermissionDenied
+from ...permission.enums import (
+    AccountPermissions,
+    CheckoutPermissions,
+    PaymentPermissions,
+)
 from ..core.context import get_database_connection_name
 from ..core.tracing import traced_resolver
 from ..core.utils import from_global_id_or_error
@@ -34,22 +39,13 @@ def resolve_checkout(info, token, id):
         .filter(token=token)
         .first()
     )
-
     if checkout is None:
         return None
-
-    # resolve checkout in active channel
+    # always return checkout for active channel
     if checkout.channel.is_active:
-        # resolve checkout for anonymous customer
-        if not checkout.user:
-            return checkout
+        return checkout
 
-        # resolve checkout for logged-in customer
-        user = info.context.user
-        if user and checkout.user == user:
-            return checkout
-
-    # resolve checkout for staff user
+    # resolve checkout for staff or app
     requester = get_user_or_app_from_context(info.context)
 
     if not requester:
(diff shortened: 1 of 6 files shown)


IKarbowiak (approved):

tomaszszymanski129 (approved):

saleor/graphql/checkout/schema.py
```diff
@@ -51,6 +55,7 @@ class CheckoutQueries(graphene.ObjectType):
             "query checkouts that belong to other users: "
             f"{CheckoutPermissions.MANAGE_CHECKOUTS.name}, "
             f"{AccountPermissions.IMPERSONATE_USER.name}. "
+            f"{PaymentPermissions.HANDLE_PAYMENTS.name}. "
```
korycins, line 58: With your changes, the description is not valid anymore. You can fetch any checkout. Permissions are required only when you want to fetch the checkout from in-active channel.

saleor/graphql/checkout/resolvers.py
```diff
@@ -34,30 +39,28 @@ def resolve_checkout(info, token, id):
...
-        if user and checkout.user == user:
-            return checkout
+        return checkout
 
-    # resolve checkout for staff user
+    # resolve checkout for staff or app
     requester = get_user_or_app_from_context(info.context)
 
     if not requester:
         return None
```
korycins, line 52: Why `None`? You already changed the way in line 63/60 to raise PermissionDenied.

saleor/graphql/checkout/schema.py
Air-t: Yes, so should I split the description into two parts - for active and inactive channel or leave it as it has no permission at all?

saleor/graphql/checkout/resolvers.py
Air-t: should we raise the same error or more specyfic for a case where the requestor is not found?

korycins: When requestor is not found, then it means that the request is anonymous, and try to fetch the details for the in-active channel (at least from what I understand 😅 )

saleor/graphql/checkout/schema.py
korycins: You require permissions for checkouts placed on in-active channels, so I would place this info in the description.

saleor/graphql/checkout/resolvers.py
Air-t: ok so throwing an PermissionDenied should be good option here

korycins (approved):

PR #16008, opened 2024-05-16
Skip address validation.

zedzior:
I want to merge this change, because there is a need to skip Saleor's default address validation.

This PR introduces `skipValidation` field in `AddressInput` type, with default value = `False`.

It adds `validation_skipped` to `Address` model, to indicate if validation has been omitted for particular address.

Validation skipping is allowed only for mutations specified in `SKIP_ADDRESS_VALIDATION_PERMISSION_MAP` with proper permissions.

This PR also refactors a bit `warehouseCreate`, `warehouseUpdate` and `customerBulkUpdate` mutations. They used to use `validate_address_form` instead of `validate_address` function, what was inconsistent among other mutations. 

Issue: https://linear.app/saleor/project/address-validation-skipping-flags-530aa688d867

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/graphql/account/i18n.py
```diff
@@ -5,8 +5,43 @@
...
+        AuthorizationFilters.AUTHENTICATED_APP,
+    ],
+    "checkoutShippingAddressUpdate": [
+        CheckoutPermissions.HANDLE_CHECKOUTS,
+        AuthorizationFilters.AUTHENTICATED_APP,
+    ],
+    "checkoutBillingAddressUpdate": [
+        CheckoutPermissions.HANDLE_CHECKOUTS,
+        AuthorizationFilters.AUTHENTICATED_APP,
+    ],
```
fowczarek, line 42: Should we add information about required permissions on those mutations when skipping validation?  
e.g. on `createWarehouse` we have information about the required permission. On the `checkoutCreate` there is no such information. 

image
image

korycins (approved):

saleor/graphql/account/i18n.py
zedzior: Thanks for comment. I am not sure if we should change the description of whole `checkoutCreate` mutation, just to  describe behaviour of single flag. Maybe we should somehow extend the description of `AddressInput.skip_validation` field? Any idea how it should looks like?

fowczarek: I don't have any idea right now. I'm not sure about adding this info to `AddressInput` other mutations have complete information about required permission in their own description not in input descriptions.

8r2y5 (approved):

saleor/graphql/checkout/mutations/checkout_complete.py
```diff
@@ -191,14 +191,15 @@ def validate_checkout_addresses(
...
                 shipping_address_data = shipping_address.as_data()
-                cls.validate_address(
-                    shipping_address_data,
-                    address_type=AddressType.SHIPPING,
-                    format_check=True,
-                    required_check=True,
-                    enable_normalization=True,
-                    instance=shipping_address,
-                )
+                if not shipping_address.validation_skipped:
```
korycins, line 194: What about case:
- Update of checkout's address with proper permission and validation_skipped = True
- Then update of the checkout's address but without proper permission. 
Will we make a validation on checkoutComplete in that case or not?

saleor/graphql/account/types.py
```diff
@@ -91,12 +96,24 @@ class AddressInput(BaseInputObjectType):
...
         description="Address public metadata." + ADDED_IN_315,
         required=False,
     )
+    skip_validation = graphene.Boolean(
+        description=(
+            "Determine if the address should be validated. "
+            "By default, Saleor accepts only address inputs matching ruleset from "
+            "[Google Address Data]{https://chromium-i18n.appspot.com/ssl-address), "
+            "using [i18naddress](https://github.com/mirumee/google-i18n-address) "
+            "library."
```
korycins, line 110: I thing that would be great to have here an additional hint, that for using the field, additional permission could be required, and each mutation that allows you to skip mutation will have more info about required permission. 
WDYT?

saleor/graphql/account/i18n.py
zedzior: The info added to `createMutation` description.

saleor/graphql/account/types.py
zedzior: I enriched the description as suggested.

PR #16033, opened 2024-05-22
Extend OrderLine with is_price_overridden field and update unit_discount_type field

IKarbowiak:
- Update `OrderLine.unit_discount_type` - make it nullable
- Etend `OrderLine` type and model with `is_price_overridden` (it's `False` by default but it will be False for old orders that might have overridden prices so maybe it should be nullable, and `None` for old instances)

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


kadewu (approved):
Just a comment, the idea of doing NULL seems better than having incorrect value.

8r2y5 (approved):

jakubkuc (approved):

kadewu (approved):

tomaszszymanski129 (approved):

Air-t (approved):

PR #16035, opened 2024-05-23
Fix KeyError when event_type was not present in the event dict.

Air-t:
I want to merge this change because it fixes the issue when `TriggerWebhook` validation can't find the proper event.
It also will throw an error now when someone will try to trigger `sync` webhook with that mutation.

Fixes: https://linear.app/saleor/issue/SHOPX-572/webhooktrigger-mutation-crashes

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/webhook/mutations/webhook_trigger.py
+++ b/saleor/graphql/webhook/mutations/webhook_trigger.py
@@ -17,7 +17,7 @@ from ...core.mutations import BaseMutation
 from ...core.types.common import WebhookTriggerError
 from ...core.utils import from_global_id_or_error, raise_validation_error
 from ..subscription_query import SubscriptionQuery
-from ..subscription_types import WEBHOOK_TYPES_MAP
+from ..subscription_types import ASYNC_WEBHOOK_TYPES_MAP
 from ..types import EventDelivery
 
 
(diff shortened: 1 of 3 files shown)


saleor/graphql/webhook/mutations/webhook_trigger.py
```diff
@@ -84,6 +84,21 @@ def validate_event_type(cls, event_type, object_id):
                 code=WebhookTriggerErrorCode.INVALID_ID,
             )
 
+    @classmethod
+    def _get_async_event_or_raise_error(cls, event_type):
+        event = ASYNC_WEBHOOK_TYPES_MAP.get(event_type)
+        if event:
+            return event
+        event_name = event_type[0].upper() + to_camel_case(event_type)[1:]
```
aniav, line 92: This is going to trigger an `IndexError` if `event_type` is None. This case was covered in line 68 previously.

aniav (commented):
One small thing 🙌

saleor/graphql/webhook/mutations/webhook_trigger.py
Air-t: this validation is done after the grapql subscription query validation is done (the event name came from there) so there should be no case where we end with empty_event name, hovewer I will add a safe method to handle case when event_name is None or empty string

aniav (approved):

saleor/graphql/webhook/mutations/webhook_trigger.py
```diff
@@ -84,6 +84,23 @@ def validate_event_type(cls, event_type, object_id):
                 code=WebhookTriggerErrorCode.INVALID_ID,
             )
 
+    @classmethod
+    def _get_async_event_or_raise_error(cls, event_type):
+        event = ASYNC_WEBHOOK_TYPES_MAP.get(event_type)
+        if event:
+            return event
```
tomaszszymanski129, line 91:
```suggestion
        if event := ASYNC_WEBHOOK_TYPES_MAP.get(event_type):
            return event
```

tomaszszymanski129 (approved):

aniav (approved):

tomaszszymanski129 (approved):

IKarbowiak (approved):

PR #16048, opened 2024-05-27
Fix issue with voucher on draft order when there is no user associated with draft order

8r2y5:
I want to merge this change because it fixes issue with adding voucher to order that is not associated with user

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/discount/utils.py
+++ b/saleor/discount/utils.py
@@ -106,7 +106,7 @@ def add_voucher_usage_by_customer(
     code: "VoucherCode", customer_email: Optional[str]
 ) -> None:
     if not customer_email:
-        raise NotApplicable("This offer is only valid for signed-in users.")
+        raise NotApplicable("Unable to apply voucher as customer details are missing.")
 
     _, created = VoucherCustomer.objects.get_or_create(
         voucher_code=code, customer_email=customer_email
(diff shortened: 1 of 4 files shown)


yellowee (approved):

zedzior (approved):

saleor/discount/utils.py
```diff
@@ -102,7 +102,12 @@ def activate_voucher_code(code: "VoucherCode") -> None:
     code.save(update_fields=["is_active"])
 
 
-def add_voucher_usage_by_customer(code: "VoucherCode", customer_email: str) -> None:
+def add_voucher_usage_by_customer(
+    code: "VoucherCode", customer_email: Optional[str]
+) -> None:
+    if not customer_email:
+        raise NotApplicable("This offer is only valid for signed-in users.")
```
korycins, line 109: It doesn't look like a truth. 👇 
image
It looks like the offer is valid, also when you don't have an assigned order.user but you have a user_email.
So it is more like: you can't add voucher that counts a usage per customer, when you don't have user_email, or user assigned to the order. 
So the offer is also valid for non-signed-in users.

korycins (approved):

PR #16076, opened 2024-06-04
Improve validation for email plugins, add support for env var USER_EMAIL_URL

8r2y5:
I want to merge this change because it improves validation for email plugins. Also it adds support for email url for `UserEmailPlugin` using env variable `USER_EMAIL_URL`.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/settings.py
+++ b/saleor/settings.py
@@ -126,7 +126,6 @@ if not EMAIL_URL and SENDGRID_USERNAME and SENDGRID_PASSWORD:
     )
 
 email_config = dj_email_url.parse(EMAIL_URL or "")
-user_email_config = dj_email_url.parse(os.environ.get("USER_EMAIL_URL", ""))
 
 EMAIL_FILE_PATH: str = email_config.get("EMAIL_FILE_PATH", "")
 EMAIL_HOST_USER: str = email_config.get("EMAIL_HOST_USER", "")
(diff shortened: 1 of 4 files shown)


saleor/settings.py
```diff
@@ -138,6 +139,16 @@ def get_bool_from_env(name, default_value):
 
 ENABLE_SSL = get_bool_from_env("ENABLE_SSL", False)
 
+USER_EMAIL_HOST_USER: str = user_email_config.get("EMAIL_HOST_USER", "")
+USER_EMAIL_HOST_PASSWORD: str = user_email_config.get("EMAIL_HOST_PASSWORD", "")
+USER_EMAIL_HOST: str = user_email_config.get("EMAIL_HOST", "")
+USER_EMAIL_PORT = user_email_config.get("EMAIL_PORT", "")
+if USER_EMAIL_PORT:
+    USER_EMAIL_PORT = str(USER_EMAIL_PORT)
```
korycins, line 147: If you're making casting here, then probably would be good to also validate if the port is `int`.

korycins (approved):

saleor/settings.py
8r2y5: Such validation is done by django-email-url when parsing email url.
This here is just handling case where `EMAIL_PORT` value would be `None` and prevents it casting that into string.

cmiacz (approved):

zedzior (approved):

PR #16086, opened 2024-06-05
Generate sync webhook payloads in the dataloaders

fowczarek:
I want to merge this change because generating sync webhook payloads in the dataloaders. 

TODO:
- [x] Adjust flow to handle multiple webhooks flows 
- [x] create pre-payloads generated dataloader
- [x] Use created dataloader in all checkout resolvers with price recalcualtion
- [x] Cover flow with prices entered with taxes

In separate PR. 
- Cover draft orders flow.
- Cover flow for other sync webhooks called via resolvers 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/webhook/subscription_payload.py
+++ b/saleor/graphql/webhook/subscription_payload.py
@@ -81,7 +81,7 @@ def generate_payload_promise_from_subscription(
     subscribable_object: is an object which have a dedicated own type in Subscription
     definition.
     subscription_query: query used to prepare a payload via graphql engine.
-    context: A dummy request used to share context between apps in order to use
+    request: A dummy request used to share context between apps in order to use
     dataloaders benefits.
     app: the owner of the given payload. Required in case when webhook contains
     protected fields.
@@ -165,7 +165,7 @@ def generate_payload_from_subscription(
     subscribable_object: is an object which have a dedicated own type in Subscription
     definition.
     subscription_query: query used to prepare a payload via graphql engine.
-    context: A dummy request used to share context between apps in order to use
+    request: A dummy request used to share context between apps in order to use
     dataloaders benefits.
     app: the owner of the given payload. Required in case when webhook contains
     protected fields.
--- a/saleor/graphql/webhook/subscription_payload.py
+++ b/saleor/graphql/webhook/subscription_payload.py
@@ -109,7 +109,7 @@ def generate_payload_promise_from_subscription(
 
     def return_payload_promise(
         results, app_id=app_id, subscription_query=subscription_query
-    ) -> Union[Promise[Optional[dict[str, Any]]], Optional[dict[str, Any]]]:
+    ):
         if hasattr(results, "errors"):
             logger.warning(
                 "Unable to build a payload for subscription. \n"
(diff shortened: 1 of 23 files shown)


szczecha (dismissed):
I tested this PR with the Avatax app and different tax configurations.

The same query that causes the recursion error (get 100 checkouts with totalPrice while AvaTax App is configured) is no longer causing a recursion error. ❤️

I have explored the checkout and webhooks areas to ensure that nothing breaks.

Unfortunately, it doesn't solve the other issue of how we can observe recursion error (Taxes are 0 when calling updateDeliveryMethod while the AvaTax App is configured and there is a custom app with a checkout updated webhook with the totalPrice field). Still, this wasn't the main goal of this fix.

saleor/graphql/webhook/subscription_payload.py
```diff
@@ -65,6 +65,89 @@ def get_event_payload(event):
     return event
 
 
+def generate_payload_promise_from_subscription(
```
IKarbowiak, line 68: Cannot we merge this logic somehow with `generate_payload_from_subscription`, it seems similar?

saleor/graphql/webhook/subscription_payload.py
```diff
@@ -65,6 +65,89 @@ def get_event_payload(event):
...
+
+        if not payload:
+            logger.warning(
+                "Subscription did not return a payload.",
+                extra={"query": subscription_query, "app": app_id},
+            )
+            return None
+
+        payload_instance = payload[0]
+        event_payload = payload_instance.data.get("event")
```
IKarbowiak, line 132: `payload_instance.data` not always has `event` so you might get an error in line 138. 
Check my fix: https://github.com/saleor/saleor/pull/16333

saleor/graphql/webhook/subscription_payload.py
fowczarek: This logic was duplicated to handle the case when the executor returns `promise` instead of `payload` 
We could extract some similar parts to helpers but I'm afraid that decrease code readability. What do you think?

IKarbowiak: I'm not insist, I am only worried that some adjustments would be done in one place and not in another.

IKarbowiak (dismissed):

saleor/graphql/webhook/subscription_payload.py
```diff
@@ -65,6 +65,89 @@ def get_event_payload(event):
...
+        root=(event_type, subscribable_object),
+        context=get_context_value(request),
+        return_promise=True,
+    )
+
+    def return_payload_promise(
+        results, app_id=app_id, subscription_query=subscription_query
+    ) -> Union[Promise[Optional[dict[str, Any]]], Optional[dict[str, Any]]]:
+        if hasattr(results, "errors"):
+            logger.warning(
```
kadewu, line 114: Just noting the codecov complains here.

kadewu (approved):

IKarbowiak (dismissed):

kadewu (dismissed):

kadewu (approved):

maarcingebala (dismissed):

IKarbowiak (dismissed):

szczecha (approved):

szczecha (approved):

IKarbowiak (approved):

kadewu (approved):

maarcingebala (dismissed):

IKarbowiak (approved):

maarcingebala (approved):

kadewu (approved):

szczecha (approved):

PR #16091, opened 2024-06-06
Log invalid address fields.

zedzior:
I want to merge this change, because it:
1. Logs the names not recognised by our validation library, which can be used then to update `VALID_ADDRESS_EXTENSION_MAP`
2. Logs address ID when trying fetching tax data for address with validation skipped.
3. Set `validation_skipped` flag to `True` only if the address is invalid. Previously, the flag was set whenever input contained `skip_validation=True`, as we were omitting the validation entirely. This will also require to rename `validation_skipped` flag.

Issue: https://linear.app/saleor/issue/MERX-489/log-invalid-address-fields-when-skipping-the-validation
Part of: https://github.com/saleor/saleor/pull/16052

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/account/i18n.py
+++ b/saleor/account/i18n.py
@@ -1,3 +1,4 @@
+import logging
 from collections import defaultdict
 
 import i18naddress
@@ -13,6 +14,7 @@ from .widgets import DatalistTextWidget
 
 COUNTRY_FORMS = {}
 UNKNOWN_COUNTRIES = set()
+ADDRESS_FIELDS_TO_LOG = ["country_area", "city_area", "city", "postal_code"]
 
 AREA_TYPE = {
     "area": "Area",
@@ -36,6 +38,8 @@ AREA_TYPE = {
     "village_township": "Village/township",
     "zip": "ZIP code",
 }
+
+logger = logging.getLogger(__name__)
 
 
 class PossiblePhoneNumberFormField(forms.CharField):
@@ -185,11 +189,22 @@ class CountryAwareAddressForm(AddressForm):
                 del data["sorting_code"]
         except i18naddress.InvalidAddressError as exc:
             self.add_field_errors(exc.errors)
+            self.log_errors()
         return data
 
     def clean(self):
         data = super().clean()
         return self.validate_address(data)
+
+    def log_errors(self):
+        errors = self.errors
+        fields = {}
+        for field, _ in errors.items():
+            fields[field] = (
+                self.data.get(field) if field in ADDRESS_FIELDS_TO_LOG else "invalid"
+            )
+        fields["skip_validation"] = self.data.get("skip_validation")
+        logger.warning("Invalid address input: %s", fields)
 
 
 def get_address_form_class(country_code):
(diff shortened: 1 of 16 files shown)


saleor/plugins/avatax/__init__.py
```diff
@@ -652,6 +651,15 @@ def get_order_tax_data(
         data, f"order_{order.id}", config, force_refresh
     )
     if response and "error" in response:
+        msg = response.get("error", {}).get("message", "")
+        error_code = response.get("error", {}).get("code", "")
+        logger.warning(
+            "Unable to calculate taxes for order %s, error_code: %s, " "error_msg: %s",
```
8r2y5, line 657: This could be one string

saleor/account/i18n.py
```diff
@@ -185,12 +189,24 @@ def validate_address(self, data):
...
         return self.validate_address(data)
 
+    def log_errors(self):
+        errors = self.errors
+        fields = {}
+        for field, _ in errors.items():
+            fields[field] = (
+                self.data.get(field) if field in ADDRESS_FIELDS_TO_LOG else "invalid"
+            )
+        fields["skip_validation"] = self.data.get("skip_validation")
```
korycins, line 206: Can't you just add `skip_validation` and `country` to `ADDRESS_FIELDS_TO_LOG` ?

saleor/checkout/utils.py
```diff
@@ -1025,3 +1025,22 @@ def get_checkout_metadata(checkout: "Checkout"):
         return checkout.metadata_storage
     else:
         return CheckoutMetadata(checkout=checkout)
+
+
+def log_address_if_validation_skipped_for_checkout(
+    checkout_info: "CheckoutInfo", logger
+):
+    address = get_address_for_checkout_taxes(checkout_info)
+    if address and address.validation_skipped:
```
korycins, line 1034: Does it mean, that we will always log the address ID, when the validation was skipped? It means multiple warning log records for a single checkout. 
I think, that we should log the warning only when: validation is skipped and the address looks incorrect from our library perspective.

saleor/plugins/webhook/plugin.py
```diff
@@ -3049,6 +3051,7 @@ def get_taxes_for_checkout(
         self, checkout_info, lines, app_identifier, previous_value
     ) -> Optional["TaxData"]:
         event_type = WebhookEventSyncType.CHECKOUT_CALCULATE_TAXES
+        log_address_if_validation_skipped_for_checkout(checkout_info, logger)
```
korycins, line 3054: This means, multiple warning loggers. We should only log the address when:
- our library thinks that the address is incorrect, and we received the flag to skip the validation (input action)
- we send a tax webhook, but we don't receive a correct response **and** address has an incorrect format. ( see `TaxEmptyData` exception)

saleor/account/i18n.py
zedzior: No, because `skip_validation` and `country` fields will not contain any error. I want to log it always when at least single error occur

saleor/checkout/utils.py
zedzior: The name `validation_skipped` is misleading at the moment and I am working on renaming it in separate PR.  `validation_skipped` will be set to True only when address contains some field with invalid format.

8r2y5 (approved):

korycins (approved):

PR #16095, opened 2024-06-07
Throw a ValidationError on `CheckoutRemovePromoCode` mutation if code was not removed.

Air-t:
I want to merge this change because it throws a `ValidationError` on `CheckoutRemovePromoCode` mutation if the promo code cannot be removed from a `Checkout` object.

Fixes: https://linear.app/saleor/issue/SHOPX-761/bug-removing-non-existing-voucher-from-a-checkout-doesnt-return-error

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -808,12 +808,8 @@ def remove_promo_code_from_checkout_or_error(
         remove_gift_card_code_from_checkout_or_error(checkout_info.checkout, promo_code)
     else:
         raise ValidationError(
-            {
-                "promo_code": ValidationError(
-                    "Promo code does not exists.",
-                    code=CheckoutErrorCode.NOT_FOUND.value,
-                )
-            }
+            "Promo code does not exists.",
+            code=CheckoutErrorCode.NOT_FOUND.value,
         )
 
 
@@ -828,12 +824,8 @@ def remove_voucher_code_from_checkout_or_error(
         checkout_info.voucher = None
     else:
         raise ValidationError(
-            {
-                "promo_code": ValidationError(
-                    "Cannot remove a voucher not attached to this checkout.",
-                    code=CheckoutErrorCode.VOUCHER_NOT_APPLICABLE.value,
-                )
-            }
+            "Cannot remove a voucher not attached to this checkout.",
+            code=CheckoutErrorCode.VOUCHER_NOT_APPLICABLE.value,
         )
 
 
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -38,7 +38,7 @@ from ..discount.utils import (
 )
 from ..giftcard.utils import (
     add_gift_card_code_to_checkout,
-    remove_gift_card_code_from_checkout,
+    remove_gift_card_code_from_checkout_or_error,
 )
 from ..plugins.manager import PluginsManager
 from ..product import models as product_models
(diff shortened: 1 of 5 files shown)


saleor/graphql/checkout/mutations/checkout_remove_promo_code.py
```diff
@@ -84,31 +84,28 @@ def perform_mutation(
...
-            lines, _ = fetch_checkout_lines(checkout)
-            invalidate_checkout(
-                checkout_info,
-                lines,
-                manager,
-                recalculate_discount=True,
-                save=True,
-            )
-            cls.call_event(manager.checkout_updated, checkout)
+        # if this step is reached, it means promo code was removed
```
IKarbowiak, line 99: I'm not sure about that. There might be the case when the voucher code exists but it's not assigned to checkout, so we are not removing the promo code from checkout and not raising an error, right? In such case, we do not want to reach the below code.

saleor/giftcard/utils.py
```diff
@@ -53,16 +55,23 @@ def add_gift_card_code_to_checkout(
...
-    return False
+    else:
+        raise ValidationError(
+            {
+                "promo_code": ValidationError(
+                    "Cannot remove a gift card not attached to this checkout.",
+                    code=CheckoutErrorCode.GIFT_CARD_NOT_APPLICABLE.value,
+                )
+            }
+        )
```
IKarbowiak, line 74: You shouldn't wrap the `ValidationError` here with the field name reference, as it might be different for different mutations. You should catch this error in mutation and there add a additional layer with the input field name
```suggestion
    else:
        raise ValidationError(
              "Cannot remove a gift card not attached to this checkout.",
              code=CheckoutErrorCode.GIFT_CARD_NOT_APPLICABLE.value,
        )
```

saleor/checkout/utils.py
```diff
@@ -797,33 +797,44 @@ def add_voucher_to_checkout(
...
+        remove_gift_card_code_from_checkout_or_error(checkout_info.checkout, promo_code)
+    else:
+        raise ValidationError(
+            {
+                "promo_code": ValidationError(
+                    "Promo code does not exists.",
+                    code=CheckoutErrorCode.NOT_FOUND.value,
+                )
+            }
+        )
```
IKarbowiak, line 817: You shouldn't wrap the `ValidationError` here with the field name reference, as it might be different for different mutations. You should catch this error in mutation and there add a additional layer with the input field name
```suggestion
        raise ValidationError(
                "Promo code does not exists.",
                code=CheckoutErrorCode.NOT_FOUND.value,
        )
```

saleor/checkout/utils.py
```diff
@@ -797,33 +797,44 @@ def add_voucher_to_checkout(
...
-    return False
+    else:
+        raise ValidationError(
+            {
+                "promo_code": ValidationError(
+                    "Cannot remove a voucher not attached to this checkout.",
+                    code=CheckoutErrorCode.VOUCHER_NOT_APPLICABLE.value,
+                )
+            }
+        )
```
IKarbowiak, line 837: You shouldn't wrap the `ValidationError` here with the field name reference, as it might be different for different mutations. You should catch this error in mutation and there add a additional layer with the input field name
```suggestion
        raise ValidationError(
                "Cannot remove a voucher not attached to this checkout.",
                code=CheckoutErrorCode.VOUCHER_NOT_APPLICABLE.value,
        )
```

saleor/graphql/checkout/mutations/checkout_remove_promo_code.py
```diff
@@ -84,31 +84,28 @@ def perform_mutation(
...
-
         checkout = get_checkout(cls, info, checkout_id=checkout_id, token=token, id=id)
 
         manager = get_plugin_manager_promise(info.context).get()
         checkout_info = fetch_checkout_info(checkout, [], manager)
 
-        removed = False
         if promo_code:
-            removed = remove_promo_code_from_checkout(checkout_info, promo_code)
+            remove_promo_code_from_checkout_or_error(checkout_info, promo_code)
```
IKarbowiak, line 93: You should add `try, except` here and add a info about invalid field, like:
```suggestion
try:
    remove_promo_code_from_checkout_or_error(checkout_info, promo_code)
except ValidationError as error:
    raise ValidationError({
        "promo_code": error
    })
```

saleor/graphql/checkout/tests/mutations/test_checkout_remove_promo_code.py
```diff
@@ -104,18 +183,59 @@ def test_checkout_remove_gift_card_code(api_client, checkout_with_gift_card):
...
+    variables = {
+        "id": to_global_id_or_none(checkout_with_gift_card),
+        "promoCode": gift_card_with_metadata.code,
+    }
+
+    # when
+    data = _mutate_checkout_remove_promo_code(api_client, variables)
+
+    # then
+    assert data["errors"][0]["field"] is None
```
IKarbowiak, line 218: It shouldn't be None, but me suggestion from above, should fix that

saleor/graphql/checkout/tests/mutations/test_checkout_remove_promo_code.py
```diff
@@ -87,15 +155,26 @@ def test_checkout_remove_voucher_code_with_inactive_channel(
         "promoCode": checkout_with_voucher.voucher_code,
     }
 
+    # when
     data = _mutate_checkout_remove_promo_code(api_client, variables)
 
+    # then
+    assert data["errors"][0]["field"] is None
```
IKarbowiak, line 162: It shouldn't be None, but me suggestion from above, should fix that

saleor/graphql/checkout/tests/mutations/test_checkout_remove_promo_code.py
```diff
@@ -65,19 +71,81 @@ def test_checkout_remove_voucher_code(api_client, checkout_with_voucher):
...
+    variables = {
+        "id": to_global_id_or_none(checkout_with_voucher),
+        "promoCode": checkout_with_voucher.voucher_code,
+    }
+
+    # when
+    data = _mutate_checkout_remove_promo_code(api_client, variables)
+
+    # then
+    assert data["errors"][0]["field"] is None
```
IKarbowiak, line 105: It shouldn't be None, but me suggestion from above, should fix that

tomaszszymanski129 (approved):

IKarbowiak (approved):

PR #16109, opened 2024-06-10
Throw a ValidationError on `CheckoutRemovePromoCode` mutation.

Air-t:
It also fixes previously broken logic for `Voucher.code` to look just for the last `VoucherCode.code`. Now mutation checks if given `promo_code` or `promo_code_id` are included in possible `Voucher.codes` codes.

Port: https://github.com/saleor/saleor/pull/16095
Fixes: https://linear.app/saleor/issue/SHOPX-761/bug-removing-non-existing-voucher-from-a-checkout-doesnt-return-error
Fixes: https://linear.app/saleor/issue/SHOPX-865/issue-with-calling-checkoutremovepromocodewith-promocodeid

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -42,7 +42,7 @@ from ..discount.utils import (
 )
 from ..giftcard.utils import (
     add_gift_card_code_to_checkout,
-    remove_gift_card_code_from_checkout,
+    remove_gift_card_code_from_checkout_or_error,
 )
 from ..plugins.manager import PluginsManager
 from ..product import models as product_models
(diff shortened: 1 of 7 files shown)


IKarbowiak (approved):

saleor/checkout/utils.py
```diff
@@ -815,33 +815,36 @@ def add_voucher_to_checkout(
...
     checkout_info: "CheckoutInfo", voucher_code: str
-) -> bool:
-    """Remove voucher data from checkout by code.
+) -> None:
+    """Remove voucher data from checkout by code or raise an error."""
 
-    Return information whether promo code was removed.
-    """
     existing_voucher = checkout_info.voucher
     if existing_voucher and existing_voucher.code == voucher_code:
```
korycins, line 840: I think there is a bug here. Voucher can have multiple codes, as we moved them to separate model `VoucherCode(models.Model)`. Here, we're using `.code` property, which takes the `voucher.codes.last()`. This means that we are unable to remove the code which is returned on different place than the last one.

saleor/checkout/utils.py
```diff
@@ -815,33 +815,36 @@ def add_voucher_to_checkout(
...
         remove_voucher_from_checkout(checkout_info.checkout)
         checkout_info.voucher = None
-        return True
-    return False
+    else:
+        raise ValidationError(
+            "Cannot remove a voucher not attached to this checkout.",
+            code=CheckoutErrorCode.VOUCHER_NOT_APPLICABLE.value,
```
korycins, line 846: Are you sure that this is the correct code? You're removing the voucher, and we're returning "Voucher not applicable" - but we don't want to apply the voucher. We want to remove it. It is more like `invalid` IMO, as we receive `voucher_code` which doesn't match to the code assigned to the checkout.

saleor/giftcard/utils.py
```diff
@@ -53,16 +55,19 @@ def add_gift_card_code_to_checkout(
...
-    """
     if gift_card := checkout.gift_cards.filter(code=gift_card_code).first():
         checkout.gift_cards.remove(gift_card)
         checkout.save(update_fields=["last_change"])
-        return True
-    return False
+    else:
+        raise ValidationError(
+            "Cannot remove a gift card not attached to this checkout.",
+            code=CheckoutErrorCode.GIFT_CARD_NOT_APPLICABLE.value,
```
korycins, line 69: Same as with Voucher. Are you sure that the code is correct? We are not adding the GC, so it could be hard to understand why in case of removing the GC, we're returning NOT_APPLICABLE error code.

saleor/checkout/utils.py
Air-t: Now I think you may be right here and the `INVALID` code might be more adequate here.

saleor/checkout/utils.py
Air-t: Yes, it looks like a bug and it was not fixed after a separate model of `VoucherCode` was created

IKarbowiak (approved):

tomaszszymanski129 (approved):

PR #16114, opened 2024-06-10
Rename `Address.validation_skipped` field

zedzior:
I want to merge this change, because after changing address validation skipping logic, the name `validation_skipped` can be misleading. Now we want to store information if some address didn't pass one of validators - the format check.
Therefor I want to rename `validation_skipped` into `invalid_format`

Issue: https://linear.app/saleor/issue/MERX-545/rename-validation-skipped-flag

# Impact

- [x] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


patrys (changes requested):
I don't follow. This field correctly indicates that validation was skipped. Why do we need to rename it?

korycins:
> I don't follow. This field correctly indicates that validation was skipped. Why do we need to rename it?

@patrys , AFAIR `validation_skipped` was set to `True`, when you provided an input flag responsible for skipping the validation. No matter if the address was valid or not from Saleor perspective ([PR](https://github.com/saleor/saleor/pull/16008)). Now we want to track the addresses that failed the validation - it can happen when the flag is used, and our validation logic says that the **format** is invalid. With the[ new approach](https://github.com/saleor/saleor/pull/16052), the address input only turns off the format check. It doesn't impact normalization or required field check (mentioned these checks as they have their flags in CheckoutAddressValidationRules). So we assumed that it would be good to re-name the field in the model, to suggest that the address failed the format check validation. 

So in Address model we don't want to track which addresses had turned off validation, but only the ones that had turned off validation and Saleor's format validation would raise ValidationError for them.

Did we miss something?

patrys:
So you want to track which addresses have effectively bypassed validation. Why is `validation_skipped` such a bad name for those cases that it's worth changing the schema?

korycins:
> So you want to track which addresses have effectively bypassed validation. Why is `validation_skipped` such a bad name for those cases that it's worth changing the schema?

It is just unclear to me. I mean `validation_skipped` sounds to me that we track ALL addresses where we turned off the validation. But the real value that we want to keep in this field is: validation was skipped and address validation failed on Saleor side.

zedzior:
Decided to keep the current field name.

Issue #13634, opened 2023-08-07
Bug: Missing `issuingPrincipal` in `TRANSACTION_INITIALIZE_SESSION`, `TRANSACTION_PROCESS_SESSION` webhook requests


witoszekdev:
### What are you trying to achieve?

I'm trying to use `issuingPrincipal.id` in order to identify users making requests to `TRANSACTION_PROCESS_SESSION` and `TRANSACTION_INITIALIZE_SESSION` webhooks in order to tokenize payment methods.

### Steps to reproduce the problem

In order to replicate this issue, use Dummy Payment Server: https://github.com/witoszekdev/dummy-payment-server

1. Install the app in Saleor (as described in README)
2. Create checkout as logged-in user
3. Run `transactionInitialize` mutation as logged-in user

```graphql
mutation TransactionInitalize(
  $amount: PositiveDecimal
  $checkoutId: ID!
  $data: JSON
) {
  transactionInitialize(
    amount: $amount
    id: $checkoutId
    paymentGateway: { id: "witoszekdev.dummy-payment-app", data: $data }
  ) {
    transaction {
      id
      actions
      message
      pspReference
    }
    data
    errors {
      field
      message
      code
    }
  }
}
```

Variables:

```json
{
  "checkoutId": "<created checkout id>",
  "amount": 1
}
```

4. In the logs of the application, `issuingPrincipal` received from Saleor is set to `null`
5. Run `transactionProcess` mutation as logged-in user

```graphql
mutation TransactionProcess($transactionId: ID!) {
  transactionProcess(data: {final: true}, id: $transactionId) {
    transaction {
      id
    }
    transactionEvent {
      id
      createdAt
      pspReference
      message
      externalUrl
      amount {
        currency
        amount
      }
      type
    }
    data
    errors {
      field
      message
      code
    }
  }
}
```

Variables:

```json
{
  "transactionId": "<id from transactionInitialize>"
}
```
6. In the logs of application, `issuingPrincipal` is also set to `null`

### What did you expect to happen?

Saleor should return `issuingPrincipal`, which is `User` object in the described webhooks.

#### Transaction Initialize

**Subscription**

```graphql
subscription {
  event {
    ... on TransactionInitializeSession {
      issuedAt
      version
      issuingPrincipal {
        ... on Node {
          id
        }
        __typename
      }
      recipient {
        id
        name
      }
      sourceObject {
        ... on Checkout {
          user {
            id
          }
        }
        ...on Order {
          user {
            id
[...]

PR #16152, opened 2024-06-14
Fix issuingPrincipal field resolution

tomaszszymanski129:
I want to merge this change because it fixes resolution of `issuingPrincipal` field in subscription queries.
Internal issue: https://linear.app/saleor/issue/SHOPX-862/bug-issuingprincipal-field-is-null-in-response-from-subscription

Resolves https://github.com/saleor/saleor/issues/13634

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/plugins/webhook/tests/test_list_stored_payment_methods_webhook.py
+++ b/saleor/plugins/webhook/tests/test_list_stored_payment_methods_webhook.py
@@ -1,3 +1,4 @@
+import json
 from unittest import mock
 
 import graphene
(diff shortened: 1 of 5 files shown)


IKarbowiak (approved):

saleor/plugins/webhook/tests/test_list_stored_payment_methods_webhook.py
```diff
@@ -85,6 +89,100 @@ def test_list_stored_payment_methods_with_static_payload(
     )
 
 
+@mock.patch("saleor.webhook.transport.synchronous.transport.cache.get")
+@mock.patch("saleor.webhook.transport.synchronous.transport.send_webhook_request_sync")
+def test_list_stored_payment_methods_subscription_issuing_principal(
```
Air-t, line 94: Do we need a separate test for that or that is enough to extend a `subscription_query` for other tests within that file and not create a test that will test just a single field wdyt?

Air-t (approved):

saleor/plugins/webhook/tests/test_list_stored_payment_methods_webhook.py
tomaszszymanski129: If you look at [subscription_queries.py](https://github.com/saleor/saleor/pull/16152/files#diff-53f051aa33a53924f0f33a8922f2dbed773347264e94fa66c1cc6bd551af5020), you'll see these are already splitted like that anyway (mostly per particular event / group of functionally related events).

PR #16202, opened 2024-06-25
Add custom `DateTime` scalar

korycins:
I want to merge this change because it adds our own DateTime scalar. The graphene scalar doesn't validate if provided datetime, after converting to UTC time doesn't exceed the datetime's min, max values.
For example when using input like this one:
`"0001-01-01T00:00:01+07:00"`, it will be saved on DB as -> `0001-12-31 17:00:01+00 BC`. When fetching the details from DB and casting it to datetime instance, datetime module will raise an exception, as the value is lower than the datetime.min.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/types.py
+++ b/saleor/graphql/account/types.py
@@ -43,7 +43,7 @@ from ..core.doc_category import DOC_CATEGORY_USERS
 from ..core.enums import LanguageCodeEnum
 from ..core.federation import federated_entity, resolve_federation_references
 from ..core.fields import ConnectionField, PermissionsField
-from ..core.scalars import UUID
+from ..core.scalars import UUID, DateTime
 from ..core.tracing import traced_resolver
 from ..core.types import (
     BaseInputObjectType,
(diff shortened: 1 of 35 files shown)


IKarbowiak (approved):
Maybe we should extend `Date` scalar with this validation as well?

maarcingebala (approved):

korycins:
> Maybe we should extend `Date` scalar with this validation as well?

@IKarbowiak I don't believe this is a case when using `Date`. The issue here was related to providing Date+Time, when we convert the time to UTC format. It means that we can have `0001-01-01T00:00:01+07:00` that in UTC is: `datetime(0001-01-01T00:00:01) - timedelta(hours=7) == 0001-12-31 17:00:01+00 BC`. It means, that we will hit the min/max limits. In case of Date, you don't have timezones, that could have an impact on final value.

8r2y5 (approved):

jakubkuc (approved):

PR #16213, opened 2024-06-26
Fix catalogue discount application on custom price in checkout and draft orders

IKarbowiak:
- Fix applying catalogue discounts on checkout instances. The `priceOverride` is treated as overridden base price, so it should be the base price on which the catalogue discounts are applied
- Fix applying catalogue discount on draft order instances.

Port of https://github.com/saleor/saleor/pull/16191

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/discount/utils.py
+++ b/saleor/discount/utils.py
@@ -548,7 +548,9 @@ def prepare_line_discount_objects_for_catalogue_promotions(lines_info):
             continue
 
         # check if the line price is discounted by catalogue promotion
-        discounted_line = _is_discounted_line(line_info.channel_listing)
+        discounted_line = _is_discounted_line_by_catalogue_promotion(
+            line_info.channel_listing
+        )
 
         # delete all existing discounts if the line is not discounted or it is a gift
         if not discounted_line or line.is_gift:
@@ -600,7 +602,7 @@ def prepare_line_discount_objects_for_catalogue_promotions(lines_info):
     )
 
 
-def _is_discounted_line(
+def _is_discounted_line_by_catalogue_promotion(
     variant_channel_listing: "ProductVariantChannelListing",
 ) -> bool:
     """Return True when the price is discounted by catalogue promotion."""
--- a/saleor/discount/utils.py
+++ b/saleor/discount/utils.py
@@ -19,7 +19,7 @@ from ..checkout.base_calculations import (
     base_checkout_delivery_price,
     base_checkout_subtotal,
 )
-from ..checkout.models import Checkout
+from ..checkout.models import Checkout, CheckoutLine
 from ..core.db.connection import allow_writer
 from ..core.exceptions import InsufficientStock
 from ..core.taxes import zero_money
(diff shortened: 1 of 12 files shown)


saleor/discount/utils.py
```diff
@@ -610,20 +612,44 @@ def _get_discount_amount(
...
+            Money(line.price_override, channel.currency_code)
+            if line.price_override is not None
+            else None
+        )
+    else:
+        price_override = (
+            line.undiscounted_base_unit_price if line.is_price_overridden else None
+        )
+
+    if price_override:
```
korycins, line 645: Shouldn't be:
```suggestion
    if price_override is None:
```
The override price can be Decimal(0)

korycins: Having test for that case would be also good

saleor/order/calculations.py
```diff
@@ -525,6 +525,60 @@ def order_line_tax_rate(
...
+def order_line_unit_discount(
+    order: Order,
+    order_line: OrderLine,
+    manager: PluginsManager,
+    lines: Optional[Iterable[OrderLine]] = None,
+    force_update: bool = False,
+) -> Decimal:
+    """Return the line unit discount.
+
+    It takes into account all plugins.
```
korycins, line 537: Maybe would be worth extending the description to explicitly write, what exactly `unit_discount` means - like including vouchers, promotions etc. This is something that I would have to double-check in case of using the function.

saleor/discount/utils.py
```diff
@@ -599,9 +600,10 @@ def prepare_line_discount_objects_for_catalogue_promotions(lines_info):
     )
 
 
-def _get_discount_amount(
-    variant_channel_listing: "ProductVariantChannelListing", line_quantity: int
-) -> Decimal:
+def _is_discounted_line(
```
korycins, line 603: What do you think about extending the name to `_is_discounted_line_by_catalogue_promotion` or something similar that will give more context that this function checks only catalogue promotions. 
Just to be as explicit as possible. (also, I don't think that we need to port this to the previous version - as I saw that we already merged this to 3.19. It is only visual improvement)

saleor/order/calculations.py
```diff
@@ -525,6 +525,65 @@ def order_line_tax_rate(
...
+    manager: PluginsManager,
+    lines: Optional[Iterable[OrderLine]] = None,
+    force_update: bool = False,
+) -> Decimal:
+    """Return the line unit discount.
+
+        It takes into account all plugins.
+        If the prices are expired, call all order price calculation methods
+        and save them in the model directly.
+    g
```
korycins, line 540: Same comments as for: https://github.com/saleor/saleor/pull/16218#pullrequestreview-2145012256

korycins (approved):

tomaszszymanski129 (approved):

Air-t (approved):

PR #16217, opened 2024-06-27
Webhook payload storage cleanup

cmiacz:
Changes made to original PR:
- added `EventPayloadManager` with transaction.atomic
- added `app-x/webhook-x/` prefix to payload files
- implemented payload files cleaups

### DIFF
--- a/saleor/core/models.py
+++ b/saleor/core/models.py
@@ -7,6 +7,8 @@ from django.contrib.postgres.indexes import GinIndex
 from django.core.files.base import ContentFile
 from django.db import models, transaction
 from django.db.models import F, JSONField, Max, Q
+from django.utils.crypto import get_random_string
+from storages.utils import safe_join
 
 from . import EventDeliveryStatus, JobStatus, private_storage
 from .utils.json_serializer import CustomJsonEncoder
@@ -181,10 +183,10 @@ class EventPayload(models.Model):
         return self.payload
 
     def save_payload_file(self, payload_data: str):
-        self.payload_file.save(
-            f"payload-{self.pk}-{self.created_at}.json",
-            ContentFile(payload_data),
-        )
+        prefix = get_random_string(length=12)
+        file_name = f"payload-{self.pk}-{self.created_at}.json"
+        file_path = safe_join(prefix, file_name)
+        self.payload_file.save(file_path, ContentFile(payload_data))
 
 
 class EventDelivery(models.Model):
--- a/saleor/core/models.py
+++ b/saleor/core/models.py
@@ -1,8 +1,9 @@
 import datetime
-from typing import Any, TypeVar
+from typing import Any, Iterable, Sequence, TypeVar
 
 import pytz
 from django.contrib.postgres.indexes import GinIndex
+from django.core.files.base import ContentFile
 from django.db import models, transaction
 from django.db.models import F, JSONField, Max, Q
 
@@ -143,6 +144,26 @@ class Job(models.Model):
         abstract = True
 
 
+class EventPayloadManager(models.Manager["EventPayload"]):
+    @transaction.atomic
+    def create_with_payload_file(self, payload: str) -> "EventPayload":
+        obj = super().create()
+        obj.payload_file.save(
+            f"payload-{obj.pk}-{obj.created_at}.json",
+            ContentFile(payload),
+        )
+        return obj
+
+    @transaction.atomic
+    def bulk_create_with_payload_files(
+        self, objs: Iterable["EventPayload"], payloads=Iterable[str]
+    ) -> list["EventPayload"]:
+        created_objs = self.bulk_create(objs)
+        for obj, payload_data in zip(created_objs, payloads):
+            obj.save_payload_file(payload_data)
+        return created_objs
+
+
 class EventPayload(models.Model):
     payload = models.TextField(default="")
     payload_file = models.FileField(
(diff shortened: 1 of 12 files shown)


saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -196,7 +207,11 @@ def trigger_webhooks_async(
             raise NotImplementedError("No payload was provided for regular webhooks.")
 
         with allow_writer():
-            payload = EventPayload.objects.create(payload=data)
+            payload = EventPayload.objects.create()
```
korycins, line 210: Shouldn't be in `transaction.atomic`?
What will happen if `payload.payload_file.save(` raise an exception? It communicates with external storage so it could be possible, right?

saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -133,7 +137,14 @@ def create_deliveries_for_subscriptions(
         )
 
     with allow_writer():
-        EventPayload.objects.bulk_create(event_payloads)
+        created_event_payloads = EventPayload.objects.bulk_create(event_payloads)
+        for event_payload, payload_data in zip(
+            created_event_payloads, event_payloads_data
+        ):
+            event_payload.payload_file.save(
+                f"payload-{event_payload.pk}-{event_payload.created_at}.json",
```
fowczarek, line 145: I'm not sure about adding `created_at` to the file name. Why do we need it? I think we should remove it. `event_payload.pk` is unique and should be enough to identify the file.

saleor/webhook/transport/asynchronous/transport.py
fowczarek: If we age that transaction is required here we should also add it to similar code inside `create_deliveries_for_subscriptions` of this file and similar code inside `saleor/webhook/transport/synchronous/transport.py`

fowczarek (approved):

saleor/webhook/transport/asynchronous/transport.py
maarcingebala: Then also maybe we can add some small helper function as there will be three things:
- transaction
- create empty payload instance
- save file

maarcingebala: + also error handling in case the file save goes wrong, as Maciek pointed.

saleor/core/tasks.py
```diff
@@ -48,3 +55,21 @@ def delete_event_payloads_task(expiration_date=None):
 def delete_files_from_storage_task(paths):
     for path in paths:
         default_storage.delete(path)
+
+
+@app.task
```
korycins, line 60: See my comment above. If we want to delete files in celery task, this task should have retries (in case of an issue with connection with storage) and should take into account that the provided path could not exist (assuming a case, where we have 10 paths, we deleted 2 of them, then received an error. Celery task should retry and logic should be able to handle a case when path doesn't exist) (not sure if .delete will raise exception in that case or handle it silently).
Also IMO, in case all retires tries were not able to delete the files we should raise an exception with a proper message that would allow us to find these files manually.

saleor/core/models.py
```diff
@@ -143,20 +149,74 @@ class Meta:
...
+    @transaction.atomic
+    def bulk_create_with_payload_files(
+        self,
+        objs: Iterable["EventPayload"],
+        payloads: Iterable[str],
+        webhooks: Iterable["Webhook"],
+    ) -> list["EventPayload"]:
+        created_objs = self.bulk_create(objs)
+        for obj, payload_data, webhook in zip(created_objs, payloads, webhooks):
+            obj.save_payload_file(payload_data, webhook)
```
korycins, line 170: In case of fail to save the file, there is a possibility that we will not generate the proper webhook event. In that case, we can lose the data - for example updating fulfillment will happen but we will not trigger a proper event. 
So my question is:  Do we take this into account? How do we want to be sure that we will always send an event?
I see that you didn't delete the `payload` field from the model. In that case, are we still going to use them? Maybe we could treat a `payload` field as a fallback for storing the data?

saleor/core/tasks.py
```diff
@@ -37,8 +39,13 @@ def delete_event_payloads_task(expiration_date=None):
     qs = EventPayload.objects.filter(pk__in=ids)
     if ids:
         if expiration_date > timezone.now():
+            files_to_delete = [
```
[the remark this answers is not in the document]
cmiacz: Good point, but I'd prefer to make files deletion more bulletproof rather then blocking DB cleanup.

saleor/core/models.py
cmiacz: @maarcingebala What do you think. In my option it's good idea to use legacy payload field as fallback for potential storage issues.

cmiacz: In this case we can miss potential issues with storage config e.g. missing permissions. Also, legacy `payload` field is going to be deprecated and deleted in next Saleor versions.

fowczarek (approved):

saleor/core/tasks.py
cmiacz: @korycins I've checked Django storages - they ignore non-existing files quietly so there's no need to implement any exection handling here

saleor/core/tasks.py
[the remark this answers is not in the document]
cmiacz: https://github.com/saleor/saleor/pull/16217#discussion_r1679307016

saleor/core/models.py
```diff
@@ -143,20 +147,47 @@ class Meta:
...
     def get_payload(self):
         if self.payload_file:
             with self.payload_file.open("rt") as f:
                 payload = f.read()
                 return payload
         return self.payload
 
+    def save_payload_file(self, payload_data: str):
+        prefix = get_random_string(length=12)
+        file_name = f"{self.pk}.json"
```
cmiacz, line 187: @fowczarek I've shortened file name to make sure that we won't hit 256 KB message size limit while deleting payloads in batches by 1000

saleor/core/tasks.py
```diff
@@ -37,8 +38,13 @@ def delete_event_payloads_task(expiration_date=None):
...
+                event_payload.payload_file.name
+                for event_payload in qs
+                if event_payload.payload_file
+            ]
             qs.delete()
-            # TODO: delete payload files
+            delete_files_from_private_storage_task(files_to_delete)
```
tomaszszymanski129, line 47: shouldn't that call .delay()?

PR #16219, opened 2024-06-27
Implement login throttling

zedzior:
This PR implements login throttling. Details can be found in [RFC](https://github.com/saleor/saleor/discussions/16201)

### Flow

```mermaid
flowchart TD
    A("createToken")
    G("Increase attempts number for case A")
    I("Increase attempts number for case B")
    B("Lookup cache for blocked IP")
    C{"Is user blocked?"}
    D("Raise INVALID_CREDENTIALS error")
    E{"Log in attempt"}
    F("Return token")
    H{"Does username exist?"}
    L("Reset attempts")
    M("Calculate and block next attempt")
    Y("Raise LOGIN_ATTEMPT_DELAYED error")
    Z("Block next attempt by min value")

    A --> B --> C -- yes --> Y
    C -- no --> Z --> E
    E -- success --> L --> F
    E -- fail --> G --> H
    H -- yes --> I --> M
    H -- no --> M
    M --> D
```

# Impact

- [ ] New migrations

# Docs

- [x] Link to documentation: https://github.com/saleor/saleor-docs/pull/1212

# Pull Request Checklist

### DIFF
--- a/saleor/account/tests/test_throttling.py
+++ b/saleor/account/tests/test_throttling.py
@@ -54,6 +54,7 @@ def test_authenticate_successful_first_attempt(
     # given
     dummy_cache = {}
     setup_mock_for_cache(dummy_cache, mocked_cache)
+    now = timezone.now()
 
     ip = "123.123.123.123"
     request = rf.request(HTTP_X_FORWARDED_FOR=ip)
@@ -67,9 +68,17 @@ def test_authenticate_successful_first_attempt(
 
     # then
     assert user
+
     assert mocked_cache.get(block_key) is None
     assert mocked_cache.get(ip_key) is None
     assert mocked_cache.get(ip_user_key) is None
+
+    next_attempt = now + timedelta(seconds=MIN_DELAY)
+    mocked_cache.add.assert_called_once_with(block_key, next_attempt, timeout=MIN_DELAY)
+    mocked_cache.set.assert_not_called()
+    mocked_cache.incr.assert_not_called()
+    mock_delete_arg_set = {arg.args[0] for arg in mocked_cache.delete.call_args_list}
+    assert {block_key, ip_key, ip_user_key} == mock_delete_arg_set
 
 
 @freeze_time("2024-05-31 12:00:01")
@@ -81,6 +90,7 @@ def test_authenticate_successful_subsequent_attempt(
     # given
     dummy_cache = {}
     setup_mock_for_cache(dummy_cache, mocked_cache)
+    now = timezone.now()
 
     ip = "123.123.123.123"
     request = rf.request(HTTP_X_FORWARDED_FOR=ip)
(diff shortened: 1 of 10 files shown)


saleor/account/throttling.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+
+    ip_key = get_cache_key_failed_ip(ip)
+    block_key = get_cache_key_blocked_ip(ip)
+
+    # block the IP address before the next attempt to prevent concurrent requests
+    if add_block(block_key, MIN_DELAY) is False:
+        next_attempt_time = cache.get(block_key) or timezone.now() + timedelta(
+            seconds=MIN_DELAY
+        )
```
zedzior, line 38: Here is a short period, between ADD and GET, when we could display shorter login delay time. IMO it is irrelevant

zedzior:
The solution was tested with locust. I f you want to play around you can use this:
```python
from locust import HttpUser, task, between

from common.mutations.account import LOGIN_MUTATION
from locustfiles.base_cloud_behavior import BaseCloudBehavior

EXISTING_EMAIL = "***"
NON_EXISTING_EMAIL = "non-existing@saleor.io"
VALID_PASSWORD = "***"
INVALID_PASSWORD = "invalid-password"

class UserBehavior(BaseCloudBehavior):
    @task(2)
    def task_fail_login(self):
        response = self.post_graphql(
            self.client,
            LOGIN_MUTATION,
            variables={
                "email": EXISTING_EMAIL,
                "password": INVALID_PASSWORD,
            },
            name="FAIL_LOGIN",
        )
        print(response)

    @task(1)
    def task_success_login(self):
        return self.post_graphql(
            self.client,
            LOGIN_MUTATION,
            variables={
                "email": EXISTING_EMAIL,
                "password": VALID_PASSWORD,
            },
            name="SUCCESS_LOGIN",
        )


class WebsiteUser(HttpUser):
    tasks = [UserBehavior]
    host = "http://localhost:8000/"
    wait_time = between(0.1, 0.5)
  ```

File: saleor/account/tests/test_throttling.py [hunk 2]
```diff
@@ -0,0 +1,387 @@
...
+    block_key = get_cache_key_blocked_ip(ip)
+    ip_key = get_cache_key_failed_ip(ip)
+    ip_user_key = get_cache_key_failed_ip_with_user(ip, customer_user.id)
+
+    # when
+    user = authenticate_with_throttling(request, EXISTING_EMAIL, CORRECT_PASSWORD)
+
+    # then
+    assert user
+    assert mocked_cache.get(block_key) is None
```
korycins, line 70: If you expect that this values will not be set by your logic, then would be good to make an explicit assert:
`assert not mocked.set.called`
This will confirm that you didn't call any `set` in your logic. 
If you expect some `set` calls, but you want to check that the `set` was not called for `block_key`, you could do something like:
`assert block_key not in mocked_cache.set.call_args_list`

This makes your test more explicit, as I will already know what should happen with the cache. Currently I know that these fields are None, but I don't know if they are None, because you didn't call `.set` for them or because you call `set` with None values

saleor/account/throttling.py
```diff
@@ -0,0 +1,144 @@
...
+logger = logging.getLogger(__name__)
+
+
+def authenticate_with_throttling(request, email, password) -> Optional[models.User]:
+    ip = get_client_ip(request)
+    if not ip:
+        logger.warning("Unknown request's IP address.")
+        raise ValidationError(
+            {
+                "email": ValidationError(
```
korycins, line 28: Is it an error related to `email` field? Shouldn't be just 
```
raise ValidationError("msg", code=UNKNOWN_IP_ADDRESS.value)
```

saleor/account/throttling.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # block the IP address before the next attempt to prevent concurrent requests
+    if add_block(block_key, MIN_DELAY) is False:
+        next_attempt_time = cache.get(block_key) or timezone.now() + timedelta(
+            seconds=MIN_DELAY
+        )
+        raise ValidationError(
+            {
+                "email": ValidationError(
+                    f"Logging has been suspended till {next_attempt_time}.",
```
korycins, line 46: Same question as above. Is the error related to the multiple logins for the same email or related to blocked IP? If related to IP I would not "connect" it with `email` field. 
---
Additionally, would be good to provide in the message more info why we blocked the login process. Like to many requests from the same IP, or too many request for the same email etc

saleor/account/tests/test_throttling.py
8r2y5: Perhaps we could make this test more explicit by using `assert mocked_cache.mock_calls = [...]`? This would verify order of the calls and what was passed to each call thus any change to order of calls or passed args would make this test fail

zedzior: More assertions were added.

saleor/account/throttling.py
korycins: > Two concurrent login attempts
> Login blocked due to too many failed attempts

These two means in general too many attempts right? It can be too many attempts to log in for single user, or for single IP. So maybe adding too many attempts in general would be enough?
So something like:
`Too many attempts to log-in. Logging has been suspended till {next_attempt_time}`
🤔

korycins (approved):

8r2y5 (dismissed):

8r2y5 (approved):

8r2y5 (approved):

korycins (approved):

korycins (approved):

8r2y5 (approved):

PR #16242, opened 2024-07-03
Fix Order and click-and-collect address data migration

przlada:
This change addresses a bug in the data migration, which involves two separate issues:

1. Addresses are created using the `bulk_create` method with the `ignore_conflicts=True` argument. However, this prevents primary keys from being assigned to Address objects, making it impossible to assign them to Orders properly. We discussed that `ignore_conflicts=True` is unnecessary here and can be turned off, which fixes this issue.
2. Due to a bug in Django (fixed in version 4.1+), addresses need to be reassigned for the `bulk_update` of Orders to work correctly. More details can be found here: [https://code.djangoproject.com/ticket/33322](https://code.djangoproject.com/ticket/33322)

Related task [SHOPX-1002](https://linear.app/saleor/issue/SHOPX-1002)

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/order/migrations/0172_update_order_cc_addresses.py
+++ b/saleor/order/migrations/0172_update_order_cc_addresses.py
@@ -38,7 +38,11 @@ def update_order_addresses(apps, schema_editor):
                 order_address = Address(**model_to_dict(cc_address, exclude=["id"]))
                 order.shipping_address = order_address
                 addresses.append(order_address)
-        Address.objects.bulk_create(addresses, ignore_conflicts=True)
+        Address.objects.bulk_create(addresses, ignore_conflicts=False)
+        # Due to a bug in Django, addresses need to be reassigned for bulk_update of Orders to work correctly.
+        # For more information about the bug, see: https://code.djangoproject.com/ticket/33322
+        for order in orders:
+            order.shipping_address = order.shipping_address
         Order.objects.bulk_update(orders, ["shipping_address"])
 
 


saleor/order/migrations/0172_update_order_cc_addresses.py
```diff
@@ -38,7 +38,11 @@ def update_order_addresses(apps, schema_editor):
                 order_address = Address(**model_to_dict(cc_address, exclude=["id"]))
                 order.shipping_address = order_address
                 addresses.append(order_address)
-        Address.objects.bulk_create(addresses, ignore_conflicts=True)
+        Address.objects.bulk_create(addresses, ignore_conflicts=False)
+        # Due to a bug in Django, addresses need to be reassigned for bulk_update of Orders to work correctly.
+        # For more information about the bug, see: https://code.djangoproject.com/ticket/33322
+        for order in orders:
+            order.shipping_address = order.shipping_address
         Order.objects.bulk_update(orders, ["shipping_address"])
```
Air-t, line 47: Since this did not work previously on environments that already applied this migration, should we also provide migration that will fix this unlinked `shipping_addresses`?

IKarbowiak: Yes, we should check how many `Orders` are broken and prepare a fix if needed

aniav:
@przlada are you indending to address the comments here?

aniav:
I'll close this PR for now because it hasn't been active for some time. Feel free to reopen when you get to it again.

PR #16272, opened 2024-07-05
Prevent from installing two apps with the same `App.identifier`

Air-t:
Currently, there is no limit to apps with the same Manifest Identifier that can be installed. 
It's now too easy to install the same app twice. 
This leads to many bugs - e.g. payment or taxes apps are routed to specific apps via identifier (which, in case of a duplicate, will be chosen randomly)

A validation for `app.identifier` was added.

Fixes: https://linear.app/saleor/issue/SHOPX-114/prevent-installation-of-the-app-twice

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/app/mutations/app_install.py
+++ b/saleor/graphql/app/mutations/app_install.py
@@ -1,9 +1,7 @@
 import graphene
-from django.core.exceptions import ValidationError
 
 from ....app import models
 from ....app.manifest_validations import clean_manifest_url
-from ....app.models import App
 from ....app.tasks import install_app_task
 from ....permission.enums import AppPermission, get_permissions
 from ....webhook.event_types import WebhookEventAsyncType
@@ -64,17 +62,6 @@ class AppInstall(ModelMutation):
     def clean_input(cls, info, instance, data, **kwargs):
         manifest_url = data.get("manifest_url")
         clean_manifest_url(manifest_url)
-        breakpoint()
-        if app := App.objects.filter(manifest_url=manifest_url).first():
-            raise ValidationError(
-                {
-                    "manifest_url": ValidationError(
-                        "App with the same manifest_url is already installed: "
-                        f"{app.name}"
-                    )
-                }
-            )
-
         cleaned_input = super().clean_input(info, instance, data, **kwargs)
 
         # clean and prepare permissions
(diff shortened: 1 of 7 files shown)


saleor/graphql/app/mutations/app_install.py
```diff
@@ -63,6 +65,16 @@ def clean_input(cls, info, instance, data, **kwargs):
         manifest_url = data.get("manifest_url")
         clean_manifest_url(manifest_url)
 
+        if app := App.objects.filter(manifest_url=manifest_url).first():
```
korycins, line 68: You can have the same app deployed in different places. All of them will result in the same `app_identifier`. I don't think that we should rely on the uniqueness of `manifest_url`. It should be uniqness for `app.identifier`

korycins: Additionally: You can have local env, when you use the same URL localhost for serving different app's manifest (apps with different identifiers). This will cause, you will not be able to install them.

korycins: @Air-t  lets also change the title of the PR to match to the modified content :)

IKarbowiak (approved):

saleor/graphql/app/tests/mutations/cassettes/test_app_fetch_manifest/test_fetch_manifest_fail_when_app_with_same_identifier_already_installed.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - gzip, deflate, br
+      Connection:
+      - keep-alive
+      User-Agent:
+      - python-requests/2.31.0
+    method: GET
+    uri: http://localhost:3000/api/manifest
+  response:
+    body:
+      string: !!binary |
```
tomaszszymanski129, line 17: I see this is gzip encoded and I can decode it, but it would be great to just have JSON here, for reference to see how app response looks like and maintainability of such test.

tomaszszymanski129 (approved):

saleor/graphql/app/tests/mutations/cassettes/test_app_fetch_manifest/test_fetch_manifest_fail_when_app_with_same_identifier_already_installed.yaml
Air-t: I managed to fix it and left a hint in our notion tutorial.

PR #16311, opened 2024-07-12
Control webhook queue name based on webhook protocol

korycins:
I want to merge this change because it allows to extract pubsub/sqs webhook requests to separate celery queue.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


magul (approved):
I heve limited understanding of context here (I'm assuming that `queue` passed to trigger_webhooks_async will contain either `WEBHOOK_CELERY_QUEUE_NAME`, `CHECKOUT_WEBHOOK_EVENTS_CELERY_QUEUE_NAME` or `ORDER_WEBHOOK_EVENTS_CELERY_QUEUE_NAME`, so dealing with priority queue is done somewhere above in stack trace).

Generally this change will do the trick, but I personally would consider just calculating queue per `delivery` basis when triggering `send_webhook_request_async.apply_async` instead of creating dictionary in advace (I'm assuming that for particular `event_type` we don't have one to many relation between delivery and webhook there, so we're not avoiding calculating queue multiple time for webhook when not going with maps).

maarcingebala:
@korycins We will need this ported to 3.20 and 3.19 as well.

korycins:
> @korycins We will need this ported to 3.20 and 3.19 as well.

@maarcingebala Yes, I know :) I don't want to create a PR with ports until receiving full review.

saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -164,6 +164,17 @@ def create_event_delivery_list_for_webhooks(
...
 
+def get_queue_name_for_webhook(webhook, default_queue):
+    scheme = urlparse(webhook.target_url).scheme.lower()
+    if scheme == WebhookSchemes.AWS_SQS:
+        queue_name = settings.WEBHOOK_SQS_CELERY_QUEUE_NAME
+    elif scheme == WebhookSchemes.GOOGLE_CLOUD_PUBSUB:
+        queue_name = settings.WEBHOOK_PUBSUB_CELERY_QUEUE_NAME
+    else:
+        queue_name = default_queue
+    return queue_name
```
magul, line 175: This is pure nitpicking, but I don't know why we need to assign to `queue_name` and then return it. Why not:
```suggestion
def get_queue_name_for_webhook(webhook, default_queue):
    scheme = urlparse(webhook.target_url).scheme.lower()
    if scheme == WebhookSchemes.AWS_SQS:
       return settings.WEBHOOK_SQS_CELERY_QUEUE_NAME
    if scheme == WebhookSchemes.GOOGLE_CLOUD_PUBSUB:
        return settings.WEBHOOK_PUBSUB_CELERY_QUEUE_NAME
    return default_queue
```

magul: or even abandon if-else approach
```suggestion
def get_queue_name_for_webhook(webhook, default_queue):
    return {
        WebhookSchemes.AWS_SQS: settings.WEBHOOK_SQS_CELERY_QUEUE_NAME,
        WebhookSchemes.GOOGLE_CLOUD_PUBSUB: settings.WEBHOOK_PUBSUB_CELERY_QUEUE_NAME,
    }.get(
        urlparse(webhook.target_url).scheme.lower(),
        default_queue,
    )
```

magul (approved):

magul (approved):

IKarbowiak (approved):

fowczarek (approved):

Issue #15029, opened 2023-12-14
User-facing mutation to add a note to their checkout


khalibloo:
## Problem

The checkout model has a `note` field that could theoretically allow the user to supply additional notes to their order. Real world use cases could be extra delivery instructions or food allergies. However, there does not appear to be any mutation that allows the user to add/edit the note.

---
### Mutations

```graphql
mutation checkoutNoteUpdate (note: "blah blah blah") {
  errors
  checkout
}
```

timuric:
Hey @khalibloo, alternatively you could try to use metadata

khalibloo:
Thanks @timuric. I considered that too, but I'm not sure if metadata gets carried over when a checkout is converted into an order. I want the information to be visible in the admin dashboard when the order is being processed.

pitkes22:
Duplicate of #9591

maarcingebala:
Fixed in https://github.com/saleor/saleor/pull/16315.

timuric:
Thank you for your contribution @pitkes22 !

Issue #9591, opened 2022-04-22
Mutation for adding a note to the checkout


DidoMarchet:
### What I'm trying to achieve
Have a mutation for adding a note.
Nowadays there is no way to add a note to the checkout despite the GQL Schema have it.

### Describe a proposed solution
Create a mutation named `checkoutAddNote`

### Ref.
[https://github.com/saleor/saleor/discussions/9580#discussioncomment-2614909](https://github.com/saleor/saleor/discussions/9580#discussioncomment-2614909)

Thanks and regards, 

Davide

maarcingebala:
Fixed in https://github.com/saleor/saleor/pull/16315.

PR #16315, opened 2024-07-14
Add CheckoutCustomerNoteUpdate mutation

pitkes22:
I want to merge this change because there is currently no mutation that allows updating the `note` field on the checkout object, even though existing functionality copies this note to the `customer_note` field in the order object and displays it in the dashboard.

Requested by: #9580, #9591 and #15029

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


korycins (commented):
Thanks for the PR! <3 It looks great, left some comments. Could you also add `#given, #when and #then` in tests where this form is missing?

korycins (dismissed):

IKarbowiak (approved):

saleor/graphql/checkout/mutations/checkout_note_update.py
```diff
@@ -0,0 +1,53 @@
...
+
+class CheckoutNoteUpdate(BaseMutation):
+    checkout = graphene.Field(Checkout, description="An updated checkout.")
+
+    class Arguments:
+        id = graphene.ID(
+            description="The checkout's ID." + ADDED_IN_321,
+            required=False,
+        )
+        note = graphene.String(required=True, description="New note content.")
```
IKarbowiak, line 22: One more request, could you rename this input field with `customer_note`, we want to unify it with order mutations that are using `customerNote` 🙏

pitkes22: Do you think I should also rename the mutation to `checkoutCustomerNoteUpdate`? I Was thinking about this and the naming of the `note` fields is currently quite confusing. 

In checkout, there is `note` field, that gets renamed to `customer_note` when checkout is converted to the order.

At the same time, there are new `orderNoteAdd` and `orderNoteUpdate` mutation that updates the notes in the order's `events` list. And there is actually no mutation to update the `customer_note` on the order object.

I was thinking about opening a PR to also implementing mutation on the order object but I was not sure about the correct way to do this, because there is an `orderUpdate` mutation that I could add this field to, but at the same time I think you are recently trying to do more specific update mutations that update single thing on the object instead of one big update mutation.

IKarbowiak: @pitkes22 great points! Thank you <3

Yes, I think we can rename it to `checkoutCustomerNoteUpdate`, as it could be misleading with `orderNoteAdd` and `orderNoteUpdate` that are creating the staff notes.
And as you correctly noticed, we are also missing a mutation for customer note update on orders.

We are planning to deprecate `note` on `Checkout` type and add a `customerNote` field instead, to make it clear.

maarcingebala (dismissed):

IKarbowiak (approved):

saleor/graphql/checkout/mutations/checkout_customer_note_update.py
```diff
@@ -0,0 +1,57 @@
...
+from ..types import Checkout
+
+
+class CheckoutCustomerNoteUpdate(BaseMutation):
+    checkout = graphene.Field(Checkout, description="An updated checkout.")
+
+    class Arguments:
+        id = graphene.ID(
+            description="The checkout's ID." + ADDED_IN_321,
+            required=False,
```
korycins, line 20: As you support only `id` field, then this should be `required=True`

saleor/graphql/checkout/mutations/checkout_customer_note_update.py
```diff
@@ -0,0 +1,57 @@
...
+from ...plugins.dataloaders import get_plugin_manager_promise
+from ..types import Checkout
+
+
+class CheckoutCustomerNoteUpdate(BaseMutation):
+    checkout = graphene.Field(Checkout, description="An updated checkout.")
+
+    class Arguments:
+        id = graphene.ID(
+            description="The checkout's ID." + ADDED_IN_321,
```
korycins, line 19: Sorry, in the previous comment, I suggested changing it to ADDED_IN_321, but as this mutation is fully new, we don't need `ADDED_IN` in the arguments list. The one in `Meta.description` covers this.

korycins (commented):
Two small comments and we are ready to merge

IKarbowiak (dismissed):

maarcingebala:
@pitkes22 After the recent change the tests are failing, most likely you need to update GraphQL queries for this mutation to use `$id: ID!` instead of `$id: ID`, as the field is now required.

pitkes22:
@maarcingebala Oh, sorry for all the hassle. My bad, I forgot to rerun the tests. :facepalm: It should all be fixed now.

maarcingebala (approved):

IKarbowiak (approved):

korycins (approved):
Great work @pitkes22! ❤️  Thanks for the PR once again!

PR #16323, opened 2024-07-15
Add filterable subscription for order webhooks

korycins:
I want to merge this change because it introduces the possibility of filtering out order webhooks via channel slug. 
So providing the subscription query like this:
```graphql
subscription {
  orderCreated(channels: ["default-channel"]) {
    order {
      id
      number
      lines {
        id
        variant {
          id
        }
      }
    }
  }
}
```
Will send `order-created` event only for orders that belong to the channel (slug:default-channel).

> [!NOTE]  
> Only one top field can be requested as a subscription query. This means that in a single subscription, you cannot define more than one filterable event or mix it with the `event` field. In that case, a new webhook with a new subscription should be created.

# Impact

- [x] New migrations

# Docs

- [ ] Link to documentation: https://github.com/saleor/saleor-docs/pull/1228

# Pull Request Checklist


saleor/graphql/webhook/subscription_query.py
```diff
@@ -48,22 +48,44 @@ def validate_query(self) -> list[Union[GraphQLSyntaxError, ValidationError]]:
...
 
-    def get_events_from_subscription(self) -> list[str]:
-        subscription = self._get_subscription(self.ast)
-        if not subscription:
-            raise ValidationError(
-                message="Subscription operation can't be found.",
-                code=WebhookErrorCode.MISSING_SUBSCRIPTION.value,
-            )
+    def get_events_from_subscription(self, subscription) -> list[str]:
+        subscription_name = subscription.selection_set.selections[0].name.value
```
IKarbowiak, line 83: Are you sure that you will always have something in `subscription.selection_set.selections`?

IKarbowiak (dismissed):

8r2y5 (approved):

saleor/graphql/webhook/subscription_types.py
```diff
@@ -2700,11 +2701,212 @@ class Meta:
...
         Event,
         description="Look up subscription event." + ADDED_IN_32,
     )
+    order_created = graphene.Field(
+        OrderCreated,
+        description=(
+            "Event sent when new order is created." + ADDED_IN_320 + PREVIEW_FEATURE
+        ),
+        resolver=filter_order_over_channels(WebhookEventAsyncType.ORDER_CREATED),
+        channels=graphene.Argument(
```
kadewu, line 2732: I believe we can extract that argument definition and reuse it.

saleor/graphql/webhook/subscription_query.py
korycins: I tested this, and I wasnt able to get something different. Additionally we already fetch selections in this way, since beginning of 2023
image

saleor/graphql/webhook/subscription_query.py
```diff
@@ -48,22 +78,44 @@ def validate_query(self) -> list[Union[GraphQLSyntaxError, ValidationError]]:
...
+                ValidationError(
+                    message="Subscription operation can't be found.",
+                    code=WebhookErrorCode.MISSING_SUBSCRIPTION.value,
+                )
+            ]
+
+        field_names = [
+            selection.name.value for selection in subscription.selection_set.selections
+        ]
+        if len(field_names) > 1 and set(field_names) != {"event"}:
```
kadewu, line 94: Can we extract this if with some meaningful name to increase readability? Maybe it's just me, but it took me a little to understand the idea here.

kadewu (approved):

saleor/graphql/webhook/tests/test_subscription_query.py
fowczarek, line 1: How about extracting those queries to variables?
example:
image

saleor/plugins/webhook/plugin.py
```diff
@@ -728,11 +728,38 @@ def gift_card_export_completed(
...
+        event_type: str,
+        order: "Order",
+        webhooks: Optional[Iterable["Webhook"]] = None,
+    ) -> Iterable["Webhook"]:
+        """Get webhooks for order events.
+
+        Fetch all valid webhooks and filter out the ones that have a subscription query
+        with filter that doesn't match to the order.
+        """
+        order_channel_slug = order.channel.slug
```
fowczarek, line 742: Does the Saleor make an additional query? Maybe we could pass the channel here for outside code?

saleor/plugins/webhook/tests/subscription_webhooks/filterable_webhooks/test_order_cancelled.py
```diff
@@ -0,0 +1,227 @@
...
+    "saleor.webhook.transport.asynchronous.transport.send_webhook_request_async.apply_async"
+)
+@override_settings(PLUGINS=["saleor.plugins.webhook.plugin.WebhookPlugin"])
+def test_order_cancelled(mocked_async, order_line, subscription_webhook):
+    # given
+    manager = get_plugins_manager(False)
+    order = order_line.order
+    channel = order.channel
+    channel.slug = "default-channel"
+    channel.save()
```
fowczarek, line 40: This action is not required. `order` fixture uses `channel_USD` with `DEFAULT_CHANNEL_SLUG` as a slug.  It should be fixed in all tests. 
```
def channel_USD(db):
    slug = settings.DEFAULT_CHANNEL_SLUG
``` 
https://github.com/saleor/saleor/blob/64d413d87b7e3f1b7732956e966cddb9721608d1/saleor/tests/fixtures.py#L1418

korycins: yes, I know, but wanted to do this explicitly, to highlight the slug of the channel :)

kadewu (dismissed):

saleor/plugins/webhook/tests/subscription_webhooks/filterable_webhooks/test_order_cancelled.py
fowczarek: If you want to have it explicate add assert instead of save. (Save makes an extra call to the database.)

saleor/graphql/webhook/subscription_types.py
```diff
@@ -2713,6 +2828,24 @@ class Meta:
...
 
+    @staticmethod
+    def resolve_order_bulk_created(root, info: ResolveInfo, channels=None):
+        event_type, orders = root
+        if event_type != WebhookEventAsyncType.ORDER_BULK_CREATED:
+            return Observable.from_([])
+
+        orders_to_return = []
+        if channels:
+            channel_ids = Channel.objects.filter(slug__in=channels).values_list(
```
maarcingebala, line 2839: This could use replica DB.

saleor/plugins/webhook/tests/subscription_webhooks/filterable_webhooks/test_order_cancelled.py
korycins: assert sounds good

maarcingebala (approved):

kadewu (approved):

PR #16325, opened 2024-07-15
Add logs when checkout is paid by gift cards.

zedzior:
I want to merge this change because it adds logs when a gift card was used to pay for the checkout.

Issue: https://linear.app/saleor/issue/SHOPX-1000/add-logs-when-gift-card-was-used-to-pay-for-the-checkout

### DIFF
--- a/saleor/checkout/calculations.py
+++ b/saleor/checkout/calculations.py
@@ -134,7 +134,7 @@ def calculate_checkout_total_with_gift_cards(
                 total_with_gift_cards.gross.amount
             ),
         }
-        logger.info("Gift card payment: %s", details)
+        logger.info("Gift card payment.", extra=details)
         return total_with_gift_cards
 
     return total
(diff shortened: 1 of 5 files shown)


korycins (approved):

saleor/checkout/calculations.py
```diff
@@ -117,9 +117,27 @@ def calculate_checkout_total_with_gift_cards(
...
+        )
+        gift_card_compensation = min(total.gross, gift_cards_balance)
+        details = {
+            "checkout_id": str(checkout_info.checkout.pk),
+            "gift_card_compensation": str(gift_card_compensation.amount),
+            "total_after_gift_card_compensation": str(
+                total_with_gift_cards.gross.amount
+            ),
+        }
+        logger.info("Gift card payment.", extra=details)
```
IKarbowiak, line 137: I'm not sure if this is the right place, as we just have a calculation here. And it will be logged each time prices for checkout are recalculated. Also, the many from gift card is not reduced here. Maybe `add_gift_cards_to_order` is a better place?

zedzior: make sense, thanks

IKarbowiak (approved):

Air-t (approved):

PR #16326, opened 2024-07-16
Fix CollectionsByProductIdLoader if collection is missing

Air-t:
In some cases, the collections list may contain `None` instance. Fixes `AttributeError` when that happens.

Fixes: https://linear.app/saleor/issue/SHOPX-683/attributeerror-nonetype-object-has-no-attribute-id

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/dataloaders/products.py
+++ b/saleor/graphql/product/dataloaders/products.py
@@ -523,10 +523,8 @@ class CollectionByIdLoader(DataLoader):
     context_key = "collection_by_id"
 
     def batch_load(self, keys):
-        collections = (
-            Collection.objects.using(self.database_connection_name)
-            .using(self.database_connection_name)
-            .in_bulk(keys)
+        collections = Collection.objects.using(self.database_connection_name).in_bulk(
+            keys
         )
         return [collections.get(collection_id) for collection_id in keys]
 
@@ -537,7 +535,6 @@ class CollectionsByProductIdLoader(DataLoader):
     def batch_load(self, keys):
         product_collection_pairs = list(
             CollectionProduct.objects.using(self.database_connection_name)
-            .using(self.database_connection_name)
             .filter(product_id__in=keys)
             .order_by("id")
             .values_list("product_id", "collection_id")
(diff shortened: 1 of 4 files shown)


saleor/graphql/product/dataloaders/products.py
```diff
@@ -548,7 +547,7 @@ def batch_load(self, keys):
             product_collection_map[pid].append(cid)
 
         def map_collections(collections):
-            collection_map = {c.id: c for c in collections}
+            collection_map = {c.id: c for c in collections if c is not None}
```
korycins, line 550: It won't solve the problem. You will change the line where an error can occur. The error is raised as probably the collection was removed in the meantime. In line 552, there is fetch via key, which will raise an exception, when the collection ID doesn't exist - and there is a possibility that the collection will not exist as you added validation for `None`.

Air-t: thanks, I checked the case when a Collection is removed. Now there is no issue with handling that situation.

saleor/graphql/product/dataloaders/products.py
```diff
@@ -548,9 +547,13 @@ def batch_load(self, keys):
...
             return [
-                [collection_map[cid] for cid in product_collection_map[pid]]
+                [
+                    collection_map[cid]
+                    for cid in product_collection_map[pid]
+                    if cid in collection_map
```
fowczarek, line 555: @Air-t Dataloader should always return a list of the same length as the list of keys. This solution doesn't meet this requirement. I think you should start by creating a unit test with bug reproduction.

Air-t: I managed to reproduce it removing a collection while a dataloader `batch_load` was partially executed and `product_collection_pairs` list was created. Now both keys and return value lengths match. 
In a reproduced example, other functions had to be fixated to not fail while the collection was resolved as None by dataloader

fowczarek: Could you write unit tests using the `before_after` module to reproduce the described case? We use this module in some tests e.g `saleor/checkout/tests/test_checkout_complete.py::test_complete_checkout_checkout_was_deleted_before_completing`

fowczarek (approved):

zedzior (approved):

IKarbowiak (approved):

PR #16331, opened 2024-07-16
Fix invalid gift card calculation on checkout when shipping voucher is used

IKarbowiak:
The gift card balance was wrongly calculated when the shipping voucher was used on checkout.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/tests/mutations/test_checkout_complete_with_transactions.py
+++ b/saleor/graphql/checkout/tests/mutations/test_checkout_complete_with_transactions.py
@@ -8,6 +8,7 @@ import pytest
 import pytz
 from django.db.models.aggregates import Sum
 from django.utils import timezone
+from prices import TaxedMoney
 
 from .....account.models import Address
 from .....channel import MarkAsPaidStrategy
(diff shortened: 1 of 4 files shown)


saleor/graphql/checkout/tests/mutations/test_checkout_complete_with_transactions.py
```diff
@@ -1640,6 +1653,131 @@ def test_checkout_complete_gift_card_bought(
     assert Fulfillment.objects.count() == 1
 
 
+@patch("saleor.plugins.manager.PluginsManager.order_confirmed")
```
Air-t, line 1656: an integration marker is missing

Air-t (approved):

saleor/graphql/checkout/tests/mutations/test_checkout_complete_with_transactions.py
IKarbowiak: Basically, it doesn't matter as we are not excluding or running only integration test, I'm also not sure if it's integration test at all.

zedzior (approved):

PR #16386, opened 2024-07-22
Improve filter_customer on orders

8r2y5:
I want to merge this change because it will improve filtering by customer when querying orders or draft orders

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/filters.py
+++ b/saleor/graphql/order/filters.py
@@ -3,6 +3,7 @@ from uuid import UUID
 import django_filters
 import graphene
 from django.core.exceptions import ValidationError
+from django.core.validators import validate_email
 from django.db.models import Exists, OuterRef, Q
 from django.utils import timezone
 from graphql.error import GraphQLError
@@ -79,12 +80,23 @@ def filter_status(qs, _, value):
 
 
 def filter_customer(qs, _, value):
-    qs = qs.filter(
-        Q(user_email__ilike=value)
-        | Q(user__email__trigram_similar=value)
-        | Q(user__first_name__trigram_similar=value)
-        | Q(user__last_name__trigram_similar=value)
-    )
+    try:
+        validate_email(value)
+        qs = qs.filter(Q(user_email__iexact=value) | Q(user__email__iexact=value))
+    except ValidationError:
+        try:
+            first_name, last_name = value.split()
+            qs = qs.filter(
+                user__first_name__iexact=first_name, user__last_name__iexact=last_name
+            )
+        except ValueError:
+            qs = qs.filter(
+                Q(user_email__ilike=value)
+                | Q(user__email__ilike=value)
+                | Q(user__first_name__ilike=value)
+                | Q(user__last_name__ilike=value)
+            )
+
     return qs
 
 
--- a/saleor/graphql/order/tests/test_filters.py
+++ b/saleor/graphql/order/tests/test_filters.py
@@ -1,0 +1,94 @@
+import pytest
+
+from ....order.models import Order
+from ..filters import filter_customer
+
+
+@pytest.fixture
+def similar_customers_with_orders(order, customer_user, customer_user2, channel_USD):
+    customer_user.first_name = "Test"
+    customer_user.last_name = "User"
+    customer_user.save()
+    customer_user2.first_name = "tester"
+    customer_user2.last_name = "users"
+    customer_user2.save()
+    order2 = Order.objects.create(
+        billing_address=customer_user2.default_billing_address,
+        channel=channel_USD,
+        user_email=customer_user2.email,
+        user=customer_user2,
+    )
+    return order, order2, customer_user, customer_user2
+
+
+def test_filter_customer_by_email(similar_customers_with_orders):
+    # given
+    order, order2, customer_user, customer_user2 = similar_customers_with_orders
+    qs = Order.objects.all()
+    value = customer_user.email
+
+    # when
+    qs = filter_customer(qs, None, value)
+
+    # then
+    assert qs.count() == 1
+    assert qs.first() == order
+
+
+def test_filter_customer_by_first_name(similar_customers_with_orders):
+    # given
+    order, order2, customer_user, customer_user2 = similar_customers_with_orders
+    qs = Order.objects.all()
+    value = customer_user.first_name
+
+    # when
+    qs = filter_customer(qs, None, value)
+
+    # then
+    assert qs.count() == 2
+    assert order in qs
+    assert order2 in qs
+
+
+def test_filter_customer_by_last_name(similar_customers_with_orders):
+    # given
+    order, order2, customer_user, customer_user2 = similar_customers_with_orders
+    qs = Order.objects.all()
+    value = customer_user.last_name
+
+    # when
+    qs = filter_customer(qs, None, value)
+
+    # then
+    assert qs.count() == 2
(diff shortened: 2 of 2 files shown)


saleor/graphql/order/filters.py
```diff
@@ -79,12 +80,23 @@ def filter_status(qs, _, value):
...
-    )
+    try:
+        validate_email(value)
+        qs = qs.filter(Q(user_email__iexact=value) | Q(user__email__iexact=value))
+    except ValidationError:
+        try:
+            first_name, last_name = value.split()
+            qs = qs.filter(
+                user__first_name__iexact=first_name, user__last_name__iexact=last_name
+            )
```
patrys, line 91: Is this using a DB index?

8r2y5: No, there is no index on `first_name` or `last_name` fields. Would you suggest adding those indexes or utilize `User.search_document` for this search?

saleor/account/migrations/0086_auto_20240722_1753.py
```diff
@@ -0,0 +1,34 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("account", "0085_address_validation_skipped"),
+    ]
+
+    operations = [
+        migrations.AddIndex(
```
korycins, line 14: Should be `AddIndexConcurrently`

kadewu, line 14: Please add indexes concurrently cause of zero-downtime.

saleor/graphql/order/filters.py
```diff
@@ -79,12 +80,23 @@ def filter_status(qs, _, value):
...
-        | Q(user__email__trigram_similar=value)
-        | Q(user__first_name__trigram_similar=value)
-        | Q(user__last_name__trigram_similar=value)
-    )
+    try:
+        validate_email(value)
+        qs = qs.filter(Q(user_email__iexact=value) | Q(user__email__iexact=value))
+    except ValidationError:
+        try:
+            first_name, last_name = value.split()
```
korycins, line 88: This means the assumption that you will always provide value in format `{first_name} {last_name}`. What if I provide it in reverse order? Or only last_name? Is it expected that we will call the previous filter in that case?

saleor/account/migrations/0086_auto_20240722_1753.py
```diff
@@ -0,0 +1,34 @@
...
+            model_name="user",
+            index=django.contrib.postgres.indexes.GinIndex(
+                django.contrib.postgres.indexes.OpClass(
+                    django.db.models.functions.text.Upper("first_name"),
```
kadewu, line 18: trgm indexes are case insensitive, no reason for this step.
Tip: check locally `SELECT show_trgm('Name')` vs `SELECT show_trgm(UPPPER('Name'))`

saleor/graphql/order/filters.py
8r2y5: Per conversation, indexes will be added for `first_name` and `last_name`.

saleor/graphql/order/filters.py
8r2y5: > This means the assumption that you will always provide value in format {first_name} {last_name}. What if I provide it in reverse order?

Good catch, i will adjust it to also handle `{last_name} {first_name}`.

> Or only last_name? Is it expected that we will call the previous filter in that case?

It is not exactly previous filter, previous filter were doing trigram similar search, while new one is doing full text search using `ilike`. 
Handling it this way it is expected.
1. if email is valid then search by email,
2. if not, try to split `value` into first and last name and try match using those values aka full name match,
3. if there was problem splitting value (aka not a full name), search using whole `value`.

If value is `last_name` it will be handled by point 3.

This is different from previous filter which was doing just search using stemming.

saleor/account/models.py
```diff
@@ -316,6 +333,10 @@ def can_login(self, site_settings: SiteSettings):
             or self.is_confirmed
         )
 
+    def save(self, *args, **kwargs):
```
fowczarek, line 336: I believe it may not be sufficient to cover all possible customer updates. When using `update` or `bulk_update` in Saleor, the `save` function is not called automatically.

You should add/extend unit tests of any possible way of updating user `first_name` and `last_name` to check if `full_name` is updated.

fowczarek (changes requested):

saleor/account/models.py
```diff
@@ -190,6 +190,7 @@ class User(
     )
     search_document = models.TextField(blank=True, default="")
     uuid = models.UUIDField(default=uuid4, unique=True)
+    full_name = models.CharField(max_length=512, blank=True, default="")
```
patrys, line 193: I believe we decided against adding a new field for this purpose.

saleor/graphql/order/filters.py
```diff
@@ -79,12 +81,30 @@ def filter_status(qs, _, value):
...
+                Q(user_full_name__iexact=value)
+                | Q(user_full_name__iexact=f"{last} {first}")
+            )
+        except ValueError:
+            qs = qs.filter(
+                Q(user_email__ilike=value)
+                | Q(user__email__ilike=value)
+                | Q(user__first_name__ilike=value)
+                | Q(user__last_name__ilike=value)
+            )
```
IKarbowiak, line 106: I do not like these nested `try except` parts. In `try` you should have only the part that might raise an error. If you want to call something when error is not raised you should use `else` part.
Maybe we should split it into multiple methods like that:

```suggestion
def filter_customer(qs, _, value):
    try:
        validate_email(value)
    except ValidationError:
       return  _filter_customer_by_full_name(value)
   # filter by user email if the provided value is an email
    return  qs.filter(Q(user_email__iexact=value) | Q(user__email__iexact=value))


def _filter_customer_by_full_name(value):
        try:
            first, last = value.split(" ", 1)
        except ValueError:
            return qs.filter(
                Q(user_email__ilike=value)
                | Q(user__email__ilike=value)
                | Q(user__first_name__ilike=value)
                | Q(user__last_name__ilike=value)
            )
        return  qs.alias(
                user_full_name=FlatConcat(
                    "user__first_name",
                    Value(" "),
                    "user__last_name",
                )
            ).filter(
                Q(user_full_name__iexact=value)
                | Q(user_full_name__iexact=f"{last} {first}")
            )
```
WDYT?

patrys (approved):
Approving but I like Iga's suggestion

saleor/graphql/order/tests/test_filters.py
```diff
@@ -0,0 +1,108 @@
+import pytest
+
+from ....order.models import Order
+from ..filters import filter_customer
```
korycins, line 4: I would recommend to do `@mock.patch("..._filter_customer_by_email_first_or_last_name", wraps=_filter_customer_by_email_first_or_last_name)`. (example for tests when you're looking for first or last name). 
Then `assert mocked.called`. `wraps` will execute `_filter_customer_by_email_first_or_last_name`, but you will be also able to fully confirm that your expected logic was called.
Thanks to this we will be sure that we always trigger a proper search logic.

korycins (approved):

IKarbowiak (approved):

korycins (approved):

PR #16393, opened 2024-07-23
Fix empty sync webhooks for checkout type

korycins:
I want to merge this change because it should solve the problem of empty subscription payloads for sync webhooks when called inside async webhook. 

In the case of having subscription for checkout's sync and async webhooks, and invalid prices/shipping methods, Saleor generates the subscription for async webhook. Inside the async subscription some fields can trigger sync webhook like the one responsible for external shipping methods or calculating the taxes. This tries to build the subscription query for sync webhooks, but due to calling `promise.get()` inside the subscription logic, we break the promise loop. As a result, we don't send sync webhooks as the generated payload is empty.
As a temporary workaround, we're adding the wrapper for async events that would trigger sync webhooks first (if needed), then async events. With this approach, we are sure, that the sync webhook is sent, and async webhook can use denormalized/cached data to build the payload. 

Scope of changes:
- adding new wrapper to trigger async checkout events like `checkout_created`, `checkout_updated`. Trigger the sync webhooks only when they exist and there is a subscription for async event. 
- blocking the default `call_event` from using it with `Checkout` CheckoutInfo object.
- extending all checkout mutations and logic responsible for triggering async events to use a new wrapper
- Fixing filter shipping method cache logic as it didn't work and it was breaking my tests. 

Internal task link: https://linear.app/saleor/issue/MERX-156/application-with-order-updated-webhook-subscribe-for-ordertotal-may

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/actions.py
+++ b/saleor/checkout/actions.py
@@ -26,11 +26,19 @@ if TYPE_CHECKING:
     from ..webhook.models import Webhook
 
 
-def checkout_event_requires_additional_action(event_name, webhook_event_map):
-    # No need for additional actions as we don't have any webhook for the `event_name`
+def checkout_event_requires_sync_webhooks_to_trigger(
+    event_name: str, webhook_event_map: dict[str, set["Webhook"]]
+) -> bool:
+    """Check if calling the event requires additional actions.
+
+    In case of having active webhook with the `event_name`, the function will return
+    True, when any sync checkout's webhooks are active. `True` means that sync webhook
+    should be triggered first, before calling async webhook.
+    """
+
     if event_name not in webhook_event_map:
         return False
-    # No need for additional actions as we don't have any sync webhook for the checkout
+
     if not set(WebhookEventSyncType.CHECKOUT_EVENTS).intersection(
         webhook_event_map.keys()
     ):
@@ -47,7 +55,9 @@ def call_checkout_event_for_checkout(
     webhook_event_map = get_webhooks_for_multiple_events(
         [event_name, *WebhookEventSyncType.CHECKOUT_EVENTS]
     )
-    if not checkout_event_requires_additional_action(event_name, webhook_event_map):
+    if not checkout_event_requires_sync_webhooks_to_trigger(
+        event_name, webhook_event_map
+    ):
         call_event_including_protected_events(event_func, checkout)
         return
 
@@ -77,7 +87,7 @@ def call_checkout_event_for_checkout_info(
     checkout_info: "CheckoutInfo",
     lines: Iterable["CheckoutLineInfo"],
     address: Optional["Address"] = None,
-    webhook_event_map: Optional[dict[str, list["Webhook"]]] = None,
+    webhook_event_map: Optional[dict[str, set["Webhook"]]] = None,
 ):
     checkout = checkout_info.checkout
     if webhook_event_map is None:
@@ -87,7 +97,9 @@ def call_checkout_event_for_checkout_info(
 
     # No need to trigger additional sync webhook when we don't have active webhook or
     # we don't have active sync checkout webhooks
-    if not checkout_event_requires_additional_action(event_name, webhook_event_map):
+    if not checkout_event_requires_sync_webhooks_to_trigger(
+        event_name, webhook_event_map
+    ):
         call_event_including_protected_events(event_func, checkout)
         return None
 
--- a/saleor/checkout/actions.py
+++ b/saleor/checkout/actions.py
@@ -1,14 +1,116 @@
+from collections.abc import Iterable
+from datetime import timedelta
 from decimal import Decimal
-from typing import cast
+from typing import TYPE_CHECKING, Callable, Optional, cast
 
-from ..core.utils.events import call_event
+from django.utils import timezone
+
+from ..core.utils.events import call_event_including_protected_events
 from ..payment.models import TransactionItem
-from ..plugins.manager import PluginsManager
+from ..webhook.event_types import WebhookEventAsyncType, WebhookEventSyncType
+from ..webhook.utils import get_webhooks_for_multiple_events
 from . import CheckoutChargeStatus
 from .calculations import fetch_checkout_data
-from .fetch import fetch_checkout_info, fetch_checkout_lines
+from .fetch import (
+    CheckoutInfo,
+    CheckoutLineInfo,
+    fetch_checkout_info,
+    fetch_checkout_lines,
+)
 from .models import Checkout
 from .payment_utils import update_refundable_for_checkout
+
+if TYPE_CHECKING:
+    from ..account.models import Address
+    from ..plugins.manager import PluginsManager
+    from ..webhook.models import Webhook
+
+
+def checkout_event_requires_additional_action(event_name, webhook_event_map):
+    # No need for additional actions as we don't have any webhook for the `event_name`
+    if event_name not in webhook_event_map:
+        return False
+    # No need for additional actions as we don't have any sync webhook for the checkout
+    if not set(WebhookEventSyncType.CHECKOUT_EVENTS).intersection(
+        webhook_event_map.keys()
+    ):
+        return False
+    return True
+
+
+def call_checkout_event_for_checkout(
+    manager: "PluginsManager",
+    event_func: Callable,
+    event_name: str,
+    checkout: "Checkout",
+):
+    webhook_event_map = get_webhooks_for_multiple_events(
+        [event_name, *WebhookEventSyncType.CHECKOUT_EVENTS]
+    )
+    if not checkout_event_requires_additional_action(event_name, webhook_event_map):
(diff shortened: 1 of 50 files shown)


saleor/graphql/checkout/mutations/checkout_add_promo_code.py
```diff
@@ -120,6 +121,12 @@ def perform_mutation(  # type: ignore[override]
             recalculate_discount=False,
             save=True,
         )
-        cls.call_event(manager.checkout_updated, checkout)
+        call_checkout_event_for_checkout_info(
+            manager=manager,
+            event_func=manager.checkout_updated,
```
korycins, line 126: I am aware this is "ugly" way of passing arguments. Instead of passing `=manager.checkout_updated,` and `event_name=WebhookEventAsyncType.CHECKOUT_UPDATED,` the best way would be passing single variable. I will address this in a separate PR. First, need to figure out a better way of doing this.

IKarbowiak (dismissed):

saleor/checkout/actions.py
```diff
@@ -1,15 +1,117 @@
...
+        call_event_including_protected_events(event_func, checkout)
+        return None
+
+    _ = checkout_info.all_shipping_methods
+
+    # + timedelta(seconds=10) to confirm that triggered webhooks will still have a
+    # valid prices. Triggered only when we have active sync tax webhook
+    if (
+        WebhookEventSyncType.CHECKOUT_CALCULATE_TAXES in webhook_event_map
+        and checkout.price_expiration < timezone.now() + timedelta(seconds=10)
```
kadewu, line 100: is there a reason to not to use `CHECKOUT_PRICES_TTL`?

saleor/core/utils/events.py
```diff
@@ -1,13 +1,26 @@
...
+    """Call webhook event with given args.
+
+    Ensures that in atomic transaction event is called on_commit.
+    """
+    is_checkout_instance = any(
+        [isinstance(arg, (Checkout, CheckoutInfo)) for arg in func_args]
+    )
+
+    if is_checkout_instance:
+        raise NotImplementedError("`call_event` doesn't support checkout/order events.")
```
kadewu, line 25: I believe we check only for checkout instances above?

saleor/webhook/utils.py
```diff
@@ -54,13 +48,90 @@ def get_webhooks_for_event(
...
+    event_map = defaultdict(list)
+    for webhook in webhooks:
+        app_permission_codenames = app_perm_map.get(webhook.app_id, [])
+        events: set[str] = webhook_events_map.get(webhook.id, set())
+        for event_type in events:
+            required_permission = WebhookEventAsyncType.PERMISSIONS.get(
+                event_type, WebhookEventSyncType.PERMISSIONS.get(event_type)
+            )
+            if not required_permission:
+                continue
```
kadewu, line 133: codecov

kadewu (dismissed):

saleor/checkout/actions.py
```diff
@@ -1,15 +1,117 @@
...
 from .models import Checkout
 from .payment_utils import update_refundable_for_checkout
 
+if TYPE_CHECKING:
+    from ..account.models import Address
+    from ..plugins.manager import PluginsManager
+    from ..webhook.models import Webhook
+
+
+def checkout_event_requires_additional_action(event_name, webhook_event_map):
```
maarcingebala, line 29: By "additional actions" you mean sync webhooks?

saleor/webhook/utils.py
```diff
@@ -54,13 +48,90 @@ def get_webhooks_for_event(
...
+    apps_identifier: Optional[list[str]] = None,
+) -> dict[str, list[Webhook]]:
+    if webhooks is None:
+        # For this QS replica usage is applied later, as this QS could be also passed
+        # as parameter.
+        webhooks = Webhook.objects.all()
+
+    filter_expresion = Q()
+    for event_type in set(event_types):
+        filter_expresion |= Q(
```
maarcingebala, line 96: If there were many events, this query could result in a quite complex query plan, I'm afraid. Maybe it won't be the case for now, as I guess we won't have many webhooks or events at once. Anyway, I think in the long term, a dedicated SQL query could be better to handle cases with multiple events than joining filters from the single-event case.

saleor/checkout/actions.py
fowczarek: We want to be sure that the price should be actual before the end of the request when we use `CHECKOUT_PRICES_TTL`, which means that any of the checkouts will meet the `checkout.price_expiration < timezone.now() + CHECKOUT_PRICES_TTL`  condition. 
checkout.price_expiration is calculated from: `timezone.now() + timedelta(seconds=10)`.

saleor/graphql/meta/extra_methods.py
```diff
@@ -1,12 +1,58 @@
...
+            webhook_event_map=webhook_event_map,
+        )
+        call_checkout_event_for_checkout_info(
+            manager=manager,
+            event_func=manager.checkout_metadata_updated,
+            event_name=WebhookEventAsyncType.CHECKOUT_METADATA_UPDATED,
+            checkout_info=checkout_info,
+            lines=lines_info,
+            webhook_event_map=webhook_event_map,
+        )
```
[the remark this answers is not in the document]
korycins: Yeah, I was also thinking about this. I will try to figure out better usage in separate PR (first need to finish the same thing for order part).

saleor/core/utils/events.py
korycins: I can change this, but actually working on separate PR for handling the same thing in order.

saleor/webhook/utils.py
korycins: You're right, if we want to change this, I would need to fetch all apps, all their permissions, then all webhooks assigned to these apps, and all webhook events assigned to webhook. Then I could get the same result as |= filter_expression. 
I am working on the order part right now, so I will try to figure out if I can improve this right now.

saleor/core/utils/events.py
kadewu: Should be fine then.

szczecha (dismissed):

saleor/graphql/checkout/mutations/checkout_add_promo_code.py
korycins: Added this to my to-do list for the task: https://linear.app/saleor/issue/MERX-156/application-with-order-updated-webhook-subscribe-for-ordertotal-may

saleor/graphql/meta/extra_methods.py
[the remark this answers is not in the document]
korycins: Added this to my to-do list for the task: https://linear.app/saleor/issue/MERX-156/application-with-order-updated-webhook-subscribe-for-ordertotal-may

saleor/webhook/utils.py
korycins: Added this to my to-do list for the task: https://linear.app/saleor/issue/MERX-156/application-with-order-updated-webhook-subscribe-for-ordertotal-may
I will double check if we can make this call better/cost of current query in case of having multiple webhooks

saleor/checkout/actions.py
```diff
@@ -1,15 +1,117 @@
...
+    # No need for additional actions as we don't have any webhook for the `event_name`
+    if event_name not in webhook_event_map:
+        return False
+    # No need for additional actions as we don't have any sync webhook for the checkout
+    if not set(WebhookEventSyncType.CHECKOUT_EVENTS).intersection(
+        webhook_event_map.keys()
+    ):
+        return False
```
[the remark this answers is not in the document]
korycins: Added more tests. It should be covered right now

saleor/webhook/utils.py
korycins: Added more tests. It should be covered now.

saleor/core/utils/events.py
```diff
@@ -1,13 +1,26 @@
...
 
+from ...checkout.fetch import CheckoutInfo
+from ...checkout.models import Checkout
 
-def call_event(func_obj, *func_args, **func_kwargs):
-    """Call webhook event with given args.
 
-    Ensures that in atomic transaction event is called on_commit.
-    """
+def call_event_including_protected_events(func_obj, *func_args, **func_kwargs):
```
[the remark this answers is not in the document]
korycins: Added a docstring

saleor/checkout/actions.py
korycins: Added docstring and rename the function to `checkout_event_requires_sync_webhooks_to_trigger`

kadewu (approved):

maarcingebala (approved):

fowczarek (dismissed):

fowczarek (approved):

kadewu (approved):

PR #16413, opened 2024-07-25
Disallow changing voucher usage limit setting when voucher has been already used.

zedzior:
I want to merge this change because I want to disallow changing voucher usage limit when voucher has been already used

Issue: https://linear.app/saleor/issue/SHOPX-1089/count-current-voucher-usage-when-setting-the-usage-limit-for-voucher

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/discount/mutations/voucher/voucher_update.py
+++ b/saleor/graphql/discount/mutations/voucher/voucher_update.py
@@ -60,21 +60,11 @@ class VoucherUpdate(VoucherCreate):
 
     @classmethod
     def clean_voucher_usage_setting(cls, instance, data):
-        """Ensure single use setting is not changed if voucher was already used."""
+        """Ensure voucher usage settings are not changed if voucher was already used."""
+        is_voucher_used = None
         if "single_use" in data and instance.single_use != data["single_use"]:
-            voucher_codes = instance.codes.all()
-            used_codes = voucher_codes.filter(
-                Exists(order_models.Order.objects.filter(voucher_code=OuterRef("code")))
-                | Exists(
-                    order_models.OrderLine.objects.filter(voucher_code=OuterRef("code"))
-                )
-                | Exists(
-                    checkout_models.Checkout.objects.filter(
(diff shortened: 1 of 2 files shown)


saleor/graphql/discount/mutations/voucher/voucher_update.py
```diff
@@ -85,12 +91,69 @@ def clean_voucher_usage_setting(cls, instance, data):
                     }
                 )
 
+    @classmethod
+    def count_voucher_usage(cls, voucher: Voucher) -> dict[str, int]:
+        voucher_codes = [code.code for code in voucher.codes.all()]
+        orders = order_models.Order.objects.select_related("channel").filter(
```
korycins, line 97: Why do you need to take this value from order/checkout? In the case of vouchers used in multiple orders/checkouts this query can be super heavy on DB, and also from the memory usage perspective. Why can't we use `VoucherCode.used`?

IKarbowiak: @korycins `code.used` is updated only when the voucher has `usage_limit` set. Here is the case where the limit is set. 
But you are right, this query looks very heavy. Maybe we should consider blocking such behavior when the voucher has already been used.

zedzior: We decided to disallow to change the setting if the voucher has been already used.

saleor/graphql/discount/mutations/voucher/voucher_update.py
```diff
@@ -84,6 +74,33 @@ def clean_voucher_usage_setting(cls, instance, data):
                         )
                     }
                 )
+        if "usage_limit" in data and instance.usage_limit != data["usage_limit"]:
```
IKarbowiak, line 77: What if the user updates the voucher with the same value as before? 
You should add a test case to cover such a case.

IKarbowiak (approved):

Air-t (approved):

PR #16451, opened 2024-07-29
Downgrade the dependecies which uses setuptools in the newest version

korycins:
I want to merge this change because setuptools removed some of the deprecated logic which is still used in some packages.  Because of this, we are not able to call `docker build` command, as `poetry install` raises an exception. 
I tried different way of handling this, but I am not sure, why the downgraded packages are using the setuptools ==72. Currently we're using 70, and we shouldn't have an issue with this. The docker image from `docker build` has installed setuptools < 60 so this is also not a case. As a result, I end up with downgrading the packages which were raising the exception. 

Port of changes: #16450 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


korycins:
No needed anymore. setuptools 72 was marked as yanky.

PR #16456, opened 2024-07-29
Fix voucher total usage count.

zedzior:
I want to merge this change because it counts total usage of the voucher (sum of all codes usage) instead of usage of a single code.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/discount/models.py
+++ b/saleor/discount/models.py
@@ -56,8 +56,10 @@ class VoucherQueryset(models.QuerySet["Voucher"]):
     def active(self, date):
         subquery = (
             VoucherCode.objects.filter(voucher_id=OuterRef("pk"))
+            .order_by()
+            .values("voucher_id")
             .annotate(total_used=Sum("used"))
-            .values("total_used")[:1]
+            .values("total_used")
         )
         return self.filter(
             Q(usage_limit__isnull=True) | Q(usage_limit__gt=Subquery(subquery)),
--- a/saleor/discount/models.py
+++ b/saleor/discount/models.py
@@ -82,8 +82,10 @@ class VoucherQueryset(models.QuerySet["Voucher"]):
     def expired(self, date):
         subquery = (
             VoucherCode.objects.filter(voucher_id=OuterRef("pk"))
+            .order_by()
+            .values("voucher_id")
             .annotate(total_used=Sum("used"))
-            .values("total_used")[:1]
+            .values("total_used")
         )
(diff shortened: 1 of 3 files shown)


saleor/graphql/order/tests/mutations/test_draft_order_create.py
```diff
@@ -3387,3 +3387,62 @@ def test_draft_order_create_create_no_shipping_method(
...
+    voucher.usage_limit = 1
+    voucher.save(update_fields=["usage_limit"])
+
+    code_1, code_2, _, _, _ = voucher.codes.all()
+    code_1.used = 1
+    code_1.save(update_fields=["used"])
+
+    user_id = graphene.Node.to_global_id("User", customer_user.id)
+    variant_id = graphene.Node.to_global_id("ProductVariant", variant.id)
+    variant_qty = 2
```
IKarbowiak, line 3419: We need more extended test.
1) We should have more vouchers
2) Each voucher should have more voucher code with `used` set
3) We should have a test when a limit is not exceeded in the voucher usage (once we should use the code with `used` and the second one we should use the new that was never used)
4) We should have a test when a limit is exceeded in such case
5) We should consider adding e2e test for such scenario - IMO it's worth

saleor/discount/models.py
```diff
@@ -56,8 +56,10 @@ class VoucherQueryset(models.QuerySet["Voucher"]):
     def active(self, date):
         subquery = (
             VoucherCode.objects.filter(voucher_id=OuterRef("pk"))
+            .order_by()
+            .values("voucher_id")
             .annotate(total_used=Sum("used"))
-            .values("total_used")[:1]
+            .values("total_used")
```
IKarbowiak, line 62: I'm not 100% sure that it will be working when the qs have multiple vouchers, and each voucher has multiple codes with `used`. Test needed for that

saleor/graphql/order/tests/mutations/test_draft_order_create.py
zedzior: Tests added. E2E tests will be added in separate PR

saleor/discount/models.py
zedzior: Tests added

IKarbowiak (approved):

Air-t (approved):

PR #16466, opened 2024-08-01
Fix empty sync webhooks for Order type

korycins:
I want to merge this change because it should solve the problem of empty subscription payloads for sync webhooks when called inside async webhook.

In the case of having subscription for order's sync and async webhooks, and invalid prices/shipping methods, Saleor generates the subscription for async webhook. Inside the async subscription some fields can trigger sync webhook like the one responsible for filtering shipping methods or calculating the taxes. This tries to build the subscription query for sync webhooks, but due to calling promise.get() inside the subscription logic, we break the promise loop. As a result, we don't send sync webhooks as the generated payload is empty.
As a temporary workaround, we're adding the wrapper for async events that would trigger sync webhooks first (if needed), then async events. With this approach, we are sure, that the sync webhook is sent, and async webhook can use denormalized/cached data to build the payload.

This is similar PR as we did for the checkout: https://github.com/saleor/saleor/pull/16393

Scope of changes:

- adding new wrapper to trigger async order events like order_updated. Trigger the sync webhooks only when they exist and there is async event.
- blocking the default call_event from using it with Order, OrderInfo object.
- extending all order mutations and logic responsible for triggering async events to use a new wrapper

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/mutations/order_confirm.py
+++ b/saleor/graphql/order/mutations/order_confirm.py
@@ -7,12 +7,18 @@ from django.db import transaction
 from ....account.models import User
 from ....core.tracing import traced_atomic_transaction
 from ....order import OrderStatus, models
-from ....order.actions import order_charged, order_confirmed
+from ....order.actions import (
+    WEBHOOK_EVENTS_FOR_ORDER_CHARGED,
+    WEBHOOK_EVENTS_FOR_ORDER_CONFIRMED,
+    order_charged,
+    order_confirmed,
+)
 from ....order.error_codes import OrderErrorCode
 from ....order.fetch import fetch_order_info
 from ....order.utils import update_order_display_gross_prices
 from ....payment import gateway
 from ....permission.enums import OrderPermissions
+from ....webhook.utils import get_webhooks_for_multiple_events
 from ...app.dataloaders import get_app_promise
 from ...core import ResolveInfo
 from ...core.mutations import ModelMutation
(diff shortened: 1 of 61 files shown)


saleor/graphql/order/tests/mutations/test_order_update_shipping.py
```diff
@@ -533,3 +538,84 @@ def test_order_shipping_update_mutation_properly_recalculate_total(
...
+    mocked_send_webhook_request_async.assert_called_once_with(
+        kwargs={"event_delivery_id": order_delivery.id},
+        queue=settings.ORDER_WEBHOOK_EVENTS_CELERY_QUEUE_NAME,
+        bind=True,
+        retry_backoff=10,
+        retry_kwargs={"max_retries": 5},
+    )
+    mocked_send_webhook_request_sync.assert_has_calls(
+        [
+            # FIXME: This is the issue with current way of caching filter shipping
```
korycins, line 609: https://linear.app/saleor/issue/MERX-805/proper-way-of-handling-the-filter-shipping-methods

saleor/graphql/order/mutations/order_confirm.py
```diff
@@ -73,11 +79,19 @@ def perform_mutation(cls, root, info: ResolveInfo, /, **data):
         payment = order_info.payment
         manager = get_plugin_manager_promise(info.context).get()
         app = get_app_promise(info.context).get()
+        webhook_events_for_map = WEBHOOK_EVENTS_FOR_ORDER_CONFIRMED
```
IKarbowiak, line 82: Maybe just `webhook_events` instead of `webhook_events_for_map`? I struggled with the meaning of this variable.

IKarbowiak (approved):

IKarbowiak (approved):

kadewu (approved):

kadewu (approved):

zedzior (approved):

Issue #12583, opened 2023-04-17
Webhook refactoring


timuric:
• Identify missing webhooks
• Move webhook away from plugins
• Make it easier to add webhooks

maarcingebala:
For missing webhooks, we will be adding dedicated issues. This task aims at refactoring the webhooks mechanism to move it away from plugins and make it a regular core module. Implementing webhooks as a plugin adds unnecessary complications and makes it harder to add new events.

- Drop WebhookPlugin and move the entire webhook mechanism into a core module.
- Reimplement all webhook events to be called directly, not via the plugin manager.

maarcingebala:
Below is the proposition for moving logic from the WebhookPlugin into the core modules.

**New files structure**

```
saleor
  webhook
    events  // functions that directly trigger events; moved from plugins/webhook/plugin.py
      |_ product.py  // product events
      |_ order.py    // order events
      |_ ...         // other event files
    transport
      |_ async_transport.py  // functions to send async events (such as `trigger_webhooks_async`, `send_webhook_request_async` task etc.)
      |_ sync_transport.py   // functions to send sync events
      |_ utils.py            // helpers needed for the above two modules to work
```

**Todo**

- Move functions from `saleor/plugins/webhook/plugin.py` to `saleor/webhook/events`. Add webhook calls in the logic to use new functions, keeping the manager calls for other plugins to work.
    - All the transport and utility functions can be imported from the plugin at this step.
    - This work can be split into smaller PRs, module by module.
- Move transport functions from `saleor/plugins/webhook/tasks.py` to `saleor/webhook/transport`.
- Split `saleor/graphql/webhook/subscription_types.py` into smaller files (https://github.com/saleor/saleor/issues/13470).

korycins:
Few questions according to the structure. 
- Can you also include here, where test files will be moved? (some of them are in graphql/webhook/, some in `saleor/webhook` and some in `plugins/webhook`. I know it is a detail, but would be good to clean up this part also. 
- celery tasks should be placed in `tasks.py` file (it is celery pattern). The celery task should trigger the logic from `async_transport.py` or `sync_transport.py`
- where do we store subscription_types `saleor/graphql/webhook/subscription_types.py`?
- where do we store the logic that builds the payload based on subscription ?
- Are we ok in making imports in `saleor/webhooks` from `saleor/graphql` - we will need to do that to prepare the subscription types?

maarcingebala:
[...]

PR #16487, opened 2024-08-06
POC: Move webhooks from plugin to core

maarcingebala:
POC of webhook refactoring - move webhook events outside of the webhook plugin into regular core modules. This approach implements the RFC https://github.com/saleor/saleor/issues/12583, where webhook events are represented as classes. 

The goal of this POC is to show examples of two webhook implementations - an async one and a sync one. The webhooks were changed:
- async `product_created`
- sync `shipping_list_methods_for_checkout`

Additional SQL queries in the benchmark are caused by fetching webhooks for the event `shipping_list_methods_for_checkout`, which was not visible before because these benchmarks were not using the webhook plugin.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/webhook/base.py
```diff
@@ -0,0 +1,65 @@
...
+        webhooks: Optional[Iterable[Webhook]] = None,
+        allow_replica: bool = True,
+        legacy_data_generator: Optional[Callable] = None,
+    ):
+        from .transport.asynchronous.transport import trigger_webhooks_async
+        from .utils import get_webhooks_for_event
+
+        if settings.USE_LEGACY_WEBHOOK_PLUGIN and cls.legacy_manager_func is not None:
+            # Trigger webhook via legacy webhook plugin
+            manager = get_plugins_manager(allow_replica)
```
maarcingebala, line 36: Note - this line loads the manager directly, while before, it used a promise:
```
```

saleor/graphql/api.py
```diff
@@ -173,6 +175,36 @@ def serialize_webhook_event(value):
...
+    """Load webhooks types so they can be used in the schema.
+
+    This function inspects webhooks by iterating over subclasses of `AsyncWebhookBase`
+    and `SyncWebhookBase` classes, and then imports the subscription type for each
+    webhook.
+    """
+    from ..webhook.base import AsyncWebhookBase, SyncWebhookBase
+
+    def load_webhooks_for_base_cls(base_cls):
+        for cls in base_cls.__subclasses__():
```
korycins, line 190: How will it work in case of having more advanced relations like:
```
AsyncWebhookBase -> AsyncOrderWebhookBase -> OrderCreated
```
?
Wouldn't be simpler to have something like `@register` that would be used to decorate the webhooks that should be included? 
```python
@register
class OrderCreated(AsyncOrderWebhookBase):
...
```

saleor/graphql/checkout/subscriptions.py
```diff
@@ -0,0 +1,69 @@
...
+        description="Shipping methods that can be used with this checkout."
+        + ADDED_IN_36,
+    )
+
+    @staticmethod
+    @plugin_manager_promise_callback
+    def resolve_shipping_methods(root, info: ResolveInfo, manager):
+        _, checkout = root
+        database_connection_name = get_database_connection_name(info.context)
+        return resolve_shipping_methods_for_checkout(
```
korycins, line 60: Isn't like you have here a circular? `resolve_shipping_methods_for_checkout(` -> `get_all_shipping_methods_list(` -> `ShippingListMethodsForCheckout.list_shipping_methods(`?

saleor/graphql/checkout/subscriptions.py
```diff
@@ -0,0 +1,69 @@
+from django.conf import settings
```
korycins, line 1: Maybe worth to have 1:1 stucture representation that we have for webhook?
So we have `checkout/webhooks/list_shipping_methods.py`, then here we could have `graphql/checkout/subscriptions/list_shipping_methods.py` ?

saleor/graphql/checkout/subscriptions.py
```diff
@@ -0,0 +1,69 @@
...
+
```
korycins, line 2: What about tests related to this subscription? Should we have it here? Like the test for building the subscription payload?

saleor/graphql/api.py
maarcingebala: I tried the solution with `register` but it didn't work well with loading modules into the GraphQL schema - some "register" calls were running after the schema was created, so the subscription was not added to the schema. I think it's related to order of imports but didn't debug this deeper. For now I thought this solution should be enough, and when we need to add such case as in your example, then this function can be easily updated.

saleor/graphql/checkout/subscriptions.py
maarcingebala: I just moved it, there is a task to deprecate this field but it's not in scope of this PR.

saleor/graphql/checkout/subscriptions.py
maarcingebala: Which tests exactly?

saleor/webhook/base.py
```diff
@@ -0,0 +1,67 @@
...
+    # For now this class is only use to distinguish between sync and async webhooks.
+    # When migrating more webhooks to the new approach, this class could implement
+    # generic function to trigger sync webhooks.
+    pass
+
+
+class AsyncWebhookBase(WebhookBase):
+    @classmethod
+    def legacy_payload_func(cls) -> Optional[Callable]:
+        return None
```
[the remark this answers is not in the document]
maarcingebala: I think `None` is OK since this method is optional and only certain webhooks will implement it. Also code easily handles cases when this returns `None`.

fowczarek (approved):

kadewu (approved):

saleor/checkout/webhooks/shipping_list_methods_for_checkout.py
```diff
@@ -0,0 +1,90 @@
...
+        cls,
+        checkout: Checkout,
+        requestor: Optional[Requestor] = None,
+        allow_replica: bool = True,
+    ) -> Iterable[ShippingMethodData]:
+        """Fetch list of external shipping methods for checkout."""
+
+        from ...webhook.payloads import generate_checkout_payload
+        from ...webhook.transport.shipping import (
+            get_cache_data_for_shipping_list_methods_for_checkout,
```
korycins, line 35: This is something related only to this webhook. I don't think we need to still keep this in `transport.shipping`. IMO this should be a part of this webhook implementation. After migrating all webhooks we should have the possibility to fully drop any custom logic from `transport` module and move them to webhook-related files.
(we can also handle this in a separate task, but let's make sure that we have todo point/task about this.)

korycins (approved):

saleor/checkout/webhooks/shipping_list_methods_for_checkout.py
maarcingebala: Yep, that's the plan to do it like that.

saleor/graphql/checkout/subscriptions.py
maarcingebala: I'll move these tests then.

maarcingebala:
FYI - we decided to pause this work for now, as first we need to deliver webhook circuit breaker.

PR #16499, opened 2024-08-06
Add discount distribution flags for tax payload.

zedzior:
I want to merge this change because it adds `distribute_over_subtotal` and `distribute_over_shipping` flags to `TaxableObjectDiscount` and tax app payloads, which will inform Avatax app how to distribute order level discounts.

Issue: https://linear.app/saleor/issue/SHOPX-1145

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/core/types/taxes.py
+++ b/saleor/graphql/core/types/taxes.py
@@ -402,25 +402,22 @@ class TaxableObject(BaseObjectType):
                 if not checkout.discount:
                     return []
 
-                if is_order_level_voucher(checkout_info.voucher):
+                has_order_promotion = has_checkout_order_promotion(checkout_info)
+                has_order_level_voucher = is_order_level_voucher(checkout_info.voucher)
+                # order promotion (currently we handle SUBTOTAL_PROMOTION only) applies
+                # to subtotal only
+                distribute_over_shipping = not has_order_promotion
+
+                if has_order_level_voucher or has_order_promotion:
                     return [
                         {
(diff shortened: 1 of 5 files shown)


saleor/graphql/core/types/taxes.py
```diff
@@ -384,15 +398,27 @@ def resolve_discounts(root: Union[Checkout, Order], info: ResolveInfo):
...
+                # order promotion (currently we handle SUBTOTAL_PROMOTION only) applies
+                # to subtotal only
+                distribute_over_shipping = not has_order_promotion
+
+                if has_order_level_voucher or has_order_promotion:
+                    return [
+                        {
+                            "name": discount_name,
+                            "amount": checkout.discount,
+                            "distribute_over_subtotal": True,
```
korycins, line 416: What is the reason for sending a flag which always is `True`? If you provide True/False flag it also means, that the app needs to have a logic to handle `False` action - app owner will not go through the core code to check that this flag is actually always set to True. Also app would need to cover different combination of these flags.
Also having these two flags means less extensibility. For example, what if you will have a third condition to cover new discount for `TaxableObjectDiscount` ? Then you will need to add third flag (or even more). 
Can't we create a new ENUM like `type` that would store the details of these two flags? You could go even further and provide there a details like: `MANUAL`, `VOUCHER`, `ORDER_PROMOTION` etc. Then each value could get a proper description, of what needs to be done.

Issue #16449, opened 2024-07-29
Allow for localized/translated slugs


delemeator:
## Problem

For multi language e-commerce store I would like to be able to use localized urls for each language. At the moment this requires creating a separate service with slug translation logic which would impact both performance and convenience for the staff (having to translate slugs in another application)

Url design is a matter preference, but it is not uncommon to use loclalized urls and it is recommended by Google:
```
Recommended: Localized words in the URL, if applicable.

https://www.example.com/lebensmittel/pfefferminz
```
https://developers.google.com/search/docs/crawling-indexing/url-structure

---

## 1. General Assumptions

Allow to translate slugs and to query on them

### Acceptance Criteria

- [ ] Allow to translate slugs for product, category, collection and page
- [ ] Allow to query product/category/collection/page by translated slug
- [ ] Ensure backward compatibility - current way of using slug should not be affected by the changes

### Types

Add slug to TranslationInput and PageTranslationInput for use in mutations:
- categoryTranslate
- collectionTranslate
- productTranslate
- pageTranslate

### Queries

Add optional argument slugLanguageCode for queries:
- product
- category
- collection
- page

If slugLanguageCode is provided, those queries should return product with specified slug translation. Otherwise, it would use regular slug field

## 3. Database changes

Add field slug (nullable) to:
- product_collectiontranslation
- product_categorytranslation
- product_producttranslation
- page_pagetranslation

Add unique constraint on languageCode and slug (conditional on slug not null)

Please let me know if you would accept such change. I would be happy to contribute on that

korycins:
Related discord thread: https://discord.com/channels/864066819866624010/1266378575250980955

kadewu:
LGTM!

From my side, when implementing this feature please remember about adding indexes concurrently: https://docs.saleor.io/developer/community/zero-downtime-migrations#adding-the-index-to-the-database

I'm mentioning this cause unique constraint creates `UNIQUE INDEX` and such index needs to be added as explained above.

And maybe small note, no need for explicitly creating conditional unique constraint depending if slug is null or not.

PR #15986, opened 2024-05-14
Fix uniqness on attribute values models

kadewu:
https://linear.app/saleor/issue/PE-223/ir-162-adding-unique-index-fails
Workarounds will be removed in `main` version, but backported to 3.19 version for zerodowntime compatibility.

# Impact

- [ ] New migrations
- [ ] New/Updated API fields or mutations
- [ ] Deprecated API fields or mutations
- [ ] Removed API types, fields, or mutations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

- [ ] Privileged queries and mutations are either absent or guarded by proper permission checks
- [ ] Database queries are optimized and the number of queries is constant
- [ ] Database migrations are either absent or optimized for zero downtime
- [ ] The changes are covered by test cases
- [ ] All new fields/inputs/mutations have proper labels added (`ADDED_IN_X`, `PREVIEW_FEATURE`, etc.)
- [ ] All migrations have proper dependencies
- [ ] All indexes are added concurrently in migrations
- [ ] All RunSql and RunPython migrations have revert option defined

PR #16519, opened 2024-08-08
Allow to translate slugs for product, category, collection and page

pawelromanskidajar:
I want to merge this change because it implements ability to translate slugs from issue [#16449
](https://github.com/saleor/saleor/issues/16449)

* add optional slugLanguageCode parameter for queries product, collection,
  category and page

* modify resolvers for queries product, collection, category and page to query
  by slug translation (using slug and slug_language_code)

* add slug to TranslationInput and PageTranslationInput

* create class BaseTranslateMutationWithSlug that implements handling of slug
  validation for Product/Page/Category/Collection Translate mutations

* add slug column to producttranslations, categorytranslations,
  collectiontranslation and pagetranslations

  - create index on language_code and slug
  - create unique constraint on language_code and slug

* add tests for new logic

# Impact

- [x] New migrations

# Docs

- [x] Link to documentation: https://github.com/saleor/saleor-docs/pull/1280

# Pull Request Checklist


saleor/page/migrations/0029_add_page_translation_slug.py
```diff
@@ -0,0 +1,31 @@
...
+class Migration(migrations.Migration):
+    atomic = False
+
+    dependencies = [
+        ("page", "0028_add_default_page_type"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="pagetranslation",
```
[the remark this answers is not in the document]
pawelromanskidajar: I thought this only applies to fields that are not nullable? When it would be needed to propagate default value?

saleor/page/migrations/0030_slug_translation_unique_constraint.py
```diff
@@ -0,0 +1,26 @@
...
+
+class Migration(migrations.Migration):
+    atomic = False
+
+    dependencies = [
+        ("page", "0029_add_page_translation_slug"),
+    ]
+
+    operations = [
+        AddIndexConcurrently(
```
kadewu, line 14: This needed to be done slightly differently, an example from PR here: https://github.com/saleor/saleor/pull/15986/files#diff-e4e911ec8d5a8922cc7b73fb117b5e1f8755d015d33c17e0f80969f96079c869

The point is to make sure that postgres use the index we created (and unfortunately we need to write raw sql, as there is no support for uniq indexes creation within `AddIndexConcurrently`).

I know the docs are missing this, I will add it soon there as well.

saleor/page/migrations/0030_slug_translation_unique_constraint.py
```diff
@@ -0,0 +1,26 @@
...
+            model_name="pagetranslation",
+            index=BTreeIndex(
+                fields=["language_code", "slug"], name="pagetranslation_slug_idx"
+            ),
+        ),
+        migrations.AddConstraint(
```
kadewu, line 20: And this is not covered in example. but this part you will need to add to the `state` in `migrations.SeparateDatabaseAndState`.

saleor/page/migrations/0030_slug_translation_unique_constraint.py
kadewu: Small tip from me, there is very useful command that allows us to see sql's created from django:
`python manage.py sqlmigrate page 0030`

saleor/page/migrations/0030_slug_translation_unique_constraint.py
pawelromanskidajar: Thanks for the guidance! Should look better now

saleor/page/migrations/0029_add_page_translation_slug.py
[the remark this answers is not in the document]
korycins: Right 🤦 , don't know why I thought that we should have here a default value.

saleor/seo/models.py
```diff
@@ -23,6 +23,7 @@ class SeoModelTranslation(Translation):
     seo_description = models.CharField(
         max_length=300, blank=True, null=True, validators=[MaxLengthValidator(300)]
     )
+    slug = models.SlugField(max_length=255, allow_unicode=True, null=True)
```
korycins, line 26: Not sure if this is proper place for adding `slug`. `SeoModelTranslation` is a representation of `SeoModel`, which where a model can inherit from it. 
In case of adding any new model (or adding new translation to existing model), this will automatically add `slug` field. 
```python

class NewTranslationModel(SeoModelTranslation):
    name = models.CharField(max_length=250, default="")
    # migration will add slug field
    
class NewTranslatableModel(SeoModel):
    name = models.CharField(max_length=250, unique=True)
    # doesn't have slug field
```

IMHO `slug` should be added explicitly to the translation models (or by a new abstract class).

saleor/graphql/translations/mutations/utils.py
```diff
@@ -139,11 +165,42 @@ def perform_mutation(  # type: ignore[override]
...
+            translation_instance = instance.translations.model()
+
+        validate_slug_already_exists(
+            instance, translation_instance, input_data, language_code
+        )
+
+        language_code_lookup = {"language_code": language_code}
+        input = validate_slug_and_generate_if_needed(
+            translation_instance,
+            cls.slugable_field,
```
korycins, line 190: Not sure if this proper flow:

- for non existing translations (CREATE):
    -  when input.slug provided, save it - 👍 
    - **when input.slug is not provided, but input.name is provided, use `name` to create slug** - 👎 
    - when input.slug is not provided and input.name is not provided, save slug as `None`  - 👍 
- for  existing translations (UPDATE):
    - when input.slug is provided, save it  - 👍 
    - when translation.slug==None, and input.name is provided: do not change anything in slug field - 👍 

So right now, we're implicitly using `input.name` to generate the `slug` value, but only in specific condition.
Adding translation is an optional replacement for the specific field. So saving `name` should only save the translation for `name`. Saving `slug` should only save the `slug` field. 
This should simplify your logic, I think.

saleor/seo/models.py
pawelromanskidajar: I see your argument, but I am not sure there can be a use-case where you would create new model with meta description and title were you wouldn't want to translate slug as well. The main reason why I want to translate them is to improve seo. If this is your decision, I will create a new class, but I feel like this separation is unnecessarily adding complexity to codebase (little complexity, but still)

saleor/graphql/translations/mutations/utils.py
pawelromanskidajar: I agree, I wasn't sure with this what is the best way to go. My arguments for this implementation were:
- When creating "regular" slugs, by default it is derived from name/title, so it would be consistent to do the same for translated slugs. The problem is that translated slug is optional which makes the thing a bit awkward
- I didn't want to duplicate validation code or modify existing slug too much, given that this is quite core feature to existing users and I am new to the codebase (which is probably a bad argument)  

I see your point and I will adjust accordingly

saleor/seo/models.py
korycins: Let's get second opinion in that case :) 
@fowczarek  @maarcingebala  What do you think?

fowczarek: > IMHO slug should be added explicitly to the translation models (or by a new abstract class).

I also agree with @korycins.



@pawelromanskidajar Huge kudos for amazing PR :)

korycins (approved):

saleor/product/models.py
```diff
@@ -87,14 +87,26 @@ def __str__(self) -> str:
...
-class CategoryTranslation(SeoModelTranslation):
+class CategoryTranslation(SeoModelTranslationWithSlug):
     category = models.ForeignKey(
         Category, related_name="translations", on_delete=models.CASCADE
     )
     name = models.CharField(max_length=128, blank=True, null=True)
     description = SanitizedJSONField(blank=True, null=True, sanitizer=clean_editor_js)
 
     class Meta:
+        indexes = [
```
kadewu, line 98: Just one small thing from me, this definition is redundant cause we know of this index by defining unique constraint.
So we can remove it from here, and from migration state operations (but we leave of course the database operation of index creation)

saleor/product/models.py
```diff
@@ -259,14 +271,26 @@ def sort_by_attribute_fields() -> list:
...
-class ProductTranslation(SeoModelTranslation):
+class ProductTranslation(SeoModelTranslationWithSlug):
     product = models.ForeignKey(
         Product, related_name="translations", on_delete=models.CASCADE
     )
     name = models.CharField(max_length=250, blank=True, null=True)
     description = SanitizedJSONField(blank=True, null=True, sanitizer=clean_editor_js)
 
     class Meta:
+        indexes = [
```
kadewu, line 282: Same here

saleor/product/models.py
```diff
@@ -709,14 +733,26 @@ class Meta:
...
-class CollectionTranslation(SeoModelTranslation):
+class CollectionTranslation(SeoModelTranslationWithSlug):
     collection = models.ForeignKey(
         Collection, related_name="translations", on_delete=models.CASCADE
     )
     name = models.CharField(max_length=128, blank=True, null=True)
     description = SanitizedJSONField(blank=True, null=True, sanitizer=clean_editor_js)
 
     class Meta:
+        indexes = [
```
kadewu, line 744: And here

kadewu (approved):

saleor/product/models.py
pawelromanskidajar: Thanks for the feedback - fixed

korycins (approved):

kadewu (approved):

pawelromanskidajar:
Sorry, for silly bugs, hope will be ok now

korycins:
> Sorry for silly bugs, hope will be ok now

@pawelromanskidajar FYI, you can use pre-commit to find all issues before making a commit:
https://docs.saleor.io/developer/community/contributing#coding-style

korycins (approved):

kadewu (approved):

maarcingebala (approved):

maarcingebala:
@pawelromanskidajar Thank you for this solid contribution; the PR looks really good! If you're building a project using Saleor, feel free to share it on our Discord at some point — we love to see what our community builds!

PR #16553, opened 2024-08-14
Return proper discounts in order tax calculation static payload.

zedzior:
I want to merge this change because it correct discounts returned in static tax payload for orders. We were accidentally excluding entire order vouchers (order level discount).

Internal issue: https://linear.app/saleor/issue/SHOPX-1217

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/discount/utils.py
+++ b/saleor/discount/utils.py
@@ -92,6 +92,13 @@ def is_shipping_voucher(voucher: Optional[Voucher]):
     return bool(voucher and voucher.type == VoucherType.SHIPPING)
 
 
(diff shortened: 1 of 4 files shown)


korycins (approved):
One thing to consider. Right now you have the same logic to exclude discounts in subscription and in static payload. 
Maybe to protect against any change added in a single place, we could extract this (including the added description), and use a function/method in both places?
Then we will be sure, that we always use the same discounts for both payload types.

8r2y5 (approved):

PR #16559, opened 2024-08-14
Disable Django CacheKeyWarning

wcislo-saleor:


fowczarek (commented):
Could you add a unit test for this?

wcislo-saleor:
The unit test that I added does verify that the code responsible for filtering a warning does the filtering **but** it doesn't verify that the code indeed is called (e.g. from settings). Together with @fowczarek we attempted to make a better test but we weren't able to come up with a working idea.

maarcingebala (approved):

przlada (approved):

fowczarek (approved):

PR #16560, opened 2024-08-14
BACKEND-API[create API's for pos integration with saleor]

DamodarVishwakarma:

I have integrated Stripe POS (Point of Sale) into Saleor using RESTful APIs. This integration allows for seamless management of POS transactions directly within the Saleor admin interface. The integration includes the ability to create, view, and manage orders processed through Stripe POS.

Reason for Merge
This change is crucial because it extends Saleor’s payment capabilities to include Stripe POS, offering users the flexibility to handle in-person transactions alongside online payments. By merging this PR, Saleor will support a wider range of payment methods, enhancing its utility for merchants who operate both online and physical stores.


korycins:
Hey, thanks for the PR but this is not the direction where we want to go. Few arguments:
- Saleor is fully graphql, this PR adds a partial REST API for handling Stripe's POS. 
- all Saleor's built-in payment integrations are built as a Plugin. You can set them up from dashboard, activate, deactivate, or fully remove them from your store by removing the plugin from `settings.PLUGINS`. 
- The proposed changes always provide the stripe POS API as it is built-in directly in Saleor. Not all Saleor's users use Stripe or Stripe POS, so there is no reason to have this by default
- We want to provide general functionalities and flexible API, allowing you to extend Saleor without modifying the base code.
- Your implementation doesn't extend Saleor, it is placed next to Saleor (for example, it will not work with Saleor's dashboard but separate Admin panel will be served by Saleor). 
- We are in the process of migrating any plugins (including payment integrations) into app. In that case, all built-in plugins will be replaced by App.

As explained above, your changes should be done as an App. Having Stripe PoS as App, would also allow you to render the App iframe inside the Dashboard. You can use it to render a configuration form, provide management view (as you did by extending the `router.urls` right now.) etc. Here you have docs for Apps: https://docs.saleor.io/developer/extending/apps/overview

I am closing the PR as explained above.

PR #16561, opened 2024-08-16
Change `Checkout.note` to `Checkout.customer_note` field

Air-t:
I want to merge this change because it adds a `customerNote` field to `Checkout` model to make it consistent with `Order` model.

Fixes: https://linear.app/saleor/issue/SHOPX-330/unify-checkout-note-field-api

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/checkout/migrations/0070_checkout_customer_note.py
```diff
@@ -0,0 +1,50 @@
...
+        if not pks:
+            break
+        yield pks
+        start_pk = pks[-1]
+
+
+def copy_note_to_customer_note(apps, schema_editor):
+    Checkout = apps.get_model("checkout", "Checkout")
+
+    queryset = Checkout.objects.filter(note__isnull=False).order_by("pk")
```
korycins, line 27: Shouldn't be a celery task? This could take a while only to filter out all records.

saleor/graphql/checkout/types.py
```diff
@@ -531,7 +531,14 @@ class Checkout(ModelObjectType[models.Checkout]):
         "saleor.graphql.account.types.Address",
         description="The shipping address of the checkout.",
     )
-    note = graphene.String(required=True, description="The note for the checkout.")
+    customer_note = graphene.String(
+        required=True, description="The customer note for the checkout."
```
korycins, line 535: Added in

saleor/graphql/checkout/types.py
```diff
@@ -531,7 +531,14 @@ class Checkout(ModelObjectType[models.Checkout]):
...
+    )
+    note = graphene.String(
```
korycins, line 537: This needs additional resolver, to return a value from `checkout.customer_note` field

zedzior: @korycins  I think it is not True, in the first version we still update both fields. Furthermore, if migration which copy note -> to customer note will be in progress, then we can return the default value instead of the proper one. Can you please check again if this change is needed?

zedzior (approved):

saleor/graphql/checkout/types.py
korycins: >  the first version

What is a "first version" ? I see only PR to main so assume that this is first version.

You have two fields. Old note, and new customerNote. As this is only field renaming, then both of them should return the same value. So during migrating the data from note to customerNote, fetching customerNote could return different values. So I think, both of the resolvers in that case should relay on both fields. At least temporary until migrating the data.  **BUT** This approach may be debatable in that case as we are changing something that wasn't available for modyfing before **3.21** (checkoutCustomerNoteUpdate was added to main only afaik). 

Then we can assume that all existing checkouts don't have any value in `note` field, so in that case `note`'s resolver could immediately return the value for `customerNote`. 

For both approaches, in my opinion, we should have a data migration in celery task - just to keep in mind that there is a possibility of having millions of checkouts so checking the all records could take some time.

saleor/checkout/models.py
```diff
@@ -86,6 +86,7 @@ class Checkout(models.Model):
         related_name="checkouts",
         on_delete=models.SET_NULL,
     )
+    customer_note = models.TextField(blank=True, default="")
```
IKarbowiak, line 89: I'm not sure if we should add a new field to model. IMO changing the API is enough

fowczarek: I agree with @IKarbowiak. If we want to indicate that `note` is a `customer_note` in the API, we could leave a comment above the `note` field.

zedzior (approved):

saleor/graphql/checkout/types.py
Air-t: We do not proceed with migrations for now, so zero downtime migrations will not be utilized.

korycins (approved):

PR #16574, opened 2024-08-21
fix `DatabaseError` on `checkoutDeliveryMethodUpdate`

Air-t:
I want to merge this change because it fixes a `DatabaseError` when the `checkoutComplete` mutation locks the ' Checkout` instance while someone tries to modify `Checkout` content with another mutation. 

Fixes: https://linear.app/saleor/issue/SHOPX-635/failing-remove-voucher-from-checkout-when-nothing-changed

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/models.py
+++ b/saleor/checkout/models.py
@@ -7,9 +7,9 @@ from typing import TYPE_CHECKING, Optional
 from uuid import uuid4
 
 from django.conf import settings
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.validators import MinValueValidator
-from django.db import DatabaseError, models
+from django.db import models
 from django.utils import timezone
 from django.utils.encoding import smart_str
 from django_countries.fields import Country, CountryField
@@ -275,22 +275,19 @@ class Checkout(models.Model):
             self.set_country(country_code, commit=True)
         return country_code
 
-    def save_or_database_error(self, fields_to_update: Optional[list] = None) -> None:
-        removed_error = "Save with update_fields did not affect any rows."
-        params = {"update_fields": fields_to_update} if fields_to_update else {}
+    def save_if_not_deleted(self, fields_to_update: Optional[list] = None) -> None:
         try:
-            self.save(**params)  # type: ignore[arg-type]
-        except DatabaseError as error:
-            if type(error) is DatabaseError and str(error) == removed_error:
-                raise ValidationError(
-                    {
-                        "checkout": ValidationError(
-                            "Checkout does no longer exists.",
-                            code=CheckoutErrorCode.DELETED.value,
-                        )
-                    }
-                )
-            raise
+            Checkout.objects.get(pk=self.pk)
+        except ObjectDoesNotExist:
+            raise ValidationError(
+                {
+                    "checkout": ValidationError(
+                        "Checkout does no longer exists.",
+                        code=CheckoutErrorCode.DELETED.value,
+                    )
+                }
+            )
+        self.save(**{"update_fields": fields_to_update} if fields_to_update else {})  # type: ignore[arg-type]
 
 
 class CheckoutLine(ModelWithMetadata):
--- a/saleor/checkout/models.py
+++ b/saleor/checkout/models.py
@@ -7,7 +7,6 @@ from typing import TYPE_CHECKING, Optional
 from uuid import uuid4
 
 from django.conf import settings
-from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.validators import MinValueValidator
 from django.db import models
 from django.utils import timezone
@@ -23,7 +22,6 @@ from ..giftcard.models import GiftCard
 from ..permission.enums import CheckoutPermissions
 from ..shipping.models import ShippingMethod
 from . import CheckoutAuthorizeStatus, CheckoutChargeStatus
-from .error_codes import CheckoutErrorCode
 
 if TYPE_CHECKING:
     from ..payment.models import Payment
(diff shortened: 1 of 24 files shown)


saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
```diff
@@ -266,9 +267,24 @@ def _update_delivery_method(
...
-            update_fields=checkout_fields_to_update + invalidate_prices_updated_fields
-        )
+        try:
+            checkout.save(
+                update_fields=(
+                    checkout_fields_to_update + invalidate_prices_updated_fields
+                )
+            )
+        except DatabaseError as error:
+            if type(error) is DatabaseError:
```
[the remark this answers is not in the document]
Air-t: so the best solution that I see is the one where just before `save()` we check if the item still exists. not the perfect one but it should mitigate this issue. I'm not sure capturing `DatabaseError` by its str representation is the proper way to filter errors, while it can be changed.

saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
```diff
@@ -266,9 +267,24 @@ def _update_delivery_method(
         invalidate_prices_updated_fields = invalidate_checkout(
             checkout_info, lines, manager, save=False
         )
-        checkout.save(
-            update_fields=checkout_fields_to_update + invalidate_prices_updated_fields
-        )
+        try:
+            checkout.save(
```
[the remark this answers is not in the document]
korycins: It depends on what we want to cover. If we want to cover a case when the object is deleted because of calling 2 mutations for the same thread (where one is checkoutComplete). Then we should have generic utils, which will make a lock before calling save on the object. If the object doesn't exist, then skip the actions.  This would require changes in the all places where we want to call a save on checkout with flow to handle a case where checkout doesn't exist anymore.
The second option that we could discuss is to clarify if calling multiple mutations for a single checkout is a proper way of handling it. Here we have a problem with `.save()` + `checkoutComplete`, but we could also have a situation where two different `.save()` + `.save()` are called at a similar time and produce strange results. For me, it sounds like a single checkout could be single-threaded

saleor/checkout/models.py
```diff
@@ -273,6 +275,20 @@ def get_country(self):
             self.set_country(country_code, commit=True)
         return country_code
 
+    def save_if_not_deleted(self, fields_to_update: Optional[list] = None) -> None:
```
Air-t, line 278: @korycins I know that we do not solve the problem that was mentioned on slack discussion, however this might solve the `DatabaseError` which is mainly caused by concurrency issue when checkout was removed in other transaction. Not sure if solving the general checkout issue should be resolved by this ticket

saleor/graphql/checkout/mutations/checkout_delivery_method_update.py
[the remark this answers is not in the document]
Air-t: I have adjusted the code to cover the first part of our needs. The second one will get a ticket and should be addressed in the future.

saleor/checkout/utils.py
```diff
@@ -1070,3 +1070,20 @@ def get_address_for_checkout_taxes(
...
+        if not Checkout.objects.select_for_update().filter(pk=checkout.pk).first():
+            raise ValidationError(
+                {
+                    "checkout": ValidationError(
+                        "Checkout does no longer exists.",
+                        code=CheckoutErrorCode.DELETED.value,
+                    )
+                }
+            )
+        checkout.save(**{"update_fields": fields_to_update} if fields_to_update else {})
```
patrys, line 1089: What's the benefit of locking a row immediately before updating it? You could do a savepoint, attempt an update, and if unsuccessful, rollback to the savepoint. Doing it this way is costly and unnecessarily fetches the entire row from the database.

Air-t: My first implementation was similar to what you say - save an object and if `DatabaseError` occurs, raise an error - no locking there. There was a check to filter only `DatabaseError` of certain types and message (I couldn't find a better way of filtering that error).

saleor/checkout/utils.py
```diff
@@ -1070,3 +1070,30 @@ def get_address_for_checkout_taxes(
...
+
+
+def save_checkout_with_update_fields(checkout, update_fields: list[str]) -> None:
+    """Try to save the checkout object with providing fields to update.
+
+    The method is used to handle the DatabaseError that can occur when the save with
+    `update_fields` is performed on a deleted object. In such a case, the method raises
+    a ValidationError with the proper error.
+    """
+    removed_error = "Save with update_fields did not affect any rows."
```
fowczarek, line 1082: Should we add a unit test for base `checkout.save` that returns an error with this message? This should inform us if Django changes this message in any version.

saleor/checkout/utils.py
```diff
@@ -1070,3 +1070,30 @@ def get_address_for_checkout_taxes(
...
+        # https://docs.djangoproject.com/en/5.1/topics/db/transactions/#controlling-transactions-explicitly
+        with transaction.atomic():
+            checkout.save(update_fields=update_fields)
+    except DatabaseError as error:
+        if str(error) == removed_error:
+            raise ValidationError(
+                {
+                    "checkout": ValidationError(
+                        "Checkout does no longer exists.",
+                        code=CheckoutErrorCode.DELETED.value,
```
fowczarek, line 1095: Maybe we should return `NOT_FOUND`?

fowczarek (approved):

saleor/checkout/utils.py
kadewu: IMO I like the idea of separate test only for that on `save` method, as it will be less confusing if such scenario happens in the future.

kadewu (dismissed):

IKarbowiak:
On hold for now.

maarcingebala:
Closing as there is no activity, please reopen if needed.

PR #16582, opened 2024-08-22
Add check if fulfillment does not exceed order line unfulfilled allocation

8r2y5:
I want to merge this change because those changes will prevent the creating simultaneously more than one fulfillment records that could exceed the order line quantity, thus creating a possibility to generate fulfillment records that combined exceed the number of items in the order.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/order/actions.py
+++ b/saleor/order/actions.py
@@ -9,6 +9,7 @@ from uuid import UUID
 from django.conf import settings
 from django.contrib.sites.models import Site
 from django.db import transaction
+from django.db.models import Sum
 
 from ..account.models import User
 from ..core.exceptions import AllocationError, InsufficientStock, InsufficientStockData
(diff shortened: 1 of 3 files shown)


saleor/order/actions.py
```diff
@@ -1120,6 +1121,19 @@ def automatically_fulfill_digital_lines(
         update_order_status(order)
 
 
+def _check_if_fulfillment_quantity_exceeds_order_quantity(order_line, quantity):
+    if order_line.quantity_unfulfilled - quantity < 0:
+        return True
+
+    allocated = order_line.allocations.aggregate(allocated=Sum("quantity_allocated"))[
```
korycins, line 1128: `_check_if_fulfillment_quantity_exceeds_order_quantity` is called for each order_line. Does it mean that we will make allocations for each line, right? So having many order_lines will cause that for each of them we will make an additional db query. Maybe good would be fetching the allocation first for all lines, then passing it to the function? 
Do I miss something? 🤔

korycins:
Closing, as the problem is not related to single mutation. We should figure out a generic flow to handle such cases.

PR #13410, opened 2023-07-16
Added missing description for Translation types.

Smit-Parmar:
I want to merge this change because of adding missing descriptions to translation types.
Part of #13142 

# Impact

* [ ] New migrations
* [ ] New/Updated API fields or mutations
* [ ] Deprecated API fields or mutations
* [ ] Removed API types, fields, or mutations
* [ ] Documentation needs to be updated

# Pull Request Checklist

* [ ] Privileged queries and mutations are either absent or guarded by proper permission checks
* [ ] Database queries are optimized and the number of queries is constant
* [ ] Database migrations are either absent or optimized for zero downtime
* [ ] The changes are covered by test cases

PR #16626, opened 2024-09-02
return ShippingMethodTranslation.name field to be unrequired

jakubkuc:
I want to merge this change because it reverts what was mistakenly changed in #13410 .
The `name` field of `ShippingMethodTranslation` shouldn't be required.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


korycins (commented):
@kadewu It looks ok, but maybe can we have a better PR name and description? Something like 
```
Fix `ShippingMethodTranslation.name` API field to allow return nullable values
```
or something similar. Personally, I would say that it sounds better - but this is only suggestions.

In the description provide some more context. Like we originally added this change unintended to 3.15, it broke some mutations in specific conditions, and as a fix, we need to set it to optional. You can provide an example of a mutation that causes the issue. 
Some could treat this change as a breaking change, so we would need to provide more explanation than we have currently. (PR is linked to release notes).

korycins (approved):

PR #16631, opened 2024-09-03
Drop the invoicing plugin

patrys:
It's not very useful outside of demonstrational purposes and it should be an app. It also happens to pull in some of the heaviest dependencies.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/graphql/schema.graphql
IKarbowiak, line 1: We should also update the descriptions of the `InvoiceCreate` and `InvoiceUpdate` mutations, as these mutations will have no effect. And maybe add a deprecation info in the previous version to inform about that.

patrys: I'm not sure I understand. Neither of these mutations seems to rely on the plugin. We still support the invoicing API for apps.

maarcingebala: These mutations don't use plugins; they just create and delete entities in the database. `invoiceRequest` and `invoiceRequestDelete` do use plugins, but after invoice plugin removal, they'll still use the webhook plugin, so external apps could generate invoices.

maarcingebala (approved):

aniav (approved):

PR #16647, opened 2024-09-04
Release 3.20.24

jakubkuc:
* fix unhandled none cases with shipping method list and add tests (#16642) (7a840dc55d)
* Reduce number of DB queries called for fulfillment logic (#16610) (dfb713a343)
* New release workflows (#16602) (2b680fcbad)
* fix get_product_image_thumbnail when data is None (#16601) (ab37dda2ab)
* Split dataloaders in checkout and webhooks to avoid circular imports (#16619) (21f8ceff75)
* Fix extra deprecation message (#16611) (385f896024)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "saleor",
-  "version": "3.20.23",
+  "version": "3.20.24",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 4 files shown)


maarcingebala (approved):
Make sure to format changelog according to rules in https://www.notion.so/saleor/Pull-requests-and-changelog-conventions-d209528c1059406698fe5c15517d5200

PR #16658, opened 2024-09-05
Circuit breakers for sync webhooks

tomaszszymanski129:
This change implements circuit breakers for sync webhooks.
More context: https://linear.app/saleor/issue/PE-426

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/webhook/circuit_breaker/breaker_board.py
```diff
@@ -0,0 +1,130 @@
...
+                return func(*args, **kwargs)
+
+            app_id = webhook.app.id
+
+            if not self.is_closed(app_id):
+                if not settings.BREAKER_BOARD_DRY_RUN:
+                    # Skip func execution to prevent sending webhooks
+                    return None
+                else:  # noqa: RET505
+                    logger.debug(
```
cmiacz, line 100: Debug print should be removed

saleor/graphql/app/mutations/reenable_sync_webhooks.py
```diff
@@ -0,0 +1,67 @@
...
+        app = cls.get_node_or_error(
+            info, data.get("app_id"), only_type="App", field="appId"
+        )
+        if breaker_board:
+            error = breaker_board.storage.clear_state_for_app(app.id)  # type: ignore[union-attr]
+            if error:
+                raise ValidationError(
+                    {
+                        "app_id": ValidationError(
+                            f"Cannot re-enable sync webhooks for the app {app.name}.",  # type: ignore[union-attr]
```
IKarbowiak, line 56: Try using `cast` to avoid using `# type: ignore[union-attr]`

`app = cast(app_models.App, app)`

saleor/graphql/app/mutations/reenable_sync_webhooks.py
```diff
@@ -0,0 +1,67 @@
...
+    class Arguments:
+        app_id = graphene.ID(
+            description="The app ID to re-enable sync webhooks for.", required=True
+        )
+
+    class Meta:
+        description = (
+            "Re-enable sync webhooks for provided app. "
+            "Can be used to manually re-enable sync webhooks for the app before "
+            "the cooldown period ends."
```
IKarbowiak, line 38: Missing `ADDED_IN` label

saleor/graphql/app/types.py
```diff
@@ -521,6 +537,10 @@ class App(ModelObjectType[models.App]):
         required=True,
     )
     brand = graphene.Field(AppBrand, description="App's brand data.")
+    breaker_state = CircuitBreakerStateEnum(
+        description="Circuit breaker state, if open, sync webhooks operation is disrupted.",
```
IKarbowiak, line 541: Missing `ADDED_IN` label

saleor/webhook/circuit_breaker/breaker_board.py
```diff
@@ -0,0 +1,184 @@
...
+        if settings.BREAKER_BOARD_SYNC_EVENTS == [""]:
+            raise ImproperlyConfigured("BREAKER_BOARD_SYNC_EVENTS cannot be empty.")
+        for event in settings.BREAKER_BOARD_SYNC_EVENTS:
+            if not WebhookEventSyncType.EVENT_MAP.get(event):
+                raise ImproperlyConfigured(
+                    f'Event "{event}" is not supported by circuit breaker.'
+                )
+
+    def exceeded_error_threshold(
+        self, state: CircuitBreakerState, total: int, errors: int
```
IKarbowiak, line 59: What is `total`? Could you use more meaning-full name?

saleor/webhook/circuit_breaker/breaker_board.py
```diff
@@ -0,0 +1,184 @@
...
+            self.failure_min_count_recovery
+            if state == CircuitBreakerState.HALF_OPEN
+            else self.failure_min_count
+        )
+        threshold = (
+            self.failure_threshold_recovery
+            if state == CircuitBreakerState.HALF_OPEN
+            else self.failure_threshold
+        )
+        return errors >= min_count and (errors / total) * 100 >= threshold
```
IKarbowiak, line 71: Could you add a comment to explain it? Or docstring for the method

saleor/graphql/app/types.py
```diff
@@ -52,13 +53,28 @@
...
 )
 
+# TODO: Remove the conditional when unit tests circular import is solved.
+breaker_board = None
+if settings.BREAKER_BOARD_ENABLED:
+    from ...webhook.circuit_breaker.breaker_board import (
+        initialize_breaker_board,
+    )
+
+    breaker_board = initialize_breaker_board()
```
IKarbowiak, line 76: Instead of adding this comment, consider moving the initialization to a dedicated method elsewhere.

saleor/graphql/app/types.py
```diff
@@ -52,13 +55,28 @@
...
+    CircuitBreakerStateEnum,
+)
 from .resolvers import (
     resolve_access_token_for_app,
     resolve_access_token_for_app_extension,
     resolve_app_extension_url,
 )
 
+breaker_board = None
+if settings.BREAKER_BOARD_ENABLED:
```
korycins, line 72: Do you need this? 
`initialize_breaker_board` already has:
```python
     if not settings.BREAKER_BOARD_ENABLED:
        return None
```

saleor/webhook/transport/synchronous/transport.py
```diff
@@ -330,6 +333,10 @@ def trigger_webhook_sync(
     return send_webhook_request_sync(delivery, **kwargs)
 
 
+if breaker_board := initialize_breaker_board():
+    trigger_webhook_sync = breaker_board(trigger_webhook_sync)
```
korycins, line 337: What about `_send_webhook_request_sync`?

saleor/webhook/circuit_breaker/breaker_board.py
```diff
@@ -0,0 +1,221 @@
...
+            event_type: str = args[0]
+            webhook: Webhook = args[2]
+
+            if event_type not in settings.BREAKER_BOARD_SYNC_EVENTS:
+                # Execute webhook without affecting breaker state
+                return func(*args, **kwargs)
+
+            app_id = webhook.app.id
+            state = self.update_breaker_state(app_id)
+            if state == CircuitBreakerState.OPEN:
```
korycins, line 186: What about HALF_OPEN? Right now you treat this as CLOSED. But for circuit breaker, when state is HALF_OPEN, limited number of requests should be invoked.

saleor/webhook/circuit_breaker/breaker_board.py
```diff
@@ -0,0 +1,221 @@
...
+
+if TYPE_CHECKING:
+    from ...webhook.models import Webhook
+
+
+logger = logging.getLogger(__name__)
+
+
+# Time frame for webhook events to be analyzed by the breaker board.
+BREAKER_BOARD_TTL_SECONDS: int = 5 * 60
```
korycins, line 21: Is it expected that we will not be able to set these variables without making an explicit change to the code? 
Personally, I would expect, that these values would be configurable from the settings.py - but maybe I missed some reasons why you decided to not have these values as configurable.

saleor/graphql/app/mutations/reenable_sync_webhooks.py
```diff
@@ -0,0 +1,72 @@
...
+from ...core.mutations import BaseMutation
+from ...core.types import AppError
+from ...utils import get_user_or_app_from_context
+
+logger = logging.getLogger(__name__)
+
+breaker_board = initialize_breaker_board()
+
+
+class ReenableSyncWebhooks(BaseMutation):
```
korycins, line 26: IMHO, the name should start with `App` prefix. As all app related mutations (actually almost of the mutation name starts with the prefix of object related to).

saleor/graphql/app/enums.py
```diff
@@ -20,6 +20,21 @@ def description(enum):
     return None
 
 
+class CircuitBreakerState:
```
korycins, line 23: Would be good to provide a description of each value. Right now, for someone who is not aware of circuit breaker pattern, this will require some googling. We can add some small descriptions to give more input here. Same as done above in line 6.

korycins (commented):
Are we aware that current implementation means: "block all webhooks for single app". This would mean, that in case of having multiple webhooks, when single webhook is failing, we will fully block all webhooks instead the single one that is failing.

saleor/webhook/circuit_breaker/breaker_board.py
cmiacz: That was the initial idea, however @patrys insited on hardcoding these settings.
The goal is to make sure that we don't have a dysfunctional defaults and the only valid config is hidden in production env variables.

saleor/webhook/circuit_breaker/breaker_board.py
cmiacz: In current impelentation HALF_OPEN is passing webhooks similar as CLOSED, but in this state we have different thresholds. In means that it will close faster if webhook endpoint is still down.

In first iteration we don't plan to throttle traffic and execute only some percentage of webhooks. However we may consider extending the implementation in next released if it's necessary.

korycins: Okay, but this `HALF_OPEN` current flow should be included in the enum description right? I mean, we don't tell anywhere that HALF_OPEN, means that we don't throttle the webhooks. So someone could assume, that it needs to call the mutation to "unblock" the webhooks (which are still working without any limitations).

tomaszszymanski129: It will be extra documented but you're right, I'll explicitly state it in the description.

saleor/graphql/app/mutations/app_reenable_sync_webhooks.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+
+    class Meta:
+        description = (
+            "Re-enable sync webhooks for provided app. "
+            "Can be used to manually re-enable sync webhooks for the app before "
+            "the cooldown period ends." + ADDED_IN_321
+        )
+        doc_category = DOC_CATEGORY_APPS
+        permissions = (AppPermission.MANAGE_APPS,)
```
korycins, line 44: Not sure if this is a valid scenario, but do we plan to allow the app, to re-enable its owns webhooks?

saleor/webhook/circuit_breaker/breaker_board.py
korycins: Also, do you plan to have a WebhookAsync event, to let know that the specific app was blocked by CircutBreaker?

saleor/webhook/transport/synchronous/transport.py
tomaszszymanski129: We don't plan to include payment-related webhooks in the MVP. We might decide to add them in the future, but the question will be, "Should we?". Likely the last thing that we want is a payment app, getting disabled by a feature, even if the failure rate is exceeding the threshold.

cmiacz: This is used oudside of `trigger_webhook_sync ` only in celery task so there's no need to wrap it with circuit breaker

saleor/webhook/circuit_breaker/breaker_board.py
cmiacz: @korycins eventual WebhookAsync event implementation wil be handled as separate ticket
https://linear.app/saleor/issue/SCL-727/add-async-webhook-event-for-app-being-blocked-by-circuit-breaker

cmiacz (approved):

saleor/webhook/tests/circuit_breaker/test_breaker_board.py
```diff
@@ -0,0 +1,226 @@
+import pytest
+from django.core.exceptions import ImproperlyConfigured
+
+from saleor.webhook.event_types import WebhookEventSyncType
```
korycins, line 4: Should be relative as the rest of the imports

korycins (approved):

korycins (approved):

cmiacz (approved):

PR #16661, opened 2024-09-05
reliability: fix and prevent logging exceptions

NyanKiyoshi:
This fixes Saleor logging exceptions in `logger.<func>(..., extra={})` - it is known to lead to high memory usage, due to Sentry holding the references to the thread objects. This is especially problematic due to using `asgiref`, which can create thousands of threads over a short time.

This also prevents the issue from being re-introduced by putting in place a Semgrep rule that checks for this pattern.

Follow-up and port of https://github.com/saleor/saleor/pull/16660 (3.20 -> main).

### DIFF
--- a/saleor/payment/gateways/adyen/webhooks.py
+++ b/saleor/payment/gateways/adyen/webhooks.py
@@ -213,7 +213,9 @@ def create_order(payment, checkout, manager):
         )
     except ValidationError as e:
         logger.info(
-            "Failed to create order from checkout %s.", checkout.pk, extra={"error": e}
+            "Failed to create order from checkout %s.",
+            checkout.pk,
+            extra={"error": str(e)},
         )
         return None
     # Refresh the payment to assign the newly created order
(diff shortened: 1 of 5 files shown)


saleor/payment/gateways/adyen/webhooks.py
```diff
@@ -292,7 +294,9 @@ def handle_authorization(notification: dict[str, Any], gateway_config: GatewayCo
             amount.get("value"), amount.get("currency")
         )
     except TypeError as e:
-        logger.exception("Cannot convert amount from minor unit", extra={"error": e})
+        logger.exception(
+            "Cannot convert amount from minor unit", extra={"error": str(e)}
+        )
```
patrys, line 299: This looks wrong, `logger.exception()` automatically captures exception info, it should not need any extra copies.

patrys, line 299:
```suggestion
        logger.exception("Cannot convert amount from minor unit")
```

NyanKiyoshi: Resolved at e889bd62c289f8511ceb2492648dfabc0b0f9d81

przlada (approved):

patrys (approved):

PR #16675, opened 2024-09-10
Only allow Int and Float literals to be passed to the Decimal scalar

patrys:
Saleor did not check what type of AST nodes were passed to Decimal.parse_literal, which meant you could pass all types of literals. Some, like Strings, had a chance of working despite being undocumented. Others, like Objects, would crash Saleor.

⚠️  This change is technically unsafe to backport as-is because it disallows String literals from being passed as Decimal scalars despite them never being officially supported as such.

Examples:

```graphql
mutation {
  example(
    amount1: 10 # ok
    amount2: 10.0 # ok
    amount3: "10.0" # used to work, now disallowed
    amount4: { foo: "bar" } # used to crash, now disallowed
  ) { __typename }
}
```

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/core/scalars.py
+++ b/saleor/graphql/core/scalars.py
@@ -22,10 +22,11 @@ class Decimal(graphene.Float):
 
     @staticmethod
     def parse_literal(node):
-        try:
-            return decimal.Decimal(node.value)
-        except decimal.DecimalException:
-            return None
+        if isinstance(node, (ast.FloatValue, ast.IntValue)):
+            try:
+                return decimal.Decimal(node.value)
+            except decimal.DecimalException:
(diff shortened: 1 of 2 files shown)


IKarbowiak (approved):

wcislo-saleor (approved):

tomaszszymanski129 (approved):

maarcingebala:
@patrys Shouldn't we mention that in the changelog as a potentially breaking change? I know it was a bug but if anyone relied on the fact that the float wrapped in a string worked, then we could inform about such change.

PR #16677, opened 2024-09-11
Fix restoring voucher usage limit when the voucher is removed from draft order

IKarbowiak:
- Fix decreasing voucher usage when the voucher is removed from draft order
- Add e2e for such case
- Fix restriction regarding updating voucher `usage_limit` - do not block the voucher update when exist an order that has a matching `voucher_code`, but an empty `voucher` instance. This might happened when the voucher code was applied to the order, but the corresponding voucher has been deleted. Nest the new voucher has been created with the same code value as the previous voucher code.

Port of https://github.com/saleor/saleor/pull/16664

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


kadewu (approved):
For other reviewers, `voucherCode` is already present on main in file `draft_order_update.py`

PR #16682, opened 2024-09-11
Connect circuit breaker board to target function

tomaszszymanski129:
I want to merge this change because it connects breaker board to `trigger_webhook_sync` func.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/webhook/transport/synchronous/circuit_breaker/breaker_board.py
+++ b/saleor/webhook/transport/synchronous/circuit_breaker/breaker_board.py
@@ -1,12 +1,14 @@
 import time
+from typing import TYPE_CHECKING
 
-from saleor.core import EventDeliveryStatus
-from saleor.core.models import EventDelivery
 from saleor.webhook.event_types import WebhookEventSyncType
 from saleor.webhook.transport.synchronous.circuit_breaker.storage import (
     Storage,
 )
 from saleor.webhook.transport.utils import WebhookResponse
+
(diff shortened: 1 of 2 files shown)


saleor/webhook/transport/synchronous/circuit_breaker/breaker_board.py
```diff
@@ -50,21 +58,24 @@ def register_success(self, app_id: int):
...
-            app = delivery.webhook.app
+            app = webhook.app
 
             if not self.is_closed(app.id):
                 return WebhookResponse(content=""), None
 
-            response, data = func(*args, **kwargs)
-            if response.status == EventDeliveryStatus.FAILED:
+            response = func(*args, **kwargs)
+            if response is None:
```
wcislo-saleor, line 74: We need to be absolutely sure that successful webhook executions never return `None` as a response.

wcislo-saleor (approved):

saleor/webhook/transport/synchronous/circuit_breaker/breaker_board.py
tomaszszymanski129: yes, webhook execution will return `None` only if there is an actual failure, in case of success, even with empty data, it will either be `[]` or `{}`, thus not explicitly `None`.

PR #16683, opened 2024-09-11
Ignore conflicts on stock save in `ProductVariantBulkUpdate`

kadewu:
Allow to create stocks in `errorPolicy` `REJECT_FAILED_ROWS` and `IGNORE_FAILED`.
The issue may happen if we concurrently have thread race to create stock, so we don't catch the existence of the Stock when validating input but once we try to save, the stock was already created.

The mutation will still fail when `errorPolicy` is `REJECT_EVERYTHING`.
This is temporary solution as we may need to come with general approach for all `bulk_mutations` and all `bulk_create` saves we are doing.

Other option would be to do `select_for_update` inside transaction when we try to save with `bulk_create` but that will increase the time execution and memory usage of already heavy mutation.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/tests/mutations/test_product_variant_bulk_update.py
+++ b/saleor/graphql/product/tests/mutations/test_product_variant_bulk_update.py
@@ -2,14 +2,20 @@ from unittest import mock
 from unittest.mock import patch
 
 import graphene
+import pytest
 from django.test import override_settings
 
 from .....discount.models import PromotionRule
 from .....discount.utils.promotion import get_active_catalogue_promotion_rules
+from .....graphql.core.enums import ErrorPolicyEnum
+from .....graphql.product.bulk_mutations.product_variant_bulk_update import (
+    ProductVariantBulkUpdate,
+)
 from .....graphql.webhook.subscription_payload import get_pre_save_payload_key
 from .....product.error_codes import ProductVariantBulkErrorCode
 from .....product.models import ProductChannelListing
 from .....tests.utils import flush_post_commit_hooks
+from .....warehouse.models import Stock
 from .....webhook.event_types import WebhookEventAsyncType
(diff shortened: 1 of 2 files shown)


IKarbowiak (approved):

saleor/graphql/product/tests/mutations/test_product_variant_bulk_update.py
```diff
@@ -136,6 +147,90 @@ def test_product_variant_bulk_update(
...
+        "productId": product_id,
+        "variants": variants,
+        "errorPolicy": error_policy,
+    }
+
+    # when
+    staff_api_client.user.user_permissions.add(permission_manage_products)
+    response = staff_api_client.post_graphql(
+        PRODUCT_VARIANT_BULK_UPDATE_MUTATION, variables
+    )
```
zedzior, line 219: Because of using the `mocked_save`, in fact we skip the `save_variants` function execution. In order to mimic concurrency I suggest to use `before_after` and do this:
```
    with before_after.before(
        "saleor.graphql.product.bulk_mutations.product_variant_bulk_update.ProductVariantBulkUpdate.save_variants",
        add_stock_before_save
    ):
        response = staff_api_client.post_graphql(
            PRODUCT_VARIANT_BULK_UPDATE_MUTATION, variables
        )
```

fowczarek (approved):

zedzior (approved):

saleor/graphql/product/tests/mutations/test_product_variant_bulk_update.py
kadewu: Unfortunately `before_after` is currently bugged with `classmethod`s. So I will leave this as it is.

PR #16688, opened 2024-09-12
Make `amount` optional in `transactionEventReport` for `REFUND_REVERSE`, `CHARGEBACK`, `INFO`, and all `*_FAILURE` events

IKarbowiak:
Make `amount` optional in `transactionEventReport` for all following events:
- `REFUND_REVERSE`
- `CHARGE_BACK`
- `INFO`
- `AUTHORIZATION_FAILURE`
- `CHARGE_FAILURE`
- `REFUND_FAILURE`
- `CANCEL_FAILURE`

When the amount is not provided we are trying to deduce the amount value according to following assumptions:
- In case of missing amount for event INFO, use 0
- In case of missing amount for *_FAILURE we need to take the amount from the event *_SUCCESS or *_REQUEST with the same pspReference. When looking at event, use latest one sorted by createdAt fileld.
  - REFUND_FAILURE -> we look at REFUND_SUCCESS and REFUND_REQUEST then at CHARGE_SUCCESS and CHARGE_FAILURE and CHARGE_REQUEST with the same pspReference
   - CHARGE_FAILURE -> we look at CHARGE_SUCCESS and CHARGE_REQUEST then at AUTHORIZATION_SUCCESS and AUTHORIZATION_FAILURE and AUTHORIZATION_FAILURE with the same pspReference
  - AUTHORIZATION_FAILURE -> we look at AUTHORIZATION_SUCCESS, AUTHORIZATION_REQUEST with the same pspReference
  - CANCEL_FAILURE -> we look at CANCEL_REQUESTand CANCEL_SUCCESS then at AUTHORIZATION_SUCCESS and AUTHORIZATION_FAILURE and AUTHORIZATION_REQUEST with the same pspReference
- In case of REFUND_REVERSE, we need to take the amount from REFUND_SUCCESS with the same pspReference.
- In case of CHARGEBACK we need to take the amount from CHARGE_SUCCESS with the same pspReference. 
- If the specific event for the pspReference doesn't exist, raise exception
- If amount is provided, but an event already exists with the same pspReference and it's not the same, an exception is raised (leave current behavior as is)

# Impact

- [ ] New migrations

# Docs

- [x] Link to documentation: https://github.com/saleor/saleor-docs/pull/1316

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/payment/mutations/transaction/transaction_event_report.py
+++ b/saleor/graphql/payment/mutations/transaction/transaction_event_report.py
@@ -82,7 +82,14 @@ class TransactionEventReport(ModelMutation):
             description="Current status of the event to report.",
         )
         amount = PositiveDecimal(
-            description="The amount of the event to report.", required=False
+            description=(
+                "The amount of the event to report. \n\nRequired for all `REQUEST`, "
+                "`SUCCESS`, `ACTION_REQUIRED`, and `ADJUSTMENT` events. For other events, "
+                "the amount will be calculated based on the previous events with "
+                "the same pspReference. "
+                "If not possible to calculate, the mutation will return an error."
+            ),
+            required=False,
         )
         time = DateTime(
             description=(
--- a/saleor/graphql/payment/mutations/transaction/transaction_event_report.py
+++ b/saleor/graphql/payment/mutations/transaction/transaction_event_report.py
@@ -17,13 +17,14 @@ from .....order.utils import (
     calculate_order_granted_refund_status,
     updates_amounts_for_order,
 )
-from .....payment import TransactionEventType
+from .....payment import OPTIONAL_AMOUNT_EVENTS, TransactionEventType
 from .....payment import models as payment_models
 from .....payment.transaction_item_calculations import recalculate_transaction_amounts
 from .....payment.utils import (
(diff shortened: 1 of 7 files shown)


saleor/graphql/payment/mutations/transaction/transaction_event_report.py
```diff
@@ -81,7 +82,7 @@ class Arguments:
             description="Current status of the event to report.",
         )
         amount = PositiveDecimal(
-            description="The amount of the event to report.", required=True
+            description="The amount of the event to report.", required=False
```
kadewu, line 85: I'm wondering if the information about when `amount` is required should be included in description. WDYT?

kadewu (approved):

saleor/graphql/payment/mutations/transaction/transaction_event_report.py
IKarbowiak: Good point!

IKarbowiak: I updated it

zedzior (approved):

jakubkuc (approved):

PR #16696, opened 2024-09-12
Fix order-level discounts handling when using tax app for tax calculation.

zedzior:
I want to merge this change because order calculation returns wrong results when the tax app is used for tax calculation and an order has order-level discounts.

In general, this PR does four things:
1. Make sure Saleor does not propagate the order level discounts to order lines when using tax app. Flat rates and plugin requires the propagation, but tax app does it itself.
2. Shipping vouchers, despite being associated with an order, decrease the base price, similar to line-level discounts. Therefore they shouldn't be propagated together with other order-level discounts.
3. Make sure all the tests inside `test_fetch_order_prices.py` are based on the same configuration (have flat rates tax strategy)
4. Make sure order-level discounts are not combined

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/order/base_calculations.py
```diff
@@ -95,11 +95,10 @@ def propagate_order_discount_on_order_prices(
     """
     base_subtotal = base_order_subtotal(order, lines)
     subtotal = base_subtotal
-    shipping_price = order.undiscounted_base_shipping_price
+    shipping_price = order.base_shipping_price
```
IKarbowiak, line 98: Cannot add comment in line `139`, but there `order.undiscounted_base_shipping_price` should be used. Otherwise, you will apply the manual discount on the discounted shipping price.
Also, we need a test for that

saleor/order/tests/test_apply_order_discount.py
```diff
@@ -179,48 +179,6 @@ def test_apply_order_discounts_manual_discount(order_with_lines):
     assert order_discount.amount_value == discount_amount
 
 
-def test_apply_order_discounts_shipping_voucher(
```
IKarbowiak, line 182: We shouldn't remove this test, we should update the asserts

saleor/order/tests/test_fetch_order_prices.py
```diff
@@ -23,6 +23,16 @@ def order_with_lines(order_with_lines):
     return order_with_lines
 
 
+@pytest.fixture
+def channel_USD(channel_USD):
```
IKarbowiak, line 27: We already have such fixture, IMO, you shouldn't override existing one, you should create a new one that will be fitting your requirements

saleor/order/tests/test_fetch_order_prices_tax_app.py
```diff
@@ -0,0 +1,656 @@
...
+
+
+@pytest.fixture
+def order_with_lines(order_with_lines):
+    order_with_lines.status = OrderStatus.UNCONFIRMED
+    return order_with_lines
+
+
+@pytest.fixture
+def channel_USD(channel_USD):
```
IKarbowiak, line 35: Same here

saleor/order/tests/test_fetch_order_prices_tax_app.py
```diff
@@ -0,0 +1,656 @@
...
+from ...webhook.transport.synchronous.transport import (
+    create_delivery_for_subscription_sync_event,
+)
+from .. import OrderStatus
+from ..calculations import fetch_order_prices_if_expired
+from ..utils import create_order_discount_for_order
+
+
+@pytest.fixture
+def order_with_lines(order_with_lines):
```
IKarbowiak, line 29: Same here. Create new fixture

saleor/order/tests/test_fetch_order_prices_tax_app.py
```diff
@@ -0,0 +1,656 @@
...
+
+
+@pytest.fixture
+def subscription_calculate_taxes_for_order(subscription_webhook):
+    return subscription_webhook(
+        TAXES_SUBSCRIPTION_QUERY, WebhookEventSyncType.ORDER_CALCULATE_TAXES
+    )
+
+
+def test_fetch_order_prices_catalogue_promotion(
```
IKarbowiak, line 52: We should also have such test for `percentage` discount

IKarbowiak (commented):
IMO, we should also add e2e tests for those combinations, to ensure that the manual discount is properly applied, when other discount exist

saleor/order/calculations.py
```diff
@@ -141,11 +141,17 @@ def _recalculate_prices(
 
     order.tax_error = None
 
-    # propagate the order level discount on the prices without taxes.
+    # Flat rates, plugin and tax exemption strategies require to propagate order-level
+    # discounts to order lines. Tax app does it itself.
+    is_tax_app_strategy = (
```
kadewu, line 146: I wonder if this should be called differently like just e.g. `using_tax_app`. Adding `_strategy` may be misleading, as tax_app_strategy includes usage of plugin.

kadewu (dismissed):

saleor/order/tests/test_fetch_order_prices_tax_app.py
zedzior: There is much more missing combination than this one. I created tests to cover updated logic

saleor/order/tests/test_fetch_order_prices.py
zedzior: I want to override it. Please notice I don't use it as a test argument. I need to override it in order to change other fixtures behaviour

saleor/order/tests/test_fetch_order_prices_tax_app.py
zedzior: I want to override it. Please notice I don't use it as a test argument. I need to override it in order to change other fixtures behaviour. This way I avoid saving channel settings (extra db calls, time of execution) for each test.

saleor/order/tests/test_fetch_order_prices_tax_app.py
zedzior: I want to override it. Please notice I don't use it as a test argument. I need to override it in order to change other fixtures behaviour. This way I avoid saving channel settings (extra db calls, time of execution) for each test.

saleor/order/tests/test_fetch_order_prices.py
IKarbowiak: But it's misleading. I will look for the `channel_USD` fixture in the `tests/fixtures.py` file. It will be confusing, that it was overridden in this file. 

> This way I avoid saving channel settings (extra db calls, time of execution) for each test.
In which place exactly? And still, I'm not sure if it's worth, the solution looks like anty-pattern for me.

saleor/webhook/tests/subscription_webhooks/test_create_deliveries_for_taxes_with_fetch_order_prices.py
IKarbowiak, line 1: - Test names don't fit what it's testing IMO.
- IMO we shouldn't override existing fixtures, we should create new ones

IKarbowiak (commented):
Also, please add e2e tests that will cover the scenarios

saleor/webhook/tests/subscription_webhooks/test_create_deliveries_for_taxes_with_fetch_order_prices.py
zedzior: - fixtures changed
- I merged `test_create_deliveries_for_taxes_with_fetch_order_prices.py` with `test_create_deliveries_for_taxes`
- e2e added

saleor/order/tests/test_fetch_order_prices.py
```diff
@@ -2637,18 +2696,298 @@ def test_fetch_order_prices_voucher_shipping_percentage(
...
+    assert order.total_net_amount == expected_total
+    assert order.total_gross_amount == round(expected_total * tax_rate, 2)
+    # FIXME: If discounts exceed total value, undiscounted prices are set to 0
+    # assert (
+    #     order.undiscounted_total_net_amount
+    #     == undiscounted_subtotal + undiscounted_shipping_price
+    # )
+    # assert order.undiscounted_total_gross_amount == round(
+    #     (undiscounted_subtotal + undiscounted_shipping_price) * tax_rate, 2
+    # )
```
IKarbowiak, line 2993: What about that?

IKarbowiak (dismissed):

saleor/order/tests/test_fetch_order_prices.py
zedzior: Here is tax: https://linear.app/saleor/issue/SHOPX-1390

IKarbowiak: Could you add a task number in the comment?

zedzior: done

IKarbowiak (approved):

kadewu (approved):

szczecha (approved):

kadewu (approved):

IKarbowiak (approved):

PR #16701, opened 2024-09-13
Add workflow to check patches

maarcingebala:
The action fetches code from the PR's source branch, takes the target branch, and tries to fetch and apply patches using PR's target branch. Example: if a PR targets a version branch `3.20`, the action would checkout to `3.20` branch in the patches repo and try to apply patches from that branch. If patches apply, the action succeeds, otherwise it returns an error, and indicates that patches need to be adjusted.

Other assumptions:
- Action runs on all branches.
- Action runs only when the label is applied: `check patches` (will be changed to all PRs when the workflow is more tested).

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/.github/workflows/check-patches.yml
+++ b/.github/workflows/check-patches.yml
@@ -39,4 +39,4 @@ jobs:
         id: apply_patches
         run: |
           cd workflows-test
-          git apply ../saleor-multitenant/patches/**.patch
+          git apply ../saleor-multitenant/patches/**.patch > /dev/null 2>&1
--- a/.github/workflows/check-patches.yml
+++ b/.github/workflows/check-patches.yml
@@ -13,7 +13,7 @@ name: Test Patches
 
 on:
   pull_request:
-    types: [labeled, opened, synchronize, unlabeled]
+    types: [labeled, opened, synchronize]
 
 jobs:
   apply-patches:
--- a/.github/workflows/check-patches.yml
+++ b/.github/workflows/check-patches.yml
@@ -1,0 +1,42 @@
+# The action fetches code from the PR's source branch, takes the target branch, and
+# tries to fetch and apply patches using PR's target branch.
+# Example: if a PR targets a version branch `3.20`, the action would checkout to `3.20`
+# branch in the patches repo and try to apply patches from that branch. If patches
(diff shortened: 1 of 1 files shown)


.github/workflows/check-patches.yml
```diff
@@ -0,0 +1,42 @@
...
+#
+# - Action runs on all branches.
+# - Action runs only when label is applied: `check patches` (will be changed to all PRs
+# when workflow is more tested).
+
+name: Test Patches
+
+on:
+  pull_request:
+    types: [labeled, opened, synchronize, unlabeled]
```
tomaszszymanski129, line 16: `unlabeled` seems unnecessary?

.github/workflows/check-patches.yml
```diff
@@ -0,0 +1,42 @@
...
+          repository: saleor/saleor-multitenant
+          token: ${{ secrets.GITHUB_TOKEN }}
+          path: saleor-multitenant
+          ref: ${{ github.base_ref }}
+
+      - name: Apply patches
+        id: apply_patches
+        run: |
+          cd workflows-test
+          git apply ../saleor-multitenant/patches/**.patch
```
tomaszszymanski129, line 42: perhaps we wanna skip the output? something like `> /dev/null 2>&1` should do it

.github/workflows/check-patches.yml
maarcingebala: Original version I wrote used a label to skip the check, so it's a leftover from that solution. Now it's not needed indeed.

.github/workflows/check-patches.yml
maarcingebala: Good idea, I made the changes and the output would now look like this:
image

NyanKiyoshi (changes requested):
Sent a comment at https://saleorcommerce.slack.com/archives/C06DNJPFDJB/p1726227492247399

.github/workflows/check-patches.yml
maarcingebala, line 16:
```suggestion
    types: [labeled, opened, synchronize]
```

tomaszszymanski129 (approved):

IKarbowiak (approved):

PR #16737, opened 2024-09-24
Log the warning when checkout and order are missing for provided payment in Stripe webhooks

IKarbowiak:
Log the warning instead of raising `ValueError` when both order and checkout are missing for the provided payment in Stipe webhooks.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/payment/gateways/stripe/webhooks.py
+++ b/saleor/payment/gateways/stripe/webhooks.py
@@ -115,9 +115,11 @@ def _channel_slug_is_different_from_payment_channel_slug(
     elif order is not None:
         return channel_slug != order.channel.slug
     else:
-        raise ValueError(
-            "Both payment.checkout and payment.order cannot be None"
-        )  # pragma: no cover
+        logger.warning(
+            "Both payment.checkout and payment.order cannot be None",
+            extra={"payment_id": payment.id},
+        )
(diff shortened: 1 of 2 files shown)


zedzior (approved):

saleor/payment/gateways/stripe/webhooks.py
```diff
@@ -115,9 +115,11 @@ def _channel_slug_is_different_from_payment_channel_slug(
     elif order is not None:
         return channel_slug != order.channel.slug
     else:
-        raise ValueError(
-            "Both payment.checkout and payment.order cannot be None"
-        )  # pragma: no cover
+        logger.warning(
+            "Both payment.checkout and payment.order cannot be None",
+            extra={"payment_id": payment.id},
```
zedzior, line 120: IDs in logs - this is something worth discussing. Should we use a camel case or snake case format? It should be consistent between apps and core logs in order to allow grouping the logs.

IKarbowiak: Yes, I considered that, but I checked other logs in Stripe plugin and all return DB id, so I decided to do this in the same way.

PR #16738, opened 2024-09-24
Do not stack manual line discounts with line-level vouchers.

zedzior:
I want to merge this change because manual line discounts should override voucher line discounts, same as it behaves with catalog promotions.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,3 +29,4 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fixed a bug when saving webhook payload to Azure Storage - #16585 by @delemeator
 - Added validation for tax data received from tax app - #16720 by @zedzior
 - Fixed order-level discounts handling when using tax app for tax calculation - #16696 by @zedzior
+- Fixed bug when manual line discount doesn't override line-level vouchers - #16738 by @zedzior
--- a/saleor/tests/e2e/orders/discounts/test_order_voucher_specific_product_and_manual_line_discount.py
+++ b/saleor/tests/e2e/orders/discounts/test_order_voucher_specific_product_and_manual_line_discount.py
@@ -1,0 +1,331 @@
+import pytest
+
+from .....product.tasks import recalculate_discounted_price_for_products_task
+from ... import DEFAULT_ADDRESS
+from ...product.utils.preparing_product import prepare_product
+from ...shop.utils.preparing_shop import prepare_shop
+from ...taxes.utils import update_country_tax_rates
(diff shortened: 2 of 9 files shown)


patrys:
It's tagged with `skip changelog`, but it seems like a significant change in Saleor's behavior.

CHANGELOG.md
```diff
@@ -27,3 +27,4 @@ All notable, unreleased changes to this project will be documented in this file.
 - Add slugs to product/category/collection/page translations. Allow to query by translated slug - #16449 by @delemeator
 - Fixed a crash when the Decimal scalar is passed a non-normal value - #16520 by @patrys
 - Fixed a bug when saving webhook payload to Azure Storage - #16585 by @delemeator
+- Fixed bug when manual line discount doesn't override line-level vouchers - #16738 by @zedzior
```
fowczarek, line 30: Should it be part of `breaking changes`? 
If yes, I think we should add at least one example of how the discount behavior will change.

fowczarek (commented):
Should we mentor those changes in the upgrading guide in our docs?

IKarbowiak (dismissed):
Could we also add e2e test for this scenario?

CHANGELOG.md
zedzior: Thanks for the remark. I would say this was a temporary bug rather than breaking change. We have never documented that manual line discount can be combined with other discounts. Since we introduced `sales`, the manual line discount has overridden other discounts. This will be documented in upcoming `Price Calculation` section: https://github.com/saleor/saleor-docs/pull/1328/files

fowczarek (approved):

saleor/tests/e2e/orders/discounts/test_order_voucher_specific_product_and_manual_line_discount.py
```diff
@@ -0,0 +1,331 @@
...
+        e2e_staff_api_client,
+        voucher_id,
+        channel_listing,
+    )
+
+    return voucher_code, voucher_discount_value
+
+
+@pytest.mark.e2e
+def test_draft_order_with_voucher_specific_product_and_manual_line_discount(
```
szczecha, line 60: Add test case ID at the end `test_draft_order_with_voucher_specific_product_and_manual_line_discount_CORE_0250`

szczecha (approved):

szczecha (approved):

IKarbowiak (approved):

PR #16775, opened 2024-09-27
Fix missing voucher_code assignment in order

Dv-del:
- Corrected missing assignment to voucher_code in checkout process.
- Ensures voucher is properly tracked during order creation.

I want to merge this change because, without the assignment of the voucher_code, the system loses track of the voucher code during order creation, making it hard to track which voucher code is being used in an order.

# Impact

- [ ] New migrations

# Docs

No docs updates required.

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/complete_checkout.py
+++ b/saleor/checkout/complete_checkout.py
@@ -1231,7 +1231,9 @@ def _create_order_from_checkout(
 
     # voucher
     voucher = checkout_info.voucher
-    voucher_code = checkout_info.checkout.voucher_code
+    voucher_code = (
+        checkout_info.voucher_code.code if checkout_info.voucher_code else None
+    )
 
     # shipping
     undiscounted_base_shipping_price = base_checkout_undiscounted_delivery_price(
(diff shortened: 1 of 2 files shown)


fowczarek (commented):
Hello, Thanks for Your contribution!
Could you also extend unit tests to check it?

saleor/checkout/complete_checkout.py
```diff
@@ -1231,6 +1231,7 @@ def _create_order_from_checkout(
 
     # voucher
     voucher = checkout_info.voucher
+    voucher_code = checkout_info.checkout.voucher_code
```
zedzior, line 1234:
```suggestion
    voucher_code = (
        checkout_info.voucher_code.code if checkout_info.voucher_code else None
    )
```
Checkouts are processed with prefetched data to prevent race conditions. Let's use the code from `checkout_info.voucher_code` instance.

Dv-del: Thanks @zedzior

zedzior (commented):
Hey, as @fowczarek previously requested, can you please extend existing tests to prove it works as expected? I would suggest to look for voucher related tests inside `saleor/graphql/checkout/tests/mutations/test_checkout_complete_with_transactions.py`

korycins (approved):

kadewu (approved):

tomaszszymanski129 (approved):

PR #16785, opened 2024-10-01
Automatic confirm paid checkouts

IKarbowiak:
Introduce automatic checkout completion.

- The automatic completion might happen:
    - as a result of the `TransactionCreate` mutation, when the created transaction covers the checkout total
    - as a result of the `TransactionEventReport` mutation, when the reported event causes the checkout total to be covered
    - as a result of the `TransactionUpdate` mutation, when the updates causes the checkout total to be covered
    - as a result of `TransactionInitialize`
    - as a result of `TransactionProcess`
    - as a result of `TransactionRequestAction`, after processing the response for sync webhooks: `TRANSACTION_CHARGE_REQUESTED`
- The checkout is considered ready to complete when:
    - `checkout.authorization_status` is `FULL` .
- Log the whole process of the automatic completion

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/actions.py
+++ b/saleor/checkout/actions.py
@@ -247,7 +247,7 @@ def transaction_amounts_for_checkout_updated(
 
     channel = checkout_info.channel
     if (
-        channel.automatically_complete_paid_checkouts
+        channel.automatically_complete_fully_paid_checkouts
         and checkout.authorize_status == CheckoutAuthorizeStatus.FULL
     ):
         complete_checkout(
--- a/saleor/checkout/actions.py
+++ b/saleor/checkout/actions.py
@@ -207,7 +207,7 @@ def transaction_amounts_for_checkout_updated(
     user: Optional["User"],
     app: Optional["App"],
 ):
-    from .complete_checkout import complete_checkout
+    from .tasks import automatic_checkout_completion_task
 
     if not transaction.checkout_id:
         return
@@ -250,12 +250,6 @@ def transaction_amounts_for_checkout_updated(
         channel.automatically_complete_fully_paid_checkouts
         and checkout.authorize_status == CheckoutAuthorizeStatus.FULL
     ):
-        complete_checkout(
-            manager=manager,
-            checkout_info=checkout_info,
-            lines=lines,
-            payment_data={},
-            store_source=False,
-            user=user,
-            app=app,
-        )
+        user_id = user.id if user else None
+        app_id = app.id if app else None
+        automatic_checkout_completion_task.delay(checkout.pk, user_id, app_id)
--- a/saleor/checkout/actions.py
+++ b/saleor/checkout/actions.py
@@ -12,7 +12,7 @@ from ..core.utils.events import (
 from ..payment.models import TransactionItem
 from ..webhook.event_types import WebhookEventAsyncType, WebhookEventSyncType
 from ..webhook.utils import get_webhooks_for_multiple_events
-from . import CheckoutChargeStatus
+from . import CheckoutAuthorizeStatus, CheckoutChargeStatus
 from .calculations import fetch_checkout_data
 from .fetch import (
     CheckoutInfo,
@@ -24,7 +24,8 @@ from .models import Checkout
 from .payment_utils import update_refundable_for_checkout
 
 if TYPE_CHECKING:
-    from ..account.models import Address
+    from ..account.models import Address, User
+    from ..app.models import App
     from ..webhook.models import Webhook
 
 from ..plugins.manager import PluginsManager
@@ -203,7 +204,11 @@ def update_last_transaction_modified_at_for_checkout(
 def transaction_amounts_for_checkout_updated(
     transaction: TransactionItem,
     manager: "PluginsManager",
+    user: Optional["User"],
+    app: Optional["App"],
 ):
+    from .complete_checkout import complete_checkout
+
     if not transaction.checkout_id:
         return
     checkout = cast(Checkout, transaction.checkout)
@@ -238,4 +243,19 @@ def transaction_amounts_for_checkout_updated(
             event_name=WebhookEventAsyncType.CHECKOUT_FULLY_PAID,
             checkout_info=checkout_info,
             lines=lines,
+        )
+
+    channel = checkout_info.channel
+    if (
+        channel.automatically_complete_paid_checkouts
+        and checkout.authorize_status == CheckoutAuthorizeStatus.FULL
+    ):
+        complete_checkout(
+            manager=manager,
+            checkout_info=checkout_info,
+            lines=lines,
+            payment_data={},
+            store_source=False,
+            user=user,
+            app=app,
         )
--- a/saleor/checkout/tasks.py
+++ b/saleor/checkout/tasks.py
@@ -1,13 +1,20 @@
 import logging
 from decimal import Decimal
 
+import graphene
 from celery.utils.log import get_task_logger
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.db.models import Exists, OuterRef, Q, QuerySet, Subquery
 from django.utils import timezone
 
+from ..account.models import User
+from ..app.models import App
 from ..celeryconf import app
 from ..payment.models import TransactionItem
+from ..plugins.manager import get_plugins_manager
+from .complete_checkout import complete_checkout
+from .fetch import fetch_checkout_info, fetch_checkout_lines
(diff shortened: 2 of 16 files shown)


saleor/checkout/tasks.py
```diff
@@ -109,3 +116,57 @@ def delete_expired_checkouts(
...
+                "checkout_id": checkout_id,
+                "error": str(error),
+            },
+        )
+    else:
+        task_logger.info(
+            "Automatic checkout completion succeeded for checkout: %s.",
+            checkout_id,
+            extra={"checkout_id": checkout_id},
+        )
```
zedzior, line 172: Why do you use `else` here, instead of logging the success directly under `try` statement? If any other error than `ValidationError` occur during `complete_checkout` we will log success.

saleor/checkout/tests/test_tasks.py
```diff
@@ -454,3 +461,87 @@ def test_aborts_deleting_checkouts_when_invocation_count_exhausted(
...
+    assert not Checkout.objects.filter(pk=checkout_pk).exists()
+    assert Order.objects.filter(checkout_token=checkout_pk).exists()
+
+    checkout_id = graphene.Node.to_global_id("Checkout", checkout_pk)
+    assert len(caplog.records) == 2
+    assert caplog.records[0].message == (
+        f"Automatic checkout completion triggered for checkout: {checkout_id}."
+    )
+    assert caplog.records[0].checkout_id == checkout_id
+    assert caplog.records[1].levelno == logging.INFO
```
zedzior, line 494:
```suggestion
    assert caplog.records[0].levelno == logging.INFO
```

saleor/graphql/payment/tests/mutations/test_transaction_create.py
```diff
@@ -1222,6 +1228,181 @@ def test_transaction_create_for_checkout_fully_paid(
...
+
+    # then
+    mocked_checkout_fully_paid.assert_called_once_with(checkout, webhooks=set())
+    with pytest.raises(Checkout.DoesNotExist):
+        checkout.refresh_from_db()
+
+    order = Order.objects.get(checkout_token=checkout_token)
+    assert order.charge_status == CheckoutChargeStatus.FULL
+    assert order.authorize_status == CheckoutAuthorizeStatus.FULL
+    # TODO: check order events
```
zedzior, line 1290: 👮‍♂️

saleor/graphql/payment/tests/mutations/test_transaction_create.py
```diff
@@ -1222,6 +1228,181 @@ def test_transaction_create_for_checkout_fully_paid(
...
+    # then
+    mocked_checkout_fully_paid.assert_not_called()
+
+    with pytest.raises(Checkout.DoesNotExist):
+        checkout.refresh_from_db()
+
+    order = Order.objects.get(checkout_token=checkout_token)
+    assert order.charge_status == CheckoutChargeStatus.NONE
+    assert order.authorize_status == CheckoutAuthorizeStatus.FULL
+    # TODO: check order events
```
zedzior, line 1405: 👮‍♂️

saleor/graphql/payment/tests/mutations/test_transaction_event_report.py
```diff
@@ -1473,6 +1483,372 @@ def test_transaction_event_updates_checkout_full_paid_with_pending_charge_amount
...
+        query, variables, permissions=[permission_manage_payments]
+    )
+
+    # then
+    get_graphql_content(response)
+    with pytest.raises(Checkout.DoesNotExist):
+        checkout.refresh_from_db()
+
+    order = Order.objects.get(checkout_token=checkout_token)
+    # TODO:  to verify if it's correct
```
zedzior, line 1630: 👮‍♂️

saleor/graphql/payment/tests/mutations/test_transaction_event_report.py
```diff
@@ -1473,6 +1483,372 @@ def test_transaction_event_updates_checkout_full_paid_with_pending_charge_amount
...
+    )
+
+    # then
+    get_graphql_content(response)
+    with pytest.raises(Checkout.DoesNotExist):
+        checkout.refresh_from_db()
+
+    order = Order.objects.get(checkout_token=checkout_token)
+    assert order.charge_status == CheckoutChargeStatus.NONE
+    # TODO: to verify if it's correct
```
zedzior, line 1849: 👮‍♂️

saleor/graphql/payment/tests/mutations/test_transaction_create.py
IKarbowiak: It will be handled in the next PR

saleor/graphql/payment/tests/mutations/test_transaction_create.py
IKarbowiak: It will be handled in the next PR

saleor/checkout/tasks.py
IKarbowiak: In a `try` block, you should only include code that might raise an exception and needs to be handled. The goal is to keep the try block as minimal as possible, containing only the code that you expect might fail. That's why I logging in the `else` block

zedzior (approved):

fowczarek (approved):

saleor/checkout/tasks.py
```diff
@@ -109,3 +116,57 @@ def delete_expired_checkouts(
         else:
             task_logger.warning("Invocation limit reached, aborting task")
     return total_deleted, has_more
+
+
+@app.task
```
korycins, line 121: IMHO, this should be on a separate celery queue (probably this one: `settings.py::CHECKOUT_WEBHOOK_EVENTS_CELERY_QUEUE_NAME`). In case having a lot of different events on queue, this could cause a delay in conversion, which could lead to unexpected behaviors (like missing orders, out of stock etc).

saleor/checkout/tests/test_actions.py
```diff
@@ -77,6 +166,116 @@ def test_transaction_amounts_for_checkout_updated_with_already_fully_paid(
...
+    channel.automatically_complete_fully_paid_checkouts = True
+    channel.save(update_fields=["automatically_complete_fully_paid_checkouts"])
+
+    # when
+    transaction_amounts_for_checkout_updated(
+        transaction, manager=plugins_manager, user=staff_user, app=None
+    )
+
+    # then
+    flush_post_commit_hooks()
```
korycins, line 271: I would recommend to not using `flush_post_commit_hooks` in new tests. In case of failure, the DB is not clean up correctly and it causes that next tests are also failing. Pytest-django has dedicated fixture for handling on_commit actions [ref](https://pytest-django.readthedocs.io/en/latest/helpers.html#django-capture-on-commit-callbacks). 
```python
with django_capture_on_commit_callbacks(execute=True) as callbacks:
    transaction_amounts_for_checkout_updated(
        transaction, manager=plugins_manager, user=staff_user, app=None
    )
```
(We also have a task to fully drop the usage of `flush_post_commit_hooks()`)

saleor/checkout/actions.py
```diff
@@ -239,3 +244,12 @@ def transaction_amounts_for_checkout_updated(
             checkout_info=checkout_info,
             lines=lines,
         )
+
+    channel = checkout_info.channel
+    if (
+        channel.automatically_complete_fully_paid_checkouts
```
korycins, line 250: Just to be aware: The current implementation allows you to trigger multiple tasks for the same checkout. 
For example:
- You receive a transaction that caused that the checkout is fully-authorized
- You triggered the `automatic_checkout_completion_task`
- Before celery pick up the task, the new transaction for the same checkout was created
- if the `transaction_amounts_for_checkout_updated` will be called, the second celery task will be placed on the queue. 
(this is why a few lines above there is an if `if not previous_charge_status_is_fully_paid and current_status_is_fully_paid:` for charge and triggering the fully_paid event.)

In the worst scenario, this would cause `checkout_complete` will be called for the same checkout at the similar time. Conversion is locked so we should be safe, but maybe worth confirming that we're calling this event only once.

saleor/checkout/tasks.py
korycins: Edit:  

> (probably this one: settings.py::CHECKOUT_WEBHOOK_EVENTS_CELERY_QUEUE_NAME)

Just saw `WEBHOOK` word  in the name, so probably we need different one

PR #16786, opened 2024-10-01
Add error handling when removing orders with protected related object

kadewu:
I want to merge this change because currently we return 500 error in case when DraftOrder has protected relation when trying to remove it from Saleor.
Changes contains only error handling and proper resposne in mutations.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/bulk_mutations/draft_orders.py
+++ b/saleor/graphql/order/bulk_mutations/draft_orders.py
@@ -1,14 +1,14 @@
 from collections.abc import Iterable
-from typing import Optional, Union
+from typing import Union
 from uuid import UUID
 
 import graphene
 from django.core.exceptions import ValidationError
-from django.db.models import ProtectedError
 
 from ....channel import models as channel_models
 from ....order import OrderStatus, models
 from ....order.error_codes import OrderErrorCode
+from ....payment.models import Payment, TransactionItem
 from ....permission.enums import OrderPermissions
 from ...core import ResolveInfo
 from ...core.mutations import (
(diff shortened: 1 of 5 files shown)


saleor/graphql/order/bulk_mutations/draft_orders.py
```diff
@@ -51,6 +52,30 @@ def get_channel_ids(cls, instances) -> Iterable[Union[UUID, int]]:
         """Get the instances channel ids for channel permission accessible check."""
         return [order.channel_id for order in instances]
 
+    @classmethod
+    def perform_mutation(  # type: ignore[override]
+        cls, _root, info: ResolveInfo, /, *, ids, **data
+    ) -> tuple[int, Optional[ValidationError]]:
+        # code relies on the fact that both protected Models have FK named as `order`
```
kadewu, line 59: It is possible to deal with that problem - I can introduce method on User model sth like:
``` python
def get_protected_relations(cls):
    protected_relations = []
    for field in cls._meta.get_fields(include_hidden=True):
        if hasattr(field, "on_delete") and field.on_delete.__name__ == "PROTECT":
            protected_relations.append(field.related_model)
    return protected_relations
```
And then introduce another method to iterate through them and find exact name that is relating to `Order`. But not sure if we should be that careful here, any opinions?

IKarbowiak (approved):

saleor/graphql/order/mutations/draft_order_delete.py
```diff
@@ -54,12 +55,27 @@ def clean_instance(cls, info: ResolveInfo, instance):
...
-                order,
+        try:
+            with traced_atomic_transaction():
+                response = super().perform_mutation(_root, info, **data)
+                call_order_event(
+                    manager,
+                    WebhookEventAsyncType.DRAFT_ORDER_DELETED,
+                    order,
+                )
+        except ProtectedError as e:
```
maarcingebala, line 66: What types of models can be protected when deleting draft orders? IMO, if users want to remove a draft order, they shouldn't care about related objects; it's on us to delete them accordingly. Or is it a business requirement to inform them about related objects, and they should handle deleting them manually?

kadewu: The protected objects are: `Payment` and `TransactionItem`. Just for now to stop the 500 errors we are introducing the error handling and separately we want to expend logic with e.g. flags to control the flow. The future is yet to be agreed, but we need to ensure the control over it, the same way as we do in checkout, that why for now it's just proper error handling for the users.

maarcingebala: Right now, the error message reveals the internals of Saleor (names of models) which is irrelevant to the user. If you want to keep it that way I would at least change the error message to some generic message explaining why the order cannot be removed.

kadewu: I'm going to move the error-handling to `clean_input` - this way it will more readable code and easier to properly provide error message.

korycins (approved):

fowczarek (approved):

PR #16808, opened 2024-10-02
Async payload generation

maarcingebala:
This PR adds the ability to generate subscription payload for webhook in the async task. Currently enabled for the following async events:
- `checkout_created`
- `checkout_updated`

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -48,10 +53,37 @@
 OBSERVABILITY_QUEUE_NAME = "observability"
 
 
+def create_deliveries_for_subscriptions_deferred_payload(
```
korycins, line 56: Question: as we don't make any subscription here (this is done by celery task), maybe worth to consider dropping `subscriptions` from here?
I mean something like `create_deliveries_for_deferred_events`? 🤔 
(If you don't like this, just mark it as resolved 😄 )

saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -251,13 +317,76 @@ def trigger_webhooks_async(
...
+        RequestorModelName.USER,
+    ):
+        model = apps.get_model(args_obj.requestor_model_name)
+        requestor = model.objects.filter(pk=args_obj.requestor_model_id).first()
+
+    subscribable_object = (
+        apps.get_model(args_obj.model_name).objects.filter(pk=args_obj.model_id).first()
+    )
+    if not subscribable_object:
+        logger_obj.error(
```
korycins, line 340: IMHO, this could be quite a common case. Like checkout-updated to deliver, but the order was created and checkout was deleted. Are you sure, that we want to raise this as an error?

saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -251,13 +317,76 @@ def trigger_webhooks_async(
...
+
+            # Save the payload in the EventPayload object for later reuse in case of
+            # automatic or manual retry.
+            with allow_writer():
+                event_payload = EventPayload.objects.create_with_payload_file(data_json)
+                delivery.payload = event_payload
+                delivery.save(update_fields=["payload"])
+
+            return data_json
+    return None
```
korycins, line 380: Shouldn't we have a test for cases when `None` is returned? 
How the flow will behave when `not subscribable_object`?  Will it return `None`?
image

saleor/webhook/transport/asynchronous/transport.py
maarcingebala: I'm not sure yet; I can change this to a warning as I think we should monitor situations when this happens first.

saleor/webhook/transport/utils.py
```diff
@@ -86,6 +88,42 @@ class WebhookResponse:
...
+class RequestorModelName:
+    # lowercase, as it is returned as such by `model._meta.model_name`
+    APP = "app.app"
+    USER = "account.user"
+
+
+@dataclass
+class DeferredPayloadData:
+    model_name: str
+    model_id: Union[int, UUID]
```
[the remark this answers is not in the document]
maarcingebala: All fields are about fetching two different models so they should use common naming. Other options we could use are:
- `instance_name`, `instance_id`, `requestor_instance_name`, `requestor_istance_id`
- `object_name`, `object_id`, `requestor_object_name`, `requestor_object_id` 

I'm not sure if there is much difference TBH, so for now I'll stick to original naming. If anyone thinks otherwise, then feel free to comment and propose a version that works better.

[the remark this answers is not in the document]
korycins: `model_name` suggest the `class` name. And this is ok. Renaming `model_name` to `instance_name` will suggest that we are talking about `instance.name`, which is not correct. The dataclass accept `model_name` (class(models.Model) and instance_id of this class. So IMO we can't use values like `instance_name` and `instance_id` together. The model_name is correct IMO, only `model_id` suggests that we are storing ID of the `class(models.Model)`.

saleor/webhook/transport/asynchronous/transport.py
korycins: I think this will be quite common case. Like `checkout_updated` events + checkout->order. This could raise this `error`. This will also be handled by active sentry. So could cause some false positives. I would consider changing this to the warning.

saleor/webhook/transport/utils.py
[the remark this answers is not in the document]
maarcingebala: Fixed to use the form from the comment above.

saleor/webhook/transport/asynchronous/transport.py
maarcingebala: Added

maarcingebala: When we fail to get the subscribable object the payload would be None. In such case `send_webhook_request_async` will not deliver the webhook, the delivery and attempt objects will be saved with "failed" status.

IKarbowiak (approved):

saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -282,32 +307,69 @@ def trigger_webhooks_async_for_multiple_objects(
...
+            event_delivery_ids = [delivery.id for delivery, _ in deferred_deliveries]
+
+            # Trigger deferred payload generation task for each subscribable object.
+            generate_deferred_payloads.apply_async(
+                kwargs={
+                    "event_delivery_ids": event_delivery_ids,
+                    "deferred_payload_data": asdict(deferred_payload_data),
+                    "queue": queue,
+                },
+                queue=queue or settings.WEBHOOK_CELERY_QUEUE_NAME,
```
korycins, line 369: This means: "use the same celery queue as the one for sending the webhooks". 
And IMO this is wrong as it means, that single queue will be shared between sending webhooks & building the webhook payloads. Few downsizes of using the same queue:
- in case of high traffic and a huge amount of webhooks to send in the queue, the building payload could be delayed, which means that the body could be changed, or object could not exist anymore. 
- sending webhooks could be delayed as right now the same queue also builds the payload. So assuming that you have checkout with sync webhooks to call, this could take a few seconds (max 20 secs) to process single task. It means, that this could downgrade the maximal number of webhooks that we're able to send by `WEBHOOK_CELERY_QUEUE_NAME`

IMHO, generating payloads should have its own queue, as we want to build a payload as fast as possible. Or at least using the default one, but using the same as the one for sending webhooks could cause a problem in case of huge amount of webhooks on queue.

saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -282,32 +307,69 @@ def trigger_webhooks_async_for_multiple_objects(
...
-                allow_replica=allow_replica,
-                pre_save_payloads=pre_save_payloads,
-                request_time=request_time,
+        subscribable_objects = [
+            webhook_payload_data.subscribable_object
+            for webhook_payload_data in webhook_payloads_data
+        ]
+        if is_deferred_payload:
+            _deferred_deliveries_to_create = []
+            for subscribable_object in subscribable_objects:
```
korycins, line 329: Can we move this logic to separate function? Like we already have for `create_deliveries_for_multiple_subscription_objects(`. It would be more readable IMO.

saleor/webhook/transport/utils.py
```diff
@@ -360,19 +397,39 @@ def handle_webhook_retry(
...
+    )
+
+    active_deliveries = {}
+    inactive_delivery_ids = set()
+
+    not_found_delivery_ids = set(event_delivery_ids) - set(
+        delivery.pk for delivery in deliveries
+    )
+    for not_found_delivery_id in not_found_delivery_ids:
+        logger.warning("Event delivery id: %r not found", not_found_delivery_id)
```
korycins, line 418: Maybe instead of multiple warnings, we could make single one? Right now this could cause a lot of entries, where the same feedback we could get by receiving single entry 🤔

saleor/webhook/transport/utils.py
```diff
@@ -360,19 +397,39 @@ def handle_webhook_retry(
...
+
+    for delivery in deliveries:
+        if delivery.webhook.is_active:
+            active_deliveries[delivery.pk] = delivery
+        else:
+            logger.info("Event delivery id: %r webhook is disabled.", delivery.pk)
```
korycins, line 424: Same question as above

saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -364,6 +426,81 @@ def trigger_webhooks_async(
...
+            request=request,
+            app=webhook.app,
+        )
+
+        if data_promise:
+            data = data_promise.get()
+            if data:
+                data_json = json.dumps({**data})
+                with allow_writer():
+                    event_payload = EventPayload.objects.create_with_payload_file(
```
korycins, line 483: In each place, payload generation is wrapped in transaction.atomic(). Shouldn't we also do this here?
Also this means, that for each delivery in deliveries, we are calling 2 or 3 db queries. Can't we make it in bulk? For example, using: `bulk_create_with_payload_files` ?

saleor/webhook/transport/utils.py
maarcingebala: I had to reimplement this function to support getting multiple deliveries, but I don't intend to change behavior about this in this PR.

saleor/webhook/transport/utils.py
maarcingebala: Not scope of this PR as this is already a case in 3.20.

saleor/webhook/transport/asynchronous/transport.py
patrys: >in case of high traffic and a huge amount of webhooks to send in the queue, the building payload could be delayed, which means that the body could be changed, or object could not exist anymore.

That should not be an issue as there is similarly little value in delivering long-obsolete data.

>sending webhooks could be delayed as right now the same queue also builds the payload. So assuming that you have checkout with sync webhooks to call, this could take a few seconds (max 20 secs) to process single task.

That is a valid concern, generating events would happen at the expense of our webhook delivery capacity. We could use the default queue for payload generation until we implement the delivery fairness improvements we're working on.

fowczarek: I also agree to use the default queue for this task. 
I don't see any more objections.

saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -364,6 +426,95 @@ def trigger_webhooks_async(
...
+        data_promise = generate_payload_promise_from_subscription(
+            event_type=event_type,
+            subscribable_object=subscribable_object,
+            subscription_query=webhook.subscription_query,  # type: ignore
+            request=request,
+            app=webhook.app,
+        )
+
+        if data_promise:
+            data = data_promise.get()
```
fowczarek, line 483: Could we rewrite it to call get for all promises after the loop?

saleor/webhook/transport/asynchronous/transport.py
maarcingebala: Change to use default queue for payload generation.

fowczarek (approved):

PR #16830, opened 2024-10-07
Circuit breaker reset mutation

tomaszszymanski129:
I want to merge this change because it adds `ReenableSyncWebhooks` mutation.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/app/mutations/reenable_sync_webhooks.py
+++ b/saleor/graphql/app/mutations/reenable_sync_webhooks.py
@@ -2,13 +2,13 @@ import graphene
 from django.conf import settings
 from django.core.exceptions import ValidationError
 
-from ....core.error_codes import ShopErrorCode
+from ....app.error_codes import AppErrorCode
 from ....graphql.app.types import App
 from ....permission.enums import SitePermissions
 from ...core import ResolveInfo
-from ...core.doc_category import DOC_CATEGORY_SHOP
+from ...core.doc_category import DOC_CATEGORY_APPS
 from ...core.mutations import BaseMutation
-from ...core.types import ShopError
+from ...core.types import AppError
 
 breaker_board = None
 if settings.ENABLE_BREAKER_BOARD:
@@ -22,11 +22,7 @@ if settings.ENABLE_BREAKER_BOARD:
 class ReenableSyncWebhooks(BaseMutation):
     app = graphene.Field(
         App,
-        description=(
-            "Re-enable sync webhooks for the app if circuit breaker was tripped. "
-            "Can be used to manually re-enable sync webhooks for the app before "
-            "the cooldown period ends."
-        ),
+        description="App for which sync webhooks were re-enabled. ",
     )
 
     class Arguments:
@@ -35,11 +31,15 @@ class ReenableSyncWebhooks(BaseMutation):
         )
 
     class Meta:
-        description = "Re-enable sync webhooks for provided app."
-        doc_category = DOC_CATEGORY_SHOP
+        description = (
+            "Re-enable sync webhooks for provided app."
+            "Can be used to manually re-enable sync webhooks for the app before "
+            "the cooldown period ends."
+        )
+        doc_category = DOC_CATEGORY_APPS
         permissions = (SitePermissions.MANAGE_SETTINGS,)
-        error_type_class = ShopError
-        error_type_field = "shop_errors"
+        error_type_class = AppError
+        error_type_field = "app_errors"
 
     @classmethod
     def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
@@ -51,7 +51,7 @@ class ReenableSyncWebhooks(BaseMutation):
                     {
                         "app_id": ValidationError(
                             f"Cannot re-enable sync webhooks for the app {app.name}.",  # type: ignore[union-attr]
-                            code=ShopErrorCode.INVALID.value,
+                            code=AppErrorCode.INVALID.value,
                         )
                     }
                 )
--- a/saleor/graphql/app/mutations/reenable_sync_webhooks.py
+++ b/saleor/graphql/app/mutations/reenable_sync_webhooks.py
@@ -4,7 +4,7 @@ from django.core.exceptions import ValidationError
 
 from ....app.error_codes import AppErrorCode
 from ....graphql.app.types import App
-from ....permission.enums import SitePermissions
+from ....permission.enums import AppPermission
 from ...core import ResolveInfo
 from ...core.doc_category import DOC_CATEGORY_APPS
 from ...core.mutations import BaseMutation
@@ -37,13 +37,15 @@ class ReenableSyncWebhooks(BaseMutation):
             "the cooldown period ends."
         )
         doc_category = DOC_CATEGORY_APPS
-        permissions = (SitePermissions.MANAGE_SETTINGS,)
+        permissions = (AppPermission.MANAGE_APPS,)
         error_type_class = AppError
         error_type_field = "app_errors"
 
     @classmethod
     def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
-        app = cls.get_node_or_error(info, data.get("app_id"), only_type="App")
+        app = cls.get_node_or_error(
+            info, data.get("app_id"), only_type="App", field="appId"
+        )
         if breaker_board:
             error = breaker_board.storage.clear_state_for_app(app.id)  # type: ignore[union-attr]
             if error:
--- a/saleor/graphql/app/mutations/reenable_sync_webhooks.py
+++ b/saleor/graphql/app/mutations/reenable_sync_webhooks.py
@@ -10,6 +10,7 @@ from ...core.doc_category import DOC_CATEGORY_APPS
 from ...core.mutations import BaseMutation
 from ...core.types import AppError
 
+# TODO: Remove the conditional when unit tests circular import is solved.
 breaker_board = None
 if settings.ENABLE_BREAKER_BOARD:
     from ....webhook.transport.synchronous.circuit_breaker.breaker_board import (
@@ -32,7 +33,7 @@ class ReenableSyncWebhooks(BaseMutation):
 
     class Meta:
         description = (
-            "Re-enable sync webhooks for provided app."
+            "Re-enable sync webhooks for provided app. "
             "Can be used to manually re-enable sync webhooks for the app before "
             "the cooldown period ends."
         )
--- a/saleor/graphql/app/tests/mutations/test_reenable_sync_webhooks.py
+++ b/saleor/graphql/app/tests/mutations/test_reenable_sync_webhooks.py
@@ -1,0 +1,90 @@
+from unittest.mock import patch
+
+import graphene
+
+from .....app.error_codes import AppErrorCode
+from ....tests.utils import get_graphql_content
+
+REENABLE_BREAKER_MUTATION = """
+    mutation ReenableSyncWebhooks($appId: ID!) {
+        reenableSyncWebhooks(appId: $appId) {
+            app {
+                name
+            }
+            errors {
+                field
+                code
+                message
+            }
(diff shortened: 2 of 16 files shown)


saleor/graphql/schema.graphql
```diff
@@ -15905,6 +15905,16 @@ type Mutation {
     input: GiftCardSettingsUpdateInput!
   ): GiftCardSettingsUpdate @doc(category: "Gift cards")
 
+  """
+  Reset circuit breaker for provided app. 
+  
+  Requires one of the following permissions: MANAGE_SETTINGS.
+  """
+  resetCircuitBreaker(
```
patrys, line 15913: I don't think we should be using "circuit breaker" in our API naming. It's a design pattern, we don't have objects named "singleton," etc.

wcislo-saleor: @tomaszszymanski129

I see this conversation is already resolved but there is no comment regarding what's the resolution here.

Are we renaming this with a different name?

saleor/graphql/shop/mutations/close_circuit_breaker.py
```diff
@@ -0,0 +1,56 @@
...
+
+from ....core.error_codes import ShopErrorCode
+from ....graphql.app.types import App
+from ....permission.enums import SitePermissions
+from ...core import ResolveInfo
+from ...core.doc_category import DOC_CATEGORY_SHOP
+from ...core.mutations import BaseMutation
+from ...core.types import ShopError
+
+# TODO: make breaker_board instance a singleton
```
wcislo-saleor, line 14: There is nothing wrong with `breaker_board` not being singleton of `BreakerBoard` assuming it's not being constructed before every request. Currently it doesn't hold any data nor logic that'd require it.

saleor/graphql/shop/mutations/close_circuit_breaker.py
```diff
@@ -0,0 +1,56 @@
...
+breaker_board = None
+if settings.ENABLE_BREAKER_BOARD:
+    from ....webhook.transport.synchronous.circuit_breaker.breaker_board import (
+        BreakerBoard,
+    )
+
+    breaker_board = BreakerBoard(
```
wcislo-saleor, line 21: We instantiate `BreakerBoard` in the same way in more than one place. Maybe we can extract this code so it's not repeated.

saleor/graphql/schema.graphql
```diff
@@ -14289,6 +14289,16 @@ type Mutation {
     input: GiftCardSettingsUpdateInput!
   ): GiftCardSettingsUpdate @doc(category: "Gift cards")
 
+  """
+  Close circuit breaker for provided app. 
+  
+  Requires one of the following permissions: MANAGE_SETTINGS.
+  """
+  closeCircuitBreaker(
```
wcislo-saleor, line 14297: I also have some doubts about the naming. What do you think about name that explains a feature, not the design pattern? Would something along the line of reenabling temporarily disabled sync webhooks for an app be better for the end user?

saleor/graphql/shop/mutations/close_circuit_breaker.py
```diff
@@ -0,0 +1,56 @@
...
+
+    class Meta:
+        description = "Close circuit breaker for provided app."
+        doc_category = DOC_CATEGORY_SHOP
+        permissions = (SitePermissions.MANAGE_SETTINGS,)
+        error_type_class = ShopError
+        error_type_field = "shop_errors"
+
+    @classmethod
+    def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
```
wcislo-saleor, line 44: I'm wondering whether telling end-user insights into configuration of Saleor is a good idea.

We can either tell them or tell them everything is alright. I'd be more inclined to telling them everything is alright (that is: if breaker board is disabled then return HTTP 200, if breaker is enabled then return the result of the operation).

I'd suggest consulting that with Python guild elders.

saleor/webhook/transport/synchronous/circuit_breaker/storage.py
```diff
@@ -118,3 +131,11 @@ def register_event_returning_count(
         except (RedisError, IndexError):
             logger.warning(self.WARNING_MESSAGE, exc_info=True)
             return 0
+
+    def close_breaker(self, app_id: int):
```
wcislo-saleor, line 135: I understand the use of `*` (wildcard) to obtain keys because it protects us in future, had we add any more counters in the storage (apart from the current main key, `total` and `error`).

Although I'd recommend having a fixed list of keys to delete here (main key, `total` key and `error` key). We would not make two requests to Redis and it'd be much more clearer to the person who reads the code what is actually happening.

Related but nothing actionable for this part of the code. If we do more than more one operation with Redis consider using pipelines.

wcislo-saleor, line 135: Storage classes themselves are not really concerned with the idea of breakers, they don't even have to be used in the context of circuit breakers. Therefore maybe the naming relating to clearing the state for a given app would be more fitting.

wcislo-saleor (commented):
Let me know what do you think about the review comments.

saleor/graphql/shop/mutations/close_circuit_breaker.py
tomaszszymanski129: @wcislo-saleor I was wondering about singleton because if not, then we have issues with InMemoryStorage. It's not possible to test breaker reset mutation if these have different instances of in-mem storage.

saleor/graphql/schema.graphql
tomaszszymanski129: Indeed, something weird happened to the schema, though for now I renamed it to `ReenableSyncWebhooks`, though we might change it later if we find a better name, it's a quick refactor.

saleor/graphql/shop/mutations/close_circuit_breaker.py
tomaszszymanski129: On second thought I agree. No need to share configuration details. I'm preparing a refactor so it will work differently.

saleor/webhook/transport/synchronous/circuit_breaker/storage.py
tomaszszymanski129: Changed to `clear_state_for_app`.

saleor/graphql/schema.graphql
tomaszszymanski129: Changed to `reenableSyncWebhooks` for now. I provided more details in Patrys's comment on the top.

saleor/webhook/transport/synchronous/transport.py
```diff
@@ -326,15 +325,9 @@ def trigger_webhook_sync(
...
 if settings.ENABLE_BREAKER_BOARD:
-    trigger_webhook_sync = BreakerBoard(
-        storage=import_string(settings.BREAKER_BOARD_STORAGE_CLASS_STRING)(),  # type: ignore[arg-type]
-        failure_threshold=settings.BREAKER_BOARD_FAILURE_THRESHOLD_PERCENTAGE,
-        failure_min_count=settings.BREAKER_BOARD_FAILURE_MIN_COUNT,
-        cooldown_seconds=settings.BREAKER_BOARD_COOLDOWN_SECONDS,
-        ttl=settings.BREAKER_BOARD_TTL,
-    )(trigger_webhook_sync)
+    breaker_board = initialize_breaker_board()
+    trigger_webhook_sync = breaker_board(trigger_webhook_sync)
```
maarcingebala, line 330: I wonder if we could apply the breaker board as a decorator:

```
@breaker_bord()
def trigger_webhook_sync(...)
```

This could be possible if we moved the if-condition to the initialization. Then all usages of breaker would be as simple as this.

tomaszszymanski129: Ain't so simple. Since now initialize_breaker_board can return None, it cannot be used like this.

saleor/graphql/app/mutations/reenable_sync_webhooks.py
```diff
@@ -0,0 +1,60 @@
...
+from ...core.mutations import BaseMutation
+from ...core.types import AppError
+
+breaker_board = None
+if settings.ENABLE_BREAKER_BOARD:
+    from ....webhook.transport.synchronous.circuit_breaker.breaker_board import (
+        initialize_breaker_board,
+    )
+
+    breaker_board = initialize_breaker_board()
```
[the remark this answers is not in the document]
tomaszszymanski129: Added a todo note

maarcingebala (approved):

saleor/webhook/transport/tests/circuit_breaker/test_redis_storage.py
```diff
@@ -80,3 +94,11 @@ def test_storage_raises_on_non_redis_cache_url(settings):
 
     with pytest.raises(ImproperlyConfigured):
         RedisStorage()
+
+
+def test_storage_clear_state_raises_error(storage):
```
IKarbowiak, line 99: Missing `given, when, then`

saleor/graphql/app/tests/mutations/test_reenable_sync_webhooks.py
```diff
@@ -0,0 +1,90 @@
...
+    breaker_board_mock.storage.clear_state_for_app.side_effect = lambda id: None
+    staff_user.user_permissions.set([permission_manage_apps])
+    variables = {
+        "appId": graphene.Node.to_global_id("App", app.id),
+    }
+    # when
+    response = staff_api_client.post_graphql(REENABLE_BREAKER_MUTATION, variables)
+    content = get_graphql_content(response)
+    data = content["data"]["reenableSyncWebhooks"]
+    breaker_board_mock.storage.clear_state_for_app.assert_called_once_with(app.id)
```
IKarbowiak, line 42: Missing `# then`

saleor/graphql/app/tests/mutations/test_reenable_sync_webhooks.py
```diff
@@ -0,0 +1,90 @@
...
+    breaker_board_mock.storage.clear_state_for_app.side_effect = lambda id: 1
+    staff_user.user_permissions.set([permission_manage_apps])
+    variables = {
+        "appId": graphene.Node.to_global_id("App", app.id),
+    }
+    # when
+    response = staff_api_client.post_graphql(REENABLE_BREAKER_MUTATION, variables)
+    content = get_graphql_content(response)
+    data = content["data"]["reenableSyncWebhooks"]
+    breaker_board_mock.storage.clear_state_for_app.assert_called_once_with(app.id)
```
IKarbowiak, line 64: Missing `# then`

saleor/graphql/app/tests/mutations/test_reenable_sync_webhooks.py
```diff
@@ -0,0 +1,90 @@
...
+    breaker_board_mock.storage.clear_state_for_app.side_effect = lambda id: 1
+    staff_user.user_permissions.set([permission_manage_apps])
+    variables = {
+        "appId": graphene.Node.to_global_id("App", 9999),
+    }
+    # when
+    response = staff_api_client.post_graphql(REENABLE_BREAKER_MUTATION, variables)
+    content = get_graphql_content(response)
+    data = content["data"]["reenableSyncWebhooks"]
+    breaker_board_mock.storage.clear_state_for_app.assert_not_called()
```
IKarbowiak, line 87: Missing `# then`

IKarbowiak (approved):

fowczarek (approved):

saleor/graphql/app/tests/mutations/test_reenable_sync_webhooks.py
tomaszszymanski129: Will quickly add those in separate PR as I'm also adding more tests there.

PR #16932, opened 2024-10-21
Fix failing import-path for promotion schedule for self-hosted instances

korycins:
I want to merge this change because when starting celery with `--scheduler saleor.schedulers.schedulers:DatabaseScheduler`, the `saleor.schedulers.models.CustomSchedule` raises an exception:

`django.core.exceptions.SuspiciousOperation: Expected type of 'saleor.core.schedules.promotion_webhook_schedule' to be inheriting from BaseScheduler but found: <class 'type'> ((<class 'object'>,))`.

The import path should be the path to the instance, not to the class. This caused, that we can't run our own `DatabaseScheduler`.

Port of changes from: #16915

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/core/schedules.py
+++ b/saleor/core/schedules.py
@@ -36,7 +36,7 @@ class promotion_webhook_schedule(CustomSchedule):
             schedule=self,
             nowfun=nowfun,
             app=app,
-            import_path="saleor.core.schedules.promotion_webhook_schedule",
+            import_path="saleor.core.schedules.initiated_promotion_webhook_schedule",
         )
         # Seconds left to next batch processing
         self.NEXT_BATCH_RUN_TIME = 5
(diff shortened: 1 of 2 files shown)


fowczarek (approved):

kadewu (approved):

bnkwines:
there's a chance that in the db the old scheduler is left, what's the process of creating a migration for something like:

```
saleor-db=# select * from schedulers_customschedule
saleor-db-# ;
 id |               schedule_import_path
----+--------------------------------------------------
  1 | saleor.core.schedules.promotion_webhook_schedule
(1 row)

saleor-db=# update schedulers_customschedule set schedule_import_path = 'saleor.core.schedules.initiated_promotion_webhook_schedule' where id = 1;
```

PR #16942, opened 2024-10-24
Fix crash on checkout resolver when variant listing is not present

korycins:
I want to merge this change because it solves the issue when checkout-line has variant without existing channel-listing.
This should address: https://linear.app/saleor/issue/MERX-1043/attributeerror-nonetype-object-has-no-attribute-price and https://linear.app/saleor/issue/MERX-957/attributeerror-nonetype-object-has-no-attribute-currency

Current flow:

Removing channel support via `ProductVariantBulkUpdate`, causes that we have checkouts with the lines that do not have enough details to list the prices. This causes unhandled exception returned by API.

Solution:

We will correctly return the lines without channel-listing. To do that we will use de-normalized Variant prices stored in `CheckoutLine`.

Error `UNAVAILABLE_VARIANT_IN_CHANNEL` will be raised in the same places as it is raised right now (for example checkoutComplete). This blocks actions like finalizing checkout process when checkout contains invalid lines, user will need to make an action to remove the non-available lines - this already working in that way.

`checkout.problems` and `checkoutLine.problems` will return a proper error that the given line is not available  - this already working in that way.

Changes:

CheckoutLine:

`undiscounted_unit_price`: de-normalized price of the variant, updated to the newest during the calculation, only when channel_listing is preset. If the listing is missing, the value will represent the previous price assigned to checkout line.

CheckoutLineInfo:

- `channel_listing` filed will become Optional (actually because of `skip_recalculation` in `fetch_checkout_lines` channel_listing could already be optional).
- cached_property `undiscounted_unit_price`, if channel listing is missing, use the value from checkoutLine object.
- cached_property `variant_discounted_price`: discounted_price fetched from `ProductVariantChannelListing.discounted_price`, if missing calculated based on denormalize variant price and line.discounts.

> [!CAUTION]
> `fetch_checkout_lines` will return lines without listing as valid line, but it will also include the PK in `unavailable_variant_pks`

Calculations:

During the calculation process, we will also assign the price from `ProductVariantChannelListing` to `CheckoutLine`.

Migrations:

New field: `undiscounted_unit_price`
The new field will receive the value from channel listing or `checkout_line.price_override`. If both missing, the `undiscounted_unit_price` will be calculated based on the `line.total_price`.

# Impact

- [x] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/checkout/fetch.py
```diff
@@ -56,6 +54,41 @@ class CheckoutLineInfo(LineInfo):
     discounts: list["CheckoutLineDiscount"]
     tax_class: Optional["TaxClass"] = None
 
+    @cached_property
+    def variant_discounted_price(self) -> Money:
+        """Return the discounted variant price.
+
+        If listing is present return the discounted price from the listing.
```
IKarbowiak, line 61:
```suggestion
        If listing is present return the discounted price from the listing,
```

saleor/checkout/fetch.py
```diff
@@ -56,6 +54,41 @@ class CheckoutLineInfo(LineInfo):
     discounts: list["CheckoutLineDiscount"]
     tax_class: Optional["TaxClass"] = None
 
+    @cached_property
+    def variant_discounted_price(self) -> Money:
+        """Return the discounted variant price.
```
IKarbowiak, line 59: Maybe we should specify that this price include just catalogue promotion

saleor/checkout/fetch.py
```diff
@@ -56,6 +54,41 @@ class CheckoutLineInfo(LineInfo):
...
+
+        If listing is present return the discounted price from the listing.
+        if listing is not present, calculate current unit price based on the details
+        assigned to the line. Checkout cannot be converted to new order when contains
```
IKarbowiak, line 63:
```suggestion
       if listing is not present, calculate current unit price based on the `undiscounted_unit_price` and catalogue promotion discounts
        assigned to the line. Checkout cannot be converted to new order when contains
```

saleor/checkout/fetch.py
korycins: Added explanation in line below. Thx 👍

IKarbowiak (dismissed):

saleor/checkout/migrations/0072_alter_checkoutline_undiscounted_unit_price_amount.py
```diff
@@ -0,0 +1,29 @@
...
+            ),
+        ),
+        migrations.RunSQL(
+            """
+            ALTER TABLE checkout_checkoutline
+            ALTER COLUMN undiscounted_unit_price_amount
+            SET DEFAULT 0;
+            """,
+            migrations.RunSQL.noop,
+        ),
```
kadewu, line 28: We should split state and db operation here:
```suggestion
        migrations.SeparateDatabaseAndState(
            state_operations=[
                migrations.AlterField(
                    model_name="checkoutline",
                    name="undiscounted_unit_price_amount",
                    field=models.DecimalField(
                        decimal_places=3, default=Decimal("0"), max_digits=12
                    ),
                ),
            ],
            database_operations=[
                migrations.RunSQL(
                    """
                    ALTER TABLE checkout_checkoutline
                    ALTER COLUMN undiscounted_unit_price_amount
                    SET DEFAULT 0;
                    """,
                    migrations.RunSQL.noop,
                ),
                migrations.RunSQL(
                    """
                    ALTER TABLE checkout_checkoutline
                    ALTER COLUMN undiscounted_unit_price_amount
                    SET NOT NULL;
                    """
                )
            ]
        )
```

This way we will not generate this query in the migrations:
`UPDATE "checkout_checkoutline" SET "undiscounted_unit_price_amount" = 0 WHERE "undiscounted_unit_price_amount" IS NULL; SET CONSTRAINTS ALL IMMEDIATE;`

kadewu (approved):

saleor/checkout/migrations/0072_alter_checkoutline_undiscounted_unit_price_amount.py
korycins: Thanks @kadewu  ❤️

szczecha (approved):

saleor/graphql/product/mutations/channels.py
```diff
@@ -318,8 +318,6 @@ def remove_variants(
         ).exists():
             product_channel_listing.delete()
 
-        cls.perform_checkout_lines_delete(variants, [channel.id])
```
fowczarek, line 321: Should you remove the body of function?

korycins: @fowczarek I can't right now. It is still used when removing ProductChannelListing. So need to wait till applying similar changes for product-channel-listing

kadewu (approved):

fowczarek (approved):

PR #16948, opened 2024-10-28
Fix crash on checkout resolver when variant listing is not present

korycins:
I want to merge this change because it solves the issue when checkout-line has variant without existing channel-listing.
This should address: https://linear.app/saleor/issue/MERX-1043/attributeerror-nonetype-object-has-no-attribute-price and https://linear.app/saleor/issue/MERX-957/attributeerror-nonetype-object-has-no-attribute-currency

Port of changes from: #16942
> [!CAUTION]
> [Commit](https://github.com/saleor/saleor/pull/16948/commits/8da6397eeaa03957172802e524af8ecf7501b741) includes the changes with zero-downtime migration

Current flow:

Removing channel support via `ProductVariantBulkUpdate`, causes that we have checkouts with the lines that do not have enough details to list the prices. This causes unhandled exception returned by API.

Solution:

We will correctly return the lines without channel-listing. To do that we will use de-normalized Variant prices stored in `CheckoutLine`.

Error `UNAVAILABLE_VARIANT_IN_CHANNEL` will be raised in the same places as it is raised right now (for example checkoutComplete). This blocks actions like finalizing checkout process when checkout contains invalid lines, user will need to make an action to remove the non-available lines - this already working in that way.

`checkout.problems` and `checkoutLine.problems` will return a proper error that the given line is not available  - this already working in that way.

Changes:

CheckoutLine:

`undiscounted_unit_price`: de-normalized price of the variant, updated to the newest during the calculation, only when channel_listing is preset. If the listing is missing, the value will represent the previous price assigned to checkout line.

CheckoutLineInfo:

- `channel_listing` filed will become Optional (actually because of `skip_recalculation` in `fetch_checkout_lines` channel_listing could already be optional).
- cached_property `undiscounted_unit_price`, if channel listing is missing, use the value from checkoutLine object.
- cached_property `variant_discounted_price`: discounted_price fetched from `ProductVariantChannelListing.discounted_price`, if missing calculated based on denormalize variant price and line.discounts.

> [!CAUTION]
> `fetch_checkout_lines` will return lines without listing as valid line, but it will also include the PK in `unavailable_variant_pks`

Calculations:

During the calculation process, we will also assign the price from `ProductVariantChannelListing` to `CheckoutLine`.

Migrations:

New field: `undiscounted_unit_price`
The new field will receive the value from channel listing or `checkout_line.price_override`. If both missing, the `undiscounted_unit_price` will be calculated based on the `line.total_price`.

# Impact

- [x] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


kadewu (approved):

saleor/celeryconf.py
```diff
@@ -33,5 +33,11 @@ def setup_celery_logging(loglevel=None, **kwargs):
     ],
     related_name="saleor3_20",
 )
+app.autodiscover_tasks(
+    packages=[
+        "saleor.checkout.migrations.tasks",
+    ],
+    related_name="saleor3_20",
+)
```
IKarbowiak, line 41: Should me merged with the previous list

fowczarek (approved):

IKarbowiak (approved):

PR #16953, opened 2024-10-31
Handle refund OR cancel reported from payment app

IKarbowiak:
- Allow to report `REFUND_OR_CANCEL_*` events from payment app
- Handle the reported event, deduce the event type based on the following conditions
   - Map to corresponding REFUND event if:
      - there is an event with the `REFUND` related type with the same pspReference
      - there is no `CANCEL` related event with the same pspReference
      - OR transaction.chargedAmount is above 0

    - Map to corresponding CANCEL event if:
      - there is an event with the `CANCEL` related type with the same pspReference
      - there is no `REFUND` related event with the same pspReference
      - OR transaction.chargedAmount is 0
- When any REFUND or CANCEL event appears in the meantime, the refund or cancel event type is reavulated based on the previous events

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


IKarbowiak:
Canceled.

PR #16967, opened 2024-11-06
Fix ProductVariantUpdate not to perform save when not necessary

tomaszszymanski129:
I want to merge this change because it skips saving of product variant on ProductVariantUpdate mutation when it's not necessary changes for now ignores case when there are attributes since it's bit more complicated, will be posted as separate pr along with update_fields.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/product/mutations/product_variant/product_variant_update.py
+++ b/saleor/graphql/product/mutations/product_variant/product_variant_update.py
@@ -149,15 +149,23 @@ class ProductVariantUpdate(ProductVariantCreate, ModelWithExtRefMutation):
     @classmethod
     def _save(cls, info: ResolveInfo, instance, cleaned_input, base_fields_changed):
         new_variant = instance.pk is None
-        cls.set_track_inventory(info, instance, cleaned_input)
         variant_modified = False
+        product_search_index_marked_dirty = False
         with traced_atomic_transaction():
             if base_fields_changed:
                 instance.save()
                 variant_modified = True
             if not instance.product.default_variant:
                 instance.product.default_variant = instance
-                instance.product.save(update_fields=["default_variant", "updated_at"])
+                instance.product.search_index_dirty = True
+                instance.product.save(
+                    update_fields=[
+                        "default_variant",
+                        "updated_at",
+                        "search_index_dirty",
+                    ]
+                )
+                product_search_index_marked_dirty = True
             if stocks := cleaned_input.get("stocks"):
                 cls.create_variant_stocks(instance, stocks)
                 variant_modified = True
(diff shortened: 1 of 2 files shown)


cmiacz (approved):
Overal concept LGTM

saleor/graphql/product/mutations/product_variant/product_variant_update.py
```diff
@@ -162,11 +200,28 @@ def perform_mutation(  # type: ignore[override]
...
-            input=input,
+        instance = cls.get_instance(
+            info, id=id, sku=sku, external_reference=external_reference, input=input
         )
+        old_instance_data = model_to_dict(instance).copy()
+
+        cleaned_input = cls.clean_input(info, instance, input)
+        metadata_list = cleaned_input.pop("metadata", None)
+        private_metadata_list = cleaned_input.pop("private_metadata", None)
+        instance = cls.construct_instance(instance, cleaned_input)
```
korycins, line 211: If the approach from this PR is "common" thing that we want to introduce, then I would consider extending ModelMutation to have something like :
```python
def update_instance(cls, instance, cleaned_data)-> tuple[Object, list[CHANGED_FIELD_NAME]]:
```
Then as a result you could get the `instance` and fields that have been changed (you could use similar logic from `construct_instance` but with additional tracking of changed field names). 
Then instead of passing `base_fields_changed` to the `_save`, you could pass these modified fields.
Then in `_save`
```python
@classmethod
def _save(cls, info: ResolveInfo, instance, cleaned_input, modifed_field_names):
    ...
    if modifed_field_names:
        instance.save(updated_fields=modifed_field_names)
```
So instead of making whole `save` you will call save only on modified fields. Additionally, introducing similar approach for mutations like ProductUpdated etc should be easier

tomaszszymanski129: Yes, I will do so, but in separate PR, for now we just want to quickly optimize case "If nothing changed, do not save at all".

korycins: Sure, sounds good 👍

saleor/graphql/product/mutations/product_variant/product_variant_update.py
```diff
@@ -142,6 +146,40 @@ def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
         ).values_list("channel_id", flat=True)
         cls.call_event(mark_active_catalogue_promotion_rules_as_dirty, channel_ids)
 
+    @classmethod
+    def _save(cls, info: ResolveInfo, instance, cleaned_input, base_fields_changed):
+        new_variant = instance.pk is None
+        cls.set_track_inventory(info, instance, cleaned_input)
```
fowczarek, line 152: This change will not be saved when `base_fields_changed` is False. 
IMO this code should be executed before calculating `base_fields_changed`.

saleor/graphql/product/mutations/product_variant/product_variant_update.py
```diff
@@ -142,6 +146,40 @@ def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
...
+            if stocks := cleaned_input.get("stocks"):
+                cls.create_variant_stocks(instance, stocks)
+                variant_modified = True
+            if attributes := cleaned_input.get("attributes"):
+                AttributeAssignmentMixin.save(instance, attributes)
+                variant_modified = True
+
+            if not instance.name:
+                generate_and_set_variant_name(instance, cleaned_input.get("sku"))
+                variant_modified = True
```
fowczarek, line 170: Those changes will not be saved. (Save is in line 156.) 
IMO this code should be executed before calculating base_fields_changed.

saleor/graphql/product/mutations/product_variant/product_variant_update.py
```diff
@@ -142,6 +146,40 @@ def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
...
+
+            if variant_modified:
+                manager = get_plugin_manager_promise(info.context).get()
+                instance.product.search_index_dirty = True
+                instance.product.save(update_fields=["search_index_dirty"])
```
fowczarek, line 175: Duplicated save from line 160. We could merge them.

fowczarek (changes requested):

fowczarek (approved):

PR #16986, opened 2024-11-12
Recalculate prices even when tax app returns empty data.

zedzior:
I want to merge this change to revert the change introduced in https://github.com/saleor/saleor/pull/16696. Skipping price assignment for tax app strategy is redundant as `TaxableObjectLine` return base prices, which don't include order-level discounts anyway.

When taxes are calculated with tax app, we update prices with data received from tax app. In case of tax data is empty (ie. lack of address) we currently skip all the calculation. This PR ensure we calculate at least untaxed values. 

This PR also covers order price calculation with tax app strategy path with tests.

Issue: https://linear.app/saleor/issue/SHOPX-1594
Issue: https://linear.app/saleor/issue/SHOPX-1362

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/order/calculations.py
+++ b/saleor/order/calculations.py
@@ -148,17 +148,9 @@ def _recalculate_prices(
 
     order.tax_error = None
 
-    # Flat rates, plugin and tax exemption strategies require to propagate order-level
-    # discounts to order lines. Tax app does it itself.
-    using_tax_app = (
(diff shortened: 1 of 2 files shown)


zedzior:
Changelog skipped, because this is a fix for changes introduced in the same version (#16696)

korycins (approved):

PR #16997, opened 2024-11-13
Fix risk of updating the already deleted checkout line

korycins:
I want to merge this change because it covers the situation when we want to update the CheckoutLine which does not exist anymore.  This can happen when checkoutLineUpdate is called at similar time as any other mutations that could delete checkout or checkout-line.

Internal task: https://linear.app/saleor/issue/MERX-1268/attributeerror-nonetype-object-has-no-attribute-price

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/mutations/checkout_lines_update.py
+++ b/saleor/graphql/checkout/mutations/checkout_lines_update.py
@@ -3,14 +3,13 @@ from django.forms import ValidationError
 
 from ....checkout.error_codes import CheckoutErrorCode
 from ....checkout.fetch import CheckoutLineInfo
+from ....core.exceptions import NonExistingCheckoutLines
 from ....warehouse.reservations import is_reservation_enabled
 from ....webhook.event_types import WebhookEventAsyncType
 from ...app.dataloaders import get_app_promise
 from ...checkout.types import CheckoutLine
 from ...core import ResolveInfo
-from ...core.descriptions import (
-    DEPRECATED_IN_3X_INPUT,
-)
+from ...core.descriptions import DEPRECATED_IN_3X_INPUT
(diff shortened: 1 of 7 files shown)


IKarbowiak (approved):

fowczarek (approved):

saleor/graphql/checkout/mutations/checkout_lines_update.py
```diff
@@ -162,10 +160,47 @@ def clean_input(
...
+                checkout,
+                variants,
+                checkout_lines_data,
+                checkout_info,
+                replace=True,
+                # set to true, as during the update we want to be sure that any deleted line
+                # in the meantime will raise an exception instead of creating the new line.
+                raise_error_for_missing_lines=True,
+            )
+        except NonExistingCheckoutLines as e:
```
maarcingebala, line 188: Maybe I'm missing something, but why do we have error handling in `checkoutLinesUpdate`, but not in the 
 `checkoutLinesAdd`?

korycins: @maarcingebala  because the flow is different in both cases. I mean:
- when calling `add`, you expect to add the variant to the checkout.  You don't relay in the input on `CheckoutLine.id`, but on the provided `variant.id`, so as a result you want to have a line with the variant from the input.  No matter if the `CheckoutLine` with `variant.id` is present or not, you want to have a single line with this `variant.id`
- but when calling `update`, you provide specific `CheckoutLine.id` that you want to update. If the line is not present as it was deleted in the meantime, we raise the same exception as we did when validating the input.

PR #17028, opened 2024-11-18
Fix decreasing voucher code usage after changing `includeDraftOrderInVoucherUsage` to false.

zedzior:
When the `includeDraftOrderInVoucherUsage` flag is changed from `true` to `false`, multiple-use voucher code usages are reduced by 1, but should be reduced by the number of draft orders that utilize the code.

internal issue: https://linear.app/saleor/issue/SHOPX-1090

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/discount/tasks.py
+++ b/saleor/discount/tasks.py
@@ -227,12 +227,12 @@ def decrease_voucher_code_usage_of_draft_orders(channel_id: int):
         .filter(code__in=codes)
         .values_list("pk", flat=True)
     )
-    decrease_voucher_codes_usage_task.delay(list(voucher_code_ids))
+    decrease_voucher_codes_usage_task.delay(list(voucher_code_ids), list(codes))
 
 
 @app.task
 @allow_writer()
-def decrease_voucher_codes_usage_task(voucher_code_ids):
+def decrease_voucher_codes_usage_task(voucher_code_ids, codes):
     # Batch of size 1000 takes ~1sec and consumes ~20mb at peak
     BATCH_SIZE = 1000
     ids = voucher_code_ids[:BATCH_SIZE]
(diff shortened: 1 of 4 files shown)


IKarbowiak (approved):
Docs need to be updated as well

saleor/discount/tasks.py
```diff
@@ -227,12 +227,12 @@ def decrease_voucher_code_usage_of_draft_orders(channel_id: int):
...
     )
-    decrease_voucher_codes_usage_task.delay(list(voucher_code_ids))
+    decrease_voucher_codes_usage_task.delay(list(voucher_code_ids), list(codes))
 
 
 @app.task
 @allow_writer()
-def decrease_voucher_codes_usage_task(voucher_code_ids):
+def decrease_voucher_codes_usage_task(voucher_code_ids, codes):
     # Batch of size 1000 takes ~1sec and consumes ~20mb at peak
```
kadewu, line 236: Did you check if the comment here is still valid?

Side note: I also wonder if using `only` in this task still makes sense (not quite sure looking just at the code), so I will also point, that it may be worth double-checking the generated SQL's in this task if there will be high memory increase after changes.

kadewu (approved):

saleor/discount/tasks.py
```diff
@@ -227,12 +227,12 @@ def decrease_voucher_code_usage_of_draft_orders(channel_id: int):
...
         .values_list("pk", flat=True)
     )
-    decrease_voucher_codes_usage_task.delay(list(voucher_code_ids))
+    decrease_voucher_codes_usage_task.delay(list(voucher_code_ids), list(codes))
 
 
 @app.task
 @allow_writer()
-def decrease_voucher_codes_usage_task(voucher_code_ids):
+def decrease_voucher_codes_usage_task(voucher_code_ids, codes):
```
korycins, line 235: Just to remind you. If you're going to port this to existing versions, then `code` should be optional. 
(You can have a task in the queue, scheduled by previous implementation, where the new implementation will pick it up).

zedzior: Noted

saleor/discount/tasks.py
zedzior: The sql queries are not changed. This PR pass the result of the query which was already evaluated and there is no significant impact on the task performance

zedzior:
> Docs need to be updated as well

Docs PR: https://github.com/saleor/saleor-docs/pull/1398

PR #17034, opened 2024-11-19
Handle long message when reporting transaction event

IKarbowiak:
Previously, when the transaction event with the long massage (above 512 characters) was reported, the error was raised.
After the changes, the massage is truncated, and the warning is logged.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation: https://github.com/saleor/saleor-docs/pull/1396#pullrequestreview-2447716806

# Pull Request Checklist

### DIFF
--- a/saleor/payment/utils.py
+++ b/saleor/payment/utils.py
@@ -880,6 +880,16 @@ def parse_transaction_event_data(
         logger.warning(missing_msg, "result")
         error_field_msg.append(missing_msg % "result")
 
+    message = event_data.get("message", "")
+    if len(message) > 512:
+        message = truncate_message(message)
+        field_limit_exceeded_msg = (
+            "Value for field: %s in response of transaction action webhook "
+            "exceeds the character field limit. Message has been truncated."
+        )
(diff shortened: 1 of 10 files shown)


kadewu (approved):

saleor/payment/utils.py
```diff
@@ -977,6 +985,10 @@ def parse_transaction_action_data(
     )
 
 
+def truncate_message(message: str):
+    return message[:509] + "..." if len(message) >= 512 else message
```
korycins, line 989: shouldn't be `if len(message) > 512` or `if len(message) >= 513`?  Right now It looks like that providing message with 512 characters will cut it to 509 🤔

korycins (commented):
IMO, would be also good to have the same test for `transactionProcess`, and the logic that calls request-event (like charge request) - `saleor.webhook.transport.synchronous.transport.handle_transaction_request_task`.
Also, what about `transactionCreate` and `transactionUpdate`?

korycins (approved):

kadewu (approved):

PR #17036, opened 2024-11-20
Improve performance of bulk translate mutations

kadewu:
Improving number of queries and time execution of bulk translate mutations.

I've run the following tests against 2000 attributes translations.

Current main:
```
Queries: 9000 in 17.2680 (9000+ in fact, as this is limit set in django.db)

Memory during query execution:
resident size: 285MB
heap size: 228MB
```

After changes:
```
Queries: 149 in 0.6390

Memory during query execution:
resident size: 287MB
heap size: 228MB
```

I've also checked the same against smaller number (300), just to confirm that we don't have bigger delta in memory consumption against main when input drastically was increased:

Current main:
```
Memory during query execution:
resident size: 265MB
heap size: 217MB
```

After changes:
```
Memory during query execution:
resident size: 270MB
heap size: 220MB
```

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/plugins/webhook/plugin.py
+++ b/saleor/plugins/webhook/plugin.py
@@ -2483,6 +2483,62 @@ class WebhookPlugin(BasePlugin):
             )
         return previous_value
 
+    def translations_created(
+        self,
+        translations: list["Translation"],
+        previous_value: None,
+        webhooks=None,
+    ) -> None:
+        return self._handle_translations(
+            translations,
+            previous_value,
+            event_type=WebhookEventAsyncType.TRANSLATION_CREATED,
+            webhooks=webhooks,
+        )
+
+    def translations_updated(
+        self,
+        translations: list["Translation"],
+        previous_value: None,
+        webhooks=None,
+    ) -> None:
+        return self._handle_translations(
+            translations,
+            previous_value,
+            event_type=WebhookEventAsyncType.TRANSLATION_UPDATED,
+            webhooks=webhooks,
+        )
+
+    def _handle_translations(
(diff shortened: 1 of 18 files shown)


saleor/plugins/webhook/plugin.py
```diff
@@ -2483,6 +2483,62 @@ def translation_updated(
             )
         return previous_value
 
+    def translations_created(
```
korycins, line 2486: What about the previously used methods? `translation_created` and `translation_updated`?

korycins: We should also change `BasePlugin` definition with replacement of old methods

saleor/plugins/webhook/plugin.py
```diff
@@ -2483,6 +2483,62 @@ def translation_updated(
...
+        webhooks=None,
+    ) -> None:
+        return self._handle_translations(
+            translations,
+            previous_value,
+            event_type=WebhookEventAsyncType.TRANSLATION_UPDATED,
+            webhooks=webhooks,
+        )
+
+    def _handle_translations(
```
fowczarek, line 2512: We should add tests for this function 🚓

saleor/plugins/webhook/plugin.py
kadewu: I will remove usage of the old methods and add definition to `BasePlugin` as it is missing currently.

IKarbowiak (approved):

PR #17038, opened 2024-11-20
Do not update EventAttempt for success event delivery

korycins:
I want to merge this change because right now when we have a successful webhook delivery, we delete the `EventDelivery` object at the end of the task. But in the meantime, we also always update `DeliveryAttempt` and `EventDelivery` with new status. It means two additional write db  queries, for the object that will be deleted a few lines later. 

With this PR, we skip updating EventAttempt for successful deliveries and update the status for EventDelivery.  As both of them will be deleted.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/webhook/transport/asynchronous/transport.py
+++ b/saleor/webhook/transport/asynchronous/transport.py
@@ -567,7 +567,6 @@ def send_webhook_request_async(self, event_delivery_id) -> None:
     webhook = delivery.webhook
     domain = get_domain()
     attempt = create_attempt(delivery, self.request.id)
-    delivery_status = EventDeliveryStatus.SUCCESS
 
     try:
         if not delivery.payload:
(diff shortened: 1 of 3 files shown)


fowczarek (dismissed):

saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -567,7 +567,6 @@ def send_webhook_request_async(self, event_delivery_id) -> None:
     webhook = delivery.webhook
     domain = get_domain()
     attempt = create_attempt(delivery, self.request.id)
```
przlada, line 569: Theoretically, we could optimize this further by using `create_attempt` with `with_save=False`, eliminating one more DB operation. 

This would require modifying `attempt_update` into `attempt_create_or_update` and save the attempt object if it doesn't exist in the DB yet. However, there are two potential issues:
- The attempt's `created_at` field would show a timestamp after the webhook was sent (due to `auto_now_add`), though this could be fixed
- For successful attempts, observability would report the attempt's ID as `EventDeliveryAttempt:None` (base64 encoded) since the object wouldn't be ever saved to the DB. But I think this should be acceptable and non-breaking

However, if we decide to proceed with this idea, it might be better addressed in a separate PR.

kadewu (approved):

saleor/webhook/transport/asynchronous/transport.py
korycins: Sounds like a good idea. I've created a ticket for that: https://linear.app/saleor/issue/MERX-1288/delay-creation-of-eventdeliveryattempt

przlada (approved):

IKarbowiak (approved):

PR #17039, opened 2024-11-20
Update test for automatic checkout completion to not overcharge the order

IKarbowiak:
Update test for automatic checkout completion to not overcharge the order

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/tests/test_tasks.py
+++ b/saleor/checkout/tests/test_tasks.py
@@ -8,9 +8,14 @@ import graphene
 import pytest
 from django.utils import timezone
 
-from ...order import OrderEvents
+from ...core.taxes import zero_money
+from ...order import OrderChargeStatus, OrderEvents
 from ...order.models import Order
+from ...plugins.manager import get_plugins_manager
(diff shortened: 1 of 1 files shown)


saleor/checkout/tests/test_tasks.py
```diff
@@ -619,6 +636,7 @@ def test_automatic_checkout_completion_task_transaction_flow(
     # then
     assert not Checkout.objects.filter(pk=checkout_pk).exists()
     order = Order.objects.filter(checkout_token=checkout_pk).first()
+    assert order.charge_status
```
korycins, line 639: Shouldn't be `assert order.charge_status == <expected-charge-status>? Right now it will always be `true` as default value is not `None`

korycins: And also, assert from line 640 should be called first 🤔

kadewu (approved):

korycins (approved):

PR #17051, opened 2024-11-21
Truncate lenghty responses in EventDeliveryAttempt objects

wcislo-saleor:


przlada (dismissed):

tomaszszymanski129 (approved):

saleor/webhook/transport/utils.py
```diff
@@ -466,7 +466,11 @@ def attempt_update(
     webhook_response: "WebhookResponse",
 ):
     attempt.duration = webhook_response.duration
-    attempt.response = webhook_response.content
+    attempt.response = webhook_response.content[
```
korycins, line 469: Wouldn't be safer to additionally confirm that the `if webhook_response.content` first?  I am know, that based on typing `WebhookResponse.content` should be always str, but we use it with external request. On the DB `attempt.response` is also nullable field.
I am thinking about a situation where somehow the one interface would return `content` as Null. This could cause an unhandled exception

wcislo-saleor: Thanks for the comment. I reviewed all usages of `WebhookResponse` as well as `requests` source code (wrt. to `text` property) and to my understanding we're protected against `None` in all these places.

korycins: Did you also confirm that the logic for sending webhooks to Pub/Sub and SQS won't allow the `None` value to be returned?

wcislo-saleor: I did not check exactly that. Although I checked the following.

For SQS we're doing: `json.dumps(...)` so in case `None` is returned:
```
>>> json.dumps(None)
'null'
```

For Google Cloud PubSub we're just reading the response from `publish` method call, where return type is `concurrent.futures._base.Future[str]` (so basically a `str`).
https://cloud.google.com/python/docs/reference/pubsublite/latest/google.cloud.pubsublite.cloudpubsub.publisher_client.PublisherClient#google_cloud_pubsublite_cloudpubsub_publisher_client_PublisherClient_publish

I understand your concern and inclination to be on the safer side. So now I'm wondering whether `if webhook_response.content` is enough. What if a `content` is _something_ but not a `str`? What do you think about checking for `str` type here?

korycins: As for DB `EventDeliveryAttempt.response` is nullable `TextField`. I would say, that adding `if response:` would be enough IMO,  but  you could also make additional validation if you want. For me, I just want to be sure, that something simple like truncating the response content, would not cause unhandled exception.

wcislo-saleor: https://github.com/saleor/saleor/pull/17067

PR #17055, opened 2024-11-21
Truncate lenghty responses in EventDeliveryAttempt objects

wcislo-saleor:


przlada (approved):
Why aren't we porting the tests here?

PR #17059, opened 2024-11-22
Fix missing currency in voucher discount objects

zedzior:
I want to merge this change because adds missing currency in `OrderDiscount` objects of type `VOUCHER`

Internal issue: https://linear.app/saleor/issue/SHOPX-1616

# Impact

- [x] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


IKarbowiak (approved):

korycins (commented):
IMHO, this change should also include the migration to fix the existing vouchers where currency is missing 🤔

korycins (approved):

saleor/discount/migrations/0084_set_missing_currency.py
IKarbowiak, line 1: If you want to port the changes to the older versions you should firstly create the migration on the older version as the last migration number on 3.19 might differ from the one on main, and you will have inconsistency. So I suggest creating a ports for older versions before merging it.

saleor/discount/migrations/0084_set_missing_currency.py
```diff
@@ -0,0 +1,49 @@
...
+BATCH_SIZE = 5000
+
+
+def set_discount_currency(apps, _schema_editor):
+    OrderDiscount = apps.get_model("discount", "OrderDiscount")
+    qs = (
+        OrderDiscount.objects.filter(currency="")
+        .select_related("order")
+        .order_by("id")
+        .only("id", "currency", "order__currency")
```
kadewu, line 16: This statement is redundant, as we will do `values_list` on that `qs` anyway in `queryset_in_batches`

kadewu (approved):

PR #17103, opened 2024-11-28
Add timeout to pubsub interface used by webhook events

korycins:
I want to merge this change because it adds missing timeout for pubsub webhook handler. 
Without this, the `future.result()` can wait 600 seconds.  This PR reduces the timeout to the default value that we use. 
Our retry flow will pick up the event which failed with the timeout. 
Internal task link: https://linear.app/saleor/issue/MERX-1285/log-timeout-on-webhooks-sent-via-pubsub

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/settings.py
```diff
@@ -964,8 +964,10 @@ def SENTRY_INIT(dsn: str, sentry_opts: dict):
 # Default timeout for external requests.
 COMMON_REQUESTS_TIMEOUT = (REQUESTS_CONN_EST_TIMEOUT, 18)
 
-WEBHOOK_TIMEOUT = (REQUESTS_CONN_EST_TIMEOUT, 18)
-WEBHOOK_SYNC_TIMEOUT = (REQUESTS_CONN_EST_TIMEOUT, 18)
+WEBHOOK_WAITING_FOR_RESPONSE_TIMEOUT = 18
+
+WEBHOOK_TIMEOUT = (REQUESTS_CONN_EST_TIMEOUT, WEBHOOK_WAITING_FOR_RESPONSE_TIMEOUT)
+WEBHOOK_SYNC_TIMEOUT = (REQUESTS_CONN_EST_TIMEOUT, WEBHOOK_WAITING_FOR_RESPONSE_TIMEOUT)
```
IKarbowiak, line 970: Cannot it be one setting as both are the same?

kadewu (approved):

saleor/settings.py
korycins: I don't have an opinion here. From one side, this could be a single env.  On the second side, having separate envs give you more flexibility. Like you would like to have a longer timeout on async webhook, as it takes some time to process the request by your app for asynchronous webhook, but you want to keep the default timeout for sync webhook.

korycins: We agreed to keep this as it is for now. Marking as resolved

PR #17125, opened 2024-12-02
Use denormalized prices during order update for catalogue discounts

zedzior:

I want to merge this change because we shouldn't refresh order base variant prices during the order update. By base variant price I mean the channel listing price with eventual catalogue discount. This price should be constant during the entire order lifecycle.

The main assumption of the PR is that we should retrieve order line's variant price (from channel listings) only when adding new line. All future recalculations should use denormalised price. Same for the catalog promotions. Only the amount of the catalogue discount can be refreshed when updating order line with new quantity or custom price. This means that we don't need to retrieve actual channel listings and promotion rules every time we trigger `fetch_order_prices_if_expired`. The whole logic for creating/updating catalogue promotion is moved to mutations which allow to add new order line. The logic for checkout remains the same.

This PR makes quite a lot of changes to existing fixtures. It is because the order line with the properly applied catalog discount should have:
- `orderLineDiscount` object updated
- `VariantChannelListingPromotionRule` object updated
- `VariantChannelListing` object updated
- `line.base_unit_price` updated

Internal issue: https://linear.app/saleor/issue/SHOPX-1672

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,4 +54,3 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fix checkout funds releasing task - #17198 by @IKarbowiak
 - Fixed 'healthcheck' middleware (`/health/` endpoint) not forwarding incoming traffic whenever the protocol wasn't HTTP (such as WebSocket or Lifespan) - #17248 by @NyanKiyoshi
 - Use denormalized discount prices during order update - #17160 by @zedzior
-- Use denormalized prices during order update for catalogue discounts - #17125 by @zedzior
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,3 +54,4 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fix checkout funds releasing task - #17198 by @IKarbowiak
 - Fixed 'healthcheck' middleware (`/health/` endpoint) not forwarding incoming traffic whenever the protocol wasn't HTTP (such as WebSocket or Lifespan) - #17248 by @NyanKiyoshi
 - Use denormalized discount prices during order update - #17160 by @zedzior
+- Use denormalized prices during order update for catalogue discounts - #17125 by @zedzior
--- a/saleor/discount/tests/test_utils/test_create_or_update_discount_objects_from_promotion_for_order.py
+++ b/saleor/discount/tests/test_utils/test_create_or_update_discount_objects_from_promotion_for_order.py
@@ -1,7 +1,6 @@
 from decimal import Decimal
 
 import graphene
-import pytest
 
 from ....order.fetch import fetch_draft_order_lines_info
 from ....product.models import (
@@ -399,71 +398,6 @@ def test_multiple_rules_no_discount_applied(
     assert not OrderDiscount.objects.exists()
 
 
-def test_update_catalogue_discount_updates_discount_amount_only(
-    order_with_lines_and_catalogue_promotion, catalogue_promotion_without_rules
-):
-    # given
-    order = order_with_lines_and_catalogue_promotion
-    promotion = catalogue_promotion_without_rules
-
-    channel = order.channel
-    line = order.lines.first()
-    new_line_quantity = line.quantity + 1
-    line.quantity = new_line_quantity
-    line.save(update_fields=["quantity"])
-
-    variant = line.variant
-    assert OrderLineDiscount.objects.count() == 1
-    discount_to_update = line.discounts.get()
-
-    new_reward_value = Decimal(6)
-    initial_rule = promotion.rules.first()
-    initial_rule_reward_value = initial_rule.reward_value
-    assert new_reward_value > initial_rule_reward_value
-    new_rule = promotion.rules.create(
-        name="New catalogue rule fixed",
-        catalogue_predicate={
-            "variantPredicate": {
-                "ids": [graphene.Node.to_global_id("ProductVariant", variant.id)]
-            }
-        },
-        reward_value_type=RewardValueType.FIXED,
-        reward_value=new_reward_value,
-    )
-    new_rule.channels.add(channel)
-
-    variant_channel_listing = variant.channel_listings.get(channel=channel)
-    undiscounted_price = variant_channel_listing.price_amount
-    variant_channel_listing.discounted_price_amount = (
-        undiscounted_price - new_reward_value
-    )
-    variant_channel_listing.save(update_fields=["discounted_price_amount"])
-
-    variant_rule_listing = variant_channel_listing.variantlistingpromotionrule.get()
-    variant_rule_listing.discount_amount = new_reward_value
-    variant_rule_listing.promotion_rule = new_rule
-    variant_rule_listing.save(update_fields=["discount_amount", "promotion_rule"])
-
-    lines_info = fetch_draft_order_lines_info(order)
-
-    # when
-    create_or_update_discount_objects_from_promotion_for_order(order, lines_info)
-
-    # then
-    assert OrderLineDiscount.objects.count() == 1
-    discount = OrderLineDiscount.objects.get()
-    assert discount.amount_value == initial_rule_reward_value * new_line_quantity
-
-    # other fields shoudn't be changed
-    assert discount_to_update.id == discount.id
-    assert discount.line == line
-    assert discount.promotion_rule == initial_rule
-    assert discount.type == DiscountType.PROMOTION
-    assert discount.value_type == RewardValueType.FIXED
-    assert discount.value == initial_rule_reward_value
-    assert discount.currency == channel.currency_code
-
-
 def test_update_order_discount_subtotal(
     order_with_lines_and_order_promotion, order_promotion_without_rules
 ):
@@ -543,7 +477,6 @@ def test_update_gift_discount_new_gift_available(
     assert gift_line.variant == variant
 
 
-@pytest.mark.django_db(transaction=True)
 def test_create_multiple_catalogue_discounts_for_the_same_line_is_not_allowed(
     order_line,
     catalogue_promotion,
--- a/saleor/discount/tests/test_utils/test_create_or_update_discount_objects_from_promotion_for_order.py
+++ b/saleor/discount/tests/test_utils/test_create_or_update_discount_objects_from_promotion_for_order.py
@@ -2,7 +2,6 @@ from decimal import Decimal
 
 import graphene
 import pytest
-from django.db import IntegrityError
 
 from ....order.fetch import fetch_draft_order_lines_info
 from ....product.models import (
@@ -545,7 +544,7 @@ def test_update_gift_discount_new_gift_available(
 
 
 @pytest.mark.django_db(transaction=True)
-def test_create_multiple_catalogue_discounts_for_the_same_line_raise_unique_error(
+def test_create_multiple_catalogue_discounts_for_the_same_line_is_not_allowed(
     order_line,
     catalogue_promotion,
     promotion_translation_fr,
@@ -603,9 +602,10 @@ def test_create_multiple_catalogue_discounts_for_the_same_line_raise_unique_erro
         ),
     ]
 
-    # when & then
-    # TODO zedzior: sprawdz czy w funkcji nie trzeba dodac ignore_conflicts=True
-    with pytest.raises(IntegrityError):
-        create_order_line_discount_objects_for_catalogue_promotions(
-            order_line, rules_info, order.channel
-        )
+    # when
+    create_order_line_discount_objects_for_catalogue_promotions(
+        order_line, rules_info, order.channel
+    )
+
+    # then
+    assert order_line.discounts.count() == 1
--- a/saleor/discount/tests/test_utils/test_create_or_update_discount_objects_from_promotion_for_order.py
+++ b/saleor/discount/tests/test_utils/test_create_or_update_discount_objects_from_promotion_for_order.py
@@ -1,6 +1,8 @@
 from decimal import Decimal
 
 import graphene
+import pytest
+from django.db import IntegrityError
 
 from ....order.fetch import fetch_draft_order_lines_info
 from ....product.models import (
@@ -9,8 +11,12 @@ from ....product.models import (
 )
 from ....warehouse.models import Stock
 from ... import DiscountType, RewardType, RewardValueType
+from ...interface import VariantPromotionRuleInfo
 from ...models import OrderDiscount, OrderLineDiscount
-from ...utils.order import create_or_update_discount_objects_from_promotion_for_order
+from ...utils.order import (
+    create_or_update_discount_objects_from_promotion_for_order,
+    create_order_line_discount_objects_for_catalogue_promotions,
+)
 
 
 def test_create_catalogue_discount_fixed(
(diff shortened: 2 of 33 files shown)


IKarbowiak:
@zedzior the docs should be updated as well

saleor/discount/utils/order.py
```diff
@@ -251,3 +315,49 @@ def create_or_update_line_discount_objects_for_manual_discounts(lines_info):
...
+
+def create_order_line_discount_objects_for_catalogue_promotions(
+    line: "OrderLine",
+    rules_info: Iterable[VariantPromotionRuleInfo],
+    channel: Channel,
+) -> Iterable["OrderLineDiscount"]:
+    from ...order.utils import order_qs_select_for_update
+
+    line_discounts_to_create_inputs: list[dict] = []
+    for rule_info in rules_info:
```
IKarbowiak, line 328: As there is assumption that only one catalogue discount object can be created, why are you iterating here through all `rules_info`? Maybe line field should be changed from `rules_info` to `rule_info` - single one, instead of list.

saleor/order/utils.py
```diff
@@ -585,6 +563,14 @@ def change_order_line_quantity(
             update_fields.extend(fields)
         else:
             line.save(update_fields=fields)
+
+        if catalogue_discount := line.discounts.filter(
```
korycins, line 567: Why do you take this directly from DB instead of using `line_info.line_discounts`?

saleor/discount/utils/checkout.py
```diff
@@ -86,6 +93,96 @@ def create_checkout_line_discount_objects_for_catalogue_promotions(
     )
 
 
+def prepare_checkout_line_discount_objects_for_catalogue_promotions(lines_info):
```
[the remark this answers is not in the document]
korycins: Typing for returning value would be also great

CHANGELOG.md
```diff
@@ -46,3 +46,4 @@ All notable, unreleased changes to this project will be documented in this file.
 - Drop tax data line number validation for Avatax plugin - #16917 by @zedzior
 - Fix decreasing voucher code usage after changing `includeDraftOrderInVoucherUsage` to false - #17028 by @zedzior
 - Fix undiscounted price taxation when prices are entered with taxes - #16992 by @zedzior
+- Use denormalized prices during order update for catalogue discounts - #17125 by @zedzior
```
IKarbowiak, line 49: Are you gonna port those changes to 3.20?

saleor/discount/utils/order.py
```diff
@@ -48,19 +53,56 @@ def create_or_update_discount_objects_from_promotion_for_order(
     lines_info: list["EditableOrderLineInfo"],
     database_connection_name: str = settings.DATABASE_CONNECTION_DEFAULT_NAME,
 ):
-    create_order_line_discount_objects_for_catalogue_promotions(lines_info)
-    # base unit price must reflect all actual catalogue discounts
+    delete_order_line_discount_objects_for_catalogue_promotions(lines_info)
+    # Update base price with catalogue discount reward. This price will be further used:
+    # - to qualify order for order promotion
+    # - to apply vouchers
     _update_base_unit_price_amount_for_catalogue_promotion(lines_info)
```
korycins, line 60: Do we still need this line? It makes:
```python
line_info.line.base_unit_price = line_info.variant_discounted_price
```
But as we operating on denormalized values, this looks like something not needed 🤔

IKarbowiak (commented):
Not sure if those changes are enough. In case the `OrderUpdate` is called with `shipping_address` or `billing_address` in the input data, the order prices will be invalidated, and when order prices will be fetched, the `fetch_order_prices_if_expired` will ba called, and the voucher discount will be updated. In case something changed with voucher, the prices will be changed.
The `DraftOrderComplete` also calls `fetch_order_prices_if_expired`.

Also, we should definitely add e2e tests for those changes, like the draft order is created, there is catalogue discount, and voucher assigned, then the new line is added, shipping address update, completing draft ordrer etc.

saleor/discount/utils/order.py
```diff
@@ -48,19 +53,56 @@ def create_or_update_discount_objects_from_promotion_for_order(
     lines_info: list["EditableOrderLineInfo"],
     database_connection_name: str = settings.DATABASE_CONNECTION_DEFAULT_NAME,
 ):
-    create_order_line_discount_objects_for_catalogue_promotions(lines_info)
-    # base unit price must reflect all actual catalogue discounts
+    delete_order_line_discount_objects_for_catalogue_promotions(lines_info)
```
korycins, line 56: As we operate on denoramilzed value, what is the reason for calling this each time when `fetch_order_prices_if_expired` is called?
I mean, any logic related to order_promotion or catalogue_promotion is applicable when creating new line for order, right? In that case we don't need to call this when we need to re-calculate prices, right?
Also, I've checked for example `OrderLineDiscountUpdate`. When calling it, we create new `OrderLineDiscount` object. As MANUAL overwrites the catalogue promotion, this should happen there IMHO. Right now, some part of the logic happens there [ `OrderLineDiscountUpdate`] (adding MANUAL), and some part of the logic here (delete CATALOGUE).

IKarbowiak:
Also, we should probably add a mutation that allow forcing the order price recalculations including fetching the current variant prices

saleor/discount/tests/test_utils/test_create_or_update_discount_objects_from_promotion_for_order.py
```diff
@@ -129,9 +134,6 @@ def test_create_catalogue_discount_percentage(
     assert discount.name == f"{promotion.name}: {rule.name}"
     assert discount.reason == f"Promotion: {promotion_id}"
 
-    line = [line_info.line for line_info in lines_info if line_info.line == line_1][0]
-    assert line.base_unit_price_amount == Decimal(5)
```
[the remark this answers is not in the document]
zedzior: It is because this function doesn't change the base price anymore. Its purpose is to create disccount object

CHANGELOG.md
zedzior: Yes, I do plan to port it to 3.19. Does it mean it should't be here?

saleor/discount/utils/order.py
zedzior: It is here to make sure we won't endup with two discounts which shouldn't be combined. But maybe you are right, it should be done only when handling manual line discount and gift.

> I mean, any logic related to order_promotion or catalogue_promotion is applicable when creating new line for order, right?
This is not true for order_promotion.

saleor/discount/utils/order.py
zedzior: This ensure the proper order of discount application. First we apply catalogue discounts what results in first decreasing of line base price. Next we apply vouchers or order promotion to the decreased base prices.

The next two PRs will handle vouchers and order promotions, so maybe it will be removed. Do you suggest to handle all the disocunts in single PR?

saleor/discount/utils/order.py
```diff
@@ -48,19 +53,56 @@ def create_or_update_discount_objects_from_promotion_for_order(
...
+        return
+
+    for line_info in lines_info:
+        line = line_info.line
+
+        # get the existing catalogue discount for the line
+        if existing_discounts := line_info.get_catalogue_discounts():
+            # Line should never have multiple catalogue discounts associated. Before
+            # introducing unique_type on discount models, there was such a possibility.
+            line_discounts_to_remove.extend(existing_discounts[1:])
```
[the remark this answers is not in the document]
zedzior: This code was introduced before. I just split the function for order and checkout. 

As Maciek suggested, this can be probably deleted and do the all deletions directly in manual line discount and gift logic.

saleor/discount/utils/order.py
zedzior: Our whole codebase assumes multiple discounts application to single line. This is not the only place we do that and imo it should be corrected in separate PR

saleor/order/utils.py
zedzior: Good point, my bad

zedzior:
> Not sure if those changes are enough. In case the `OrderUpdate` is called with `shipping_address` or `billing_address` in the input data, the order prices will be invalidated, and when order prices will be fetched, the `fetch_order_prices_if_expired` will ba called, and the voucher discount will be updated. In case something changed with voucher, the prices will be changed. The `DraftOrderComplete` also calls `fetch_order_prices_if_expired`.
> 
> Also, we should definitely add e2e tests for those changes, like the draft order is created, there is catalogue discount, and voucher assigned, then the new line is added, shipping address update, completing draft ordrer etc.

I have 2 separate tasks to handle vouchers (https://linear.app/saleor/issue/SHOPX-1673/use-denormalised-prices-during-order-update-for-vouchers) and order promotions (https://linear.app/saleor/issue/SHOPX-1674/use-denormalised-prices-during-order-update-for-order-discounts). Do you suggest to make all the changes in a single PR?

zedzior:
> Also, we should probably add a mutation that allow forcing the order price recalculations including fetching the current variant prices

Do you suggest to handle it within this PR?

IKarbowiak:
> > Not sure if those changes are enough. In case the `OrderUpdate` is called with `shipping_address` or `billing_address` in the input data, the order prices will be invalidated, and when order prices will be fetched, the `fetch_order_prices_if_expired` will ba called, and the voucher discount will be updated. In case something changed with voucher, the prices will be changed. The `DraftOrderComplete` also calls `fetch_order_prices_if_expired`.
> > Also, we should definitely add e2e tests for those changes, like the draft order is created, there is catalogue discount, and voucher assigned, then the new line is added, shipping address update, completing draft ordrer etc.
> 
> I have 2 separate tasks to handle vouchers (https://linear.app/saleor/issue/SHOPX-1673/use-denormalised-prices-during-order-update-for-vouchers) and order promotions (https://linear.app/saleor/issue/SHOPX-1674/use-denormalised-prices-during-order-update-for-order-discounts). Do you suggest to make all the changes in a single PR?

For me it's ok to do it in separate PRs.

CHANGELOG.md
zedzior: I removed the entry

saleor/discount/utils/order.py
korycins: I am just thinking about the current logic called inside `fetch_order_prices_if_expired`. As we agreed that we should use de-normalized values on Order/OrderLine models, it doesn't make sense to make any calculations that would change the `base_price`, or any other fields that store the Money representation.
For example, what is the reason for calling `create_or_update_discount_objects_for_order` inside `fetch_order_prices_if_expired` in case of making an update on `shippingAddress`?

saleor/discount/utils/order.py
korycins: > This is not true for order_promotion.

I forgot about this one 😄 

But still, I think that discount objects re-creation should be triggered when we know that something has been changed. By "Something" I mean the action that would have an impact on base_price/total/discount objects. 

We can also handle this as a separate task, but please discuss it with your team. Right now, some logic for creating discount objects is called inside the mutation/adding order line, and some of the logic is called inside `fetch_order_prices_if_expired`.

saleor/discount/utils/order.py
zedzior: I don't know what is a rationale of invalidating prices after `shippingAddress` update, but let's have an example of updating `shippingMethod`. This for sure results in different shipping price. Let's look on examples:
1. We have shipping voucher applied to order -> we need to do an update of shipping discount amount. Where would you like to do the update?
2. Increased shipping price results in higher total price of an order. This change can qualify the order for some order promotion.

Next PR will handle vouchers to use denormalised discount prices. It will require to move the whole logic of voucher application out of  `fetch_order_prices_if_expired`. This change will probably get rid of manipulating with `base_price` inside the `create_or_update_discount_objects_for_order`.

saleor/order/utils.py
zedzior: My bad, this id `OrderLineData` type:
```
@dataclass
class OrderLineData:
    variant_id: Optional[str] = None
    variant: Optional[ProductVariant] = None
    line_id: Optional[str] = None
    price_override: Optional[Decimal] = None
    quantity: int = 0
    rules_info: Optional[Iterable[VariantPromotionRuleInfo]] = None
```

saleor/discount/utils/checkout.py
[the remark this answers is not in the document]
zedzior: Done

saleor/order/utils.py
zedzior: It was confusing so I added typing for orderLinesCreate and its utils :) Indeed it is `OrderLineInfo` type, however it only contains the line and quantity and it is constructed in the loop. The adding lines logic requires optimisation, to not create the lines in the loop, but using `bulk_create` instead. I cerated a ticket for that: https://linear.app/saleor/issue/SHOPX-1709

zedzior:
WIll be continued in Q1

saleor/discount/utils/promotion.py
```diff
@@ -371,6 +262,40 @@ def _update_promotion_discount(
     )
 
 
+def _update_catalogue_promotion_discount(
```
IKarbowiak, line 265: Looks like the function is missing tests 🚔

IKarbowiak (approved):

saleor/discount/utils/promotion.py
zedzior: This function is not used at all. Let me drop it

saleor/order/fetch.py
```diff
@@ -86,16 +82,7 @@ class EditableOrderLineInfo(LineInfo):
 
     @cached_property
     def variant_discounted_price(self) -> Money:
-        """Return the discounted variant price.
```
korycins, line 89: Actually, after this change, I am not sure if we still need `variant_discounted_price`. 
With previous logic, the main purpose of this property was: return the price from listing if present. If not present, calculated it. 
But with your changes, the `order_line.base_unit_price` will be static, for the whole process for single line. So I think, that we could drop this logic 🤔 
(Correct me if I'm am wrong)

saleor/order/utils.py
```diff
@@ -273,11 +275,16 @@ def create_order_line(
...
+            line_discounts = (
+                create_order_line_discount_objects_for_catalogue_promotions(
+                    line, rules_info, channel
+                )
+            )
             promotion = rules_info[0].promotion
             line.sale_id = get_sale_id(promotion)
             line.unit_discount_reason = (
-                prepare_promotion_discount_reason(promotion, line.sale_id)
+                prepare_promotion_discount_reason(rules_info[0].rule)
```
korycins, line 287: Isn't like this will trigger additional DB call?
`prepare_promotion_discount_reason` calls `rule.promotion`. `rule` i `PromotionRule` object. You already have the promotion inside `rules_info[0].promotion`

saleor/order/utils.py
```diff
@@ -363,6 +343,7 @@ def add_variant_to_order(
     channel = order.channel
 
     if line_data.line_id:
+        # It means there is an update of the order line with new quantity
```
korycins, line 346: It is good to have additional context, but that better would be to have self-explaining code. So instead of an additional comment, we could have:
```python
is_new_line = not line_data.line_id
if not is_new_line:
    ...
```

saleor/order/utils.py
```diff
@@ -565,21 +546,21 @@ def change_order_line_quantity(
         line.quantity = new_quantity
         total_price_net_amount = line.quantity * line.unit_price_net_amount
         total_price_gross_amount = line.quantity * line.unit_price_gross_amount
-        line.total_price_net_amount = total_price_net_amount.quantize(Decimal("0.001"))
-        line.total_price_gross_amount = total_price_gross_amount.quantize(
-            Decimal("0.001")
+        line.total_price_net_amount = quantize_price(total_price_net_amount, currency)
```
korycins, line 549: Hhm, do we actually need to make these calculations here? 
I mean, we agree that the base price is static, but we still need to get tax values from TaxPlugin/TaxApp/FlatRates. So when we change the `order_line_quantity`, we change the `_price` fields, then later we set `price_expiration` to True, and we calculate prices in the resolver. Right now I am thinking: do we get something here? The prices are marked as expired, so they will be calculated once again, right?

saleor/order/utils.py
```diff
@@ -1318,3 +1297,7 @@ def clean_order_line_quantities(order_lines, quantities_for_lines):
                     )
                 }
             )
+
+
+def order_qs_select_for_update():
```
korycins, line 1302: Can you move this closer to `order_lines_qs_select_for_update`? Having both closed, would be helpful in case, of searching for the function name, or understanding that, the usage of this helper should be mandatory or general code readability.

saleor/product/utils/variants.py
```diff
@@ -64,7 +64,7 @@ def fetch_variants_for_promotion_rules(
     for rule in rules.iterator():
         variants = get_variants_for_catalogue_predicate(
             rule.catalogue_predicate,
-            database_connection_name=settings.DATABASE_CONNECTION_REPLICA_NAME,
+            database_connection_name=database_connection_name,
```
korycins, line 67: This will cause the celery task like, `set_promotion_rule_variants_task`, will start using `DEFAULT` database. IMHO, you should change the default `database_connection_name` to `REPLICA`

saleor/order/utils.py
zedzior: Make sense, thanks

saleor/order/utils.py
zedzior: Good point

saleor/order/utils.py
zedzior: Correct. IMO this should be handled in another PR. There is already an internal task for this: https://linear.app/saleor/issue/EXT-1729/all-order-mutation-should-be-lazy. 
I don't want to extend this PR further as it already introduces a massive change. What do you think?

saleor/order/utils.py
zedzior: Moved

saleor/product/utils/variants.py
zedzior: Changed back

saleor/order/utils.py
korycins: IMO separate task sounds good

korycins (approved):

PR #17145, opened 2024-12-05
Circuit breakers observability

tomaszszymanski129:
I want to merge this change because it adds basic observability for circuit breakers. Target state would be keeping breaker events in the database but for now it should be sufficient.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/app/types.py
+++ b/saleor/graphql/app/types.py
@@ -54,13 +54,18 @@ from .dataloaders import (
     ThumbnailByAppInstallationIdSizeAndFormatLoader,
     app_promise_callback,
 )
-from .enums import AppExtensionMountEnum, AppExtensionTargetEnum, AppTypeEnum
+from .enums import (
+    AppExtensionMountEnum,
+    AppExtensionTargetEnum,
+    AppTypeEnum,
+    CircuitBreakerState,
+    CircuitBreakerStateEnum,
+)
 from .resolvers import (
     resolve_access_token_for_app,
     resolve_access_token_for_app_extension,
     resolve_app_extension_url,
 )
-from .utils import CircuitBreakerState
 
 # TODO: Remove the conditional when unit tests circular import is solved.
 breaker_board = None
@@ -535,7 +540,10 @@ class App(ModelObjectType[models.App]):
         required=True,
     )
     brand = graphene.Field(AppBrand, description="App's brand data.")
-    breaker_state = graphene.String(description="App sync webhooks operation state.")
+    breaker_state = CircuitBreakerStateEnum(
+        description="Circuit breaker state, if open, sync webhooks operation is disrupted.",
+        required=True,
+    )
 
     class Meta:
         description = "Represents app data."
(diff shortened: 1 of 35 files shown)


saleor/graphql/app/types.py
```diff
@@ -597,6 +609,16 @@ def resolve_brand(root: models.App, _info: ResolveInfo):
             return root
         return None
 
+    @staticmethod
+    def resolve_breaker_state(root: models.App, _info: ResolveInfo):
+        if not breaker_board:
+            return CircuitBreakerState.CLOSED
```
maarcingebala, line 615: The field is nullable so maybe when no breaker is used, it should return null to indicate that?

saleor/graphql/app/types.py
```diff
@@ -524,6 +535,7 @@ class App(ModelObjectType[models.App]):
         required=True,
     )
     brand = graphene.Field(AppBrand, description="App's brand data.")
+    breaker_state = graphene.String(description="App sync webhooks operation state.")
```
maarcingebala, line 538: This looks more like an enum in the schema. I also think the description is rather vague, and we should explicitly say that this is the state of the app's circuit breaker. If we had an enum in the schema, we could also describe what each state value exactly means.

I know that there was a suggestion not to use the "circuit breaker" term in the schema, but in this case, how should we call it to make sure that it's understandable to the user? @patrys Do you have any suggestion?

saleor/graphql/app/utils.py
```diff
@@ -40,3 +40,8 @@ def validate_app_is_not_removed(
                 )
             }
         )
+
+
+class CircuitBreakerState:
```
maarcingebala, line 45: Enums or enum-like classes are usually stored in the module's main `__init__` file, so I suggest `saleor/app/__init__.py` for this one.

fowczarek (approved):

saleor/graphql/app/types.py
tomaszszymanski129: I could do that, but I'm not sure we want to expose configuration details like this through the API?

maarcingebala: I'm just wondering if it's not confusing when no breaker is used and the field will return one of the breaker's states. But you can also look at this from this perspective - when there is no breaker, the state is always "closed". But then the field shouldn't be nullable.

saleor/graphql/app/types.py
```diff
@@ -53,13 +54,28 @@
...
+    CircuitBreakerState,
+    CircuitBreakerStateEnum,
+)
 from .resolvers import (
     resolve_access_token_for_app,
     resolve_access_token_for_app_extension,
     resolve_app_extension_url,
 )
 
+# TODO: Remove the conditional when unit tests circular import is solved.
```
maarcingebala, line 70: We should be able to remove this conditional now.

maarcingebala (approved):

PR #17176, opened 2024-12-18
Always create checkout metadata for new checkouts

korycins:
I want to merge this change because it adds logic, to always create the checkout's metadata object when creating the checkout.
Additionally, I've drop some UPDATE queries, where there is no need to make an update on CheckoutMetadata.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/utils.py
+++ b/saleor/checkout/utils.py
@@ -8,7 +8,7 @@ from uuid import UUID
 import graphene
 from django.conf import settings
 from django.core.exceptions import ValidationError
-from django.db import transaction
+from django.db import IntegrityError, transaction
 from django.db.models import prefetch_related_objects
 from django.utils import timezone
 from prices import Money
@@ -1012,7 +1012,6 @@ def clear_delivery_method(checkout_info: "CheckoutInfo"):
         shipping_channel_listings=checkout_info.shipping_channel_listings,
     )
 
-    delete_external_shipping_id(checkout=checkout)
     checkout.save(
         update_fields=[
             "shipping_method",
@@ -1020,7 +1019,8 @@ def clear_delivery_method(checkout_info: "CheckoutInfo"):
             "last_change",
         ]
     )
-    get_checkout_metadata(checkout).save()
+    if delete_external_shipping_id(checkout=checkout):
+        checkout.metadata_storage.save()
 
 
 def is_fully_paid(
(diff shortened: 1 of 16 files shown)


saleor/checkout/utils.py
```diff
@@ -1090,19 +1090,32 @@ def set_external_shipping_id(checkout: Checkout, app_shipping_id: str):
...
         metadata.save(update_fields=["private_metadata"])
+    return field_deleted
+
+
+@allow_writer()
+def create_checkout_metadata(checkout: "Checkout"):
+    try:
+        return CheckoutMetadata.objects.create(checkout=checkout)
+    except IntegrityError:
+        return checkout.metadata_storage
```
patrys, line 1118: We should never catch `IntegrityError`s in our code. Why not use `get_or_create` instead?

IKarbowiak (dismissed):

saleor/payment/gateways/adyen/utils/common.py
```diff
@@ -356,7 +356,12 @@ def request_data_for_gateway_config(
         country_code = country.code
     else:
         country_code = Country(settings.DEFAULT_COUNTRY).code
-    channel = get_checkout_metadata(checkout).get_value_from_metadata("channel", "web")
+    checkout_metadata = get_checkout_metadata(checkout)
+    if checkout_metadata:
+        channel = checkout_metadata.get_value_from_metadata("channel", "web")
+    else:
+        channel = "web"
```
kadewu, line 363: I know it's not in the scope of that pr, but what is this "web" value?
I see we have more occurrences of that string, maybe it's worth to extract it to some variable?

kadewu (approved):

saleor/payment/gateways/adyen/utils/common.py
korycins: @kadewu  It is a value sent to Adyen. It is used to determine the channel related to the processed payment. At least this is what I remember 😅 
Honestly, there is no reason to invest time in AdyenPlugin right now, as we are moving from plugins to apps.

jakubkuc (dismissed):

fowczarek (approved):

zedzior (approved):

PR #17198, opened 2024-12-31
Fix checkout funds releasing task

IKarbowiak:
Fix checkout funds releasing.
Missing `authorized_value` and `charged_value` in filters causes the task to clog, we were fetching still the same checkouts with transactions that cannot be released.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

- [ ] ~~Database migrations are either absent or optimized for zero downtime~~
- [ ] ~~All new fields/inputs/mutations have proper labels added (`ADDED_IN_X`, `PREVIEW_FEATURE`, etc.)~~
- [ ] ~~All migrations have proper dependencies~~
- [ ] ~~All indexes are added concurrently in migrations~~
- [ ] ~~All RunSql and RunPython migrations have revert option defined~~

### DIFF
--- a/saleor/payment/tasks.py
+++ b/saleor/payment/tasks.py
@@ -3,11 +3,12 @@ import logging
 import uuid
 
 from django.conf import settings
-from django.db.models import OuterRef, Q, Subquery
+from django.db import transaction
+from django.db.models import Exists, OuterRef, Q
 
 from ..celeryconf import app
 from ..channel.models import Channel
-from ..checkout import CheckoutAuthorizeStatus, CheckoutChargeStatus
+from ..checkout import CheckoutAuthorizeStatus
 from ..checkout.models import Checkout
 from ..core.db.connection import allow_writer
 from ..payment.models import TransactionEvent, TransactionItem
@@ -18,7 +19,7 @@ from .gateway import request_cancelation_action, request_refund_action
 logger = logging.getLogger(__name__)
 
 
-def checkouts_with_funds_to_release():
+def transactions_to_release_funds():
     """Fetch checkouts that are ready release the funds.
 
     Fetch checkouts with the funds where the last modification was more than defined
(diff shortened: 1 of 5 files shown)


kadewu (dismissed):

saleor/payment/tasks.py
```diff
@@ -18,7 +19,7 @@
 logger = logging.getLogger(__name__)
 
 
-def checkouts_with_funds_to_release():
+def transactions_to_release_funds():
     """Fetch checkouts that are ready release the funds.
```
korycins, line 23: Would be also worth to update the description of the function. As after the changes it doesn't return checkouts but transactions

saleor/payment/tasks.py
```diff
@@ -30,131 +31,143 @@ def checkouts_with_funds_to_release():
...
+        transactions.values_list("pk", "checkout_id")[:TRANSACTION_BATCH_SIZE]
+    )
+    transaction_pks = [pk for pk, _checkout_id in transactions_data]
+    checkout_ids = [checkout_id for _, checkout_id in transactions_data]
+
+    checkouts_data = Checkout.objects.filter(pk__in=checkout_ids).values_list(
+        "pk", "channel_id"
     )
-    checkout_pks = [pk for pk, _ in checkouts_data]
     checkout_channel_ids = [channel_id for _, channel_id in checkouts_data]
```
korycins, line 76: It is only a detail, as we don't have a lot of `checkouts_data` fetched at once, but to keep it cleaner, I would suggest adding `set()` here. No need to add in `filter()` duplicated values.

saleor/payment/tasks.py
```diff
@@ -30,131 +31,143 @@ def checkouts_with_funds_to_release():
...
-            TransactionEvent.objects.bulk_create(
-                [event for _tr, event in transactions_with_cancel_request_events]
-                + [event for _tr, event in transactions_with_charge_request_events]
-            )
+            with transaction.atomic():
+                TransactionEvent.objects.bulk_create(
+                    [event for _tr, event in transactions_with_cancel_request_events]
+                    + [event for _tr, event in transactions_with_charge_request_events]
+                )
+                # Mark transactions as not refundable to avoid multiple refund requests
```
korycins, line 131: Just to be more clear, I would add the following info: 
```suggestion
                # Mark transactions as not refundable to avoid multiple automatic refund requests
```

saleor/payment/tasks.py
```diff
@@ -30,131 +31,143 @@ def checkouts_with_funds_to_release():
...
                     )
                 except PaymentError as e:
                     logger.warning(
                         "Unable to refund transaction %s. %s",
-                        transaction.token,
+                        transaction_item.token,
                         str(e),
                     )
+        else:
+            logger.warning("No transactions to release funds.")
```
korycins, line 173: Maybe worth to be more explicit here 🤔  For example, In case of having 30 transactions, where `transactions_with_charge_request_events` is empty and we will have `transactions_with_cancel_request_events` single elemenent, we won't see error msg. But for some reasons, we had 30 fetched transactions, and in "standard" scenario all of them should have a charge or cancel event.

fowczarek (approved):

Issue #16950, opened 2024-10-30
Add pricePrior field to ProductVariantChannelListing


delemeator:
## Problem

EU Omnibus directive (https://eur-lex.europa.eu/eli/dir/2019/2161/oj) creates an obligation to present "prior price" in case of products with discount. This price should cover period of 30 days and include all price changes (prior promotions or changes of regular price). Details of the implementation vary between countries. Currently there is no good way to provide this information to storefront from Saleor and storing it in CMS could be messy (as this information is required in product pages and cart / checkout)

---

## 1. General Assumptions

Create pricePrior of type Money field in ProductVariantChannelListing. This would be set via ProductVariantChannelListingAddInput so that the logic on how pricePrior is calculated is kept outside of Saleor. This field would be then populated to VariantPricingInfo and CheckoutLine (as priorTotalPrice and priorUnitPrice of type Money)

### Acceptance Criteria

- [ ] It is possible to modify priorPrice for a productVariant and channel combination
- [ ] Prior price is available in queries related to variant pricing and checkout

### Mutations

- Save priorPrice using productVariantChannelListingUpdate mutation

### Types

- Add field priorPrice (PositiveDecimal) to ProductVariantChannelListingAddInput 
- Add field priorPrice (TaxedMoney) to VariantPricingInfo
- Add field priorUnitPrice and priorTotalPrice (Money) to CheckoutLine

## 3. Database changes

Add prior_price (DecimalField) and prior_price_amount (MoneyField) to ProductVariantChannelListing

maarcingebala:
Closing as done 🎉

PR #17202, opened 2025-01-02
Implement price prior to promotion (for EU Omnibus Directive)

delemeator:
I want to merge this change because it implements priorPrice field as discussed in https://github.com/saleor/saleor/issues/16950

Changes:
- Added db columns related to prior price to `ProductVariantChannelListing` and `CheckoutLine`
- Added input fields to update prior price to mutations `productVariantChannelListingUpdate` and `productVariantBulkUpdate`
- Added fields to display prior price in `VariantPricingInfo`, `ProductPricingInfo` and `CheckoutLine`

Notes: 
- Logic for updating prior price is pretty much the same as for cost price. In some cases I have slightly refactored the code to avoid excessive copy/pasting 
- I think it would make sense to allow in input of productVariantChannelListingUpdate to skip `price` field and update only `costPrice` / `preorderThreshold` / `priorPrice`, but currently it is not possible as it uses type `ProductVariantChannelListingAddInput` where `price` is `PositiveDecimal!`. I think it would make sense to create input type `ProductVariantChannelListingUpdateInput` with `price` as `PositiveDecimal`, but I didn't do it as it would be a breaking change

# Impact

- [x] New migrations

# Docs

- [x] Link to documentation: https://github.com/saleor/saleor-docs/pull/1432

# Pull Request Checklist

- [x] All indexes are added concurrently in migrations (not applicable)
- [x] All RunSql and RunPython migrations have revert option defined (not applicable)


delemeator:
Hey - I can see that 2 tests are failing, but they don't seem to be related to changes from this PR. It's more about IP security features / infrastructure. Is this some broader problem?

korycins:
Hey @delemeator  can you rebase your branch with the newest `main`? The failing tests were fixed in this PR: https://github.com/saleor/saleor/pull/17262

delemeator:
@korycins - thanks - should be ok now

saleor/product/migrations/0197_productvariantchannellisting_prior_price_amount.py
IKarbowiak, line 1: This mutation also need to be ported to the previous version 3.20 to ensure zero downtime migration. You can find details about his approach [here](https://docs.saleor.io/developer/community/zero-downtime-migrations).

We need to ensure that the prior_price_amount field is already present in the database when users migrate from version 3.20 to 3.21

saleor/checkout/models.py
```diff
@@ -313,6 +313,17 @@ class CheckoutLine(ModelWithMetadata):
...
+    prior_unit_price_amount = models.DecimalField(
+        max_digits=settings.DEFAULT_MAX_DIGITS,
+        decimal_places=settings.DEFAULT_DECIMAL_PLACES,
+        blank=True,
+        null=True,
+    )
+
+    prior_unit_price = MoneyField(
+        amount_field="prior_unit_price_amount", currency_field="currency"
+    )
```
IKarbowiak, line 325: I’m wondering if we need this value on CheckoutLine at all. What is the reason behind it?
Isn't the info about this price on variant listing enough?

saleor/product/models.py
```diff
@@ -526,6 +535,16 @@ class ProductVariantChannelListing(models.Model):
...
 
+    prior_price_amount = models.DecimalField(
+        max_digits=settings.DEFAULT_MAX_DIGITS,
+        decimal_places=settings.DEFAULT_DECIMAL_PLACES,
+        blank=True,
+        null=True,
+    )
+    prior_price = MoneyField(
+        amount_field="prior_price_amount", currency_field="currency"
+    )
```
IKarbowiak, line 546: If you agree for the another field name, the field names on DB should ba changed as well.

saleor/graphql/product/types/channels.py
```diff
@@ -350,6 +350,11 @@ class ProductVariantChannelListing(
     )
     price = graphene.Field(Money, description="The price of the variant.")
     cost_price = graphene.Field(Money, description="Cost price of the variant.")
+    prior_price = graphene.Field(
+        Money,
+        description="Prior price of the variant used for discount calculations."
+        + ADDED_IN_321,
+    )
```
IKarbowiak, line 357: I’m unsure about the field name—it suggests that the price is before any promotion, which isn’t accurate. As I understand it, the purpose of this field is to keep a historical record of the lowest price from the last 30 days, correct?
Perhaps reference_price would be a better name?

Additionally, the description should be updated accordingly, clarifying that it represents a kind of reference price.

saleor/checkout/models.py
delemeator: For compliance with Omnibus, all calculations showing "-X%" should use prior price as reference. So if you have product with price 100, prior price 90 and discounted price 80, you can show only -11% discount in checkout / cart. So this is why prior price is needed in checkout. 

When it comes to where we take prior price info from, I have assumed there is probably a performance reason why checkout stores undiscounted_unit_price_amount rather than taking it from listing, so I assumed adding prior_unit_price_amount is the right way to proceed. If this should be resolved differently, I would need a guidance how you would implement it

saleor/graphql/product/types/channels.py
delemeator: Sure - I can rename prior into reference, given that it is called "Prior price" in directives I thought this name would be more meaningful for users, but I am fine with name "reference" which would be more general

For sake of accuracy `keep a historical record of the lowest price from the last 30 days` is not 100% correct in case of EU. Omnibus directive states: `The prior price means the lowest price applied by the trader during a period of time not shorter than 30 days prior to the application of the price reduction.`. So this is not price from last 30 days, but price from last 30day prior to (current) promotion

saleor/product/models.py
delemeator: Sure, please let me know if you agree internally that reference price is a better name. Then I will do the change

delemeator:
@IKarbowiak @korycins Did you get a chance to look on my comments?

saleor/graphql/product/mutations/channels.py
```diff
@@ -386,6 +386,11 @@ class ProductVariantChannelListingAddInput(BaseInputObjectType):
         required=True, description="Price of the particular variant in channel."
     )
     cost_price = PositiveDecimal(description="Cost price of the variant in channel.")
+    prior_price = PositiveDecimal(
```
korycins, line 389: Right now, you're unable to remove the `prior_price`. As PositiveDecimal would not accept `null`.  Not sure if this is something that we should worry about. It works the same way as `cost_price`, but the difference here is that `cost_price` is not available for customer, and `prior_price` will be available in `variant.pricing` object 🤔 
cc: @patrys @maarcingebala  WDYT?

saleor/product/migrations/0197_productvariantchannellisting_prior_price_amount.py
korycins: Yes, the nullable field doesn't need additional migration :)

korycins:
> @IKarbowiak @korycins Did you get a chance to look on my comments?

Sorry, for the delay! I additionally pushed the discussion internally.

saleor/graphql/checkout/tests/mutations/test_checkout_lines_add.py
IKarbowiak, line 1: I think we should also check `checkoutLinesUpdate` to ensure that the prior prices will be set, in case there were 0 before update.

saleor/product/tests/test_product_availability.py
```diff
@@ -431,3 +432,43 @@ def test_filter_not_published_product_without_assigned_channel(
...
+    )
+
+    # then
+    assert availability.price_range_prior is not None
+    assert (
+        availability.price_range_prior.start.gross.amount
+        == variant_channel_listing.prior_price_amount
+    )
+
+    if prior_price > discounted_price:
```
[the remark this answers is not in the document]
delemeator: Sure - I have split it into two tests and also added prior_price = 0 to make sure nothing brakes in such case

saleor/product/utils/availability.py
```diff
@@ -174,8 +198,25 @@ def get_variant_availability(
...
+    prior_price_taxed = None
+    if prior_price:
+        prior_price_taxed = (
+            _calculate_product_price_with_taxes(
+                prior_price,
+                tax_rate,
+                tax_calculation_strategy,
+                prices_entered_with_tax,
+            )
+            if prior_price
```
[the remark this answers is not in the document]
delemeator: Done, I don't know how I didn't notice this

saleor/product/utils/availability.py
```diff
@@ -174,8 +198,25 @@ def get_variant_availability(
         tax_calculation_strategy,
         prices_entered_with_tax,
     )
+    prior_price = variant_channel_listing.prior_price
+    prior_price_taxed = None
+    if prior_price:
```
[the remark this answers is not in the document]
delemeator: Good point, I thought you can't pass zero via PositiveDecimal in gql, but anyhow it should work now with 0

saleor/graphql/checkout/tests/mutations/test_checkout_lines_add.py
delemeator: Good point, I assumed that given that `CheckoutLinesUpdate` inherits from `CheckoutLinesAdd` it will re-do all the calculations. Now I see I was wrong ;) 

I have added logic to update prior price in `fetch_checkout_data`

saleor/graphql/product/mutations/channels.py
delemeator: For my workflow, I would set prior price equal to price if I want to "deactivate" it, but I can see your point

korycins (approved):

IKarbowiak (approved):

korycins:
@delemeator Merged. Really **great** work! Thanks a lot! 🙏 <3

delemeator:
Thank you for reviewing and accepting! :)

PR #17253, opened 2025-01-13
Fix undiscounted price taxation inside an order calculations when the Avatax plugin is used

zedzior:
I want to merge this change, because it fixes missing tax calculation for undiscounted prices when Avatax plugin is used and an order origins from draft order.

Internal issue: https://linear.app/saleor/issue/SHOPX-1775

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/order/calculations.py
+++ b/saleor/order/calculations.py
@@ -21,7 +21,7 @@ from ..payment.model_helpers import get_subtotal
 from ..plugins import PLUGIN_IDENTIFIER_PREFIX
 from ..plugins.manager import PluginsManager
 from ..tax import TaxCalculationStrategy
-from ..tax.calculations import get_taxed_undiscounted_price
+from ..tax.calculations import calculate_flat_rate_tax, get_taxed_undiscounted_price
 from ..tax.calculations.order import update_order_prices_with_flat_rates
 from ..tax.utils import (
     get_charge_taxes_for_order,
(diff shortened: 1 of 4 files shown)


IKarbowiak (approved):

saleor/order/calculations.py
```diff
@@ -345,22 +345,21 @@ def _recalculate_with_plugins(
...
-    ):
-        get_taxed_undiscounted_price(
-            line_base_price,
-            line_price.undiscounted_price,
-            tax_rate,
-            prices_entered_with_tax,
+) -> TaxedMoney:
+    if tax_rate > 0 and price.undiscounted_price.net == price.undiscounted_price.gross:
+        return quantize_price(
+            calculate_flat_rate_tax(
```
maarcingebala, line 355: A few things are unclear to me:
- before, `get_taxed_undiscounted_price` was missing the return statement which was obviously wrong
- now we're replacing `get_taxed_undiscounted_price` with `calculate_flat_rate_tax` which was done anyway in `get_taxed_undiscounted_price`, why are we doing this? Is there something wrong with the implementation of `get_taxed_undiscounted_price` which looks like should be used here?

zedzior: Just adding the missing return statement will not solve the issue because inside `get_taxed_undiscounted_price` function, this comparison: `if undiscounted_base_price == price_to_compare:` is always `true`.
Since the `get_taxed_undiscounted_price` function is used in checkout flow (where everything works correctly) I didn't want to refactor the function.
Does it make sense?

maarcingebala: Is it a different use case in this place vs. other usages in checkout where this comparison makes sense? I didn't check but I'm wondering if we're sure about the use-cases.

zedzior: OK, now I see we can reuse the `get_taxed_undiscounted_price` function, but we need to pass  `price=line_price.price_with_discounts` instead of `price=line_price.undiscounted_price`. Let me rewrite it

IKarbowiak (approved):

maarcingebala (approved):

PR #17268, opened 2025-01-16
Prevent tax sync webhooks in checkout bulk queries

korycins:
When calling resolvers like `me{checkouts}` , `checkouts` , `checkoutLines`  there is a possibility that some of the objects would trigger the sync webhooks. In case of having dozens of objects requested in the query, this could cause a massive slow down in the response, as we need to wait for sync webhooks. As a solution, we want to block the possibility of calling sync webhooks when requesting data via bulk queries.

We want to also cover cases like:
```graphql
query {
  checkout(id:"Q2hlY2tvdXQ6ZGRkNDQ3ZGUtZmQyNy00MDdkLTk2YTYtZGNmZTgwMDk3ODhi"){
    id
    totalPrice{
      gross{
        amount
      }
    }
  }
  checkouts(first:100, sortBy:{direction:ASC, field:CREATION_DATE}){
    edges{
      node{
        id
        totalPrice{
          tax{
            amount
          }
        }
      }
    }
  }
}
```

(where `checkout` should be able to trigger the sync webhooks, and for `checkouts` webhooks should be skipped).

Similar case is for different query indentations:
```graphql
query {
  checkout(id:"Q2hlY2tvdXQ6ZGRkNDQ3ZGUtZmQyNy00MDdkLTk2YTYtZGNmZTgwMDk3ODhi"){
    id
    totalPrice{
      gross{
        amount
      }
    }
  }
  me{
    checkouts(first:10){
      edges{
        node{
          id
        }
      }
    }
  }
}

```

The webhooks that will be skipped for bulk queries by this PR:
~- `SHIPPING_LIST_METHODS_FOR_CHECKOUT`~ (Will be handled in separate PR)
~- `CHECKOUT_FILTER_SHIPPING_METHODS`~ (Will be handled in separate PR)
- `CHECKOUT_CALCULATE_TAXES`

When calling `me{checkouts}` , `checkouts` , `checkoutLines` 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation: https://github.com/saleor/saleor-docs/pull/1459

# Pull Request Checklist


korycins:
> IMO you should also prevent generating pre-generated payload when sync webhooks are blocked.

Valid point, I will create a sub-task to handle this in separate PR.

IKarbowiak (approved):

fowczarek (approved):

PR #17273, opened 2025-01-18
Test/saleor docker setup

Rafi1115:


maarcingebala:
Thanks for contribution, but we're not planning to add docker-compose to core, as it is currently handled by saleor-platform repo.

PR #17276, opened 2025-01-20
Fix incorrect order for reordering collection products when sort_order is null

korycins:
I want to merge this change because it fixes the issue when calling `collectionReorderProducts` for products with `sort_order == Null`. By default `collection.products` are sorted by `sort_order, product_id`, but in case of reordering, the `id` of the many-to-many object was used (`CollectionProduct`). This caused that the first call to reorder objects, caused results in different order than expected.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/core/utils/reordering.py
+++ b/saleor/graphql/core/utils/reordering.py
@@ -23,10 +23,11 @@ class FinalSortOrder:
 
 
 class Reordering:
-    def __init__(self, qs: QuerySet, operations: dict[int, int], field: str):
+    def __init__(
+        self, qs: QuerySet, operations: dict[int, int], second_sort_field: str
+    ):
         self.qs = qs
         self.operations = operations
-        self.field = field
 
         # Will contain the original data, before sorting
         # This will be useful to look for the sort orders that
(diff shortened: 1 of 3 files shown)


kadewu (dismissed):

saleor/graphql/core/utils/reordering.py
```diff
@@ -36,13 +37,14 @@ def __init__(self, qs: QuerySet, operations: dict[int, int], field: str):
...
         self.ordered_pks: list[int] = []
+        self.second_sort_field = second_sort_field
 
     @cached_property
     def ordered_node_map(self):
         ordering_map = OrderedDict(
             self.qs.select_for_update()
             .values_list("pk", "sort_order")
-            .order_by(F("sort_order").asc(nulls_last=True), "id")
+            .order_by(F("sort_order").asc(nulls_last=True), self.second_sort_field)
```
fowczarek, line 47:
```suggestion
            .order_by(F("sort_order").asc(nulls_last=True), self.second_sort_field, "id")
```
Maybe? The last ordering field should be unique to avoid breaking the pagination.

korycins: @fowczarek Thanks, added :)

IKarbowiak (approved):

kadewu (approved):

fowczarek (approved):

PR #17300, opened 2025-01-29
Adjust order fulfillment

IKarbowiak:
Solution:
- Call gift card creation and order status updates to the transaction where fulfillment in the same transaction where fulfillment lines are created - currently, there is a risk that fulfillment might be created, but if an error occurs while sending webhooks, the fulfillment event may not be recorded, and the order status might remain unchanged.
- Ensure that `update_order_status` is performed in transaction to ensure that the status is not overridden by another thread.
- Extend parameters of `FULFILLMENT_FULFILLED_ITEMS` order event with `auto` that is saying if the fulfillments was done automatically or not

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/graphql/checkout/tests/deprecated/test_checkout_complete.py
```diff
@@ -115,6 +116,7 @@ def test_checkout_complete(
     variables = {"checkoutId": checkout_id, "redirectUrl": redirect_url}
     response = user_api_client.post_graphql(MUTATION_CHECKOUT_COMPLETE, variables)
 
+    flush_post_commit_hooks()
```
korycins, line 119: `with django_capture_on_commit_callbacks(execute=True):` would be better, as `flush_post_commit_hooks()` doesn't handle well a rollback. (IMHO this helper should be fully drop from our core, as AFAIK it does the same what `django_capture_on_commit_callbacks`)

saleor/graphql/order/mutations/order_confirm.py
```diff
@@ -72,9 +72,9 @@ def perform_mutation(cls, root, info: ResolveInfo, /, **data):
         user = cast(User, user)
         order = cls.get_instance(info, **data)
         cls.check_channel_permissions(info, [order.channel_id])
-        order.status = OrderStatus.UNFULFILLED
+        order = update_order_status(order)
         update_order_display_gross_prices(order)
```
korycins, line 76: Hhm, it sounds strange, that we can change the behavior of the `order.display_gross_prices` when calling orderConfirm 🤔 
`update_order_display_gross_prices` is called when we convert checkout to order, or when calling draft create/draft complete. The unconfirmed order already has this value, so not sure why we want to change it here.

I know that this is not the scope of this PR, but I pointed it out, as in case of removing this, we will be able to also remove additional `.save()` call.

saleor/order/utils.py
```diff
@@ -168,35 +169,41 @@ def _calculate_quantity_including_returns(order):
 
 def update_order_status(order: Order):
```
korycins, line 170:
```suggestion
def update_order_status(order: Order) -> Order:
```

saleor/graphql/order/mutations/order_confirm.py
IKarbowiak: It's in case order billing or shipping address has changed. So seems like the `update_order_display_gross_prices` should be moved to `OrderUpdate` mutation instead of updating it here. I will create a ticket for that to handle it separately.

fowczarek (approved):

korycins (approved):

PR #17301, opened 2025-01-29
Adjust order fulfillment

IKarbowiak:
Solution:
- Call gift card creation and order status updates to the transaction where fulfillment in the same transaction where fulfillment lines are created - currently, there is a risk that fulfillment might be created, but if an error occurs while sending webhooks, the fulfillment event may not be recorded, and the order status might remain unchanged.
- Ensure that `update_order_status` is performed in transaction to ensure that the status is not overridden by another thread.
- Extend parameters of `FULFILLMENT_FULFILLED_ITEMS` order event with `auto` that is saying if the fulfillments was done automatically or not

Port of https://github.com/saleor/saleor/pull/17300

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/checkout/tests/test_checkout_complete.py
+++ b/saleor/checkout/tests/test_checkout_complete.py
@@ -1258,6 +1258,7 @@ def test_complete_checkout_0_total_with_transaction_for_mark_as_paid(
         )
 
     # then
+    order.refresh_from_db()
     assert order
     assert order.authorize_status == OrderAuthorizeStatus.FULL
     assert order.charge_status == OrderChargeStatus.FULL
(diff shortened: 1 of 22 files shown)


fowczarek (approved):

korycins (commented):
Do we need a separate event for that? I mean, the most important information here is that the lines were fulfilled. If we want to track if the fulfillment was created by Saleor, or by staff user/app, then I would consider additional info in the event that would determine if the event was created automatically or not.

As with proposed changes we have:
Staff/App: FULFILLMENT_FULFILLED_ITEMS
Auto: FULFILLMENT_AUTOMATIC_FULFILLED_ITEMS

With parameter, we would have:
Staff/App: FULFILLMENT_FULFILLED_ITEMS - {"auto":false}
Auto: FULFILLMENT_FULFILLED_ITEMS - {"auto": true}

Same thing we would propabably need to add in case of `if order_info.channel.automatically_fulfill_non_shippable_gift_card:`

korycins (dismissed):

fowczarek (dismissed):

korycins (approved):

fowczarek (approved):

PR #17309, opened 2025-01-30
Remove reference cycle in Saleor context

fowczarek:
- allows memory to be freed immediately without needing a deep garbage collection cycle.
- Drop usage of SaleorContext in `transaction.on_commit`. 

Garbage before:
garbage_before

Garbage after
garbage_afterr

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/context.py
+++ b/saleor/graphql/context.py
@@ -28,6 +28,7 @@ def get_context_value(request: HttpRequest) -> SaleorContext:
 
 def clear_context(context: SaleorContext):
     context.dataloaders.clear()
+    del context.user
 
 
 class RequestWithUser(HttpRequest):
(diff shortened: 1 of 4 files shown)


saleor/graphql/context.py
```diff
@@ -28,6 +28,7 @@ def get_context_value(request: HttpRequest) -> SaleorContext:
 
 def clear_context(context: SaleorContext):
     context.dataloaders.clear()
+    del context.user
```
korycins, line 31: As you're deleting the assigned user here, don't you also need to `del` the potential app?
Also, what in a situation where `context.user` is None?

fowczarek: We don't need to delete the `app`. The user is wrapped in `SimpleLazyObject`. The reference cycle is generated via `SimpleLazyObject`. When a user is `None` it's still wrapped with `SimpleLaysObject`.

korycins: Ach I see. Based on the typing in SaleorContext, I assumed that this is "plain" user

korycins (approved):

korycins (approved):

IKarbowiak (approved):

PR #17311, opened 2025-01-31
[POC] Telemetry instrumentation abstract layer with OpenTelemetry adapter

przlada:
This PR is a proof of concept for a telemetry instrumentation interface that provides an API for traces and metrics collection. By default it comes with an OpenTelemetry implementation.
- Signals can be gathered by `tracer` and `meter` in two distinct scopes:
  - `core`: For internal system operations and core functionality
  - `service`: For business logic and service-level operations
- Alternative implementations can be used by changing `TELEMETRY_TRACER_CLASS` and `TELEMETRY_METER_CLASS` settings
- Functionality for adding "global" attributes that are automatically set on all spans and metric measurements
- Telemetry components are lazily initialized to ensure fork safety in multi-process environments

### Traces:
```python
from saleor.core.telemetry import SpanAttributes, tracer

# Create and activate span in the service scope
with tracer.start_as_current_span("graphql_query", service_scope=True) as span:
    span.set_attribute(SpanAttributes.HTTP_URL, url)
    # ... your code here
```

### Metrics:
```python
from saleor.core.telemetry import MetricType, Unit, meter

# Define metrics
meter.create_metric(
    "saleor.graphql_queries",
    service_scope=True,
    type=MetricType.COUNTER,
    unit=Unit.REQUEST,
    description="Number of GraphQL queries.",
)

meter.create_metric(
    "saleor.graphql_query_duration",
    service_scope=True,
    type=MetricType.HISTOGRAM,
    unit=Unit.MILLISECOND,
    description="Duration of GraphQL queries.",
)

# Report duration measurement using context manager
with meter.record_duration("saleor.graphql_query_duration"):
    # Create simple measurement with provided value and attributes
    meter.record("saleor.graphql_queries", 1, attributes=attributes)
```

To start Saleor with OTEL enabled, you can use the standard method, for example:
```
opentelemetry-instrument --traces_exporter console --metrics_exporter console --service_name saleor uvicorn saleor.asgi:application --loop=uvloop --host=0.0.0.0 --port=8000
```

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/asgi/__init__.py
```diff
@@ -32,7 +33,8 @@ def preload_app() -> None:
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "saleor.settings")
 
 application = get_asgi_application()
-application = health_check(application, "/health/")  # type: ignore[arg-type] # Django's ASGI app is less strict than the spec # noqa: E501
+application = otel_context(application)  # type: ignore[arg-type] # Django's ASGI app is less strict than the spec # noqa: E501
```
patrys, line 36: Shouldn't this be the outermost middleware? What if other middleware uses telemetry?

saleor/asgi/otel_context.py
```diff
@@ -0,0 +1,31 @@
...
+
+def get_hostname(scope: Scope) -> str:
+    hostname: str = ""
+    if scope["type"] == "http":
+        for header, value in scope["headers"]:
+            if header == b"host":
+                hostname = value.decode("latin1")
+                break
+    hostname = hostname.split(":")[0]  # remove port
+    hostname = hostname.removeprefix("www.")
```
patrys, line 20: Why alter the hostname?

saleor/core/otel/metrics.py
```diff
@@ -0,0 +1,52 @@
...
+
+from opentelemetry import metrics
+from opentelemetry.metrics import Meter
+from opentelemetry.util.types import Attributes
+
+from ... import __version__
+from .context import enrich_with_trace_attributes
+
+
+class _ContextAwareMeter(Meter):
```
patrys, line 11: This class inherits from `Meter` but does not use that. It's a facade, not a subclass.

saleor/graphql/views.py
```diff
@@ -204,7 +213,9 @@ def handle_query(self, request: HttpRequest) -> JsonResponse:
                     span.set_attribute(f"ip_{additional_ip_header}", request_ips[:100])
 
             response = self._handle_query(request)
-            span.set_attribute(SpanAttributes.HTTP_STATUS_CODE, response.status_code)
+            tracer.get_current_span().set_attribute(
```
maarcingebala, line 216: Isn't `tracer.get_current_span()` the same as `span` here?

przlada: You are right. It is just an example of using `tracer.get_current_span()` which I forgot to remove

saleor/graphql/views.py
```diff
@@ -273,7 +288,11 @@ def parse_query(
             return None, ExecutionResult(errors=[e], invalid=True)
 
     def execute_graphql_request(self, request: HttpRequest, data: dict):
-        with tracer.start_as_current_span("graphql_query") as span:
+        with (
+            tracer.start_as_current_span("graphql_query", service_scope=True) as span,
```
fowczarek, line 292: Are we sure all data in `graphql_query` should be exposed to `SERVICE` tracker?

saleor/graphql/views.py
```diff
@@ -162,7 +175,9 @@ def _handle_query(self, request: HttpRequest) -> JsonResponse:
         return JsonResponse(data=result, status=status_code, safe=False)
 
     def handle_query(self, request: HttpRequest) -> JsonResponse:
-        with tracer.start_as_current_span("http", kind=trace.SpanKind.SERVER) as span:
+        with tracer.start_as_current_span(
+            "http", service_scope=True, kind=SpanKind.SERVER
```
fowczarek, line 179: Are we sure all data in `http` should be exposed to `SERVICE` tracker?

saleor/graphql/views.py
```diff
@@ -39,11 +37,26 @@
...
+    description="Number of GraphQL queries.",
+)
+meter.create_metric(
+    "saleor.graphql_query_duration",
+    service_scope=True,
+    type=MetricType.HISTOGRAM,
+    unit=Unit.MILLISECOND,
+    description="Duration of GraphQL queries.",
+)
+
```
fowczarek, line 54: Maybe we should define metrics in separate files? e.g `saleor/graphql/metrics.py` ?

saleor/core/tracing.py
```diff
@@ -35,7 +34,7 @@ def webhooks_otel_trace(
     :param payload_size: size of the payload in bytes
     """
     with tracer.start_as_current_span(
-        f"webhooks.{span_name}", kind=trace.SpanKind.CLIENT
+        f"webhooks.{span_name}", service_scope=True, kind=SpanKind.CLIENT
```
fowczarek, line 37: Are we sure all data in `webhooks.*`  should be exposed to `SERVICE` tracker?

saleor/core/telemetry/metric.py
```diff
@@ -0,0 +1,205 @@
...
+            type: Type of the metric
+            unit: Unit of the metric
+            service_scope: If True, use service scope instead of default core
+            description: Optional description of the metric
+
+        """
+        meter = self._service_tracer if service_scope else self._core_tracer
+        with self._lock:
+            if name in self._instruments:
+                raise DuplicateMetricError(name)
```
fowczarek, line 106: Should we add a test for this?

saleor/core/telemetry/metric.py
```diff
@@ -0,0 +1,205 @@
...
+        Args:
+            metric_name: Name of the metric to record a measurement
+            amount: Value to record
+            unit: Optional unit of the measurement (converted if different from metric unit)
+            attributes: Attributes to record with the measurement
+
+        """
+        attributes = enrich_with_global_attributes(attributes)
+        if metric_name not in self._instruments:
+            raise RuntimeError(f"Metric {metric_name} must be created first")
```
fowczarek, line 129: We should also test this case.

fowczarek (commented):
Sending unnecessary tags in spans to the `SERVICE` tracer could increase costs. 
e.g., In the `graphql` span, we have `pod_name` in my opinion, this information is useless.

saleor/graphql/views.py
```diff
@@ -39,11 +37,26 @@
 
 INT_ERROR_MSG = "Int cannot represent non 32-bit signed integer value"
 
+meter.create_metric(
```
maarcingebala, line 40: This is good for now, but while reading I had an idea that I wanted to share about how to organize the metrics for easier usage in code.

- Have the `metrics.py` files in specific modules where `meter.create_metric` calls would be defined. In this case, we would have `saleor/graphql/metrics.py`.
- Next to metrics definitions have functions to actually increment the metric e.g. `record_graphql_query_duration` - this way in the actual module where you increment the metric you will only import the function, without the details such as `MetricType` or `Unit`. Also, we wouldn't use strings such as `"saleor.graphql_query_duration"` (but this could be replaced with a constant as well).

Then, adding new metrics will always include changes in the `metrics.py` files, importing the increment functions and calling it in the right place.

przlada: As discussed, I moved this example metrics creation code to a separate module `saleor/graphql/metrics.py` and created some helper functions that can be used to operate on them

saleor/core/telemetry/metric.py
przlada: Yes, great point! We plan to add tests after confirming that the proposed API is acceptable, or in a separate PR to the feature branch

saleor/graphql/metrics.py
```diff
@@ -0,0 +1,30 @@
...
+    "saleor.graphql_query_duration",
+    service_scope=True,
+    type=MetricType.HISTOGRAM,
+    unit=Unit.MILLISECOND,
+    description="Duration of GraphQL queries.",
+)
+
+
+# Helper functions
+def incr_graphql_queries(amount: int = 1) -> None:
```
maarcingebala, line 23: I was also thinking about some naming convention for the functions that record metrics. Since meters use "record", maybe we could also name all these helpers using the `record_` prefix:
- `record_graphql_queries_count`
- `record_graphql_query_duration`

saleor/graphql/views.py
przlada: Yes, we agreed that all the attributes are suitable for the service scope

saleor/graphql/views.py
przlada: Yes, we agreed that all the attributes are suitable for the `service` scope

saleor/core/tracing.py
przlada: Yes, we agreed that all the attributes are suitable for the `service` scope

saleor/graphql/metrics.py
przlada: Fixed

maarcingebala (approved):

PR #17160, opened 2024-12-10
Use denormalized discount prices during order update

zedzior:
This is feature branch for using denormalised discount prices when updating an order.

To do list:
- [ ] Use denormalised prices during order update for catalogue discounts (https://linear.app/saleor/issue/SHOPX-1672)
- [ ] Use denormalised prices during order update for vouchers (https://linear.app/saleor/issue/SHOPX-1673)
- [ ] use denormalised prices during order update for order discounts (https://linear.app/saleor/issue/SHOPX-1674)

Internal issue: https://linear.app/saleor/issue/SHOPX-1586/use-denormalised-prices-during-order-update

# Impact

- [ ] New migrations
- [ ] New/Updated API fields or mutations
- [ ] Deprecated API fields or mutations
- [ ] Removed API types, fields, or mutations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

- [ ] Privileged queries and mutations are either absent or guarded by proper permission checks
- [ ] Database queries are optimized and the number of queries is constant
- [ ] Database migrations are either absent or optimized for zero downtime
- [ ] The changes are covered by test cases
- [ ] All new fields/inputs/mutations have proper labels added (`ADDED_IN_X`, `PREVIEW_FEATURE`, etc.)
- [ ] All migrations have proper dependencies
- [ ] All indexes are added concurrently in migrations
- [ ] All RunSql and RunPython migrations have revert option defined

PR #17326, opened 2025-02-06
Add a logic to force the order price refresh with the latest channel listing prices

zedzior:
Issue: https://linear.app/saleor/issue/SCL-704/add-a-logic-to-force-the-order-price-refresh-with-the-latest-channel

This PR adds a logic, which allows to update the order prices (all the lines or particular lines), based on the latest conditions including update of a discount objects. By latest conditions, I mean latest channel listing prices, catalog promotion and voucher details.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/discount/tests/test_utils/test_refresh_order_prices_and_discounts.py
+++ b/saleor/discount/tests/test_utils/test_refresh_order_prices_and_discounts.py
@@ -100,7 +100,7 @@ def test_refresh_order_base_prices_single_line(order_with_lines):
 
 
 def test_refresh_order_base_prices_catalogue_discount(
-    order_with_lines_and_catalogue_promotion, plugins_manager
+    order_with_lines_and_catalogue_promotion,
 ):
     # given
     order = order_with_lines_and_catalogue_promotion
@@ -217,7 +217,7 @@ def test_refresh_order_base_prices_catalogue_discount(
 
 
 def test_refresh_order_base_prices_catalogue_discount_single_line(
-    order_with_lines_and_catalogue_promotion, plugins_manager
+    order_with_lines_and_catalogue_promotion,
 ):
     # given
     order = order_with_lines_and_catalogue_promotion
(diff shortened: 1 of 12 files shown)


saleor/discount/utils/order.py
```diff
@@ -277,3 +285,213 @@ def create_order_line_discount_objects_for_catalogue_promotions(
...
+    for line_info in lines_info:
+        line = line_info.line
+        channel_listing = line_info.channel_listing
+        # TODO zedzior: should we do anything if the listing is not available anymore
+        if channel_listing and channel_listing.price_amount:
+            line.undiscounted_base_unit_price_amount = channel_listing.price_amount
+            line.base_unit_price_amount = channel_listing.price_amount
+
+
+def refresh_order_line_discount_objects_for_catalogue_promotions(
```
zedzior, line 366: Functions below are restored from main

saleor/order/fetch.py
```diff
@@ -145,22 +182,63 @@ def fetch_draft_order_lines_info(
...
+
+
+def _get_variant_listing(
+    variant: ProductVariant | None, channel_id: int
+) -> ProductVariantChannelListing | None:
+    if not variant:
+        return None
+    for channel_listing in variant.channel_listings.all():
+        if channel_listing.channel_id == channel_id:
+            return channel_listing
```
zedzior, line 235: This is to avoid additional db call (no use of `filter`), as channel_listings are prefetched

saleor/discount/tests/test_utils/test_refresh_order_prices_and_discounts.py
```diff
@@ -0,0 +1,1079 @@
...
+from ....product.utils.variant_prices import update_discounted_prices_for_promotion
+from ....product.utils.variants import fetch_variants_for_promotion_rules
+from ... import DiscountType, DiscountValueType, RewardValueType, VoucherType
+from ...models import OrderLineDiscount, Promotion, PromotionRule
+from ...utils.order import refresh_order_base_prices_and_discounts
+from ...utils.voucher import create_or_update_voucher_discount_objects_for_order
+
+
+@pytest.fixture
+def order_with_lines(order_with_lines):
```
IKarbowiak, line 19: You shouldn't override existing fixture. Create a new one instead. 
```suggestion
def draft_order_with_lines(order_with_lines):
````

File: saleor/order/fetch.py [hunk 4]
```diff
@@ -107,13 +116,28 @@ def get_manual_line_discount(
...
 def fetch_draft_order_lines_info(
-    order: "Order", lines: Iterable["OrderLine"] | None = None
+    order: "Order",
+    lines: Iterable["OrderLine"] | None = None,
+    extend: bool = False,
 ) -> list[EditableOrderLineInfo]:
+    """Fetch the necessary order lines info in order to recalculate its prices.
+
+    `extend` argument determines if the function should additionally retrieve the latest
+    variant channel listing prices
```
IKarbowiak, line 126: Please use some more meaningful name, like `prefetch_channel_prices`, or something different that explain what is this flag used for

saleor/order/utils.py
```diff
@@ -819,6 +819,7 @@ def create_order_discount_for_order(
     current_total: TaxedMoney = order.undiscounted_total
     currency = order.currency
 
+    # TODO zedzior: why gross price is taken?
```
korycins, line 822: I would say it is a bug (but you can also double check this with someone). I mean, we should use .net when channel is set to .net, and use .gross when channel is seto to .gross.

saleor/discount/tests/test_utils/test_refresh_order_prices_and_discounts.py
zedzior: Creating a new fixture will not give the same result. Overwriting the existing `order_with_lines` will also change the order status of all the fixtures which inherit from `order_with_lines`. This way we save multiple database calls. What do you think?

saleor/discount/utils/order.py
```diff
@@ -277,3 +285,213 @@ def create_order_line_discount_objects_for_catalogue_promotions(
                     new_line_discounts, ignore_conflicts=True
                 )
     return []
+
+
+def refresh_order_base_prices_and_discounts(
```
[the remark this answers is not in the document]
zedzior: > I assume that this function will be used in next PRs - as I don't see the usage in the code.

You are right

> So my assumption is that this function, should accept the all order's lines instead of IDs

IDs is enough. The thing is I will always need to fetch all the lines details in order to properly handle various types of discounts (ie. voucher for the cheapest product).

The next PR will give you more light on it

saleor/discount/utils/order.py
```diff
@@ -277,3 +285,213 @@ def create_order_line_discount_objects_for_catalogue_promotions(
...
+            "undiscounted_base_unit_price_amount",
+        ],
+    )
+
+
+def _set_channel_listing_prices(lines_info: list["EditableOrderLineInfo"]):
+    for line_info in lines_info:
+        line = line_info.line
+        channel_listing = line_info.channel_listing
+        # TODO zedzior: should we do anything if the listing is not available anymore
```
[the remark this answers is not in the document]
zedzior: I will leave it to be handled in separate task

saleor/discount/utils/order.py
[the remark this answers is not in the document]
korycins: I meant, that in case of having a flag on orderLine, you should refresh all expired lines based on this flag. If you provide single ID in the list, but orders will have 4 expired lines, then IMHO all 4 of them should be recalculated. This is why I assumed that list of IDs is obsolete.
But as you said, I will wait for next PR.

saleor/discount/tests/test_utils/test_refresh_order_prices_and_discounts.py
zedzior: To avoid rebases, I change it in next PR of the project: https://github.com/saleor/saleor/pull/17419/commits/16b6d3e78ccd1531f0d6aadc07a1c0080518e2be

PR #17329, opened 2025-02-06
Fix default value of settings.AWS_S3_URL_PROTOCOL

magul:
I want to merge this change because without proper default value media URLs are broken

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/core/utils/tests/test_url.py
```diff
@@ -51,3 +54,48 @@ def test_get_default_storage_root_url(site_settings):
...
+    when we want to test how settings will behave based on different environment variables.
+    """
+    importlib.reload(sys.modules["saleor.settings"])
+    importlib.reload(sys.modules["saleor.tests.settings"])
+    del settings.STORAGES
+    settings._setup()
+    storages._backends = None
+    storages._storages = {}
+    del storages.backends
+    default_storage._setup()
```
patrys, line 71: We shouldn't do that, it will affect other tests. If we want to test a code path, we should extract that piece of code to a function.

saleor/core/utils/tests/test_url.py
```diff
@@ -51,3 +54,48 @@ def test_get_default_storage_root_url(site_settings):
...
+
+
+def test_aws_s3_url_protocol_not_set(monkeypatch):
```
korycins, line 74: In case of having the custom settings.py as a default, where you don't use `os.environ` to set `AWS_S3_URL_PROTOCOL`, the proposed tests flow with `m.setenv()` will fail (even if everything works as expected). 


What about something like:
```python
from django.test import override_settings

@override_settings(
    AWS_MEDIA_BUCKET_NAME="media-test-bucket",
    AWS_MEDIA_CUSTOM_DOMAIN="media.localhost",
)
def test_aws_s3_url_protocol_not_set(monkeypatch):
    assert get_default_storage_root_url() == "https://media.localhost/"

@override_settings(
    AWS_MEDIA_BUCKET_NAME="media-test-bucket",
    AWS_MEDIA_CUSTOM_DOMAIN="media.localhost",
    AWS_S3_URL_PROTOCOL="custom:"
)
def test_aws_s3_url_protocol_custom(monkeypatch):
    assert get_default_storage_root_url() == "custom://media.localhost/"
```

It doesn't cover the os.env usage, but you have covered the case when  url_protocol is not provided - so use the default value (`https:`). Second covers the case when you provided the protocol, so you expect the `custom:`.

magul: > In case of having the custom settings.py as a default, where you don't use os.environ to set AWS_S3_URL_PROTOCOL, the proposed tests flow with m.setenv() will fail (even if everything works as expected).

That's true, but I assumed that default `settings.py` is a reasonable starting point, if you want to extend our settings (and I also assumed, that if you really want to tinker with settings, you should use env variables and not overwrite settings inside whatever you will pass as `DJANGO_SETTINGS_MODULE`).
This is kind of question, what we can expect to be a Saleor user public interface.

Also I see what you proposed there, but my intention was to test how logic inside `settings.py` will behave, not logic, that's based on `django.conf.settings`. Assuming, that we treat chaning `DJANGO_SETTINGS_MODULE` as proper publicly accessible interface to setup `django.conf.settings`, and we will `@override_settings` in tests, then we don't coverage whatever logic we have in our default `settings.py`.

I will remove these tests for now, but we should spend some time to figure out, how to potentially defend ourselves against issues with wrong defaults in `settings.py`.

saleor/core/utils/tests/test_url.py
magul: I agree it looks terrible, AFAIK whole 13k tests passes (I tested that locally), but I also agree, that there's no justifiable value provided be these tests to maintain this refresh logic (and ensure it's not going to affect other tests).

> If we want to test a code path, we should extract that piece of code to a function.

Here I would defer making that decision to Django 5, as it looks there is major rewrite in a way Django treats `STORAGES`.

korycins (approved):

IKarbowiak (approved):

PR #17346, opened 2025-02-12
Add saving strategies for `Checkout`

IKarbowiak:
- Add `save_shipping_address` to `Checkout` model
- Add `save_billing_address` to `Order` model
- Take into consideration saving settings during checkout completion - with payment, transaction, and order from checkout completion
- Ensure the shipping address for checkouts with click and collect is never saved 

💡 
PR to feature branch

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/checkout/complete_checkout.py
```diff
@@ -180,11 +180,16 @@ def _process_shipping_data_for_order(
...
-    if (
+    # allow saving shipping address in case the delivery method is not collection point
+    # or the delivery method is not provided, but the shipping address is set
+    save_shipping_address = checkout_info.checkout.save_shipping_address and (
         delivery_method_info.store_as_customer_address
-        and checkout_info.user
-        and shipping_address
-    ):
+        or not delivery_method_info.delivery_method
+    )
```
lkostrowski, line 188: suggestion: such complex conditions, especially near business logic, would be much more readable if simplified outside of the if statement

such as

```python
address_marked_to_be_saved_via_mutation = checkout_info.checkout.save_shipping_address
delivery_allows_storing_as_address = delivery_method_info.store_as_customer_address
delivery_method_doesnt_exist = delivery_method_info.delivery_method

save_shipping_address = address_marked_to_be_saved_via_mutation 
  and (delivery_allows_storing_as_address or delivery_method_doesnt_exist)

```

which is easier to understand how the logic behaves and easier to find a logical bug

saleor/order/models.py
```diff
@@ -153,6 +153,8 @@ class Order(ModelWithMetadata, ModelWithExternalReference):
         null=True,
         on_delete=models.SET_NULL,
     )
+    draft_save_billing_address = models.BooleanField(null=True, blank=True)
+    draft_save_shipping_address = models.BooleanField(null=True, blank=True)
```
lkostrowski, line 157: If this part is confusing, maybe it's worth to add comment explaining that this flag only should exist in draft order and in Order is null?

IKarbowiak: Comment added

saleor/checkout/complete_checkout.py
IKarbowiak: I applied suggestion and moved the logic to the separate method

saleor/checkout/complete_checkout.py
```diff
@@ -211,11 +219,38 @@ def _process_shipping_data_for_order(
...
+) -> bool:
+    """Determine whether the shipping address should be stored in the user's addresses.
+
+    The address should be stored if:
+    - the user is authenticated
+    - the shipping address is set
+    - the address is marked to be saved
+    - the delivery method allows storing the address as a user address
+    (the delivery method is not a collection point) or the delivery method is not set
+    """
```
lkostrowski, line 233: awesome 😍

lkostrowski (approved):

saleor/checkout/complete_checkout.py
```diff
@@ -180,15 +180,23 @@ def _process_shipping_data_for_order(
...
-    if (
-        delivery_method_info.store_as_customer_address
-        and checkout_info.user
-        and shipping_address
+    # allow saving shipping address in case the delivery method is not provided,
+    # but the shipping address is set
+    if not delivery_method_info.delivery_method:
+        shipping_address = checkout_info.shipping_address
+
+    if _should_store_shipping_address_in_user_addresses(
```
korycins, line 188: Not sure if this should be a part of this feature or separate issue, but shouldn't we move this part to on_commit action of checkout-complete? 
Right now, for plugins flow, the address can be created even if we raise an exception (plugin flow is split into two transaction blocks. This happens in first transaction block. So in case of having an exception in the second block, we will make a store-address action, but the order will not be created).
For transaction flow, it should work, as we have there single transaction block, but here I would also expect on_commit. I mean, converting checkout into order is our main action for checkout-complete. Saving address is an additional action that should not have a potential impact on rolling back the whole order creation process.

fowczarek (approved):

saleor/checkout/complete_checkout.py
IKarbowiak: I'll create a ticket for that.

PR #17350, opened 2025-02-12
Do not return external shipping methods with different currency

korycins:
I want to merge this change because it filters external shipping methods to return only those with the same currency as checkout's currency. 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation: https://github.com/saleor/saleor-docs/pull/1473

# Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,3 +63,4 @@ All notable, unreleased changes to this project will be documented in this file.
 - Added support for the AWS_S3_URL_PROTOCOL environment variable - #17305 by @p-febis
 - Fixed pycurl dependency and required system libraries to fix Celery worker issues when using SQS by @mariobrgomes
 - Added [`alg`](https://datatracker.ietf.org/doc/html/rfc7517#section-4.4) key to JWKS available at `/.well-known/jwks.json` - #17363 by @lkostrowski
+- Fix the issue of returning external shipping methods in a currency different from the checkout currency. - #17350 by @korycins
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,5 +63,4 @@ All notable, unreleased changes to this project will be documented in this file.
 - Added support for the AWS_S3_URL_PROTOCOL environment variable - #17305 by @p-febis
 - Fixed pycurl dependency and required system libraries to fix Celery worker issues when using SQS by @mariobrgomes
 - Added [`alg`](https://datatracker.ietf.org/doc/html/rfc7517#section-4.4) key to JWKS available at `/.well-known/jwks.json` - #17363 by @lkostrowski
-- Fix the issue of returning external shipping methods in a currency different from the checkout currency. - #17350 by @korycins
 - `checkout.shippingMethods` and `checkout.availableShippingMethods` will no longer return external shipping methods if their currency differs from the checkout's currency - #17350 by @korycins
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,3 +64,4 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fixed pycurl dependency and required system libraries to fix Celery worker issues when using SQS by @mariobrgomes
 - Added [`alg`](https://datatracker.ietf.org/doc/html/rfc7517#section-4.4) key to JWKS available at `/.well-known/jwks.json` - #17363 by @lkostrowski
 - Fix the issue of returning external shipping methods in a currency different from the checkout currency. - #17350 by @korycins
+- `checkout.shippingMethods` and `checkout.availableShippingMethods` will no longer return external shipping methods if their currency differs from the checkout's currency - #17350 by @korycins
--- a/saleor/graphql/checkout/tests/mutations/test_checkout_delivery_method_update.py
+++ b/saleor/graphql/checkout/tests/mutations/test_checkout_delivery_method_update.py
@@ -419,11 +419,8 @@ def test_checkout_delivery_method_update_external_shipping_invalid_currency(
     errors = data["errors"]
     assert len(errors) == 1
     assert errors[0]["field"] == "deliveryMethodId"
-    assert errors[0]["code"] == CheckoutErrorCode.DELIVERY_METHOD_NOT_APPLICABLE.name
-    assert (
-        errors[0]["message"]
-        == "Cannot choose shipping method with different currency than the checkout."
-    )
+    assert errors[0]["code"] == CheckoutErrorCode.NOT_FOUND.name
+    assert errors[0]["message"] == f"Couldn't resolve to a node: ${method_id}"
 
 
 @patch(
--- a/saleor/plugins/manager.py
+++ b/saleor/plugins/manager.py
@@ -2583,7 +2583,12 @@ class PluginsManager(PaymentInterface):
                 # https://github.com/python/mypy/issues/9975
                 getattr(plugin, "get_shipping_methods_for_checkout")(checkout, None)
             )
-        return shipping_methods
+        return list(
+            filter(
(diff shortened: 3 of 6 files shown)


IKarbowiak (dismissed):

saleor/plugins/manager.py
```diff
@@ -2583,7 +2583,12 @@ def list_shipping_methods_for_checkout(
                 # https://github.com/python/mypy/issues/9975
                 getattr(plugin, "get_shipping_methods_for_checkout")(checkout, None)
             )
-        return shipping_methods
+        return list(
```
krzysztofzuraw, line 2586: **Question:** I assume that instead of validation error after this change we will filter out shipping methods with different currency?

I yes - lets add this information to docs so no one will be confused

saleor/plugins/tests/sample_plugins.py
```diff
@@ -366,6 +367,24 @@ def payment_method_initialize_tokenization(self, request_data, previous_value):
...
+        return [
+            ShippingMethodData(
+                id="123",
+                price=Money(Decimal(10), currency=different_currency),
+                name="EUR shipping",
+            ),
+            ShippingMethodData(
+                id="123",
+                price=Money(Decimal(10), currency=checkout.currency),
+                name="Defualt shipping",
```
krzysztofzuraw, line 384:
```suggestion
                name="Default shipping",
```

saleor/plugins/tests/sample_plugins.py
```diff
@@ -366,6 +367,24 @@ def payment_method_initialize_tokenization(self, request_data, previous_value):
     def payment_method_process_tokenization(self, request_data, previous_value):
         return previous_value
 
+    def get_shipping_methods_for_checkout(
```
krzysztofzuraw, line 370: **Question:** this change is needed so we can test https://github.com/saleor/saleor/pull/17350/files#diff-a5848cf43f14aaf7a32dc81e09e933bf1b3c0fb9225bf3db7cd45026ec47b1deR1639-R1643 ?

krzysztofzuraw (dismissed):

CHANGELOG.md
```diff
@@ -58,3 +58,4 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fix checkout funds releasing task - #17198 by @IKarbowiak
 - Fixed 'healthcheck' middleware (`/health/` endpoint) not forwarding incoming traffic whenever the protocol wasn't HTTP (such as WebSocket or Lifespan) - #17248 by @NyanKiyoshi
 - Added support for the AWS_S3_URL_PROTOCOL environment variable - #17305 by @p-febis
+- Fix the issue of returning external shipping methods in a currency different from the checkout currency. - #17350 by @korycins
```
lkostrowski, line 61: But what was the issue? Let's ensure we have changelogs that explain before/after behavior if possible

saleor/plugins/manager.py
lkostrowski: yup, it's especially tricky because we dont have mechanism for saleor to inform app about invalid payload

korycins: > I assume that instead of validation error after this change we will filter out shipping methods with different currency?

This works on `BasePlugin` interface, so there is a possibility of having a custom shipping plugin. It means, that this is a single place where we can filter out the shipping methods used later by Saleor. We also can't raise a validation error, as the issue could be related to a single shipping method, where the next 10 are correct. In case of sync webhook - we receive the shipping methods in the webhook response, so we don't have the possibility to provide feedback to the app. As webhooks are not channel-specific, we need to take into account, that there is a possibility that someone would have a single app to support multiple currencies.

saleor/plugins/tests/sample_plugins.py
korycins: SamplePlugin is our test dummy plugin, that you can extend, to test the behaviour of PluginsManager, without working with mocks. 
So yes, this method is added here to, have a possibility to test the flow on pluginsManager, without using built-in plugins or mocking the code.

saleor/plugins/tests/sample_plugins.py
```diff
@@ -366,6 +367,24 @@ def payment_method_initialize_tokenization(self, request_data, previous_value):
...
+        self, checkout: "Checkout", previous_value: Any
+    ) -> list["ShippingMethodData"]:
+        different_currency = "EUR"
+        assert checkout.currency != different_currency
```
lkostrowski, line 374: @IKarbowiak this can be a nice addition to my statement about tests being explicit. So, we dont have to rely on "some nexted fixtures" being in "given" state we want. We also dont have to re-introduce them in every test, but we can assert fixtures being in valid state. If fixture changes, we fail fast (we know that "given" is broken before we check "then")

lkostrowski (changes requested):

saleor/plugins/manager.py
lkostrowski: >As webhooks are not channel-specific, we need to take into account, that there is a possibility that someone would have a single app to support multiple currencies.

Good point, lets ensure we have documented behavior that saleor will filter out currencies that don't match channel currency

korycins: https://github.com/saleor/saleor-docs/pull/1473

saleor/graphql/checkout/tests/mutations/test_checkout_delivery_method_update.py
```diff
@@ -419,11 +419,8 @@ def test_checkout_delivery_method_update_external_shipping_invalid_currency(
     errors = data["errors"]
     assert len(errors) == 1
     assert errors[0]["field"] == "deliveryMethodId"
-    assert errors[0]["code"] == CheckoutErrorCode.DELIVERY_METHOD_NOT_APPLICABLE.name
-    assert (
-        errors[0]["message"]
-        == "Cannot choose shipping method with different currency than the checkout."
-    )
+    assert errors[0]["code"] == CheckoutErrorCode.NOT_FOUND.name
```
fowczarek, line 422: Shouldn't it be described as breaking change? Someone could have implemented a notification based on this error code.

fowczarek: Also, we should add it to the migration guide in [this PR. ](https://github.com/saleor/saleor-docs/pull/1473/files)

lkostrowski (approved):

saleor/plugins/tests/sample_plugins.py
IKarbowiak: BTW, should we do such assertion in the `SamplePlugin`?

Assertion should in specific test unit IMO.

korycins: `assert` is not unit-test specific. You can find multiple assertions in the core code. 
In that case, I can't make `assert` inside the test, as sample plugin returns shipping methods with different currencies, but it is not available anymore for Saleor (PluginsManager, will filter it out), so only the `assert` here can confirm that  sample plugin returns the shipping methods different currency than checkout's currency (where single record has the same currency as the checkout's currency)

saleor/graphql/checkout/tests/mutations/test_checkout_delivery_method_update.py
korycins: It is not a breaking change from the API perspective. Anything that could cause such error is not possible anymore - we don't return external shipping with different currency than the checkout's currency. 

But adding it to migration guide is a valid point, in case someone would have an additional implementation, to filter out the shipping methods with incorrect currency on the frontend/app side. I've already added it :)

fowczarek (dismissed):

fowczarek (approved):

IKarbowiak (approved):

PR #17353, opened 2025-02-12
Improve setup for local development

witoszekdev:
I want to merge this change because it fixes `docker-compose.yaml` config, by exposing missing ports for redis and mailpit

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -20,7 +20,7 @@ services:
     image: ghcr.io/saleor/saleor-dashboard:3.20.5
     restart: unless-stopped
     ports:
-      - 9000:80
+      - 9000:9000
 
   db:
     image: library/postgres:15-alpine
@@ -38,12 +38,14 @@ services:
     restart: unless-stopped
     volumes:
       - saleor-redis:/data
+    ports:
+      - 6379:6379
 
   mailpit:
     image: axllent/mailpit
     ports:
-      - "1025" # SMTP Server
-      - "8025" # Mailpit UI
+      - 1025:1025 # SMTP Server
+      - 8025:8025 # Mailpit UI
     restart: unless-stopped
 
 volumes:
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -99,6 +98,26 @@ Install pre commit hooks:
 pre-commit install
 ```
 
+Set environment variables for your setup. You can use [direnv](https://direnv.net/) to do this automatically.
```
korycins, line 101: Maybe we can drop the env from here that has the same pre-defined value? Like `DEFAULT_CHANNEL_SLUG` (it already has `default-channel` value by default)
Also, instead of setting up some envs without explanation, I would suggest providing a link with an explanation for each of the envs that you need to set up. Most of the envs are described here: https://docs.saleor.io/setup/configuration#general-configuration

CONTRIBUTING.md
```diff
@@ -109,6 +128,12 @@ To start server:
 uvicorn saleor.asgi:application --reload
 ```
 
+To start Celery worker:
+
+```shell
+celery --app=saleor.celeryconf:app worker --beat
```
korycins, line 134: This would not work correctly as we're using a custom scheduler to gather the tasks (after first cycle, the celery beat will just stop working).  The proper way of setting up the celery beat is described here: https://docs.saleor.io/developer/running-saleor/task-queue#periodic-tasks

CONTRIBUTING.md
witoszekdev: I'll put better docs in poe setup, since it supports loading from env and you don't need `direnv` for that

CONTRIBUTING.md
witoszekdev: I've removed it, will add a separate beat command to poe setup

.devcontainer/docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
     image: ghcr.io/saleor/saleor-dashboard:3.20.5
     restart: unless-stopped
     ports:
-      - 9000:80
+      - 9000:9000
```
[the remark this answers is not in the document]
witoszekdev: you're right, my bad fix it back to `80`

krzysztofzuraw (approved):

korycins (approved):

PR #17351, opened 2025-02-12
Add Makefile with common dev commands

krzysztofzuraw:
I want to merge this change because this PR adds `Makefile` to store & easy reuse of common dev commands. You can read more on this [blog post](https://krzysztofzuraw.com/blog/2016/makefiles-in-python-projects/) 😄 

# Impact

- [ ] New migrations
- [ ] New/Updated API fields or mutations
- [ ] Deprecated API fields or mutations
- [ ] Removed API types, fields, or mutations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

- [ ] Privileged queries and mutations are either absent or guarded by proper permission checks
- [ ] Database queries are optimized and the number of queries is constant
- [ ] Database migrations are either absent or optimized for zero downtime
- [ ] The changes are covered by test cases
- [ ] All new fields/inputs/mutations have proper labels added (`ADDED_IN_X`, `PREVIEW_FEATURE`, etc.)
- [ ] All migrations have proper dependencies
- [ ] All indexes are added concurrently in migrations
- [ ] All RunSql and RunPython migrations have revert option defined

PR #17356, opened 2025-02-12
Add common tasks via PoeThePoet

witoszekdev:
I want to merge this change because it adds common tasks for running Saleor:
- uvicorn
- celery
- npm release script
- building GraphQL schema (removed from `package.json` scripts)

by using [poethepoet](https://poethepoet.natn.io/index.html) task runner.

It also automatically loads environment variables from `.env` file, see docs: https://poethepoet.natn.io/tasks/options.html#loading-environment-variables-from-an-env-file - this removes the need to use external tools like `dotenv` for loading env variables.

## Example usage

CleanShot 2025-02-13 at 15 14 35@2x

CleanShot 2025-02-13 at 15 15 16@2x


patrys:
My personal wishlist item: please replace the remaining tasks from `npm` too 🙏

lkostrowski:
https://github.com/saleor/saleor/pull/17351

:D

CONTRIBUTING.md
```diff
@@ -106,19 +106,19 @@ You are ready to go 🎉.
...
 ```shell
-python manage.py migrate
+poe migrate
 ```
 
 To populate database with example data and create the admin user:
 
 ```shell
-python manage.py populatedb --createsuperuser
+poe populatedb
```
krzysztofzuraw, line 121: **Suggestion:** lets add info and remove paragraph from line 124:

```
> [!NOTE]
> `populatedb` populates database with example data and creates an admin account for `admin@example.com` with the password set to `admin`.*
```

pyproject.toml
```diff
@@ -2,6 +2,21 @@
...
+worker.cmd = "celery --app=saleor.celeryconf:app worker --beat"
+
+migrate = "python manage.py migrate"
+populatedb = "python manage.py populatedb --createsuperuser"
+
+build-schema.shell = "python manage.py get_graphql_schema > saleor/graphql/schema.graphql"
+build-schema.help = "Build GraphQL schema to schema.graphql file"
+
+release.shell = "npm run release"
+
```
krzysztofzuraw, line 22: **Suggestion:** we can add `python manage.py shell` as well

witoszekdev:
@patrys I've migrated package.json scripts into poe ;)

CONTRIBUTING.md
```diff
@@ -103,26 +103,46 @@ You are ready to go 🎉.
 
 ### Common commands
 
+> [!TIP]
+> To run these commands you don't need to activate virtualenv via `poetry env activate`
```
krzysztofzuraw, line 107: I would add info that if you are in devcontainer you don't need to active env 😄

pyproject.toml
```diff
@@ -2,6 +2,41 @@
...
+release.shell = "npm run release"
+release.help = """
+Release new Saleor verison
+Prerequisities:
+- Node installed on machine
+- Run npm install to install dependencies
+"""
+
+shell.cmd = "python manage.py shell"
+shell.help = "Start Python interactive interpreter"
```
krzysztofzuraw, line 31:
```suggestion
shell.help = "Start Django shell in interactive mode"
```

krzysztofzuraw (dismissed):

lkostrowski (dismissed):

IKarbowiak (dismissed):

CONTRIBUTING.md
korycins: Is that sentence correct? You added poe to `pyproject`, which will install the `poe` to the env which is active when you call `poetry install` right? Then to use it, you need to have active virtual env (assuming that you're using virutalenv).  Additionally, the poe's commands, calls django commands, which would require the virtualenv. Do I miss something? 🤔

witoszekdev: @korycins you were right, it worked _for me_ because I've installed `poe` as their docs suggested via `pipx` globally 😅

if it's installed only for Saleor, then this won't work, I've removed this tip.

krzysztofzuraw (dismissed):

krzysztofzuraw (approved):

lkostrowski (approved):

PR #17366, opened 2025-02-13
Unify assignment of delivery method to checkout

korycins:
- unify the logic for assigning the delvieryMethod to checkout (shippingMethod, externalShippingMethod, Warehouse). Previously the `checkoutDeliveryMethod` and `checkoutShippingMethodUpdate` had own implementation of assignment process. It was hard to read and understand (a lot of `if` condtions also), and in some specific cases caused different behaviour.
-  Reduce the amount of data that we fetch to confirm if delviery emthod is valid (if method is WH, we don't need shipping methods, and if method is ShippingMethod, we don't need to fetch C&C points) - [ref](https://github.com/saleor/saleor/pull/17366/files#diff-df51b1fe10bd3e3ceb77b16b763c9da977a1e92555d28868c1ca98d03578c7fbR92) 
- add multiple tests to cover different cases like: switching the delivery method from C&C to externalShippingMethod
- invalidate-taxes and db save on checkout will happen only in situation where we change the delivery method - no need to invalidate them if someone tried to assign the same delivery method (or remove delivery method, from the checkout whout delivery method).
- do not call checkout-updated webhook if nothing changed (same case as with `invalidate-taxes` ☝️ )

> [!NOTE]  
> There is a change for pluginsManager in the file changes. This is something addressed in separate [PR](https://github.com/saleor/saleor/pull/17350). As these changes are going to feature-branch (not main), I decided to keep it. I will rebase the feature-branch, after merging the mentioned changes. So please just ignore it for now 🙏 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/graphql/checkout/tests/mutations/test_checkout_delivery_method_update.py
```diff
@@ -1461,3 +1462,729 @@ def test_checkout_delivery_method_update_external_shipping_triggers_webhooks(
...
+    # when
+    response = api_client.post_graphql(
+        MUTATION_UPDATE_DELIVERY_METHOD,
+        {"id": to_global_id_or_none(checkout), "deliveryMethodId": method_id},
+    )
+
+    # then
+    data = get_graphql_content(response)["data"]["checkoutDeliveryMethodUpdate"]
+    errors = data["errors"]
+    assert not errors
```
IKarbowiak, line 1515: Shouldn't we also check the response from the query? 
It's important to have a valid data on DB, but we should ensure that proper data are returned in the response as well.

IKarbowiak: The comment concerns all added tests in this file

saleor/graphql/checkout/tests/mutations/test_checkout_shipping_method_update.py
IKarbowiak, line 1: Same here

fowczarek (approved):

IKarbowiak (approved):

IKarbowiak (approved):

PR #17369, opened 2025-02-14
Add poetry lock check to pre commit

krzysztofzuraw:
I want to merge this change because it adds [`poetry lock`](https://python-poetry.org/docs/pre-commit-hooks/#poetry-lock) to pre commit configuration. It will help us with not merging outdated poetry lock file.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,6 +53,11 @@ repos:
         # Replace --quiet with --verbose to debug issues.
         - --quiet
 
+  - repo: https://github.com/python-poetry/poetry
+    rev: 2.0.1
+    hooks:
+      - id: poetry-lock
+
   - repo: local
     hooks:
       - id: migrations-check
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,6 +57,7 @@ repos:
     rev: 2.0.1
     hooks:
       - id: poetry-lock
+        files: "^pyproject.toml$"
 
   - repo: local
     hooks:


.pre-commit-config.yaml
```diff
@@ -53,6 +53,11 @@ repos:
         # Replace --quiet with --verbose to debug issues.
         - --quiet
 
+  - repo: https://github.com/python-poetry/poetry
+    rev: 2.0.1
+    hooks:
+      - id: poetry-lock
```
lkostrowski, line 59: lock or check?  I think it's the same dilemma with build-schema?

krzysztofzuraw: According to poetry docs:
> The poetry-lock hook calls the poetry lock command to make sure the lock file is up-to-date when committing changes.
There is also `poetry-check` but I don't know if we need it right now

krzysztofzuraw: I think we can start with check and then allow developer to manually update - the same as `pnpm install --frozen-lockfile` fails without update 🤔

IKarbowiak:
I don't think that poetry lock should be a part of the pre commit. It's already quite long and I would avoid extending it. 
1. It's rather rare situation that something like that would happened
2. We have a bot that is doing a PRs with poetry lock update
3. This check will run with your commit, but poetry lock usually not depend on your changes.

korycins:
> I don't think that poetry lock should be a part of the pre commit. It's already quite long and I would avoid extending it.
> 
> 1. It's rather rare situation that something like that would happened
> 2. We have a bot that is doing a PRs with poetry lock update
> 3. This check will run with your commit, but poetry lock usually not depend on your changes.

Valid point. So maybe worth extending the check to be called only, when we extend the `pyproject.toml` file (as this is the one responsible for building the lock, afaik).

krzysztofzuraw:
@IKarbowiak & @korycins:
1. AFAIK `pre-commit` tool works on only changed files so if there is no change in pyproject.toml it won't run and slow pre-commit execution.
2. We already have situation where we merged PR without lock update and it makes `CI` on `main` branch not working - this is the main issue I wanted to avoid by adding this check so we won't relay on human to check if lock is regenerated but automated check.

lkostrowski (approved):

IKarbowiak (approved):

korycins (approved):

PR #17372, opened 2025-02-14
Allow providing saving address strategy in order mutations 

IKarbowiak:
Allow providing saving address option in the following checkout mutations:

- `DraftOrderCreate`
- `DraftOrderUpdate`
- ~~`OrderBulkCreate`~~

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


fowczarek (approved):

saleor/graphql/order/bulk_mutations/order_bulk_create.py
```diff
@@ -1050,6 +1074,20 @@ def get_instances_related_to_order(
                         code=OrderBulkCreateErrorCode.INVALID,
                     )
                 )
+        if (
+            not shipping_address
```
korycins, line 1078: What if `user` is None? Shouldn't we also raise an exception?

saleor/graphql/order/tests/mutations/test_order_bulk_create.py
```diff
@@ -4053,3 +4057,185 @@ def test_order_bulk_create_shipping_price_zero(
...
+
+    order = data[0]["order"]
+    assert order["created"]
+
+    db_order = Order.objects.get()
+    assert db_order.billing_address
+    assert db_order.shipping_address
+    assert db_order.draft_save_billing_address is None
+    assert db_order.draft_save_shipping_address is None
+    assert customer_user.addresses.count() == user_address_count + 2
```
korycins, line 4106: Can we also confirm that the `db_order.shipping_address.id != user.addresses[x].id`? (same for billing)

saleor/graphql/order/bulk_mutations/order_bulk_create.py
IKarbowiak: We are ignoring it in such scenario. I would like to have the behaviour consistent with other mutation. In case of user is not set we are ignoring the option.

IKarbowiak: I removed support for `OrderBulkCreate` as it was missing checking the address duplication and it's not easy to implement it with good performance. The support will be added if there will be the need for that.

korycins (approved):

PR #17376, opened 2025-02-17
Use get_or_create instead of save when creating a customer

lkostrowski:
I want to merge this change because it fixes issue with race condition when concurrent customerCreate calls were executed.

A 2nd call was run before validation kick in with email uniqueness but after 1st call created a user in DB

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/graphql/account/mutations/base.py
```diff
@@ -324,7 +328,21 @@ def save(cls, info: ResolveInfo, instance, cleaned_input):
             instance.default_billing_address = default_billing_address
 
         is_creation = instance.pk is None
-        super().save(info, instance, cleaned_input)
+
+        try:
+            instance.save()
+        except IntegrityError as e:
```
fowczarek, line 334: I don't think catching `IntegrityError` is the best solution. Integrit error now always means that email exists. It could be released also when the Non-Null field is empty. 

Maybe we could use [get_or_create](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#get-or-create)? All data except email should be passed as `defaults`.

fowczarek: > it happens that nun-null is raised, we will know on Sentry

In the current Django implementation when a nun-null error is detected Django throws an IntegrityError. Then your code will return `ValidationError` with a duplicated email message. No error will be logged and no data is sent to Sentry.

lkostrowski: Ah, that's a good point - is there a way to extract this specific error so we can map to our email validaiton, other than parsing error message?

korycins: Owczar is right here. We shouldn't catch IntegrityError and dummy assume that the `email` caused the problem. 

> If it happens that nun-null is raised, we will know on Sentry

You already hide any potential integrity error here. You won't see them on Sentry. No matter if the problem is related to the `email` or to something else.

> For example we will have to unpack existing instance to get metadata and pack it again to defaults. We wanted to reduce complexity of change

Why?  If `get_or_create` return `user`, `false`. It means that the user is already present, and we should abort the mutation. You don't need to pack anything to `defaults` except the input provided by user. It will be used only in case when `User` does not exist. 

If you're thinking about Update mutations which uses the `BaseCreate` mutation, then I would say, that this logic should be separated. I mean right now `BaseCreate` has implementation strictly related to children class(Update) which is wrong, as it should not know anything about update action - it is class **for create** not base for create or update.

If the problem is related how the ModelMutation works, then I think we should drop usage of it, or "force" it to work in expected way for us.

fowczarek: ~~We could use `model_to_dict` to create `defaults` from `instance`.~~ This solution doesn't work with relations. 
e.g. 
```python
        from django.forms.models import model_to_dict
        from saleor.account.models import User as UserModel

        defaults = model_to_dict(instance)
        del defaults["email"]

        instance, created = UserModel.objects.get_or_create(
            email=instance.email, defaults=defaults
        )
        if created is False:
            raise ValidationError(
                {
                    # This validation error mimics built-in validation error
                    # So graphQL response is the same
                    "email": ValidationError(
                        "User with this Email already exists.",
                        code=AccountErrorCode.UNIQUE.value,
                    )
                }
            )
```

Other option is to use similar error handling like Django use in `get_or_create`. 
https://github.com/django/django/blob/0bf412111be686b6b23e00863f5d449d63557dbf/django/db/models/query.py#L942
```python
        from saleor.account.models import User as UserModel

        try:
            instance.save()
        except IntegrityError:
            try:
                user = UserModel.objects.get(email=instance.email)
                raise ValidationError(
                    {
                        # This validation error mimics built-in validation error
                        # So graphQL response is the same
                        "email": ValidationError(
                            "User with this Email already exists.",
                            code=AccountErrorCode.UNIQUE.value,
                        )
                    }
                )
            except self.model.DoesNotExist:
                pass
            raise
```

lkostrowski: changes applied

lkostrowski:
closing - I will only merge 3.20 port, this will be solved different way

PR #17359, opened 2025-02-13
Add fields to checkout to store external shipping methods

korycins:
I want to merge this change because it ports changes from: https://github.com/saleor/saleor/pull/17339 to cover the compatibility in DB between 3.20 and main

> It adds new fields `external_shipping_method_id`, `shipping_method_name`  and `undiscounted_base_shipping_price`. All of these fields are required to block the webhooks in bulk queries (as already prepared for tax part: https://github.com/saleor/saleor/pull/17268).
> These fields will store the information to provide enough context to return the response when making a bulk queries.
> 
> Internal task link: https://linear.app/saleor/issue/EXT-1880/denormalize-shipping-details-on-checkout

# Impact

- [ ] New migrations
- [ ] New/Updated API fields or mutations
- [ ] Deprecated API fields or mutations
- [ ] Removed API types, fields, or mutations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

- [ ] Privileged queries and mutations are either absent or guarded by proper permission checks
- [ ] Database queries are optimized and the number of queries is constant
- [ ] Database migrations are either absent or optimized for zero downtime
- [ ] The changes are covered by test cases
- [ ] All new fields/inputs/mutations have proper labels added (`ADDED_IN_X`, `PREVIEW_FEATURE`, etc.)
- [ ] All migrations have proper dependencies
- [ ] All indexes are added concurrently in migrations
- [ ] All RunSql and RunPython migrations have revert option defined

PR #17381, opened 2025-02-18
Denormalize assigned shipping method to checkout

korycins:
The branch contains the changes form two PRs:

https://github.com/saleor/saleor/pull/17359

> I want to merge this change because it ports changes from: https://github.com/saleor/saleor/pull/17339 to cover the compatibility in DB between 3.20 and main
It adds new fields external_shipping_method_id, shipping_method_name and undiscounted_base_shipping_price. All of these fields are required to block the webhooks in bulk queries (as already prepared for tax part: https://github.com/saleor/saleor/pull/17268).
These fields will store the information to provide enough context to return the response when making a bulk queries.
Internal task link: https://linear.app/saleor/issue/EXT-1880/denormalize-shipping-details-on-checkout

And https://github.com/saleor/saleor/pull/17366

> unify the logic for assigning the delvieryMethod to checkout (shippingMethod, externalShippingMethod, Warehouse). Previously the checkoutDeliveryMethod and checkoutShippingMethodUpdate had own implementation of assignment process. It was hard to read and understand (a lot of if condtions also), and in some specific cases caused different behaviour.
> Reduce the amount of data that we fetch to confirm if delviery emthod is valid (if method is WH, we don't need shipping methods, and if method is ShippingMethod, we don't need to fetch C&C points) - [ref](https://github.com/saleor/saleor/pull/17366/files#diff-df51b1fe10bd3e3ceb77b16b763c9da977a1e92555d28868c1ca98d03578c7fbR92)
> add multiple tests to cover different cases like: switching the delivery method from C&C to externalShippingMethod
> invalidate-taxes and db save on checkout will happen only in situation where we change the delivery method - no need to invalidate them if someone tried to assign the same delivery method (or remove delivery method, from the checkout whout delivery method).
> do not call checkout-updated webhook if nothing changed (same case as with invalidate-taxes ☝️ )

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


korycins:
(Skipping changelog, as it doesn't change anything from the API perspective)

IKarbowiak (approved):

fowczarek (approved):

PR #17382, opened 2025-02-19
Add metadata to CheckoutCreateInput

lkostrowski:
I want to merge this change because it allows to set `metadata` and `privateMetadata` when checkout is created by `checkoutCreate`

It also introduces a small refactor - extracted "MetadataManager" to keep utils connected to metadata & private_metadata together.

Old code (base mutations) still contain methods for metadata, but these methods use MetadataManager now.
New code, can directly use MetadataManager, to avoid relying on inherited behavior.

For example, Checkout metadata is *not* saved on Checkout model, but CheckoutCreate mutation inherits the behavior, not matter what.

Now, CheckoutCreate can use MetadataManager under control, while the rest of the code is not changed.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/graphql/checkout/mutations/checkout_create.py
```diff
@@ -376,7 +394,35 @@ def save(cls, info: ResolveInfo, instance: models.Checkout, cleaned_input):
                 instance.billing_address = billing_address.get_copy()
 
             instance.save()
-            create_checkout_metadata(instance)
+
+    @classmethod
+    def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
+        # Write metadata in post-save action because in "save" metadata doesn't exist in
+        # cleaned_input.
```
lkostrowski, line 401: this should be refactored once we refactor this inheritance-hell

saleor/graphql/checkout/tests/mutations/test_checkout_create.py
```diff
@@ -2791,3 +2791,46 @@ def test_checkout_create_triggers_webhooks(
     assert WebhookEventSyncType.CHECKOUT_CALCULATE_TAXES in sync_deliveries
     tax_delivery = sync_deliveries[WebhookEventSyncType.CHECKOUT_CALCULATE_TAXES]
     assert tax_delivery.webhook_id == tax_webhook.id
+
+
+def test_checkout_create_with_metadata(api_client, graphql_address_data, channel_USD):
```
IKarbowiak, line 2796: Maybe we should also add a test for providing only `private_metadata` and only `metadata`.

Also missing `given, when, then`.

saleor/graphql/checkout/mutations/checkout_create.py
```diff
@@ -159,6 +159,18 @@ class CheckoutCreateInput(BaseInputObjectType):
         description=("The checkout validation rules that can be changed."),
     )
 
+    metadata = NonNullList(
+        MetadataInput,
+        description="Checkout public metadata." + ADDED_IN_321,
+        required=False,
+    )
+
+    private_metadata = NonNullList(
```
korycins, line 168: Adding anything to private_metadata should require proper permission.  `MANAGE_CHECKOUTS` in that case.

saleor/graphql/checkout/tests/mutations/test_checkout_create.py
lkostrowski: done

saleor/graphql/checkout/mutations/checkout_create.py
lkostrowski: done + added test for that

saleor/graphql/checkout/mutations/checkout_create.py
```diff
@@ -376,7 +411,35 @@ def save(cls, info: ResolveInfo, instance: models.Checkout, cleaned_input):
...
+        # Write metadata in post-save action because in "save" metadata doesn't exist in
+        # cleaned_input.
+
+        checkout_metadata = create_checkout_metadata(instance)
+
+        metadata_from_input = cleaned_input.get("metadata", [])
+        private_metadata_from_input = cleaned_input.get("private_metadata", [])
+
+        nothing_to_write = (
+            len(metadata_from_input) == 0 and len(private_metadata_from_input) == 0
```
aniav, line 426: Nitpick: you can also do
```
if not metadata_from_input and not private_metadata_from_input:
    return
```

It's a bit easier to read.

saleor/graphql/checkout/mutations/checkout_create.py
```diff
@@ -376,7 +411,35 @@ def save(cls, info: ResolveInfo, instance: models.Checkout, cleaned_input):
...
+        )
+
+        if nothing_to_write:
+            return
+
+        metadata_dict = {
```
aniav, line 432: Why do you need to go through that metadata if it's already in key, value pairs? If there is a reason maybe you can add a comment about it?

korycins: > Why do you need to go through that metadata if it's already in key, value pairs?

Input is a list of objects: [ {key:"key provided by user", "value": "value provided by user}]

saleor/graphql/checkout/mutations/checkout_create.py
```diff
@@ -376,7 +411,35 @@ def save(cls, info: ResolveInfo, instance: models.Checkout, cleaned_input):
                 instance.billing_address = billing_address.get_copy()
 
             instance.save()
-            create_checkout_metadata(instance)
+
+    @classmethod
+    def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
```
korycins, line 416: I think, that you shouldn't make your own logic here, but re-use the existing one. 
In that case, it could be something like this:
```python
    @classmethod
    def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
        # Write metadata in post-save action because in "save" metadata doesn't exist in
        # cleaned_input.
        checkout_metadata = CheckoutMetadata(checkout=checkout)
        cls.validate_and_update_metadata(
            checkout_metadata, 
            cleaned_input.get("metadata"), 
            cleaned_input.get("private_metadata")
        )
        checkout_metadata.save()
```

Your current implementation allows me to provide `key` == `""`, where the `cls.validate_and_update_metadata` will raise an exception

Then I think, we can also drop `create_checkout_metadata` function (as it is used only in this place right now).

saleor/graphql/checkout/mutations/checkout_create.py
lkostrowski: Okay 😄 In JS empty arrays work different 👍

saleor/graphql/checkout/mutations/checkout_create.py
korycins: You can just make at the end:
```python
with allow_writer():
    checkout_metadata.save()
```
You will have a single DB query instead of two (`create_checkout_metadata` and later `checkout_metadata`)

saleor/graphql/core/utils/metadata_manager.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def update_metadata_on_instance(
+    instance: ModelWithMetadata,
+    metadata: list[dict] | None,
+    private_metadata: list[dict] | None,
+):
+    if metadata is not None:
+        instance.store_value_in_metadata(
+            {data["key"]: data["value"] for data in metadata}
```
lkostrowski, line 33: I changed it to read from dict (["key"]) but previously it was

```
instance.store_value_in_metadata(
                {data.key: data.value for data in meta_data_list}```

not sure if I just broke something...

PR #17383, opened 2025-02-19
Remove reference cycle when query returns error.

fowczarek:
I want to merge this change because it allows memory to be freed immediately without needing a deep garbage collection cycle when the API returns an error. 

We don't create a bugfix in the original repository because `GraphQL-core-legacy` is not supported. 

`--dist loadgroup` is added to `pytest` to prevent test Interference. [Docs. ](https://pytest-xdist.readthedocs.io/en/stable/distribution.html)

> `--dist loadgroup`: Tests are grouped by the xdist_group mark. Groups are distributed to available workers as whole units. This guarantees that all tests with same xdist_group name run in the same worker.

Garbage before when `PermissionDenied` on 1 field:
garbage_permission_error_1_field

Garbage before when an unexpected error appears in the resolver:
garbage_resolver_error

Garbage before when an unexpected error appears in the dataloader:
garbage_dataloader_error

Garbage after changes in all cases:
garbage_after

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


patrys:
Can you fix the title of this PR?

IKarbowiak (approved):

korycins (approved):

PR #17384, opened 2025-02-19
[POC] Telemetry - propagate context for Celery tasks

przlada:

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/core/telemetry/utils.py
+++ b/saleor/core/telemetry/utils.py
@@ -1,7 +1,12 @@
+from collections.abc import Callable
 from contextlib import contextmanager
 from contextvars import ContextVar
+from dataclasses import dataclass, field
 from enum import Enum
+from functools import wraps
+from typing import Any
 
+from opentelemetry.trace import Link, SpanContext
 from opentelemetry.util.types import Attributes, AttributeValue
 
 Amount = int | float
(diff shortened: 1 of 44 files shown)


saleor/core/telemetry/utils.py
```diff
@@ -49,6 +56,81 @@ def set_global_attributes(attributes: dict[str, AttributeValue]):
...
+    def from_dict(cls, data: dict | None) -> "TelemetryContext":
+        if not data:
+            return cls(global_attributes={})
+        try:
+            links = [
+                Link(
+                    context=SpanContext(
+                        trace_id=link["context"]["trace_id"],
+                        span_id=link["context"]["span_id"],
+                        is_remote=True,
```
maarcingebala, line 103: When looking into this I wasn't sure whether "remote" in this context refers to the fact that the parent span comes from distributed tracing (e.g. external app) or not. Celery tasks are somehow remote to span produced by core, but they are part of the same app. Do you have any reference in docs for this?

przlada: Good catch! I made an assumption and forgot about this
Based on this https://opentelemetry.io/docs/specs/otel/trace/api/#isremote
> An API called IsRemote, that returns a boolean value, which is true if the SpanContext was propagated from a remote parent, MUST be provided. When extracting a SpanContext through the Propagators API, IsRemote MUST return true, whereas for the SpanContext of any child spans it MUST return false.

I understand that `is_remote` should be `True` whenever any "context propagation" happens (our case with Celery) and `False` only if we are creating local child spans. I also believe that the `is_remote` property is fully ignored when [linking spans](https://opentelemetry.io/docs/specs/otel/overview/#links-between-spans), but I would need to dig into the code to fully verify that

saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -470,11 +479,13 @@ def trigger_webhooks_async(
 
 @app.task(bind=True)
 @allow_writer()
+@task_with_telemetry_context
 def generate_deferred_payloads(
     self,
     event_delivery_ids: list,
     deferred_payload_data: dict,
     send_webhook_queue: str | None = None,
+    span_links: Sequence[Link] | None = None,
```
maarcingebala, line 488: I have second thoughts about this bit where it looks like the task is expecting `span_links`, while we call it with `telemetry_context`, which gets rewritten in the decorator. Would it be possible to supply the global attributes inside the `get_current_context` function? Then I think whole decorator could be skipped and we could simply pass `telemetry_context` (as dict) between tasks.

saleor/core/telemetry/utils.py
```diff
@@ -49,6 +56,81 @@ def set_global_attributes(attributes: dict[str, AttributeValue]):
...
+
+
 def enrich_with_global_attributes(attributes: Attributes) -> Attributes:
-    trace_attributes = _GLOBAL_ATTRS.get({})
-    return {**(attributes or {}), **trace_attributes}
+    return {**(attributes or {}), **get_global_attributes()}
+
+
+@dataclass(frozen=True)
+class TelemetryContext:
```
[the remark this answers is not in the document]
przlada: That is a very good point. As this class represents telemetry data (span links and global attributes) which we want to propagate to async Celery tasks , I propose to change the name to `TelemetryTaskContext` / `get_task_context`, what do you think?

saleor/webhook/transport/asynchronous/transport.py
przlada: I agree with you, I changed it. But I propose to stick with the `task_with_telemetry_context` decorator, as apart from deserialization it also sets global attributes for the whole task. I fixed the decorator's docstring to describe it better. Is it ok for you?

maarcingebala (approved):
LGTM 👍

PR #17387, opened 2025-02-19
Block shipping webhooks for bulk checkouts queries

korycins:
I want to merge this change because it blocks webhook calls to attach external shipping methods, filter shipping methods, when using bulk queries like `checkouts`, `me.checkouts` etc. If external shipping is attached, then the de-normalized data will be used to present the shipping object. 

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation: https://github.com/saleor/saleor-docs/pull/1459

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/tests/test_checkouts_query.py
+++ b/saleor/graphql/checkout/tests/test_checkouts_query.py
@@ -1,0 +1,289 @@
+from decimal import Decimal
+from unittest.mock import patch
+
+from django.utils import timezone
+
+from saleor.graphql.core.utils import to_global_id_or_none
+from saleor.graphql.tests.utils import get_graphql_content
+from saleor.shipping.models import ShippingMethod
+from saleor.webhook.event_types import WebhookEventSyncType
+
+CHECKOUTS_QUERY = """
+query CheckoutsQuery {
+  checkouts(first: 1) {
+    edges {
+      node {
+        deliveryMethod {
+          __typename
+          ... on ShippingMethod {
+            id
+            name
+            price {
+              amount
+            }
+          }
+        }
+        shippingPrice {
+          gross {
+            amount
+          }
+        }
+        totalPrice {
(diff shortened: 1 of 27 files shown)


korycins:
Note: There is `pregenerated_payloads_for_excluded_shipping_methods` which is always called right now.  I am addressing it right now in separate PR

IKarbowiak (dismissed):

saleor/graphql/checkout/tests/test_checkouts_query.py
```diff
@@ -0,0 +1,289 @@
+from decimal import Decimal
+from unittest.mock import patch
+
+from django.utils import timezone
+
+from saleor.graphql.core.utils import to_global_id_or_none
+from saleor.graphql.tests.utils import get_graphql_content
+from saleor.shipping.models import ShippingMethod
+from saleor.webhook.event_types import WebhookEventSyncType
```
zedzior, line 9: Absolute paths

saleor/graphql/checkout/tests/test_checkouts_query.py
```diff
@@ -0,0 +1,289 @@
+from decimal import Decimal
```
zedzior, line 1: To be consistent, this file should be moved and renamed to: `saleor/graphql/checkout/tests/queries/test_checkouts.py`

zedzior (dismissed):

kadewu (dismissed):

saleor/graphql/checkout/tests/test_checkouts_query.py
korycins: To be consistent with what? Checkout doesn't have `queries`.  I would need to remove all query tests and place them inside `queries`.

IKarbowiak (approved):

kadewu (approved):

PR #17405, opened 2025-02-25
Fix app resolver calling update state

tomaszszymanski129:
I want to merge this change because it fixes the `App` type resolver of `breakerState` calling `update_breaker_state` which it should not.

Also refactored some breaker methods - because we had separate for updating last open, last change, state, fetching all three. Now there is simply `set_app_state` / `get_app_state`.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/app/types.py
+++ b/saleor/graphql/app/types.py
@@ -621,7 +621,8 @@ class App(ModelObjectType[models.App]):
     def resolve_breaker_state(root: models.App, _info: ResolveInfo):
         if not breaker_board:
             return CircuitBreakerState.CLOSED
-        return breaker_board.update_breaker_state(root)
+        _, state = breaker_board.storage.last_open(root.id)
+        return state
 
     @staticmethod
     def resolve_breaker_last_state_change(root: models.App, _info: ResolveInfo):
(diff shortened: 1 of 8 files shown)


saleor/graphql/app/types.py
```diff
@@ -621,7 +621,8 @@ def resolve_brand(root: models.App, _info: ResolveInfo):
     def resolve_breaker_state(root: models.App, _info: ResolveInfo):
         if not breaker_board:
             return CircuitBreakerState.CLOSED
-        return breaker_board.update_breaker_state(root)
+        _, state = breaker_board.storage.last_open(root.id)
```
cmiacz, line 624: `last_open` method name and return value is confusing. It returns timestamp and state, but this timestamp seems useless and is ignored. It can also return `CLOSED` state with timestamp 0. On the other hand we have `retrieve_last_state_change` methods.

I'd suggest to refactor and merge these methods into one e.g. `get_current_state` returning both state and a timestamp of last state change. In this case returing tuple of `(CircuitBreakerState.CLOSED, 0)` is case when no key is found in redis is fine. 

This refactoring would also require minor changes in `BreakerBoard.set_breaker_state` method.
This line doesn't seem valid as we can set `CLOSED` state too and timestamp is a valid inf then
https://github.com/saleor/saleor/blob/525afacd01f05e97ad754818e4d5dc50aec9c53d/saleor/webhook/circuit_breaker/breaker_board.py#L113

cmiacz (changes requested):

cmiacz (approved):

cmiacz (approved):

PR #17415, opened 2025-02-26
Add `draftOrderLinePriceFreezePeriod`

zedzior:

This PR introduces `channel.draftOrderLinePriceFreezePeriod` setting which determine the time in hours after which the draft order line prices should be refreshed.

# Impact

- [x] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/channel/models.py
```diff
@@ -48,6 +48,9 @@ class Channel(ModelWithMetadata):
 
     use_legacy_error_flow_for_checkout = models.BooleanField(default=True)
     automatically_complete_fully_paid_checkouts = models.BooleanField(default=False)
+    draft_order_line_price_freeze_period = models.IntegerField(
```
IKarbowiak, line 51: Maybe we should add a comment that the value is in hours?

IKarbowiak: You should also use `PositiveIntegerField` instead of `IntegerField`

korycins (approved):

saleor/channel/models.py
zedzior: corrected

PR #17418, opened 2025-02-26
Add `draft_base_price_expire_at` field to `OrderLine`

zedzior:

This PR: 
- introduces `OrderLine.draft_base_price_expire_at` field, which store the date of base prices expiration
- the expiration time is set when adding new line and during the price refresh
- the price has no expiration time for lines with price overridden

# Impact

- [x] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/order/models.py
```diff
@@ -710,6 +710,10 @@ class OrderLine(ModelWithMetadata):
     # Fulfilled when sale was applied to product in the line
     sale_id = models.CharField(max_length=255, null=True, blank=True)
 
+    # The date time when the line should refresh its prices.
+    # It depends on channel.draft_order_line_price_freeze_period setting.
+    price_expire_at = models.DateTimeField(blank=True, null=True)
```
korycins, line 715: Shouldn't we also use the prefix `draft_` here?  As this is something related only to the draft.

saleor/discount/utils/order.py
```diff
@@ -341,6 +343,11 @@ def refresh_order_base_prices_and_discounts(
     # update unit discount fields based on updated discounts
     update_unit_discount_data_on_order_line(lines_info)
 
+    # set price expiration time
+    expiration_time = get_order_line_price_expiration_date(order.channel)
+    for line_info in lines_info_to_update:
+        line_info.line.price_expire_at = expiration_time
```
korycins, line 349: Shouldn't you also take into account `is_price_overridden`? When provided, do not set `price_expire_at`?

saleor/order/models.py
zedzior: For non draft orders this will be `null` so IMO the prefix is not needed

saleor/discount/utils/order.py
zedzior: In can add extra guard here, but the lines with `is_price_overridden` will never have `base_price_expire_at` set, therefore will never be included in `lines_info_to_update`

zedzior: I also think that we should have an option to refresh the line with `is_price_overridden` over a mutation, when explicitly pass the id of the line with custom price. Does it make sense?

saleor/order/models.py
korycins: I know that this flag is only for `draft`.  But I am talking about better readability. 
For example, Iga is adding a save_address to drafts. The flags are NULL for order!=DRAFT, but the name still has a prefix `draft_`. Just to emphasize, that this field is used by draft orders.
image

You know that this is only for draft, I know that this is only for draft, but we know that because we are 
engaged in the process of adding this. Someone who is not familiar with this feature, would not know what this value means. 
If someone would investigate something with orders, and will fetch this flag directly from the database, will not know if this is related to all order lines, or only in the specific case. This is why IMHO, we should be more specific here. 
This name would be OK, if we would have a separate models for `OrderLine` & `DraftOrderLine`, and you would add it directly to the `DraftOrderLine`

> For non draft orders this will be null

BTW, can you point me a place where you clean up this value? I mean conversation from Draft -> non-draft.

saleor/discount/utils/order.py
korycins: > In can add extra guard here, but the lines with `is_price_overridden` will never have `base_price_expire_at` set, therefore will never be included in `lines_info_to_update`

As keeping the price_overridden is a business requirement, I would always skip such lines. I would also probably add the logic to the `get_expired_line_ids` to explicitly filter out the lines with price_overridden.

> therefore will never be included in `lines_info_to_update`

Yes, if you assume that this function will get the expired lines from `get_expired_line_ids`. If I add  all lines as `expired_lines` input, I will also refresh the ones with price_overriden, right? I know that it works now as expected, but It would be really great to have explicit confirmation, that we don't change the overridden price, no matter what is provided in the input.

saleor/order/models.py
zedzior: IMO we shouldn't be more precise here. Ie. if at some point we decide to expire the price for `UNCONFIRMED` orders, the name with prefix will be misleading.

`draft_save_billing_address` - for me this unclear. What `draft save` means? 
Same here: If you want to add prefix we end up with `draft_base_price_expire_at` - what `draft base price` means?

zedzior: Good point about cleaning during conversion - let me add it

saleor/discount/utils/order.py
zedzior: Ok I will add additional check here, but in my opinion we should be able to force the update of lines with overridden price when explicitly requested. Need to check it with product and it can wait for implementing the mutation

saleor/order/models.py
fowczarek: We should add the information in the comment that this field only affects the order draft.
I prefer to add the `draft_` prefix to fields that affect only draft orders. 

>  if at some point we decide to expire the price for UNCONFIRMED orders, the name with prefix will be misleading.

If this field changes their behaviour we should also rename it.

maarcingebala: Regarding the naming - I understand both sides here, but I would say `price_expire_at` is a fine name, with extra comment that describes the purpose. It's something that you check once when unsure, and then it becomes clear. DB doesn't have to represent that IMO.

IKarbowiak (approved):

korycins (approved):

PR #17419, opened 2025-02-26
Refresh order prices after freeze period has passed

zedzior:

After the freeze period has passed for particular order line, the refresh logic introduced in https://github.com/saleor/saleor/pull/17326 will be fired. The price refresh is only triggered for DRAFT orders

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/order/calculations.py
+++ b/saleor/order/calculations.py
@@ -50,7 +50,7 @@ from .fetch import (
 from .interface import OrderTaxedPricesData
 from .models import Order, OrderLine
 from .utils import (
-    get_order_line_price_expiration_date,
+    calculate_draft_order_line_price_expiration_date,
     log_address_if_validation_skipped_for_order,
     order_info_for_logs,
 )
@@ -564,7 +564,9 @@ def refresh_order_base_prices_and_discounts(
     update_unit_discount_data_on_order_line(lines_info)
 
     # set price expiration time
-    expiration_time = get_order_line_price_expiration_date(order)
+    expiration_time = calculate_draft_order_line_price_expiration_date(
+        order.channel, order.status
+    )
     for line_info in lines_info_to_update:
         line_info.line.draft_base_price_expire_at = expiration_time
 
--- a/saleor/order/calculations.py
+++ b/saleor/order/calculations.py
@@ -5,7 +5,6 @@ from uuid import UUID
 
 from django.conf import settings
 from django.db import transaction
-from django.db.models import prefetch_related_objects
 from django.utils import timezone
 from prices import Money, TaxedMoney
 
@@ -19,7 +18,7 @@ from ..core.taxes import (
     zero_taxed_money,
 )
 from ..discount.utils.order import (
-    create_order_discount_objects_for_order_promotions,
+    handle_order_promotion,
     refresh_order_base_prices_and_discounts,
     update_unit_discount_data_on_order_line,
 )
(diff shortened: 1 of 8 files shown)


saleor/order/calculations.py
```diff
@@ -122,6 +139,18 @@ def fetch_order_prices_if_expired(
         return order, lines
 
 
+def get_expired_line_ids(order: Order, lines: Iterable[OrderLine] | None) -> list[UUID]:
+    if order.status != OrderStatus.DRAFT:
+        return []
+
+    if lines is None:
```
korycins, line 146: If you move this condition to `fetch_order_prices_if_expired`, then you could re-use already fetched lines in later part of the code:
```python
    if order.status not in ORDER_EDITABLE_STATUS:
        return order, lines

    if lines is None:
        lines = order.lines.all()

    expired_line_ids = get_expired_line_ids(order, lines)
```
Then rest of the calls will always have lines instead of None

saleor/order/calculations.py
```diff
@@ -57,16 +63,27 @@ def fetch_order_prices_if_expired(
     if order.status not in ORDER_EDITABLE_STATUS:
         return order, lines
 
-    if not force_update and not order.should_refresh_prices:
+    expired_line_ids = get_expired_line_ids(order, lines)
+    if not force_update and not order.should_refresh_prices and not expired_line_ids:
```
korycins, line 67: As right now you need to match two conditions to exit from price recalculation, maybe it would be good for the readability to separate them to two variables? Something like this:

```suggestion
    no_need_to_calculate_taxes = not force_update and not order.should_refresh_prices
    no_need_to_calculate_expired_base = not expired_line_ids
    if no_need_to_calculate_taxes and no_need_to_calculate_expired_base:
```

Right now it is easier to see, that to go inside this `if`, we need to have valid taxes, and valid base_prices.

korycins: Also, for better readability, I would also recommend to rename `force_update` to `force_tax_update`. As this could be misleading, that `force_update` is responsible for re-calculate taxes and recalculate base prices

saleor/order/calculations.py
zedzior: This way you make an extra db call for recalculations of orders with status `UNCONFIRMED`. That is why I moved it here. What do you think?

saleor/order/calculations.py
zedzior: What about something like:
```python
need_recalculation = (
   order.status in ORDER_EDITABLE_STATUS 
   and (force_update or order.should_refresh_prices or expired_line_ids)
   )
if not need_recalculation:
   return order, lines
```

zedzior: https://github.com/saleor/saleor/pull/17428 here is a draft PR with proposal to increase readability of whole `fetch_order_pirces_if_expired` function

saleor/order/utils.py
```diff
@@ -1354,9 +1354,11 @@ def clean_order_line_quantities(order_lines, quantities_for_lines):
             )
 
 
-def get_order_line_price_expiration_date(channel: "Channel") -> datetime | None:
-    freeze_period = channel.draft_order_line_price_freeze_period
-    if freeze_period is not None and freeze_period > 0:
+def get_order_line_price_expiration_date(order: "Order") -> datetime | None:
```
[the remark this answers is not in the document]
zedzior: corrected

saleor/order/utils.py
```diff
@@ -1354,9 +1354,11 @@ def clean_order_line_quantities(order_lines, quantities_for_lines):
...
+    if order.status != OrderStatus.DRAFT:
+        return None
+    freeze_period = order.channel.draft_order_line_price_freeze_period
+    if isinstance(freeze_period, int) and freeze_period > 0:
```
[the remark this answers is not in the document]
zedzior: corrected

korycins (approved):

IKarbowiak (approved):

PR #17420, opened 2025-02-26
Verify if updating the address after checkout complete do not change the address in customer address book

IKarbowiak:
Test ID: 0131

This test verifies that when a shipping address is updated during checkout, the new address is stored in the customer’s address book even if the `SAVE_ADDRESS` flag is not provided, the initial flag is reset.

Steps
1. Create a checkout with an initial shipping and billing address. Set save_address flags to False
2. Update the shipping address, do not provide save address setting.
3. Complete checkout
**Expected**
The shipping address should be saved in customer address book.

The billing address should not be saved in customer address book.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


fowczarek (approved):
IMO we should use descriptive titles. Test ID should be part of the test description.

PR #17423, opened 2025-02-26
Verify if the checkout complete take into consideration address settings

IKarbowiak:
Test ID: CORE_0132

Verifies that when a new shipping address is added during checkout with save_address flag set to True, and billing with save address set to False, the shipping address is stored in the user’s address book, and the billing address isn't.

1. Create checkout with shipping and billing address, save_shipping_address=True, save_billing_address=False
2. Set the delivery method
3. Complete checkout
**Expected**
Ensure the shipping_address is stored in customer address book.
Ensure the billing_address is not stored in customer address book.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/tests/e2e/checkout/test_checkout_complete_respect_saving_address_settings.py
+++ b/saleor/tests/e2e/checkout/test_checkout_complete_respect_saving_address_settings.py
@@ -1,0 +1,144 @@
+import pytest
+
+from ..account.utils import get_own_data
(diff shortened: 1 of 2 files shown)


korycins (approved):

fowczarek (approved):
IMO we should use descriptive titles. Test ID should be part of the test description.

PR #17440, opened 2025-02-28
Separate lint & test jobs

lkostrowski:
- it runs linters and tests in parallel
- it saves extra 2-3 minutes on the entire workflow
- it will cancel both if one of them fails to save time and resources

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,8 +1,8 @@
-name: Pytest
+name: Tests & Linters
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened]
+    types: [ opened, synchronize, reopened ]
     paths:
       - "**.py"
       - Dockerfile
@@ -34,10 +34,10 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  units_and_linters:
+  unit_tests:
     runs-on: ubuntu-latest
     container: python:3.12
-    name: Run units & linters
+    name: Run unit tests
 
     services:
       postgres:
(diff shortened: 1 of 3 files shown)


IKarbowiak (approved):

.github/workflows/pytest.yml
```diff
@@ -135,3 +120,76 @@ jobs:
...
         if: github.event_name == 'push' && github.repository == 'saleor/saleor'
+
+      # Do not continue linters if tests failed earlier
+      - name: Cancel workflow on failure
+        if: failure()
+        uses: vishnudxb/cancel-workflow@c3c77eb4383ba7d023e6614a07d94fe990501ac6 # 1.2
+        with:
+          repo: ${{ github.repository }}
+          workflow_id: ${{ github.run_id }}
+          access_token: ${{ github.token }}
```
NyanKiyoshi, line 131: I don't think it matters to cancel the workflow, it doesn't really bring much value IMO, it might even be confusing the the PR reviewers as it will not show as failed

It will also not work with PRs coming from external contributors due to requiring write permissions (whereas they are read-only)

Suggestion: do nothing, just let the jobs fail by themselves

lkostrowski: @korycins I think you pointed me that previous workflow was written this way to indeed cancel tests if linter fails. I personally agree with @NyanKiyoshi but I don't know what were the previous decisions

korycins: To fix linters you need to push new changes. This is why in the previous flow, we called linters first, then tests (in single job). So, in case of failure on linters, you had faster info about the failure and required action as there was no need to wait for the pytest.

lkostrowski: So this flow will work similarly:
1. You will have separate 3-minutes job for linter (and 15 minutes for pytest)
2. If linters fail, you will immediately see job failed (and you know its lint, no need to check logs) and get the email notification
3. You will push fix, which will restart linters & pytest (already we have such behavior)

So, I assume we have very similar behavior and we can do what @NyanKiyoshi proposed

korycins (approved):

PR #17456, opened 2025-03-05
Add OpenTelemetry

maarcingebala:
Adds support for OpenTelemetry to Saleor Core. Scope of changes includes:
- Replaced OpenTracing with OpenTelemetry.
- Added the ability to define custom metrics, which was not supported with OpenTracing.
- Added `Tracer` and `Meter` wrappers for creating traces and metrics.

Docs: https://github.com/saleor/saleor-docs/pull/1515

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/core/telemetry/metric.py
```diff
@@ -0,0 +1,208 @@
...
+            amount: Value to record
+            unit: Optional unit of the measurement (converted if different from metric unit)
+            attributes: Attributes to record with the measurement
+
+        """
+        attributes = enrich_with_global_attributes(attributes)
+        if metric_name not in self._instruments:
+            raise RuntimeError(f"Metric {metric_name} must be created first")
+        instrument_unit, instrument = self._instruments[metric_name]
+        amount = convert_unit(amount, unit, instrument_unit)
```
korycins, line 134: It might be stupid question, but:
Do you actually need `unit` as an input and this conversion? Do you have any use case where the same metric would call one record in `unit1` and later in `unit2`?
Why not assume, that `meter.record` should be always called with the value defined when creating the metric?

saleor/core/telemetry/utils.py
```diff
@@ -0,0 +1,139 @@
...
+    @property
+    def is_service(self):
+        return self == Scope.SERVICE
+
+
+class Unit(Enum):
+    SECOND = "s"
+    MILLISECOND = "ms"
+    NANOSECOND = "ns"
+    REQUEST = "{request}"
```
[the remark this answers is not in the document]
przlada: > What type of unit is this?

These are the units used when metrics are defined (using our `meter.create_metric` method). They follow the "Unified Code for Units of Measure" ([more details here](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units)).

> Why is it needed here? 

We need them, as it is recommended by OpenTelemetry to assign a unit to a metric while creating it. For convenience, we wrapped it in an Enum (rather than passing plain strings around). The idea is to extend this enum whenever we need to apply a new type of unit.

> Is it request_duration or request_count?

In case of `{request}`, it is a "non-unit" annotation (as defined by the "Unified Code for Units of Measure") which is used to measure request **count**. Duration is time, so to measure it the unit should be seconds (`s`). Here is more context from the OpenTelemetry docs:
> - All non-units that use curly braces to annotate a quantity need to match the grammatical number of the quantity it represent. For example if measuring the number of individual requests to a process the unit would be {request}, not {requests}.
> - Instruments that measure an integer count of something SHOULD only use annotations with curly braces to give additional meaning without the leading default unit (1). For example, use {packet}, {error}, {fault}, etc.


> At first glance, it appears to mix units with unrelated data, which could be confusing. 

I agree, but this is just a utility for grouping unit names from a unified standard. We could possibly split units (`s`, `ms`, `kg`) and non-units (`{request}`, `{packet}`, `{error}`) into separate Enums, but I'm not sure if that would reduce the confusion (as both are used to declare a metric's unit).

Thanks to your comment, I realized that OpenTelemetry now recommends using seconds as the unit when measuring durations (they deprecated standard metrics based on milliseconds), so we should change that for ours. I will dig deeper into this

saleor/graphql/metrics.py
```diff
@@ -0,0 +1,30 @@
+from contextlib import AbstractContextManager
+
+from ..core.telemetry import AttributeValue, MetricType, Scope, Unit, meter
+
+# Initialize metrics
+METRIC_GRAPHQL_QUERIES = meter.create_metric(
+    "saleor.graphql_queries",
+    scope=Scope.SERVICE,
+    type=MetricType.COUNTER,
```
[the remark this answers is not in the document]
przlada: That is totally valid case - you can declare metrics with different combinations of `type` and `unit`. More about metric kinds here: [https://opentelemetry.io/docs/concepts/signals/metrics/#metric-instruments](https://opentelemetry.io/docs/concepts/signals/metrics/#metric-instruments)

COUNTER & MILLISECOND - for example, a metric of total time spent doing something. The MILLISECOND unit represents time which is typically used with type HISTOGRAM as it's better suited to handle "durations". But there's nothing wrong with COUNTER & MILLISECOND metric if you have such a use case.

HISTOGRAM & REQUEST -> same situation as above, it is totally ok to declare such a metric. It's harder to come up with an example (as such combination is less common, histograms are usually used for durations) - but let's say "Distribution of the number of requests processed in a batch".

> or if I will call COUNTER metric with unit NANOSECOND `meter.record(METRIC_GRAPHQL_QUERIES, amount, unit=Unit.NANOSECOND)`?

In this case, you're recording a measurement in nanoseconds for a metric declared with the unit "request". Our idea was to raise an exception in such cases (there's no way of converting nanoseconds to requests).

saleor/core/telemetry/metric.py
```diff
@@ -0,0 +1,208 @@
...
+        return name
+
+    def record(
+        self,
+        metric_name: str,
+        amount: Amount,
+        *,
+        unit: Unit | None = None,
+        attributes: Attributes = None,
+    ) -> None:
```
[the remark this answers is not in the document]
przlada: Our idea here was for it to be a safety mechanism -> raising an exception if there is no conversion rate declared. That can happen when:
- There is just a need to add a missing conversion rate to the table
- There is a bug, and a measurement is created for the wrong metric or with the wrong unit, which can lead to incorrect data being reported and potentially corrupted metric
Thanks to this approach, such scenarios should result in failing tests and CI going red 

But I also get your point - tests should break, but production shouldn't. So I propose to do something similar to `allow_writer()`: change the behavior with a setting that logs warnings by default but raises exceptions in tests. Great insight! Thank you for catching that!

saleor/core/telemetry/metric.py
przlada: It's a totally valid question! OpenTelemetry doesn't do conversions by default, and we only added it as a utility for our convenience. While experimenting with metrics, we quite commonly encounter situations where, for example: we have a histogram metric with milliseconds as the unit and a need to report a measurement in a different unit (seconds, nanoseconds) because it came that way from somewhere (`time.monotonic_ns()`, `datetime.timedelta.total_seconds()`). In such situations, developers are required to check the metric definition for the unit and make the proper conversion before reporting.

That's why we came up with this helper function - to reduce this burden. However, while responding to your question, I realized that for our mechanism to effectively work and secure our telemetry, we should make providing the unit for `meter.record` required rather than optional. Thank you! Excellent reviewer!

saleor/core/telemetry/utils.py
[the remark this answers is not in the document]
korycins: It seems I wasn’t familiar with the OpenTelemetry docs, so the {request} placeholder felt strange and non-intuitive. I wasn’t sure why you didn’t use something like count, especially since {request} could refer to the number of requests, but it might also relate to something else entirely connected to the request.

saleor/core/telemetry/tests/test_metric.py
```diff
@@ -0,0 +1,416 @@
...
+    # when
+    meter_proxy.record(
+        "test_metric",
+        1,
+        unit=Unit.REQUEST,
+        attributes={"attr1": "value1"},
+    )
+
+    # then
+    # should not raise an error, but do nothing
```
kadewu, line 393: It this test really testing anything?

saleor/core/telemetry/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+
+def initialize_telemetry() -> None:
+    """Initialize telemetry components lazily to ensure fork safety in multi-process environments."""
+
+    # To avoid circular imports.
+    from ... import __version__ as saleor_version
+
+    tracer_cls = load_object(settings.TELEMETRY_TRACER_CLASS)
+    if not issubclass(tracer_cls, Tracer):
+        raise ValueError(
```
kadewu, line 35: Can we add tests for those two raises?

kadewu (commented):
Shouldn't we provide PR to the `saleor-platform` as well with those changes?

saleor/core/telemetry/utils.py
```diff
@@ -0,0 +1,139 @@
...
+
+def enrich_with_global_attributes(attributes: Attributes) -> Attributes:
+    return {**(attributes or {}), **get_global_attributes()}
+
+
+@dataclass(frozen=True)
+class TelemetryTaskContext:
+    """Carries telemetry context when propagated to Celery tasks."""
+
+    # TODO add TraceState support
```
IKarbowiak, line 71: Do you have task for that? Maybe worth linking the issue number here

IKarbowiak (approved):

saleor/core/telemetry/metric.py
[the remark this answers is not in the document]
przlada: Fixed with https://github.com/saleor/saleor/pull/17579

saleor/core/telemetry/metric.py
przlada: Fixed with https://github.com/saleor/saleor/pull/17579

saleor/core/telemetry/utils.py
maarcingebala: We're tracking this in Linear, for now I'll skip linking the task here and we can fix it later.

przlada (approved):

kadewu (approved):

PR #17466, opened 2025-03-07
[WIP] Add metric for webhook delivery delay

kadewu:
Metrics to measure performance of webhook delivery

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/webhook/transport/asynchronous/transport.py
+++ b/saleor/webhook/transport/asynchronous/transport.py
@@ -35,6 +35,7 @@ from ....graphql.webhook.subscription_payload import (
 from ....graphql.webhook.subscription_types import WEBHOOK_TYPES_MAP
 from ... import observability
 from ...event_types import WebhookEventAsyncType, WebhookEventSyncType
+from ...metrics import record_first_delivery_attempt_delay
 from ...observability import WebhookData
 from ..utils import (
     DeferredPayloadData,
@@ -603,6 +604,8 @@ def send_webhook_request_async(
         data = data if isinstance(data, bytes) else data.encode("utf-8")
         # Count payload size in bytes.
         payload_size = len(data)
+        if self.request.retries == 0:
+            record_first_delivery_attempt_delay(delivery)
         with webhooks_otel_trace(
             delivery.event_type,
             domain,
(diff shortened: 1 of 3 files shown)


saleor/webhook/transport/utils.py
```diff
@@ -523,6 +551,93 @@ def clear_successful_delivery(delivery: "EventDelivery"):
         delete_files_from_private_storage_task(files_to_delete)
 
 
+@allow_writer()
+def clear_successful_deliveries(deliveries: list["EventDelivery"]):
```
korycins, line 555: We alreayd have a `clear_successful_delivery`. Now we have separate log for single delviery and separate logic for batch of deliveries. The "core" logic should always be the same. Instead of adding new helper function, I would suggest to modify the existing one to support single and list. Like


```python
def clear_successful_deliveries(delvieries: list []):
    # main logic

def clear_successful_delivery(delivery: "EventDelivery"):
    clear_successful_deliveries([delivery])
```

Same questions to `process_failed_deliveries` and `create_attempts_for_deliveries`. It is not new functionality, it is something that we extend from single object to multiple object. In that case, why we add fully new implementation, instead of making the common part for single and multiple?

saleor/webhook/transport/utils.py
```diff
@@ -523,6 +551,93 @@ def clear_successful_delivery(delivery: "EventDelivery"):
...
+    files_to_delete = []
+    for delivery in deliveries:
+        # skip deliveries that cannot be deleted
+        if not delivery.id or delivery.status != EventDeliveryStatus.SUCCESS:
+            continue
+
+        delivery_ids_to_delete.append(delivery.id)
+
+        payload_id = delivery.payload_id
+        if payload_id:
```
tomaszszymanski129, line 567:
```suggestion
        if payload_id := delivery.payload_id:
```

saleor/webhook/transport/asynchronous/transport.py
```diff
@@ -652,6 +671,98 @@ def send_webhook_request_async(
...
+        return
+
+    attempts_for_deliveries = create_attempts_for_deliveries(
+        deliveries, self.request.id
+    )
+    failed_deliveries_attempts = []
+    successful_deliveries = []
+
+    for delivery_id, delivery_with_count in deliveries.items():
+        delivery = delivery_with_count.delivery
```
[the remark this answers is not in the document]
kadewu: I've decided to address that in the current changes that I'm applying to the batch task.

fowczarek (approved):

tomaszszymanski129 (approved):

PR #17470, opened 2025-03-11
Refactor metadata operations

lkostrowski:
- current metadata implementation is limiting and hard to extend to other mutations and objects
- this PR extracts its self-contain abstraction that can be operated on between graphql input and django model

This is a minimal scope of the refactor that allows me to continue working on metadata effectively.

In the future following improvements can be proposed:
- clean_input can automatically parse "dict" input to MetadataItemCollection, so every mutation will receive type-safe, validated model
- More elastic error handling - injecting them from the outside or better configuration. Also more validators can be introduced eventually

It *almost* doesn't have any visible changes. There are only few changes in error code paths - which is rather a proper change - in some responses `error.field` will be `metadata` or `privateMetadata` instead generic "input" value

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/core/utils/metadata_manager.py
lkostrowski, line 1: This file is only addition. Rest of the PR is applying it instead previous behaviors

saleor/core/tests/test_metadata_manager.py
```diff
@@ -0,0 +1,131 @@
+from sympy.testing.pytest import raises
+
+from saleor.core.models import ModelWithMetadata
```
fowczarek, line 3: For code consistency, you should use relative imports.

saleor/core/tests/test_metadata_manager.py
```diff
@@ -0,0 +1,131 @@
+from sympy.testing.pytest import raises
```
fowczarek, line 1: It should be imported directly from `pytest` module. 
In other tests we have:
```python
import pytest

def test_abc():
    with pytest.raises(ValidationError) as e:
        assert False
```

saleor/graphql/payment/mutations/payment/checkout_payment_create.py
```diff
@@ -175,9 +175,10 @@ def validate_return_url(cls, input_data):
                 {"redirect_url": e}, code=PaymentErrorCode.INVALID.value
             ) from e
 
+    # TODO This should be unified with metadata_manager and MetadataItemCollection
```
fowczarek, line 178: Should it be fixed in this PR?

saleor/core/utils/metadata_manager.py
```diff
@@ -0,0 +1,72 @@
...
+    class MetadataItem:
+        key: str
+        value: str
+
+        def __init__(self, key: str, value: str):
+            if not key.strip():
+                raise MetadataEmptyKeyError()
+
+            self.key = key
+            self.value = value
```
IKarbowiak, line 29: Why you nested `MetadataItem` inside `MetadataItemCollection`? IMO those should be independent classes

saleor/core/tests/test_metadata_manager.py
```diff
@@ -0,0 +1,131 @@
...
+
+    assert instance.metadata.get(metadata_list[1].key) == metadata_list[1].value
+
+    # Check if the key with the same value was overwritten
+    assert instance.metadata.get(metadata_list[0].key) == overwritten_value
+
+
+def test_throws_for_invalid_metadata_target():
+    class TestInstance(ModelWithMetadata):
+        pass
```
IKarbowiak, line 122: You are using `TestInstance` multiple times, so can be defined once globally instead in each test

saleor/graphql/payment/mutations/payment/checkout_payment_create.py
lkostrowski: No

saleor/core/utils/metadata_manager.py
lkostrowski: Agree, it was some part of discussions with @korycins , not sure where we standed. I want to keep MetadataItem class to ensure it's typed. We did have some discussions about keeping 2 classes (which I think is fine)

@korycins WDYT to extract this MetadataItem outside parent class?

saleor/core/tests/test_metadata_manager.py
lkostrowski: Fixing + opening a discussion about adding this to linter (https://docs.astral.sh/ruff/rules/relative-imports/) and hopefully changing to absolutes ;)

saleor/core/utils/metadata_manager.py
```diff
@@ -0,0 +1,72 @@
...
+                "Unknown argument, provide MetadataType.PRIVATE or MetadataType.PUBLIC"
+            )
+
+
+def create_from_graphql_input(
+    items: list[MetadataInput] | None,
+) -> MetadataItemCollection:
+    """Use with care.
+
+    This method is eventually returning custom error, so if it's used directly
```
korycins, line 62: OR
```suggestion
    This method is eventually raising custom error, so if it's used directly
```

(returning suggests `return SomeCustomError`. Your logic makes error `raise`)

saleor/core/utils/metadata_manager.py
```diff
@@ -0,0 +1,72 @@
...
+        case _:
+            raise ValueError(
+                "Unknown argument, provide MetadataType.PRIVATE or MetadataType.PUBLIC"
+            )
+
+
+def create_from_graphql_input(
+    items: list[MetadataInput] | None,
+) -> MetadataItemCollection:
+    """Use with care.
```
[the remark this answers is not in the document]
korycins: Or maybe something like: `Converts graphql metadata input into dataclass metadata structure.`

saleor/graphql/account/bulk_mutations/customer_bulk_update.py
```diff
@@ -360,8 +365,14 @@ def _get_customer(cls, customer_id, external_ref):
...
         address = getattr(instance, field) or models.Address()
-        address_metadata = data.pop("metadata", [])
-        cls.update_metadata(address, address_metadata)
+        address_metadata: list[MetadataInput] = data.pop("metadata", [])
+
+        metadata_collection = cls.create_metadata_from_graphql_input(
+            address_metadata, error_field_name="metadata"
+        )
+
+        store_on_instance(metadata_collection, address, MetadataType.PUBLIC)
```
korycins, line 374: Maybe we could import the file instead of methods? This could be more meaningful, like:
`metadata_manager.store_on_instance()`
or `store_on_instance` -> `store_metadata_on_instance`

saleor/graphql/core/mutations.py
```diff
@@ -609,6 +594,29 @@ def check_channel_permissions(
                 message="You don't have access to some objects' channel."
             )
 
+    @classmethod
+    def create_metadata_from_graphql_input(
+        cls, metadata_list: list[MetadataInput] | None, *, error_field_name: str
+    ) -> MetadataItemCollection:
+        """Wrap the creation of metadata and map the error to graphql error.
+
+        In case of metadata - we need to pass field name, because it can be nested in the other path than "metadata" or "privateMetadata"
```
korycins, line 603:
```suggestion
        In case of metadata - we need to pass `error_field_name`, because it can be nested in the other path than "metadata" or "privateMetadata"
```
It would be more meaningful to use exact name of the parameter that you're writing about

saleor/graphql/core/mutations.py
```diff
@@ -609,6 +594,29 @@ def check_channel_permissions(
                 message="You don't have access to some objects' channel."
             )
 
+    @classmethod
+    def create_metadata_from_graphql_input(
+        cls, metadata_list: list[MetadataInput] | None, *, error_field_name: str
+    ) -> MetadataItemCollection:
+        """Wrap the creation of metadata and map the error to graphql error.
```
korycins, line 601: > and map the error to graphql error.

IMHO, for me more important information here would be: this method can raise ValidationError. The conversion from metadata's error  is more like implementation detail.

saleor/core/tests/test_metadata_manager.py
```diff
@@ -0,0 +1,126 @@
...
+
+invalid_metadata_item = MetadataInput()
+
+# Key can't be empty
+invalid_metadata_item.key = ""
+invalid_metadata_item.value = "value"
+
+valid_list = [valid_metadata_item]
+invalid_list = [invalid_metadata_item]
+invalid_list_with_one_valid = [
```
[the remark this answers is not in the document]
lkostrowski: Agree, I will wrap this with functions to create instance instead sharing objects

saleor/graphql/payment/mutations/payment/checkout_payment_create.py
fowczarek: If we have a separate task to fix maybe it would be a good idea also to write down his ID next to this comment?

saleor/core/tests/test_metadata_manager.py
lkostrowski: I tried, but moving this class to test body fails the tests - they need database migration just for that. I dont think it's worth it. Having it instead inside the test body doesn't raise any issues

saleor/graphql/account/bulk_mutations/customer_bulk_update.py
lkostrowski: done

saleor/core/utils/metadata_manager.py
lkostrowski: done

saleor/graphql/core/mutations.py
lkostrowski: done

fowczarek (dismissed):

korycins (approved):

fowczarek (approved):

PR #17471, opened 2025-03-11
Improve order status logic for fulfillments waiting for approval

delemeator:
I want to merge this change because I have came across strange behavior for WAITING_FOR_APPROVAL fulfillments. Consider this example from recent 3.21 version (saleor git hash 4d0ab24):
```
{
  "data": {
    "order": {
      "fulfillments": [
        {
          "status": "WAITING_FOR_APPROVAL",
          "lines": [
            {
              "quantity": 1
            }
          ]
        }
      ],
      "lines": [
        {
          "quantityFulfilled": 1,
          "quantityToFulfill": 0
        }
      ],
      "status": "PARTIALLY_FULFILLED"
    }
  }
}
```

Here we have an order with one WAITING_FOR_APPROVAL fulfillment and the status for entire order is PARTIALLY_FULFILLED which is counter intuitive. This PR updates this logic so that:
- If **all** fulfillments are WAITING_FOR_APPROVAL then order is UNFULFILLED
- If **some**  fulfillments are WAITING_FOR_APPROVAL then order is PARTIALLY_FULFILLED

EDIT: I have also added a fix to prevent from overriding UNCONFIRMED status when WAITING_FOR_APPROVAL fulfillment is created. 

# Impact

No migrations or schema changes.

Needed to update one test. It was having **empty** waiting for approval fulfillment. Under new logic it doesn't affect anything. Also I suppose that empty waiting for approval fulfillment shouldn't really exist? 

# Docs

Update for migration guide: https://github.com/saleor/saleor-docs/pull/1498

# Pull Request Checklist


korycins (commented):
Thanks for the PR! 🎉

- It would be great to include information about this change in the [Upgrade Guide (3.20 to 3.21)](https://docs.saleor.io/upgrade-guides/core/3-20-to-3-21).

- Additionally, a migration to correct the status for existing orders with awaiting-approve fulfillment would be needed. Let us know if you're able to handle this - otherwise, we can take care of it in a separate PR. 🙂

delemeator:
> Thanks for the PR! 🎉
> 
>     * It would be great to include information about this change in the [Upgrade Guide (3.20 to 3.21)](https://docs.saleor.io/upgrade-guides/core/3-20-to-3-21).
> 
>     * Additionally, a migration to correct the status for existing orders with awaiting-approve fulfillment would be needed. Let us know if you're able to handle this - otherwise, we can take care of it in a separate PR. 🙂

Hey @korycins - this migration sounds fun, but if you could take care of it, it would be great ;) 
I will take care of migration guide

korycins (approved):

saleor/order/utils.py
```diff
@@ -176,15 +177,22 @@ def _calculate_quantity_including_returns(order):
...
+    return (
+        total_quantity,
+        quantity_fulfilled,
+        quantity_returned,
+        quantity_awaiting_approval,
+    )
 
 
-def update_order_status(order: Order):
+def update_order_status(order: Order, confirm: bool = False):
```
korycins, line 195: Maybe we could be more explicit with the name of `confirm` variable? It looks like it is related only to the unconfirmed orders. It could also be misleading, as we use `confrim` word for orders with status != [draft, unconfirmed]. 
Maybe something like `keep_unconfirmed`/`update_unconfirmed`, or something different and more meaningful?

saleor/order/actions.py
```diff
@@ -392,7 +392,7 @@ def handle_fully_paid_order(
         )
 
     if not order.is_draft() and order.channel.automatically_confirm_all_new_orders:
-        update_order_status(order)
+        update_order_status(order, confirm=True)
```
korycins, line 395: Actually, I am not sure if this is correct way.  We have a flow, where you can create unpaid order. unpaid order is always marked as unconfirmed (`saleor.checkout.complete_checkout._create_order_from_checkout`):
```python
# status
status = (
    OrderStatus.UNFULFILLED
    if (
        checkout_info.channel.automatically_confirm_all_new_orders
        and checkout_info.checkout.payment_transactions.exists()
    )
    else OrderStatus.UNCONFIRMED
)
# I am aware that the `checkout_info.checkout.payment_transactions.exists()` should 
# be fixed as in specific case this could not be 
# correct - will create the separate task for this)
```

In that case, we will have unpaid order with unconfirmed status. Previously, when receiving the payment and `automatically_confirm_all_new_orders==True`, we were converting the order to "confirmed" status.
Right now, this will stop working, as you block the conversion from `unconfirmed` to `unfulfilled`.

saleor/order/utils.py
```diff
@@ -195,19 +203,16 @@ def update_order_status(order: Order):
...
-
         status = determine_order_status(
             order,
             total_quantity,
             quantity_fulfilled,
             quantity_returned,
-            awaiting_approval,
+            quantity_awaiting_approval,
+            confirm,
         )
```
IKarbowiak, line 216: Maybe instead of checking `confirm` flag and `order.is_unconfirmed()` condition in  `determine_order_status` function, you can quit the method earlier. I mean:

```suggestion
        if not confirm and order.is_unconfirmed():
            return
     
        status = determine_order_status(
            order,
            total_quantity,
            quantity_fulfilled,
            quantity_returned,
            quantity_awaiting_approval,
        )
```
As in such case we do not want to change the status at all, right?

Actually, we can even extend the condtiton with `total_quantity == 0` check.
```suggestion
        all_products_replaced = total_quantity == 0
        skip_update_for_unconfirmed = not confirm and order.is_unconfirmed()
        if skip_update_for_unconfirmed or all_products_replaced:
            return
     
        status = determine_order_status(
            order,
            total_quantity,
            quantity_fulfilled,
            quantity_returned,
            quantity_awaiting_approval,
        )
```

saleor/order/actions.py
IKarbowiak: In which scenario exactly? Here the `confirm` flag is set to `True`, which mean that the status will be updated to `UNFULFILLED`.

saleor/order/utils.py
delemeator: As you prefer. I actually did something similar initially, but then decided against it to keep all the conditions in determine_order_status function. My reasoning was that in such case, it's easier to understand what is happening and why. But if you prefer this way, I don't object :)

delemeator: @IKarbowiak When I do this change, maybe I will incorporate „automatically_confirm_all_new_orders” check in „determine_order_status”? Then we don’t have extra attribute and we can remove copy pasted check from 3 different ifs

saleor/order/actions.py
korycins: Ohh, my bad! You're right, I messed up with the `confirm` value.  (I was before my first coffee - this is my excuse 😅 )

delemeator:
Thank you for the feedback - I have implemented the changes

delemeator:
Hey @korycins: 
- I would opt to raise error in graphql if someone is trying to confirm fulfillment on unconfirmed order. 

Let me briefly describe our setup:
- We don't use automatic confirmation of neither order nor fulfillment
- We use transaction flow with separate authorization and charge 

Currently when we get an order:
- order management app creates draft/unaproved fulfillment 
- in the meantime payment app is doing charging
- after several minutes order gets confirmed from ERP (then orderConfirm will be called, but Saleor is always faster to confirm the order by itself ;) )

When all of this happens, something is confirming the order. And I am not sure what (with suspicion around transaction or draft fulfillment or some combination of both). I had some theories, but then I was finding some counter example in our test orders. 

I think it should be explicit when order is confirmed.

delemeator:
Alternatively, I can add flag `update_unconfirmed` in two more places. So that creating approved or approving draft fulfillment confirms order. From my perspective error less surprising (actually I thought it is implemented), but for my use-case it doesn't matter (error or automatic order confirmation for approved fulfillment)

korycins:
> Saleor is always faster to confirm the order by itself

Is it something that always happens? Does it also happens on this branch?
I tried to reproduce it with:
- turned off auto confirm&auto fulfillment
- order from checkout with transaction: AUTHORIZATION_SUCCESS
- create awaiting-fulfillment for single quantity
- call `transactionEventReport` with CHARGE_SUCCESS

And it always keeps the order in UNCONFIRMED status 🤔 - (same results when I add charge-success first, and awaiting fulfillment later)

I see that on `main` when calling `CANCEL-FULFILLMENT` mutation for awaiting-fulfillment, the order status is converted to PARTIALLY_FULFILLED. This is wrong imo, but on your branch it keeps the UNCONFIRMED.


> Alternatively, I can add flag update_unconfirmed in two more places. So that creating approved or approving draft fulfillment confirms order. IMHO error is less surprising (actually I thought it is implemented), but for my use-case it doesn't matter (error or automatic order confirmation for approved fulfillment)

Personally, I would say that adding confirmed fulfillment should mark order as confirmed, but let me double check this with the team.

delemeator:
No, I mean on `main`. But I think you might be right, when I look on my test cases CANCEL-FULFILLMENT looks like it might be it (on unapproved fufillment). 

Then I will add the `confirm_unconfirmed` flag to places where fulfillment is approved (or created as approved). And I will also mention this in migration docs, so that there is definitive list of cases when order gets approved

korycins:
> But I think you might be right, when I look on my test cases, CANCEL-FULFILLMENT looks like a valid hypothesis (on unapproved fulfillment).
Then I will add the update_unconfirmed flag to places where fulfillment is approved (or created as approved). And I will also mention this in migration docs, so that there is definitive list of cases when order gets approved

As you changed the way how we treat `WAITING_FOR_APPROVAL` fulfillments. Now such fulfillments don't have an impact on order's status (previously they forced `PARTIALLY_FULLFILLMENT`).
Maybe in that case, we could fully drop `update_unconfirmed` flag? I mean, on `main` the `cancel-waiting-fulfillment` was the only way for me to force `UNCONFIRMED` order to be converted into `FULFILLED`.

`WAITING_FOR_APPROVAL` do not trigger `update_order_status`, so `saleor.order.actions.cancel_waiting_fulfillment` should not call `update_order_status` also. Removing it from `cancel_waiting_fulfillment` should solve your problem (I guess 🤔 )

delemeator:
I have removed short lived `update_unconfirmed` flag and removed `update_order_status` from `cancel_waiting_fulfillment`. Also added a test that checks if cancelling waiting fulfillment changes `UNCONFIRMED` status (to avoid potential regression)

I think it should be good to go now

korycins (dismissed):
Looks good!
As always is it a pleasure to work with you! ❤️

IKarbowiak (approved):

korycins (approved):

PR #17480, opened 2025-03-13
fix webhook emitted from orderUpdateShipping

lkostrowski:
I want to merge this change because it fixed webhook emitted from `orderUpdateShipping` to be DRAFT_ORDER_UPDATED if order was draft

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/order/mutations/order_update_shipping.py
+++ b/saleor/graphql/order/mutations/order_update_shipping.py
@@ -7,6 +7,7 @@ from ....order.error_codes import OrderErrorCode
 from ....permission.enums import OrderPermissions
 from ....shipping import models as shipping_models
 from ....shipping.utils import convert_to_shipping_method_data
+from ....webhook.deprecated_event_types import WebhookEventType
 from ....webhook.event_types import WebhookEventAsyncType
 from ...core import ResolveInfo
 from ...core.context import SyncWebhookControlContext
@@ -74,8 +75,15 @@ class OrderUpdateShipping(
                 "lines", "channel__shipping_method_listings"
             ),
         )
+
+        # Below this Order is properly typed Order, instead of Model
+        if not isinstance(order, models.Order):
+            raise Exception("Received node is not Order")
+
         cls.check_channel_permissions(info, [order.channel_id])
         cls.validate_order(order)
+
+        is_draft_order = order.is_draft()
 
         if "shipping_method" not in input:
             raise ValidationError(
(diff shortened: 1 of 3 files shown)


saleor/graphql/order/mutations/order_update_shipping.py
```diff
@@ -74,9 +75,16 @@ def perform_mutation(  # type: ignore[override]
                 "lines", "channel__shipping_method_listings"
             ),
         )
+
+        # Below this Order is properly typed Order, instead of Model
+        if not isinstance(order, models.Order):
+            raise Exception("Received node is not Order")
```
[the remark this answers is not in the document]
lkostrowski: I will do that, but this is not the same thing

```
def cast(typ, val):
    """Cast a value to a type.

    This returns the value unchanged.  To the type checker this
    signals that the return value has the designated type, but at
    runtime we intentionally don't check anything (we want this
    to be as fast as possible).
    """
    return val
```

which means it solves the typing issue, but it's not  "defensive", so it will not actually check instance.

However, I believe we trust the values here and type casting will be enough. Thanks 👍

IKarbowiak (approved):

saleor/graphql/order/mutations/order_update_shipping.py
```diff
@@ -137,5 +145,13 @@ def perform_mutation(  # type: ignore[override]
 
         order.save(update_fields=SHIPPING_METHOD_UPDATE_FIELDS)
         # Post-process the results
-        call_order_event(manager, WebhookEventAsyncType.ORDER_UPDATED, order)
+
+        event_to_emit = (
+            WebhookEventAsyncType.DRAFT_ORDER_UPDATED
```
korycins, line 150: You should also emit the event for the `return` from the line 122

saleor/graphql/order/tests/mutations/test_order_update_shipping.py
```diff
@@ -705,3 +705,60 @@ def test_order_update_shipping_triggers_webhooks(
...
+    permission_group_manage_orders.user_set.add(staff_api_client.user)
+    order = order_with_lines
+    order.base_shipping_price = zero_money(order.currency)
+    order.status = OrderStatus.DRAFT
+    order.save()
+
+    query = ORDER_UPDATE_SHIPPING_QUERY
+    order_id = graphene.Node.to_global_id("Order", order.id)
+    method_id = graphene.Node.to_global_id("ShippingMethod", shipping_method.id)
+    variables = {"order": order_id, "shippingMethod": method_id}
```
korycins, line 747: Would be good to also have a test when you call with  `shippingMethod == null`

saleor/graphql/order/mutations/order_update_shipping.py
lkostrowski: as discussed - it was out of the initial scope, but added + incldued in changelog as a separate change

korycins (approved):

IKarbowiak (approved):

PR #17499, opened 2025-03-19
Harden admin email plugin config

cmiacz:
I want to merge this change to avoid mixing user-provided admin email config with default settings from env variables.
Currently all null values in user-provided config are getting overwritten by defaults.
From now defaults are going to be used only when none of required fields are provided.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/plugins/admin_email/tests/test_plugin.py
+++ b/saleor/plugins/admin_email/tests/test_plugin.py
@@ -11,6 +11,7 @@ from ....graphql.tests.utils import get_graphql_content
 from ...email_common import (
     DEFAULT_EMAIL_CONFIG_STRUCTURE,
     DEFAULT_EMAIL_VALUE,
+    EmailConfig,
     get_email_template,
 )
 from ...manager import get_plugins_manager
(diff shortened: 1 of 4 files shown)


NyanKiyoshi (dismissed):

patrys:
Can we have a test for the config behavior?

IKarbowiak (approved):

saleor/plugins/admin_email/tests/test_plugin.py
```diff
@@ -324,3 +325,60 @@ def test_plugin_dont_change_default_help_text_config_value():
...
+    data_to_save = {
+        "configuration": [
+            {"name": "host", "value": "localhost"},
+            {"name": "port", "value": "1025"},
+            {"name": "sender_address", "value": "noreply@exmaple.com"},
+        ]
+    }
+
+    plugin.save_plugin_configuration(plugin_configuration, data_to_save)
+    mocked_validate_email_config.assert_called_once_with(
```
NyanKiyoshi, line 373: We should add a comment explaining what the test is checking otherwise it will make it difficult to maintain:

```suggestion
    plugin.save_plugin_configuration(plugin_configuration, data_to_save)

    # When the user changes the host and/or port number, we the default
    # configuration from settings.EMAIL_URL shouldn't be passed, such as
    # the password, username, etc.
    mocked_validate_email_config.assert_called_once_with(
```

saleor/plugins/admin_email/tests/test_plugin.py
NyanKiyoshi, line 1: We are missing a test case that ensures Django doesn't fall back to the `settings.EMAIL_*` inside `EmailBackend` (Saleor could break the logic from this PR accidentally, or Django could change the behavior), specifically these lines: https://github.com/django/django/blob/afbb8c709d40e77b3f71c152d363c5ad95ceec2d/django/core/mail/backends/smtp.py#L36-L37

NyanKiyoshi (approved):

saleor/plugins/admin_email/tests/test_plugin.py
```diff
@@ -324,3 +325,71 @@ def test_plugin_dont_change_default_help_text_config_value():
...
+        username="",  # empty as there's no username in data_to_save
+        password="",  # empty as there's no password in data_to_save
+        use_tls=False,
+        use_ssl=False,
+    )
+
+    plugin.save_plugin_configuration(plugin_configuration, data_to_save)
+    mocked_validate_email_config.assert_called_once_with(expected_config)
+
+    # Assert that Django's EmailBackend does not override empty username and password
```
wcislo-saleor, line 385: These asserts below look like they could be a separate test.

saleor/plugins/admin_email/tests/test_plugin.py
```diff
@@ -324,3 +325,71 @@ def test_plugin_dont_change_default_help_text_config_value():
...
+        sender_name="",
+        sender_address="noreply@exmaple.com",
+        username="",  # empty as there's no username in data_to_save
+        password="",  # empty as there's no password in data_to_save
+        use_tls=False,
+        use_ssl=False,
+    )
+
+    plugin.save_plugin_configuration(plugin_configuration, data_to_save)
+    mocked_validate_email_config.assert_called_once_with(expected_config)
```
wcislo-saleor, line 383: Maybe it's a minor thing but in fact we're not testing whether particular user-defined config was **saved** here, we only check what was **validated**. How about adding assert against `plugin.config`?

wcislo-saleor (approved):

PR #17501, opened 2025-03-20
Call `DRAFT_ORDER_CREATED` webhhok in `FulfillmentReturnProducts` in case the replace order is created

IKarbowiak:
In case the `DraftOrder` is created as a result of `FulfillmentReturnProducts`mutation the `DRAFT_ORDER_CREATED` will be called.
The source of the draft order creation can be check in `order.origin` field. The `REISSUE` value suggest that order has be created as a replacement.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/order/actions.py
+++ b/saleor/order/actions.py
@@ -1957,6 +1957,10 @@ def create_fulfillments_for_returned_products(
         ).delete()
 
         call_order_event(manager, WebhookEventAsyncType.ORDER_UPDATED, order)
+        if new_order:
+            call_order_event(
+                manager, WebhookEventAsyncType.DRAFT_ORDER_CREATED, new_order
+            )
     return return_fulfillment, replace_fulfillment, new_order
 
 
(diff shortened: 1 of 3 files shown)


saleor/order/actions.py
```diff
@@ -1957,6 +1957,10 @@ def create_fulfillments_for_returned_products(
         ).delete()
 
         call_order_event(manager, WebhookEventAsyncType.ORDER_UPDATED, order)
+        if new_order:
+            call_order_event(
```
korycins, line 1961: Sugestion:
Maybe we could create `draft_order_created()` handler, and call it here, and in  `draftOrderCreate`. 
This could give us confirmation, that we have a single place of logic that should be called for created draft order.

korycins (approved):

fowczarek (approved):

PR #17503, opened 2025-03-20
Add metadata to CheckoutCreateInput

lkostrowski:
I want to merge this change because it allows to set metadata and privateMetadata when checkout is created by checkoutCreate

It also introduces a small refactor - extracted "MetadataManager" to keep utils connected to metadata & private_metadata together.

Old code (base mutations) still contain methods for metadata, but these methods use MetadataManager now.
New code, can directly use MetadataManager, to avoid relying on inherited behavior.

For example, Checkout metadata is not saved on Checkout model, but CheckoutCreate mutation inherits the behavior, not matter what.

Now, CheckoutCreate can use MetadataManager under control, while the rest of the code is not changed.

- [ ] Benchmark test failing

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation: Docs will automatically reflect built reference

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/mutations/checkout_create.py
+++ b/saleor/graphql/checkout/mutations/checkout_create.py
@@ -334,6 +334,10 @@ class CheckoutCreate(DeprecatedModelMutation, I18nMixin):
     def clean_input(cls, info: ResolveInfo, instance: models.Checkout, data, **kwargs):
         user = info.context.user
         channel = data.pop("channel")
+
+        metadata_list: list[MetadataInput] = data.pop("metadata", None)
+        private_metadata_list: list[MetadataInput] = data.pop("private_metadata", None)
+
         cleaned_input = super().clean_input(info, instance, data, **kwargs)
 
         can_set_private_metadata = cls.check_permissions(
@@ -344,12 +348,21 @@ class CheckoutCreate(DeprecatedModelMutation, I18nMixin):
             ],
         )
 
-        trying_to_set_private_metadata = cleaned_input.get("private_metadata")
+        trying_to_set_private_metadata = private_metadata_list is not None
 
         if trying_to_set_private_metadata and (not can_set_private_metadata):
             raise PermissionDenied(
                 "You need MANAGE_CHECKOUTS or HANDLE_CHECKOUTS permission to set privateMetadata"
             )
+
+        cleaned_input["metadata_collection"] = cls.create_metadata_from_graphql_input(
+            metadata_list, error_field_name="metadata"
+        )
+        cleaned_input["private_metadata_collection"] = (
+            cls.create_metadata_from_graphql_input(
+                private_metadata_list, error_field_name="private_metadata"
+            )
+        )
 
         cleaned_input["channel"] = channel
         cleaned_input["currency"] = channel.currency_code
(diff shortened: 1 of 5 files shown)


saleor/graphql/checkout/mutations/checkout_create.py
```diff
@@ -308,8 +336,37 @@ def retrieve_billing_address(
     def clean_input(cls, info: ResolveInfo, instance: models.Checkout, data, **kwargs):
         user = info.context.user
         channel = data.pop("channel")
+
+        metadata_list: list[MetadataInput] = data.pop("metadata", None)
+        private_metadata_list: list[MetadataInput] = data.pop("private_metadata", None)
```
IKarbowiak, line 341: The types of those fields are invalid. In case the `metadata` or `private_metadata` are not present in `data`, the `None` will be assigned, so the type of those fields is actually `None | list[MetadataInput]`.

IKarbowiak (approved):

saleor/graphql/checkout/mutations/checkout_create.py
```diff
@@ -180,6 +182,26 @@ class CheckoutCreateInput(BaseInputObjectType):
...
+        ),
+        required=False,
+    )
+
+    private_metadata = NonNullList(
+        MetadataInput,
+        description=(
+            "Checkout private metadata. Requires one of the following permissions:"
+            f"{CheckoutPermissions.MANAGE_CHECKOUTS.name}, "
+            f"{CheckoutPermissions.HANDLE_CHECKOUTS.name} {ADDED_IN_321}"
```
korycins, line 199:
```suggestion
            f"{CheckoutPermissions.HANDLE_CHECKOUTS.name} {ADDED_IN_321} "
```
See the schema. There is missing space between ADDED_IN and MetadataInputDescription

saleor/graphql/checkout/mutations/checkout_create.py
```diff
@@ -180,6 +182,26 @@ class CheckoutCreateInput(BaseInputObjectType):
         description=("The checkout validation rules that can be changed."),
     )
 
+    metadata = NonNullList(
+        MetadataInput,
+        description=(
+            f"Checkout public metadata. {ADDED_IN_321}"
```
korycins, line 188:
```suggestion
            f"Checkout public metadata. {ADDED_IN_321} "
```

saleor/graphql/checkout/mutations/checkout_create.py
lkostrowski: done

saleor/graphql/checkout/tests/mutations/test_checkout_create.py
```diff
@@ -2910,3 +2910,211 @@ def test_checkout_create_no_billing_address_providing_save_address_raising_error
...
+            "metadata": [{"key": metadata_key, "value": metadata_value}],
+            # Make it more diverse, set 2 keys
+            "privateMetadata": [
+                {"key": metadata_key, "value": metadata_value},
+                {"key": metadata_key_2, "value": metadata_value},
+            ],
+        }
+    }
+
+    assert not Checkout.objects.exists()
```
fowczarek, line 2939: I'm unsure if we should check if the database is empty in each test. IMO empty database is a default state that we could assume as a starting point for each test.

lkostrowski: For me it's fine because it additionally ensures that our initial state is valid

But it's out of my scope- every test in this file does it, so probably it's separate discussion either we need this or should remove it

fowczarek: Fully agreed. I just thinking about the test performance problem. Each optional database operation in the test increases execution time. We should clear those tests in separate PR.

IKarbowiak (approved):

PR #17523, opened 2025-03-24
Metadata support in CheckoutLinesUpdate

lkostrowski:
I want to merge this change because it allows to update metadata in CheckoutLinesUpdate

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/graphql/checkout/mutations/checkout_lines_update.py
```diff
@@ -55,6 +56,15 @@ class CheckoutLineUpdateInput(BaseInputObjectType):
         required=False,
     )
 
+    metadata = NonNullList(
+        MetadataInput,
+        description=(
+            f"Checkout line public metadata. Will add and update keys. To delete keys use deleteMetadata mutation. {ADDED_IN_321} "
```
korycins, line 62: > Will add and update keys

To keep it consistent, shouldn't we update all metadata input descriptions? Right now this can be misleading. For example, here we explicitly say that this is `add and update`, but we don't say that in `orderUpdate` mutation. This can suggest, that `checkoutLinesUpdate` works in a different way than the rest of the metadata inputs

korycins: > To keep it consistent, shouldn't we update all metadata input descriptions?

(or just create a task, I am not suggesting to extend the scope)

lkostrowski: I will create a task, I think we need to scan a code - and I believe it's still better to leave better description in one, than to be consistent but not saying enough

lkostrowski: https://linear.app/saleor/issue/EXT-2072

lkostrowski:
Good catch @korycins , the implementation for validation was missing. It's just too easy to forget about it (hopefully we can eventually enrich graphql layer for validation like this)

saleor/graphql/checkout/mutations/checkout_lines_add.py
```diff
@@ -185,6 +188,21 @@ def clean_input(
...
+                metadata_manager.create_from_graphql_input(line.metadata)
+        except metadata_manager.MetadataEmptyKeyError:
+            raise ValidationError(
+                {
+                    "metadata": ValidationError(
+                        "Metadata key cannot be empty.",
+                        code=MetadataErrorCode.REQUIRED.value,
+                    )
+                }
+            ) from None
```
korycins, line 204:
```suggestion
            )
```

saleor/graphql/checkout/tests/mutations/test_checkout_lines_update.py
```diff
@@ -1622,3 +1622,181 @@ def add_variants_to_checkout_wrapper(*args, **kwargs):
...
     assert data["errors"][0]["field"] == "lineId"
     assert data["errors"][0]["code"] == CheckoutErrorCode.GRAPHQL_ERROR.name
+
+
+def test_checkout_lines_update_with_metadata(
+    user_api_client,
+    checkout_with_item,
+):
+    checkout = checkout_with_item
+    lines, _ = fetch_checkout_lines(checkout)
```
korycins, line 1632: I don't think that you need usage of `fetch_checkout_lines` in your test. It is just a wrapper on the `CheckoutLine`. 
You just need to confirm if the metadata was set. So you can just make `checkout.lines.first()` (as you're already doing). You can fully drop the `fetch_checkout_lines` from your tests - actually, you don't do anything with them.

saleor/graphql/checkout/tests/mutations/test_checkout_lines_update.py
```diff
@@ -1622,3 +1622,181 @@ def add_variants_to_checkout_wrapper(*args, **kwargs):
     assert len(data["errors"]) == 1
     assert data["errors"][0]["field"] == "lineId"
     assert data["errors"][0]["code"] == CheckoutErrorCode.GRAPHQL_ERROR.name
+
+
+def test_checkout_lines_update_with_metadata(
```
korycins, line 1627: Also for better readability and separating part of the test, I would suggest aligning with our contributing guide: https://github.com/saleor/saleor/blob/main/CONTRIBUTING.md#given-when-then.
It's a small change, but it significantly improves the readability of tests.

fowczarek (approved):

saleor/graphql/checkout/mutations/checkout_lines_add.py
lkostrowski: why?
IIRC
- it's a good practice, to be explicit we don't leak the cause?
- We have ruff rule that is enforcing that

fowczarek (approved):

IKarbowiak (approved):

saleor/graphql/checkout/mutations/utils.py
```diff
@@ -472,6 +473,11 @@ def group_lines_input_data_on_update(
             line_data.custom_price = line["price"]
             line_data.custom_price_to_update = True
 
+        if line_data.metadata_list and metadata_list:
+            line_data.metadata_list += metadata_list
```
korycins, line 477: Codecov says that this if is not tested. From what I see, we have some tests for validating the input, adding a new field, overwriting the field, but this report could be valid, as I don't see a test, where we have a key-1 in meta, and we adding key-2.

PR #17534, opened 2025-03-25
Refactor: BaseCustomerCreate inheritance

lkostrowski:
I want to merge this change because it refactors derived classes from BaseCustomerCreate

Previously base class containes "save" method that needs to know about what is it's child classes, checking if it's either creation or updating operation

Now each child class has it's own `save` method implementation that is not shared

This change is also needed for upcoming fixes in these mutations

- [x] Verify if save() is called once
- [x] Refactor address management to shared logic

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/account/mutations/account/account_update.py
+++ b/saleor/graphql/account/mutations/account/account_update.py
@@ -17,7 +17,7 @@ from ....meta.inputs import MetadataInput, MetadataInputDescription
 from ....plugins.dataloaders import get_plugin_manager_promise
 from ...mixins import AppImpersonateMixin
 from ...types import AddressInput, User
-from ..base import BILLING_ADDRESS_FIELD, SHIPPING_ADDRESS_FIELD, BaseCustomerCreate
+from ..base import BaseCustomerCreate
 from .base import AccountBaseInput
 
 
@@ -98,21 +98,18 @@ class AccountUpdate(AddressMetadataMixin, BaseCustomerCreate, AppImpersonateMixi
     @classmethod
     @traced_atomic_transaction()
     def save(cls, info: ResolveInfo, instance: models.User, cleaned_input):
-        default_shipping_address = cleaned_input.get(SHIPPING_ADDRESS_FIELD)
         manager = get_plugin_manager_promise(info.context).get()
-        if default_shipping_address:
-            default_shipping_address.save()
-            instance.default_shipping_address = default_shipping_address
-        default_billing_address = cleaned_input.get(BILLING_ADDRESS_FIELD)
-        if default_billing_address:
-            default_billing_address.save()
-            instance.default_billing_address = default_billing_address
+
+        default_addresses = cls.save_and_apply_addresses_from_input(
+            cleaned_input=cleaned_input, user_instance=instance
+        )
 
         super().save(info, instance, cleaned_input)
-        if default_billing_address:
-            instance.addresses.add(default_billing_address)
-        if default_shipping_address:
-            instance.addresses.add(default_shipping_address)
+
+        if default_addresses["billing"]:
+            instance.addresses.add(default_addresses["billing"])
+        if default_addresses["shipping"]:
+            instance.addresses.add(default_addresses["shipping"])
 
         instance.search_document = prepare_user_search_document_value(instance)
         instance.save(update_fields=["search_document", "updated_at"])
--- a/saleor/graphql/account/mutations/account/account_update.py
+++ b/saleor/graphql/account/mutations/account/account_update.py
@@ -13,7 +13,7 @@ from ....core.descriptions import ADDED_IN_319
 from ....core.doc_category import DOC_CATEGORY_USERS
 from ....core.types import AccountError, NonNullList
 from ....core.utils import WebhookEventInfo
-from ....meta.inputs import MetadataInput
+from ....meta.inputs import MetadataInput, MetadataInputDescription
 from ....plugins.dataloaders import get_plugin_manager_promise
 from ...mixins import AppImpersonateMixin
 from ...types import AddressInput, User
--- a/saleor/graphql/account/mutations/account/account_update.py
+++ b/saleor/graphql/account/mutations/account/account_update.py
@@ -104,14 +104,12 @@ class AccountUpdate(AddressMetadataMixin, BaseCustomerCreate, AppImpersonateMixi
             cleaned_input=cleaned_input, user_instance=instance
         )
 
-        instance.save()
-
         if default_addresses["billing"]:
             instance.addresses.add(default_addresses["billing"])
         if default_addresses["shipping"]:
             instance.addresses.add(default_addresses["shipping"])
 
         instance.search_document = prepare_user_search_document_value(instance)
-        instance.save(update_fields=["search_document", "updated_at"])
+        instance.save()
 
         cls.call_event(manager.customer_updated, instance)
--- a/saleor/graphql/account/mutations/account/account_update.py
+++ b/saleor/graphql/account/mutations/account/account_update.py
@@ -1,19 +1,13 @@
 from typing import cast
-from urllib.parse import urlencode
 
 import graphene
-from django.contrib.auth.tokens import default_token_generator
 
 from .....account import models
-from .....account.error_codes import AccountErrorCode
-from .....account.notifications import send_set_password_notification
 from .....account.search import prepare_user_search_document_value
 from .....core.tracing import traced_atomic_transaction
-from .....core.utils.url import prepare_url
 from .....permission.auth_filters import AuthorizationFilters
 from .....webhook.event_types import WebhookEventAsyncType
 from ....account.mixins import AddressMetadataMixin
-from ....channel.utils import clean_channel, validate_channel
 from ....core import ResolveInfo
 from ....core.descriptions import ADDED_IN_319
 from ....core.doc_category import DOC_CATEGORY_USERS
@@ -103,7 +97,7 @@ class AccountUpdate(AddressMetadataMixin, BaseCustomerCreate, AppImpersonateMixi
 
     @classmethod
     @traced_atomic_transaction()
-    def save(cls, info: ResolveInfo, instance, cleaned_input):
+    def save(cls, info: ResolveInfo, instance: models.User, cleaned_input):
         default_shipping_address = cleaned_input.get(SHIPPING_ADDRESS_FIELD)
         manager = get_plugin_manager_promise(info.context).get()
         if default_shipping_address:
@@ -126,27 +120,9 @@ class AccountUpdate(AddressMetadataMixin, BaseCustomerCreate, AppImpersonateMixi
         cls.call_event(manager.customer_updated, instance)
 
         if redirect_url := cleaned_input.get("redirect_url"):
-            channel_slug = cleaned_input.get("channel")
-            if not instance.is_staff:
-                channel_slug = clean_channel(
-                    channel_slug, error_class=AccountErrorCode, allow_replica=False
-                ).slug
-            elif channel_slug is not None:
-                channel_slug = validate_channel(
-                    channel_slug, error_class=AccountErrorCode
-                ).slug
-            send_set_password_notification(
-                redirect_url,
-                instance,
-                manager,
-                channel_slug,
-            )
-            token = default_token_generator.make_token(instance)
-            params = urlencode({"email": instance.email, "token": token})
-            cls.call_event(
-                manager.account_set_password_requested,
-                instance,
-                channel_slug,
-                token,
-                prepare_url(params, redirect_url),
+            cls.process_account_confirmation(
+                redirect_url=redirect_url,
+                instance=instance,
+                channel_slug_from_input=cleaned_input.get("channel"),
+                plugins_manager=manager,
             )
--- a/saleor/graphql/account/mutations/account/account_update.py
+++ b/saleor/graphql/account/mutations/account/account_update.py
@@ -104,7 +104,7 @@ class AccountUpdate(AddressMetadataMixin, BaseCustomerCreate, AppImpersonateMixi
             cleaned_input=cleaned_input, user_instance=instance
         )
 
-        super().save(info, instance, cleaned_input)
+        instance.save()
 
         if default_addresses["billing"]:
             instance.addresses.add(default_addresses["billing"])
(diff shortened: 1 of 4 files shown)


saleor/graphql/account/mutations/base.py
```diff
@@ -414,6 +359,43 @@ def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
             if user_gift_cards := get_user_gift_cards(instance):
                 mark_gift_cards_search_index_as_dirty(user_gift_cards)
 
+    @classmethod
+    def process_account_confirmation(
```
korycins, line 363: I would explicitly use name `set_password`. It is the exact action that this method is doing. `account_confirmation` can be misleading, as we already have an event like: `send_account_confirmation`. 
Additionally as this is the event triggered by staff (at least it is what I see in the code), maybe worth to add `customer` instead of `account`. `account` suggest me that I am operating on my own account (as customer's mutation to manage the account, starts with prefix `account`)

saleor/graphql/account/mutations/account/account_update.py
```diff
@@ -91,3 +94,35 @@ def perform_mutation(cls, root, info: ResolveInfo, /, **data):
...
+        if default_shipping_address:
+            instance.addresses.add(default_shipping_address)
+
+        instance.search_document = prepare_user_search_document_value(instance)
+        instance.save(update_fields=["search_document", "updated_at"])
+
+        cls.call_event(manager.customer_updated, instance)
+
+        if redirect_url := cleaned_input.get("redirect_url"):
+            cls.process_account_confirmation(
```
korycins, line 123: you don't have `redirect_url` in the input, so this part will not be called in any case.

saleor/graphql/account/mutations/staff/customer_update.py
```diff
@@ -162,3 +168,38 @@ def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
...
+            instance.addresses.add(default_billing_address)
+        if default_shipping_address:
+            instance.addresses.add(default_shipping_address)
+
+        instance.search_document = prepare_user_search_document_value(instance)
+        instance.save(update_fields=["search_document", "updated_at"])
+
+        cls.call_event(manager.customer_updated, instance)
+
+        if redirect_url := cleaned_input.get("redirect_url"):
```
korycins, line 199: You don't have `redirece_url` in the input. This part will not be called in any case.

saleor/graphql/account/mutations/staff/customer_create.py
```diff
@@ -38,3 +43,39 @@ class Meta:
...
+            instance.addresses.add(default_shipping_address)
+
+        instance.search_document = prepare_user_search_document_value(instance)
+        instance.save(update_fields=["search_document", "updated_at"])
+
+        cls.call_event(manager.customer_created, instance)
+        account_events.customer_account_created_event(user=instance)
+
+        if redirect_url := cleaned_input.get("redirect_url"):
+            cls.process_account_confirmation(
```
korycins, line 76: This is called only for `CustomerCreate`. We can fully extract it from the "Base" and place it here.

saleor/graphql/account/mutations/staff/customer_update.py
```diff
@@ -162,3 +168,38 @@ def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
...
+
+        if default_shipping_address:
+            default_shipping_address.save()
+            instance.default_shipping_address = default_shipping_address
+
+        default_billing_address = cleaned_input.get(BILLING_ADDRESS_FIELD)
+
+        if default_billing_address:
+            default_billing_address.save()
+            instance.default_billing_address = default_billing_address
```
IKarbowiak, line 186: Seems like address saving part is the same for all mutations, so maybe can be extracted to function and reused

saleor/graphql/account/mutations/account/account_update.py
```diff
@@ -91,3 +94,24 @@ def perform_mutation(cls, root, info: ResolveInfo, /, **data):
...
+    @classmethod
+    @traced_atomic_transaction()
+    def save(cls, info: ResolveInfo, instance: models.User, cleaned_input):
+        manager = get_plugin_manager_promise(info.context).get()
+
+        default_addresses = cls.save_and_apply_addresses_from_input(
+            cleaned_input=cleaned_input, user_instance=instance
+        )
+
+        super().save(info, instance, cleaned_input)
```
IKarbowiak, line 107: Do we need to call the `save` of super class? It's probably saving all object (all fields instead of that that have been changed). 
But maybe is a topic for a separate PR

saleor/graphql/account/mutations/staff/customer_create.py
```diff
@@ -38,3 +43,39 @@ class Meta:
...
+            default_shipping_address.save()
+            instance.default_shipping_address = default_shipping_address
+
+        default_billing_address = cleaned_input.get(BILLING_ADDRESS_FIELD)
+
+        if default_billing_address:
+            default_billing_address.save()
+            instance.default_billing_address = default_billing_address
+
+        super().save(info, instance, cleaned_input)
```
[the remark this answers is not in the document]
IKarbowiak: I guess that instead of calling `save` of super class we can just call `instance.save()`.

saleor/graphql/account/mutations/staff/customer_create.py
```diff
@@ -38,3 +52,70 @@ class Meta:
...
+
+        if default_addresses["billing"]:
+            instance.addresses.add(default_addresses["billing"])
+        if default_addresses["shipping"]:
+            instance.addresses.add(default_addresses["shipping"])
+
+        instance.search_document = prepare_user_search_document_value(instance)
+        instance.save(update_fields=["search_document", "updated_at"])
```
IKarbowiak, line 73: No need for saving the instance twice. Probably we can only call `instance.save()` here instead of calling it in line 65.

saleor/graphql/account/mutations/staff/customer_update.py
```diff
@@ -162,3 +166,24 @@ def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
...
+    @classmethod
+    @traced_atomic_transaction()
+    def save(cls, info: ResolveInfo, instance, cleaned_input):
+        manager = get_plugin_manager_promise(info.context).get()
+
+        default_addresses = cls.save_and_apply_addresses_from_input(
+            cleaned_input=cleaned_input, user_instance=instance
+        )
+
+        super().save(info, instance, cleaned_input)
```
IKarbowiak, line 179: Same here with saving the whole object instead of changed fields. 
Also the save is called twice, in this line and in 187 and can probably be called once

saleor/graphql/account/mutations/account/account_update.py
IKarbowiak: Also the save is called twice (line 107 and 115), and probably can be called once

saleor/graphql/account/mutations/staff/customer_create.py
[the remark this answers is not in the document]
fowczarek: `super().save` call `save` function of a parent of this class.  In this example, it should be `BaseCustomerCreate.super`. `save_and_apply_addresses_from_input` coantins duplicated code with `BaseCustomerCreate.super`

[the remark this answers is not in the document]
lkostrowski: > super().save call save function of a parent of this class. In this example, it should be BaseCustomerCreate.super

Yes but there is no "save" on BaseCustomerCreate anymore. So it should call "save" from `DeprecatedModelMutation` which calls instance.save()

@IKarbowiak you are right, I should call instance.save(), it's more readable 👍 But I think I don't execute anything twice...

> save_and_apply_addresses_from_input coantins duplicated code with BaseCustomerCreate.super

This method is defined on base class and used in child classes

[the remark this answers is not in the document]
lkostrowski: Replaced super().save() with instance.save() and removed duplicated saves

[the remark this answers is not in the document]
lkostrowski: In last commit I had to restore 2nd save() for customerCreate, because it failed on many to many relations

fowczarek (dismissed):

saleor/graphql/account/mutations/base.py
```diff
@@ -414,6 +354,31 @@ def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
             if user_gift_cards := get_user_gift_cards(instance):
                 mark_gift_cards_search_index_as_dirty(user_gift_cards)
 
+    @classmethod
+    def save_and_apply_addresses_from_input(
```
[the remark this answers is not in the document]
lkostrowski: This function is

1. Taking input (hence "from_input", it explictly needs this and we dont have typing)
2. Saves addresses (so "save")
3. adding address to instance (this is my "apply", but maybe it's not a good word)

saleor/graphql/account/mutations/staff/customer_create.py
```diff
@@ -38,3 +52,70 @@ class Meta:
...
+    @classmethod
+    @traced_atomic_transaction()
+    def save(cls, info: ResolveInfo, instance, cleaned_input):
+        manager = get_plugin_manager_promise(info.context).get()
+
+        default_addresses = cls.save_and_apply_addresses_from_input(
+            cleaned_input=cleaned_input, user_instance=instance
+        )
+
+        instance.save()
```
IKarbowiak, line 65: Do we need tis save at all? It seems like save from line 74 should be enough

saleor/graphql/account/mutations/staff/customer_update.py
```diff
@@ -162,3 +166,22 @@ def perform_mutation(cls, _root, info: ResolveInfo, /, **data):
...
+            cleaned_input=cleaned_input, user_instance=instance
+        )
+
+        if default_addresses["billing"]:
+            instance.addresses.add(default_addresses["billing"])
+        if default_addresses["shipping"]:
+            instance.addresses.add(default_addresses["shipping"])
+
+        instance.search_document = prepare_user_search_document_value(instance)
+        instance.save()
```
IKarbowiak, line 185: Could we save only updated fields instead of the whole instance? If it will extend the scope too much, we can create a ticket for that

saleor/graphql/account/mutations/account/account_update.py
```diff
@@ -91,3 +94,22 @@ def perform_mutation(cls, root, info: ResolveInfo, /, **data):
...
+            cleaned_input=cleaned_input, user_instance=instance
+        )
+
+        if default_addresses["billing"]:
+            instance.addresses.add(default_addresses["billing"])
+        if default_addresses["shipping"]:
+            instance.addresses.add(default_addresses["shipping"])
+
+        instance.search_document = prepare_user_search_document_value(instance)
+        instance.save()
```
IKarbowiak, line 113: Could we save only updated fields instead of the whole instance? If it will extend the scope too much, we can create a ticket for that

saleor/graphql/account/mutations/staff/customer_create.py
```diff
@@ -38,3 +52,70 @@ class Meta:
...
+
+        if default_addresses["billing"]:
+            instance.addresses.add(default_addresses["billing"])
+        if default_addresses["shipping"]:
+            instance.addresses.add(default_addresses["shipping"])
+
+        instance.search_document = prepare_user_search_document_value(instance)
+        instance.save()
```
korycins, line 73:
```suggestion
        instance.save(update_fields=["search_document"])
```

saleor/graphql/account/mutations/staff/customer_create.py
```diff
@@ -38,3 +52,70 @@ class Meta:
...
+    def save(cls, info: ResolveInfo, instance, cleaned_input):
+        manager = get_plugin_manager_promise(info.context).get()
+
+        default_addresses = cls.save_and_apply_addresses_from_input(
+            cleaned_input=cleaned_input, user_instance=instance
+        )
+
+        instance.save()
+
+        if default_addresses["billing"]:
```
korycins, line 67: I am not conviced that current function separation is correct. In `save_and_apply_addresses_from_input` you have  one part of saving process:
- `user_instance.default_shipping_address`

Then as a "post" function `save_and_apply_addresses_from_input` calls, you always have `if`s like:
```python
if default_addresses["billing"]:
    instance.addresses.add(default_addresses["billing"])
```

Calling any of these logic separately, can cause an unexpected results. IMHO  adding to `addresses` & `instance.default_address=` should be wrapped in single function as this is atomic assignment.

saleor/graphql/account/mutations/staff/customer_create.py
lkostrowski: It fails without both of them, because (according to my understanding) some objects must be saved before m2m are saved

saleor/graphql/account/mutations/staff/customer_update.py
lkostrowski: I think it's not an easy way to deduct that quickly and it's not the scope of my refactor (I'm breaking inheritance, not optimizing writes) - so I will create a ticket

lkostrowski: https://linear.app/saleor/issue/EXT-2149/optimize-save-operations-in-customeraccount-createupdate

saleor/graphql/account/mutations/staff/customer_create.py
lkostrowski: done

fowczarek (approved):

saleor/graphql/account/mutations/base.py
```diff
@@ -414,6 +354,34 @@ def post_save_action(cls, info: ResolveInfo, instance, cleaned_input):
...
+        default_billing_address: models.Address | None = cleaned_input.get(
+            BILLING_ADDRESS_FIELD
+        )
+
+        if default_billing_address:
+            default_billing_address.save()
+            user_instance.default_billing_address = default_billing_address
+
+        if save_user:
+            user_instance.save()
```
IKarbowiak, line 378: I think this save shouldn't be here. I was looking for save in `CustomerCreate` and I won't expect that the user will be saved in `save_default_addresses` method.

IKarbowiak (approved):

saleor/graphql/account/mutations/base.py
korycins: I think that this is for handling the case, when you need to save addresses on the instance without the `pk`. 
As addresses are palced in `default_...address` fields, and as a foreign relation: `addresses`. You need to have user represantation on the database side, to perfomr proper address save

PR #17541, opened 2025-03-27
Introduce static typing for shipping sync webhook responses

IKarbowiak:
- Add `pydantic` package
- Introduce static typing for shipping sync webhook responses:
  - `SHIPPING_LIST_METHODS_FOR_CHECKOUT`
  - `CHECKOUT_FILTER_SHIPPING_METHODS`
  - `ORDER_FILTER_SHIPPING_METHODS`

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/webhook/transport/shipping_helpers.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def to_shipping_app_id(app: App, shipping_method_id: str) -> str:
+    app_identifier = app.identifier or app.id
+    return base64.b64encode(
+        str.encode(f"{APP_ID_PREFIX}:{app_identifier}:{shipping_method_id}")
+    ).decode("utf-8")
+
+
+def convert_to_app_id_with_identifier(shipping_app_id: str) -> None | str:
```
lkostrowski, line 14: looks like these are not tested

saleor/plugins/webhook/tests/test_shipping_webhook.py
```diff
@@ -422,8 +425,9 @@ def test_parse_excluded_shipping_methods_response(app):
 
     # then
     assert len(excluded_methods) == 2
-    assert excluded_methods[0]["id"] == "2"
-    assert excluded_methods[1]["id"] == external_id
+    assert excluded_methods[0].id == "2"
+    assert excluded_methods[1].id == external_id
+    assert mocked_schema_logger.call_count == 3
```
lkostrowski, line 430: we really want to log these? 🤔

saleor/webhook/response_schemas.py
```diff
@@ -0,0 +1,87 @@
...
+        return True
+
+    @property
+    def price(self) -> Money:
+        return Money(self.amount, self.currency)
+
+    def get_shipping_method_data(self, app: "App"):
+        # metadata might be `None` in case it was not provided in the input
+        # in such case the `field_validator` is not called and the self.metadata
+        # is `None` instead of an empty dict
```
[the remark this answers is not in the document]
IKarbowiak: In such case the mypy is crying.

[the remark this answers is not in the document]
IKarbowiak: Like I would like to say that `metadata` can be `None` or `dict`, in case I'm providing default value with `{}` is saying that is not true, as the value is always `dict`

saleor/webhook/transport/shipping_helpers.py
IKarbowiak: I copied it from other place.

saleor/plugins/webhook/tests/test_shipping_webhook.py
IKarbowiak: There are logs for invalid entries, and it was logged before, I think that we can provide such logs to customers in case the response data are wrong, and we are skipping them

saleor/webhook/response_schemas.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class ShippingMethodSchema(BaseModel):
+    id: str | int
+    name: str
+    amount: Decimal
+    currency: str
+    maximum_delivery_days: int | None = None
+    minimum_delivery_days: int | None = None
+    description: str | None = None
```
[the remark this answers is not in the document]
IKarbowiak: Basically we can match the database limits, but those values are not saved in DB, so there is no limit

saleor/plugins/webhook/tests/test_shipping_webhook.py
lkostrowski: If it was logged before, we can leave it to avoid unrelated scope. But I think this would be eventually removed and we should use observability mechanism for that (like metric `saleor.webhook.shipping.errors`)

saleor/webhook/transport/shipping_helpers.py
lkostrowski: Adding tests should be quick here 🙏

saleor/webhook/transport/shipping.py
```diff
@@ -1,106 +1,39 @@
...
-
-def validate_shipping_method_data(shipping_method_data):
-    if not isinstance(shipping_method_data, dict):
-        return False
-    keys = ["id", "name", "amount", "currency"]
-    return all(key in shipping_method_data for key in keys)
+    valid_methods = []
+    for method_data in response_data:
+        try:
+            shipping_method_schema = ShippingMethodSchema.model_validate(method_data)
```
lkostrowski, line 31: nitpick, but I think naming is wrong. `ShippingMethodSchema` doesn't return schema. The class itself is schema. What is returned is actual model (not django model, but some sort of DTO) or "payload" or something like this

saleor/webhook/response_schemas.py
```diff
@@ -0,0 +1,87 @@
...
+    @field_validator("metadata", mode="before")
+    @classmethod
+    def clean_metadata(cls, value: Any) -> dict[str, str]:
+        metadata = value
+        if metadata:
+            metadata = metadata if cls.method_metadata_is_valid(metadata) else {}
+        return metadata or {}
+
+    @classmethod
+    def method_metadata_is_valid(cls, metadata: Any) -> bool:
```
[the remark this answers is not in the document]
IKarbowiak: Done

saleor/webhook/response_schemas.py
```diff
@@ -0,0 +1,87 @@
...
+from .transport.shipping_helpers import to_shipping_app_id
+
+logger = logging.getLogger(__name__)
+
+
+class ShippingMethodSchema(BaseModel):
+    id: str | int
+    name: str
+    amount: Decimal
+    currency: str
```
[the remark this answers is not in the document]
IKarbowiak: Done

saleor/webhook/response_schemas.py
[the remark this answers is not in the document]
IKarbowiak: Resolved

saleor/core/utils/metadata_manager.py
```diff
@@ -73,3 +74,12 @@ def create_from_graphql_input(
     return MetadataItemCollection(
         [MetadataItem(item.key, item.value) for item in items]
     )
+
+
+def method_metadata_is_valid(metadata: Any) -> bool:
```
lkostrowski, line 79: missing test

saleor/plugins/webhook/tests/test_shipping_webhook.py
```diff
@@ -1275,6 +1284,29 @@ def test_parse_list_shipping_methods_with_metadata(app):
...
+            "currency": "ABC",
+            "name": "shipping",
+            "description": "Description",
+            "maximum_delivery_days": 10,
+            "minimum_delivery_days": 2,
+            "metadata": {"field": "value"},
+        }
+    ]
+    # when
+    response = parse_list_shipping_methods_response(response_data_with_meta, app)
```
lkostrowski, line 1303: Opinion:

I'm thinking if `parse_list_shipping_methods_response` shouldn't actually raise? I know it may be only a syntax here if we check against "None", but I think raise+except is more readable that we are processing some invalid state

```
try:
  parse_list_shipping_methods_response(response_data_with_meta, app)

except ValidationError:
  logger.warning(...)
  ...

```

saleor/core/utils/metadata_manager.py
```diff
@@ -73,3 +74,12 @@ def create_from_graphql_input(
...
+    if not isinstance(metadata, dict):
+        return False
+    for key, value in metadata.items():
+        if not isinstance(key, str) or not isinstance(value, str) or not key.strip():
```
fowczarek, line 83: I'm not sure that one invalid metadata item should cause that whole metadata to be saved as `{}`?

fowczarek, line 83: Are you sure that value is always `str`? What happens if someone sends any numeric value?

saleor/webhook/response_schemas.py
```diff
@@ -0,0 +1,79 @@
...
+    amount: Decimal = Field(..., ge=0)
+    currency: CurrencyEnum
+    maximum_delivery_days: int | None = Field(None, ge=0)
+    minimum_delivery_days: int | None = Field(None, ge=0)
+    description: str | None = None
+    metadata: dict[str, str] | None = {}
+
+    @field_validator("metadata", mode="before")
+    @classmethod
+    def clean_metadata(cls, value: Any) -> dict[str, str]:
```
lkostrowski, line 35: is value really Any? Isn't it `dict[str, str] | None`?

saleor/webhook/response_schemas.py
```diff
@@ -0,0 +1,79 @@
...
+        metadata = value
+        if metadata:
+            metadata = metadata if method_metadata_is_valid(metadata) else {}
+        return metadata or {}
+
+    @property
+    def price(self) -> Money:
+        return Money(self.amount, self.currency)
+
+    def get_shipping_method_data(self, app: "App"):
```
lkostrowski, line 45: question: why you use @property once and method second time?

I assume you need to pass "app" to the method which can't be done in property getter?

Is there any difference if you use `get_price()` instead?

(It's not a change request, I'm learning python now ;) )

saleor/webhook/response_schemas.py
```diff
@@ -0,0 +1,79 @@
...
+            maximum_delivery_days=self.maximum_delivery_days,
+            minimum_delivery_days=self.minimum_delivery_days,
+            description=self.description,
+            metadata=metadata,
+        )
+
+
+class ExcludedShippingMethodSchema(BaseModel):
+    id: str
+    reason: str | None = ""
```
lkostrowski, line 63: Looks like response is actually different
https://docs.saleor.io/developer/extending/webhooks/synchronous-events/shipping#response-format-1

I guess its fine to have class for single method, but IMO validator should be applied on entire response object

like this

```
excluded_methods = list[ExcludedShippingMethodSchema]
```

response can have typo in `excluded_methods` etc. So we should have single layer that validates entire shape, not only nested shape

saleor/webhook/transport/shipping.py
```diff
@@ -209,36 +142,27 @@ def get_excluded_shipping_data(
 
 def get_excluded_shipping_methods_from_response(
     response_data: dict,
-) -> list[dict]:
+) -> list[ExcludedShippingMethodSchema]:
     excluded_methods = []
     for method_data in response_data.get("excluded_methods", []):
```
lkostrowski, line 147: as in previous comment, this part "leaks" validation - nested part is checked in pydantic, but wrapping key here

lkostrowski (changes requested):

saleor/webhook/response_schemas.py
IKarbowiak: The validation is run before the pydantic validation, so at this moment basically anything can be here

saleor/core/utils/metadata_manager.py
IKarbowiak: This part is just moved from other place. But I will add tests

saleor/plugins/webhook/tests/test_shipping_webhook.py
IKarbowiak: I don't understand you concern. 
The `parse_list_shipping_methods_response` is logging the warnings, to not break the previous behaviour.

saleor/webhook/response_schemas.py
IKarbowiak: You can define `get_price()` as well. Only difference is that with property you can use it like a instance field `self.price` in case of method you need to call it to get the response `self.get_price()`

saleor/core/utils/metadata_manager.py
IKarbowiak: I just moved the method to different place. Prepared a ticket to address the concerns:
https://linear.app/saleor/issue/EXT-2110

IKarbowiak: I just moved the method to different place. Prepared a ticket to address the concerns:
https://linear.app/saleor/issue/EXT-2110__

saleor/core/utils/metadata_manager.py
IKarbowiak: done

lkostrowski (approved):

fowczarek (approved):

lkostrowski (approved):

saleor/webhook/response_schemas.py
```diff
@@ -0,0 +1,138 @@
...
+
+        # in case the data are not list, handle validation by pydantic
+        if not isinstance(value, list):
+            return value
+
+        excluded_methods = []
+        for method_data in value:
+            try:
+                excluded_methods.append(
+                    ExcludedShippingMethodSchema.model_validate(method_data)
```
korycins, line 133: You should be able to fully drop the validation by adding `OnErrorOmit`. It will skip the incorrect entries. 

```
root: list[OnErrorOmit[ExcludedShippingMethodSchema]]
```

saleor/core/utils/metadata_manager.py
korycins, line 79: What is `method`? 🤔  It looks like after moving to different file, the function lost the context. Maybe worth to add it 🤔

saleor/webhook/transport/shipping.py
```diff
@@ -1,106 +1,43 @@
...
-                id=to_shipping_app_id(app, method_id),
-                name=method_name,
-                price=Money(method_amount, method_currency),
-                maximum_delivery_days=method_maximum_delivery_days,
-                minimum_delivery_days=method_minimum_delivery_days,
-                description=method_description,
-                metadata=method_metadata,
-            )
+    try:
+        list_shipping_method_model = ListShippingMethodsSchema.model_validate(
```
korycins, line 34: Is there any reason why you use `model_validate` explicitly? 
Personally I would just call `ListShippingMethodsSchema(response_data)` as this would be more friendly from me (but maybe I am missing something 😓 )

saleor/webhook/transport/shipping.py
```diff
@@ -209,36 +146,31 @@ def get_excluded_shipping_data(
...
-                )
-                continue
-
-        except (KeyError, ValueError, TypeError, GraphQLError) as e:
-            logger.warning("Malformed ShippingMethod id was provided: %s", e)
-            continue
-        excluded_methods.append(
-            {"id": method_id, "reason": method_data.get("reason", "")}
+    try:
+        excluded_methods.extend(
```
korycins, line 153: [Sugestions]
Maybe worth to split it into two lines? Right now it looks like a lot of things happens here 😅 

```python
        parsed_app_response = FilterShippingMethodsSchema(response_data) 
        excluded_methods.extend(
            parsed_app_response.excluded_methods
        )
```

saleor/webhook/response_schemas.py
```diff
@@ -0,0 +1,138 @@
...
+
+        methods = []
+        for method_data in value:
+            try:
+                methods.append(ShippingMethodSchema.model_validate(method_data))
+            except ValidationError as e:
+                logger.warning("Skipping invalid shipping method: %s", e)
+        return methods
+
+    def get_shipping_methods_data(self, app: "App") -> list[ShippingMethodData]:
```
korycins, line 82: IMHO I am not fan of `get_shipping_methods_data`, `get_shipping_method_data`. The PydanticModels responsible for data validating received from app, has also embedded logic responsible for converting the pydantic model into our internal dataclass. 
Personally I would recommend to keep pydantic models as flat as possible. The conversion to our internal structures is not part of their job, I would suggest something similar to what you already added (helper function accepts pydantic -> converts it to our dataclass):

```python
def parse_excluded_shipping_methods(
    excluded_methods: list[ExcludedShippingMethodSchema],
) -> dict[str, list[ExcludedShippingMethod]]:
```

With this the layers are separated. Pydantic doesn't know anything about our internal representation of shipping methods. And our dataclassases doesn't know anything about the pydantic.

saleor/webhook/response_schemas.py
```diff
@@ -0,0 +1,138 @@
...
+logger = logging.getLogger(__name__)
+
+name_max_length = ShippingMethod._meta.get_field("name").max_length
+
+
+class ShippingMethodSchema(BaseModel):
+    id: str | int
+    name: str = Field(..., max_length=name_max_length)
+    amount: Decimal = Field(..., ge=0)
+    currency: CurrencyEnum
```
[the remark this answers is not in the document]
IKarbowiak: Will be added separately: https://linear.app/saleor/issue/EXT-2140

saleor/webhook/response_schemas.py
IKarbowiak: Not sure if we want to use it, as we will not able to log the issues for customers. So IMO, we should rather log that the response is missing instead of totally omitting the errors

saleor/core/utils/metadata_manager.py
IKarbowiak: changed to `metadata_is_valid`

saleor/webhook/transport/shipping.py
IKarbowiak: Done

saleor/webhook/response_schemas.py
IKarbowiak: I removed those methods from `schema_responses`

saleor/webhook/transport/tests/test_response_schemas.py
```diff
@@ -0,0 +1,352 @@
...
+
+
+@pytest.mark.parametrize(
+    "data",
+    [
+        # All fields provided
+        {
+            "id": "1",
+            "name": "Standard Shipping",
+            "amount": Decimal("10.00"),
```
przlada, line 29: Maybe we should extend the tests with input data of different types for `amount: Decimal = Field(ge=0)` field? `int`, `float` and `str` should be supported

saleor/webhook/response_schemas.py
```diff
@@ -0,0 +1,115 @@
...
+            try:
+                methods.append(ShippingMethodSchema.model_validate(method_data))
+            except ValidationError as e:
+                logger.warning("Skipping invalid shipping method: %s", e)
+        return methods
+
+
+class ExcludedShippingMethodSchema(BaseModel):
+    id: str
+    reason: str | None = ""
```
przlada, line 73:
```suggestion
    reason: DefaultIfNone[str] = ""
```
I am not aware if later we make use of distinction between empty string and `None` set as the reason value, which is totally fine, but if `None` is added as accepted type only for backward compatibility, I propose to make use of something like `DefaultIfNone` introduced in other comment

saleor/webhook/transport/shipping.py
przlada: Using `model_validate()` is the recommended way, as it allows to pass additional validation parameters like `by_alias` or `context`. We should consider using `model_validate_json` and passing raw JSON directly to it as it is [recommended performance-wise](https://docs.pydantic.dev/2.11/concepts/performance/#in-general-use-model_validate_json-not-model_validatejsonloads)

saleor/webhook/transport/shipping.py
```diff
@@ -1,106 +1,55 @@
...
-    if not isinstance(shipping_method_data, dict):
-        return False
-    keys = ["id", "name", "amount", "currency"]
-    return all(key in shipping_method_data for key in keys)
+    except ValidationError:
+        logger.warning("Skipping invalid shipping method response: %s", response_data)
+        return []
+    return [
+        ShippingMethodData(
+            id=to_shipping_app_id(app, str(shipping_method.id)),
```
przlada, line 43: This approach is totally fine, but I was wondering if conversion `to_shipping_app_id` shouldn't be a part of Pydantic model as shipping method should be assigned to only one specific app. This is only an idea to consider, but that can potentially be achieved using [Validation context](https://docs.pydantic.dev/latest/concepts/validators/#validation-context)
We can add something like this to the model:
```python
class ShippingMethodSchema(BaseModel):  
    ...
    @field_validator("id", mode="after")  
    @classmethod  
    def clean_id(cls, shipping_method_id: str, info: ValidationInfo) -> str:  
        app: App | None = info.context.get("app", None) if info.context else None  
        if not app:  
            raise RuntimeError("Missing app in context")  
        return to_shipping_app_id(app, shipping_method_id)
```
And then set app object in the validation context this way:
```python
ShippingMethodSchema.model_validate(data, context={"app": app})
```
Or through the parent:
```python
ListShippingMethodsSchema.model_validate(data, context={"app": app})
```

saleor/webhook/response_schemas.py
korycins: > Not sure if we want to use it, as we will not able to log the issues for customers. So IMO, we should rather log that the response is missing instead of totally omitting the errors

I proposed it here, as I saw that you don't make any logging here - but now I see that you're added something :)

saleor/webhook/transport/shipping.py
IKarbowiak: As I'm using `ListShippingMethodsSchema.model_validate` I would like provide the context there, but the context is in such case not available in child elements, I would have to propagate the context ot child so I decided to not change it for know.

saleor/webhook/response_schemas/annotations.py
```diff
@@ -0,0 +1,27 @@
...
+)
+from pydantic_core import PydanticUseDefault
+
+from ...core.utils.metadata_manager import metadata_is_valid
+
+
+def skip_invalid_metadata(value: Any) -> Any:
+    if not metadata_is_valid(value):
+        raise PydanticUseDefault()
+    return value
```
lkostrowski, line 14: this is not correct typing, instead can be type safe with generic, try this:

```python
M = TypeVar("M")


def skip_invalid_metadata(value: M) -> M:
    if not metadata_is_valid(value):
        raise PydanticUseDefault()
    return value
```

lkostrowski (approved):

korycins (approved):

fowczarek (approved):

saleor/webhook/response_schemas/annotations.py
IKarbowiak: I will take care of that into separate PR

lkostrowski: Ok. I'd also recommend to use reasoanble names for type vars, I think "T" is only for example (like `(for i)` ) but is not actually something readable. Maybe it's better to call it with semantics like normal variables, e.g. `MaybeMetadata` or something like this

PR #17545, opened 2025-03-27
Process async webhooks with scheduler

zedzior:
I want to merge this change because it add celery beat schedule for processing async webhooks in batches per app.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


saleor/webhook/tasks.py
```diff
@@ -0,0 +1,23 @@
...
+from ..core import EventDeliveryStatus
+from ..core.models import EventDelivery
+from .transport.asynchronous.transport import send_webhooks_async_for_app
+
+
+@app.task
+def process_async_webhooks_task():
+    app_ids = get_app_ids_with_pending_deliveries()
+    for app_id in app_ids:
+        send_webhooks_async_for_app(app_id=app_id)
```
kadewu, line 12: You need to `apply_assync`
e.g:
```
send_webhooks_async_for_app.apply_async(
    kwargs={"app_id": delivery.webhook.app_id},
    bind=True,
)
```

saleor/webhook/tasks.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def get_app_ids_with_pending_deliveries() -> list[int]:
+    app_ids = (
+        EventDelivery.objects.select_related("webhook")
```
kadewu, line 17: I think we should use subquery in where condition, to avoid join here.

saleor/webhook/tests/test_tasks.py
```diff
@@ -1149,3 +1152,57 @@ def test_handle_transaction_request_task_request_event_included_in_calculations(
...
+
+
+@mock.patch("saleor.webhook.tasks.send_webhooks_async_for_app")
+def test_process_async_webhooks_task(
+    mock_send_webhooks_for_app, app_with_webhook, event_payload
+):
+    # given
+    app_1, app_2, app_3, app_4 = App.objects.bulk_create(
+        [
+            App(name="App_1", is_active=True),
```
kadewu, line 1164: What if app is inactive?

zedzior: I think we should process webhooks no matter the app is active or not. Otherwise we can end up with unprocessed webhooks in the queue in case someone deactivate the app just after generating some webhook. What do you think ?

zedzior: I discussed it with a product team - you are right, we should only process webhooks for active apps. Undelivered webhooks should wait for the app to be re-enabled or uninstalled

saleor/webhook/tasks.py
```diff
@@ -0,0 +1,38 @@
...
+                EventDelivery.objects.filter(status=EventDeliveryStatus.PENDING).filter(
+                    webhook_id=OuterRef("pk")
+                )
+            )
+        )
+        .filter(
+            Exists(App.objects.filter(is_active=True).filter(id=OuterRef("app_id")))
+        )
+        .values_list("app_id", flat=True)
+        .distinct()
```
tomaszszymanski129, line 36: is the distinct really needed here? It often brings much worse performance compared to exists, plus there are no explicit joins?

saleor/webhook/tasks.py
```diff
@@ -0,0 +1,38 @@
...
+    app_ids = (
+        Webhook.objects.using(settings.DATABASE_CONNECTION_REPLICA_NAME)
+        .filter(
+            Exists(
+                EventDelivery.objects.filter(status=EventDeliveryStatus.PENDING).filter(
+                    webhook_id=OuterRef("pk")
+                )
+            )
+        )
+        .filter(
```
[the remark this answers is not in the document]
zedzior: We will test the performance of the query during load tests of the whole solution

aniav:
@maarcingebala can you make a decision on what's the future of this PR? :)

PR #17552, opened 2025-03-31
Improve checkout lines metadata overwriting conditions

lkostrowski:
- it fixes invalid condition that would overwrite metadata in some cases
- adds extra tests for that

This behavior didn't really trigger the issue, because in another part of code it was "healed"

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/graphql/checkout/mutations/utils.py
+++ b/saleor/graphql/checkout/mutations/utils.py
@@ -2,7 +2,7 @@ import datetime
 import uuid
 from collections import defaultdict
 from collections.abc import Iterable
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from decimal import Decimal
 from typing import TYPE_CHECKING, Any, cast
 
@@ -68,7 +68,7 @@ class CheckoutLineData:
     quantity_to_update: bool = False
     custom_price: Decimal | None = None
     custom_price_to_update: bool = False
-    metadata_list: list | None = None
+    metadata_list: list = field(default_factory=list)
 
 
 def clean_delivery_method(
@@ -376,7 +376,7 @@ def group_lines_input_on_add(
     for line in lines:
         variant_id = cast(str, line.get("variant_id"))
         force_new_line = line.get("force_new_line")
-        metadata_list_from_input = line.get("metadata")
+        metadata_list_from_input = line.get("metadata", [])
 
         _, variant_db_id = graphene.Node.from_global_id(variant_id)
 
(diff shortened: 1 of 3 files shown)


saleor/graphql/checkout/mutations/utils.py
```diff
@@ -396,23 +396,23 @@ def group_lines_input_on_add(
...
                     line_data.variant_id = find_variant_id_when_line_parameter_used(
                         line_db_id, existing_lines_info
                     )
 
-                    if line_data.metadata_list and metadata_list:
-                        line_data.metadata_list += metadata_list
-                    else:
-                        line_data.metadata_list = metadata_list
+                    if metadata_list_from_input:
+                        line_data.metadata_list = line_data.metadata_list or []
```
IKarbowiak, line 408: Maybe instead of doing this, the default value of `CheckoutLineData` should be changed to `list`? Currently it's `None`, and I suppose it's the root of the issue

saleor/graphql/checkout/tests/mutations/test_checkout_lines_update.py
```diff
@@ -1809,3 +1809,109 @@ def test_checkout_lines_update_with_invalid_metadata(
...
+    variant_id = graphene.Node.to_global_id("ProductVariant", variant.pk)
+
+    # when
+    variables = {
+        "id": to_global_id_or_none(checkout_with_item),
+        "lines": [
+            {
+                "variantId": variant_id,
+                "quantity": 1,
+                # Leave input empty to ensure it will not affect existing entries
```
fowczarek, line 1889: Test name, and this comment says that passed metadata should be empty.

saleor/graphql/checkout/mutations/utils.py
lkostrowski: I'd prefer to avoid another refactoring - it's working. We have introduced MetadataCollection dataclass that should be used to persist and operate on metadata - IMO if we refactor this, it should be to this dataclass

And it will also encapsulate such behavior, eg. `m = MetadataCollection(...); m.appendItems(...)`

saleor/graphql/checkout/tests/mutations/test_checkout_lines_update.py
lkostrowski: invalid test name, it should merge old and new

previous test in this file is merging empty. updated

fowczarek (dismissed):

saleor/graphql/checkout/mutations/utils.py
lkostrowski: done

IKarbowiak (approved):

fowczarek (approved):

PR #17562, opened 2025-04-02
Introduce static typing for tax sync webhook responses

IKarbowiak:
- Introduce static typing for shipping sync webhook responses:
  - `ORDER_CALCULATE_TAXES`
  - `CHECKOUT_CALCULATE_TAXES`

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist

### DIFF
--- a/saleor/core/utils/__init__.py
+++ b/saleor/core/utils/__init__.py
@@ -153,3 +153,14 @@ def prepare_unique_attribute_value_slug(attribute: "Attribute", slug: str):
         "slug", flat=True
     )
     return prepare_unique_slug(slug, value_slugs)
+
+
+def truncate_msg(message: str, max_length: int) -> str:
+    """Truncate a message to a specified maximum length.
+
+    If the message exceeds the maximum length, it will be truncated
+    and '...' will be appended to indicate truncation.
+    """
+    if len(message) > max_length:
+        return f"{message[: max_length - 3]}..."
+    return message
--- a/saleor/core/utils/__init__.py
+++ b/saleor/core/utils/__init__.py
@@ -153,14 +153,3 @@ def prepare_unique_attribute_value_slug(attribute: "Attribute", slug: str):
         "slug", flat=True
     )
     return prepare_unique_slug(slug, value_slugs)
-
-
-def truncate_msg(message: str, max_length: int) -> str:
-    """Truncate a message to a specified maximum length.
-
-    If the message exceeds the maximum length, it will be truncated
-    and '...' will be appended to indicate truncation.
-    """
-    if len(message) > max_length:
-        return f"{message[: max_length - 3]}..."
-    return message
--- a/saleor/plugins/webhook/plugin.py
+++ b/saleor/plugins/webhook/plugin.py
@@ -18,7 +18,7 @@ from ...core import EventDeliveryStatus
 from ...core.models import EventDelivery
 from ...core.notify import NotifyEventType
 from ...core.taxes import TaxData, TaxDataError, TaxType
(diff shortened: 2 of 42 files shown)


saleor/core/utils/__init__.py
```diff
@@ -153,3 +153,14 @@ def prepare_unique_attribute_value_slug(attribute: "Attribute", slug: str):
         "slug", flat=True
     )
     return prepare_unique_slug(slug, value_slugs)
+
+
+def truncate_msg(message: str, max_length: int) -> str:
```
fowczarek, line 158: Should we use `truncatechars` instead of creating a new function for truncating?

fowczarek (approved):

saleor/plugins/webhook/plugin.py
```diff
@@ -159,6 +162,10 @@
 # (FedEx, UPS, TNT, DHL).
 CACHE_TIME_SHIPPING_LIST_METHODS_FOR_CHECKOUT: Final[int] = 3600 * 12
 
+MAX_TAX_ERROR_LGTH: int = min(  # type: ignore[type-var, assignment]
```
korycins, line 165: It might be worth "reversing" how we define the tax error length variable. Instead of fetching it dynamically with something like `Order._meta.get_field("tax_error").max_length`, let's use a constant value to define the length for both error model fields. For example:

```python
TAX_ERROR_FIELD_LENGTH = 255
...
Checkout:
    tax_error = models.CharField(max_length=TAX_ERROR_FIELD_LENGTH, null=True, blank=True)
...
Order:
    tax_error = models.CharField(max_length=TAX_ERROR_FIELD_LENGTH, null=True, blank=True)
```

Then, in your code, you can use `TAX_ERROR_FIELD_LENGTH` to limit the input.

---

In my opinion, this approach is more explicit because:
- By looking at the model, you can easily see that both fields rely on the same length.
- It "connects" the Order and Checkout field lengths. When creating an order, you'll have `tax_error=checkout_info.checkout.tax_error`, so the length should definitely be the same.

WDYT?

saleor/plugins/webhook/tests/test_tax_webhook.py
korycins, line 1: Question: do we have somewhere tests for case when `response_data=None`? 
Not sure if I miss them, or we don't have them. 

If missing, maybe worth to add them, as having `response_data` is common case - lack of response from app, http_error from app etc

saleor/webhook/response_schemas/taxes.py
```diff
@@ -0,0 +1,36 @@
...
+
+class CalculateTaxesSchema(BaseModel):
+    shipping_tax_rate: Annotated[Decimal, Field(ge=0, le=100)]
+    shipping_price_gross_amount: Annotated[Decimal, Field(ge=0, le=MAXIMUM_PRICE)]
+    shipping_price_net_amount: Annotated[Decimal, Field(ge=0, le=MAXIMUM_PRICE)]
+    lines: list[LineCalculateTaxesSchema] = []
+
+    @field_validator("lines")
+    @classmethod
+    def validate_lines_length(
```
korycins, line 23: It is only suggestion, but as for `ExcludedShippingMethodSchema` we have `clean_id` for field_validator, then maybe worth to keep the same naming here?  - like `clena_lines` or `clean_lines_length`, or `ExcludedShippingMethodSchema.validate_id`

fowczarek (dismissed):

fowczarek (approved):

korycins (approved):

PR #17563, opened 2025-04-03
Allow to change Admin email plugin custom templates back to default

wcislo-saleor:
It has to be backported to previous versions.

Changes in unrelated tests were necessary due to the way fixtures are written and used (`admin_email_plugin` fixture upon its use deletes template created by `admin_email_template` fixture).


kadewu (approved):

korycins (approved):
Looks good! 💪 
Question: Maybe worth to also check `UserEmailPlugin`? There is a possibility that the same issue is also there.

PR #17567, opened 2025-04-03
Drop `django-prices`

patrys:
In preparation for upgrading to Django 5, stop depending on `django-prices`. We hardly use any of the Django capabilities of that package and it's not an easy task to make it compatible with Django 5 and Python 3.13. Instead of trying to fix the package, add our own virtual fields for Money and TaxedMoney and drop the dependency.

# Impact

- [ ] New migrations

# Docs

- [ ] Link to documentation:

# Pull Request Checklist


maarcingebala (approved):

patrys:
I have no idea why `poetry lock` gives a different result on CI than on my local machine. It should not depend on the architecture, yet it seems it somehow does.

maarcingebala:
@patrys Try clearing the cache and generating the lock again locally:
```
poetry cache clear --all pypi
```
I didn't debug this issue, but we already saw it a couple of times, and the above helped for now.

korycins (approved):

